PR #1, opened 2015-12-15
Implement auth

minglis:
First pass at implementing API authentication using new JWT tokens …
- NOTE - this does not manage secrets. There is only one URL and there is no functionality implemented
- prior to rolling out we need to store secrets properly

Uses the JWT libraries in [https://github.com/alphagov/notifications-python-client](https://github.com/alphagov/notifications-python-client)

- Tokens are checked on every request and will be rejected if token is invalid as per the rules in the python clients.


app/main/authentication/auth.py
```diff
@@ -0,0 +1,51 @@
...
+    auth_scheme = auth_header[:7]
+
+    if auth_scheme != 'Bearer ':
+        return authentication_response('Unauthorized, authentication bearer scheme must be used', 401)
+
+    try:
+        auth_token = auth_header[7:]
+        api_client = fetch_client(get_token_issuer(auth_token))
+
+        if api_client is None:
```
robyoung: This cannot be true, `fetch_client` always returns a dictionary.

app/main/authentication/auth.py
```diff
@@ -0,0 +1,51 @@
+from flask import request, jsonify, _request_ctx_stack
+from client.jwt import decode_jwt_token, get_token_issuer
```
robyoung: Generic names for top level packages can become a pain if you have a clash.

app/main/__init__.py
```diff
@@ -1,7 +1,15 @@
...
+AUTHORIZATION_SCHEME = 'Bearer'
+WINDOW = 1
 
 main = Blueprint('main', __name__)
 
 
+main.before_request(requires_auth)
+
+
 from .views import notifications, index
```
servingUpAces: Can we use absolute imports, please?

app/main/errors.py
```diff
@@ -0,0 +1,36 @@
+from flask import jsonify
+
+from . import main
```
servingUpAces: Use absolute path for the import

servingUpAces:
Do we want to build this on travis?

PR #4, opened 2016-01-12
Service creation

mannickutd:
Lots of changes, so please take 2 minutes to go through all changes.


app/models.py
```diff
@@ -0,0 +1,61 @@
+from . import db
+import datetime
+
+
```
mannickutd: This can probably be removed now as serialise and deserialise will be done in the schema class.

migrations/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
mannickutd: Need to improve on this.

quis:
I wouldn’t worry about it for this pull request, but worth reading https://github.com/alphagov/styleguides/blob/master/git.md

Especially:
> You are encouraged to make liberal use of Git's history rewriting features while working locally, in order to arrange your commits into appropriate logical chunks that will make sense to your fellow developers.

PR #6, opened 2016-01-13
Service tokens

servingUpAces:
This pull request creates the token model and endpoints to create a token and delete a token. There is not a notion of being able to update the token. If the create token endpoint is called with a token already the old token is expired and a new token is created.


app/service/views/rest.py
```diff
@@ -67,3 +72,49 @@ def get_service(service_id=None):
         return jsonify(result="error", message="Service not found"), 404
     data, errors = services_schema.dump(services) if isinstance(services, list) else service_schema.dump(services)
     return jsonify(data=data)
+
+
+# TODO auth to be added
+@service.route('/<int:service_id>/token', methods=['POST'])
+def create_token(service_id=None):
```
mannickutd: Is a token created through a separate api call, or automatically when a service is created?

app/service/views/rest.py
```diff
@@ -67,3 +72,49 @@ def get_service(service_id=None):
...
+                                                         'expiry_date': datetime.now()})
+        except NoResultFound:
+            pass
+        save_token_model(Token(service_id=service_id, token=token))
+    except DAOException as e:
+        return jsonify(result='error', message=str(e)), 400
+    return jsonify(token=str(token)), 201
+
+
+@service.route('/<int:service_id>/token', methods=['DELETE'])
```
mannickutd: Is there a need for a delete token? perhaps a renew token method?

app/models.py
```diff
@@ -59,3 +59,12 @@ class Service(db.Model):
...
         backref=db.backref('user_to_service', lazy='dynamic'))
     restricted = db.Column(db.Boolean, index=False, unique=False, nullable=False)
+
+
+class Token(db.Model):
+    __tablename__ = 'tokens'
+
+    id = db.Column(db.Integer, primary_key=True)
+    token = db.Column(db.String, unique=True, nullable=False)
+    service_id = db.Column(db.Integer, db.ForeignKey('services.id'), index=True, nullable=False)
```
mannickutd: Can we add a service sqlalchemy relationship column.

app/dao/tokens_dao.py
```diff
@@ -0,0 +1,22 @@
...
+def save_token_model(token, update_dict={}):
+    if update_dict:
+        del update_dict['id']
+        db.session.query(Token).filter_by(id=token.id).update(update_dict)
+    else:
+        db.session.add(token)
+    db.session.commit()
+
+
+def get_model_tokens(service_id=None):
```
mannickutd: When is the need for more than one token? Should a service have more than one token?

app/dao/tokens_dao.py
```diff
@@ -0,0 +1,22 @@
+from app import db
+from app.models import Token
+
+
+def save_token_model(token, update_dict={}):
```
mannickutd: This function can probably have the update part removed, we should allow updating of the token, just creating/renewing.

app/service/views/rest.py
mannickutd: Perhaps this is just renew token?
GET /<int:service_id>/token/renew?

tests/app/dao/test_tokens_dao.py
```diff
@@ -0,0 +1,45 @@
+import uuid
+from app.dao import tokens_dao
+
+from app.models import Token
+
+
+def test_should_create_token(notify_api, notify_db, notify_db_session, sample_service):
```
mannickutd: A token is not passed to the create api endpoint, we generate a token and pass it back.

app/dao/services_dao.py
```diff
@@ -7,6 +7,7 @@
 
 
 def save_model_service(service, update_dict=None):
+    print(update_dict)
```
mannickutd: Please remove print statement

app/dao/tokens_dao.py
mannickutd: my mistake it will need that part.

app/dao/tokens_dao.py
```diff
@@ -0,0 +1,21 @@
...
+    if update_dict:
+        del update_dict['id']
+        db.session.query(Token).filter_by(id=token.id).update(update_dict)
+    else:
+        db.session.add(token)
+    db.session.commit()
+
+
+def get_model_tokens(service_id=None, raise_=True):
```
mannickutd: Can you please describe the raise_ parameter, it might be misleading.

app/service/views/rest.py
```diff
@@ -67,3 +73,54 @@ def get_service(service_id=None):
...
+        service = get_model_services(service_id=service_id)
+    except DataError:
+        return jsonify(result="error", message="Invalid service id"), 400
+    except NoResultFound:
+        return jsonify(result="error", message="Service not found"), 404
+
+    service_token = get_model_tokens(service_id=service_id, raise_=False)
+    if service_token:
+        save_token_model(service_token, update_dict={'id': service_token.id, 'expiry_date': datetime.now()})
+    return jsonify(data=token_schema.dump(service_token)), 202
```
mannickutd: Probably don't want to return this json.

tests/app/dao/test_tokens_dao.py
```diff
@@ -0,0 +1,52 @@
...
+    all_tokens = tokens_dao.get_model_tokens()
+    assert len(all_tokens) == 1
+    assert all_tokens[0].expiry_date == now
+
+
+def test_get_token_should_raise_exception_when_service_does_not_exist(notify_api, notify_db, notify_db_session,
+                                                                      sample_service):
+    try:
+        tokens_dao.get_model_tokens(sample_service.id)
+        fail()
```
mannickutd: Add a message to the fail, eg. "Failed to throw this exception"

tests/app/service/views/test_rest.py
```diff
@@ -261,3 +262,87 @@ def test_delete_service_not_exists(notify_api, notify_db, notify_db_session, sam
...
+        with notify_api.test_client() as client:
+            response = client.post(url_for('service.renew_token', service_id=sample_service.id),
+                                   headers=[('Content-Type', 'application/json')])
+            assert response.status_code == 201
+            assert response.get_data is not None
+            saved_token = Token.query.first()
+            assert saved_token.service_id == sample_service.id
+
+
+def test_renew_token_should_expire_the_old_token_and_create_a_new_token(notify_api, notify_db, notify_db_session,
```
mannickutd: Can we get a sample_token py.test fixture created so we can easily add tokens to tests in the future.
It would also be useful for this test as well.

PR #9, opened 2016-01-15
Added model and dao for Jobs.

ashimali:


migrations/versions/0004_create_jobs.py
```diff
@@ -0,0 +1,40 @@
+"""empty message
+
+Revision ID: c929d81b9e4a
```
mannickutd: Should also be 0004_create_jobs

PR #16, opened 2016-01-19
Add rest of user model fields to api.

ashimali:
First step to moving user interactions to api.


app/user/rest.py
```diff
@@ -17,8 +17,13 @@
 @user.route('', methods=['POST'])
 def create_user():
     user, errors = user_schema.load(request.get_json())
+    req_json = request.get_json()
+    if not req_json.get('password'):
+        return jsonify(result="error", message={'error': 'password missing'}), 400
```
mannickutd: Just to make this consistent with the marshmallow errors object, the dictionary value ('password missing') can that be a list.

app/user/rest.py
```diff
@@ -37,15 +42,8 @@ def update_user(user_id):
...
-        # TODO there has got to be a better way to do the next three lines
-        update_user, errors = user_schema.load(request.get_json())
-        if errors:
-            return jsonify(result="error", message=errors), 400
-        update_dict, errors = user_schema.dump(update_user)
-        # TODO FIX ME
-        # Remove update_service model which is added to db.session
         db.session.rollback()
-        save_model_user(user, update_dict=update_dict)
+        save_model_user(user, update_dict=request.get_json())
```
mannickutd: Will this fail if you pass in extra variables via http call?
The extra values will get passed to the sql alchemy and an exception will be raised.
That was the reason for the load and dump calls?

mannickutd:
This looks fine, I have chatted with Adam and a future task will be to implement state handling in the api for the User. Eg. 'pending', 'active' state changes will be handled in the api.

app/user/rest.py
ashimali: Yes but that causes exceptions with non nullable fields on model if load is called. There will need to be another way of doing that rather than calling load to get validation.

PR #17, opened 2016-01-19
Replace Tokens with Api keys

servingUpAces:
This pull request changes the name of Tokens to ApiKeys. 
ApiKeys require a name as well.
The next to come is the functionality to have multiple ApiKeys for a service. And creating an admin client for authentication that does not require a service and is not persisted for the database.


app/dao/api_key_dao.py
```diff
@@ -0,0 +1,51 @@
+from flask import current_app
+from itsdangerous import URLSafeSerializer
+
+from app import db
+from app.models import ApiKey
+
+
+def save_model_api_key(api_key, update_dict={}):
+    if update_dict:
+        del update_dict['id']
```
mannickutd: Adam highlighted a problem with this line, if the 'id' is not in the dict object, this throws a KeyError exception.
Can you check for the key before deleting.

app/models.py
```diff
@@ -85,13 +85,14 @@ class Service(db.Model):
...
+class ApiKey(db.Model):
+    __tablename__ = 'api_key'
 
     id = db.Column(db.Integer, primary_key=True)
-    token = db.Column(db.String(255), unique=True, nullable=False)
+    name = db.Column(db.String(255), nullable=False)
+    secret = db.Column(db.String(255), unique=True, nullable=False)
     service_id = db.Column(db.Integer, db.ForeignKey('services.id'), index=True, nullable=False)
-    service = db.relationship('Service', backref=db.backref('tokens', lazy='dynamic'))
+    service = db.relationship('Service', backref=db.backref('api_key', lazy='dynamic'))
```
mannickutd: shouldn't the backref be plural?

tests/app/conftest.py
```diff
@@ -64,15 +64,15 @@ def sample_template(notify_db,
...
+def sample_api_key(notify_db,
+                   notify_db_session,
+                   service=None):
     if service is None:
         service = sample_service(notify_db, notify_db_session)
-    data = {'service_id': service.id}
-    token = Token(**data)
-    save_model_token(token)
-    return token
+    data = {'service_id': service.id, 'name': service.name}
```
mannickutd: Shouldn't this just be 'service' not 'service_id'?

mannickutd:
Looks good apart from the few small comments.

app/models.py
servingUpAces: doesn't the backref match the table name?

PR #23, opened 2016-01-20
Allow multiple api keys

servingUpAces:
Allow services to have multiple api keys.
/service/<service_id>/api-key/renew has been renamed to /service/<service_id>/api-key
/service/<service_id>/api-key now creates a token and no longer expires the existing api key.
Moved test for this endpoint to it's own file.


mannickutd:
Couple things to pick up in future tasks.
1. Consistency among response structure for error.
2. Extend test client with methods to automatically add in auth for a request.

PR #32, opened 2016-01-21
All tests working, second time around.

mannickutd:


app/user/rest.py
```diff
@@ -57,19 +61,81 @@ def verify_user_password(user_id):
...
     try:
-        text_pwd = request.get_json()['password']
+        txt_pwd = request.get_json()['password']
     except KeyError:
         return jsonify(
             result="error",
             message={'password': ['Required field missing data']}), 400
-    if user.check_password(text_pwd):
+    if user.check_password(txt_pwd):
         return jsonify(), 204
```
ashimali: Bear in mind at the moment an empty json responses cause client to return InvalidResponse to admin.

mannickutd: Yes I noticed this morning, I will fix this.

PR #35, opened 2016-01-25
Record and persist failed login count on api.

ashimali:
This pr should wait until after hack day


servingUpAces:
We need to reset the failed login count when the user successfully logs in. Will do that in another PR.

PR #38, opened 2016-01-28
Create queue for service

servingUpAces:
When a service sends a sms notification, the notification is added to a queue for the service.
The message id, service is, to, message type template id are encrypted as the MessageBody. MessageAttributes are message id, service id, message type, and template id.


app/service/rest.py
```diff
@@ -28,6 +28,9 @@ def create_service():
     # I believe service is already added to the session but just needs a
     # db.session.commit
     try:
+        # add service name to here or in dao?
+        import uuid
+        service.queue_name = uuid.uuid4()
```
mannickutd: You can probably do this through the on_creation method of the db model.

mannickutd:
Only one comment otherwise it looks good.

app/service/rest.py
servingUpAces: Done

PR #46, opened 2016-02-01
version file has been updated to include a travis commit and time of …

mannickutd:
…build.


app/version.py
```diff
@@ -1 +1,2 @@
-__version__ = '0.1.1'
+__build__ = "12345"
+__time__ = "2016-01-01:10:10:10"
```
quis: Would be better to put non-real-looking placeholders here, otherwise in 3 months time the page would return `api_built_time: "2016-01-01:10:10:10"` and you’d be left wondering why the app had reverted to such an old version. And it would make your `sed` command simpler.

PR #61, opened 2016-02-09
[WIP] New model class and dao for notification. This will be used for

ashimali:
recording status and outcome of sending notifications.


app/models.py
```diff
@@ -189,3 +189,32 @@ def code(self, cde):
...
+
+class Notification(db.Model):
+
+    __tablename__ = 'notifications'
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True)
+    to = db.Column(db.String, nullable=False)
+    job_id = db.Column(UUID(as_uuid=True), db.ForeignKey('jobs.id'), index=True, unique=False, nullable=False)
+    job = db.relationship('Job', backref=db.backref('notifications', lazy='dynamic'))
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), index=True, unique=False)
```
mannickutd: Why doesn't the service and template, have a relationship?

tests/app/conftest.py
```diff
@@ -180,3 +181,31 @@ def _create():
...
+    notificaton_id = uuid.uuid4()
+    to = '+44709123456'
+
+    data = {
+        'id': notificaton_id,
+        'to': to,
+        'job_id': job.id,
+        'service_id': service.id,
+        'template_id': template.id
+    }
```
mannickutd: To keep consistent with the other parts of the marshmallow schemas, we should remove the '_id' from job, server and template.

PR #70, opened 2016-02-10
Return notification is when adding notification to queue

servingUpAces:
- Return notification_id on create notification endpoints
     - /notification/sms
     - /notification/email
     - /notificaiton/sms/service/<service_id>
- Update message attribute on SQS to notification_id from message_id
- Fixed intermittent test


ashimali:
:+1:

PR #71, opened 2016-02-12
Validate template for a service

mannickutd:


app/schemas.py
```diff
@@ -116,6 +120,11 @@ def validate_schema(self, data):
                         break
                 if not valid:
                     raise ValidationError('Invalid phone number for restricted service', 'restricted')
+            # Assert the template is valid for the service which made the request.
+            service = api_user['client']
+            if service != current_app.config.get('ADMIN_CLIENT_USER_NAME'):
+                if template.service != models.Service.query.filter_by(id=service).first():
```
quis: This would be better with an `and` rather than nested `if` statements.

mannickutd: Yes I agree.

PR #73, opened 2016-02-16
Add environment.sh

quis:
Structure copied from the delivery app


quis:
This needs more things to work properly, but it’s a start that someone else can pick up.

PR #78, opened 2016-02-17
Move sms notifications into celery

minglis:
Initial move from direct access to SQS to using [celery](http://www.celeryproject.org/) as task manager.

- This Pull request moves the sending of SMS messages for 3rd party clients into the celery model.
- This requires the SMS clients to be accessable to the API application, so the Twilio client is lifted across. This is now duplicated, but this will last only a short while until migration is complete.

Celery code is help in the app/celery folder. The celery instance is wrapped in the standard Flask module way as per the docs: (http://flask.pocoo.org/docs/0.10/patterns/celery/)[http://flask.pocoo.org/docs/0.10/patterns/celery/]
- this notify_celery object is available across the application.

Tasks are defined in app/celery/tasks. Each task encapsulates a single action for our system. For this pull request it sends the Notification to Twilio, and persists in the database.

Celery is configured in the the config.Config class as per the rest of the application.

Running celery locally:
- Celery is build in the run_celery.py script in the application root
- Celery is ran using the script run_celery.sh - in the scripts folder.
- Additionally aws_run_celery.py in the root is used in AWS to run the application - it uses the credstash secrets model.

Celery will run with the following command:
celery -A run_celery.notify_celery worker --loglevel=INFO --logfile=/var/log/notify/application.log --concurrency=4 -Q sms

Which means:
-  it logs into the application.log file as per Flask apps
- Runs with a concurreny of 4 workers
- Listens ONLY to the queue named 'sms'.

Queues
- Currently celery is aware of 1 queue - SMS. There will be more queues as we go - jobs/emails/2fa and so on. Each of which celery must know about. The model for this on this pull request is hard coded. This will need to migrate as we add more queues.


app/notifications/rest.py
```diff
@@ -6,30 +6,51 @@
...
-    resp_json = request.get_json()
-
-    notification, errors = sms_template_notification_schema.load(resp_json)
+    notification, errors = sms_template_notification_schema.load(request.get_json())
     if errors:
         return jsonify(result="error", message=errors), 400
 
-    notification_id = add_notification_to_queue(api_user['client'], notification['template'], 'sms', notification)
+    try:
+        templates_dao.get_model_templates(template_id=notification['template'], service_id=api_user['client'])
```
mannickutd: I was fiddling with the idea that this can all be done in the schema class but for now this is fine.

tests/app/celery/test_tasks.py
```diff
@@ -0,0 +1,76 @@
...
+    assert persisted_notification.status == 'sent'
+
+
+def test_should_persist_notification_as_failed_if_sms_client_fails(sample_template, mocker):
+    notification = {
+        "template": sample_template.id,
+        "to": "+441234123123"
+    }
+    mocker.patch('app.encryption.decrypt', return_value=notification)
+    mocker.patch('app.twilio_client.send_sms', side_effect=TwilioClientException())
```
mannickutd: Duplicate code, we usually just create a fixture to mock out the function

PR #80, opened 2016-02-17
Fire text client

minglis:
Pulls in a very (too) simple client to send SMS via Firetext

Firetext is a G-Cloud SMS supplier. This will deliver SMS's that are created using the client API - so not the validation code ones.

Requires new properties:
- FIRETEXT_NUMBER - From number for the text
- FIRETEXT_API_KEY - API key for firetext

This replaces twilio for those calls.

Will need a more detailed set of clients for full on beta delivery. This is to get us off twilio to see if delivery is better. The start not the end of SMS integrations.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -8,12 +8,14 @@ from werkzeug.local import LocalProxy
 from utils import logging
 from app.celery.celery import NotifyCelery
 from app.clients.sms.twilio import TwilioClient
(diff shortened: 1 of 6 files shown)


minglis:
Note properties are on preview - so this can be merged if it's OK.

mannickutd:
Only one general comment and that is to create a fixture for class mocks.

PR #81, opened 2016-02-17
Use Celery to send sms code

servingUpAces:


app/celery/tasks.py
```diff
@@ -30,3 +30,13 @@ def send_sms(service_id, notification_id, encrypted_notification):
 
     except SQLAlchemyError as e:
         current_app.logger.debug(e)
+
+
+@notify_celery.task(name='send-sms-code')
+def send_sms_code(encrypted_notification):
+    notification = encryption.decrypt(encrypted_notification)
```
minglis: Is this a 'notification'? Should it be called something like 'validation code'. Notifications have a specific meaning in the code and we have loads of confusion, might be useful to tidy up our domain language.

tests/app/celery/test_tasks.py
```diff
@@ -74,3 +77,24 @@ def test_should_not_send_sms_if_db_peristance_failed(sample_template, mocker):
...
+
+
+def test_should_log_firetext_client_exception(mocker):
+    notification = {'to': '+441234123123',
+                    'secret_code': '12345'}
+
+    encrypted_notification = encryption.encrypt(notification)
+    mocker.patch('app.firetext_client.send_sms', side_effect=FiretextClientException)
+    send_sms_code(encrypted_notification)
+    firetext_client.send_sms.assert_called_once_with(notification['to'], notification['secret_code'])
```
minglis: Where does the assert about logging happen? Am I being dumb?

app/user/rest.py
```diff
@@ -135,8 +137,8 @@ def send_user_code(user_id):
     create_user_code(user, secret_code, verify_code.get('code_type'))
```
minglis: Do you want to move this into the task? Guess we could do it here - and the application fails and the user knows? Also not a performance bottleneck on our application?

app/celery/tasks.py
```diff
@@ -30,3 +30,13 @@ def send_sms(service_id, notification_id, encrypted_notification):
...
+
+    try:
+        firetext_client.send_sms(notification['to'], notification['secret_code'])
+    except FiretextClientException as e:
+        current_app.logger.debug(e)
```
minglis: Should we take action here to deal with the now orphaned code?

servingUpAces: I think we should. Should I create the failed queue now.

tests/app/celery/test_tasks.py
servingUpAces: I didn't actually know what how to test that the logging happened. If there is a failure we might want to put it in a retry queue

app/user/rest.py
servingUpAces: As discussed this is okay here.

PR #84, opened 2016-02-19
Refactor user/<user_id>/code into two endpoints.

servingUpAces:
- Created new endpoint user/<user_id>/sms-code to send the sms verification code to the user.
- Create new endpoirtn user/<user_id>/email-code to send the email verifcation code to the user.
- Marked the old methods, schema, tests with a TODO to be deleted when the admin app is no longer sending messages to /user/<user_id>/code
- Added error handlers for DataError and NoResultFound. Data error catches invalid input errors.
- Added error handler for SqlAlchemyError which catches any other database errors.
- Removed the need for the try catches around the db calls in the user endpoints with the addition of the db error handlers.
- We may want to wrap db exceptions in the dao, if we want the No results found message to be more specific and say no result found for user.

### DIFF
--- a/app/errors.py
+++ b/app/errors.py
@@ -2,6 +2,8 @@ from flask import (
     jsonify,
     current_app
 )
+from sqlalchemy.exc import SQLAlchemyError, DataError
+from sqlalchemy.orm.exc import NoResultFound
 
 
 def register_errors(blueprint):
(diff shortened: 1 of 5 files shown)


app/errors.py
```diff
@@ -32,3 +34,18 @@ def limit_exceeded(e):
     def internal_server_error(e):
         current_app.logger.exception(e)
         return jsonify(error="Internal error"), 500
+
+    @blueprint.app_errorhandler(NoResultFound)
+    def no_result_found(e):
```
mannickutd: This is good. In the future we could add to the error message by examining exception to determine exactly what object they were trying to retrieve.

PR #85, opened 2016-02-19
Permissions added.

mannickutd:


mannickutd:
The admin side is to be added.

PR #87, opened 2016-02-22
Email templates part 2

minglis:
- depends on https://github.com/alphagov/notifications-api/pull/86

- extends the previous pull request by adding the subject line to email templates.


app/service/rest.py
```diff
@@ -1,93 +1,126 @@
...
 
 
+@service.route('', methods=['GET'])
+def get_services():
+    user_id = request.args.get('user_id', None)
+    if user_id:
+        services = dao_fetch_all_services_by_user(user_id)
+    else:
+        services = dao_fetch_all_services()
+    data, errors = services_schema.dump(services)
```
mannickutd: Can probably remove this line and pass this directly to jsonify.

app/service/rest.py
```diff
@@ -1,93 +1,126 @@
...
-        if errors:
-            return jsonify(result="error", message=errors), 400
-        try:
-            save_model_service(service, update_dict=update_dict)
-        except DAOException as e:
-            return jsonify(result="error", message=str(e)), 500
-    return jsonify(data=service_schema.dump(service).data), status_code
+    fetched_service = dao_fetch_service_by_id(service_id)
+    if not fetched_service:
+        return jsonify(result="error", message="not found"), 404
```
mannickutd: This error message is inconsistent with the other 404 messages in the api. Should we create task to update the other messages?

app/template/rest.py
```diff
@@ -1,36 +1,84 @@
...
+    if errors:
+        return jsonify(result="error", message=errors), 400
+    new_template.service = fetched_service
     try:
-        templates = get_model_templates(template_id=template_id)
-    except DataError:
-        return jsonify(result="error", message="Invalid template id"), 400
-    except NoResultFound:
+        dao_create_template(new_template)
+    except IntegrityError as ex:
```
mannickutd: Why isn't this already handled by the schema class??

app/template/rest.py
```diff
@@ -1,36 +1,84 @@
...
+        current_app.logger.debug(ex)
+        message = "Failed to create template"
+        if "templates_subject_key" in str(ex):
+            message = 'Duplicate template subject'
```
mannickutd: This message is inconsistent with how the schema would return error messages.

app/template/rest.py
```diff
@@ -1,36 +1,84 @@
...
+    fetched_template = dao_get_template_by_id_and_service_id(template_id=template_id, service_id=service_id)
+    if not fetched_template:
         return jsonify(result="error", message="Template not found"), 404
-    if isinstance(templates, list):
-        data, errors = templates_schema.dump(templates)
-    else:
-        data, errors = template_schema.dump(templates)
+
+    current_data = dict(template_schema.dump(fetched_template).data.items())
+    current_data.update(request.get_json())
```
mannickutd: The two lines above don't really make sense to me??
Why load a schema from a json object to then update the schema with whats passed in from the JSON?
Have a look at the "_load_json" schemas, we don't have to do the template dump and load process?

minglis: I could be wrong, but the update endpoints now can handle a subset of data - they don't need the whole object - this means that the load_from_json fails as it's missing lots of fields. I had to get the current data, update with the JSON, then validate to ensure it was all OK.

mannickutd: I will break this to show why it is wrong.

PR #91, opened 2016-02-22
Send Email via the API

minglis:
- uses the new subject/email from fields present on the templates / service tables
- brings the send email api call into line with the sms one.
- same fields (to/template_id)
- same rules regarding restricted services
- wired in as a task into celery

Requires
- new celery queue
- new env property (NOTIFY_EMAIL_DOMAIN)


app/notifications/rest.py
```diff
@@ -47,6 +54,12 @@ def create_sms_notification():
     except NoResultFound:
         return jsonify(result="error", message={'template': ['Template not found']}), 400
 
+    service = services_dao.dao_fetch_service_by_id(api_user['client'])
+
+    if service.restricted:
+        if notification['to'] not in [user.email_address for user in service.users]:
+            return jsonify(result="error", message={'to': ['Invalid phone number for restricted service']}), 400
+
```
mannickutd: Why are we doing this outside of the schema?

minglis: Thinking that the schema shouldn't do DB lookups? Was a bit worried about relatively hidden DB calls, wanted it all up front. Maybe not the best option, happy to refactor again when the service is on the context, as that will affect what calls we need to make.

PR #95, opened 2016-02-23
New endpoint to get users for service id.

servingUpAces:
New endpoint: /service/<service_id>/users returns a list of all users associated with the service

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -161,7 +161,7 @@ def get_api_keys(service_id, key_id=None):
 def get_users_for_service(service_id):
     fetched = dao_fetch_service_by_id(service_id)
     if not fetched:
-        return jsonify(result="error", message="Service not found"), 404
-    print(fetched.users)
+        return jsonify(data=[])
+
     result = users_schema.dump(fetched.users)
     return jsonify(data=result.data)
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -27,8 +27,8 @@ from app.models import ApiKey
 from app.schemas import (
     services_schema,
     service_schema,
-    api_keys_schema
-)
+    api_keys_schema,
+    users_schema)
 from app import email_safe
 
 from flask import Blueprint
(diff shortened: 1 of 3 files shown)


tests/app/service/test_rest.py
```diff
@@ -341,3 +342,27 @@ def test_update_service_should_404_if_id_is_invalid(notify_api, notify_db):
...
+
+            sample_service = create_service(notify_db=notify_db, notify_db_session=notify_db_session,
+                                            service_name='Sample service', user=sample_user)
+            auth_header = create_authorization_header(
+                path='/service/{}/users'.format(sample_service.id),
+                method='GET'
+            )
+
+            resp = client.get(
+                '/service/{}/users'.format(sample_service.id),
```
mannickutd: Why not url_for?

app/service/rest.py
```diff
@@ -155,3 +155,13 @@ def get_api_keys(service_id, key_id=None):
...
 
     return jsonify(apiKeys=api_keys_schema.dump(api_keys).data), 200
+
+
+@service.route('/<service_id>/users', methods=['GET'])
+def get_users_for_service(service_id):
+    fetched = dao_fetch_service_by_id(service_id)
+    if not fetched:
+        return jsonify(result="error", message="Service not found"), 404
+    print(fetched.users)
```
minglis: think this should go

tests/app/service/test_rest.py
```diff
@@ -341,3 +342,27 @@ def test_update_service_should_404_if_id_is_invalid(notify_api, notify_db):
                 headers=[('Content-Type', 'application/json'), auth_header]
             )
             assert resp.status_code == 404
+
+
+def test_get_users_by_service(notify_api, notify_db, notify_db_session, sample_user):
```
minglis: any tests for fail cases (empty list etc)

tests/app/service/test_rest.py
minglis: I'm not 100% on the benefit of url for

- using it means if I change the method sig I change the tests
- not using it means I change the path I change the tests

By all means use it in application code - the right answer - using it in tests to save typing is not 100% benefit.

tests/app/service/test_rest.py
servingUpAces: Test has been added

tests/app/service/test_rest.py
servingUpAces: If the compromise is to write unit tests for url_for, then those tests are slightly annoying to write and maintain.

PR #99, opened 2016-02-25
Celery jobs

minglis:
Creates a task for processing jobs

- Create job API call makes a task to process job in the future
- Task will read CSV from S3, process it, and make a task for each notification

Tasks for notifications used by Jobs are the same as used by the API.

Implemented both for Email and for SMS. API endpoints / jobs / etc all set up.

**Breaking change **
Header in the CSV file is now 'to' not number. This matches the API and makes CSV format the same for both SMS and EMAIL. Template type allows the choice of delivery to be made.

Additional fields added to both Jobs and Notifications to capture process time / length on queue / etc


minglis:
Depends on https://github.com/alphagov/notifications-admin/pull/199

PR #101, opened 2016-02-25
Add services permissions

mannickutd:


app/permission/rest.py
```diff
@@ -0,0 +1,56 @@
...
+    if errors:
+        abort(500, errors)
+    return jsonify(data=data)
+
+
+@permission.route('/<permission_id>', methods=['GET'])
+def get_permission(permission_id):
+    inst = permission_schema.get_query(filter_by_dict={'id': permission_id}).first()
+    if not inst:
+        abort(404, 'Permission not found')
```
servingUpAces: Please format errors as {result:error, message:Permission not found}. The base client will then do the right thing with the error (hopefully :).

servingUpAces: Actually we want to start adding more details for logging, so this could read 'Permission not found for id: {}'.format(permission_id)

migrations/versions/0027_add_service_permission.py
```diff
@@ -0,0 +1,41 @@
+"""empty message
+
+Revision ID: 0027_add_service_permission
+Revises: 0023_drop_token
```
servingUpAces: Alembic cares about the comment

tests/app/job/test_rest.py
```diff
@@ -50,7 +50,6 @@ def test_get_job_with_invalid_job_id_returns404(notify_api, sample_template):
             response = client.get(path, headers=[auth_header])
             assert response.status_code == 404
             resp_json = json.loads(response.get_data(as_text=True))
-            print(resp_json)
             assert resp_json['result'] == 'error'
```
servingUpAces: print statements can go

PR #104, opened 2016-02-26
Add user permissions

mannickutd:


app/permission/rest.py
```diff
@@ -0,0 +1,54 @@
...
+from flask import (jsonify, request, abort, Blueprint, current_app)
+from app.schemas import permission_schema
+from app.errors import register_errors
+from app.dao.permissions_dao import permission_dao
+
+permission = Blueprint('permission', __name__)
+register_errors(permission)
+
+
+@permission.route('', methods=['GET'])
```
minglis: Don't think we need a rest endpoint for permissions - we'd only operate on them via user actions.

app/dao/__init__.py
```diff
@@ -1,6 +1,54 @@
 from sqlalchemy.exc import SQLAlchemyError
+from werkzeug.datastructures import MultiDict
```
minglis: Guess this is superceded by the next pull request

app/models.py
```diff
@@ -265,3 +264,25 @@ class InvitedUser(db.Model):
         default=datetime.datetime.now)
     status = db.Column(
         db.Enum(*INVITED_USER_STATUS_TYPES, name='invited_users_status_types'), nullable=False, default='pending')
+
+
+class Permission(db.Model):
+    __tablename__ = 'permissions'
```
minglis: Probably need to model permissions as a list - like we do with user status etc - enums?

PR #105, opened 2016-02-26
Create permissions on service creation

mannickutd:


app/dao/services_dao.py
```diff
@@ -20,7 +20,14 @@ def dao_fetch_service_by_id_and_user(service_id, user_id):
...
-    service.users.append(user)
+    try:
+        from app.dao.permissions_dao import permission_dao
+        service.users.append(user)
+        permission_dao.add_default_service_permissions_for_user(user, service)
+    except Exception as e:
+        # Proper clean up
+        db.session.rollback()
+        raise e
     db.session.add(service)
```
minglis: What happens if this fails? Does the permsission DAO commit? Can we get into a weird state with the permissions persisted and the service not?

app/dao/__init__.py
```diff
@@ -1,6 +1,55 @@
...
+from sqlalchemy.orm.relationships import RelationshipProperty
+from app import db
 
 
 # Should I use SQLAlchemyError?
 class DAOException(SQLAlchemyError):
     pass
+
+
+class DAOClass(object):
```
minglis: Thought we talked about not doing this sort of stuff now, pushing queries into the DAO, as explicit queries, until we understood the bounds of our requirements better?

app/schemas.py
```diff
@@ -21,6 +22,30 @@ def __init__(self, *args, load_json=False, **kwargs):
         self.load_json = load_json
         super(BaseSchema, self).__init__(*args, **kwargs)
 
+    __envelope__ = {
```
minglis: Whats the envelope stuff for?

app/dao/__init__.py
minglis: We could push some of this shared logic down onto the Model classes, and remove the DAO stuff maybe. For now lets not get into this we need the features more than the code refactors.

app/dao/services_dao.py
mannickutd: The reason for the rollback was to stop the permission getting committed.
inside the 'create' method there is a commit parameter if you want to commit the session or not.

app/dao/__init__.py
mannickutd: I will come over to your desk to see so we can get what you want done.

app/schemas.py
mannickutd: This is precursor for wrapping the responses and possible the requests instead of using 'data'
e.g.

app/dao/permissions_dao.py
```diff
@@ -7,5 +16,30 @@ class PermissionDAO(DAOClass):
     class Meta:
         model = Permission
 
+    def get_query(self, filter_by_dict={}):
```
minglis: Would we query permissions by anything other than the user? Is there a user.persmissions mapping, so when you get a user you get the permisssions / save the user update the permissions?

app/schemas.py
```diff
@@ -58,6 +59,18 @@ def make_instance(self, data):
 
 
 class UserSchema(BaseSchema):
+
+    permissions = fields.Method("user_permissions", dump_only=True)
+
+    def user_permissions(self, usr):
```
minglis: Can this be done with a join? So that the user object always has permissions? Sorry if I'm missing some marshmallow stuff here that does this.

PR #109, opened 2016-02-29
Replace placeholders with personalisation

quis:
## Refactor send notification into one route

Using a URL parameter means that sending a notification can be done in one route, rather than two separate routes and an extra method.

This commit also refactors that one remaining method to be shorter/cleaner/more readable (or I think so anyway).

No functional changes in this commit.

## Accept and validate personalisation

This commit allows the send notification endpoint to accept an extra parameter, `personalisation`, the contents of which will be used (later) to replace the placeholders in the template.

It does validation in the following places:
- at the schema level, to validate the type and (optional) presence of personalisation
- at the endpoint, to check whether the personalisation provided matches exactly the placeholders in the template

It does not do validation when processing CSV files, as these are assumed to already have been validated by the admin app.

It explicitly does not persist either the names of the placeholders (these should always be derived from the template contents unless it really becomes a performance concern) or the values of the placeholders (because they might be personal data).

## Replace placeholders before sending a message

This commit replaces placeholders in a template with the user’s data, using the Template class from utils (https://github.com/alphagov/notifications-utils/tree/master/utils#template)

It also extracts the personalisation data from the CSV, taking account of the different column headings that SMS and email CSVs will have.

## Remove CSV utils

These are in the utils repo since https://github.com/alphagov/notifications-utils/releases/tag/0.2.1


quis:
## Prefix all SMS messages with service name

Implements alphagov/notifications-utils#4

app/celery/tasks.py
```diff
@@ -78,12 +86,14 @@ def send_sms(service_id, notification_id, encrypted_notification, created_at):
...
         dao_create_notification(notification_db_object)
 
         try:
-            client.send_sms(notification['to'], template.content)
+            client.send_sms(
+                notification['to'],
+                template.replaced
+            )
         except FiretextClientException as e:
             current_app.logger.debug(e)
```
mannickutd: Maybe this should be logger.exception?
On ever environment we probably want to know about this error.

quis: That line was added by here: https://github.com/alphagov/notifications-api/commit/a2341be0e2c200c1e8dc541b0608876f6b09739c#diff-201d2d704e6c87ac11b55391f2da412fR28

@minglis is there a good reason for it not to be an exception?

quis: Spoke with @minglis, no objection so changed this to log an exception in 2da8a72

PR #111, opened 2016-02-29
Fixed some tests.

mannickutd:


migrations/versions/0031_add_manage_team_permission.py
```diff
@@ -0,0 +1,40 @@
...
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    conn = op.get_bind()
+    user_services = conn.execute("SELECT * FROM user_to_service").fetchall()
+    for entry in user_services:
+        id_ = uuid.uuid4()
```
servingUpAces: id_ needs to be unique for each insert.

PR #113, opened 2016-03-01
Fetch endpoints for notifications

minglis:
Allow clients to fetch notifications

Allow admins to fetch notifications by service / job-service

Includes pagination on the fetch, also there is a decorator to ensure that fetch by service / job can only be done by admin JWT tokens.


minglis:
Requires new property on preview/staging (page size) so before merge must do this.

app/dao/notifications_dao.py
```diff
@@ -16,9 +18,24 @@ def get_notification_for_job(service_id, job_id, notification_id):
     return Notification.query.filter_by(service_id=service_id, job_id=job_id, id=notification_id).one()
 
 
-def get_notifications_for_job(service_id, job_id):
-    return Notification.query.filter_by(service_id=service_id, job_id=job_id).all()
+def get_notifications_for_job(service_id, job_id, page=1):
+    query = Notification.query.filter_by(service_id=service_id, job_id=job_id)\
+        .order_by(asc(Notification.created_at))\
```
servingUpAces: I wonder if order by should be desc, see last notification first on the list?

app/dao/notifications_dao.py
```diff
@@ -16,9 +18,24 @@ def get_notification_for_job(service_id, job_id, notification_id):
...
+        )
+    return query
 
 
 def get_notification(service_id, notification_id):
     return Notification.query.filter_by(service_id=service_id, id=notification_id).one()
+
+
+def get_notifications_for_service(service_id, page=1):
+    print(service_id)
```
servingUpAces: print statement

app/dao/notifications_dao.py
```diff
@@ -16,9 +18,24 @@ def get_notification_for_job(service_id, job_id, notification_id):
...
+    query = Notification.query.filter_by(service_id=service_id).order_by(asc(Notification.created_at)).paginate(
```
servingUpAces: I wonder if order by should be desc, see last notification first on the list?

app/notifications/rest.py
```diff
@@ -40,6 +42,86 @@ def get_notifications(notification_id):
         return jsonify(result="error", message="not found"), 404
 
 
+@notifications.route('/notifications', methods=['GET'])
+def get_all_notifications():
+    page = get_page_from_request()
+
+    if not page:
+        return jsonify(result="error", message="Invalid page"), 400
```
servingUpAces: Is it possible there are no notifications? Return empty list

app/notifications/rest.py
```diff
@@ -40,6 +42,86 @@ def get_notifications(notification_id):
...
+    ), 200
+
+
+@notifications.route('/service/<service_id>/notifications', methods=['GET'])
+@require_admin()
+def get_all_notifications_for_service(service_id):
+    page = get_page_from_request()
+
+    if not page:
+        return jsonify(result="error", message="Invalid page"), 400
```
servingUpAces: Is is possible there are no notifications for service? Return empty list, otherwise return message='No notifications for service id: {service_id}'

app/notifications/rest.py
```diff
@@ -40,6 +42,86 @@ def get_notifications(notification_id):
...
+    ), 200
+
+
+@notifications.route('/service/<service_id>/job/<job_id>/notifications', methods=['GET'])
+@require_admin()
+def get_all_notifications_for_service_job(service_id, job_id):
+    page = get_page_from_request()
+
+    if not page:
+        return jsonify(result="error", message="Invalid page"), 400
```
servingUpAces: Should we be more specific with the error message: 'No notifications for job id {job_id}'

tests/app/notifications/test_rest.py
```diff
@@ -47,6 +48,204 @@ def test_get_notifications_empty_result(notify_api, sample_api_key):
...
+            response = client.get(
+                '/notifications',
+                headers=[auth_header])
+
+            notifications = json.loads(response.get_data(as_text=True))
+            assert len(notifications['notifications']) == 3
+            assert notifications['notifications'][0]['to'] == notification_3.to
+            assert notifications['notifications'][1]['to'] == notification_2.to
+            assert notifications['notifications'][2]['to'] == notification_1.to
+            assert response.status_code == 200
```
servingUpAces: Do we want to check if the sort order is correct?

app/dao/notifications_dao.py
minglis: Chris and I chatted about this and went for newest first. We'll see it this makes sense to users in due course.

app/notifications/rest.py
minglis: Think it's a 200 with an empty list.

app/notifications/rest.py
minglis: Job ID is in the URL, so logging of URL should be enough? Thoughts? It's not consistent.

tests/app/notifications/test_rest.py
minglis: It does:

assert notifications['notifications'][0]['to'] == notification_3.to
assert notifications['notifications'][1]['to'] == notification_2.to
assert notifications['notifications'][2]['to'] == notification_1.to

Is in reverse order to creation.

app/notifications/rest.py
servingUpAces: I'd prefer it if the log message had the id but I'd would most prefer consistent messages.

app/notifications/rest.py
servingUpAces: Ah right - the error message is for invalid request arguments, fine.

PR #115, opened 2016-03-01
[WIP] Start of api for accepting invite.

ashimali:
This only handles the case of an existing user being invited and added to an existing service.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -19,8 +19,10 @@ from app.dao.services_dao import (
     dao_fetch_all_services,
     dao_create_service,
     dao_update_service,
-    dao_fetch_all_services_by_user
+    dao_fetch_all_services_by_user,
+    dao_add_user_to_service
 )
+
 from app.dao.users_dao import get_model_users
 from app.models import ApiKey
 from app.schemas import (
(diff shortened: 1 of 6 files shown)


app/service/rest.py
```diff
@@ -155,5 +155,23 @@ def get_users_for_service(service_id):
     return jsonify(data=result.data)
 
 
+@service.route('/<service_id>/users/<user_id>', methods=['POST'])
+def add_user_to_service(service_id, user_id):
+    service = dao_fetch_service_by_id(service_id)
+    if not service:
+        return _service_not_found(service_id)
+    user = get_model_users(user_id=user_id)
+
```
servingUpAces: get returns None  - should add a check here.

app/service/rest.py
```diff
@@ -155,5 +155,23 @@ def get_users_for_service(service_id):
...
+    if user in service.users:
+        return jsonify(result='error',
+                       message='User id: {} already part of service id: {}'.format(user_id, service_id)), 400
+
+    service.users.append(user)
+    dao_update_service(service)
```
servingUpAces: dao_add_user_to_service does the same thing

tests/app/service/test_rest.py
```diff
@@ -456,3 +456,127 @@ def test_default_permissions_are_added_for_user_service(notify_api,
...
+            resp = client.get(
+                '/service/{}/users'.format(sample_service.id),
+                headers=[('Content-Type', 'application/json'), auth_header]
+            )
+
+            assert resp.status_code == 200
+            result = json.loads(resp.get_data(as_text=True))
+            assert len(result['data']) == 1
+            assert result['data'][0]['email_address'] == user_already_in_service.email_address
+
```
servingUpAces: Why call the get first here?

ashimali:
Updated as per comments above.

tests/app/service/test_rest.py
ashimali: Verify only one user on service. Call post to add, verify 2 users on service, the original one and the one added.

PR #121, opened 2016-03-02
Revert queue config

servingUpAces:


config.py
```diff
@@ -29,7 +29,7 @@ class Config(object):
     BROKER_URL = 'sqs://'
     BROKER_TRANSPORT_OPTIONS = {
         'region': 'eu-west-1',
-        'polling_interval': 1,  # 1 second
```
mannickutd: Should probably leave this at 1.

PR #125, opened 2016-03-03
Ensure restricted service are respected by tasks:

minglis:
This is checked on 3rd party API calls, but jobs (CSV files) were able expected to only allow valid files.

Change in tack means we want to have restricted notification failures reported in the UI.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -207,7 +207,7 @@ class VerifyCode(db.Model):
         return check_hash(cde, self._code)
 
 
-NOTIFICATION_STATUS_TYPES = ['sent', 'failed']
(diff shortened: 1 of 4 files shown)


app/models.py
```diff
@@ -207,7 +207,7 @@ def check_code(self, cde):
         return check_hash(cde, self._code)
 
 
-NOTIFICATION_STATUS_TYPES = ['sent', 'failed']
+NOTIFICATION_STATUS_TYPES = ['sent', 'failed', 'failed - restricted service']
```
servingUpAces: The new status type is not being used

PR #139, opened 2016-03-08
Reset password

servingUpAces:
The password reset functionality was broken when the notifications-python-client stopped taking content as an option in the send_email method. This pull request creates a celery task to create and send the email for the reset password url. A new endpoint to reset-password has been created to call the celery task. There was also a problem with the save_user_model, where the password_changed_at datetime was being set to the value in the update_dict overriding the value when changing the password. 

story: https://www.pivotaltracker.com/story/show/114940405

Changes include: 
- Added an endpoint and celery task to email a reset password url	
- Send email address in the data rather than the user_id as a path param. 
- Remove unused OldRequestVerifyCodeSchema.
- Include token creation date in the url token
- Remove password_changed_at from the update_dict in users_dao
- Format dates in UserSchema
- Properly formatted subject and message body for the password reset email
- Add name to the message for reset password


app/schemas.py
```diff
@@ -209,6 +197,16 @@ class Meta:
...
 
+class EmailDataSchema(ma.Schema):
+    email = fields.Str(required=False)
+
+    @validates('email')
+    def validate_email(self, value):
+        try:
+            validate_email_address(value)
+        except InvalidEmailError:
+            raise ValidationError('Invalid email')
```
mannickutd: It might be better to pass the message form the exception InvalidEmailError so that  the returned response puts the more specific reason in the response json.

minglis:
I have seen this:

👍

PR #142, opened 2016-03-09
Accept CSV files with additional columns

quis:
Currently when the Celery task processes a CSV it will call the API with the values for all the non-recipient columns in the `personalisation` field. This means that those API calls would fail, even though the CSV has been processed ‘successfully’.

This was not being caught by the tests, so this commit adds extra checks to make the task is passing the right data to the API call.

It then updates utils to version 2.0.1 which brings in this fix: https://github.com/alphagov/notifications-utils/pull/10


ashimali:
:+1:

PR #143, opened 2016-03-09
Rate limit

minglis:
Ensure clients have rate limit enforced  …
- rate limiting is a hard number per day
- not limited in terms of rate of request
- limit is a single number held against the service
- every notification counts against the limit, regardless of type
- return a 429 if limit exceeded.

	Job processing respects sendlimits  …
- If a job starts it MUST be able to fit into the days sending limits
- So if service limit is 10, and we've sent 5 messages and the current job is 4 then it's OK.
- If the job is 6 then it's over the limit and it should fail
- Job should NOT start if can't complete in the limit


minglis:
Demoed to @ashimali

minglis:
As per discussions if someone could code review this and give me a👍 I'll do the merge.

thanks

ashimali:
@minglis sorry you conflicts

minglis:
Fixed

quis:
:+1: from me.

PR #145, opened 2016-03-10
Duplicate service name check added and all tests passing.

mannickutd:


mannickutd:
Demoed to @servingUpAces

PR #147, opened 2016-03-11
Be agnostic about format when comparing phone numbers

quis:
This pull request fixes a bug introduced here: https://github.com/alphagov/notifications-api/commit/832375744185f35cdb7d739a49ccbaefa44ce6e6

***

If a service is in restricted mode then a user can’t send messages to anyone other than themselves and members of their team. To do this the API has to compare the numbers they are sending to with those of their team members.

It will (falsely) say the numbers do not match if they are in a different format, eg 07700 900849 vs +447700900849

This commit uses the code we use elsewhere for formatting phone numbers to make sure that both numbers are in a consistent format before doing a comparison.

I have a strong preference for doing it this way, rather than formatting numbers before we store them:

1. https://en.wikipedia.org/wiki/Robustness_principle
2. It’s confusing to a user to see their own phone number formatted in a different way to that which they entered it, and the alternative, storing the phone number in two different formats, is grim

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -24,7 +24,7 @@ from sqlalchemy.exc import SQLAlchemyError
 from app.aws import s3
 from datetime import datetime
 from utils.template import Template
-from utils.recipients import RecipientCSV
+from utils.recipients import RecipientCSV, validate_phone_number, format_phone_number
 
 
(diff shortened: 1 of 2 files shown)


servingUpAces:
:+1:

PR #148, opened 2016-03-11
Error handing

servingUpAces:
Changed db queries to use one, which throws NoResultFound exception, this exception is dealt with in our error handlers. Now a lot of the if none checks can be removed.


mannickutd:
:+1:

PR #149, opened 2016-03-14
SES Callback testing

minglis:
- SES/AWS JSON is horrible and not valid.
- JSON in tests did not test accurately what it looked like in reality
- Using very odd looking bytes/strings as input into API which is more accurate


ashimali:
demoed to me :+1:

PR #156, opened 2016-03-15
Updated aws start/stop scripts to handle multiple start/stop scripts

minglis:
Primarily to handle the two celery processes

- On the environments we have start/stop scripts for the 3 API types

- Normal API
- Celery Workers
- Celery Beat

We now have a different upstart script for each, and there may be more than one on a box, so check for the upstart scripts and if present use them.

### DIFF
--- a/scripts/aws_start_app.sh
+++ b/scripts/aws_start_app.sh
@@ -1,4 +1,19 @@
 #!/bin/bash
 
-echo "Starting application"
-sudo service notifications-api start
+if [ -e "/etc/init/notifications-api.conf" ]
+then
+  echo "Starting api"
+  sudo service notifications-api start
+fi
+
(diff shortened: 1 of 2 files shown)


scripts/aws_start_app.sh
```diff
@@ -1,4 +1,19 @@
 #!/bin/bash
 
-echo "Starting application"
-sudo service notifications-api start
\ No newline at end of file
+if [ -e "/etc/init/notifications-api.conf" ]
+then
+  echo "Starting api"
+  sudo service notifications-api start
```
mannickutd: I would normally use this the other way round, run the script with sudo and then if you need to run something as a specific user use that in the command eg. http://unix.stackexchange.com/questions/190571/sudo-in-non-interactive-script

PR #158, opened 2016-03-16
Pass application name in from start scripts

minglis:
- allows logger to log as correct application

This allows the delivery api to set app name as 'delivery' to allow easier log filtering


servingUpAces:
Do we want to give the api app a name too?

servingUpAces:
:+1:

PR #162, opened 2016-03-17
Add platform_admin boolean on the User data model.

servingUpAces:


mannickutd:
A script to add a platform admin might be useful but for now this is fine.
:+1:

PR #164, opened 2016-03-17
Allowing overrides on a per environment basis

minglis:
In the production AWS account we run both staging and production instances. This means in practise that a single dynamo db instance/credstash implementation is the source of the secrets.

We need therefore to distinguish between properties for staging and those for live. To do this we use different named properties where appropriate - prefixed by environment.

there are two new config files - Live and Staging - that reference these new properties. To choose the right one we use the fact that each AWS instance has an 'environment' file present in the home folder. This contains the name of the environment it is deployed into. 

We can use this to choose the right config object to start the application.


aws_run_celery.py
```diff
@@ -2,11 +2,21 @@
 from app import notify_celery, create_app
 from credstash import getAllSecrets
 import os
+from config import configs
+
+default_env_file = '/home/ubuntu/environment'
+environment = 'live'
+
+if os.path.isfile(default_env_file):
+    environment_file = open(default_env_file, 'r')
```
mannickutd: with open(default_env_file, 'r') as environment_file:
    environment = environment_file.readline().strip()

This handles the the closing of the file as well. Python context managers :)

wsgi.py
```diff
@@ -2,13 +2,24 @@
 
 from app import create_app
 from credstash import getAllSecrets
+from config import configs
+
+
+default_env_file = '/home/ubuntu/environment'
+environment = 'live'
+
+if os.path.isfile(default_env_file):
```
mannickutd: Same see above.

wsgi.py
```diff
@@ -2,13 +2,24 @@
...
+    environment_file = open(default_env_file, 'r')
+    environment = environment_file.readline().strip()
 
 # on aws get secrets and export to env
 secrets = getAllSecrets(region="eu-west-1")
 for key, val in secrets.items():
```
mannickutd: os.environ.update(secrets)

ashimali:
apart from the above :+1:

aws_run_celery.py
minglis: Gotcha - done

wsgi.py
minglis: Done

PR #165, opened 2016-03-17
Added task for sending email verification links

ashimali:
On initial registration user is sent a link to click on to verify email address.

Left original email code endpoint in as it is still used for things like email change.

The work in this pr is needed for https://github.com/alphagov/notifications-admin/pull/293


servingUpAces:
:+1:

PR #176, opened 2016-03-21
Capture delivery outcomes

minglis:
Adds notification stats update into the callback process
- when a provider callback occurs and we update the status of the notification, also update the statistics table

Adds:
- Mapping object to the clients to handle mapping to various states from the response codes, this replaces the map.
- query lookup in the DAO to get the query based on response type / template type

Tests around rest class and dao to check correct updating of stats

Missing:
- multiple client callbacks will keep incrementing the counts of success/failure. This edge case needs to be handle in a future story.


app/clients/__init__.py
```diff
@@ -11,3 +11,28 @@ class Client(object):
...
     '''
     pass
+
+
+STATISTICS_REQUESTED = 'requested'
+STATISTICS_DELIVERED = 'delivered'
+STATISTICS_FAILURE = 'failure'
+
+
+class ClientResponse:
```
minglis: Base response class.

- Used to map status codes from SMS/Email providers back into Notify statuses

app/clients/email/aws_ses.py
```diff
@@ -1,22 +1,33 @@
...
-        "success": True,
-        "notify_status": 'delivered'
-    },
-    'Complaint': {
-        "success": False,
-        "notify_status": 'complaint'
-    }
-}
+
+class AwsSesResponses(ClientResponse):
```
minglis: SES Version of status mapper

app/clients/sms/firetext.py
```diff
@@ -6,26 +6,34 @@
...
-        "notify_status": 'failed'
-    },
-    '2': {
-        "firetext_message": 'Undelivered (Pending with Network)',
-        "success": False,
-        "notify_status": 'sent'
-    }
-}
+
+class FiretextResponses(ClientResponse):
```
minglis: Firetext verison. Note here the "pending" state doesn't update the stats table (None state)

app/dao/notifications_dao.py
```diff
@@ -49,20 +55,22 @@ def dao_create_notification(notification, notification_type):
...
 
-def update_notification_stats(notification, notification_type):
-    if notification_type == TEMPLATE_TYPE_SMS:
-        update = {
-            NotificationStatistics.sms_requested: NotificationStatistics.sms_requested + 1
-        }
-    else:
-        update = {
-            NotificationStatistics.emails_requested: NotificationStatistics.emails_requested + 1
+def update_query(notification_type, status):
```
minglis: Update queries mapped by type/outcome. Methods use notification type/status to pick correct update

app/notifications/rest.py
```diff
@@ -11,8 +11,8 @@
 )
```
minglis: Changes in this file are about using the new response objects defined in the clients.

servingUpAces:
:+1:

PR #184, opened 2016-03-24
Align notifications api with docs and other endpoints

ashimali:
Post endpoint to create notification and get endpoint to retrieve notification by id return 
data in shape more consistent with other api endpoints.


mannickutd:
:+1:

PR #186, opened 2016-03-30
Add view activity permission

servingUpAces:
Added view_activity as a new permission. All users will have this permission as a default permission. This will allow them to view activity, team members but not be able to edit or send messages.

Access_developer_docs has been removed as a valid permission - everyone is able to view the documentation. 

The https://github.com/alphagov/notifications-admin/pull/358 pull request is the required after this pull request.


mannickutd:
:+1:

PR #189, opened 2016-03-31
Remove email_safe function

servingUpAces:
-The admin app now sends the email from when creating a service and when updating the service name.
- This PR removes the need for the email_safe function. The api does not create the email_from field for the service.
- Tests were updated to reflect this change.
- Removed the validation in the schema - it was adding complexity, let the unique constraint on the db throw the exception. This should only ever happen on a race condition which seems unlikely (two people changing a service to the same name at the same time)
- Do no set debug=true to the test config. If debug=true it changes the behaviour of the error handlers, throwing the exception rather than returning a 500.


mannickutd:
:+1: 
Looks alright, was expecting a few more tests testing the collision of service names and email_from's that we experienced.

PR #196, opened 2016-04-04
Additional sort order by template name for template statistics.

ashimali:
Sort was previously on date only. This pr adds case insensitive additional ordering on template name.


mannickutd:
:+1:

mannickutd:
:+1:

PR #198, opened 2016-04-05
Successful notifications are deleted after a week now. All tests pass…

mannickutd:
…ing.


mannickutd:
Demoed to @servingUpAces

PR #201, opened 2016-04-05
Remove csv after process job is finished.

mannickutd:


mannickutd:
Demoed to @servingUpAces

PR #203, opened 2016-04-05
Catch sending to restricted recipients in Celery

quis:
The Celery `send_sms` and `send_email` tasks _could_ assume that all the recipients it gets are safe, because they have been checked either:
- when the admin app processes the CSV
- in the `/notifications/sms|email` endpoint

**However**, it’s probably safer to make the check again when the Celery task runs and passes the message off to the third party.

This also means that changing a service _back_ to restricted would have an effect on messages that were queued, as well as all subsequent messages.

This commit:
- restores the test that was removed here: https://github.com/alphagov/notifications-api/commit/e56aee5d1d77142b78434f3faeddfd2cb1a6e9bf#diff-e5627619e387fccc04783c32a23e6859L346
- adds a similar test for sending email while a service is in restricted mode
- adds checks back into the Celery tasks for sending email and SMS, using the `allowed_to_send_to` function that was introduced into utils in https://github.com/alphagov/notifications-utils/pull/16

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,3 +1,4 @@
+import itertools
 from datetime import datetime
 
 from flask import current_app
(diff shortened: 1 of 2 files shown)


servingUpAces:
:+1:

PR #204, opened 2016-04-06
Change sort order for templates

ashimali:
Instead of by name sort by date using full timestamp, so be most recently used at top.


servingUpAces:
:+1:

PR #205, opened 2016-04-07
Fix for date range required for template stats

ashimali:
It should always be last n days, whether or not there is data.


mannickutd:
:+1:

PR #225, opened 2016-04-14
Stop GOV.UK becoming a link in email clients

quis:
Implements (and depends on) https://github.com/alphagov/notifications-utils/pull/24

Works for:
- emails we send to users, eg invite, forgot password
- email templates that users have created

***

image

***

image


servingUpAces:
:+1:

PR #226, opened 2016-04-15
Simplify jwt

servingUpAces:
This pull request removes the need for the req and pay in the claims JWT.
Use the new version of the notifications-python-client. This version no longer adds the req and pay to the claims of the jwt.

The change is backward compatible so an older client that sends a JWT with the extra claims will pass authentication.
Once all the clients have been updated to not include the extra claims some updates to exclude them from the method signatures will happen as well.

https://www.pivotaltracker.com/story/show/116971293


ashimali:
:+1:

PR #230, opened 2016-04-15
Update master with all the staging changes for aws deployments.

servingUpAces:


quis:
Reproduced the files that should be changed by patching from the live branch. :+1:

PR #236, opened 2016-04-18
A slightly manual way of recording history.

ashimali:
To record history for a given class, you'd need to add the versioned mixin in models.py 
to the class you want to record history for.

Then create history model to mirror the versioned class with fields you want to record from the versioned class.

You can then use versioned decorator from dao_utils on the dao methods where your class is created or updated.

The history table records current state on create or update. This removes need to try and diff changes.

A consequence of this means that versioned objects should be given their id cols manually. By doing this the history models can be saved in the same transaction and rolled back with the versioned object in case of errors.


app/models.py
```diff
@@ -64,14 +73,30 @@ def check_password(self, password):
...
     db.Column('user_id', UUID(as_uuid=True), db.ForeignKey('users.id')),
     db.Column('service_id', UUID(as_uuid=True), db.ForeignKey('services.id')),
-
     UniqueConstraint('user_id', 'service_id', name='uix_user_to_service')
 
 )
 
 
-class Service(db.Model):
+class ServiceHistory(db.Model):
```
quis: Why can’t this inherit from the `Service` model and override the things that are different?

app/service/rest.py
```diff
@@ -1,3 +1,4 @@
+import uuid
```
quis: Why is this import being added?

app/service/rest.py
```diff
@@ -72,6 +73,9 @@ def create_service():
 
     data.pop('user_id', None)
 
+    # import pdb
```
quis: Why is this here?

app/vendor/sqlalchemy_history_meta.py
```diff
@@ -0,0 +1,261 @@
+"""
```
mannickutd: Is this file still needed?

quis: Are you still using this file?

quis:
I’ve left some comments on the code. Someone else should take a look at it as well, but I think the concept is good.

migrations/versions/0002_add_service_history.py
```diff
@@ -0,0 +1,49 @@
+"""empty message
```
servingUpAces: This upgrade does not work on a populated db.

app/dao/dao_utils.py
```diff
@@ -0,0 +1,34 @@
+from functools import wraps
+
```
mannickutd: I do like these decorators.
One scenario which we may have to alter their functionality in the future is to handle situations where we want to commit and situations where it is one of a many db operations and only on the last db operation do we want to commit. How do we handle the stacked decorated methods?

app/models.py
ashimali: Most of the fields are nullable, the main difference is the composite primary key of id and version, so I think inheritance wouldn't be the way to go for this model.

app/service/rest.py
ashimali: stray import

app/service/rest.py
ashimali: removed

app/vendor/sqlalchemy_history_meta.py
ashimali: No will remove it.

migrations/versions/0002_add_service_history.py
ashimali: Good point. Will do an update to set existing models version to 1 and also create a history record for each.

app/dao/dao_utils.py
ashimali: Yep there is a required order for the decorators.

tests/app/service/test_rest.py
```diff
@@ -212,15 +213,16 @@ def test_create_service(notify_api, sample_user):
...
     with notify_api.test_request_context():
         with notify_api.test_client() as client:
             data = {
                 'email_from': 'service',
                 'name': 'created service',
                 'message_limit': 1000,
                 'restricted': False,
-                'active': False
+                'active': False,
+                'created_by': str(fake_uuid)
```
mannickutd: Don't need to str the fake_uuid but doesn't matter for now.

mannickutd:
:+1:

PR #241, opened 2016-04-19
Allow filtering of notification stats by days

quis:
On the dashboard we want to show counts of notifications sent in the last 7 days, plus today. So the API endpoint needs to accept an argument to limit how many days worth of statistics it will return.

This is a bit fiddly at the DAO level because the date is just stored as a string.

https://www.pivotaltracker.com/story/show/117920839


ashimali:
Not sure why the day on NotificationsStatistics is a string. Why not change it to a date?

ashimali:
👍 
I'm happy to circle back and handle changing string day to proper date as a tech debt/bug fix so as not not hold this up.

PR #242, opened 2016-04-20
Update to processing the the response from MMG

servingUpAces:
MMG changed the datatype and the status codes they send us for the delivery receipts.
This PR accounts for that change.


mannickutd:
:+1:

PR #248, opened 2016-04-20
Update MMG responses so that 3 is success.

servingUpAces:
Status codes from MMG are:
2: UNDELIVERABLE Message is undeliverable.
3: DELIVERED Message is delivered.
4: EXPIRED Message is expired.
5: REJECTED Message is rejected.


mannickutd:
:+1:

PR #255, opened 2016-04-21
Added version history to api keys.

ashimali:
This needed a bit of change to create history to handle foreign keys better. 

There may yet be a better way of doing this that I have not found yet in sqlalchemy
docs.


app/dao/api_key_dao.py
```diff
@@ -1,19 +1,30 @@
...
+    versioned
+)
 
+
+@transactional
+@versioned
 def save_model_api_key(api_key, update_dict={}):
     if update_dict:
-        if update_dict['id']:
+        if update_dict.get('id'):
```
mannickutd: one line pop('id')

mannickutd:
:+1:

PR #256, opened 2016-04-21
Temp history of endpoint

ashimali:
This is a temporary endpoint to get history of service and api key updates. 

This will be surfaced on a simple log page on admin app until product team work out who/how/when etc. of viewing history data.


mannickutd:
:+1:

PR #258, opened 2016-04-25
Working tests and provider stats table.

mannickutd:
Fix for tests and import error.


app/schemas.py
```diff
@@ -58,6 +48,25 @@ def make_instance(self, data):
         return super(BaseSchema, self).make_instance(data)
 
 
+class CreatedBySchema(ma.Schema):
```
ashimali: May as well at this to ApiKey as well

tests/app/dao/test_templates_dao.py
```diff
@@ -11,12 +11,13 @@
 import pytest
 
 
-def test_create_template(sample_service):
+def test_create_template(sample_service, sample_user):
```
ashimali: Probably add an assert that template history record exists

ashimali:
You might want to add this to get_service_and_api_key_history in /service/rest.py and surface in admin app so Pete/Henry can sign off story.

app/schemas.py
mannickutd: Done

tests/app/dao/test_templates_dao.py
mannickutd: Separate tests added.

ashimali:
👍

PR #260, opened 2016-04-25
Use short dates when selection notifications for deletion.

ashimali:
This means we will retain notifications for a full week and not
delete records that are 7 x 24 hours older than the time of the run of
the deletion task.

Also the task only needs to run once a day now, so I have changed
the celery config for the deletion tasks.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -1,5 +1,5 @@
 import math
-from sqlalchemy import desc
+from sqlalchemy import desc, func
 
 from datetime import (
     datetime,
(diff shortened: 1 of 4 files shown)


mannickutd:
:+1:

app/dao/notifications_dao.py
```diff
@@ -257,18 +257,20 @@ def filter_query(query, filter_dict=None):
 
 
 def delete_notifications_created_more_than_a_day_ago(status):
```
servingUpAces: This is only used in tests, we could remove

PR #261, opened 2016-04-25
Added functionality to archive a template.

mannickutd:
Fixed migration file name.


migrations/versions/archive_template.py
```diff
@@ -0,0 +1,33 @@
...
+
+Revision ID: archive_template
+Revises: 0007_template_history
+Create Date: 2016-04-25 14:16:49.787229
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'archive_template'
+down_revision = '0007_template_history'
```
servingUpAces: Why is the migration script named without the version number?

servingUpAces:
👍

PR #262, opened 2016-04-25
Send emails with a friendly from name

quis:
image

It’s nicer to have emails with a sender name, as well as the raw email address.

Amazon SES can acheive this by using the format
```
"Sender name" <sender.name@domain.com>
```
— http://docs.aws.amazon.com/ses/latest/DeveloperGuide/email-format.html

We also have to remove all non-ASCII characters from the sender name, because SMTP only supports 7-bit ASCII:

> A field name MUST be composed of printable US-ASCII characters (i.e., characters that have values between 33 and 126, inclusive), except colon.

— http://www.ietf.org/rfc/rfc5322.txt

We use the service name as the sender name when:
- sending emails from the API
- sending emails from a CSV file

We use GOV.UK Notify as the sender name when:
- sending invitation emails
- sending password reset emails

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -173,7 +173,11 @@ def process_job(job_id):
             send_email.apply_async((
                 str(job.service_id),
                 str(create_uuid()),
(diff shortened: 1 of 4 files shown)


mannickutd:
:+1:

PR #263, opened 2016-04-25
Remove send email code task

quis:
We don’t send email codes any more.

No corresponding API client method here: https://github.com/alphagov/notifications-admin/blob/master/app/notify_client/user_api_client.py

So it’s definitely not being used.


servingUpAces:
👍   Always good to delete some code :)

PR #265, opened 2016-04-26
Fix test for duplicate service name

quis:
There is a uniqueness constraint on service `name` and `email_from`.

When you try to insert a row which has _both_ constraints, there is no guarantee which one the operation will fail on.

This means that, when handling the exception, service `name` is not reliably available, because sometimes the operation fails on the `email_from` constraint instead.

This caused the tests to fail non-deterministically because they were looking for the service `name` in the error message. As a fix, this commit does two things:

1. Return either the service `name` or `email_from` in the error message, depending which is available.

2. Modify the test to pass on _either_ of the two possible error messages.

This is not ideal, but I can’t think of a way to maintain the original behaviour, and have reliably passing tests.

### DIFF
--- a/app/errors.py
+++ b/app/errors.py
@@ -64,6 +64,8 @@ def register_errors(blueprint):
                 'duplicate key value violates unique constraint "services_email_from_key"' in e.orig.pgerror):
             return jsonify(
                 result='error',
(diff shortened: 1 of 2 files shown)


servingUpAces:
👍   Thank you

mannickutd:
:+1:

PR #266, opened 2016-04-26
Add created by for jobs

servingUpAces:
Save the user id of the person that created the job

https://www.pivotaltracker.com/story/show/117144359


ashimali:
👍

PR #267, opened 2016-04-27
Endpoint for recording events in api.

ashimali:
An event has an id, a type and a blob of json attached.


servingUpAces:
👍 
Presumably there is a plan for a GET endpoint?

ashimali:
@servingUpAces yep will do that bit last as a temp measure

PR #268, opened 2016-04-27
Align date range for template + notification stats

quis:
We were using two different queries to filter template stats to the past 7 days, plus today.

Since we’re storing both as short dates, we can now use the same query for both.

0      | 1       | 2         | 3        | 4      | 5        | 6      | 7
-------|---------|-----------|----------|--------|----------|--------|-------
Monday | Tuesday | Wednesday | Thursday | Friday | Saturday | Sunday | Monday

So if we are on Monday, the stats should include today, plus everything back to last Monday.

Previously the template stats query was only going back to the Tuesday.

This should mean the numbers on the dashboard always line up.


ashimali:
👍

PR #271, opened 2016-04-27
Hard sms limits added to the api for send_notification and create_tem…

mannickutd:
…plate, update_template. All tests passing.


ashimali:
👍

PR #274, opened 2016-04-28
Added limit_days filter to get notifications dao.

ashimali:
This pr is needed for https://github.com/alphagov/notifications-admin/pull/519

Also added config for deletion task so that it will run at midnight, 1am and 2am to cover utc and utc+1 servers, also clock skew and possibility of servers time zones not being changed in sync. In longer term it might be worth running this job more frequently with a limit on delete in case of lots of records to delete.


servingUpAces:
👍  demoed and reviewed

PR #275, opened 2016-04-29
This pull request fixes a bug in authentication.

servingUpAces:
If the service does not have any api keys, there would be an error but it was not formed well.


mannickutd:
:+1:
Just a side note, if we don't need to generate a uuid for the test we can speed up the test suite by just using a fake one.

PR #279, opened 2016-04-29
Add delivery status endpoint

ashimali:


servingUpAces:
I suspect we need to add the new endpoint to list of no_auth_req in the required_authentication method in init_app

ashimali:
@servingUpAces already is

servingUpAces:
👍

PR #280, opened 2016-04-29
Process CSV files with column headers of any case

quis:
Implements and depends on: https://github.com/alphagov/notifications-utils/pull/27

Should be merged and deployed before the same change in the admin app.

Object passed to Celery tasks must be JSON serialisable, hence the `dict` comprehension to convert from a `Columns` object.


minglis:
👍

PR #285, opened 2016-05-04
Update python client to version 1.0.0.

servingUpAces:
This version of the client removed the request method, path and body from the encode and decode methods.
The biggest changes here is to the unit tests.


mannickutd:
:+1:

PR #286, opened 2016-05-04
Group by Notification Statistics added and all tests working.

mannickutd:


quis:
Seems to be working… 

image

quis:
Happy with the output of this. Would be good if someone else could look at the code.

PR #289, opened 2016-05-06
Template history endpoint added. All tests passing.

mannickutd:
The private api will support retrieving a template history endpoint and for the existing GET template endpoints will now return a list of versions for each template.


ashimali:
👍

PR #290, opened 2016-05-09
Get all versions for a template endpoint added.

mannickutd:


ashimali:
👍 pending green build

PR #295, opened 2016-05-11
Primary provider

minglis:
Adds the notion of priority to the delivery partners.

- New DB table to store the data
- Updates the other provider tables to use the new table as root table for provider models.
- Celery tasks use the primary provider in all delivery tasks

Tests:
- will fail build if any providers in the database who are not modelled in code
- new providers can be in code and not in the database

Reference Data:
- Providers are created in a migration script


app/__init__.py
```diff
@@ -48,6 +50,7 @@ def create_app(app_name=None):
     aws_ses_client.init_app(application.config['AWS_REGION'])
     notify_celery.init_app(application)
     encryption.init_app(application)
+    clients.init_app(sms_clients=[firetext_client, mmg_client], email_clients=[aws_ses_client])
```
minglis: We store all the clients in an object, built at start up. Clients in the database MUST be in the object, mappings are between 'identifier' in the database and 'name' in the client object.

app/celery/tasks.py
```diff
@@ -206,7 +203,8 @@ def remove_job(job_id):
 def send_sms(service_id, notification_id, encrypted_notification, created_at):
     notification = encryption.decrypt(encrypted_notification)
     service = dao_fetch_service_by_id(service_id)
-    client = mmg_client
+
+    provider = provider_to_use('sms', notification_id)
```
minglis: All tasks use this method to pick the correct provider at run time. Not cached.

app/celery/tasks.py
```diff
@@ -454,3 +460,17 @@ def service_allowed_to_send_to(recipient, service):
             [user.mobile_number, user.email_address] for user in service.users
         )
     )
+
+
+def provider_to_use(notification_type, notification_id):
```
minglis: Helper method to pick the right provider. If no active providers blow up with an exception.

app/dao/notifications_dao.py
```diff
@@ -125,15 +126,15 @@ def dao_create_notification(notification, notification_type, provider):
     update_count = db.session.query(ProviderStatistics).filter_by(
         day=date.today(),
         service_id=notification.service_id,
-        provider=provider
+        provider_id=provider.id
```
minglis: All now use the foreign key not the identifier.

app/dao/provider_rates_dao.py
```diff
@@ -1,9 +1,11 @@
...
 from app.dao.dao_utils import transactional
 
 
 @transactional
-def create_provider_rates(provider, valid_from, rate):
-    provider_rates = ProviderRates(provider=provider, valid_from=valid_from, rate=rate)
+def create_provider_rates(provider_identifier, valid_from, rate):
+    provider = ProviderDetails.query.filter_by(identifier=provider_identifier).one()
+
+    provider_rates = ProviderRates(provider_id=provider.id, valid_from=valid_from, rate=rate)
```
minglis: Foreign key reference

app/provider_details/rest.py
```diff
@@ -0,0 +1,42 @@
...
+def update_provider_details(provider_details_id):
+    fetched_provider_details = get_provider_details_by_id(provider_details_id)
+
+    current_data = dict(provider_details_schema.dump(fetched_provider_details).data.items())
+    current_data.update(request.get_json())
+    update_dict, errors = provider_details_schema.load(current_data)
+    if errors:
+        return jsonify(result="error", message=errors), 400
+
+    if "identifier" in request.get_json().keys():
```
minglis: We can't change the identifier on an API call. This would break the relationships with the code.

migrations/versions/0011_ad_provider_details.py
```diff
@@ -0,0 +1,74 @@
...
+
+    op.add_column('provider_rates', sa.Column('provider_id', postgresql.UUID(as_uuid=True), nullable=True))
+    op.create_index(op.f('ix_provider_rates_provider_id'), 'provider_rates', ['provider_id'], unique=False)
+    op.create_foreign_key(None, 'provider_rates', 'provider_details', ['provider_id'], ['id'])
+    op.add_column('provider_statistics', sa.Column('provider_id', postgresql.UUID(as_uuid=True), nullable=True))
+    op.create_index(op.f('ix_provider_statistics_provider_id'), 'provider_statistics', ['provider_id'], unique=False)
+    op.create_foreign_key(None, 'provider_statistics', 'provider_details', ['provider_id'], ['id'])
+
+    op.execute(
+        "INSERT INTO provider_details (id, display_name, identifier, priority, notification_type, active) values ('{}', 'MMG', 'mmg', 10, 'sms', true)".format(str(uuid.uuid4()))
```
minglis: Data inserts and migrations to move to the new model.

requirements_for_test.txt
```diff
@@ -1,6 +1,6 @@
 -r requirements.txt
 pep8==1.5.7
-pytest==2.8.1
+pytest==2.8.3
```
minglis: Pytest bumped - stops those 'refs' errors when the model is broken.

app/celery/tasks.py
```diff
@@ -454,3 +460,17 @@ def service_allowed_to_send_to(recipient, service):
...
+    active_providers_in_order = [
+        provider for provider in get_provider_details_by_notification_type(notification_type) if provider.active
+    ]
+
+    if len(active_providers_in_order) == 0:
```
mannickutd: Java if statement
if not active_providers_in_order:

app/models.py
```diff
@@ -191,13 +191,18 @@ class Template(db.Model, Versioned):
...
+
 
 class ProviderStatistics(db.Model):
     __tablename__ = 'provider_statistics'
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     day = db.Column(db.Date, nullable=False)
-    provider = db.Column(db.Enum(*PROVIDERS, name='providers'), nullable=False)
+    provider_id = db.Column(UUID(as_uuid=True), db.ForeignKey('provider_details.id'), index=True, nullable=False)
+    provider_stats_to_provider = db.relationship(
```
mannickutd: I would prefer if it would keep the existing name 'provider'.
Most of our history, marshmallow expect 'provider' and then a 'provider_id'.

app/provider_details/rest.py
mannickutd: Why not do this check in the schema?

minglis: Thinking it's more request validation not object validation?

app/celery/tasks.py
minglis: :-)

app/models.py
minglis: notifications table has template_id / service_id and so on, is this not how we identify foreign keys?

PR #298, opened 2016-05-11
Add created_by for jobs

servingUpAces:
Add a load only and dump only for the created_by attribute of the JobSchema

https://www.pivotaltracker.com/story/show/117144359


mannickutd:
:+1:

PR #301, opened 2016-05-11
Add server_commands and update readme.md

mannickutd:
The server commands is a file to run management commands on the servers.
I will look to remove the db.py file when I get back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,3 +73,19 @@ scripts/run_celery.sh
 ```
 scripts/run_celery_beat.sh
 ```
+
+
+## To remove functional test data
+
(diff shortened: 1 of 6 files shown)


ashimali:
Might be easier rather than take command line options of prefix, just look for any users that have email with pattern notify-tests-preview[+someuuid]@digital.cabinet-office.gov.uk, and take those users are where to start clearing up. In other words find service, templates, verify keys, events etc that are associated directly or indirectly with that user.

mannickutd:
I will update the commands from Adams suggestion and include deletion of history for a service.

ashimali:
👍

PR #310, opened 2016-05-16
Add template version for jobs and notifications

servingUpAces:
Use the template version at the time the notification is created or at the time the job is created.

Update notifications/sms|email endpoint to send the template version to the queue.
Update the process_job celery talk to send the template version to the queue.
When the send_sms|send_email task runs it will get the template by id and version.

Created a data migration script to add the template_vesion column for jobs and notifications.
The existing jobs and notifications are given the template_version of the current template.
There is a chance this is the wrong template version, but deemed okay since the application is not live.

Create unit test for the dao_get_template_versions method.
Rename /template/<id>/version to /template/<id>/versions which returns all versions for that template id and service id.


app/celery/tasks.py
```diff
@@ -313,8 +316,8 @@ def send_email(service_id, notification_id, from_address, encrypted_notification
         except EmailClientException as e:
             current_app.logger.exception(e)
             notification_db_object.status = 'failed'
+            dao_update_notification(notification_db_object)
```
quis: Why has this line moved inside the exception?

servingUpAces: That update statement is to update the status when there is an exception. If it is outside the exception block we are always updating the notification but we do not need it, as we update the notification after the notification is successfully sent to client provider.

quis:
- 👍 seen demo on Becca’s machine
- 👍 reviewed code
- 👍 run db upgrade, downgrade, upgrade locally and don’t seem to have any missing data in `template_version` column on `notifications`, `jobs`, or `templates` tables

PR #313, opened 2016-05-16
There is a problem where columns on the templates table were not set.

servingUpAces:
It is also discovered that columns that have a default value and use the version mixin must set the value when creating the db object before the insert otherwise the history table will be missing the default value.
Updated the templates_history.created_by_id with a value where missing, using the current version of the template for this value.
Update templates_history.archived to false. This is okay as we do not yet have a way to set this value to true.
Removed the versions attribute from the TemplateSchema, there is not a need for this column.


ashimali:
👍

PR #318, opened 2016-05-17
New notification status types

servingUpAces:
There are 3 new failure types: technical-failure, temporary-failure, permanent-failure.
Technical-failure: meaning notify was unable to send the message to the client provider.
Temporary-failure: meaning the client provider is unable to send the message now, but will try again. For example the mobile phone is off but it is a valid number
Permanent-failure: meaning the client provider did not send the message because the to address to mobile number is invalid.


ashimali:
👍

PR #321, opened 2016-05-17
Email failures

servingUpAces:
Update ses callback to interpret hard and soft bounces.  
If the notification has a status == sending then update the status otherwise do not update the status.
In other words do not change the status more than once.


ashimali:
👍

PR #326, opened 2016-05-18
Fix test to check for decrypted dictionary value.

mannickutd:
Refers to the pivotal task #119746991


mannickutd:
@servingUpAces can you please have a look.

quis:
👍

PR #332, opened 2016-05-19
Update template json to return created_at and updated_at. Also fixed …

mannickutd:
…a bug where updated_at was not being added to the templates_history model.


servingUpAces:
👍

PR #334, opened 2016-05-20
Make the get_user_by_email_address case insensitive.

servingUpAces:


minglis:
Sweet OK, 👍

Can you pop a story in to address the underlying data issues - outcome being that the data is stored in a consistant manner.

PR #335, opened 2016-05-20
Data migration script to fix templates_history.updated_at.

servingUpAces:
There was a bug that was not setting the updated_at for the history table.


servingUpAces:
Has been demoed and reviewed to Leo. All good to go.

PR #338, opened 2016-05-20
Only update status when status is sending.

servingUpAces:
If the sms client provider sends more than one delivery receipt only take the update for the the first one.

Only update the notification.status if status = sending.

https://www.pivotaltracker.com/story/show/119377985


quis:
Code looks 👌. Just needs a demo and good to go.

quis:
👍 To go to preview.

PR #340, opened 2016-05-23
only return unarchived templates when getting all templates

leohemsted:


app/dao/templates_dao.py
```diff
@@ -41,8 +41,10 @@ def dao_get_template_by_id(template_id, version=None):
 
 
 def dao_get_all_templates_for_service(service_id):
+    # when getting all templates only return unarchived templates
```
quis: I don’t think you need this comment, it’s clear from the code what it’s doing.

tests/app/dao/test_templates_dao.py
```diff
@@ -140,6 +140,30 @@ def test_get_all_returns_empty_list_if_no_templates(sample_service):
...
+        service=sample_service,
+        archived=False
+    )
+    archived_template = create_sample_template(
+        notify_db,
+        notify_db_session,
+        template_name='Archived Template',
+        service=sample_service
+    )
+    # sample_template fixture uses dao, which forces archived = False at creation.
```
quis: This, on the other hand, is a useful comment.

quis:
One nitpicky comment, otherwise 👍

PR #341, opened 2016-05-23
Prevent creating jobs for deleted templates

leohemsted:


app/job/rest.py
```diff
@@ -52,6 +52,9 @@ def create_job(service_id):
         "service": service_id
     })
     template = dao_get_template_by_id(data['template'])
+    if template.archived:
+        return jsonify(result="error", message="Template has been deleted"), 400
```
mannickutd: Can this 400 look like the other 400 error messages

message = {'field with problem': ['problem 1', 'problem 2']}

Because this is kind of object problem rather than field problem, use the object name as the dictionary key.

mannickutd:
Is it also possible to get the restriction on the public api as well
/notifications/<string:notification_type>

mannickutd:
:+1:

PR #344, opened 2016-05-24
Send emails using aws send_raw_email

mannickutd:
[DO NOT MERGE] - Example of how to attach headers to email sent through SES.
http://boto3.readthedocs.io/en/latest/reference/services/ses.html#SES.Client.send_raw_email
http://docs.aws.amazon.com/ses/latest/APIReference/API_SendRawEmail.html


minglis:
Failed spike - this doesn't answer our issues. :-(

PR #347, opened 2016-05-24
Add a limit days query param for get all jobs.

servingUpAces:
https://www.pivotaltracker.com/story/show/119369811


quis:
Very thorough tests 👍

PR #349, opened 2016-05-25
Remove the status from the group by clause of the update statement.

servingUpAces:
This script has been applied to all environments except live, this can be pushed through to live without issue because of how alembic works.


minglis:
👍

PR #352, opened 2016-05-26
notifications/statistics returns when no stats exist for today

leohemsted:
moved filtering from WHERE to JOIN ON so that when join suceeds but filter fails, we dont lose the service's data from the results set


mannickutd:
:+1:

PR #353, opened 2016-05-27
Permanent failure firetext

servingUpAces:
This pull request will handle the way Firetext sends us delivery receipts for the message that we send them. 
https://www.pivotaltracker.com/story/show/119376873
- Mark a declined message from Firetext as permanent-failure rather than failed.
- Firetext does not have a status code for temporary-failure.  …
In order to set a message as temporary-failure, we check if it is in pending status first.
Otherwise a delivery receipt for failure is set to permanent failure.
- Make update notification status transactional.
- Refactor methods for reusability.
- Added a comment to explain how the delivery receipt will affect the notification status.


app/models.py
```diff
@@ -345,7 +345,7 @@ class Notification(db.Model):
         nullable=True,
         onupdate=datetime.datetime.utcnow)
     status = db.Column(
-        db.Enum(*NOTIFICATION_STATUS_TYPES, name='notification_status_type'), nullable=False, default='sending')
+        db.Enum(*NOTIFICATION_STATUS_TYPES, name='notify_status_types'), nullable=False, default='sending')
```
mannickutd: Why is the name changed?

mannickutd:
:+1:

PR #354, opened 2016-05-31
Reorder updates to the notification / stats tables to match create.

minglis:
First pass at trying to address deadlocks

- ensure updates are the same order on the seperate transactions. 

- previously job was first on create and last on update, causing transactions to block

- unable to fully test locally so pull request for testing on preview.


quis:
LGTM 👍

PR #357, opened 2016-05-31
Add research mode to service/history tables

minglis:
- added as a nullable boolean column.

This pull request adds the column in a non-destructive way
- actual implementation in a subsequent pull request


mannickutd:
:+1:

PR #358, opened 2016-05-31
Profile library added, tests now run up to 10% quicker.

mannickutd:
Add command to run tests script.


quis:
- +607 −454 lines changed
- many naming changes
- formatting changes
- tests deleted with no explanation

This could really benefit from being split into multiple commits. 

`git reset head~`

`git add --patch`

http://nuclearsquid.com/writings/git-add/

PR #361, opened 2016-06-01
Check for an update before saving template

servingUpAces:
Do not create a new version of the template if there is no change to the template.

https://www.pivotaltracker.com/story/show/120427863

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -59,14 +59,18 @@ def update_template(service_id, template_id):
     fetched_template = dao_get_template_by_id_and_service_id(template_id=template_id, service_id=service_id)
 
     current_data = dict(template_schema.dump(fetched_template).data.items())
-    current_data.update(request.get_json())
-    current_data['content'] = _strip_html(current_data['content'])
+    update_template = dict(template_schema.dump(fetched_template).data.items())
+    update_template.update(request.get_json())
+    update_template['content'] = _strip_html(update_template['content'])
+    # Check if there is a change to make.
+    if current_data == update_template:
+        return jsonify(data=update_template), 200
 
-    update_dict, errors = template_schema.load(current_data)
+    update_dict, errors = template_schema.load(update_template)
     if errors:
         return jsonify(result="error", message=errors), 400
     over_limit, json_resp = _content_count_greater_than_limit(
-        current_data['content'],
+        update_template['content'],
         fetched_template.template_type)
     if over_limit:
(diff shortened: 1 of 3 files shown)


tests/app/template/test_rest.py
```diff
@@ -444,3 +426,20 @@ def test_should_return_all_template_versions_for_service_and_template_id(notify_
...
+            auth_header = create_authorization_header()
+            data = {
+                'template_type': sample_template.template_type,
+                'content': sample_template.content,
+            }
+            resp = client.post('/service/{}/template/{}'.format(sample_template.service_id, sample_template.id),
+                               data=json.dumps(data),
+                               headers=[('Content-Type', 'application/json'), auth_header])
+            assert resp.status_code == 200
+    from app.dao.templates_dao import dao_get_template_by_id
```
mannickutd: Pep 8 says:
"Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants."

app/template/rest.py
```diff
@@ -59,14 +59,18 @@ def update_template(service_id, template_id):
...
-    current_data.update(request.get_json())
-    current_data['content'] = _strip_html(current_data['content'])
-
-    update_dict, errors = template_schema.load(current_data)
+    update_template = dict(template_schema.dump(fetched_template).data.items())
+    update_template.update(request.get_json())
+    update_template['content'] = _strip_html(update_template['content'])
+    # Check if there is a change to make.
+    if current_data == update_template:
+        return jsonify(data=update_template), 200
```
mannickutd: And if you are feeling really 'http compliant' 304 is probably a better response code.

app/template/rest.py
```diff
@@ -115,3 +118,13 @@ def get_template_versions(service_id, template_id):
...
+
+
+def _template_has_not_changed(current_data, updated_template):
+    if (current_data['name'] == updated_template['name'] and
+            current_data['content'] == updated_template['content'] and
+            current_data['subject'] == updated_template['subject']and
+            current_data['archived'] == updated_template['archived']):
+        return True
+    else:
+        return False
```
quis:
```python
return all(
  current_data[key] == updated_template[key]
  for key in ('name', 'content', 'subject', 'archived')
)
```

…if you really want to be Pythonic

servingUpAces: ok - good idea, thanks

quis:
👍

PR #362, opened 2016-06-01
Provider stats only updated if the provider successfully sends the me…

mannickutd:
…ssage.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -281,24 +281,32 @@ def dao_update_notification(notification):
 def update_notification_after_sent_to_provider(id_, notification_type, provider_name, reference=None):
     provider = ProviderDetails.query.filter_by(identifier=provider_name).one()
     notification = Notification.query.filter(Notification.id == id_).one()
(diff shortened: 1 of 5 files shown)


leohemsted:
👍

app/dao/notifications_dao.py
```diff
@@ -176,22 +176,6 @@ def dao_create_notification(notification, notification_type, provider_identifier
...
-    if update_count == 0:
-        provider_stats = ProviderStatistics(
-            day=notification.created_at.date(),
-            service_id=notification.service_id,
-            provider_id=provider.id,
-            unit_count=1 if notification_type == TEMPLATE_TYPE_EMAIL else get_sms_fragment_count(
-                notification.content_char_count))
-        db.session.add(provider_stats)
-
     db.session.add(notification)
```
minglis: Can we ensure that the notification update is the last thing that happens? Deadlocks now prey on my mind.

minglis: I'm doing this now.

PR #363, opened 2016-06-01
send_sms tests

leohemsted:
cleaned up the send_sms functions in firetext.py and mmg.py, and wrote some tests to validate current behaviour


servingUpAces:
👍

PR #368, opened 2016-06-02
caching in travis

leohemsted:
[good to merge]

https://docs.travis-ci.com/user/caching/

not quite the speed improvements i was hoping for (since we're still spending quite a bit of time, by the looks of it, installing C modules), but a few seconds is a few seconds.

new speed: image


quis:
👍

PR #371, opened 2016-06-02
Add a service for Notify to use to send messages

servingUpAces:
Add a service that will be used to send Notify's messages.
Including sms code, email verification, invitation emails, and password reset emails
Added the service id and template ids to config.
Small change to dao_create_service to use id if set.

First PR for https://www.pivotaltracker.com/story/show/120282429

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -47,8 +47,6 @@ def dao_create_service(service, user):
     from app.dao.permissions_dao import permission_dao
     service.users.append(user)
     permission_dao.add_default_service_permissions_for_user(user, service)
-    if not service.id:
-        service.id = uuid.uuid4()  # must be set now so version history model can use same id
     service.id = uuid.uuid4()  # must be set now so version history model can use same id
     service.research_mode = False
     db.session.add(service)
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -47,6 +47,8 @@ def dao_create_service(service, user):
     from app.dao.permissions_dao import permission_dao
     service.users.append(user)
     permission_dao.add_default_service_permissions_for_user(user, service)
+    if not service.id:
+        service.id = uuid.uuid4()  # must be set now so version history model can use same id
     service.id = uuid.uuid4()  # must be set now so version history model can use same id
     service.research_mode = False
     db.session.add(service)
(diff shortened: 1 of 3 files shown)


app/dao/services_dao.py
```diff
@@ -47,6 +47,8 @@ def dao_create_service(service, user):
     from app.dao.permissions_dao import permission_dao
     service.users.append(user)
     permission_dao.add_default_service_permissions_for_user(user, service)
+    if not service.id:
+        service.id = uuid.uuid4()  # must be set now so version history model can use same id
     service.id = uuid.uuid4()  # must be set now so version history model can use same id
```
leohemsted: this branch doesn't matter since service.id is always overwritten

migrations/versions/0025_notify_service_data.py
```diff
@@ -0,0 +1,119 @@
...
+
+user_id= '6af522d0-2915-4e52-83a3-3690455a5fe6'
+service_id = 'd6aa2c68-a2d9-4437-ab19-3ae8eb202553'
+
+
+def upgrade():
+    password = hashpw(str(uuid.uuid4()))
+    op.get_bind()
+    user_insert = """INSERT INTO users (id, name, email_address, created_at, failed_login_count, _password, mobile_number, state, platform_admin)
+                     VALUES ('{}', 'Notify service user', 'this-does-not-work@this.does.not.work.gov.uk', '{}', 0,'{}', '+441234123412', 'active', False)
```
minglis: Can we make the email address more instructive of the reason it exists (notify-service-user@digital.cabinet-office or something. Might not work, but when I'm scanning the DB in a year might be useful.

migrations/versions/0025_notify_service_data.py
```diff
@@ -0,0 +1,119 @@
...
+    op.execute(service_history_insert.format(service_id, datetime.utcnow(), user_id))
+    service_insert = """INSERT INTO services (id, name, created_at, active, message_limit, restricted, research_mode, email_from, created_by_id, reply_to_email_address, version)
+                        VALUES ('{}', 'Notify service', '{}', True, 1000, False, False, 'notify@digital.cabinet-office.gov.uk',
+                        '{}', 'notify@digital.cabinet-office.gov.uk', 1)
+                    """
+    op.execute(service_insert.format(service_id, datetime.utcnow(), user_id))
+    user_to_service_insert = """INSERT INTO user_to_service (user_id, service_id) VALUES ('{}', '{}')"""
+    op.execute(user_to_service_insert.format(user_id, service_id))
+
+    permission_insert = """INSERT INTO permissions (id, service_id, user_id, permission, created_at)
```
minglis: Does the user need any permissions for the service?

servingUpAces: Probably not since we will not have a way to log in as that user

migrations/versions/0025_notify_service_data.py
servingUpAces: Sure. Is the service name ok?

PR #372, opened 2016-06-02
Delivery code statsd

leohemsted:
for working out what firetext and mmg actually send us

Also I fixed a test that tested email notifications that had responses from firetext (!)


servingUpAces:
👍

PR #378, opened 2016-06-06
Order templates by newest created first

quis:
When you add a new template, it’s probably the one that you want to do subsequent stuff with. But it’s also helpful to see the template in context (with its siblings) to understand that there are multiple
templates. So we don’t want to do what we do in https://github.com/alphagov/notifications-admin/pull/648
for adding a new template.

But we _can_ make your brand-new template appear first by always ordering by when the template was created.

This also removes the confusion caused by having `updated_at` affecting order, and causing the templates to move around all the time.


mannickutd:
:+1:

PR #379, opened 2016-06-06
Twilio config variables removed.

mannickutd:


mannickutd:
@servingUpAces Might be an idea to clean up credstash Twilio variables if they are still floating around.

servingUpAces:
👍

PR #383, opened 2016-06-06
Use notify to send sms codes

servingUpAces:
- Update /sms-code endpoint to use the GOV.UK Notify service and it's sms code template, send the sms code via the send_sms task. 
- The GOV.UK Notify service and template data was added in a pervious merge.

Part of story https://www.pivotaltracker.com/story/show/120282429


ashimali:
👍 demo seen by me.

PR #384, opened 2016-06-06
API receives full set of permission names

ashimali:
API receives full set of permissions names on create of user invite. This is instead of mapping from permission groups to individual permissions on user creation.

For example instead of posting 'send_messages' and mapping to 'send_emails', 'send_texts', 'send_letters', the latter are saved from the outset against the user invite, and then posted to back to api from admin to create the user.

This fixes:
https://www.pivotaltracker.com/story/show/120829927

PR relies on:
https://github.com/alphagov/notifications-admin/pull/652


servingUpAces:
👍

ashimali:
Will merge on 07.06.2016 as user testing happening at the moment and this needs to be deployed to all environments.

PR #390, opened 2016-06-07
add template_statistics endpoint for specific template

leohemsted:
`/service/<service_id>/template-statistics/<template_id>`
still requires service-id just to try and keep api tree cleaner


mannickutd:
:+1:

PR #392, opened 2016-06-07
[WIP] Permissions being removed

ashimali:
This is marked as WIP for the bug as I still need to do a migration script for existing broken accounts.

When adding a user new with permissions to a service, the permissions
dao was deleting all permissions for that user (regardless of service
id) as the last filter on the permissions dao get_query method won.

I've added a replace flag to the set_user_service_permission method
so that it can handle adding new users + permissions and editing
of existing users' permissions.

Also by pass the get_query method until it can be refactored to work
correctly.

For now execute the filter query directly on the model.

### DIFF
--- a/app/dao/permissions_dao.py
+++ b/app/dao/permissions_dao.py
@@ -32,6 +32,8 @@ class PermissionDAO(DAOClass):
     class Meta:
         model = Permission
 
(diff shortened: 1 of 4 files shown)


servingUpAces:
👍  reviewed and demoed

PR #399, opened 2016-06-09
fix template api tests being inconsistent by adding ordering

leohemsted:
they were previously returning unordered, which broke the build (once and arbitrarily)

i tried to use freezegun to make the updated_at times predictable but didn't have any luck with that, but separate commits should be enough


mannickutd:
:+1:

PR #400, opened 2016-06-09
Remove task send_sms_code.

servingUpAces:
We use a notify service to send the sms code via using a template.

https://www.pivotaltracker.com/story/show/120282429


ashimali:
👍

PR #402, opened 2016-06-09
fix firetext research mode requests

leohemsted:
requests converts dicts into query parameters anyway if you don't specify so dont try and encode the data ourselves, also hardened up tests


tests/app/celery/test_research_mode_tasks.py
```diff
@@ -20,6 +20,8 @@ def test_make_mmg_callback(notify_api, rmock):
     send_sms_response("mmg", "1234", "07811111111")
 
     assert rmock.called
+    assert rmock.request_history[0].url == 'http://localhost:6011/notifications/sms/mmg'
```
mannickutd: Hard coded values (shudder)

mannickutd:
:+1:

PR #405, opened 2016-06-10
exclude notifications from job

leohemsted:
was being lazily loaded to get notification ids, so every time a job is loaded it would potentially select thousands of notifications from the database


mannickutd:
:+1:

PR #410, opened 2016-06-13
Better error logging

ashimali:
Rest methods that explicitly return errors by pass the error handlers as they do not raise exceptions.

Introduced a simple exception that contains error messages and status
code that can be used rather than return json + status code from rest
methods directly.

The handler in errors for this exception can then log the error before
returning json.


servingUpAces:
👍

PR #411, opened 2016-06-13
Use notify to send email verification links

servingUpAces:
This PR sends the registration email through the GOV.UK Notify service, using the template created for that message.


quis:
Seen a demo, IRL comments addressed, 👍

PR #412, opened 2016-06-13
Safely change task signature - first deploy

servingUpAces:
Create a new task called send_email_v2 so that I can get rid of the from_address in the signature.

This is done to make sure we do not lose any messages in the queue during deployment.


app/celery/tasks.py
```diff
@@ -290,6 +290,11 @@ def send_sms(service_id, notification_id, encrypted_notification, created_at):
 
 @notify_celery.task(name="send-email")
 def send_email(service_id, notification_id, from_address, encrypted_notification, created_at, reply_to_addresses=None):
+    send_email_v2(service_id, notification_id, encrypted_notification, created_at, reply_to_addresses=None)
```
leohemsted: this should just be passing through `reply_to_addresses` without the `=None`

PR #413, opened 2016-06-13
Test for filtering a job by notification status

quis:
It’s going to be useful to see all the notifications for a job that are failed/delivered/etc.

The API seems to support this behaviour already, but it doesn’t seem to be tested.

This commit adds some tests for the DAO and REST layers.

### DIFF
--- a/tests/app/notifications/test_rest.py
+++ b/tests/app/notifications/test_rest.py
@@ -2,6 +2,7 @@ from datetime import datetime
 import uuid
 import random
 import string
+import pytest
 import app.celery.tasks
 from mock import ANY
 from app import encryption
@@ -12,7 +13,7 @@ from tests.app.conftest import sample_service as create_sample_service
 from tests.app.conftest import sample_email_template as create_sample_email_template
 from tests.app.conftest import sample_template as create_sample_template
 from flask import (json, current_app, url_for)
-from app.models import Service
+from app.models import Service, NOTIFICATION_STATUS_TYPES
 from app.dao.templates_dao import dao_get_all_templates_for_service, dao_update_template
 from app.dao.services_dao import dao_update_service
 from app.dao.notifications_dao import get_notification_by_id, dao_get_notification_statistics_for_service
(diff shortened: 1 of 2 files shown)


tests/app/notifications/test_rest.py
```diff
@@ -182,6 +182,49 @@ def test_get_all_notifications_for_job_in_order_of_job_number(notify_api,
...
+
+            response = client.get(
+                path='/service/{}/job/{}/notifications?status={}&status={}&status={}'.format(
+                    sample_service.id, job.id,
+                    NOTIFICATION_STATUS_TYPES[3], NOTIFICATION_STATUS_TYPES[4], NOTIFICATION_STATUS_TYPES[5]),
+                headers=[create_authorization_header()]
+            )
+            resp = json.loads(response.get_data(as_text=True))
+            assert len(resp['notifications']) == 0
+            assert response.status_code == 200
```
leohemsted: feels like two separate tests here

tests/app/notifications/test_rest.py
```diff
@@ -182,6 +182,49 @@ def test_get_all_notifications_for_job_in_order_of_job_number(notify_api,
...
+                to_field="1",
+                created_at=datetime.utcnow(),
+                status=NOTIFICATION_STATUS_TYPES[0],
+                job_row_number=1
+            )
+
+            for length_expected, status in [
+                (1, NOTIFICATION_STATUS_TYPES[0]),
+                (0, NOTIFICATION_STATUS_TYPES[1])
+            ]:
```
leohemsted: `for [int(not i), NOTIFICATION_STATUS_TYPES[i]] in range(2): `

but if you split out the test below, then i'd say make this a `@pytest.mark.parametrize`

quis: `int(not i)` 😦

leohemsted: definitely legit 😉

leohemsted: i spent far too long last night on the train home working out how you could do it with enumerate

(for reference `for length_expected, status in enumerate(reversed(NOTIFICATION_STATUS_TYPES[1::-1])):` *should* work)

quis:
@leohemsted Rewrote that test to use `parametrize` with the parameters written out: https://github.com/alphagov/notifications-api/pull/413/files#diff-3db1fa4d5e7a153d7c98b054251b1f86R186

I think it’s clearer, but 👏 for the enumerate anyway.

leohemsted:
👍

PR #414, opened 2016-06-13
Split sms and retry

minglis:
Currently we don't retry any of our tasks. This pull request brings retry logic into the sending of SMS.

In order to do this I've split the send-sms task up.

The existing task still exists, and now focuses on creating the notification in the database. This task creates the notification and handles restricted services. It hands off the actual call to the 3rd party to another task.

Note we retry this first task on a SQLAlchemy exception. This allows us to handle the race condition on notification creation where we may have an issue creating stats. This should be replaced when we use postgress upserts.

This second task is in a new file - provider_tasks. This handles the API call to the SMS provider and updating the notification with the relevant data, sent at time, provider used, stats etc,

If the API call fails we retry the task using celery countdown functionality. This is an exponential scale, starting at 1 minute and going up to 25 minutes depending on the iteration. We retry a max of 5 times, on the 6th time we through an exception (standard celery retries exceeded exception) and mark the notification as being a technical-failure, and update the stats accordingly.

statsd is called in the provider send task to log total time from the creation of the notification until the completion of the task so we know the length of time taken to do the whole process.


app/celery/tasks.py
```diff
@@ -247,51 +241,21 @@ def send_sms(service_id, notification_id, encrypted_notification, created_at):
...
 
         current_app.logger.info(
             "SMS {} created at {} sent at {}".format(notification_id, created_at, sent_at)
         )
+
         statsd_client.incr("notifications.tasks.send-sms")
         statsd_client.timing("notifications.tasks.send-sms.task-time", monotonic() - task_start)
     except SQLAlchemyError as e:
         current_app.logger.exception(e)
+        raise self.retry(queue="retry", exc=e)
```
servingUpAces: We are retrying 5 times with 5 second intervals. Is that what we want when the db write fails?
Also the job is marked as complete but we will have missing notifications. Is that right?

app/celery/provider_tasks.py
```diff
@@ -0,0 +1,109 @@
...
+        notification.sent_by = provider.get_name(),
+        notification.content_char_count = template.replaced_content_count
+        dao_update_notification(notification)
+    except SmsClientException as e:
+        try:
+            current_app.logger.error(
+                "SMS notification {} failed".format(notification_id)
+            )
+            current_app.logger.exception(e)
+            self.retry(queue="retry", countdown=retry_iteration_to_delay(self.request.retries))
```
servingUpAces: This is putting more than one task in the retry queue each time it retries. I think I am getting the same number of tasks in the queue as the number of seconds it will retry in.

app/celery/tasks.py
```diff
@@ -224,12 +223,6 @@ def send_sms(service_id, notification_id, encrypted_notification, created_at):
 
     try:
 
-        template = Template(
-            dao_get_template_by_id(notification['template'], notification['template_version']).__dict__,
-            values=notification.get('personalisation', {}),
-            prefix=service.name
-        )
-
         sent_at = datetime.utcnow()
```
servingUpAces: sent_at no longer used

PR #422, opened 2016-06-15
move all send_notification tests to separate file

leohemsted:
no changes to tests in any way (though cleaned up/re-ordered imports)


servingUpAces:
👍

PR #424, opened 2016-06-15
Make errors return via error handlers in errors.py

ashimali:
Change endpoint responses where there are marshalling, unmarshalling or param 
errors to raise invalid request or marshmallow validation exceptions. 

That will cause those responses to be handled in by errors.py, which will log the errors.

Set most of schemas to strict mode so that marshmallow will raise exception rather than checking for errors in return tuple from load.

Added handler to errors.py for marshmallow validation errors.

### DIFF
--- a/app/job/rest.py
+++ b/app/job/rest.py
@@ -1,8 +1,8 @@
 from flask import (
     Blueprint,
     jsonify,
-    request,
-    current_app)
+    request
+)
 
 from app.dao.jobs_dao import (
     dao_create_job,
@@ -43,7 +43,7 @@ def get_jobs_by_service(service_id):
         try:
             limit_days = int(request.args['limit_days'])
         except ValueError as e:
-            errors = {'error': ['{} is not an integer'.format(request.args['limit_days'])]}
+            errors = {'limit_days': ['{} is not an integer'.format(request.args['limit_days'])]}
             raise InvalidRequest(errors, status_code=400)
     else:
         limit_days = None
--- a/app/job/rest.py
+++ b/app/job/rest.py
@@ -22,7 +22,10 @@ from app.celery.tasks import process_job
 
 job = Blueprint('job', __name__, url_prefix='/service/<uuid:service_id>/job')
 
-from app.errors import register_errors
+from app.errors import (
+    register_errors,
+    InvalidRequest
+)
 
 register_errors(job)
 
@@ -30,7 +33,7 @@ register_errors(job)
 @job.route('/<job_id>', methods=['GET'])
 def get_job_by_service_and_job_id(service_id, job_id):
     job = dao_get_job_by_service_id_and_job_id(service_id, job_id)
-    data, errors = job_schema.dump(job)
+    data = job_schema.dump(job).data
     return jsonify(data=data)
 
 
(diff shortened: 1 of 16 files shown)


ashimali:
@servingUpAces and @leohemsted have a look please.

app/job/rest.py
```diff
@@ -40,16 +43,13 @@ def get_jobs_by_service(service_id):
         try:
             limit_days = int(request.args['limit_days'])
         except ValueError as e:
-            error = '{} is not an integer'.format(request.args['limit_days'])
-            current_app.logger.error(error)
-            return jsonify(result="error", message={'limit_days': [error]}), 400
+            errors = {'error': ['{} is not an integer'.format(request.args['limit_days'])]}
```
leohemsted: should be `{'limit_days': [error_msg]}`

app/notifications/rest.py
```diff
@@ -317,62 +285,48 @@ def send_notification(notification_type):
...
-                service.message_limit)), 429
+            error = 'Exceeded send limits ({}) for today'.format(service.message_limit)
+            raise InvalidRequest(error, status_code=429)
 
     notification, errors = (
         sms_template_notification_schema if notification_type == 'sms' else email_notification_schema
     ).load(request.get_json())
 
-    if errors:
-        return jsonify(result="error", message=errors), 400
```
leohemsted: EmailNotificationSchema and SmsTemplateNotificationSchema are still 'un-strict' so I think you still need this error check

app/permission/rest.py
```diff
@@ -10,17 +14,12 @@
 
 @permission.route('', methods=['GET'])
 def get_permissions():
-    data, errors = permission_schema.dump(
-        permission_dao.get_query(filter_by_dict=request.args), many=True)
-    if errors:
-        abort(500, errors)
+    data = permission_schema.dump(permission_dao.get_query(filter_by_dict=request.args), many=True).data
```
leohemsted: this'll now 400 instead of 500 - is that ok? i assume we 500'd for a reason, but not sure what that is/was

app/template/rest.py
```diff
@@ -28,27 +31,23 @@ def _content_count_greater_than_limit(content, template_type):
...
     if template_type == 'sms' and \
        template.content_count > current_app.config.get('SMS_CHAR_COUNT_LIMIT'):
-        return True, jsonify(
-            result="error",
-            message={'content': ['Content has a character count greater than the limit of {}'.format(
-                current_app.config.get('SMS_CHAR_COUNT_LIMIT'))]}
-        )
-    return False, ''
+        return True
+    return False
```
leohemsted: can just put `return` on line 32

app/permission/rest.py
ashimali: I assumed 500 was an error in this case as any real 500 in this case I guess would be db related and we have error handlers for those.

app/notifications/rest.py
ashimali: Yep I'll put those checks in

app/job/rest.py
ashimali: Will change

app/template/rest.py
ashimali: Will change

leohemsted:
👍

PR #426, opened 2016-06-15
add template body, version, and subject to send_notification api

leohemsted:
example return value:
```
{'data': {'body': "dear leo\n\nit's currently sunny.\n\nbye!",
          'notification': {'id': '71e377b6-a1fb-495b-ba51-6af4bd7bf7cb'},
          'subject': 'my subject',
          'template_version': 1}}
```

subject only sends for emails


servingUpAces:
👍  tried sending via the clients and it works fine.

PR #428, opened 2016-06-15
Remove the extra task

servingUpAces:
Remove the extra task created to handle the change to the send_email signature


ashimali:
👍

PR #429, opened 2016-06-15
Refactor permissions dao

servingUpAces:
Refactor permissions dao - removed custom query filter
Removed permissions rest - not being used.


ashimali:
👍

PR #432, opened 2016-06-16
Add a ‘preview template’ endpoint

quis:
There’s a need for users of the API to be able to take advantage of Notify’s template rendering.

For example, there’s a service that’s building a case management system. Their users are sending emails on a case-by-case basis. Before they send an email, it’s reassuring to double check that the right data is being inserted, that the right template is being used, etc.

This commit:
- adds a separate endpoint for previewing a template, with personalisation taken from the get parameters of the request
- beefs up the tests around getting a template

Not part of this pull request:
- making this endpoint publicly accessible

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -108,7 +108,7 @@ def preview_template_by_id_and_service_id(service_id, template_id):
 
     data['subject'], data['content'] = template_object.replaced_subject, template_object.replaced
 
-    return jsonify(data=data)
+    return jsonify(data)
 
 
 @template.route('/<uuid:template_id>/version/<int:version>')
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -86,6 +86,31 @@ def get_template_by_id_and_service_id(service_id, template_id):
     return jsonify(data=data)
 
 
+@template.route('/<uuid:template_id>/preview', methods=['GET'])
+def preview_template_by_id_and_service_id(service_id, template_id):
+    fetched_template = dao_get_template_by_id_and_service_id(template_id=template_id, service_id=service_id)
+    data = template_schema.dump(fetched_template).data
+    template_object = Template(data, values=request.args.to_dict())
+
+    if template_object.missing_data:
+        raise InvalidRequest(
+            {'template': [
+                'Missing personalisation: {}'.format(", ".join(template_object.missing_data))
+            ]}, status_code=400
+        )
+
+    if template_object.additional_data:
+        raise InvalidRequest(
+            {'template': [
+                'Personalisation not needed for template: {}'.format(", ".join(template_object.additional_data))
+            ]}, status_code=400
+        )
+
+    data['subject'], data['content'] = template_object.replaced_subject, template_object.replaced
+
+    return jsonify(data=data)
+
+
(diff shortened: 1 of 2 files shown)


tests/app/template/test_rest.py
```diff
@@ -324,6 +325,84 @@ def test_should_get_only_templates_for_that_service(notify_api, sample_user, ser
...
+            headers=[('Content-Type', 'application/json'), create_authorization_header()],
+            data=json.dumps({
+                'name': 'my template',
+                'template_type': 'email',
+                'subject': subject,
+                'content': content,
+                'service': str(service.id),
+                'created_by': str(sample_user.id)
+            })
+        ).get_data(as_text=True))['data']['id']
```
leohemsted: any reason you're using the endpoint here rather than calling the sample_template fixture as a function? Would rather avoid two fake requests

quis:
@leohemsted Rewrote the tests to use the `sample_template` fixture.

app/template/rest.py
```diff
@@ -86,6 +86,31 @@ def get_template_by_id_and_service_id(service_id, template_id):
     return jsonify(data=data)
 
 
+@template.route('/<uuid:template_id>/preview', methods=['GET'])
+def preview_template_by_id_and_service_id(service_id, template_id):
```
servingUpAces: Do we need the version of the template?

quis: I don’t think so, because you can only send the latest template.

tests/app/template/test_rest.py
```diff
@@ -324,6 +326,82 @@ def test_should_get_only_templates_for_that_service(notify_api, sample_user, ser
             assert len(json_resp_4['data']) == 1
 
 
+@pytest.mark.parametrize(
+    "subject, content, path, expected_subject, expected_content, expected_error", [
```
servingUpAces: I haven't seen parametrize - it's good.

app/template/rest.py
```diff
@@ -86,6 +86,31 @@ def get_template_by_id_and_service_id(service_id, template_id):
...
+    if template_object.additional_data:
+        raise InvalidRequest(
+            {'template': [
+                'Personalisation not needed for template: {}'.format(", ".join(template_object.additional_data))
+            ]}, status_code=400
+        )
+
+    data['subject'], data['content'] = template_object.replaced_subject, template_object.replaced
+
+    return jsonify(data=data)
```
servingUpAces: It would be nice if we started returning more descriptive json - I mean use jsonify(template=data)

quis:
@servingUpAces more descriptive JSON added in 0d9519c

## Remove wrapper around response object

Before:
```json
{"data": {"template": …}}
```

There’s no need to wrap the response in key because there will only ever be one valid key for the template preview endpoint.

Flatter is better:
```json
{
  "content": …,
  "subject": …,
  "template_type": …,
  …
}
```

The response will be different if there’s an error, but you should be checking the status code first anyway.

This commit:
- changes the template preview endpoint to return the above format
- adds a test to make sure that the original `/service/…/template/…` endpoint still returns JSON in the same format (with a `data` key)

servingUpAces:
👍

PR #435, opened 2016-06-16
Use notify to send forgot password and invitation emails

servingUpAces:
This PR uses the GOV.UK Notify service to send the email for password reset and invitation emails.
The templates are managed by the template and the endpoints send the send_email task the personalisation for the templates.


app/user/rest.py
```diff
@@ -166,6 +163,7 @@ def send_user_email_verification(user_id):
             'url': _create_verification_url(user_to_send_to, secret_code)
         }
     }
+    print(message)
```
quis: Errant `print()`

PR #441, opened 2016-06-17
Update auth module to return consistently formed error messages.

servingUpAces:
We are trying to get all the error messages to return in the following format:
{result: error,
 message: ['what caused error': 'reason for error']
}


leohemsted:
👍  but i'm pretty sure we don't even use `result` anywhere - since we're unifying the error message struct now does it make sense to kill the `result` kw now?

servingUpAces:
I'm going to merge this as is. Removing result can be decided later.

PR #452, opened 2016-06-20
Move scheduler tasks

servingUpAces:


app/celery/scheduled_tasks.py
```diff
@@ -0,0 +1,94 @@
...
+@notify_celery.task(name="delete-verify-codes")
+def delete_verify_codes():
+    try:
+        start = datetime.utcnow()
+        deleted = delete_codes_older_created_more_than_a_day_ago()
+        current_app.logger.info(
+            "Delete job started {} finished {} deleted {} verify codes".format(start, datetime.utcnow(), deleted)
+        )
+    except SQLAlchemyError:
+        current_app.logger.info("Failed to delete verify codes")
```
leohemsted: do we really need these log statements? They'll be one line above a stack trace that will include "delete_verify_codes" in it, right?

leohemsted:
👍

PR #454, opened 2016-06-20
Store personalisation with notification

ashimali:
This pr is for https://www.pivotaltracker.com/story/show/121672129

Personalisation is now saved with notification.

Template and personalisation content is now merged and returned with
notifications, when retrieved by notification id, or service id (i.e.
all notifications for service).

There is a new element returned at top level of notification json called
body, which is the template content merged with personalisation. This
is consistent with api to endpoint to create notification which returns
what was sent as 'body' in json response.

Merging of template with personalisation is done in the
NotificationStatusSchema.

Personalisation data in encrypted before storing in db.


app/models.py
```diff
@@ -347,6 +350,18 @@ class Notification(db.Model):
     status = db.Column(
         db.Enum(*NOTIFICATION_STATUS_TYPES, name='notify_status_types'), nullable=False, default='sending')
     reference = db.Column(db.String, nullable=True, index=True)
+    _personalisation = db.Column(db.String, nullable=True)
```
leohemsted: i'd rather we give the column the name `personalisation` - if memory serves me you should just be able to write `_personalisation = db.Column('personalisation', db.String, nullable=True)` and that will keep it mapped to `_personalisation` in the ORM but `personalisation` in the actual DB

app/schemas.py
```diff
@@ -246,12 +248,30 @@ class SmsAdminNotificationSchema(SmsNotificationSchema):
...
+
+    @post_dump
+    def handle_template_merge(self, in_data):
+        if in_data.get('personalisation'):
+            from notifications_utils.template import Template
+            merged = Template(in_data['template'], in_data['personalisation']).replaced
+            in_data['body'] = merged
+            in_data.pop('personalisation', None)
+        in_data['template'].pop('content', None)
+        return in_data
```
leohemsted: will this handle placeholders in email subjects?

now that i think about it my addition to the send_notification api the other day certainly wont. whoops.

app/models.py
ashimali: We've left other _ prefixed properties as is for other models, and I don't think it's a egregious situation.

app/schemas.py
ashimali: Probably not, and based on comment with send_notification, let's spin off another story.

app/models.py
leohemsted: ah fair enough 👍

leohemsted:
agree with comments, 👍

PR #457, opened 2016-06-21
Add created as a status for notifications

servingUpAces:
Upon creation of the notification the status of the notification will be created. When the notification is sent to the provider the notification.status will be updated to sending


minglis:
👍 this is minging, maybe we need to not do enums?

leohemsted:
@minglis in my old job we decided to just have boring 'id and name' tables of static data rather than postgres enums for exactly this reason

PR #458, opened 2016-06-21
Fix downgrade to work if any notifications have status = 'created'

servingUpAces:


leohemsted:
looks good when travis passes 👍

PR #459, opened 2016-06-21
Replaced subjects

leohemsted:
In the two new tickets to add the final compiled message body to the `POST /notifications/_ _ _ _`  and  `GET /notifications[/ _ _ _ _]` responses, we neglected to convert email subjects with placeholders in them

Tests and code for that.

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -248,7 +248,7 @@ class SmsAdminNotificationSchema(SmsNotificationSchema):
 
 class NotificationStatusSchema(BaseSchema):
 
-    template = fields.Nested(TemplateSchema, only=["id", "name", "template_type", "content"], dump_only=True)
+    template = fields.Nested(TemplateSchema, only=["id", "name", "template_type", "content", "subject"], dump_only=True)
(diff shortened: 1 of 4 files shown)


app/schemas.py
```diff
@@ -267,10 +267,13 @@ def handle_personalisation_property(self, in_data):
     def handle_template_merge(self, in_data):
         if in_data.get('personalisation'):
             from notifications_utils.template import Template
-            merged = Template(in_data['template'], in_data['personalisation']).replaced
-            in_data['body'] = merged
+            merged = Template(in_data['template'], in_data['personalisation'])
+            in_data['body'] = merged.replaced
```
quis: Why not
```
in_data['body'] = Template(in_data['template'], in_data['personalisation']).replaced
```
?

quis: One less ambiguously named variable ✊

leohemsted: i use it on the next line - so need to save it rather than the previous one-liner. happy to rename it to `template` tho

quis:
👍

PR #464, opened 2016-06-22
Refactor the api_key_dao.

servingUpAces:
The only update we should be doing to an api key is to expire/revoke the api key.
Removed the update_dict from the the save method.
Added an expire_api_key method that only updates the api key with an expiry date.


ashimali:
👍

PR #467, opened 2016-06-22
Notification created status for send_sms

servingUpAces:
- The send_sms task will created the notification with a status = created.  
- The encrypted_notification for the send_sms_to_provider task has been made optional.
- The encrypted_notificaiton is still passed through to the send_sms_to_provider task so that the code is backward compatible during deployments.


ashimali:
👍

PR #469, opened 2016-06-23
Body of notification without placeholder should have been present.

ashimali:


app/schemas.py
```diff
@@ -272,6 +272,8 @@ def handle_template_merge(self, in_data):
             if in_data['template']['template_type'] == 'email':
                 in_data['subject'] = template.replaced_subject
             in_data.pop('personalisation', None)
+        else:
+            in_data['body'] = in_data['template']['content']
```
quis: Sorry for not spotting this when I reviewed the PR yesterday.

You could get rid of the `if in_data.get('personalisation'):` entirely and it would work the same.

ashimali: No problem. I didn't notice it either. However we still need to hand personalisation being absent in in_data dictionary. 

The pre_dump receives a Notification model and if has a personalisation property that is not null, it sets self.notification. 

That self becomes the in_data in post_dump, so we need to check for personalisation key. This bug was for the case where there is no personalisation. In which case just get template content.

ashimali: Oh and I should have mentioned why any of this is necessary at all. Marshmallow doesn't seem to play nice with python properties, hence the need for pre and post jiggery pokery.

leohemsted: should `template['subject']` also be in the else clause

ashimali: Done

leohemsted:
👍

quis:
👍

PR #473, opened 2016-06-24
key_types table

leohemsted:
* single-column static data table that currently contains two types: 'normal' and 'team'
  - essentially similar to a postgres enum, but trying this as it makes it significantly easier to upgrade/downgrade/amend values than altering an enum
* key_type foreign-keyed from api_keys
  - must be not null
    + for now marshmallow will set it to `KEY_TYPE_NORMAL` if not supplied to ensure backwards compatibility
  - existing rows set to 'normal'
* key_type foreign-keyed from notifications
  - nullable
  - existing rows set to null
* api_key foreign-keyed from notifications
  - nullable
  - existing rows set to null

also i cleaned up some test files


servingUpAces:
👍

PR #474, opened 2016-06-27
Notification created status

servingUpAces:
- Stop sending the encrypted message to the send_sms_to_provider task. Everything the task needs is now stored in the db.
- Only send to the provider if the notification has a created status. 
If the notification ends up in the retry queue and the delivery app is restarted the notification will get sent twice.
This is because when the app is restarted another message will be in the retry queue as message available which is a
duplicate of the one in the queue that is a  message in flight.

This should complete https://www.pivotaltracker.com/story/show/121844427


leohemsted:
👍

PR #477, opened 2016-06-27
remove default key_type

leohemsted:
depends on https://github.com/alphagov/notifications-admin/pull/730

remove the default key_type - given that we only have one route for creating API keys, which will imminently let the user choose key_type, it makes no sense to provide this default, it's just extra unwelcome logic that we have to remember/support


servingUpAces:
👍

PR #478, opened 2016-06-28
Users.password_changed_at not nullable

servingUpAces:
Password changed at is defaulted to the current date when the user is created.

This PR set the users.password_changed_at column to not be nullable.


migrations/versions/0034_pwd_changed_at_not_null.py
```diff
@@ -0,0 +1,28 @@
...
+    ### commands auto generated by Alembic - please adjust! ###
+    op.get_bind()
+    op.execute('update users set password_changed_at = created_at where password_changed_at is null')
+    op.alter_column('users', 'password_changed_at', nullable=False)
+    ### end Alembic commands ###
+
+
+def downgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    pass
```
leohemsted: still needs alter_column in downgrade (though doesn't need to update the data)

leohemsted:
👍

PR #481, opened 2016-06-29
Add notification type

servingUpAces:
- Add notification_type to notification table. It seems like an oversight not to include the notification type in the notification.
- When updating statistics a query to the template table is required to get the type, this update will mean that query does not have to happen.


leohemsted:
👍

PR #485, opened 2016-06-29
Api user cleanup

leohemsted:
## api_user now contains the relevant ApiKey object

used in notifications/rest.py to get the current API's service id, it previously contained a dict of `{'client': service_id, 'secrets': [ all api_key secrets of that service ]}`. now contains full db object.

the api_key needs to be stored after the requires_auth decorator, for distinguishing between api_key types.

To aid me I attempted a cleanup of auth.py

### require_admin decorator removed.

it's not required - nginx performs all of its functionality
the functions that used it moved from notifications to job/rest.py and service/rest.py with no changes

### api_secret code moved to authentication/utils.py

so that it can be accessed cleanly from a new `unsigned_secret` property on the `ApiKey` model.

### tweaked authentication error messages

ensure consistency and accuracy of error messages - sometimes previously would report wrong error message if there were multiple invalid keys in the database

* **new**: if the token matches but has been revoked, now return `"Invalid token: revoked"` (previously expired)
* if the token matches but has expired (more than 30 sec since it was generated) still return "expired"
* if no token matches, errors with "signature"

### use error handler rather than returning 403

raise AuthError and catch in `errors.py` - I did this because with the code changes that the above introduced, the function was getting ridiculous - whether this has improved code legibility is up for debtate, and am happy to roll back change if people think it hasn't helped.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -47,7 +47,6 @@ def create_app(app_name=None):
     init_app(application)
     db.init_app(application)
     ma.init_app(application)
-    init_app(application)
     logging.init_app(application)
     statsd_client.init_app(application)
     firetext_client.init_app(application, statsd_client=statsd_client)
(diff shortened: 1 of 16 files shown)


servingUpAces:
👍

PR #488, opened 2016-06-30
Add breathing room around GOV.UK logo in emails

quis:
Implements and depends on:

- [x] https://github.com/alphagov/notifications-utils/pull/44


servingUpAces:
The utils PR has a 👍

PR #489, opened 2016-06-30
Use type from model

servingUpAces:
A small refactor to use the SMS_TYPE and EMAIL_TYPE in code rather than 'sms' or 'email'


leohemsted:
👍

PR #490, opened 2016-07-01
only send to team emails/phones if POST /notification/<type> with team api_key

leohemsted:
uses same restriction as a service in trial mode


leohemsted:
superceded by #496

PR #495, opened 2016-07-01
ensure consistency with reply-to addresses

leohemsted:
remove reply-to-address from the send_email task kwargs, since we can just get it from the service - also depluralise the variable name in aws_ses.send_email and reduce the amount of types we accept.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -109,7 +109,6 @@ def process_job(job_id):
                 create_uuid(),
                 encrypted,
                 datetime.utcnow().strftime(DATETIME_FORMAT)),
-                {'reply_to_addresses': service.reply_to_email_address},
                 queue='bulk-email')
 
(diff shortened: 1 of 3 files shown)


app/celery/tasks.py
```diff
@@ -176,7 +175,7 @@ def send_sms(self, service_id, notification_id, encrypted_notification, created_
 
 
 @notify_celery.task(name="send-email")
-def send_email(service_id, notification_id, encrypted_notification, created_at, reply_to_addresses=None):
```
servingUpAces: Removing the reply_to_addresses may be not compatible with the previous versions of tasks created. What happens if an old task is in the queue with reply_to_addresses and is picked up by a new task. Will that work?

leohemsted: no you're right - that won't work - i'll split this into two PRs

PR #496, opened 2016-07-01
filter GET /notifications by api type

leohemsted:
~~don't merge until #490 - it requires that for some changes to conftest and tests/__init__.py~~


servingUpAces:
👍

PR #497, opened 2016-07-01
Split send email task

servingUpAces:
Split send_email task into one task to create the notification and one to send it to the provider.
If there is an exception the task will go to the retry queue.

https://www.pivotaltracker.com/story/show/123845161


ashimali:
👍

PR #498, opened 2016-07-01
Set sms sender on service

ashimali:
Add sms sender to service to be used in sms templates in place of default numeric short code.
If not present default short code is used.


servingUpAces:
👍  reviewed and demoed

PR #504, opened 2016-07-04
dont pass reply-to-addresses about

leohemsted:
dont send reply_to_addresses around from process_job and send_email - take it from the service in send_email_to_provider. also clean up the kwarg in aws_ses.send_email to more accurately reflect what we might pass in

this is the first of two PRs - next PR will remove kwargs from celery tasks

### DIFF
--- a/tests/app/celery/test_tasks.py
+++ b/tests/app/celery/test_tasks.py
@@ -231,7 +231,6 @@ def test_should_process_email_job_if_exactly_on_send_limits(notify_db,
             "something_encrypted",
             "2016-01-01T11:09:00.061258"
         ),
-        {'reply_to_addresses': None},
         queue="bulk-email"
     )
     mock_celery_remove_job.assert_called_once_with((str(job.id),), queue="remove-job")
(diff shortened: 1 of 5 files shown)


ashimali:
👍

tests/app/celery/test_tasks.py
```diff
@@ -654,7 +649,7 @@ def test_send_email_should_go_to_retry_queue_if_database_errors(sample_email_tem
     assert 'No row was found for one' in str(e.value)
 
 
-def test_process_email_job_should_use_reply_to_email_if_present(sample_email_job, mocker, mock_celery_remove_job):
+def test_process_email_should_not_use_reply_to_email(sample_email_job, mocker, mock_celery_remove_job):
```
ashimali: Probably would have just removed this test actually.

leohemsted: yeah i removed it in next PR - had it in this one to sanity check that my two-step change still accepts the old parameters

PR #507, opened 2016-07-04
remove reply_to_addresses from task kwargs

leohemsted:
also added a test for aws_ses.send_email to prove it handles reply_to_address correctly


quis:
👍

PR #510, opened 2016-07-05
Test api keys

leohemsted:
adds a new api key_type, `test` - this key type behaves the same as research mode - does not send to ses/mmg/firetext, but to our own mock classes

I also moved all the callback tests into their own separate files so notifications/test_rest is no longer ridiculously long


servingUpAces:
👍

quis:
Can you double check with @yahoopete or @minglis about how these keys work when `get`ting notifications before merging?

leohemsted:
@minglis agreed with silo'd API key `get`s normal get returns job/normal, team returns team, and test returns test only. will double check for what we want to see on dashboard in a bit but for now merging

PR #512, opened 2016-07-06
Fix bug where all notifications were getting a type = email.

servingUpAces:
Includes a fix to notification and notification_statistics data.


leohemsted:
👍  demoed

PR #515, opened 2016-07-06
add pip-accel to travis to speed up builds

leohemsted:
* our travis builds spend several minutes building python dependencies.
* pip rebuilds dependencies, even if the version hasn't changed.
* many of our dependencies are c-compiled, and take a long time to build.
* [pip-accel](https://github.com/paylogic/pip-accel) is a library that locally caches built python libraries
* we can use travis' built-in caching system to cache that pip-accel directory, and hopefully as a result can reduce build times (after the initial build)

from ~100s down to ~25 seconds on this branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,16 @@
 sudo: false
 language: python
-cache: pip
+cache:
+  pip: true
+  directories:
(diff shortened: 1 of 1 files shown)


quis:
👍

PR #520, opened 2016-07-07
Use `PassThrough` renderer

quis:
Implements and depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/49


quis:
@servingUpAces 

### Add a test for sending email to provider

We had a test like this for sending sms, but not email. This meant that, for example, we weren’t checking that the provider was getting passed the HTML and plain text versions of the email.

servingUpAces:
👍

PR #523, opened 2016-07-08
Add a lockmode for the update_notifications_status_by_id to prevent t…

servingUpAces:
…he timeout task from updating the same notification more than once.

This happens because more than one beat process was creating the timeout task, resulting in multiple workers running the same queries at the same time.


leohemsted:
👍

PR #524, opened 2016-07-08
Email registered users

servingUpAces:
New endpoint to send a message to a user that is registering with an email address that already exists.


migrations/versions/0041_email_template_.py
```diff
@@ -0,0 +1,56 @@
...
+                                              datetime.utcnow(), content, service_id,
+                                              'Your GOV.UK Notify account', user_id))
+    op.execute(
+        template_insert.format('0880fbb1-a0c6-46f0-9a8e-36c986381ceb', 'Your GOV.UK Notify account', 'email',
+                               datetime.utcnow(), content, service_id,
+                               'Your GOV.UK Notify account', user_id))
+
+
+def downgrade():
+    op.execute("delete from notifications where template_id = '0880fbb1-a0c6-46f0-9a8e-36c986381ceb'")
```
servingUpAces: @leohemsted if I don't delete the data and we downgrade, then we will get a failure with the upgrade. So I left it in.

leohemsted: ah yes of course

tests/app/user/test_rest.py
```diff
@@ -424,6 +424,21 @@ def test_send_user_reset_password_should_send_reset_password_link(notify_api,
                 queue="email-reset-password")
 
 
+def test_send_user_reset_password_should_return_400_when_email_is_missing(notify_api):
+    with notify_api.test_request_context():
+        with notify_api.test_client() as client:
+            data = json.dumps({})
+            auth_header = create_authorization_header()
+
+        resp = client.post(
```
leohemsted: this should be indented

leohemsted:
👍

PR #525, opened 2016-07-11
updated bootstrap

minglis:


scripts/bootstrap.sh
```diff
@@ -22,7 +22,7 @@ function display_result {
 
 if [ ! $VIRTUAL_ENV ]; then
   virtualenv ./venv
-  . ./venv/bin/activate
+  virtualenv -p python3 ./venv
```
leohemsted: think you removed the wrong line?

scripts/bootstrap.sh
```diff
@@ -20,9 +20,13 @@ function display_result {
...
 }
 
+ if [ ! $VIRTUAL_ENV ]; then
+  virtualenv -p python3 ./venv
+  . ./venv/bin/activate
+
 if [ ! $VIRTUAL_ENV ]; then
   virtualenv ./venv
-  . ./venv/bin/activate
+  virtualenv -p python3 ./venv
```
leohemsted: seemingly accidentally duplicated the if block rather than editing

quis:
👍

PR #526, opened 2016-07-11
Notification history table

leohemsted:
new table `notification_history` that mimics notifications table, but with the aim that we will never delete from the table. We can use it to construct statistics in batch jobs in a straightforward manner rather than having to ensure that the current stats tables are always kept up-to-date.

It's exactly the same as the notification table except for:
* no defaults (eg id/created_by) - this table should only ever completely duplicate rows from the `notification` table.
* no personalisation - don't keep sensitive customer information (the `to` phone number/email and any template personalisation) any longer than before.

The table is only updated through manual python calls in `dao_create_notification` and `dao_update_notification` - if your code updates notifications in some other way (eg just setting a variable inside a transactional block) then you will need to either alter your code to update the notification history table yourself or use the dao functions above, as I did in the `_update_notification_stats_query` function


migrations/versions/0041_notification_history.py
```diff
@@ -0,0 +1,61 @@
+"""empty message
```
servingUpAces: There is a script 0041, you'll need to merge master

servingUpAces:
👍 tested and reviewed

PR #533, opened 2016-07-14
Daily message limit imposed for service api calls.

ashimali:
Daily message limits are imposed for csv jobs, but not api calls. This pr brings these into line by respecting limits in api calls. It may be decided later that no limit is the way to go, and if so, this can be reverted and a check for service being restricted can be added to tasks.py -> process_job.

Another pr will be coming on admin side to increase the daily limit that is set when service is un restricted.

Fix for https://www.pivotaltracker.com/story/show/126230157


quis:
👍

PR #538, opened 2016-07-15
Speed up GET services endpoints

leohemsted:
Speed up GET services:

### before
image

### after
image

Joined loads are almost always better than lazily loading - we know we need users, so load them when we grab the service, rather than when we dump the schema.

No loads are better than both, so exclude stats tables.


ashimali:
👍

PR #539, opened 2016-07-15
Upgrade bleach to fix dodgy html5lib version

quis:
Bleach 1.4.2 installs html5lib version 0.999 or greater.

This was fine, until version 0.999999999 (nine nines) was released, which doesn’t work.

Bleach 1.4.3 pins html5lib to 0.9999999 (seven nines), which stops the bad version getting installed.


leohemsted:
👍

PR #544, opened 2016-07-19
Create notification history from jobs

ashimali:
Db migration to retrospectively create notification history from jobs table.

The migration selects only jobs from after the go live date (18/05/2016) and before the earliest record in notification history table (2016-06-26 23:21:55).

This migration can only set outcome of notifications to either delivered or failed as jobs table has no data about temporary failure or other status.

Created at and sent at will be set to job creation time and job process finish time, and therefore will be different than actual timings.

Messages sent my amazon SES will have no reference ids.


ashimali:
@servingUpAces and @minglis let me know if you're ok with this and the caveats in comment above.

minglis:
Looks OK, best efforts after all. Going to get actual numbers from Prod and see what it behaves like.

ashimali:
The select would return 1630 jobs
Of those number delivered is 84k and number failed is 73k records.

minglis:
150k or so. We should test that.

PR #547, opened 2016-07-21
Only put a character limit on SMS notifications

quis:
We limit SMS to be a maximum of 495 characters at the point of an API call being made.

We were also applying this limit to emails, which is wrong.

### DIFF
--- a/app/notifications/rest.py
+++ b/app/notifications/rest.py
@@ -226,7 +226,7 @@ def send_notification(notification_type):
             raise InvalidRequest(error, status_code=429)
 
     notification, errors = (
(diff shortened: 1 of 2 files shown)


app/notifications/rest.py
```diff
@@ -256,7 +256,10 @@ def send_notification(notification_type):
         errors = {'template': [message]}
         raise InvalidRequest(errors, status_code=400)
 
-    if template_object.replaced_content_count > current_app.config.get('SMS_CHAR_COUNT_LIMIT'):
+    if (
+        template_object.template_type == 'sms' and
```
servingUpAces: You can use the constant SMS_TYPE here

leohemsted:
👍

PR #552, opened 2016-07-22
Update utils for Markdown parsing + subbranding

quis:
Markdown parsing (will be active but isn’t documented):
- [x] https://github.com/alphagov/notifications-utils/pull/54

Email sub branding (not used yet):
- [x] https://github.com/alphagov/notifications-utils/pull/55


leohemsted:
👍

PR #553, opened 2016-07-22
Statds value error

ashimali:
Statsd timings were being passed a timedelta instead for float for milliseconds.

Bugfix for https://www.pivotaltracker.com/story/show/126852733


leohemsted:
👍

PR #556, opened 2016-07-22
Fix bug with expired token error response

servingUpAces:


tests/app/authentication/test_authentication.py
```diff
@@ -224,17 +256,3 @@ def __create_post_token(service_id, request_body):
         secret=get_unsigned_secrets(service_id)[0],
         client_id=str(service_id)
     )
-
-
-def test_should_attach_the_current_api_key_to_current_app(notify_api, sample_service, sample_api_key):
```
leohemsted: why did you remove this test?

ashimali: Moved not removed?

minglis:
👍

PR #557, opened 2016-07-22
make tests run consistently

leohemsted:
Just a couple of tests that were returning inconsistent data, and sometimes screwing up my test runs.

* tests on API endpoints that we do not explicitly sort should either sort the results or use a data set where we do not rely on json list order being consistent
* tests that touch the provider_details should reset values after running, since the provider_details table is not torn down and re-created between tests (unlike most tables)


minglis:
👍

PR #559, opened 2016-07-25
Get service with stats for today

leohemsted:
This endpoint is needed by the send page, which checks to see how many messages you've sent today compared to your daily limit

blocks https://github.com/alphagov/notifications-admin/pull/809

~~blocked by #557~~

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -1,4 +1,5 @@
 import uuid
+from datetime import date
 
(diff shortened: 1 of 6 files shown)


minglis:
👍 think this is OK. Assuming indexes exist on date field propely?

leohemsted:
@minglis index does not currently exist, i'll throw it on there

PR #560, opened 2016-07-26
fix notification 500

leohemsted:
problem is due to a bug (still outstanding see pivotal #127079637) in our marshmallow notification schema when it tries to recreate the message content from personalisation if the template has been edited.

Fixed by using different schema that skips that in all places except for the public /notification endpoint.

Added an expected-fail test to test that


minglis:
👍

PR #563, opened 2016-07-28
New weekly aggregate function

leohemsted:
New function to replace `GET /service/<id>/notifications-statistics/seven_day_aggregate` - summarises status counts on a per-week basis from the `notifications_history` table - I'm not 100% sure how the old one actually worked but this one works fine - i moved as much date manipulation and aggregation back into python to try and make it more legible but let me know if it's off

apologies for lots of +- changes but they're mostly new test cases and moving a couple of fns to a different file

~~blocked by #559~~

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -157,3 +157,21 @@ def _stats_for_service_query(service_id):
         Notification.notification_type,
         Notification.status,
     )
+
+
+def dao_fetch_weekly_historical_stats_for_service(service_id):
+    monday_of_notification_week = func.date_trunc('week', NotificationHistory.created_at).label('week_start')
+    return db.session.query(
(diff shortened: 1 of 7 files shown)


app/dao/services_dao.py
```diff
@@ -145,4 +156,22 @@ def dao_fetch_stats_for_service(service_id):
...
+    return db.session.query(
+        NotificationHistory.notification_type,
+        NotificationHistory.status,
+        monday_of_notification_week,
+        func.count(NotificationHistory.id).label('count')
+    ).filter(
+        NotificationHistory.service_id == service_id
+    ).group_by(
+        NotificationHistory.notification_type,
+        NotificationHistory.status,
```
minglis: Again are these all indexed?

minglis:
👍

PR #579, opened 2016-08-03
get usage stats from notification_history

leohemsted:
an in-place replacement of old code for getting usage details

### replaced content_char_count with billable_units

content_char_count is grabbed at send_sms time, and stored in the db then. we weren't using it at all later on - so it appeared to be useful for calculating usage on the fly when users hit the dashboard. However, we don't record whether the notification was *actually* sent - if it was in research mode then we lose that information.

So I've moved the calculation of billable units from read time to write time by removing content_char_count and replacing with billable units

### the upgrade script

it's non-trivial:

* it has to convert content_char_count to billable units, following the formula detailed here: https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/template.py#L214.  

it's not perfect:

* the downgrade script is destructive - there's loss of information by replacing char count with units, so i just guess that each billable unit was worth 150 characters
* we don't know whether the original message was in research mode - so we decided to assume all SMS were sent live, except for *any* sent by a service that has *ever* been in research mode (on live, that's only services we've created for testing purposes it appears)

enjoy!


leohemsted:
@minglis thumbsed up

PR #583, opened 2016-08-05
Stats db updates

minglis:
Refactor statsd logging

Removed all existing statsd logging and replaced with:

- statsd decorator. Infers the stat name from the decorated function call. Delegates statsd call to statsd client. Calls incr and timing for each decorated method. This is applied to all tasks and all dao methods that touch the notifications/notification_history tables

- statsd client changed to prefix all stats with "notification.api."

- Relies on https://github.com/alphagov/notifications-utils/pull/61 for request logging. Once integrated we pass the statsd client to the logger, allowing us to statsd all API calls. This passes in the start time and the method to be called (NOT the url) onto the global flask object. We then construct statsd counters and timers in the following way

	notifications.api.POST.notifications.send_notification.200 

This should allow us to aggregate to the level of 
	
	- API or ADMIN
	- POST or GET etc
	- modules
	- methods 
	- status codes 

Finally we count the callbacks received from 3rd parties to mapped status.


app/celery/provider_tasks.py
```diff
@@ -101,8 +101,6 @@ def send_sms_to_provider(self, service_id, notification_id):
         current_app.logger.info(
             "SMS {} sent to provider at {}".format(notification_id, notification.sent_at)
         )
-        statsd_client.incr("notifications.tasks.send-sms-to-provider")
-        statsd_client.timing("notifications.tasks.send-sms-to-provider.task-time", monotonic() - task_start)
```
leohemsted: does statsd have the concept of migrating/renaming/simlinking the old spelling to the new one?

app/__init__.py
```diff
@@ -1,7 +1,7 @@
 import uuid
 import os
 
-from flask import request, url_for, g
+from flask import request, url_for, g, current_app
```
leohemsted: unused import `current_app`

app/celery/provider_tasks.py
minglis: I'm happy to drop the old stats and move on. They were mostly unused.

leohemsted:
👍

PR #585, opened 2016-08-05
Organisations

leohemsted:
## organisations

`organisation` is a new table that holds an organisation's branding details - its colour, a path to its logo and its display name.

services now have a nullable organisation_id. 

## branding

branding refers to how the service interprets the org's branding. it can be one of three strings (backed by a foreign key to a static table)

* `govuk` - (default) GOV.UK branding, as current
* `org` - Only use the organisation's branding
* `both` - use both the gov.uk branding and the org's branding.

## misc cleanup

the history meta wasn't ready for nullable foreign keys. now it is!

### DIFF
--- a/app/history_meta.py
+++ b/app/history_meta.py
@@ -53,7 +53,12 @@ def _history_mapper(local_mapper):
         col = col.copy()
         orig.info['history_copy'] = col
         col.unique = False
-        col.default = col.server_default = None
+
+        # if the column is nullable, we could end up overwriting an on-purpose null value with a default.
+        # if it's not nullable, however, the default may be relied upon to correctly set values within the database,
+        # so we should preserve it
+        if col.nullable:
+            col.default = col.server_default = None
         return col
 
     properties = util.OrderedDict()
@@ -201,7 +206,9 @@ def create_history(obj):
 
         elif isinstance(prop, RelationshipProperty):
             if hasattr(history, prop.key+'_id'):
-                data[prop.key+'_id'] = getattr(obj, prop.key).id
+                foreign_obj = getattr(obj, prop.key)
+                # if it's a nullable relationship, foreign_obj will be None, and we actually want to record that
+                data[prop.key+'_id'] = getattr(foreign_obj, 'id', None)
 
     if not obj.version:
         obj.version = 1
--- a/app/models.py
+++ b/app/models.py
@@ -5,7 +5,7 @@ from sqlalchemy.dialects.postgresql import (
     UUID,
     JSON
 )
-from sqlalchemy import UniqueConstraint
+from sqlalchemy import UniqueConstraint, text
 
 from app.encryption import (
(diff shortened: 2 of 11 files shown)


app/history_meta.py
```diff
@@ -53,7 +53,12 @@ def _col_copy(col):
         col = col.copy()
         orig.info['history_copy'] = col
         col.unique = False
-        col.default = col.server_default = None
+
+        # if the column is nullable, we could end up overwriting an on-purpose null value with a default.
```
minglis: These sort of hacks are a bad smell. the sooner we get this versioning plugin out the better.

app/models.py
```diff
@@ -98,12 +116,21 @@ class Service(db.Model, Versioned):
         secondary=user_to_service,
         backref=db.backref('user_to_service', lazy='dynamic'))
     restricted = db.Column(db.Boolean, index=False, unique=False, nullable=False)
-    research_mode = db.Column(db.Boolean, index=False, unique=False, nullable=True, default=False)
+    research_mode = db.Column(db.Boolean, index=False, unique=False, nullable=False, default=False)
```
minglis: Do we know this will work?

migrations/versions/0046_organisations_and_branding.py
```diff
@@ -0,0 +1,64 @@
...
+    op.add_column('services', sa.Column('organisation_id', postgresql.UUID(as_uuid=True)))
+    op.add_column('services_history', sa.Column('branding', sa.String(length=255)))
+    op.add_column('services_history', sa.Column('organisation_id', postgresql.UUID(as_uuid=True)))
+
+    op.execute("INSERT INTO branding_type VALUES ('govuk'), ('org'), ('both')")
+    # insert UKVI data as initial test data. hex and crest pulled from alphagov/whitehall
+    op.execute("""INSERT INTO organisation VALUES (
+        '9d25d02d-2915-4e98-874b-974e123e8536',
+        '#9325b2',
+        'ho_crest_27px_x2.png',
```
minglis: Where is this file? Is this a dependency between API and frontend on content?

minglis:
Couple of minor things. 

Also do we need organisation? Noone is using and branding applies directly to service?

migrations/versions/0046_organisations_and_branding.py
leohemsted: yup - it comes from here: https://www.notifications.service.gov.uk/static/images/email-template/crests/ho_crest_27px_x2.png - file in source code is https://github.com/alphagov/notifications-admin/blob/master/app/assets/images/email-template/crests/ho_crest_27px_x2.png

minglis: pivotal an alert on the branding so that if fails/missing/etc we know.

leohemsted:
comments addressed, merging

PR #589, opened 2016-08-08
Render organisational branding in emails

leohemsted:
image

* [x] blocked by #585

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -1,27 +1,27 @@
 from datetime import datetime
 from monotonic import monotonic
+from urllib.parse import urljoin
+
 from flask import current_app
+from notifications_utils.recipients import (
+    validate_and_format_phone_number
+)
+from notifications_utils.template import Template, get_sms_fragment_count
+from notifications_utils.renderers import HTMLEmail, PlainTextEmail, SMSMessage
+
 from app import notify_celery, statsd_client, clients, create_uuid
 from app.clients.email import EmailClientException
 from app.clients.sms import SmsClientException
-
 from app.dao.notifications_dao import (
     update_provider_stats,
     get_notification_by_id,
     dao_update_notification,
     update_notification_status_by_id
 )
-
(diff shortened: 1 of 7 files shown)


app/celery/provider_tasks.py
```diff
@@ -189,3 +189,22 @@ def send_email_to_provider(self, service_id, notification_id):
         statsd_client.timing("notifications.tasks.send-email-to-provider.task-time", monotonic() - task_start)
         delta_milliseconds = (datetime.utcnow() - notification.created_at).total_seconds() * 1000
         statsd_client.timing("notifications.email.total-time", delta_milliseconds)
+
+
+def get_html_email_renderer(service):
+    govuk_banner = service.branding != BRANDING_ORG
+    if service.organisation:
+        logo = '{}{}{}'.format(
+            current_app.config['ADMIN_BASE_URL'],
```
leohemsted: not sure if i want to do this - i prefer it because it means we don't have emails from different envs relying on each other, but it does mean that styled emails won't work from localhost

app/celery/provider_tasks.py
```diff
@@ -185,3 +185,22 @@ def send_email_to_provider(self, service_id, notification_id):
         )
         delta_milliseconds = (datetime.utcnow() - notification.created_at).total_seconds() * 1000
         statsd_client.timing("email.total-time", delta_milliseconds)
+
+
+def get_html_email_renderer(service):
+    govuk_banner = service.branding != BRANDING_ORG
```
leohemsted: leo check and make sure both vs org works as expected

app/celery/provider_tasks.py
servingUpAces: It would be nice if localhost worked the same as other environments, but this is a case where it probably does not matter.

app/celery/provider_tasks.py
leohemsted: update - it appears to

servingUpAces:
👍 Reviewed, demoed and ran functional tests.

PR #595, opened 2016-08-09
re-add Content char count

leohemsted:
only for the public endpoints


leohemsted:
self merging as no-one else is around to 👍

PR #596, opened 2016-08-10
Improve logging and extend time periods associated with retrying

minglis:
**Extend the retry time on the create notification DB tasks**
- Was previoulsy 5 attempts 5 seconds apart
- No 5 times 5 minutes apart.
- Gives 25 mins of retries before an error

**In event of a task retry ensure the log message is identifier as such**
- "RETRY" prefixes the messages

In event of the retry attempts completing without successfully completing the task identify message as such

- "RETRY FAILED" prefixes the messages

Applies to the send_sms|send_email and send_sms_to_provider|send_email_to_provider tasks

These are there to try and ensure we can alert on these events so that we know if we have started retrying messages

Retry messages also contain notification ids to aid debugging.

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -92,11 +92,15 @@ def send_sms_to_provider(self, service_id, notification_id):
         except SmsClientException as e:
             try:
                 current_app.logger.error(
-                    "SMS notification {} failed".format(notification_id)
+                    "RETRY: SMS notification {} failed".format(notification_id)
                 )
                 current_app.logger.exception(e)
                 self.retry(queue="retry", countdown=retry_iteration_to_delay(self.request.retries))
             except self.MaxRetriesExceededError:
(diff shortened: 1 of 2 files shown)


app/celery/provider_tasks.py
```diff
@@ -92,11 +92,15 @@ def send_sms_to_provider(self, service_id, notification_id):
         except SmsClientException as e:
             try:
                 current_app.logger.error(
-                    "SMS notification {} failed".format(notification_id)
+                    "RETRY: SMS notification {} failed".format(notification_id)
```
leohemsted: this'll log with 'RETRY' even if it's the first time running the task - is that intended behaviour? if so why bother adding 'RETRY'?

minglis: I think if it throws an exception it is a retry (it's going to) which is worth knowing about.

PR #599, opened 2016-08-10
fix GET /notifications 500 error

leohemsted:
kinda two separate parts to this:

### replacing `Versioned` with `TemplateHistory`

version table is a total pain and doesn't actually give us that much for free - but does block us from constructing sqlalchemy relationships or joining to it in an ORM manner, so remove it. Turns out this is very simple - had to do some small work to the version decorator (used on ORM create/update methods) to support this but that's all)

### notification knows about `TemplateHistory` instead of `Template`

It's not complete in this ticket, but this is the smallest section of work that fixes the 500 errors.

We shouldn't be referring to templates in the notification table - notifications are tied to a template's version at the moment in time that they're created, and if the template changes we shouldn't change that reference.

So for now solve this in a stop-gap method - add a new relationship called `template_history` and specificially reference that in `GET /notifications`, in the dao and schema layers.

A future chore is to move all of the other times we link notification and template to refer to this new relationship, and then switch over the foreign key in the database to be a composite foreign key to `templates_history` instead of `template`

### DIFF
--- a/app/dao/dao_utils.py
+++ b/app/dao/dao_utils.py
@@ -1,6 +1,7 @@
 import itertools
-from functools import wraps
-from app.history_meta import versioned_objects, create_history
+from functools import wraps, partial
+
+from app.history_meta import create_history
 
 
 def transactional(func):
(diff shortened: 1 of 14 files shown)


minglis:
👍

PR #600, opened 2016-08-10
Give better error messages when using the API to send to recipients not in your team

quis:
## Give better error when sending to non-team members

Scenario we saw in research:
- trying to send a message to someone outside your team
- service is in trial mode

Result:
- error message was terrible, no-one understood it

Solution:
- better error message

## Show separate error messages for team key

Although using a team key is functionally the same as your service being restricted, conflating the two errors is not helpful. What we typically saw in research was that someone was using a team key, got the error, used a live key and got the _same_ error.

This commit adds a new error message that specifically mentions the type of API key that you’re using.

### DIFF
--- a/app/notifications/rest.py
+++ b/app/notifications/rest.py
@@ -270,11 +270,15 @@ def send_notification(notification_type):
             [user.mobile_number, user.email_address] for user in service.users
         )
     ):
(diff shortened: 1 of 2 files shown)


leohemsted:
👍

PR #605, opened 2016-08-16
Create Docker build image, build project with Docker

bandesz:


scripts/run_tests.sh
```diff
@@ -24,8 +24,12 @@ function display_result {
   fi
 }
 
-pep8 .
+python_bin_dir=""
+if [ -d venv ]; then
+  python_bin_dir=venv/bin/
+fi
+${python_bin_dir}pep8 .
```
leohemsted: so this is saying that if a virtualenv exists, we should use it, otherwise use default? I'd rather it was written like this (like we do in `bootstrap.sh`)

```
if [ ! $VIRTUAL_ENV ]; then
  virtualenv ./venv
  . ./venv/bin/activate
fi
```

although that means that if there isn't a virtualenv active, it creates a new virtualenv in the current directory, which this one doesn't

PR #610, opened 2016-08-18
Template queryies

minglis:
DO NOT MERGE IS A BREAKING CHANGE ON ADMIN

New query to enable the template stats to be driven from notifications history.

New query in DAO and rest endpoint rewired.

Note. Does not use marshmallow, so response is serialized by hand in the rest endpoint. Response format has changed.

Previous:
{
        'template': {
            'name': 'Brine Shrimp',
            'template_type': 'sms',
            'id': 1
        },
        'id': '6005e192-4738-4962-beec-ebd982d0b03f',
        'day': '2016-04-06',
        'usage_count': 6,
        'service': '1491b86f-c950-48f5-bed1-2a55df027ecb'
    },

Current:
 {
        'template_type': 'sms',
        'template_name': 'one',
        'template_id': 'id-1',
        'count': 100,
        'day': '2016-01-01'
    },

I removed the notion of nested objects and simplified the structure. There were 3 IDs previously in the JSON. Now only template ID.

This breaks the admin and the admin pull request will need to be carefully managed. 

Feedback welcome. Not sure where all those IDs were used by the app and tests all pass in admin, on the other pull request so ¯\_(ツ)_/¯

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -137,23 +137,27 @@ def dao_get_7_day_agg_notification_statistics_for_service(service_id,
 @statsd(namespace="dao")
 def dao_get_template_usage(service_id, limit_days=None):
 
+    table = NotificationHistory
+
+    if limit_days and limit_days <= 7:  # can get this data from notifications table
+        table = Notification
+
     query = db.session.query(
-        func.count(NotificationHistory.template_id).label('count'),
-        NotificationHistory.template_id,
-        func.DATE(NotificationHistory.created_at).label('day'),
+        func.count(table.template_id).label('count'),
+        table.template_id,
         Template.name,
         Template.template_type
     )
 
-    query_filter = [NotificationHistory.service_id == service_id]
+    query_filter = [table.service_id == service_id]
     if limit_days is not None:
-        query_filter.append(NotificationHistory.created_at >= days_ago(limit_days))
+        query_filter.append(table.created_at >= days_ago(limit_days))
 
     return query.filter(*query_filter) \
         .join(Template)\
-        .group_by(NotificationHistory.template_id, func.DATE(NotificationHistory.created_at), Template.name, Template.template_type)\
-        .order_by(desc(func.DATE(NotificationHistory.created_at)), asc(Template.name))\
-        .all()  # noqa
+        .group_by(table.template_id, Template.name, Template.template_type)\
+        .order_by(asc(Template.name))\
+        .all()
 
 
 @statsd(namespace="dao")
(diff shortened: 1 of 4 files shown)


app/dao/notifications_dao.py
```diff
@@ -136,6 +135,28 @@ def dao_get_7_day_agg_notification_statistics_for_service(service_id,
 
 
 @statsd(namespace="dao")
+def dao_get_template_usage(service_id, limit_days=None):
```
leohemsted: a brief check suggests that limit_days is either not passed in or set to 7 - ideally i'd say replace with a boolean flag, then we can just either select from `Notification` or `NotificationHistory` - saves having to filter on `created_at`

app/dao/notifications_dao.py
```diff
@@ -136,6 +135,28 @@ def dao_get_7_day_agg_notification_statistics_for_service(service_id,
...
+        Template.template_type
+    )
+
+    query_filter = [NotificationHistory.service_id == service_id]
+    if limit_days is not None:
+        query_filter.append(NotificationHistory.created_at >= days_ago(limit_days))
+
+    return query.filter(*query_filter) \
+        .join(Template)\
+        .group_by(NotificationHistory.template_id, func.DATE(NotificationHistory.created_at), Template.name, Template.template_type)\
```
leohemsted: do we need to group by day for any reason? Given that we're changing the API schema anyway, I feel like we can take the extra step and just group by template_id - that then gets rid of that silly aggregate_usage function on the admin side

app/dao/notifications_dao.py
leohemsted: actually it's a bit gnarly but could do that keeping limit_days, but just saying

```
notification_table = Notification if limit_days == '7' else NotificationHistory
...
db.session.query(
    notification_table.template_id
    ...
```

app/dao/notifications_dao.py
```diff
@@ -136,6 +135,28 @@ def dao_get_7_day_agg_notification_statistics_for_service(service_id,
...
+        Template.name,
+        Template.template_type
+    )
+
+    query_filter = [NotificationHistory.service_id == service_id]
+    if limit_days is not None:
+        query_filter.append(NotificationHistory.created_at >= days_ago(limit_days))
+
+    return query.filter(*query_filter) \
+        .join(Template)\
```
leohemsted: This should really be joining to TemplateHistory - the query results shouldn't change, but it means that in future we can remove the fkey relationship completely (but only after 
https://github.com/alphagov/notifications-api/pull/599 is merged)

app/dao/notifications_dao.py
minglis: Good call on this. Done:

  table = NotificationHistory

    if limit_days and limit_days < 7:  # can get this data from notifications table
        table = Notification

leohemsted:
looks good 👍

PR #614, opened 2016-08-22
actually bump version

leohemsted:
missed a number last time 😧


leohemsted:
note: this is actually testable - see https://travis-ci.org/alphagov/notifications-api/builds/154153273#L400

we download 8.7.6

quis:
I’ve made this mistake so many times. Is there any reason we need the `egg=…` bit?

leohemsted:
good point - doesn't appear to be required at first glance but i'll check more thoroughly later

leohemsted:
@quis it appears that if you don't provide the `#egg` it clones the entire repository and fetches every tag before selecting the right one 😐

PR #617, opened 2016-08-22
update ukvi org branding spelling

leohemsted:
was previously UK Visas and Immigration
now UK Visas & Immigration


minglis:
Does that work on every environment? Is it only PROD related? If so does the build pass if that ID isn't present?

PR #618, opened 2016-08-22
Template useage per

minglis:
Requires https://github.com/alphagov/notifications-api/pull/610

Is a breaking ADMIN change.

Changes last template usage query to be based on history table and gets last row for that template to allow us to work out what the last use for that template was.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -39,11 +39,9 @@ def dao_get_notification_statistics_for_service(service_id, limit_days=None):
     query_filter = [NotificationStatistics.service_id == service_id]
     if limit_days is not None:
         query_filter.append(NotificationStatistics.day >= days_ago(limit_days))
-    return NotificationStatistics.query.filter(
-        *query_filter
-    ).order_by(
-        desc(NotificationStatistics.day)
-    ).all()
+    return NotificationStatistics.query.filter(*query_filter)\
+        .order_by(desc(NotificationStatistics.day))\
+        .all()
 
 
 @statsd(namespace="dao")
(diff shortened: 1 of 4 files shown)


app/dao/notifications_dao.py
```diff
@@ -155,28 +154,19 @@ def dao_get_template_usage(service_id, limit_days=None):
...
-        TemplateStatistics.template_id == template_id
-    ).order_by(
-        desc(TemplateStatistics.updated_at)
-    ).all()
+def dao_get_last_template_usage(template_id):
+    return NotificationHistory.query.filter(
+        NotificationHistory.template_id == template_id
+    ).join(Template) \
+        .order_by(desc(NotificationHistory.created_at)) \
+        .first()
```
leohemsted: i don't mind you using `.foo() \` for sqlalchemy line continuation (I prefer `).foo(`) but can we at least keep single queries consistent in code style, this is a pain to read

tests/app/template_statistics/test_rest.py
```diff
@@ -157,62 +153,48 @@ def test_returns_empty_list_if_no_templates_used(notify_api, sample_service):
...
                 headers=[('Content-Type', 'application/json'), auth_header],
             )
 
             assert response.status_code == 200
-            json_resp = json.loads(response.get_data(as_text=True))
-            assert len(json_resp['data']) == 2
-            assert json_resp['data'][0]['id'] == str(template_1_stats_2.id)
-            assert json_resp['data'][1]['id'] == str(template_1_stats_1.id)
+            json_resp = json.loads(response.get_data(as_text=True))['data']
+            print(json_resp)
```
leohemsted: rogue print

leohemsted:
👍  apart from that print

PR #623, opened 2016-08-23
Test fixes

leohemsted:
Cleaned up some pep8 failures that for some unknown reason weren't previously tripping on local pep8 runs or on travis - but are certainly things that weren't quite right (spacing, indentation, etc).

Also call generate-version-file as a dependancy of `make test`, cos it is one.

Also a little bit of makefile venv cleanup stuff (calling $VIRTUAL_ENV in `scripts/run_tests.sh`). there's more to do but that'd involve some thinking that i'm not prepared for right now


scripts/run_tests.sh
```diff
@@ -24,7 +24,7 @@ function display_result {
   fi
 }
 
-if [ -d venv ]; then
+if [ -z "$VIRTUAL_ENV" ]; then
```
bandesz: this should be probably ```[[ -z "$VIRTUAL_ENV" ]] && [[ -d venv ]]```, because we still need to check if venv exists

Makefile
```diff
@@ -75,7 +75,7 @@ upload-codedeploy-artifact: check-env-vars
 	aws s3 cp --region eu-west-1 target/notifications-api.zip s3://${DNS_NAME}-codedeploy/notifications-api-${DEPLOY_BUILD_NUMBER}.zip
 
 .PHONY: test
-test: venv ## Run tests
+test: venv generate-version-file ## Run tests
```
bandesz: Ok, it won't hurt if we add it here, but obviously anyone should run build (which generates the version file) and then test. We just don't want to add the build here as a dependency.

leohemsted: yeah didn't want to do that cos the build step screws up with me due to me not wanting a local venv

minglis:
Is the README up to date on how to do this?

leohemsted:
@minglis it is now

PR #624, opened 2016-08-23
Don’t prefix text messages if sender name is set

quis:
Implements:
- [x] https://github.com/alphagov/notifications-utils/pull/66


quis:
Seeing if Jenkins will rebuild this when I reopen it…

PR #627, opened 2016-08-23
Returns the outcome statistics for the job on the API call.

minglis:
BREAKING CHANGE FOR THE ADMIN APP

- Refactors the JOBS api to use the notifications history table to build up the statuses for the notifications.

- Groups by status

- returns a new block on the API response

{
 'statistics' [{
'status': 'created, 'count': 1,
...
}]
}

### DIFF
--- a/tests/app/job/test_rest.py
+++ b/tests/app/job/test_rest.py
@@ -9,7 +9,7 @@ import app.celery.tasks
 from tests import create_authorization_header
 from tests.app.conftest import (
     sample_job as create_job,
-    sample_notification as create_sample_notification)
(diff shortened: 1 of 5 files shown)


tests/app/job/test_rest.py
```diff
@@ -9,7 +9,7 @@
 from tests import create_authorization_header
 from tests.app.conftest import (
     sample_job as create_job,
-    sample_notification as create_sample_notification)
+    sample_notification as create_sample_notification, sample_notification)
```
leohemsted: importing sample_notification twice with different aliases

leohemsted:
apart from that 👍

PR #628, opened 2016-08-24
Platform admin notification query

leohemsted:
(also removed a whole bunch of unused code for getting some stats we already removed previously)

The new function's `dao_fetch_todays_stats_for_all_services`, and it's called by hitting `GET /services?detailed=True`, which returns a list of dicts that looks something like:

```
{
  'id': '1234...',
  'name': 'My Service',
  'created_by': '...',
  ...
  'statistics': {
    'sms': {'created': 0, 'delivered': 0, 'failed': 0},
    'email': {'created': 1, 'delivered': 1, 'failed': 0}
  }
}
```

I reused code from existing dashboard queries to get the stats into that format.

blocks https://github.com/alphagov/notifications-admin/issues/907

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -58,7 +58,7 @@ def create_app(app_name=None):
     encryption.init_app(application)
     clients.init_app(sms_clients=[firetext_client, mmg_client, loadtest_client], email_clients=[aws_ses_client])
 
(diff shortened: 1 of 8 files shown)


minglis:
👍

PR #632, opened 2016-08-24
cleaned up readme + environment file

leohemsted:
* sorted list in README and environment_test.sh
* removed some unused vars
* cleaned up some names to be more accurate in the readme
* removed twilio as a dependency


README.md
```diff
@@ -16,38 +16,42 @@ Get and update notification status.
 mkvirtualenv -p /usr/local/bin/python3 notifications-api
 ```
 
-Creating the environment.sh file. Replace [unique-to-environment] with your something unique to the environment. The local development environments are using the AWS on preview.
+Creating the environment.sh file. Replace [unique-to-environment] with your something unique to the environment. Your aws credentials should be set up for notify-tools (the development/ci aws account).
```
quis: These acronyms should be uppercase:
- aws
- ci

README.md
```diff
@@ -16,38 +16,42 @@ Get and update notification status.
...
 export STATSD_PORT=1000
 export STATSD_PREFIX="stats-prefix"
-export FROM_NUMBER='from_number'
+export VERIFY_CODE_FROM_EMAIL_ADDRESS='no-reply@notify.tools'
 "> environment.sh
 ```
 
-NOTE: the DELIVERY_CLIENT_USER_NAME, DELIVERY_CLIENT_SECRET, NOTIFY_JOB_QUEUE and NOTIFICATION_QUEUE_PREFIX must be the same as the ones in the [notifications-delivery](https://github.com/alphagov/notifications-delivery) app.
-The SECRET_KEY and DANGEROUS_SALT are the same in [notifications-delivery](https://github.com/alphagov/notifications-delivery) and [notifications-admin](https://github.com/alphagov/notifications-admin) app.
+NOTE: The SECRET_KEY and DANGEROUS_SALT are the same in and [notifications-admin](https://github.com/alphagov/notifications-admin) app.
```
quis: s/are the same in and/should match those in

README.md
```diff
@@ -16,38 +16,42 @@ Get and update notification status.
...
-export ADMIN_CLIENT_SECRET='dev-notify-secret-key'
 export ADMIN_CLIENT_USER_NAME='dev-notify-admin'
+export ADMIN_CLIENT_SECRET='dev-notify-secret-key'
+export API_HOST_NAME='http://localhost:6011'
+
 export AWS_REGION='eu-west-1'
+export AWS_ACCESS_KEY_ID=[MY ACCESS KEY]
+export AWS_SECRET_ACCESS_KEY=[MY SECRET]
+
+export SECRET_KEY='dev-notify-secret-key'
```
quis: This one isn’t in order if the order is alphabetical.

README.md
```diff
@@ -16,38 +16,42 @@ Get and update notification status.
...
+export MMG_URL="https://api.mmg.co.uk/json/api.php"
+export NOTIFICATION_QUEUE_PREFIX='[unique-to-environment]' #
+export NOTIFY_EMAIL_DOMAIN='notify.tools'
+export SQLALCHEMY_DATABASE_URI='postgresql://localhost/notification_api'
 export STATSD_ENABLED=True
 export STATSD_HOST="localhost"
 export STATSD_PORT=1000
 export STATSD_PREFIX="stats-prefix"
-export FROM_NUMBER='from_number'
+export VERIFY_CODE_FROM_EMAIL_ADDRESS='no-reply@notify.tools'
```
quis: Is this relevant now that we use Notify to notify?

There’s something going on here but I don’t understand why it’s still a special case: https://github.com/alphagov/notifications-api/blob/020d998baa1f8d763184ee1d29a665acc44628b7/app/notifications/rest.py#L81

leohemsted: so before since we just sent them adhoc i guess we still got callbacks? so we discarded those callbacks based on if they came from our own no-reply email address because it'd fail if it tried to look up the notification in the DB to update stats etc.

looks like that stuff can go now, but i'm hesitant because I'm not sure how the success statistics get recorded and the notification.status changes to `delivered` at all if that's the case? @servingUpAces do you know anything? Did this is_not_a_notification check ever work?

leohemsted: @quis we certainly don't appear to use it - I checked in kibana while registering and inviting, and the log call `SES callback for notify success:. source {} status {}".format(source, notification_status)` doesn't appear in the logs anywhere

quis:
👍

PR #635, opened 2016-08-24
Scheduled delivery of jobs

minglis:
Enables delayed job sending.

Requires: https://github.com/alphagov/notifications-api/pull/631

If the create job API call contains a field "scheduled_date" that is valid (not in past/not more than 24 hours in future) then the new job_status column is set to "scheduled" and the job is NOT places into the process-job queue.

A new celery beat task runs 1 minute past the hour and places all eligible scheduled jobs into the normal process-job queue to be processed in due course. Status is updated to pending.


servingUpAces:
👍

PR #638, opened 2016-08-25
Remove contested writes

minglis:
This pull request removes all writes to the contested tables. 

The affected tables are:
- provider_stats
- template_stats
- notification_stats

So on create or update to a notification we no longer record that event/outcome on these tables. These are no frozen at whatever day we ship this pull request.

Plus we no longer update the job row with the current processing state.

All these writes were optimisations to power read queries for the dashboard, but the performance hit we have seen on writes at scale is too great. Removing all of these.

All dashboard queries now powered by counting notifications. This will need to be looked at again as the data quantity grows.


leohemsted:
👍  looks good, checkout, gave a click about.

leohemsted:
now tested after platform admin changes merged

👍

PR #640, opened 2016-08-25
Scheduled delivery of jobs cleanup

minglis:
Depends on 

- [x] https://github.com/alphagov/notifications-api/pull/635

Tidies up after the DB migrations in the scheduled job.

- migrates existing statues
- drops status column
- not null on job_status

### DIFF
--- a/migrations/versions/0049_set_job_status.py
+++ b/migrations/versions/0049_set_job_status.py
@@ -1,0 +1,22 @@
+"""empty message
+
+Revision ID: 0049_set_job_status
+Revises: 0048_job_scheduled_time
+Create Date: 2016-08-24 13:21:51.744526
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '0049_set_job_status'
+down_revision = '0048_job_scheduled_time'
+
+from alembic import op
+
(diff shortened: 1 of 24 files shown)


quis:
Rebased this against master now that #635 is merged. Previous head was 7b95981.

migrations/versions/0049_set_job_status.py
```diff
@@ -0,0 +1,22 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '0049_set_job_status'
+down_revision = '0048_job_scheduled_time'
+
+from alembic import op
+
+
+def upgrade():
+    op.execute("update jobs set job_status = status where job_status is null")
```
quis: What does `= status` mean here?

minglis: Sets one to the other. So notifications.job_status becomes the value of notifications.status.

quis:
Tested upgrade and downgrade locally, works as expected. 👍

servingUpAces:
👍  Fixed the merge conflicts and migration script conflicts. This is good to go now.

PR #644, opened 2016-08-30
Seperate queues for sending and for db

minglis:
New queues for the sms/email tasks
Previously there were 4 queues for sending messages

The was based on the fact that each notification has 2 actions - persist in the database and send to provider.

Two queues supported the CSV upload - for the first of these tasks
- bulk-email
- build-sms

And there were two more queues for the tasks that make the 3rd party client calls.
- sms
- email

API Calls just used the latter two queues for both tasks

Added four new queues
- db-email
- db-sms
- send-sms
- send-email

So an API call puts a notification into the db-[type] queue first, which then puts the notification into the send-[type] queue

Build queues stay as before.

This will allow us to target processing of these tasks with separate workers to manage these differently.


leohemsted:
👍

PR #647, opened 2016-08-31
Pushed the notifications from the CSV in the DB queue

minglis:
Use the same queues for CSV notifications as for API ones. Looking to push notifications at their own workers, leaving workers for normal "admin" style stuff behind.


quis:
👍

PR #649, opened 2016-08-31
Queues for production

minglis:
Rejigged the queue config.
- the two new queues that handle delivery of notifications (db-[type] / send-[type]) Now no longer are processed by the default workers
- These workers now to the admin type queues for notify (CSV/Validation codes) etc.
- Two new workers are deployed to the AWS environments, one focused on db- tasks and one for send tasks
- These will pick up those queues explicitly.

### DIFF
--- a/config.py
+++ b/config.py
@@ -161,9 +161,9 @@ class Live(Config):
 
 
 configs = {
-    'development': Development(),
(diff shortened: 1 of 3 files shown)


leohemsted:
👍

PR #650, opened 2016-08-31
Return empty stats list for job creation

minglis:
FAO @quis 

Should fix issue we saw on preview.

Adds statistics block onto the job when a create-job is called


quis:
👍

PR #651, opened 2016-08-31
Contract tests

leohemsted:
json schemas! 🎉  like all good json schema related PRs it's kinda huge

### contract tests

added contract tests based off json schemas. had to a little finageling to get the schema to recognise $refs to other files, so I could reuse uuid definitions and such. 

### test requirements 
I've bumped all test libraries to their most recent versions - this was spurred by me requiring new functionality from pytest, which we were a year and a half and a major version behind their stable. Also added `jsonschema` for jsonschema validation, and `strict-rfc3339` which is a dependency for jsonschema date format validation.
Also, removed `mock` - unittest.mock is a python builtin, and has been since 2.7/3.3

### new `client` fixture

@quis long overdue! 
```python
def test_foo(client):
    resp = client.get(...)
    assert resp.status_code == 200
```


servingUpAces:
👍

PR #652, opened 2016-08-31
Fix test to expect correct db queue

quis:
Sending emails was moved to a different queue in https://github.com/alphagov/notifications-api/pull/647

At the time that https://github.com/alphagov/notifications-api/pull/645 was written, the previous queues were still being used. So when #645 got merged it was testing for the old queues, which broke the tests.


leohemsted:
👍  on tests passing

quis:
Jenkins couldn’t clone the repo: https://jenkins.notify.tools/job/notifications-api/99/console

Re-running manually here: https://jenkins.notify.tools/job/notifications-api/100/console

PR #653, opened 2016-08-31
add junit xml reporting to pytest

leohemsted:
so we can get test stats (including time taken, which i'm most interested in) on jenkins. creates test_results.xml in your project root.


bandesz:
👍

PR #656, opened 2016-09-01
Add an endpoint to cancel a job

quis:
If you schedule a job you might change your mind or circumstances might change. So you need to be able to cancel it. This commit adds a `DELETE` endpoint for individual jobs which sets their status to `cancelled`.

I don’t really know what I’m doing here so someone should probably take a close look at it.

### DIFF
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -46,8 +46,12 @@ def dao_get_scheduled_jobs():
 
 def dao_get_future_scheduled_job_by_id_and_service_id(job_id, service_id):
     return Job.query \
-        .filter_by(service_id=service_id, id=job_id) \
-        .filter(Job.job_status == 'scheduled', Job.scheduled_for > datetime.utcnow()) \
+        .filter(
+            Job.service_id == service_id,
+            Job.id == job_id,
+            Job.job_status == 'scheduled',
+            Job.scheduled_for > datetime.utcnow()
+        ) \
         .one()
 
 
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -39,7 +39,10 @@ def dao_get_job_by_id(job_id):
 
 def dao_get_scheduled_jobs():
     return Job.query \
-        .filter(Job.job_status == JOB_STATUS_SCHEDULED, Job.scheduled_for < datetime.utcnow()) \
+        .filter(
+            Job.job_status == JOB_STATUS_SCHEDULED,
+            Job.scheduled_for < datetime.utcnow()
+        ) \
         .order_by(asc(Job.scheduled_for)) \
         .all()
 
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -44,6 +44,13 @@ def dao_get_scheduled_jobs():
         .all()
 
 
+def dao_get_future_scheduled_job_by_id_and_service_id(job_id, service_id):
+    return Job.query \
+        .filter_by(service_id=service_id, id=job_id) \
+        .filter(Job.job_status == 'scheduled', Job.scheduled_for > datetime.utcnow()) \
+        .one()
+
+
 def dao_create_job(job):
     db.session.add(job)
     db.session.commit()
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -2,7 +2,7 @@ from datetime import date, timedelta, datetime
 from sqlalchemy import desc, asc, cast, Date as sql_date
 from app import db
 from app.dao import days_ago
-from app.models import Job, NotificationHistory
+from app.models import Job, NotificationHistory, JOB_STATUS_SCHEDULED
 from app.statsd_decorators import statsd
 from sqlalchemy import func, asc
 
@@ -39,7 +39,7 @@ def dao_get_job_by_id(job_id):
 
 def dao_get_scheduled_jobs():
     return Job.query \
-        .filter(Job.job_status == 'scheduled', Job.scheduled_for < datetime.utcnow()) \
+        .filter(Job.job_status == JOB_STATUS_SCHEDULED, Job.scheduled_for < datetime.utcnow()) \
         .order_by(asc(Job.scheduled_for)) \
         .all()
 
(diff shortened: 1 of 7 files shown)


app/dao/jobs_dao.py
```diff
@@ -44,6 +44,13 @@ def dao_get_scheduled_jobs():
         .all()
 
 
+def dao_get_future_scheduled_job_by_id_and_service_id(job_id, service_id):
+    return Job.query \
+        .filter_by(service_id=service_id, id=job_id) \
+        .filter(Job.job_status == 'scheduled', Job.scheduled_for > datetime.utcnow()) \
+        .one()
```
leohemsted: filter_by and filter are just aliases for each other so can be combined together - `filter` is probably the better one (and then use `==` instead of keyword args)

tests/app/conftest.py
```diff
@@ -289,6 +289,22 @@ def sample_job_with_placeholdered_template(
 
 
 @pytest.fixture(scope='function')
+def sample_scheduled_job(
+    notify_db,
+    notify_db_session,
+    service=None
```
leohemsted: should this be `sample_service`, so it can be used without defining a service separately?

tests/app/dao/test_jobs_dao.py
```diff
@@ -251,3 +252,10 @@ def test_get_scheduled_jobs_gets_ignores_jobs_scheduled_in_the_future(notify_db,
     sample_job(notify_db, notify_db_session, scheduled_for=one_minute_in_the_future, job_status='scheduled')
     jobs = dao_get_scheduled_jobs()
     assert len(jobs) == 0
+
+
+def test_get_future_scheduled_job_gets_a_job_yet_to_send(notify_db, notify_db_session):
+    one_hour_from_now = datetime.utcnow() + timedelta(minutes=60)
+    job = sample_job(notify_db, notify_db_session, scheduled_for=one_hour_from_now, job_status='scheduled')
```
leohemsted: should this use `sample_scheduled_job`?

app/job/rest.py
```diff
@@ -53,6 +55,15 @@ def get_job_by_service_and_job_id(service_id, job_id):
     return jsonify(data=data)
 
 
+@job.route('/<job_id>', methods=['DELETE'])
```
leohemsted: is DELETE the semantically appropriate method? The job's still GETable afterwards. I would have leant towards

`POST /service/{}/job/{}/cancel`

quis: @leohemsted I’ve changed it to `POST /service/{}/job/{}/cancel`

app/dao/jobs_dao.py
quis: Thanks. Changed in 11a4b18.

tests/app/dao/test_jobs_dao.py
quis: Yup, seems reasonable. Changed in 5d4e942.

app/dao/jobs_dao.py
```diff
@@ -44,6 +44,17 @@ def dao_get_scheduled_jobs():
         .all()
 
 
+def dao_get_future_scheduled_job_by_id_and_service_id(job_id, service_id):
+    return Job.query \
+        .filter(
+            Job.service_id == service_id,
+            Job.id == job_id,
+            Job.job_status == 'scheduled',
```
leohemsted: probs best to use the static var to save on magic strings - `from app.models import JOB_STATUS_SCHEDULED`

tests/app/dao/test_jobs_dao.py
```diff
@@ -246,8 +247,15 @@ def test_get_scheduled_jobs_gets_ignores_jobs_not_scheduled(notify_db, notify_db
     assert jobs[0].id == job_scheduled.id
 
 
-def test_get_scheduled_jobs_gets_ignores_jobs_scheduled_in_the_future(notify_db, notify_db_session):
-    one_minute_in_the_future = datetime.utcnow() + timedelta(minutes=1)
-    sample_job(notify_db, notify_db_session, scheduled_for=one_minute_in_the_future, job_status='scheduled')
+def test_get_scheduled_jobs_gets_ignores_jobs_scheduled_in_the_future(
```
leohemsted: two things:

1) you don't need to include notify_db and notify_db_session as fixture params, since they're prerequisites of sample_scheduled_job they'll be loaded anyway

2) does it make sense to add a new test that tests the inverse of this? ie:
```
def test_get_scheduled_jobs_returns_jobs_scheduled_in_the_past(sample_scheduled_job):
    with freeze_time(datetime.now() + timedelta(hours=2)):
        jobs = dao_get_scheduled_jobs()
    assert len(jobs) == 1
```

app/job/rest.py
```diff
@@ -53,6 +55,15 @@ def get_job_by_service_and_job_id(service_id, job_id):
     return jsonify(data=data)
 
 
+@job.route('/<job_id>/cancel', methods=['POST'])
+def cancel_job(service_id, job_id):
+    job = dao_get_future_scheduled_job_by_id_and_service_id(job_id, service_id)
+    job.job_status = JOB_STATUS_CANCELLED
+    dao_update_job(job)
+
+    return get_job_by_service_and_job_id(service_id, job_id)
```
servingUpAces: You don't need to do another query to get the job object, it has been updated by the dao_update_job(job).
Also it's probably better to return json version of the object:
return jsonify(data=job_schema.dump(job).data)

quis: I’m calling the method for getting a job because it saves repeating:
- adding the statistics to the job
- JSONifying the response

See: https://github.com/alphagov/notifications-api/blob/add-cancel-job-endpoint/app/job/rest.py#L50-L55

app/dao/jobs_dao.py
quis: Fixed in 2c17825

tests/app/dao/test_jobs_dao.py
quis: 1. Fixed in e28ce50
2. There’s already a test for this here: https://github.com/alphagov/notifications-api/blob/add-cancel-job-endpoint/tests/app/dao/test_jobs_dao.py#L230-L238

leohemsted:
👍

PR #658, opened 2016-09-05
Drop the status column

minglis:


leohemsted:
👍

quis:
Needs changing to version `0054` because https://github.com/alphagov/notifications-api/pull/663 has been merged.

migrations/versions/0054_perform_drop_status_column.py
```diff
@@ -0,0 +1,23 @@
...
+revision = '0054_perform_drop_status_column'
+down_revision = '0053_cancelled_job_status'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+def upgrade():
+    op.drop_column('jobs', 'status')
```
leohemsted: can we remove the `job_status_types` enum at this stage?

PR #663, opened 2016-09-06
Fix mis-matched database migrations

quis:
The cancelled job migration originally went in as version `0051`.

But `0051_set_job_status` went in first in e3fa8bffd7dd436a9d3da68058528d3ad1cd2c09

This meant that Alembic couldn’t work out what order to run the migrations in, and blew up.

This commit puts the scheduled job migration after all others.


servingUpAces:
👍

PR #664, opened 2016-09-07
Simplified the config. Aim is to get the actual secrets in credstash …

minglis:
…to be env specific, and not the random collection of things we have at the moment.

Secret definition also includes env specific things such as URLs / Queue prefixes / URLs for providers and so on.

Matches:
https://github.com/alphagov/notifications-admin/pull/928


config.py
```diff
@@ -5,27 +5,61 @@
...
+    ADMIN_CLIENT_USER_NAME = 'notify-admin'
+    AWS_REGION = 'eu-west-1'
+    INVITATION_EXPIRATION_DAYS = 2
+    INVITATION_EMAIL_FROM = 'no-reply'
     NOTIFY_APP_NAME = 'api'
     NOTIFY_LOG_PATH = '/var/log/notify/application.log'
     # Notification Queue names are a combination of a prefix plus a name
-    NOTIFICATION_QUEUE_PREFIX = os.environ['NOTIFICATION_QUEUE_PREFIX']
-    SECRET_KEY = os.environ['SECRET_KEY']
+    NOTIFICATION_QUEUE_PREFIX = 'development'
```
leohemsted: this is overriding line 41's env var

config.py
```diff
@@ -5,27 +5,61 @@
...
+    ###########################
+    # Default config values ###
+    ###########################
+
+    DEBUG = False
+    NOTIFY_ENVIRONMENT = 'development'
+    ADMIN_CLIENT_USER_NAME = 'notify-admin'
+    AWS_REGION = 'eu-west-1'
+    INVITATION_EXPIRATION_DAYS = 2
+    INVITATION_EMAIL_FROM = 'no-reply'
```
leohemsted: this isn't used anywhere

config.py
```diff
@@ -5,27 +5,61 @@
...
     # Notification Queue names are a combination of a prefix plus a name
-    NOTIFICATION_QUEUE_PREFIX = os.environ['NOTIFICATION_QUEUE_PREFIX']
-    SECRET_KEY = os.environ['SECRET_KEY']
+    NOTIFICATION_QUEUE_PREFIX = 'development'
     SQLALCHEMY_COMMIT_ON_TEARDOWN = False
-    SQLALCHEMY_DATABASE_URI = os.environ['SQLALCHEMY_DATABASE_URI']
     SQLALCHEMY_RECORD_QUERIES = True
     SQLALCHEMY_TRACK_MODIFICATIONS = True
-    NOTIFY_EMAIL_DOMAIN = os.environ['NOTIFY_EMAIL_DOMAIN']
+    NOTIFY_EMAIL_DOMAIN = 'notify.tools'
```
leohemsted: this should be defined in `Development` rather than in base config i think

config.py
```diff
@@ -96,26 +130,26 @@ class Config(object):
...
     ]
-    API_HOST_NAME = os.environ['API_HOST_NAME']
-    MMG_API_KEY = os.environ['MMG_API_KEY']
-    FIRETEXT_API_KEY = os.getenv("FIRETEXT_API_KEY")
-    LOADTESTING_NUMBER = os.getenv('LOADTESTING_NUMBER')
-    LOADTESTING_API_KEY = os.getenv("LOADTESTING_API_KEY")
-    CSV_UPLOAD_BUCKET_NAME = os.getenv("CSV_UPLOAD_BUCKET_NAME")
+    API_HOST_NAME = "http://localhost:6011"
+
+    CSV_UPLOAD_BUCKET_NAME = 'development-notifications-csv-upload'
```
leohemsted: same with this being in Dev - i just think it's a bit nicer that if we're defining a new environment it's immediately obvious that we aren't using the dev bucket

minglis: Moved to dev config. Each env now defines own bucket with no default.

config.py
minglis: Done

config.py
minglis: Removed

config.py
minglis: Deleted line 41

leohemsted:
👍

PR #665, opened 2016-09-07
Fix test key external number issue

imdadahad:
- Fix issue where sending messages to external numbers with a test key fails
- Add corresponding tests

@servingUpAces does this work as expected?


servingUpAces:
👍

PR #666, opened 2016-09-07
Write to postgres not sqs

minglis:
We wrote to the queue as a performance optimisation , however dev found it confusing to not have immediate access to the notification as it may be perished some minutes later under periods of load. Additionally we had a couple of DB issues which led to us dropping notifications.

Pushing the DB write to earlier in the flow makes the system a little  more robust in the early days, we may want to change this when the traffic increases.


app/models.py
```diff
@@ -482,6 +480,31 @@ def personalisation(self, personalisation):
         if personalisation:
             self._personalisation = encryption.encrypt(personalisation)
 
+    @staticmethod
+    def from_api_request(
```
quis: The Pythonic way to do this is with a class method, eg
```
@classmethod
def from_api_request(cls, args):
    return cls(args)
```
Means that if you ever subclass `Notification` then the method will always call its constructor, not the parent class.

minglis: Done

app/notifications/rest.py
```diff
@@ -320,3 +303,28 @@ def get_notification_statistics_for_day():
...
+        api_key_id=None,
+        key_type=KEY_TYPE_NORMAL,
+):
+    dao_create_notification(
+        Notification.from_api_request(
+            created_at, notification, notification_id, service.id, notification_type, api_key_id, key_type
+        )
+    )
+
+    if notification_type == SMS_TYPE:
```
minglis: Needs to record if this call fails - or push into a retry queue. Otherwise notification is stuck in created.

minglis: We need to ensure that
1) DB write succeeds before SQS call is made.
2) If SQS call fails we need to ensure that the DB row is removed.
3) If SQS call fails we need to ensure a 500 response to the client.

minglis: Also we need to ensure that the send to provider tasks will retry if the DB row is non existent. This allows for the commit to complete.

minglis: If SQS fails on the rest call, we need to delete the created notification.

tests/app/dao/test_notification_dao.py
```diff
@@ -729,6 +729,67 @@ def test_updating_notification_updates_notification_history(sample_notification)
...
+
+    assert Notification.query.count() == 2
+    assert NotificationHistory.query.count() == 2
+
+    dao_delete_notifications_and_history_by_id(notification_1.id)
+
+    assert Notification.query.count() == 1
+    assert NotificationHistory.query.count() == 1
+    assert Notification.query.first().id == notification_2.id
+    assert NotificationHistory.query.first().id == notification_2.id
```
leohemsted: if you change these lines from first to `.one()` then you don't need the count lines above, since .one will raise if there's not exactly one row

leohemsted:
👍

PR #667, opened 2016-09-07
Delete csv file after 7 days

minglis:
Adds new scheduled task to delete CSV files.
Deletes files for jobs older than 7 days, matching delete process for the actual notifications

Runs 1 minutes past the hour at midnight, 1 and 2 am.


leohemsted:
looks good 👍 

although would it result in cleaner code if we just make use of celery's apparent `countdown` or `eta` apply_async args and at the end of process_job just call

```python
    remove_job.apply_async((str(job_id),), queue='remove-job', countdown=timedelta(days=7))
``` 

but we would have to spend a little time testing that to see how robust the countdown argument is, especially regarding server restarts etc

minglis:
Yeah, I -think- that that lives in memory then. Not sure. SQS can delay delivery by 15mins, otherwise I was worried that a celery task keeps those around a long time.

PR #668, opened 2016-09-08
Update Python client

quis:
Just so that nobody else has to do it.

Implements:
- [x] https://github.com/alphagov/notifications-python-client/pull/29

Which is a breaking change, but the breaking bits don’t seem to affect this app.


leohemsted:
👍 

we only use python client for JWT tokens in the auth handler in api app anyway

PR #671, opened 2016-09-13
Statsd time taken to be delivered

minglis:
Record the time between sent_at and now on a provider callback into statsd

- gives elapsed time at provider


quis:
👍

PR #677, opened 2016-09-15
No jobs on api get

minglis:
Ensure that we do not return JOB created notifications in the API calls.

Unless:

It is a LIVE key and the include_jobs param is set to TRUE

Other key types CANNOT retrieve JOB created notifications from the API.

Note:
API calls made by the Admin app ALWAYS return ALL notifications (JOB|API) - this may need to change to create flexibility the in the dashboards, but this is like for like functionality for now.


servingUpAces:
Once functional tests are fixed
👍

PR #678, opened 2016-09-16
Improve error messages if you make an API call with a bad token

quis:
## Give specifc error when service doesn’t exist

If you sign a token with a service ID that doesn’t exist (say, for example, that you get service ID and API key mixed up) then you get an error saying that “no API keys exist for the service”. This is wrong because the service doesn’t even exist.

This commit adds:
- code to check if the service does exist
- a specific error message for this case

The check does mean an extra database call to look up the service. However this only happens _after_ looping through all the API keys. So it shouldn’t have a performance implication for anyone using a valid API key.

## Rewrite authentication error messages more English

giphy

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -46,7 +46,7 @@ def requires_auth():
             continue
 
         if api_key.expiry_date:
-            raise AuthError("Invalid token: revoked", 403)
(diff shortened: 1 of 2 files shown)


imdadahad:
Tested locally 👍

PR #679, opened 2016-09-16
Remove test keys from admin app

minglis:
Depends on
https://github.com/alphagov/notifications-api/pull/677

Ensure we don't return TEST key data on the various calls to the dashboard.

Note this probably requires us to look again at the DAO classes and think of a better interface as logic is getting spread about too much.


servingUpAces:
👍

PR #680, opened 2016-09-16
Update schema to validate contents of GET notifications

imdadahad:
The current schema disallows the links' object within the returned response to have any properties despite being contained in the response (as expected). This removes the aforementioned constraint and implements a check to ensure the link's object contains the expected properties.

### DIFF
--- a/tests/app/public_contracts/schemas/GET_notifications_return.json
+++ b/tests/app/public_contracts/schemas/GET_notifications_return.json
@@ -20,6 +20,9 @@
         },
         "next" : {
           "type" : "string"
+        },
+        "prev" : {
+          "type" : "string"
         }
       }
     },
--- a/tests/app/public_contracts/schemas/GET_notifications_return.json
+++ b/tests/app/public_contracts/schemas/GET_notifications_return.json
@@ -15,16 +15,17 @@
     "links": {
       "type": "object",
       "properties" : {
-        "last": {
+        "prev" : {
           "type" : "string"
         },
         "next" : {
           "type" : "string"
         },
-        "prev" : {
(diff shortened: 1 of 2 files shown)


servingUpAces:
👍

tests/app/public_contracts/schemas/GET_notifications_return.json
```diff
@@ -14,7 +14,14 @@
     },
     "links": {
       "type": "object",
-      "additionalProperties": false
```
leohemsted: do we want this to be removed? what other properties would we have other than last or next?

tests/app/public_contracts/schemas/GET_notifications_return.json
```diff
@@ -14,7 +14,14 @@
...
+      "properties" : {
+        "last": {
+          "type" : "string"
+        },
+        "next" : {
+          "type" : "string"
+        }
```
leohemsted: `"prev": {"type": "string"}` is also an option (see `app/utils.py`)

tests/app/public_contracts/schemas/GET_notifications_return.json
imdadahad: I will update to include 'prev' based on your comment below. I suppose in future if we add more we can update accordingly...

leohemsted: yeah - that's unlikely tho so we should keep in `additionalProperties`

imdadahad: Makes sense, added in 👍

PR #681, opened 2016-09-20
Deploy db migrations

bandesz:
In every environment there will be a t2.nano instance with a separate CodeDeploy configuration to run just the database migrations. This way we only run the migrations once. After the migrations are ready, we run the functional tests first, then deploying the app everywhere and finally and other round of functional tests.

Changes:
* separate appspec-db-migration.yml file to run only the database migrations
* separate zip file is built for the db migrations (only the appspec.yml is different)
* cleaning up the aws scripts

### DIFF
--- a/scripts/aws_start_app.sh
+++ b/scripts/aws_start_app.sh
@@ -1,31 +1,19 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-if [ -e "/etc/init/notifications-api.conf" ]
-then
-  echo "Starting api"
-  sudo service notifications-api start
-fi
+set -eo pipefail
 
-if [ -e "/etc/init/notifications-api-celery-worker.conf" ]
-then
-  echo "Starting celery worker"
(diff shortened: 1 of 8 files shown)


scripts/aws_start_app.sh
```diff
@@ -1,31 +1,19 @@
...
-  echo "Starting celery worker"
-  sudo service notifications-api-celery-worker-db start
-fi
-
-if [ -e "/etc/init/notifications-api-celery-beat.conf" ]
-then
-  echo "Starting celery beat"
-  sudo service notifications-api-celery-beat start
-fi
+start "notifications-api"
```
minglis: Given that the celery scripts are only on delivery nodes, does this throw errors on the API boxes?

bandesz: No, we check which init scripts exist and only start those apps

bandesz:
@minglis said reviewed

PR #684, opened 2016-09-22
Optionally get only notifications created by API

quis:
This commit adds the `include_jobs` filter to the `GET /services/…/notifications` endpoint. It defaults to `True` (i.e. show all notifications) but makes it possible to only return notifications
created by _any_ API key.

This is so that we can show a log of all notifications sent through the API in the admin app.

It does not expose this parameter to the public `GET /notifications` endpoint because this would violate our rules about keys only being able to get notifications created with keys of the same type.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -213,6 +213,7 @@ def get_all_notifications_for_service(service_id):
     page = data['page'] if 'page' in data else 1
     page_size = data['page_size'] if 'page_size' in data else current_app.config.get('PAGE_SIZE')
     limit_days = data.get('limit_days')
+    include_jobs = data.get('include_jobs', True)
 
     pagination = notifications_dao.get_notifications_for_service(
         service_id,
@@ -220,7 +221,7 @@ def get_all_notifications_for_service(service_id):
         page=page,
         page_size=page_size,
         limit_days=limit_days,
-        include_jobs=True)
+        include_jobs=include_jobs)
     kwargs = request.args.to_dict()
     kwargs['service_id'] = service_id
     return jsonify(
--- a/tests/app/service/test_rest.py
+++ b/tests/app/service/test_rest.py
@@ -1059,6 +1059,27 @@ def test_get_all_notifications_for_service_including_ones_made_by_jobs(
         assert response.status_code == 200
 
 
+def test_get_only_api_created_notifications_for_service(
+    client,
+    notify_db,
+    notify_db_session,
+    sample_service
+):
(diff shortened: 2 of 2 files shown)


app/service/rest.py
```diff
@@ -213,14 +213,15 @@ def get_all_notifications_for_service(service_id):
     page = data['page'] if 'page' in data else 1
     page_size = data['page_size'] if 'page_size' in data else current_app.config.get('PAGE_SIZE')
     limit_days = data.get('limit_days')
+    include_jobs = False if ('include_jobs' in data and data['include_jobs'] is False) else True
```
leohemsted: `include_jobs = data.get('include_jobs', False)`

tests/app/service/test_rest.py
```diff
@@ -1059,6 +1059,28 @@ def test_get_all_notifications_for_service_including_ones_made_by_jobs(
         assert response.status_code == 200
 
 
+def test_get_only_api_created_notifications_for_service(
+    notify_api,
+    notify_db,
+    notify_db_session,
+    sample_service
+):
+    with notify_api.test_request_context(), notify_api.test_client() as client:
```
leohemsted: i'm going to keep pushing my `client` pytest fixture until everyone uses it!

just swap out notify_api for client in the parameters and then you don't need this line

tests/app/service/test_rest.py
```diff
@@ -1059,6 +1059,28 @@ def test_get_all_notifications_for_service_including_ones_made_by_jobs(
...
+
+        auth_header = create_authorization_header()
+
+        response = client.get(
+            path='/service/{}/notifications?include_jobs=false'.format(sample_service.id),
+            headers=[auth_header])
+
+        resp = json.loads(response.get_data(as_text=True))
+        assert len(resp['notifications']) == 1
+        assert resp['notifications'][0]['to'] == without_job.to
```
leohemsted: they'll probably both have the same phone number - can you check their `id` instead?

app/service/rest.py
quis: It’s actually `include_jobs = data.get('include_jobs', True)` but yeah, cleaner than the one line `if`.

tests/app/service/test_rest.py
quis: Changed and squashed into 4217bd2

tests/app/service/test_rest.py
quis: Changed and squashed into 4217bd2

app/service/rest.py
quis: Changed and squashed into 4217bd2

PR #686, opened 2016-09-23
Refactor send tasks into shared code

minglis:
Refactor of the code that sends notifications to our 3rd party delivery partners.

- Moved into a new file
- Moved tests accordingly
- New directory /app/delivery to hold code
- Tasks updated to use new location
- New tasks added to use ONLY notification ID not service ID and notification ID. Old tasks remain for now to ensure we empty existing queues before removing matching tasks.

Also adds new REST endpoint to send notifications via an API call.
- also in /app/delivery
- single endpoint takes a notification ID, and passes this on to the send task.
- Synchronous sending

Finally Clients updated to have better exception handling.
- Clients through specific exceptions and check JSON now.
- Calling functions now catch ANY exception so we can retry. One or two errors missed as specific exceptions not handled meaning notifications don't get sent if some exception occurs we are not expecting. Note though at the point of sending we have validated all the data and approved it for delivery so we can be more open about catching exceptions as this notification is ready to be send so we should ensure we give it the best chance of delivery.


app/clients/sms/firetext.py
```diff
@@ -62,11 +61,27 @@ def init_app(self, current_app, statsd_client, *args, **kwargs):
...
 
+    def record_outcome(self, success, response):
+        log_message = "API {} request {} on {} response status_code {}".format(
+            "POST",
+            "succeeded" if success else "failed",
+            self.url,
+            response.status_code
+        )
+
+        if not success:
```
servingUpAces: It's easier to read when if statements are written as positive conditions "if successs then ..." rather than "if not success then .."

tests/app/clients/test_firetext.py
```diff
@@ -88,12 +89,26 @@ def test_send_sms_raises_if_firetext_rejects(mocker, mock_firetext_client):
...
         request_mock.post('https://www.firetext.co.uk/api/sendsms/json', json=response_dict, status_code=200)
         mock_firetext_client.send_sms(to, content, reference)
 
-    assert exc.value.code == 1
-    assert exc.value.description == 'Some kind of error'
+    assert exc.value.status_code == 200
+    assert '"description": "Some kind of error"' in exc.value.text
+    assert '"code": 1' in exc.value.text
+
+
```
servingUpAces: Do we want to test the clients when they do not return json in the response?

app/clients/sms/mmg.py
```diff
@@ -65,6 +67,21 @@ def init_app(self, current_app, statsd_client, *args, **kwargs):
...
 
+    def record_outcome(self, success, response):
+        log_message = "API {} request {} on {} response status_code {}".format(
+            "POST",
+            "succeeded" if success else "failed",
+            self.mmg_url,
+            response.status_code
+        )
+
+        if not success:
```
servingUpAces: It's easier to read if statement with positive conditions.

PR #689, opened 2016-09-23
Service whitelist

leohemsted:
CRUD only

`GET /service/{}/whitelist` to get a dict of existing emails/mobile numbers

`PUT /service/{}/whitelist` to put a complete array of strings of emails/mobile numbers to replace old values with


app/models.py
```diff
@@ -296,6 +329,7 @@ class ProviderDetails(db.Model):
     priority = db.Column(db.Integer, nullable=False)
     notification_type = db.Column(notification_types, nullable=False)
     active = db.Column(db.Boolean, default=False)
+    blah = db.Column
```
servingUpAces: What's a blah column?

tests/app/service/test_service_whitelist.py
```diff
@@ -0,0 +1,87 @@
...
+
+    assert response.status_code == 200
+    assert json.loads(response.get_data(as_text=True)) == {
+        'email_addresses': ['service@example.com'],
+        'mobile_numbers': ['07123456789']
+    }
+
+
+def test_get_whitelist_404s_with_unknown_service_id(client):
+    path = 'service/{}/api-keys'.format(uuid.uuid4())
```
servingUpAces: api_keys?

app/models.py
```diff
@@ -132,6 +139,32 @@ class Service(db.Model, Versioned):
...
 
 
+class ServiceWhitelist(db.Model):
+    __tablename__ = 'service_whitelist'
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), index=True, nullable=False)
+    service = db.relationship('Service', backref='whitelist')
+    email_address = db.Column(db.String(255), nullable=True)
+    mobile_number = db.Column(db.String, nullable=True)
```
servingUpAces: Is the intention of the whitelist to have both an email and mobile number or one or the other?

app/models.py
leohemsted: an incredibly important column that we absolutely can't remove

app/models.py
leohemsted: discussed - determined it's okay to have this column layout as is for now

app/service/rest.py
```diff
@@ -270,3 +280,39 @@ def get_detailed_services():
...
+@service_blueprint.route('/<uuid:service_id>/whitelist', methods=['PUT'])
+def update_whitelist(service_id):
+    # doesn't commit so if there are any errors, we preserve old values in db
+    dao_remove_service_whitelist(service_id)
+    try:
+        whitelist_objs = itertools.chain(
+            [ServiceWhitelist.from_string(service_id, MOBILE_TYPE, recipient)
+             for recipient in request.get_json().get('phone_numbers')],
+            [ServiceWhitelist.from_string(service_id, EMAIL_TYPE, recipient)
+             for recipient in request.get_json().get('email_addresses')])
```
quis: I think this would be easier to read if you factored out some of the code into helper functions, eg:

```python
def get_recipients_from_request(request_json, key, type):
  return [(type, recipient) for recipient in request_json.get(key)]

def get_whitelist_objects(service_id, request_json):
  return [
    ServiceWhitelist.from_string(service_id, type, recipient)
    for type, recipient in (
      get_recipients_from_request(request_json, 'phone_numbers', MOBILE_TYPE) +
      get_recipients_from_request(request_json, 'email_addresses', EMAIL_TYPE)
    )
  ]

whitelist_objs = get_whitelist_objects(service_id, request.get_json())
```

imdadahad: Thanks, added to a separate utils class

PR #691, opened 2016-09-23
Fix template type bug

servingUpAces:
Fixing a bug that allows a sms notification to be sent with an email template and vice versa.

This has been resolved for the post notifications endpoint


minglis:
👍

PR #696, opened 2016-09-28
Research mode queue should be used for research-mode services

minglis:
When notifications are being created, currently they go onto the normal db-sms/db-email and send-sms/send-email queues.

This is true for both research-mode and non-research mode services and for ALL key types.

This has the side effect of affecting delivery performance of the application as these 'test' messages endup being processed by workers that are also processing the genuine notifications.

This PR does 3 things:
- CSV file processing now creates the send-email/send-sms tasks in the research mode queue if the service is in research mode.
- send-email/send-sms tasks now create the deliver-sms/deliver-email tasks in the research-mode queue for research mode services
- API calls that are for research-mode services OR use test keys create the deliver-sms/deliver-email tasks in the research mode queue.

This has the effect of keeping the real queues free from test data. It does also mean however that the worker processing this queue is under more load and this worker also needs to process notify tasks. This isn't an issue as we don't do perf testing in production. But we may want another worker for this queue, esp once people start using TEST keys in anger.

### DIFF
--- a/tests/app/notifications/rest/test_send_notification.py
+++ b/tests/app/notifications/rest/test_send_notification.py
@@ -5,7 +5,7 @@ import pytest
 
 from flask import (json, current_app)
 from freezegun import freeze_time
-from mock import ANY
+from unittest.mock import ANY
 from notifications_python_client.authentication import create_jwt_token
 
 import app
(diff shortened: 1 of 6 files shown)


tests/app/notifications/rest/test_send_notification.py
```diff
@@ -5,6 +5,7 @@
 
 from flask import (json, current_app)
 from freezegun import freeze_time
+from mock import ANY
```
servingUpAces: Please use unittest.mock import ANY

PR #699, opened 2016-09-30
Add DAO and endpoint for getting billable units/financial year

quis:
In order to invoice people we need to know how many text message fragments they’ve sent per month.

This should be per (government) financial year, ie April 1<sup>st</sup> to April 1<sup>st</sup> because we’ll only ever show a page for one year (because the 250,000 allowance is topped up at the start of every financial year).

**Exposed at**
`/services/ef7a665d-11a4-425a-a180-a67ca00b69d7/billable-units?year=2016`

**Returns**
```python
{
    'April': 3,
    'July': 1,
    'August': 1,
    'January': 1
}
```

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -7,7 +7,7 @@ from datetime import (
 
 from flask import current_app
 from werkzeug.datastructures import MultiDict
-from sqlalchemy import (desc, func, or_, and_, asc)
+from sqlalchemy import (desc, func, or_, and_, asc, cast, Text)
 from sqlalchemy.orm import joinedload
 
 from app import db
@@ -210,6 +210,24 @@ def get_notifications_for_job(service_id, job_id, filter_dict=None, page=1, page
 
 
 @statsd(namespace="dao")
+def get_notification_billable_unit_count_per_month(service_id, year):
+    start, end = get_financial_year(year)
+    return db.session.query(
+        func.to_char(NotificationHistory.created_at, "FMMonth"),
+        func.sum(NotificationHistory.billable_units)
+    ).group_by(
+        func.to_char(NotificationHistory.created_at, "FMMonth"),
+        func.to_char(NotificationHistory.created_at, "YYYY-MM")
+    ).order_by(
+        func.to_char(NotificationHistory.created_at, "YYYY-MM")
+    ).filter(
+        NotificationHistory.service_id == service_id,
+        NotificationHistory.created_at >= start,
+        NotificationHistory.created_at < end
+    ).all()
+
+
+@statsd(namespace="dao")
 def get_notification_with_personalisation(service_id, notification_id, key_type):
     filter_dict = {'service_id': service_id, 'id': notification_id}
     if key_type:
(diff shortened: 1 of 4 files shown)


tests/app/service/test_rest.py
```diff
@@ -1300,3 +1300,28 @@ def test_get_detailed_services_only_includes_todays_notifications(notify_db, not
...
+        '/service/{}/billable-units?year=2012'.format(notification.service_id),
+        headers=[create_authorization_header(service_id=notification.service_id)]
+    )
+    assert response.status_code == 200
+    assert json.loads(response.get_data(as_text=True)) == {
+        'December': 1
+    }
+
+
+@freeze_time('2001-01-01T12:00:01')
```
servingUpAces: why do you need freeze_time?

quis: Removed and squashed out.

quis:
@servingUpAces Added a test for a year with no notifications in 3a01c4fa04841f6ce88313f9c4206be543cca0c4

quis:
@servingUpAces think I’ve fixed the timezone thing in 81dc8c8

quis:
## 7abe40b

When we group notifications by month, the months need to be in BST. So the line between one year and another is actually 01:00 on April 1st, _not_ 00:00 on April 1st.

There’s no way we’ve found to do this in SQLAlchemy or raw Postgres, especially because we don’t store the timestamps with a timezone in the database.

So the grouping and summing of the notifications has to be done in Python.

app/dao/notifications_dao.py
```diff
@@ -210,6 +212,30 @@ def get_notifications_for_job(service_id, job_id, filter_dict=None, page=1, page
...
+        NotificationHistory.created_at,
+        NotificationHistory.billable_units
+    ).order_by(
+        NotificationHistory.created_at
+    ).filter(
+        NotificationHistory.billable_units != 0,
+        NotificationHistory.service_id == service_id,
+        NotificationHistory.created_at >= start,
+        NotificationHistory.created_at < end
+    ).all()
```
leohemsted: if you don't call `.all()` here it will stream it into the iterator below, which should be a bit more efficient

app/dao/notifications_dao.py
```diff
@@ -319,3 +345,20 @@ def dao_timeout_notifications(timeout_period_in_seconds):
         update({'status': NOTIFICATION_TEMPORARY_FAILURE, 'updated_at': update_at}, synchronize_session=False)
     db.session.commit()
     return updated
+
+
+def get_financial_year(year):
+    return (get_april_fools(year), get_april_fools(year + 1))
```
leohemsted: minor style thing - you don't need brackets around tuples eg `return a, b`

quis:
**Think I’ve properly fixed the timezones in d352c0e**

- use `astimezone` instead of `replace` because `replace` [doesn’t handle daylight savings time correctly](http://www.saltycrane.com/blog/2009/05/converting-time-zones-datetime-objects-python/#add-timezone-localize)
- create the notifications one second before midnight in BST, because midnight is actually counted as being start of the _next_ day, month, etc

quis:
@leohemsted Addressed those two small comments.

PR #700, opened 2016-10-03
Refactor send notifications

servingUpAces:
Create query to sum notifications for the message limit check, this is a bit more efficient.

Refactored send_notifications method so that it is more readable.
Refactored the test_send_notificaitons so that it uses parametrised test to avoid duplication.


tests/app/dao/test_services_dao.py
```diff
@@ -556,3 +557,14 @@ def test_fetch_weekly_historical_stats_separates_types(notify_db,
     assert ret[1].week_start == datetime(2016, 7, 25)
     assert ret[1].count == 1
     assert ret[1].notification_type == 'sms'
+
+
+def test_dao_fetch_todays_total_message_count_returns_0_when_no_messages_for_today(notify_db,
+                                                                                   notify_db_session,
```
imdadahad: Duplicate test name (line 568)

imdadahad:
Looks good 👍

tests/app/notifications/rest/test_send_notification.py
```diff
@@ -24,21 +23,34 @@
...
-def test_create_sms_should_reject_if_missing_required_fields(notify_api, sample_api_key, mocker):
+
+@pytest.mark.parametrize('template_type',
+                         ['sms', 'email'])
+def test_create_notification_should_reject_if_missing_required_fields(notify_api,
+                                                                      sample_api_key, mocker, template_type):
     with notify_api.test_request_context():
         with notify_api.test_client() as client:
-            mocker.patch('app.celery.provider_tasks.deliver_sms.apply_async')
+            if template_type == 'sms':
```
quis: Rather than have an `if` statement here you could pass in the thing to be patched as part of the `parametrize`, eg:
```
@pytest.mark.parametrize('template_type, patch_target, task', [
    (
        'sms',
        'app.celery.provider_tasks.deliver_sms.apply_async',
        app.celery.provider_tasks.deliver_sms.apply_async
    ),
    …
])
```

PR #701, opened 2016-10-04
Update the database to match the application models

servingUpAces:
There are differences in the models to the database, this script gets the database up to speed with the models.


migrations/versions/0056_minor_updates.py
```diff
@@ -0,0 +1,52 @@
...
+               nullable=False)
+    op.alter_column('services', 'research_mode',
+               existing_type=sa.BOOLEAN(),
+               nullable=False)
+    op.alter_column('services_history', 'research_mode',
+               existing_type=sa.BOOLEAN(),
+               nullable=False)
+    op.alter_column('templates', 'version',
+               existing_type=sa.INTEGER(),
+               nullable=True)
```
leohemsted: dont think this sounds right - ideally set this to be `nullable=False` in the model instead?

servingUpAces:
None of the columns in question are null in production, all of the columns had the default value set to false.

migrations/versions/0056_minor_updates.py
servingUpAces: Ya I think your right

PR #702, opened 2016-10-04
README + bootstrap update

pgchamberlin:
I've updated the README after doing a clean install and hitting a few issues and updated `./scripts/bootstrap.sh` to generate the version file in `./app`.


pgchamberlin:
@leohemsted care to review?

README.md
```diff
@@ -32,18 +36,25 @@ export LOADTESTING_API_KEY="FIRETEXT_SIMULATION_KEY"
 export FIRETEXT_API_KEY="FIRETEXT_ACTUAL_KEY"
 export STATSD_PREFIX="FAKE_PREFIX"
 export NOTIFICATION_QUEUE_PREFIX="PREFIX-TO-IDENTIFY-SQS-QUEUE"
+export AWS_REGION=<YOUR AWS REGION>
+export AWS_ACCESS_KEY_ID=<YOUR ACCESS KEY ID>
+export AWS_SECRET_ACCESS_KEY=<YOUR SECRET ACCESS KEY>
```
leohemsted: can you put these in quotes rather than `<`angle brackets`>`, like above?

README.md
```diff
@@ -32,18 +36,25 @@ export LOADTESTING_API_KEY="FIRETEXT_SIMULATION_KEY"
 export FIRETEXT_API_KEY="FIRETEXT_ACTUAL_KEY"
 export STATSD_PREFIX="FAKE_PREFIX"
 export NOTIFICATION_QUEUE_PREFIX="PREFIX-TO-IDENTIFY-SQS-QUEUE"
+export AWS_REGION=<YOUR AWS REGION>
```
leohemsted: AWS_REGION is always going to be `"eu_west_1"`

README.md
```diff
@@ -15,7 +19,7 @@ mkvirtualenv -p /usr/local/bin/python3 notifications-api
 
 Creating the environment.sh file. Replace [unique-to-environment] with your something unique to the environment. Your AWS credentials should be set up for notify-tools (the development/CI AWS account).
 
-Create a local environment.sh file containing the following:
+Create a local environment.sh file containing the following, updating the AWS key entries with your own values:
```
leohemsted: dont think you need this addition - since you have to update other entries (eg mmg/firetest API keys) it should be apparent that they should be updated. happy to keep it if you want to reword it to not explicitly mention AWS tho

README.md
```diff
@@ -32,18 +36,25 @@ export LOADTESTING_API_KEY="FIRETEXT_SIMULATION_KEY"
...
+
+Assuming you downloaded the ZIP file to `~/Downloads` and expanded it there, you should be able to run the following to install the app:
 
-```shell
-    brew install postgres
 ```
+sudo -b ~/Downloads/Postgres.app/Contents/MacOS/Postgres
+```
+
+This will execute the app as a background process (so you can close the terminal window and it won't stop). The install process should offer to move the app to `Applications`.
```
leohemsted: i feel like this is a bit too hand-holding for a developer instruction set - could be wrong though. i'd probably just point them to postgres and tell them to install it and ensure it is running locally. happy to discuss tho

also while you're editing, we use postgres 9.5 in production, probs worth mentioning that here

README.md
pgchamberlin: Sure

README.md
pgchamberlin: I had that originally, but thought maybe I should generalise. Happy to hard code though :+1:

README.md
pgchamberlin: Sure. I think you might be right. I was feeling a bit thorough when I wrote that! :)

README.md
pgchamberlin: Sure.

pgchamberlin:
@leohemsted I made changes...

 - I removed the AWS config from `environment.sh` entirely as I've found my install plays fine with creds in `~/.aws`.
 - I copied the AWS cred instructions from the admin readme.
 - I organised things inside `## Setting up` using `###` level headings.
 - I OCD'd everything to single quotes in the `environment.sh` copy block.
 - I deleted the too-much-hand-holding.

PR #704, opened 2016-10-07
fix run_scheduled_jobs concurrency bug

leohemsted:
we weren't locking the job table, so if the task runs twice (it in fact runs three times) there is a chance it will pick up the same job to process multiple times. The odds increase as more jobs are scheduled, so this got through to live before we noticed. Bad us.

This PR:
* makes the run_scheduled_jobs lock the job table (only against updates, deletes, and other run_scheduled_jobs tasks) using `FOR UPDATE`
* makes the process_job task check if the job is pending. Note: this is _in no way_ failsafe (as it's not tied using a lock to the update that sets `job_status = 'in progress'` , however it is another line of defence.

### TODO: remember to check concurrency issues on scheduled tasks in future


minglis:
👍

PR #706, opened 2016-10-10
Sort jobs by processed time first

quis:
Say you have a dashboard with some jobs you sent. Normally looks like:

job | sent
--- | ---
file.csv | _5pm_
file.csv | 3pm
file.csv | 1pm
file.csv | 11am

However if your 5pm job was scheduled at lunchtime, then it will look like this:

job | sent
--- | ---
file.csv | 3pm
file.csv | 1pm
file.csv | _5pm_
file.csv | 11am

This is because the jobs are sorted by when they were created, not when they were sent. It looks wrong.

**For jobs that have already been sent**

This commit changes the sort order to be based on `processed_at` instead.

**For upcoming jobs**

If a job doesn’t have a `processed_at` time then it’s scheduled, but hasn’t started yet. Only in this case should we still be sorting by`created_at`.

### DIFF
--- a/tests/app/dao/test_jobs_dao.py
+++ b/tests/app/dao/test_jobs_dao.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+from functools import partial
 import uuid
 
 from freezegun import freeze_time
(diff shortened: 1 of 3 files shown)


tests/app/dao/test_jobs_dao.py
```diff
@@ -192,23 +193,26 @@ def test_get_jobs_for_service_with_limit_days_edge_case(notify_db, notify_db_ses
...
+    scheduled_jobs, processed_jobs = [], []
+
+    for index in range(0, 4):
+        scheduled_jobs.append(_create_job(
+            created_at=datetime.utcnow() - timedelta(seconds=index)
+        ))
+        processed_jobs.append(_create_job(
+            created_at=datetime.utcnow() - timedelta(minutes=(4 - index)),
+            processing_started=datetime.utcnow() - timedelta(hours=index)
+        ))
```
leohemsted: not a huge fan of this test - it's pretty hard to read and work out what the actual times involved are - can we unroll this? (we also shouldn't need 8 entries to test a sort)

leohemsted: also, don't think this currently tests ordering when a scheduled job was created before a second job that has already been processed

quis:
@leohemsted I haven’t completely unrolled it but I think I’ve made it clearer and more robust. See: df0504dddaf65568b92065c6c2e605279d65f387

tests/app/dao/test_jobs_dao.py
```diff
@@ -192,23 +193,32 @@ def test_get_jobs_for_service_with_limit_days_edge_case(notify_db, notify_db_ses
...
+    created_jobs = [
+        _create_job(
+            created_at=_time_from_hour(created_at_hour),
+            processing_started=_time_from_hour(processing_started_hour),
+        )
+        for created_at_hour, processing_started_hour in [
+            (2, None),
+            (1, None),
+            (1, 4),
+            (4, 3),
```
leohemsted: this can't (shouldn't?) exist - i think this should say created at 2am, processed at 3am or similar.

tests/app/dao/test_jobs_dao.py
```diff
@@ -192,23 +193,32 @@ def test_get_jobs_for_service_with_limit_days_edge_case(notify_db, notify_db_ses
...
+        ]
+    ]
```
leohemsted: hmm i still feel
```
from_hour = partial(datetime, 2001, 1, 1)

created_jobs = [
    _create_job(created_at=from_hour(2), processing_started=None),
    _create_job(created_at=from_hour(1), processing_started=None),
    _create_job(created_at=from_hour(1), processing_started=from_hour(4)),
    _create_job(created_at=from_hour(4), processing_started=from_hour(3)),
]
```

is a bit clearer - and it's only 7 lines rather than the 15 lines you need to set up this loop (dont mind if you don't like the partial datetime thats optional)

(less fussed about the assert loop, that's fine)

PR #709, opened 2016-10-11
Test message filtering

leohemsted:
We previously filtered test messages on the front-end - unfortunately this doesn't play well with pagination since if you have a page that only contains jobs - eg:

image

so put this on the API side so its filtered before paginating.

### DIFF
--- a/tests/app/dao/test_jobs_dao.py
+++ b/tests/app/dao/test_jobs_dao.py
@@ -300,3 +300,16 @@ def test_get_jobs_for_service_is_paginated(notify_db, notify_db_session, sample_
     assert len(res.items) == 2
     assert res.items[0].created_at == datetime(2015, 1, 1, 8)
     assert res.items[1].created_at == datetime(2015, 1, 1, 7)
+
+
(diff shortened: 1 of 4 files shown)


tests/app/dao/test_jobs_dao.py
```diff
@@ -32,15 +32,15 @@ def test_should_get_all_statuses_for_notifications_associated_with_job(
...
-    create_notification(notify_db, notify_db_session, service=sample_service, job=sample_job, status='permanent-failure')  # noqa
+    notification = partial(create_notification, notify_db, notify_db_session, service=sample_service, job=sample_job)
+    notification(status='created')
+    notification(status='sending')
+    notification(status='delivered')
+    notification(status='pending')
+    notification(status='failed')
+    notification(status='technical-failure')
+    notification(status='temporary-failure')
+    notification(status='permanent-failure')
```
imdadahad: Can this be wrapped in a for loop?

statuses = ['created', 'sending'...]
for status in statuses:
    create_notification....

imdadahad: Discussed, looks good 👍

PR #711, opened 2016-10-12
Let jobs be scheduled up to 96 hours in the future

quis:
If you want to send a job on Monday morning, you should be able to schedule it on Friday. You shouldn’t need to work on the weekend.

96 hours is a full 4 days, so you can schedule a job at any time on Friday for any time on Monday.

We’ve checked with the information assurance people, and they’re OK with us holding the data for this extra amount of time.

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -39,8 +39,8 @@ def _validate_positive_number(value, msg="Not a positive integer"):
         raise ValidationError(msg)
 
 
-def _validate_datetime_not_more_than_24_hours_in_future(dte, msg="Date cannot be more than 24hrs in the future"):
-    if dte > datetime.utcnow() + timedelta(hours=24):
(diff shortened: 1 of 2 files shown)


app/schemas.py
```diff
@@ -39,8 +39,8 @@ def _validate_positive_number(value, msg="Not a positive integer"):
         raise ValidationError(msg)
 
 
-def _validate_datetime_not_more_than_24_hours_in_future(dte, msg="Date cannot be more than 24hrs in the future"):
-    if dte > datetime.utcnow() + timedelta(hours=24):
+def _validate_datetime_not_more_than_96_hours_in_future(dte, msg="Date cannot be more than 24hrs in the future"):
```
servingUpAces: 24 hours?

quis: Good spot, squashed it out.

PR #718, opened 2016-10-27
add can_send_letters flag to services

leohemsted:
defaults to false

pretty boring otherwise


servingUpAces:
@imdadahad that permission is for the user to send letters, this boolean is to turn sending letters feature on or off at the service, so that we can show the letters features without showing it to everyone since it is still a work in progress (a feature flag)

PR #721, opened 2016-10-31
Post sms v2

servingUpAces:
The first endpoint to be created for version 2 of the public api.

Biggest change is to the package structure, the use of json schema rather than marshmallow. 
The new endpoint is not being used yet.


tests/app/notifications/test_process_notification.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def test_create_content_for_notification_fails_with_missing_personalisation(sample_template_with_placeholders):
+    template = Template.query.get(sample_template_with_placeholders.id)
+    with pytest.raises(BadRequestError):
+        create_content_for_notification(template, None)
+
+
+def test_create_content_for_notification_fails_with_additional_personalisation(sample_template_with_placeholders):
+    template = Template.query.get(sample_template_with_placeholders.id)
```
leohemsted: you don't need this query - just pass `sample_template_with_placeholders` in to the `create_content_for_notification` call

tests/app/notifications/test_process_notification.py
```diff
@@ -0,0 +1,90 @@
...
+
+def test_create_content_for_notification_with_placeholders_passes(sample_template_with_placeholders):
+    template = Template.query.get(sample_template_with_placeholders.id)
+    content = create_content_for_notification(template, {'name': 'Bobby'})
+    assert content.content == template.content
+    assert 'Bobby' in content.replaced
+
+
+def test_create_content_for_notification_fails_with_missing_personalisation(sample_template_with_placeholders):
+    template = Template.query.get(sample_template_with_placeholders.id)
```
leohemsted: you don't need this query - just pass `sample_template_with_placeholders` in to the `create_content_for_notification` call

tests/app/notifications/test_process_notification.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def test_persist_notification_creates_and_save_to_db(sample_template, sample_api_key):
+    assert Notification.query.count() == 0
+    assert NotificationHistory.query.count() == 0
+    notification = persist_notification(sample_template.id, sample_template.version, '+447111111111',
+                                        sample_template.service.id, {}, 'sms', sample_api_key.id,
+                                        sample_api_key.key_type)
+    assert Notification.query.count() == 1
+    assert Notification.query.get(notification.id).__eq__(notification)
```
leohemsted: why not use `==` rather than `__eq__`?

tests/app/notifications/test_validators.py
```diff
@@ -0,0 +1,160 @@
...
+    assert service_can_send_to_recipient("some_other_email@test.com",
+                                         key_type,
+                                         live_service) is None
+    assert service_can_send_to_recipient('07513332413',
+                                         key_type,
+                                         live_service) is None
+
+
+@pytest.mark.parametrize('key_type',
+                         ['team'])
```
leohemsted: dont need this parametrize, just hard code it in the test fn

tests/app/notifications/test_validators.py
```diff
@@ -0,0 +1,160 @@
...
+                         [('team', 'Can’t send to this recipient using a team-only API key'),
+                          ('normal',
+                           "Can’t send to this recipient when service is in trial mode – see https://www.notifications.service.gov.uk/trial-mode")])  # noqa
+def test_service_can_send_to_recipient_fails_when_recipient_is_not_on_team(recipient, key_type, error_message,
+                                                                           notify_db, notify_db_session):
+    trial_mode_service = create_service(notify_db, notify_db_session, service_name='trial mode', restricted=True)
+    with pytest.raises(BadRequestError) as exec_info:
+        assert service_can_send_to_recipient(recipient,
+                                             key_type,
+                                             trial_mode_service) is None
```
leohemsted: since you expect it to raise, you don't need this `assert is None` bit (as if it doesn't raise and does return, that'll fail the test anyway)

tests/app/notifications/test_validators.py
```diff
@@ -0,0 +1,160 @@
...
+    assert exec_info.value.link == 'link to documentation'
+    assert exec_info.value.fields == []
+
+
+def test_service_can_send_to_recipient_fails_when_mobile_number_is_not_on_team(notify_db, notify_db_session):
+    live_service = create_service(notify_db, notify_db_session, service_name='live mode', restricted=False)
+    with pytest.raises(BadRequestError) as e:
+        assert service_can_send_to_recipient("0758964221",
+                                             'team',
+                                             live_service) is None
```
leohemsted: since you expect it to raise, you don't need this `assert is None` bit (as if it doesn't raise and does return, that'll fail the test anyway)

tests/app/v2/notifications/test_notification_schemas.py
```diff
@@ -0,0 +1,91 @@
...
+@pytest.mark.parametrize("input", [valid_json, valid_json_with_optionals])
+def test_post_sms_schema_is_valid(input):
+    validate(input, post_sms_request)
+
+
+def test_post_sms_json_schema_bad_uuid_and_missing_phone_number():
+    j = {"template_id": "notUUID"}
+    try:
+        validate(j, post_sms_request)
+    except ValidationError as e:
```
leohemsted: better to use `with pytest.raises(ValidationError) as e:` else this test will silently pass if it doesn't err

tests/app/v2/notifications/test_notification_schemas.py
```diff
@@ -0,0 +1,91 @@
...
+def test_post_sms_schema_with_personalisation_that_is_not_a_dict():
+    j = {
+        "phone_number": "07515111111",
+        "template_id": str(uuid.uuid4()),
+        "reference": "reference from caller",
+        "personalisation": "not_a_dict"
+    }
+    try:
+        validate(j, post_sms_request)
+    except ValidationError as e:
```
leohemsted: better to use `with pytest.raises(ValidationError) as e:` else this test will silently pass if it doesn't err

tests/app/v2/notifications/test_notification_schemas.py
```diff
@@ -0,0 +1,91 @@
...
+def test_post_sms_response_schema_is_valid(input):
+    validate(input, post_sms_response)
+
+
+def test_post_sms_response_schema_missing_uri():
+    j = valid_response
+    del j["uri"]
+    try:
+        validate(j, post_sms_response)
+    except ValidationError as e:
```
leohemsted: better to use `with pytest.raises(ValidationError) as e:` else this test will silently pass if it doesn't err

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -0,0 +1,96 @@
...
+            response = client.post(
+                path='/v2/notifications/sms',
+                data=json.dumps(data),
+                headers=[('Content-Type', 'application/json')])
+
+            assert response.status_code == 401
+            assert response.headers['Content-type'] == 'application/json'
+            error_resp = json.loads(response.get_data(as_text=True))
+            assert error_resp['code'] == 401
+            assert error_resp['message'] == {'token': ['Unauthorized, authentication token must be provided']}
```
leohemsted: don't think this should be a dict - that's inconsistent with 'message' being a string - might have to tweak our v2 `AuthError` handler

tests/app/notifications/test_process_notification.py
servingUpAces: I tried that but it seems that the template SQLAlchemy object the test is passing into the test does not resolve the __dict__ attribute.

app/v2/notifications/notification_schemas.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def create_post_sms_response_from_notification(notification, content):
+    return {"id": notification.id,
+            "reference": None,  # not yet implemented
+            "content": content,
+            "uri": "v2/notifications/{}".format(notification.id),
+            "template": {"id": notification.template_id,
+                         "version": notification.template_version,
+                         "uri": "v2/templates/{}".format(notification.template_id)}
```
servingUpAces: Yes, I'll add that.

tests/app/notifications/test_process_notification.py
```diff
@@ -0,0 +1,90 @@
...
+        else sample_email_template(notify_db, notify_db_session)
+    notification = sample_notification(notify_db, notify_db_session, template=template, key_type=key_type)
+    send_notification_to_queue(notification=notification, research_mode=research_mode)
+
+    mocked.assert_called_once_with([str(notification.id)], queue=queue)
+
+
+def test_send_notification_to_queue_throws_exception_deletes_notification(sample_notification, mocker):
+    mocked = mocker.patch('app.celery.provider_tasks.deliver_sms.apply_async', side_effect=Boto3Error("EXPECTED"))
+    with pytest.raises(Boto3Error):
```
servingUpAces: The code catches Exception, so does it matter what exception I set the mock to throw?

tests/app/v2/notifications/test_post_notifications.py
servingUpAces: Actually I think that probably belongs in the fields of the error response.

servingUpAces: I've created a to_dict_v2 method on AuthError, better?

leohemsted: yeah that sounds cool 👍

PR #723, opened 2016-11-01
deactivate a service

leohemsted:
new endpoint which deletes templates, revokes keys, and archives a service.

there aren't any checks in the code anywhere to prevent inactive services doing things, however, without a template and/or an api key there's not much we want to block them from so i think that's okay (unless i'm forgetting something really obvious, in which case checks will be pretty trivial to put in).

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -188,7 +188,7 @@ class ApiKey(db.Model, Versioned):
     name = db.Column(db.String(255), nullable=False)
     secret = db.Column(db.String(255), unique=True, nullable=False)
     service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), index=True, nullable=False)
-    service = db.relationship('Service', backref=db.backref('api_keys', lazy='dynamic'))
+    service = db.relationship('Service', backref='api_keys')
     key_type = db.Column(db.String(255), db.ForeignKey('key_types.name'), index=True, nullable=False)
     expiry_date = db.Column(db.DateTime)
     created_at = db.Column(
--- a/app/models.py
+++ b/app/models.py
@@ -5,7 +5,7 @@ from sqlalchemy.dialects.postgresql import (
     UUID,
     JSON
 )
-from sqlalchemy import UniqueConstraint, and_, false, true
+from sqlalchemy import UniqueConstraint, and_
 from sqlalchemy.orm import foreign, remote
 from notifications_utils.recipients import (
     validate_email_address,
--- a/app/models.py
+++ b/app/models.py
@@ -58,7 +58,7 @@ class User(db.Model):
     logged_in_at = db.Column(db.DateTime, nullable=True)
     failed_login_count = db.Column(db.Integer, nullable=False, default=0)
     state = db.Column(db.String, nullable=False, default='pending')
-    platform_admin = db.Column(db.Boolean, nullable=False, default=false())
+    platform_admin = db.Column(db.Boolean, nullable=False, default=False)
 
     @property
     def password(self):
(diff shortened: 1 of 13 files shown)


migrations/versions/0059_set_services_to_active.py
```diff
@@ -0,0 +1,23 @@
...
+
+from alembic import op
+
+
+def upgrade():
+    op.execute('UPDATE services SET active = TRUE')
+
+
+def downgrade():
+    op.execute('UPDATE templates SET archived = FALSE')
```
leohemsted: uhh whoops

tests/app/service/test_deactivate.py
```diff
@@ -0,0 +1,75 @@
...
+        assert key.version == 2
+
+
+def test_deactivating_service_archives_templates(deactivated_service):
+    assert deactivated_service.templates.count() == 2
+    for template in deactivated_service.templates:
+        assert template.archived is True
+        assert template.version == 2
+
+
```
servingUpAces: Do we need to test that the updated timestamp and updated_by are updated?

tests/conftest.py
```diff
@@ -33,6 +33,8 @@ def client(notify_api):
 
 @pytest.fixture(scope='session')
 def notify_db(notify_api, request):
+    assert db.engine.url.database != 'notification_api', 'dont run tests against main db'
```
servingUpAces: Nice

tests/app/service/test_deactivate.py
leohemsted: i don't think so, the versioning code is well enough documented, and - `updated_by` isn't a thing and we've got enough tests around versioning that I think the test below where I assert that there is a `version=2` with `active=False` is enough

servingUpAces:
It would be worth while to have a check the the auth code to reject any called from an inactive service.

leohemsted:
@servingUpAces i haven't implemented any of that yet - that'll be a future PR

app/service/rest.py
```diff
@@ -312,6 +318,29 @@ def update_whitelist(service_id):
         return '', 204
 
 
+@service_blueprint.route('/<uuid:service_id>/deactivate', methods=['POST'])
+def deactivate_service(service_id):
+    service = dao_fetch_service_by_id(service_id)
+
```
servingUpAces: It would be good to wrap this in a transaction, so that a service can not get into a state where it is inactive but has active api keys or templates.

app/models.py
```diff
@@ -115,15 +115,15 @@ class Service(db.Model, Versioned):
...
+    active = db.Column(db.Boolean, index=False, unique=False, nullable=False, default=true())
     message_limit = db.Column(db.BigInteger, index=False, unique=False, nullable=False)
     users = db.relationship(
         'User',
         secondary=user_to_service,
         backref=db.backref('user_to_service', lazy='dynamic'))
     restricted = db.Column(db.Boolean, index=False, unique=False, nullable=False)
-    research_mode = db.Column(db.Boolean, index=False, unique=False, nullable=False, default=False)
-    can_send_letters = db.Column(db.Boolean, nullable=False, default=False)
+    research_mode = db.Column(db.Boolean, index=False, unique=False, nullable=False, default=false())
```
leohemsted: reverted #b2149bf

app/service/rest.py
servingUpAces: That would not solve the problem. We want this action to be all or nothing. I think you need to create one function to do all updates and commit at the end (or rollback)

leohemsted: okay that's fine, but we should probably look at everywhere else - and possibly remove the `@transactional` decorator, if we want to do this - virtually none of our other endpoints have db atomicity

servingUpAces: Typically the endpoints are only updating one resource at a time. The @transactional decorator is useful, but make it difficult to compose updates. We took the same approach when updating the resources for the notification stats (when it was more than one table)

app/models.py
```diff
@@ -5,7 +5,7 @@
     UUID,
     JSON
 )
-from sqlalchemy import UniqueConstraint, and_
+from sqlalchemy import UniqueConstraint, and_, false, true
```
servingUpAces: are false and true needed now?

PR #725, opened 2016-11-03
Refactor saving user profile

imdadahad:
A small refactor of how we save user profile details (removing the schema dependency).

### DIFF
--- a/app/dao/users_dao.py
+++ b/app/dao/users_dao.py
@@ -7,8 +7,18 @@ from app import db
 from app.models import (User, VerifyCode)
 
 
+def _remove_values_for_keys_if_present(dict, keys):
+    for key in keys:
+        dict.pop(key, None)
+
+
 def create_secret_code():
     return ''.join(map(str, random.sample(range(9), 5)))
+
+
+def save_user_attribute(usr, update_dict={}):
+    db.session.query(User).filter_by(id=usr.id).update(update_dict)
+    db.session.commit()
 
 
 def save_model_user(usr, update_dict={}, pwd=None):
@@ -16,9 +26,7 @@ def save_model_user(usr, update_dict={}, pwd=None):
         usr.password = pwd
         usr.password_changed_at = datetime.utcnow()
     if update_dict:
-        if update_dict.get('id'):
-            del update_dict['id']
-            update_dict.pop('password_changed_at')
+        _remove_values_for_keys_if_present(update_dict, ['id', 'password_changed_at'])
         db.session.query(User).filter_by(id=usr.id).update(update_dict)
     else:
         db.session.add(usr)
@@ -74,7 +82,7 @@ def delete_user_verify_codes(user):
     db.session.commit()
 
 
-def get_model_users(user_id=None):
+def get_user_by_id(user_id=None):
     if user_id:
         return User.query.filter_by(id=user_id).one()
     return User.query.filter_by().all()
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -110,8 +110,9 @@ class UserUpdateAttributeSchema(BaseSchema):
     class Meta:
         model = models.User
         exclude = (
-            "updated_at", "created_at", "user_to_service",
-            "_password", "verify_codes")
+            'id', 'updated_at', 'created_at', 'user_to_service',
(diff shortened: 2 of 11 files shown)


app/dao/users_dao.py
```diff
@@ -16,9 +21,7 @@ def save_model_user(usr, update_dict={}, pwd=None):
         usr.password = pwd
         usr.password_changed_at = datetime.utcnow()
     if update_dict:
-        if update_dict.get('id'):
-            del update_dict['id']
-            update_dict.pop('password_changed_at')
+        _remove_values_for_keys_if_present(update_dict, ['id', 'password', 'password_changed_at'])
```
quis: Why are we also `pop`ing the password field? Not saying it’s a bad idea, would just be good to know the reason.

imdadahad: Initially it was because the schema checks were being omitted resulting in the password being included in the update_dict. Anyhow I've changed to put it back in and using a separate schema for my changes.

app/user/rest.py
```diff
@@ -66,9 +69,20 @@ def update_user(user_id):
     return jsonify(data=user_schema.dump(user_to_update).data), 200
 
 
+@user.route('/<uuid:user_id>/update-attribute', methods=['PUT'])
```
minglis: Should this endpoint be:

POST /user/<id>

That would allow updating of fields without the need for a new path.

app/user/rest.py
```diff
@@ -53,7 +56,7 @@ def create_user():
 
 @user.route('/<uuid:user_id>', methods=['PUT'])
 def update_user(user_id):
```
minglis: Is this endpoint marked for deletion? Once we have POST (See other comment) and an update that allows per field updating can we remove the PUT whole user endpoint?

app/user/rest.py
minglis: that should be POST /users/123456

app/user/rest.py
```diff
@@ -4,15 +4,16 @@
 from flask import (jsonify, request, Blueprint, current_app)
```
minglis: Are we adding a "change-password" POST endpoint into this class?

imdadahad: Yeah, was planning to do so after these set of changes were in first

app/user/rest.py
imdadahad: yep, plan is to remove this completely after this change and the password change is in.

app/schemas.py
```diff
@@ -105,6 +105,41 @@ class Meta:
         strict = True
 
 
+class UserUpdateAttributeSchema(BaseSchema):
+
+    class Meta:
+        model = models.User
+        exclude = (
+            "updated_at", "created_at", "user_to_service",
```
servingUpAces: should add id to this list

tests/app/dao/test_users_dao.py
```diff
@@ -37,34 +38,34 @@ def test_create_user(notify_api, notify_db, notify_db_session):
...
-        get_model_users(user_id=fake_uuid)
+        get_user_by_id(user_id=fake_uuid)
         pytest.fail("NoResultFound exception not thrown.")
     except NoResultFound as e:
         pass
 
 
 def test_get_user_invalid_id(notify_api, notify_db, notify_db_session):
     try:
-        get_model_users(user_id="blah")
```
servingUpAces: you can use with pytest.raises(DataError) rather than a try catch

tests/app/user/test_rest.py
```diff
@@ -180,6 +181,29 @@ def test_put_user(notify_api, notify_db, notify_db_session, sample_service):
...
+    resp = client.post(
+        url_for('user.update_user_attribute', user_id=sample_user.id),
+        data=json.dumps(update_dict),
+        headers=headers)
+
+    assert resp.status_code == 200
+    json_resp = json.loads(resp.get_data(as_text=True))
+    assert json_resp['data'][user_attribute] == user_value
+
+
```
servingUpAces: We should have a test to make sure that the endpoint does not allow us to update the excluded fields in the user_update_schema_load_json

imdadahad: Added test_user_update_schema_rejects_disallowed_attribute_keys

PR #726, opened 2016-11-03
update python client to 2.0.0

leohemsted:
this is to prevent 500 errors because <2.0.0 raised AssertionError if supplied JWT tokens were incorrectly formatted

tests added

- [x] https://github.com/alphagov/notifications-python-client/pull/40


minglis:
👍 Should help with those difficult to understand errors.

PR #728, opened 2016-11-07
Make bearer prefix on auth header case insensitive

quis:
From a support ticket:

> the "Bearer" prefix on the auth header is case sensitive. Can this be made case-insensitive?

Sure can 🙃

https://www.pivotaltracker.com/story/show/133183605


minglis:
Is this Case sensitivity not part of the RFC spec? https://tools.ietf.org/html/rfc6750#section-2.1

quis:
I think Postels law applies here:
https://en.m.wikipedia.org/wiki/Robustness_principle

PR #730, opened 2016-11-07
Add ‘letter’ as a possible template type

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/731

So that you can create and get letter templates.

Fairly straightforward with few changes, apart from loads of whitespace-only changes in 1cda00b

Best to look at the commits in isolation because of that last one.


quis:
@minglis Raised a separate PR for the migration here: https://github.com/alphagov/notifications-api/pull/731

quis:
Rebased to remove the migration from this PR.

PR #731, opened 2016-11-10
Add letter to possible template types in DB

quis:
A letter type was added to the `enum` in the `Template` model at the same it was added to the `Notification` model. But the migration was only done for the `notifications` table, not the `templates` table.

See: https://github.com/alphagov/notifications-api/commit/25db1bce#diff-516aab258e161fc65e7564dabd2c625aR19

This commit does the migration to add `letter` as a possible value for the `template_type` column, which is a bit fiddly because `enum`s.

Before:
```
notification_api=# select enum_range(null::template_type);
 enum_range
-------------
 {sms,email}
(1 row)
```

After upgrade:
```
notification_api=# select enum_range(null::template_type);
     enum_range
--------------------
 {sms,email,letter}
(1 row)
```

After downgrade
```
notification_api=# select enum_range(null::template_type);
 enum_range
-------------
 {sms,email}
(1 row)
```

### DIFF
--- a/migrations/versions/0059_add_letter_template_type.py
+++ b/migrations/versions/0059_add_letter_template_type.py
@@ -1,0 +1,56 @@
+"""empty message
+
+Revision ID: f266fb67597a
+Revises: 0058_add_letters_flag
+Create Date: 2016-11-07 16:13:18.961527
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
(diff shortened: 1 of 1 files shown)


quis:
Failing due to low disk space on Jenkins.

quis:
Jenkins is happy again.

PR #732, opened 2016-11-10
Change v2 error resp

servingUpAces:
update V2 error response to  …
{
 status_code: 403,
 errors: [error: AuthError, message: token has expired}]
}

### DIFF
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -19,7 +19,7 @@ def build_error_message(errors, schema):
         field = {s[1]: s[2].strip()}
         fields.append(field)
     message = {
-        "code": "1001",
+        "status_code": 400,
         "message": "Validation error occurred - {}".format(schema['title']),
         "fields": fields
     }
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -1,5 +1,4 @@
 import json
-from collections import OrderedDict
 
 from jsonschema import Draft4Validator, ValidationError
 
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -1,4 +1,6 @@
 import json
+from collections import OrderedDict
+
 from jsonschema import Draft4Validator, ValidationError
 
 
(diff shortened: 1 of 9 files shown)


leohemsted:
i think we should always make it a string now - i can't think of a case where a dict would contain multiple entries (they should be separate items in `errors` now) and i can't think of any error message that we couldn't just make unambiguous to remove the need for the key field

servingUpAces:
I agree. The last commit changes the validation error to just read like a sentence rather than be a dict.

app/schema_validation/__init__.py
```diff
@@ -16,13 +18,11 @@ def build_error_message(errors, schema):
         field = "'{}' {}".format(e.path[0], e.schema.get('validationMessage')) if e.schema.get(
             'validationMessage') else e.message
         s = field.split("'")
-        field = {s[1]: s[2].strip()}
+        field = OrderedDict({"error": "ValidationError", "message": "{}{}".format(s[1], s[2])})
```
leohemsted: OrderedDict doesn't gain us anything here - especially in its current invocation, which would simply keep the ordering of the dict containing `error` and `message`, which as we know is arbitrary.

kill the ordered dict i say

leohemsted: if you really want to sort the keys (which i don't think is especially worthwhile as it'll be an unformatted blob, which when is read in will lose ordering), then you can call `sort_keys=True` on json.dumps


```
>>> json.dumps(x)
'{"b": 1, "a": 1, "c": 1}'
>>> json.dumps(x, sort_keys=True)
'{"a": 1, "b": 1, "c": 1}'
```

tests/app/notifications/test_validators.py
```diff
@@ -49,10 +48,9 @@ def test_check_template_is_for_notification_type_fails_when_template_type_does_n
     with pytest.raises(BadRequestError) as e:
         check_template_is_for_notification_type(notification_type=notification_type,
                                                 template_type=template_type)
-    assert e.value.code == 10400
+    e.value.status_code == 400
```
leohemsted: missing `assert`

app/schema_validation/__init__.py
```diff
@@ -1,4 +1,6 @@
 import json
+from collections import OrderedDict
```
leohemsted: still importing ordereddict

PR #733, opened 2016-11-11
Refactor for code reuse

servingUpAces:
- Refactor send_sms and send_email to use common code to persist the notificaiton.
- Use the same method to persist a notification.
- Refactored the send_sms task to use this method.
- Fix the send_sms and send_email code to send the notification id that has been persisted.
- Fix logging in send_sms and send_email

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -150,12 +150,12 @@ def send_sms(self,
         )
 
     except SQLAlchemyError as e:
-        current_app.logger.exception("RETRY: send_sms notification {}".format(saved_notification.id), e)
+        current_app.logger.exception("RETRY: send_sms notification", e)
         try:
             raise self.retry(queue="retry", exc=e)
         except self.MaxRetriesExceededError:
             current_app.logger.exception(
-                "RETRY FAILED: task send_sms failed for notification {}".format(saved_notification.id),
+                "RETRY FAILED: task send_sms failed for notification",
                 e
             )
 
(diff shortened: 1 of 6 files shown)


tests/app/celery/test_tasks.py
```diff
@@ -337,24 +330,17 @@ def test_should_send_template_to_correct_sms_task_and_persist(sample_template_wi
...
 
-    provider_tasks.deliver_sms.apply_async.assert_called_once_with(
-        [notification_id],
-        queue="send-sms"
-    )
-
-    persisted_notification = Notification.query.filter_by(id=notification_id).one()
-    assert persisted_notification.id == notification_id
+    assert Notification.query.count() == 1
+    persisted_notification = Notification.query.all()[0]
```
servingUpAces: Ah good - will do that.

tests/app/celery/test_tasks.py
```diff
@@ -737,24 +724,27 @@ def test_should_use_email_template_subject_placeholders(sample_email_template_wi
...
+    provider_tasks.deliver_email.apply_async.assert_called_once_with(
+        [persisted_notification.id], queue='send-email'
+    )
 
 
 def test_should_use_email_template_and_persist_without_personalisation(sample_email_template, mocker):
     notification = _notification_json(sample_email_template, "my_email@my_email.com")
     mocker.patch('app.celery.provider_tasks.deliver_email.apply_async')
 
+    assert Notification.query.count() == 0
```
servingUpAces: ok

app/models.py
```diff
@@ -538,52 +538,31 @@ def personalisation(self, personalisation):
...
+                     job_row_number,
+                     created_at):
         return cls(
             template_id=template_id,
             template_version=template_version,
             to=recipient,
             service_id=service_id,
             status='created',
-            created_at=datetime.datetime.utcnow(),
+            created_at=created_at,
```
leohemsted: ah right 👍

app/celery/tasks.py
```diff
@@ -122,27 +127,35 @@ def send_sms(self,
         return
 
     try:
-        dao_create_notification(
-            Notification.from_api_request(
-                created_at, notification, notification_id, service.id, SMS_TYPE, api_key_id, key_type
```
servingUpAces: Yes it is not being used now. I think I have to create a new method without the notificaiton_id, start using the new method, deploy that, then remove the old method. I was going to worry about it in another PR.

PR #735, opened 2016-11-11
Caching with redis

minglis:
Implements a caching layer in Redis.

Adds Flask-Redis to the application, and wraps this in a class to enable switching and other types of logging. Also we catch exceptions unless we want to raise them. Allows Redis failures to not fail the application as primary use case is a cache  and I didn't want to fail a request on a cache read failure.

This reinstates request limits on non-restricted services, now that we can cache what is an expensive query. Cached for an hour. We increment the cache when we do a successful API call or when we create a notification in the database on the send_email or send_sms tasks.

Cache is non-fatal if it errors.

Cache is incremented when a notification is written to the database.

Cache expires after 1 hour.  

NOTE: this PR ships the code, but does not switch it on. I'll do that as a seperate PR.


config.py
```diff
@@ -132,6 +132,9 @@ class Config(object):
     STATSD_HOST = "statsd.hostedgraphite.com"
     STATSD_PORT = 8125
 
+    REDIS_ENABLED = False
+    REDIS_URL = "redis://localhost:6379/0"
```
servingUpAces: Is the redis url different in other environments?

tests/app/celery/test_tasks.py
```diff
@@ -327,34 +327,35 @@ def test_should_process_all_sms_job(sample_job,
 
 
 def test_should_send_template_to_correct_sms_task_and_persist(sample_template_with_placeholders, mocker):
-    notification = _notification_json(sample_template_with_placeholders,
-                                      to="+447234123123", personalisation={"name": "Jo"})
+    with freeze_time("2016-01-01 11:09:00.00000"):
```
servingUpAces: Why does this test need freeze_time?

config.py
minglis: It will be. I'll do another pull request that switches it on, and handles all the configs. I want to hand hold it in preview first.

tests/app/celery/test_tasks.py
minglis: removed

config.py
servingUpAces: ok

PR #736, opened 2016-11-14
Implemented the post email notifications endpoint for v2

servingUpAces:
This PR implements version 2 of the post email notifications, /v2/notifications/email.

### DIFF
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -42,4 +42,4 @@ def build_error_message(errors):
 def __format_message(e):
     s = e.message.split("'")
     msg = "{}{}".format(s[1], s[2])
-    return msg if not e.cause else "'{}' {}".format(e.path[0], e.cause.message)
+    return msg if not e.cause else "{} {}".format(e.path[0], e.cause.message)
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -1,10 +1,21 @@
 import json
 
-from jsonschema import Draft4Validator, ValidationError
+from jsonschema import (Draft4Validator, ValidationError, FormatChecker)
+from notifications_utils.recipients import (validate_phone_number, validate_email_address)
 
 
 def validate(json_to_validate, schema):
-    validator = Draft4Validator(schema)
+    format_checker = FormatChecker()
+
+    @format_checker.checks('phone_number')
+    def validate_schema_phone_number(instance):
+        return validate_phone_number(instance)
+
+    @format_checker.checks('email_address')
+    def validate_schema_email_address(instance):
+        return validate_email_address(instance)
+
+    validator = Draft4Validator(schema, format_checker=format_checker)
     errors = list(validator.iter_errors(json_to_validate))
     if errors.__len__() > 0:
         raise ValidationError(build_error_message(errors, schema))
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -1,38 +1,45 @@
 import json
 
 from jsonschema import (Draft4Validator, ValidationError, FormatChecker)
-from notifications_utils.recipients import (validate_phone_number, validate_email_address)
+from notifications_utils.recipients import (validate_phone_number, validate_email_address, InvalidPhoneError,
+                                            InvalidEmailError)
 
 
 def validate(json_to_validate, schema):
     format_checker = FormatChecker()
 
-    @format_checker.checks('phone_number')
+    @format_checker.checks('phone_number', raises=InvalidPhoneError)
     def validate_schema_phone_number(instance):
-        return validate_phone_number(instance)
+        validate_phone_number(instance)
+        return True
 
-    @format_checker.checks('email_address')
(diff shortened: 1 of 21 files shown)


app/schema_validation/__init__.py
```diff
@@ -1,10 +1,21 @@
...
-    validator = Draft4Validator(schema)
+    format_checker = FormatChecker()
+
+    @format_checker.checks('phone_number')
+    def validate_schema_phone_number(instance):
+        return validate_phone_number(instance)
+
+    @format_checker.checks('email_address')
+    def validate_schema_email_address(instance):
+        return validate_email_address(instance)
```
leohemsted: i think these should catch `InvalidEmail` and `InvalidPhone`, and have this function just return a boolean of "is the email address valid" - if you provide a bad template id and no phone number, then it returns you a `ValidationError` that crucially contains both of those messages - if you provide a bad template id and a _bad_ phone number it returns an `InvalidPhoneError` and does not mention your bad template id. that way `validate` will only ever throw `jsonschema.ValidationError` which is nicer, I think.

app/v2/notifications/notification_schemas.py
```diff
@@ -48,7 +48,50 @@
...
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "description": "Email content for POST email notification",
+    "type": "object",
+    "title": "notification email content",
+    "properties": {
+        "from_email": {"type": "string", "format": "email_address"},
+        "body": {"type": "string"},
+        "subject": {"type": "string"}
+    },
+    "required": ["body"]
```
leohemsted: think `from_email` and `subject` should also be required

tests/app/v2/notifications/test_notification_schemas.py
```diff
@@ -90,3 +101,58 @@ def test_post_sms_response_schema_missing_uri():
     assert error['status_code'] == 400
     assert error['errors'] == [{'error': 'ValidationError',
                                'message': "uri is a required property"}]
+
+
+valid_post_email_json = {"email_address": "test@example.gov.uk",
+                         "template_id": str(uuid.uuid4())
+                         }
+valid_post_emaiL_json_with_optionals = {
```
leohemsted: rogue capital letter in `emaiL`

app/v2/notifications/notification_schemas.py
servingUpAces: good spot

app/schema_validation/__init__.py
servingUpAces: I added raises=InvalidEmailError to the decorator rather than use a try catch. 
The build_error_message method is a bit more messy because of it, I'll put a comment of why it looks like it does.

app/schema_validation/__init__.py
```diff
@@ -1,27 +1,45 @@
...
+from jsonschema import (Draft4Validator, ValidationError, FormatChecker)
+from notifications_utils.recipients import (validate_phone_number, validate_email_address, InvalidPhoneError,
+                                            InvalidEmailError)
 
 
 def validate(json_to_validate, schema):
-    validator = Draft4Validator(schema)
+    format_checker = FormatChecker()
+
+    @format_checker.checks('phone_number', raises=InvalidPhoneError)
```
leohemsted: nice, i like the `raises` thing

tests/app/v2/notifications/test_notification_schemas.py
```diff
@@ -54,6 +56,26 @@ def test_post_sms_schema_with_personalisation_that_is_not_a_dict():
...
+        validate(j, post_sms_request)
+
+
+def test_post_sms_request_invalid_phone_number_and_missing_template():
+    j = {"phone_number": '08515111111',
+         }
+    with pytest.raises(ValidationError) as e:
+        validate(j, post_sms_request)
+    error = json.loads(e.value.message)
+    print(error)
```
leohemsted: 🖨 🚫

tests/app/v2/notifications/test_notification_schemas.py
```diff
@@ -54,6 +56,26 @@ def test_post_sms_schema_with_personalisation_that_is_not_a_dict():
...
+    assert len(error.get('errors')) == 2
```
leohemsted: should we be asserting on these error message contents?

tests/app/v2/notifications/test_notification_schemas.py
```diff
@@ -54,6 +56,26 @@ def test_post_sms_schema_with_personalisation_that_is_not_a_dict():
...
 
 
+@pytest.mark.parametrize('invalid_phone_number',
+                         ['08515111111', '07515111*11', 'notaphoneumber'])
+def test_post_sms_request_invalid_phone_number(invalid_phone_number):
+    j = {"phone_number": invalid_phone_number,
+         "template_id": str(uuid.uuid4())
+         }
+    with pytest.raises(ValidationError):
+        validate(j, post_sms_request)
```
leohemsted: should we be asserting on the error message contents?

PR #737, opened 2016-11-14
archive service - don't re-delete templates etc

leohemsted:
don't re-delete already archived/revoked templates/api-keys

it was making extra versions and changing expiry dates which is bad


imdadahad:
Looks good 👍

PR #747, opened 2016-11-25
Protect send_sms send_email task from being executed twice

servingUpAces:
Sometimes a message is picked up twice of the SQS queue, we need to safe guard ourselves for that.

In this PR the id for the notification is passed in and used to created the notification, which causes a integrity error.
Normally when we get a SQLAlchemy error here we send the message to the retry queue, but if the notification already exists
we just ignore it.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,4 +1,5 @@
 from datetime import (datetime)
+
 from flask import current_app
 from notifications_utils.recipients import (
     RecipientCSV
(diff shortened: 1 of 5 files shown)


app/celery/tasks.py
```diff
@@ -205,12 +208,17 @@ def send_email(self, service_id,
...
-                "RETRY FAILED: task send_email failed for notification".format(
-                    notification.get('job', None),
-                    notification.get('row_number', None)), e)
+        if not get_notification_by_id(notification_id):
+            # Sometimes, SQS plays the same message twice. We should be able to catch an IntegrityError, but it seems
+            # SQLAlchemy is throwing a FlushError. So we check if the notification id already exists then do not
+            # send to the retry queue.
+            current_app.logger.exception("RETRY: send_email notification".format(notification.get('job', None),
+                                                                                 notification.get('row_number', None)),
+                                         e)
```
leohemsted: this won't work - the string format will fail and you shouldn't pass in `e` to a `logger.exception` call as it is automatically attached.

PR #748, opened 2016-11-28
Set up REDIS config.

minglis:
This is now ready to go

Switched OFF on DEV and by DEFAULT
Switched ON for TESTS
Switched ON on PREVIEW - STAGING - PRODUCTION

URL set via ENV file - set in environment_test.sh. DEVS will need to update environment.sh. README updated.


README.md
```diff
@@ -42,6 +42,7 @@ export LOADTESTING_API_KEY='FIRETEXT_SIMULATION_KEY'
 export FIRETEXT_API_KEY='FIRETEXT_ACTUAL_KEY'
 export STATSD_PREFIX='YOU_OWN_PREFIX'
 export NOTIFICATION_QUEUE_PREFIX='YOUR_OWN_PREFIX'
+export REDIS_URL="redis://localhost:6379/0"
```
leohemsted: should probably mention just below, after postgres, that you can install redis locally, but will need to change your local config if you want to use it

PR #751, opened 2016-11-29
Refactored a couple of tests that were slightly testing the wrong thing.

minglis:
- Ensure mock used correctly.


app/notifications/validators.py
```diff
@@ -12,6 +12,7 @@ def check_service_message_limit(key_type, service):
     if key_type != KEY_TYPE_TEST:
         cache_key = redis.daily_limit_cache_key(service.id)
         service_stats = redis_store.get(cache_key)
+        print(service_stats)
```
servingUpAces: let's get rid of the print stmt

PR #752, opened 2016-11-29
Update README

jennyd:
- the bootstrap script needs to be run before the app will run, so mention
  that first
- update number of scripts

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ Install [Postgres.app](http://postgresapp.com/). You will need admin on your mac
 
 ##  To run the application
 
+First, run `scripts/bootstrap.sh` to install dependencies and create the databases.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,9 +57,11 @@ Install [Postgres.app](http://postgresapp.com/). You will need admin on your mac
 
 ##  To run the application
 
+First, run `scripts/bootstrap.sh` to install dependencies and create the databases.
+
 You need to run the api application and a local celery instance.
```
quis: This should also say two Celery instances, rather than a Celery instance.

minglis: 1 does to get it working, 2 if you need celery beat?

jennyd: I've updated it to say explicitly that celery beat isn't required.

PR #753, opened 2016-11-30
Records the service ID if present in the request

minglis:
this is used later in logging of requests

- if a client call the service id is stored
- if an admin app call the string "admin-app" is recorded.

depends on 

- [ ] https://github.com/alphagov/notifications-utils/pull/88

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -1,11 +1,10 @@
-from flask import request, _request_ctx_stack, current_app
+from flask import request, _request_ctx_stack, current_app, g
 from sqlalchemy.exc import DataError
 from sqlalchemy.orm.exc import NoResultFound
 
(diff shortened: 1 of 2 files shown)


app/authentication/auth.py
```diff
@@ -50,6 +49,7 @@ def requires_auth():
         raise AuthError("Invalid token: iss not provided", 403)
 
     if client == current_app.config.get('ADMIN_CLIENT_USER_NAME'):
+        g.service_id = current_app.config.get('ADMIN_CLIENT_USER_NAME')
```
leohemsted: does it make sense for us to try and grab [a] service of the user who is currently logged in? is there any way we can make that useful?

leohemsted: im not sure we can meaningfully extract that info cos i think current_service is only stored on the admin

PR #761, opened 2016-12-07
Invite expiry message

servingUpAces:
Update the error message that a user will see if the invitation is expired.


tests/app/accept_invite/test_accept_invite_rest.py
```diff
@@ -20,7 +19,8 @@ def test_accept_invite_for_expired_token_returns_400(notify_api, sample_invited_
         assert response.status_code == 400
         json_resp = json.loads(response.get_data(as_text=True))
         assert json_resp['result'] == 'error'
-        assert json_resp['message'] == {'invitation': ['Invitation has expired']}
+        assert json_resp['message'] == {'invitation': [
```
imdadahad: "This invitation has expired. Please contact your service owner and have them send out another invitation." Something like that?

imdadahad:
👍

PR #765, opened 2016-12-09
Send Notify's 2FA codes via only one queue

jennyd:
At the moment Notify's 2FA codes and internal emails go first to a `notify` queue, but then are put on the shared `send-sms/email` queues along with all other notifications, and they can take some time to arrive if big jobs are being processed.

This change means that these codes won't be delayed by large jobs going through the `send-sms/email` queues. `send_user_sms_code` now works much more like the endpoints for sending notifications, by persisting the notification and only using the `deliver_sms` task (instead of using `send_sms` as well).

The workers consuming the `notify` queue should be able to handle the deliver task as well, so no change should be needed to the celery workers to support this.

I think there's also a change in behaviour here: previously, if the Notify service was in research mode, 2FA codes would not have been sent out, making it impossible to log into the admin. Now, a call to this endpoint will always send out the notification even if we've put the Notify service into research mode, since we set the notification's key type to `normal` and ignore the service's research mode setting when sending the notification to the queue.

I'm opening this now to get feedback on the approach so far for this one endpoint. If people are happy with this I'd aim to carry on to do similar things to the rest of the API endpoints which use notifications internally:

- [ ] `send_user_confirm_new_email`
- [ ] `send_user_email_verification`
- [ ] `send_already_registered_email`
- [ ] `send_user_reset_password`
- [ ] `create_invited_user`

### DIFF
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -70,19 +70,22 @@ def persist_notification(template_id,
     return notification
 
 
-def send_notification_to_queue(notification, research_mode):
-    try:
-        research_mode = research_mode or notification.key_type == KEY_TYPE_TEST
+def send_notification_to_queue(notification, research_mode, queue=None):
+    if research_mode or notification.key_type == KEY_TYPE_TEST:
+        queue = 'research-mode'
+    elif not queue:
         if notification.notification_type == SMS_TYPE:
-            provider_tasks.deliver_sms.apply_async(
-                [str(notification.id)],
-                queue='send-sms' if not research_mode else 'research-mode'
-            )
+            queue = 'send-sms'
         if notification.notification_type == EMAIL_TYPE:
-            provider_tasks.deliver_email.apply_async(
-                [str(notification.id)],
-                queue='send-email' if not research_mode else 'research-mode'
-            )
+            queue = 'send-email'
+
+    if notification.notification_type == SMS_TYPE:
+        deliver_task = provider_tasks.deliver_sms
+    if notification.notification_type == EMAIL_TYPE:
+        deliver_task = provider_tasks.deliver_email
+
+    try:
+        deliver_task.apply_async([str(notification.id)], queue=queue)
     except Exception as e:
         current_app.logger.exception(e)
         dao_delete_notifications_and_history_by_id(notification.id)
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -18,7 +18,11 @@ from app.dao.users_dao import (
 from app.dao.permissions_dao import permission_dao
 from app.dao.services_dao import dao_fetch_service_by_id
 from app.dao.templates_dao import dao_get_template_by_id
-from app.models import SMS_TYPE
+from app.models import SMS_TYPE, KEY_TYPE_NORMAL
+from app.notifications.process_notifications import (
+    persist_notification,
+    send_notification_to_queue
+)
 from app.schemas import (
     email_data_request_schema,
     user_schema,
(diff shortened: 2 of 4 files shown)


app/user/rest.py
```diff
@@ -143,20 +147,20 @@ def send_user_sms_code(user_id):
...
-    })
-    send_sms.apply_async([current_app.config['NOTIFY_SERVICE_ID'],
-                          str(uuid.uuid4()),
-                          verification_message,
-                          datetime.utcnow().strftime(DATETIME_FORMAT)
-                          ], queue='notify')
+    notify_service_id = current_app.config['NOTIFY_SERVICE_ID']
+
+    saved_notification = persist_notification(template_id=sms_code_template_id,
+                                              template_version=sms_code_template.version,
```
quis: Nitpicky, but I don’t like this style of indentation. It means that if the name of the `persist_notification` function ever changes then you get a 6 line diff instead of a 1 line diff. I prefer:
```python
function(
    arg1,
    arg2,
)
```

jennyd: I agree, I copied this from somewhere else :) updated now

app/notifications/process_notifications.py
```diff
@@ -70,19 +70,24 @@ def persist_notification(template_id,
...
 
-def send_notification_to_queue(notification, research_mode):
-    try:
-        research_mode = research_mode or notification.key_type == KEY_TYPE_TEST
+def send_notification_to_queue(notification, research_mode, queue=None):
+    if research_mode or notification.key_type == KEY_TYPE_TEST:
+        queue_to_use = 'research-mode'
+    elif queue is not None:
+        queue_to_use = queue
+    else:
```
leohemsted: other minor thing, but i'd probably use `queue` rather than `queue_to_use` inside the function - then you can get rid of the `elif queue is not None:` block if you change this to `elif not queue:`

tests/app/notifications/test_process_notification.py
```diff
@@ -135,24 +135,28 @@ def test_persist_notification_with_optionals(sample_job, sample_api_key, mocker)
...
+                          (True, None, 'research-mode', 'email', 'team'),
+                          (False, None, 'send-sms', 'sms', 'normal'),
+                          (False, None, 'send-email', 'email', 'normal'),
+                          (False, None, 'send-sms', 'sms', 'team'),
+                          (False, None, 'research-mode', 'sms', 'test'),
+                          (True, 'notify', 'research-mode', 'email', 'normal'),
+                          (False, 'notify', 'notify', 'sms', 'normal'),
+                          (False, 'notify', 'notify', 'email', 'normal'),
+                          (False, 'notify', 'research-mode', 'sms', 'test')])
 def test_send_notification_to_queue(notify_db, notify_db_session,
```
leohemsted: this test could probably be done without DB (probably just a `namedtuple('Notification', ['id', 'notification_type', 'key_type'])`), which would be a nice test perf win considering all the different parametrized test cases here

leohemsted:
barring code style suggestions above, actual flow/implementation seems good :)

app/notifications/process_notifications.py
jennyd: Yeah, I agree it's marginally clearer, and it's definitely shorter :)

tests/app/notifications/test_process_notification.py
jennyd: That makes it 0.7s instead of 2.5s for me locally 🎉  It's worth looking for other places we can do this too - the fixtures often create lots of other dependent objects which we don't always need.

PR #766, opened 2016-12-12
Fix the from number in the post_sms_response.

servingUpAces:
If the service.sms_sender is not set the FROM_NUMBER from the config is returned, which is the default for the notification.


imdadahad:
👍

PR #768, opened 2016-12-14
Update the timeout notifications to use technical-failure 

servingUpAces:
Update the timeout notifications to use technical-failure rather than temporary failure

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -22,7 +22,8 @@ from app.models import (
     NOTIFICATION_CREATED,
     NOTIFICATION_SENDING,
     NOTIFICATION_PENDING,
-    NOTIFICATION_TEMPORARY_FAILURE, KEY_TYPE_NORMAL, KEY_TYPE_TEST)
+    NOTIFICATION_TECHNICAL_FAILURE,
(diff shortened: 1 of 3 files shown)


app/dao/notifications_dao.py
```diff
@@ -353,11 +354,11 @@ def dao_timeout_notifications(timeout_period_in_seconds):
     updated = db.session.query(Notification). \
         filter(Notification.created_at < (datetime.utcnow() - timedelta(seconds=timeout_period_in_seconds))). \
         filter(Notification.status.in_([NOTIFICATION_CREATED, NOTIFICATION_SENDING, NOTIFICATION_PENDING])). \
```
leohemsted: we need to separate out logic for created vs sending/pending

leohemsted: i think at the very least, we should be doing logging here - if somethings in `sending`/`pending` then the fault is with our provider. if something's in `created`, we've dropped the ball.

servingUpAces: Agreed. If the notification is still in sending or pending that should mean we sent it to the provider and we did not get a delivery receipt. Is that a technical-failure or temporary-failure? 
I think that's a temporary-failure. @minglis do you have an opinion?

PR #772, opened 2016-12-15
Fix uri validation in v2 responses

imdadahad:
In our v2 notification responses the URI's are returned with an invalid format. An example is `"http://api.url:6011//v2/templates/be35a391-e912-42e9-82e6-3f4953f6cbb0"` where `//` appears twice. 

This pull request rectifies the responses to ensure the URIs are returned in the correct format. Additionally the JSON schemas have been updated to reflect this and tests have been added. 
These tests exposed issues in our error handling, particularly where we build an array of schema validation errors. This function has also been modified to be more bulletproof to ensure we can build this array properly and not break whilst doing so.

Existing tests have also been refactored to check that we now receive the field name as well as the value that failed validation.

### DIFF
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -45,7 +45,9 @@ def __format_message(e):
     def get_path(e):
         error_path = None
         try:
-            error_path = e.path[0]
+            error_path = e.path.popleft()
+            # no need to catch IndexError exception explicity as
+            # error_path is None if e.path has no items
         finally:
             return error_path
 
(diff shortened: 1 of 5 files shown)


app/schema_validation/__init__.py
```diff
@@ -42,4 +42,25 @@ def build_error_message(errors):
 
 
 def __format_message(e):
-    return e.message.replace("'", "") if not e.cause else "{} {}".format(e.path[0], e.cause.message)
+    def get_path(e):
+        error_path = None
```
servingUpAces: is all this code the same as return e.path[0] or None?

servingUpAces: Ok

PR #775, opened 2016-12-19
Do not write test data to the history table

minglis:
Ensure updates on a research mode service or test key don't touch the…
… history table

- note this is an unexpectedly big change.
- When we create a service we pass the service id to the persist method. This means that we don't have the service available to check if in research mode.
- All calling methods (expecting the one where we use the notify service) have the service available. So rather than reload it I changed the method signature to pass the service, not the ID to persist.
- Touches a few places.

Note this means that the update or create methods will fall over on a null service. But this seems correct.

Goes back to the story which we need to play to make the service available as the API user so that the need to load and pass around services is minimised.


servingUpAces:
I left a comment in the story https://www.pivotaltracker.com/story/show/135208989 which brings up a question of about research mode and how things can look different if you switch in and out. 
I am okay with this PR being merged before addressing that question but we should take another look at it.

PR #776, opened 2016-12-19
audit changes to provider_details table

leohemsted:
uses a history table, like `template_history` or `service_history`

doesn't use the `Versioned` thing because we're only ever updating this from one place (`dao_update_provider_details`) and it's super straightforward what we're actually doing, compared to that history_meta which is some mad shit.

I did take the liberty of making some reasonably chunky QOL changes to the provider_details tests - I added a new fixture that preserves the existing state of your provider_details/history table so that we don't need to worry about it being irreversibly torn down by a test failing, and was able to refactor some tests as a result of that

in test_rest.py, most of the diff is me unindenting, but I did make some other changes too - lots of the old tests did things like "update, set priority to 5, then set priority back to 10", so I was able to remove chunks of those, and also they were doing `GET`s to get IDs to test with, I just hit the DB directly now

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -43,15 +43,13 @@ def deliver_sms(self, notification_id):
         send_to_providers.send_sms_to_provider(notification)
     except Exception as e:
         try:
-            current_app.logger.error(
+            current_app.logger.exception(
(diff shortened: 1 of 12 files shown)


imdadahad:
👍

PR #780, opened 2016-12-21
Add celery beat task to switch providers on failure

imdadahad:
This adds a celery beat task (running every minute) to **automatically** switch providers on slow delivery of notifications. 

# **How it works**

There are some assumptions made here regarding how the task will operate and check for notifications:
- only those sent from the functional test service with a specific template id (refer to [config.py](https://github.com/alphagov/notifications-api/blob/c2a62e58f5dea6b28c15d1b09fc54489df261c55/config.py))
- in the last 10 minutes either `sending` or `delivered` where the delivery time is/took longer than four minutes

Currently to make a given provider the primary, we set the priority = x where x < all other provider priorities. The switching mechanism behaves differently depending on the current state of providers:

- Only attempt to switch between firetext and mmg (can be changed in the future)
- If the provider we want to switch to is inactive (`active=False`), we won't attempt the switch
- If the provider priorities are equal, then we will simply increase the priority of the one we want to be secondary.

# **Testing locally**

1. Make sure providers are active 
2. Set the `FUNCTIONAL_TEST_SERVICE_ID` and `FUNCTIONAL_TEST_TEMPLATE_ID` to a valid service and template id (from your local dev service). 
3. Run `scripts/run_celery_beat.sh`
3. Send an sms notification and manually edit the db to change the `sent_at` for that notification. As long as the  `(created_at - sent_at)` is more than 4 minutes it should be picked up as a 'slow delivery notification'
4. Monitor the celery beat deamon and you should see logs indicating that a provider switch is taking place. The switch can also be verified on the admin dashboard. 

Note: You can play around with the priorities and active flags to validate switching is behaving as expected.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from flask import current_app
 from sqlalchemy.exc import SQLAlchemyError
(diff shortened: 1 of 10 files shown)


imdadahad:
Closing this for now as we no longer want to run the auto-switch based on the functional test results (for now). Instead we will base it on whether the provider returns an exception on delivery. We probably want _this_ at some point in the future but not at the same time!

PR #782, opened 2016-12-28
Conftest cleanups

leohemsted:
This is a proof of concept for a significant cleanup of the db test fixtures - by separating the two use cases of `sample_*` db objects:

* invoking as a fixture parameter so pytest automatically creates objects
* calling manually to create a custom object, perhaps with some specific non-default property. Often used if we want more than one of that object in the database.

So to help this, I've refactored `sample_user` fixture into the fixture and the db call as two separate functions. `sample_user` was chosen as it has no downstream dependencies on other fixtures - this refactor has to be done from leaf nodes else we'll get into cyclical import hell. The fixture just wraps the db call. Those who want to use the fixture still can, with zero changes. 

Those who want to create a custom user instead can write code such as this:

```
from tests.app.db import create_user

my_user = create_user(email_address="foo@bar.com")
```

one caveat: When using this function, the onus is on the developer to ensure that the `notify_db_session` fixture is being used - the sample_user fixture still depends on it, so will tear down, but the create_user function does not, so if you don't use the fixture the data *will persist to the next test*.

### DIFF
--- a/tests/app/dao/test_users_dao.py
+++ b/tests/app/dao/test_users_dao.py
@@ -46,16 +46,17 @@ def test_get_all_users(notify_db_session):
 
 
 def test_get_user(notify_db_session):
-    user = create_user(email='1@test.com')
+    email = '1@test.com'
+    user = create_user(email=email)
     assert get_user_by_id(user_id=user.id).email_address == email
 
 
-def test_get_user_not_exists(notify_db, fake_uuid):
+def test_get_user_not_exists(notify_db_session, fake_uuid):
     with pytest.raises(NoResultFound):
         get_user_by_id(user_id=fake_uuid)
 
 
-def test_get_user_invalid_id(notify_db):
+def test_get_user_invalid_id(notify_db_session):
     with pytest.raises(DataError):
         get_user_by_id(user_id="blah")
 
--- a/tests/app/dao/test_users_dao.py
+++ b/tests/app/dao/test_users_dao.py
@@ -1,10 +1,10 @@
 from datetime import datetime, timedelta
+
 from sqlalchemy.exc import DataError
 from sqlalchemy.orm.exc import NoResultFound
-
-from app import db
 import pytest
 
+from app import db
 from app.dao.users_dao import (
     save_model_user,
     save_user_attribute,
@@ -21,7 +21,8 @@ from app.models import User, VerifyCode
 from tests.app.db import create_user
 
 
-def test_create_user(notify_db_session):
+@pytest.mark.usefixtures('notify_db_session')
+def test_create_user():
     email = 'notify@digital.cabinet-office.gov.uk'
     data = {
         'name': 'Test User',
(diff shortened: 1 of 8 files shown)


tests/app/dao/test_users_dao.py
```diff
@@ -36,62 +37,50 @@ def test_create_user(notify_api, notify_db, notify_db_session):
...
 
 
-def test_get_all_users(notify_api, notify_db, notify_db_session, sample_user):
-    assert User.query.count() == 1
-    assert len(get_user_by_id()) == 1
-    email = "another.notify@digital.cabinet-office.gov.uk"
-    another_user = create_sample_user(notify_db,
-                                      notify_db_session,
-                                      email=email)
+def test_get_all_users(notify_db_session):
```
leohemsted: note that i still need `notify_db_session` as a fixture here to ensure cleanup

tests/app/dao/test_users_dao.py
```diff
@@ -36,62 +38,55 @@ def test_create_user(notify_api, notify_db, notify_db_session):
...
 
 
-def test_get_all_users(notify_api, notify_db, notify_db_session, sample_user):
-    assert User.query.count() == 1
-    assert len(get_user_by_id()) == 1
-    email = "another.notify@digital.cabinet-office.gov.uk"
-    another_user = create_sample_user(notify_db,
-                                      notify_db_session,
-                                      email=email)
+@pytest.mark.usefixtures('notify_db_session')
```
leohemsted: i'm not entire sold on this `usefixtures` thing - my aim with using it is to signify that "this is a fixture with side effects" as opposed to "this is a fixture that gives you something useful" (and also to supress linter warnings about unused functions)

thoughts?

tests/app/db.py
```diff
@@ -0,0 +1,17 @@
...
+
+def create_user(mobile_number="+447700900986", email="notify@digital.cabinet-office.gov.uk"):
+    data = {
+        'name': 'Test User',
+        'email_address': email,
+        'password': 'password',
+        'mobile_number': mobile_number,
+        'state': 'active'
+    }
+    usr = User.query.filter_by(email_address=email).first()
```
quis: us**e**r – I think we can spare the bytes.

tests/app/dao/test_users_dao.py
leohemsted: yeah that's a fair point - and for now, i imagine every almost single other usage of `notify_db_session` is when its being passed through to various fixtures, which will be staying for a while at least

could be more useful in admin, where we've got all of the stubbed out API calls, especially the login ones that we don't care about

leohemsted: this only ever works if you don't have any funcarg method names, otherwise it's not considered, it seems :(

leohemsted: i'm choosing to remove this because I think it's too confusing when it's applicable

PR #783, opened 2016-12-28
Added a date range filter for the get all services end point.

servingUpAces:
When the start_date and end_date query argruments exists in the request,
the query will return the results from the NotificationHistory table for the given date range.
We will need to check the performance of this query, but this will only be used by the platform admin page.

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -262,7 +262,7 @@ def dao_fetch_todays_stats_for_all_services(include_from_test_key=True):
     if not include_from_test_key:
         query = query.filter(Notification.key_type != KEY_TYPE_TEST)
 
-    return query
+    return query.all()
 
 
 @statsd(namespace='dao')
@@ -290,4 +290,4 @@ def fetch_stats_by_date_range_for_all_services(start_date, end_date, include_fro
     if not include_from_test_key:
         query = query.filter(NotificationHistory.key_type != KEY_TYPE_TEST)
 
-    return query
+    return query.all()
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -263,3 +263,31 @@ def dao_fetch_todays_stats_for_all_services(include_from_test_key=True):
         query = query.filter(Notification.key_type != KEY_TYPE_TEST)
 
     return query
+
+
+@statsd(namespace='dao')
+def fetch_stats_by_date_range_for_all_services(start_date, end_date, include_from_test_key=True):
+    query = db.session.query(
+        NotificationHistory.notification_type,
+        NotificationHistory.status,
+        NotificationHistory.service_id,
+        func.count(NotificationHistory.id).label('count')
+    ).select_from(
+        Service
+    ).join(
+        NotificationHistory
+    ).filter(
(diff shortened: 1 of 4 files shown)


app/dao/services_dao.py
```diff
@@ -263,3 +263,31 @@ def dao_fetch_todays_stats_for_all_services(include_from_test_key=True):
...
+        NotificationHistory.status,
+        NotificationHistory.service_id
+    ).order_by(
+        NotificationHistory.service_id
+    )
+
+    if not include_from_test_key:
+        query = query.filter(NotificationHistory.key_type != KEY_TYPE_TEST)
+
+    return query
```
leohemsted: can you make this return `query.all()`?

i wouldn't mind (and indeed, would even encourage it as it results in less data being held in memory), except for the fact we're wrapping this in the `@statsd` timing decorator, and the query won't actually execute until you call .all on it, leading to the timing being misleading otherwise

app/service/rest.py
```diff
@@ -63,11 +64,14 @@ def get_services():
     detailed = request.args.get('detailed') == 'True'
     user_id = request.args.get('user_id', None)
     include_from_test_key = request.args.get('include_from_test_key', 'True') != 'False'
+    start_date = request.args.get('start_date', None)
+    end_date = request.args.get('end_date', None)
```
leohemsted: do we ever expect these to be `None`?

i'm not sure what comparing a date to null does in sql, but i don't like the sound of it.

in the `elif detailed:` block i'd put a  `if not start_date or not end_date: abort(400)` or whatever (i'll leave error message wording as an exercise to the reader)

app/service/rest.py
```diff
@@ -269,18 +273,22 @@ def get_detailed_service(service_id, today_only=False):
     return detailed_service_schema.dump(service).data
 
 
-def get_detailed_services(only_active=False, include_from_test_key=True):
+def get_detailed_services(only_active=False, include_from_test_key=True, start_date=None, end_date=None):
```
leohemsted: start_date and end_date definitely shouldn't default to `None` here

app/service/rest.py
```diff
@@ -269,18 +273,22 @@ def get_detailed_service(service_id, today_only=False):
...
+    if start_date:
+        stats = fetch_stats_by_date_range_for_all_services(start_date=start_date,
+                                                           end_date=end_date,
+                                                           include_from_test_key=include_from_test_key)
+    else:
+        stats = dao_fetch_todays_stats_for_all_services(include_from_test_key=include_from_test_key)
 
     for service_id, rows in itertools.groupby(stats, lambda x: x.service_id):
-        services[service_id].statistics = statistics.format_statistics(rows)
+            services[service_id].statistics = statistics.format_statistics(rows)
```
leohemsted: rogue extra indent?

app/service/rest.py
leohemsted: ah i completely missed that branch, buhhh 👍

app/dao/services_dao.py
servingUpAces: Roger that

PR #785, opened 2017-01-04
Removed the oneOf validation in the get_notification_response schema.

servingUpAces:
Small PR to remove the oneOf validation in the v2 get_notification_response schema. It doesn't work properly with the required element of the schema.


leohemsted:
So to clarify, the checks never worked, so we've just removed it and have no checks on what pairs of things are null now?

I'm fine with that but just checking. :+1:

PR #787, opened 2017-01-09
Performance fix for the monthly billing summary query

servingUpAces:
Calculating the billing units per month was taking too long for a services with a lot of data.

The query now does the sum per month.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -1,14 +1,13 @@
+
 import pytz
 from datetime import (
     datetime,
     timedelta,
-    date
-)
-from itertools import groupby
+    date)
 
 from flask import current_app
 from werkzeug.datastructures import MultiDict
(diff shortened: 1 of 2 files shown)


app/dao/notifications_dao.py
```diff
@@ -217,23 +216,16 @@ def get_notification_billable_unit_count_per_month(service_id, year):
     start, end = get_financial_year(year)
 
     notifications = db.session.query(
-        NotificationHistory.created_at,
-        NotificationHistory.billable_units
-    ).order_by(
-        NotificationHistory.created_at
+        func.date_trunc("month", (NotificationHistory.created_at + extract("timezone", func.timezone("Europe/London", NotificationHistory.created_at)) * timedelta(seconds=1))),
```
leohemsted: if you give this a variable on a line before the query, you can use that variable in the `group_by` statement

PR #788, opened 2017-01-09
Fix performance of billable units query

servingUpAces:
Calculating the billing units per month was taking too long for a services with a lot of data.
The query now does the sum per month.

https://www.pivotaltracker.com/story/show/137027485


app/dao/notifications_dao.py
```diff
@@ -217,23 +216,29 @@ def get_notification_billable_unit_count_per_month(service_id, year):
...
 
     notifications = db.session.query(
-        NotificationHistory.created_at,
-        NotificationHistory.billable_units
-    ).order_by(
-        NotificationHistory.created_at
+        func.date_trunc("month",
+                        (NotificationHistory.created_at +
+                         extract("timezone", func.timezone("Europe/London",
+                                                           NotificationHistory.created_at)) * timedelta(seconds=1))),
```
leohemsted: if you make this a variable above then we won't need the ridiculous group_by and order_by duplication.

also this really needs its own comment explaining what it's doing i think, given the amount of time and trouble we had figuring out the SQL in the first place

app/dao/notifications_dao.py
```diff
@@ -384,13 +389,5 @@ def get_financial_year(year):
...
-        year, 4, 1, 0, 0, 0, 0,
-        pytz.timezone("Europe/London")
-    ).astimezone(pytz.utc)
-
-
-def get_bst_month(datetime):
-    return pytz.utc.localize(datetime).astimezone(
-        pytz.timezone("Europe/London")
-    ).strftime('%B')
+    return pytz.timezone('Europe/London').localize(datetime(year, 4, 1, 0, 0, 0)).astimezone(pytz.UTC).replace(
```
leohemsted: i think this would benefit from a docstring/comment explaining why we need to jump through this hoop, and also splitting this out into two lines, eg 
```
april_start = pytz.timezone('Europe/London').localize(datetime(year, 4, 1, 0, 0, 0))
return april_start.astimezone(pytz.UTC).replace(tzinfo=None)
```

tests/app/dao/test_notification_dao.py
```diff
@@ -1201,7 +1204,9 @@ def test_should_exclude_test_key_notifications_by_default(
...
-    assert start.tzinfo == pytz.utc
-    assert start.isoformat() == '2000-04-01T00:01:00+00:00'
-    assert end.tzinfo == pytz.utc
-    assert end.isoformat() == '2001-04-01T00:01:00+00:00'
+    assert str(start) == '2000-03-31 23:00:00'
+    assert str(end) == '2001-03-31 23:00:00'
+
+
+def test_get_april_fools():
+    assert str(get_april_fools(2016)) == '2016-03-31 23:00:00'
```
leohemsted: can you also assert that ` assert get_april_fools(2016).tzinfo is None`? you can just do it in the same test if you want

app/dao/notifications_dao.py
servingUpAces: Fair point, I'm on it.

app/dao/notifications_dao.py
```diff
@@ -216,24 +215,30 @@ def get_notifications_for_job(service_id, job_id, filter_dict=None, page=1, page
...
+        month,
+        func.sum(NotificationHistory.billable_units)
     ).filter(
         NotificationHistory.billable_units != 0,
         NotificationHistory.service_id == service_id,
         NotificationHistory.created_at >= start,
         NotificationHistory.created_at < end
-    )
+    ).group_by(month
+               ).order_by(month).all()
```
leohemsted: we've just got this funky indentation left 😄

PR #789, opened 2017-01-09
bump utils to 13.0.1

leohemsted:
brings in a fix to InvalidEmail/Phone/AddressExceptions not being instantiated correctly. `exception.message` is not a python standard, so we shouldn't be relying on it to transmit exception reasons - rather we should be using `str(exception)` instead. This involved a handful of small changes to the schema validation

### DIFF
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -30,8 +30,10 @@ def validate(json_to_validate, schema):
 def build_error_message(errors):
     fields = []
     for e in errors:
-        field = "{} {}".format(e.path[0], e.schema.get('validationMessage')) if e.schema.get(
-            'validationMessage') else __format_message(e)
+        field = (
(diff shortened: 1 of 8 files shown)


app/schema_validation/__init__.py
```diff
@@ -52,13 +54,11 @@ def get_path(e):
...
-        error_message = None
-        try:
-            error_message = e.cause.message
-        except AttributeError:
-            error_message = e.message
-        finally:
-            return error_message.replace("'", '')
+        # e.cause is an exception (such as InvalidPhoneError). if it's not present it was a standard jsonschema error
+        # such as a required field not being present
+        print('msg', e.message, 'exc', type(e.cause), e.cause)
```
imdadahad: leftover print

imdadahad:
Everything else looks good 👍

app/schema_validation/__init__.py
leohemsted: 🙈

PR #794, opened 2017-01-17
Priority by template

servingUpAces:
- Update templates.process_type to be `normal` and set the column as not null.
- dB script to default templates/templates_history.process_type = 'normal' and set the column to not null
- Refactor v2 post_notification to use a single method for sms and email.
- Added the `simulate` notification logic to version 2. We have 3 email addresses and phone numbers that are used
to simulate a successful post to /notifications. This was missed out of the version 2 endpoint.
- Added a test to template_dao to check for the default value of normal for new templates
- in v2 get_notifications, casted the path param to a uuid, if not uuid abort(404)
	If the template.process_type = PRIROITY send the message to the notify queue.

We are using the notify queue in this iteration because that queue is a low volume queue with it's own dedicated workers. This just saves us from building a new queue at this point, and a new queue may not be necessary.
- Make the update template endpoint work when process_type is present.


app/notifications/process_notifications.py
```diff
@@ -87,3 +89,9 @@ def send_notification_to_queue(notification, research_mode, queue=None):
     current_app.logger.info(
         "{} {} created at {}".format(notification.notification_type, notification.id, notification.created_at)
```
imdadahad: Should this be here...or be in `persist_notification`?

app/notifications/rest.py
```diff
@@ -233,23 +235,29 @@ def send_notification(notification_type):
...
-                                                  personalisation=notification.get('personalisation', None),
-                                                  notification_type=notification_type,
-                                                  api_key_id=api_user.id,
-                                                  key_type=api_user.key_type)
-        send_notification_to_queue(saved_notification, service.research_mode)
-
-    notification_id = create_uuid() if saved_notification is None else saved_notification.id
+    # Do not persist or send notification to the queue if it is a simulated recipient
+    simulated = simulated_recipient(notification['to'], notification_type)
+    saved_notification = persist_notification(template_id=template.id,
```
imdadahad: Not sure I'm a fan of this. I like the abstraction but if I was reading this on first glance I'd automatically assume that the notification was persisted. 

I would have to zoom into `persist_notification` to know that it isn't.

app/v2/notifications/get_notifications.py
```diff
@@ -1,15 +1,23 @@
...
-@notification_blueprint.route("/<uuid:id>", methods=['GET'])
+@notification_blueprint.route("/<id>", methods=['GET'])
 def get_notification_by_id(id):
+    try:
+        casted_id = uuid.UUID(id)
+    except ValueError:
+        abort(404)
     notification = notifications_dao.get_notification_with_personalisation(
-        str(api_user.service_id), id, key_type=None
+        api_user.service_id, casted_id, key_type=None
```
imdadahad: It might be safer to do `except Exception` as `uuid.UUID(0)` raises an AttributeError.

tests/app/notifications/test_rest.py
```diff
@@ -13,34 +13,43 @@
...
+            'template_type': sample_notification.template.template_type,
+            'version': 1
+        }
+        assert notification['to'] == '+447700900855'
+        assert notification['service'] == str(sample_notification.service_id)
+        assert notification['body'] == "This is a template:\nwith a newline"
+        assert not notification.get('subject')
+
+
+def test_get_sms_notification_by_invalid_id(client, sample_notification):
```
imdadahad: We should probably add a case for `0` based on my comment further above.

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -145,24 +151,76 @@ def test_post_email_notification_returns_201(client, sample_email_template_with_
...
+@pytest.mark.parametrize('recipient, notification_type', [
+    ('simulate-delivered@notifications.service.gov.uk', 'email'),
+    ('simulate-delivered-2@notifications.service.gov.uk', 'email'),
+    ('simulate-delivered-3@notifications.service.gov.uk', 'email'),
+    ('07700 900000', 'sms'),
+    ('07700 900111', 'sms'),
+    ('07700 900222', 'sms')
+])
+def test_should_not_persist_notification_or_send_notification_if_simulated_recipient(
+        client,
```
imdadahad: `test_should_not_persist_or_send_notification_if_simulated_recipient`

app/notifications/process_notifications.py
servingUpAces: I've moved this log message to the persist method and changed this to talk about sending it to the queue.

PR #795, opened 2017-01-17
Auto-switch providers on error

imdadahad:
This will auto-switch to a new provider when the existing provider returns an error in `deliver_sms`.

# **Dependencies (do not have to be merged first)**

- [ ] [Alert through AWS CloudWatch when we switch providers ](https://github.com/alphagov/notifications-aws/pull/116)
- [x] [Show when a provider was last updated](https://github.com/alphagov/notifications-admin/pull/1074)

# **How it works**

When we send an sms notification for delivery to the provider, they could raise an exception (due to some error). In this case we want to switch to another provider. This will account for edge cases where the provider we want to switch to is already set as primary or they are inactive.

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -5,7 +5,6 @@ from sqlalchemy.orm.exc import NoResultFound
 from app import notify_celery
 from app.dao import notifications_dao
 from app.dao.notifications_dao import update_notification_status_by_id
-from app.dao.provider_details_dao import dao_toggle_sms_provider
 from app.statsd_decorators import statsd
 from app.delivery import send_to_providers
 
@@ -44,7 +43,6 @@ def deliver_sms(self, notification_id):
         send_to_providers.send_sms_to_provider(notification)
     except Exception as e:
         try:
-            dao_toggle_sms_provider()
             current_app.logger.exception(
                 "RETRY: SMS notification {} failed".format(notification_id)
             )
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -5,6 +5,7 @@ from sqlalchemy.orm.exc import NoResultFound
 from app import notify_celery
 from app.dao import notifications_dao
 from app.dao.notifications_dao import update_notification_status_by_id
+from app.dao.provider_details_dao import dao_toggle_sms_provider
 from app.statsd_decorators import statsd
 from app.delivery import send_to_providers
 
@@ -43,6 +44,7 @@ def deliver_sms(self, notification_id):
         send_to_providers.send_sms_to_provider(notification)
     except Exception as e:
         try:
+            dao_toggle_sms_provider()
             current_app.logger.exception(
                 "RETRY: SMS notification {} failed".format(notification_id)
             )
--- a/app/dao/provider_details_dao.py
+++ b/app/dao/provider_details_dao.py
@@ -9,8 +9,7 @@ from app import db
 
 
 def get_provider_details():
-    providers = ProviderDetails.query.order_by(asc(ProviderDetails.priority), asc(ProviderDetails.notification_type)).all()
-    return providers
+    return ProviderDetails.query.order_by(asc(ProviderDetails.priority), asc(ProviderDetails.notification_type)).all()
 
 
 def get_provider_details_by_id(provider_details_id):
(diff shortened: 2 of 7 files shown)


tests/app/dao/test_provider_details_dao.py
```diff
@@ -21,15 +28,13 @@ def test_can_get_sms_providers(restore_provider_details):
...
 
 
-def test_can_get_sms_providers_in_order(restore_provider_details):
+def test_can_get_sms_providers_in_order_of_priority(restore_provider_details):
     providers = get_provider_details_by_notification_type('sms')
 
-    assert providers[0].identifier == "mmg"
-    assert providers[1].identifier == "firetext"
-    assert providers[2].identifier == "loadtesting"
+    assert providers[0].priority < providers[1].priority < providers[2].priority
```
leohemsted: nice 👌

tests/app/conftest.py
```diff
@@ -629,13 +634,34 @@ def fake_uuid():
...
+def current_sms_provider():
+    return ProviderDetails.query.filter_by(
+        notification_type='sms'
+    ).order_by(
+        asc(ProviderDetails.priority)
+    ).first()
+
+
+@pytest.fixture(scope='function')
+def set_primary_sms_provider(identifier='mmg'):
```
leohemsted: this isn't used as a fixture anywhere - just pop this function up the top of `test_provider_details_dao`

tests/app/conftest.py
```diff
@@ -629,13 +634,34 @@ def fake_uuid():
...
+
+@pytest.fixture(scope='function')
 def ses_provider():
     return ProviderDetails.query.filter_by(identifier='ses').one()
 
 
 @pytest.fixture(scope='function')
 def firetext_provider():
-    return ProviderDetails.query.filter_by(identifier='mmg').one()
+    return ProviderDetails.query.filter_by(identifier='firetext').one()
```
leohemsted: 😅

tests/app/dao/test_provider_details_dao.py
```diff
@@ -78,3 +83,72 @@ def test_update_adds_history(restore_provider_details):
...
+
+@pytest.mark.parametrize('provider_identifier', ['firetext', 'mmg'])
+def test_get_alternative_sms_provider_returns_expected_provider(notify_db, provider_identifier):
+    provider = get_alternative_sms_provider(provider_identifier)
+    assert provider.identifier != provider
+
+
+def test_switch_sms_provider_to_current_provider_does_not_switch(
+    notify_db,
+    notify_db_session,
```
leohemsted: don't need these `notify_db*` fixtures here (or elsewhere in this file) - they're used by `restore_provider_details`

app/celery/provider_tasks.py
```diff
@@ -43,6 +44,7 @@ def deliver_sms(self, notification_id):
         send_to_providers.send_sms_to_provider(notification)
     except Exception as e:
         try:
+            dao_toggle_sms_provider()
```
servingUpAces: I agree. But at the moment the `SMSClientResponseException` is thrown when we do not understand the response from the provider. But when we get timeout exceptions we get throw a different type of exception. Or there could be an unknown reason for the exception. So it will not work to catch `SMSClientResponseException`, without some refactoring.

app/provider_details/switch_providers.py
```diff
@@ -0,0 +1,48 @@
...
+
+def update_provider_priorities(current_provider, new_provider):
+    # Swap priority to change primary provider
+    if new_provider.priority > current_provider.priority:
+        new_provider.priority, current_provider.priority = current_provider.priority, new_provider.priority
+        _print_provider_switch_logs(current_provider, new_provider)
+        return [current_provider, new_provider]
+
+    # Incease other provider priority if equal
+    elif new_provider.priority == current_provider.priority:
```
leohemsted: do we have any tests that test this branch works?

app/dao/provider_details_dao.py
```diff
@@ -14,6 +19,48 @@ def get_provider_details_by_id(provider_details_id):
...
+def dao_switch_sms_provider_to_provider_with_identifier(identifier):
+    current_provider = get_current_provider('sms')
+    new_provider = get_provider_details_by_identifier(identifier)
+
+    if provider_is_already_primary_or_inactive(current_provider, new_provider, identifier):
+        return current_provider
+    else:
+        updated_providers = update_provider_priorities(current_provider, new_provider)
+        for provider in updated_providers:
+            db.session.add(provider)
```
imdadahad: 👍

PR #796, opened 2017-01-19
Don’t strip HTML when saving templates

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/109
- [x] https://github.com/alphagov/notifications-admin/pull/1078

***

Right now we strip HTML from templates at the point of saving them. This also converts stuff like ampersands to their entity form (eg &amp;) and this is what we save in the database.

This is a bad idea when you’re sending a text message or a letter, in which an HTML entity makes no sense. But we still need to encode HTML in the body of HTML emails.

The right place to do this is when rendering the templates. The code to do this is now in utils. So this commit:
- pull in this new utils code
- removes the old
- adds some integration tests to make sure that everything is working as expected (more thorough unit tests are happening in utils)

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -4,7 +4,6 @@ from flask import (
     request,
     current_app
 )
-import bleach
 
 from app.dao.templates_dao import (
     dao_update_template,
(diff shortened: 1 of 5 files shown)


leohemsted:
(I approve this btw)

PR #797, opened 2017-01-19
Persist letters

leohemsted:
### Refactor tasks code

I pulled out the per-row handling from process_job into a separate function for readability/testability. I also pulled the exception handling stuff from send_sms and send_email out because it's literally identical.

### Refactor tasks tests

I tried not to do too much, but moved a couple of tests around so they're in logical sections within the file, and changed a couple of mock points. I also created some new functions in `tests/db.py` to create templates and notifications - these don't require `notify_db`/`notify_db_session` so can be used anywhere cleanly - though you'll need to take care to ensure your test cleans up after itself.

### New letter stuff

Added new task `persist_letter` - I know it's not named at all related to the `send_sms` and `send_email` tasks it mimics, but those tasks aren't really fair - those tasks persist to the database, and then queue up the `provider_tasks.deliver_xxx` tasks - if renaming celery tasks wasn't such a bother i'd rename those to persist_sms and persist_email too.

The task itself is super straightforward - the only thing of note it does is set the `to` field to equal `personalisation['addressline1']`, which is a required field, and will hopefully generally be a person's name or company name.

also, I didn't include the api_key and key_type params from send_xxx tasks - they're not used cos we don't use this task for api calls

todo:

- [x] test using front end. some front end work required first but it should work, so i'm happy for people to start reviewing before i've finished admin work
- [x] https://github.com/alphagov/notifications-aws/pull/118

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -220,20 +220,24 @@ def persist_letter(
     created_at
 ):
     notification = encryption.decrypt(encrypted_notification)
+
+    # we store the recipient as just the first item of the person's address
+    recipient = notification['personalisation']['addressline1']
+
     service = dao_fetch_service_by_id(service_id)
     try:
         saved_notification = persist_notification(
             template_id=notification['template'],
             template_version=notification['template_version'],
-            recipient=notification['to'],
+            recipient=recipient,
             service=service,
-            personalisation=notification.get('personalisation'),
-            notification_type=EMAIL_TYPE,
+            personalisation=notification['personalisation'],
+            notification_type=LETTER_TYPE,
             api_key_id=None,
             key_type=KEY_TYPE_NORMAL,
             created_at=created_at,
-            job_id=notification.get('job', None),
-            job_row_number=notification.get('row_number', None),
+            job_id=notification['job'],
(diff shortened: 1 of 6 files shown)


app/celery/tasks.py
```diff
@@ -213,22 +207,69 @@ def send_email(self, service_id,
...
+        # SQLAlchemy is throwing a FlushError. So we check if the notification id already exists then do not
+        # send to the retry queue.
+        current_app.logger.exception('Retry' + retry_msg)
+        try:
+            task.retry(queue="retry", exc=exc)
+        except task.MaxRetriesExceededError:
+            current_app.logger.exception('Retry' + retry_msg)
+
+
+def get_template_class(template):
```
servingUpAces: Could just pass the template_type here rather than the whole template object.

tests/app/celery/test_tasks.py
```diff
@@ -766,19 +794,23 @@ def test_should_use_email_template_and_persist_without_personalisation(sample_em
                                                                      queue='send-email')
 
 
+class MyException(Exception):
```
servingUpAces: Is MyException used?

PR #798, opened 2017-01-19
Turns out the numbers in the research mode task were delivering.

servingUpAces:
This updates them to the dedicated "TV" numbers. Should never be real.


leohemsted:
we should check and see if anyone else has been using these numbers (@minglis performance tests?)

PR #799, opened 2017-01-20
fix simulate research mode logic

servingUpAces:
The notification.to field is being formatted to be +447...
This meant that the research mode task would never work.
Also the way we mark data as "temporary-failure" with firetext is with first a pending status callback then a declined callback.
This PR changes the research mode task to account for that situation.

### DIFF
--- a/tests/app/celery/test_research_mode_tasks.py
+++ b/tests/app/celery/test_research_mode_tasks.py
@@ -1,3 +1,4 @@
+import pytest
 from flask import json
 from app.celery.research_mode_tasks import (
     send_sms_response,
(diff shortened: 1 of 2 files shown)


tests/app/celery/test_research_mode_tasks.py
```diff
@@ -50,42 +51,49 @@ def test_make_ses_callback(notify_api, rmock):
...
     assert data['status'] == "4"
     assert data['reference'] == "mmg_reference"
     assert data['CID'] == "1234"
 
 
-def test_delivered_firetext_callback():
-    assert firetext_callback('1234', '07700900001') == {
-        'mobile': '07700900001',
+@pytest.mark.parametrize("phone_number", ["07700900001", "+447700900001", "7700900001", "+44 7700900001",
+                                          "+4407513453456"])
```
leohemsted: hope this isn't someone on the team's personal phone number?

PR #801, opened 2017-01-24
Change to research mode.

minglis:
Previously research mode created a task to fake the callback from the providers. This meant there is an extra task for each notification than would be generated in a live situation.

This PR changes that, so that a research mode/test key notification calls the callback API rather than make a task to do that .

This ensures that the flow for research mode more closely mimics that of live, and removes a task from the process so we can more accurately test throughput,

### DIFF
--- a/tests/app/delivery/test_send_to_providers.py
+++ b/tests/app/delivery/test_send_to_providers.py
@@ -165,7 +165,7 @@ def test_should_call_send_sms_response_task_if_research_mode(notify_db, sample_s
                                                              research_mode, key_type):
     mocker.patch('app.mmg_client.send_sms')
     mocker.patch('app.mmg_client.get_name', return_value="mmg")
-    mocker.patch('app.celery.research_mode_tasks.send_sms_response.apply_async')
+    mocker.patch('app.delivery.send_to_providers.send_sms_response')
 
     if research_mode:
         sample_service.research_mode = True
@@ -178,8 +178,8 @@ def test_should_call_send_sms_response_task_if_research_mode(notify_db, sample_s
         sample_notification
     )
     assert not mmg_client.send_sms.called
-    send_to_providers.send_sms_response.apply_async.assert_called_once_with(
-        ('mmg', str(sample_notification.id), sample_notification.to), queue='research-mode'
+    app.delivery.send_to_providers.send_sms_response.assert_called_once_with(
+        'mmg', str(sample_notification.id), sample_notification.to
     )
 
     persisted_notification = notifications_dao.get_notification_by_id(sample_notification.id)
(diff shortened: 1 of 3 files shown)


tests/app/delivery/test_send_to_providers.py
```diff
@@ -165,7 +165,7 @@ def test_should_call_send_sms_response_task_if_research_mode(notify_db, sample_s
                                                              research_mode, key_type):
     mocker.patch('app.mmg_client.send_sms')
     mocker.patch('app.mmg_client.get_name', return_value="mmg")
-    mocker.patch('app.celery.research_mode_tasks.send_sms_response.apply_async')
+    mocker.patch('app.delivery.send_to_providers.send_sms_response')
```
servingUpAces: Looks like you are mocking the method you are trying to test.

tests/app/delivery/test_send_to_providers.py
```diff
@@ -200,7 +200,7 @@ def test_should_set_billable_units_to_zero_in_research_mode_or_test_key(
 
     mocker.patch('app.mmg_client.send_sms')
     mocker.patch('app.mmg_client.get_name', return_value="mmg")
-    mocker.patch('app.celery.research_mode_tasks.send_sms_response.apply_async')
+    mocker.patch('app.delivery.send_to_providers.send_sms_response')
```
servingUpAces: Mocking the method under test?

tests/app/delivery/test_send_to_providers.py
```diff
@@ -222,14 +222,14 @@ def test_should_not_send_to_provider_when_status_is_not_created(notify_db, notif
                                        status='sending')
     mocker.patch('app.mmg_client.send_sms')
     mocker.patch('app.mmg_client.get_name', return_value="mmg")
-    mocker.patch('app.celery.research_mode_tasks.send_sms_response.apply_async')
+    mocker.patch('app.delivery.send_to_providers.send_sms_response')
```
servingUpAces: again?

tests/app/delivery/test_send_to_providers.py
```diff
@@ -222,14 +222,14 @@ def test_should_not_send_to_provider_when_status_is_not_created(notify_db, notif
...
 
     send_to_providers.send_sms_to_provider(
         notification
     )
 
     app.mmg_client.send_sms.assert_not_called()
-    app.celery.research_mode_tasks.send_sms_response.apply_async.assert_not_called()
+    app.delivery.send_to_providers.send_sms_response.assert_not_called()
```
servingUpAces: here too.

tests/app/delivery/test_send_to_providers.py
```diff
@@ -284,7 +284,7 @@ def test_send_email_to_provider_should_call_research_mode_task_response_task_if_
     mocker.patch('app.uuid.uuid4', return_value=reference)
     mocker.patch('app.aws_ses_client.send_email')
     mocker.patch('app.aws_ses_client.get_name', return_value="ses")
-    mocker.patch('app.celery.research_mode_tasks.send_email_response.apply_async')
+    mocker.patch('app.delivery.send_to_providers.send_email_response')
```
servingUpAces: you get the point

PR #802, opened 2017-01-25
Return success of the callback is a duplicate or for an id that does not exist

servingUpAces:
This PR changes the response to POST /notifications/sms/<mmg | firetext> from a 400 response to a 200 response.
If we get a callback for a notification more than once or for a notification we log that but we return a 200 success response to the provider.
We have found that there is a situation where the send to provider throws a timeout exception but the provider did get the message, but we still send it to them again.
In which case they send the message twice, and callback for the message twice.
Another case where we may get duplicate callbacks is that the network gave the provider two callbacks meaning they pass those two callbacks onto us.

So it is really difficult to know if we sent to the provider twice or just got two callbacks.

The test_callback has many changes because I took the opportunity to use the client conftest fixture rather than the notify_api fixture.
The only 2 tests really changed are test_mmg_callback_returns_200_when_notification_id_not_found_or_already_updated and test_firetext_callback_returns_200_when_notification_id_not_found_or_already_updated

### DIFF
--- a/tests/app/notifications/rest/test_callbacks.py
+++ b/tests/app/notifications/rest/test_callbacks.py
@@ -1,9 +1,9 @@
 import uuid
 
 from datetime import datetime
+
 from flask import json
 from freezegun import freeze_time
-from unittest.mock import call
 
 import app.celery.tasks
 from app.dao.notifications_dao import (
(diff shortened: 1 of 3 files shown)


tests/app/notifications/rest/test_callbacks.py
```diff
@@ -13,712 +13,670 @@
...
+                           headers=[('Content-Type', 'application/json')])
+    assert response.status_code == 400
+    json_data = json.loads(response.data)
+    assert json_data['result'] == 'error'
+    assert len(json_data['message']) == 2
+    assert "{} callback failed: {} missing".format('MMG', 'status') in json_data['message']
+    assert "{} callback failed: {} missing".format('MMG', 'CID') in json_data['message']
+
+
+def test_ses_callback_should_not_need_auth(client):
```
leohemsted: this test is duplicated

PR #803, opened 2017-01-26
Allow filtering of billing info by financial year

quis:
We already filter the usage-by-month query by financial year. When we show the total usage for a service, we should be able to filter this by financial year.

Then, when the two lots of data are put side by side, it all adds up.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -190,7 +190,13 @@ def remove_user_from_service(service_id, user_id):
 
 @service_blueprint.route('/<uuid:service_id>/fragment/aggregate_statistics')
 def get_service_provider_aggregate_statistics(service_id):
-    return jsonify(data=get_fragment_count(service_id))
+    year = request.args.get('year')
+    if year is not None:
+        try:
+            year = int(year)
+        except ValueError:
+            raise InvalidRequest('Year must be a number', status_code=400)
+    return jsonify(data=get_fragment_count(service_id, year=year))
 
 
 # This is placeholder get method until more thought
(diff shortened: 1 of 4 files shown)


app/service/rest.py
```diff
@@ -190,7 +190,13 @@ def remove_user_from_service(service_id, user_id):
 
 @service_blueprint.route('/<uuid:service_id>/fragment/aggregate_statistics')
 def get_service_provider_aggregate_statistics(service_id):
-    return jsonify(data=get_fragment_count(service_id))
+    year = request.args.get('year')
+    if year is not None:
+        try:
+            year = int(year)
+        except TypeError:
```
imdadahad: Think this should be `ValueError`

quis: Good point. Have changed and squashed out.

app/service/rest.py
```diff
@@ -190,7 +190,13 @@ def remove_user_from_service(service_id, user_id):
 
 @service_blueprint.route('/<uuid:service_id>/fragment/aggregate_statistics')
 def get_service_provider_aggregate_statistics(service_id):
-    return jsonify(data=get_fragment_count(service_id))
+    year = request.args.get('year')
+    if year is not None:
+        try:
+            year = int(year)
+        except ValueError:
+            raise InvalidRequest('Year must be a number', status_code=400)
```
leohemsted: should throw in a quick test for this - `test_aggregate_statistics_rejects_invalid_year_param` or something

quis: Tests added in 4e60265.

PR #804, opened 2017-01-27
Add client and job to update the performance platform daily

imdadahad:
This adds a daily task that will update the performance platform with the latest email and sms notification counts 

# **Dependencies**

- [x] [Update production configs to activate the task]

# **How it works**

The task is scheduled to run each day at 00:30AM. It will calculate the number of sent (failed, delivered) notifications from the previous day. 

E.g. When the task runs on Thur, Mar, 2017 at 00:30AM, it will retrieve counts for the day before until midnight (Wed, Mar, 2017 00:00AM to Thu, Mar, 2017 00:00AM). These will then be sent to the performance platform.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -6,7 +6,7 @@ from datetime import (
 
 from flask import current_app
 from werkzeug.datastructures import MultiDict
-from sqlalchemy import (desc, func, or_, and_, asc, extract)
+from sqlalchemy import (desc, func, or_, and_, asc)
 from sqlalchemy.orm import joinedload
 
 from app import db, create_uuid
@@ -29,9 +29,8 @@ from app.models import (
 from app.dao.dao_utils import transactional
 from app.statsd_decorators import statsd
 from app.utils import (
-    get_midnight_for_date,
-    get_midnight_for_day_before
-)
+    get_midnight_for_day_before,
+    get_london_midnight_in_utc)
 
 
 def dao_get_notification_statistics_for_service_and_day(service_id, day):
@@ -432,7 +431,7 @@ def get_total_sent_notifications_in_date_range(start_date, end_date, notificatio
 def get_total_sent_notifications_yesterday():
     today = datetime.utcnow()
     start_date = get_midnight_for_day_before(today)
-    end_date = get_midnight_for_date(today)
+    end_date = get_london_midnight_in_utc(today)
 
     return {
         "start_date": start_date,
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -22,15 +22,11 @@ from app.models import (
     NOTIFICATION_PENDING,
     NOTIFICATION_TECHNICAL_FAILURE,
     NOTIFICATION_TEMPORARY_FAILURE,
-    NOTIFICATION_STATUS_TYPES_COMPLETED,
     KEY_TYPE_NORMAL, KEY_TYPE_TEST
 )
 
 from app.dao.dao_utils import transactional
 from app.statsd_decorators import statsd
-from app.utils import (
-    get_midnight_for_day_before,
-    get_london_midnight_in_utc)
 
 
 def dao_get_notification_statistics_for_service_and_day(service_id, day):
@@ -419,27 +415,9 @@ def get_total_sent_notifications_in_date_range(start_date, end_date, notificatio
         func.count(NotificationHistory.id).label('count')
     ).filter(
         NotificationHistory.key_type != KEY_TYPE_TEST,
-        NotificationHistory.status.in_(NOTIFICATION_STATUS_TYPES_COMPLETED),
         NotificationHistory.created_at >= start_date,
         NotificationHistory.created_at <= end_date,
         NotificationHistory.notification_type == notification_type
-    ).first()
+    ).scalar()
 
-    return 0 if result is None else result.count
-
-
-def get_total_sent_notifications_yesterday():
-    today = datetime.utcnow()
-    start_date = get_midnight_for_day_before(today)
-    end_date = get_london_midnight_in_utc(today)
-
-    return {
-        "start_date": start_date,
-        "end_date": end_date,
-        "email": {
-            "count": get_total_sent_notifications_in_date_range(start_date, end_date, 'email')
-        },
-        "sms": {
-            "count": get_total_sent_notifications_in_date_range(start_date, end_date, 'sms')
-        }
-    }
+    return result or 0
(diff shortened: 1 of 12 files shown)


app/dao/notifications_dao.py
```diff
@@ -408,3 +413,34 @@ def get_april_fools(year):
...
+
+
+def get_total_sent_notifications_in_date_range(start_date, end_date, notification_type):
+    result = db.session.query(
+        func.count(NotificationHistory.id).label('count')
+    ).filter(
+        NotificationHistory.key_type != KEY_TYPE_TEST,
+        NotificationHistory.status.in_(NOTIFICATION_STATUS_TYPES_COMPLETED),
+        NotificationHistory.created_at >= start_date,
+        NotificationHistory.created_at <= end_date,
```
leohemsted: this should refer `updated_at` - or a notification that is created at 11:59 on 1st jan but not delivered until 12:31 2nd jan (ie: after the job collecting for 1st jan has finished running) will never get captured

app/dao/notifications_dao.py
```diff
@@ -408,3 +413,34 @@ def get_april_fools(year):
...
+        NotificationHistory.notification_type == notification_type
+    ).first()
```
leohemsted: http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.scalar

`.scalar()` does what you need (though you'll still need to say `return result or 0` to cover the 0 rows case)

app/dao/notifications_dao.py
```diff
@@ -408,3 +413,34 @@ def get_april_fools(year):
...
+    end_date = get_midnight_for_date(today)
+
+    return {
+        "start_date": start_date,
+        "end_date": end_date,
+        "email": {
+            "count": get_total_sent_notifications_in_date_range(start_date, end_date, 'email')
+        },
+        "sms": {
+            "count": get_total_sent_notifications_in_date_range(start_date, end_date, 'sms')
```
leohemsted: if you `group_by(NotificationHistory.notification_type)` then you won't need to do this twice - two separate queries with query overheads and compute times and all that - is that a concern?

leohemsted: actually it might not be they're not hugely long running queries (yesterday's data set of ~39k messages on live takes 42 + 24ms, whereas with a groupby it takes 45ms )

since it only saves 20ms on a scheduled task that runs at unsavoury hours i'm not bothered, but its just something to keep in mind - grouping is good

app/dao/notifications_dao.py
```diff
@@ -408,3 +413,34 @@ def get_april_fools(year):
...
+        NotificationHistory.status.in_(NOTIFICATION_STATUS_TYPES_COMPLETED),
+        NotificationHistory.created_at >= start_date,
+        NotificationHistory.created_at <= end_date,
+        NotificationHistory.notification_type == notification_type
+    ).first()
+
+    return 0 if result is None else result.count
+
+
+def get_total_sent_notifications_yesterday():
```
leohemsted: i think this function should live in `scheduled_tasks.py` rather than the dao, as it deals with date time manipulation and marshalling data into the required json format

app/utils.py
```diff
@@ -26,3 +28,13 @@ def get_template_instance(template, values):
     return {
         SMS_TYPE: SMSMessageTemplate, EMAIL_TYPE: PlainTextEmailTemplate
     }[template['template_type']](template, values)
+
+
+def get_midnight_for_date(date):
```
leohemsted: can you rename this to `get_utc_midnight_for_date` (and the same with the function below) just so we're clear that it needs adjusting if you care about what day it is (pete ruled that it doesn't matter for this for now but a rename would be good 👍 )

app/dao/notifications_dao.py
```diff
@@ -408,3 +413,34 @@ def get_april_fools(year):
...
     return pytz.timezone('Europe/London').localize(datetime(year, 4, 1, 0, 0, 0)).astimezone(pytz.UTC).replace(
         tzinfo=None)
+
+
+def get_total_sent_notifications_in_date_range(start_date, end_date, notification_type):
+    result = db.session.query(
+        func.count(NotificationHistory.id).label('count')
+    ).filter(
+        NotificationHistory.key_type != KEY_TYPE_TEST,
+        NotificationHistory.status.in_(NOTIFICATION_STATUS_TYPES_COMPLETED),
```
servingUpAces: We probably want all messages here, otherwise will never count those message in sending for a long time (because the phone is off or what have you).

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -150,10 +175,42 @@ def test_will_remove_csv_files_for_jobs_older_than_seven_days(notify_db, notify_
...
     with freeze_time('2016-10-17T00:00:00'):
         remove_csv_files()
     s3.remove_job_from_s3.assert_called_once_with(job_1.service_id, job_1.id)
+
+
+@freeze_time("2016-01-11 12:30:00")
+def test_send_daily_performance_stats_calls_with_correct_totals(notify_db, notify_db_session, sample_template, mocker):
+    perf_mock = mocker.patch('app.celery.scheduled_tasks.performance_platform_client.send_performance_stats')
+
+    notification_history = partial(
```
leohemsted: nice use of higher order functions 👌 👌 👌

app/dao/notifications_dao.py
leohemsted: just discussed with @yahoopete and @servingUpAces - we think that running by `created_at` makes more sense - so to correct this if we remove the `status in NOTIFICATION_STATUS_TYPES_COMPLETED` bit it'll be accurate

tests/app/dao/test_notification_dao.py
```diff
@@ -1306,3 +1320,180 @@ def test_get_april_fools():
...
+        notification_history(notification_type='email')
+        notification_history(notification_type='sms')
+        notification_history(notification_type='sms')
+        notification_history(notification_type='sms')
+
+    total_count = get_total_sent_notifications_in_date_range(start_date, end_date, 'sms')
+    assert total_count == 3
+
+
+def test_get_total_sent_notifications_for_sms_excludes_sms_counts(
```
leohemsted: should read `test_get_total_sent_notifications_for_email_excludes_sms_counts`

app/dao/notifications_dao.py
servingUpAces: 👌

imdadahad:
Updated based on comments 👍

PR #807, opened 2017-01-30
Add endpoint for breakdown of activity by month

quis:
This endpoint will eventually replace the weekly breakdown one. By month for a given financial year is better, because it gives us consistency with the breakdown of financial usage (and eventually consistency with the template usage).

The code to do this is a bit convoluted, in order to fill out the counts for months and statuses where we don’t have notifications.

This will make the admin side of this easier, because we can rely on there always being numbers available. The admin side will deal with summing the statuses (eg `temporary-failure` > `failed`) because this is presentational.

This commit also modifies the usage count to use `.between()` for consistency.

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -9,6 +9,7 @@ from app.dao.dao_utils import (
     transactional,
     version_class
 )
+from app.dao.notifications_dao import get_financial_year
 from app.models import (
     NotificationStatistics,
     TemplateStatistics,
@@ -24,7 +25,10 @@ from app.models import (
     User,
     InvitedUser,
     Service,
-    KEY_TYPE_TEST)
+    KEY_TYPE_TEST,
+    NOTIFICATION_STATUS_TYPES,
+    TEMPLATE_TYPES,
+)
 from app.statsd_decorators import statsd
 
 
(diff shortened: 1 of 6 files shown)


app/dao/services_dao.py
```diff
@@ -238,6 +242,56 @@ def dao_fetch_weekly_historical_stats_for_service(service_id):
     ).all()
 
 
+@statsd(namespace="dao")
+def dao_fetch_monthly_historical_stats_for_service(service_id, year):
+    monday_of_notification_week = func.date_trunc('week', NotificationHistory.created_at).label('week_start')
+    start, end = get_financial_year(year)
+
+    month = func.date_trunc(
```
servingUpAces: Can we pull this month into a method since it's used more than once?

tests/app/dao/test_services_dao.py
```diff
@@ -506,6 +507,50 @@ def test_fetch_weekly_historical_stats_separates_weeks(notify_db, notify_db_sess
...
+def test_fetch_monthly_historical_stats_separates_weeks(notify_db, notify_db_session, sample_template):
+    notification_history = functools.partial(
+        create_notification_history,
+        notify_db,
+        notify_db_session,
+        sample_template
+    )
+    _before_start_of_financial_year = notification_history(created_at=datetime(2016, 3, 31))
+    start_of_financial_year = notification_history(created_at=datetime(2016, 4, 1))
+    start_of_summer = notification_history(created_at=datetime(2016, 6, 20))
```
servingUpAces: we should test that sept 30, 23:00 counts in October. And March 31 11:30 resolves to April 1

quis: Tests added in fe267f3

app/dao/services_dao.py
quis: Split it out in 3153950

PR #808, opened 2017-01-30
Suspend/resume service feature

servingUpAces:
This is the first PR in the story to create a suspend service feature.
- Renaming /service/<id>/deactivate to /service/<id>/archive to match language on the UI.
  - Will need to update admin before deleting the deactivate service method
- Created dao and endpoint methods to suspend and resume a service.
- I confirm the use of suspend and resume with a couple people on the team, seems to be the right choice.

The idea is that if you archive a service there is no coming back from that.
To suspend a service is marking it as inactive and revoking the api keys. To resume a service is to mark the service as active, the service will need to create new API keys.
It makes sense that if a service is under threat that the API keys should be renewed.

The next PR will update the code to check that the service is active before sending the request or allowing any actions by the service.

https://www.pivotaltracker.com/story/show/138598461

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -27,9 +27,10 @@ from app.dao.services_dao import (
     dao_fetch_todays_stats_for_service,
     dao_fetch_weekly_historical_stats_for_service,
     dao_fetch_todays_stats_for_all_services,
-    dao_deactive_service,
-    fetch_stats_by_date_range_for_all_services
-)
+    dao_archive_service,
+    fetch_stats_by_date_range_for_all_services,
+    dao_suspend_service,
+    dao_resume_service)
 from app.dao.service_whitelist_dao import (
     dao_fetch_service_whitelist,
     dao_add_and_commit_whitelisted_contacts,
@@ -335,6 +336,7 @@ def update_whitelist(service_id):
         return '', 204
 
 
+# Renaming this endpoint to archive
 @service_blueprint.route('/<uuid:service_id>/deactivate', methods=['POST'])
 def deactivate_service(service_id):
     service = dao_fetch_service_by_id(service_id)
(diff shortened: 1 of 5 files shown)


tests/app/service/test_suspend_resume_service.py
```diff
@@ -0,0 +1,72 @@
...
+                           headers=[auth_header])
+    assert response.status_code == 204
+    assert sample_api_key.expiry_date.date() == datetime.utcnow().date()
+
+
+def test_resume_service_leaves_api_keys_revokes(client, sample_service, sample_api_key):
+    sample_api_key.expiry_date = datetime.utcnow()
+    sample_service.active = False
+    auth_header = create_authorization_header()
+    response = client.post("/service/{}/archive".format(sample_service.id),
```
leohemsted: this should be `/service/{}/resume`

tests/app/service/test_suspend_resume_service.py
```diff
@@ -0,0 +1,72 @@
...
+    response = client.post("/service/{}/{}".format(sample_service.id, action),
+                           headers=[auth_header])
+    assert response.status_code == 204
+    mocked.assert_not_called()
+    assert mocked.called == 0
+
+
+def test_suspending_service_revokes_api_keys(client, sample_service, sample_api_key):
+    auth_header = create_authorization_header()
+    response = client.post("/service/{}/archive".format(sample_service.id),
```
leohemsted: `/service/{}/suspend`

tests/app/service/test_suspend_resume_service.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.mark.parametrize("action, active", [("suspend", False), ("resume", True)])
+def test_has_not_effect_when_service_is_already_that_state(client, sample_service, action, active, mocker):
+    mocked = mocker.patch("app.service.rest.dao_{}_service".format(action))
+    sample_service.active = active
+    auth_header = create_authorization_header()
+    response = client.post("/service/{}/{}".format(sample_service.id, action),
+                           headers=[auth_header])
+    assert response.status_code == 204
+    mocked.assert_not_called()
+    assert mocked.called == 0
```
leohemsted: also is it worth checking `assert sample_service.active == active` to check we havent accidentally changed status? not sure but maybe /shrug

tests/app/service/test_suspend_resume_service.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def test_suspending_service_revokes_api_keys(client, sample_service, sample_api_key):
+    auth_header = create_authorization_header()
+    response = client.post("/service/{}/archive".format(sample_service.id),
+                           headers=[auth_header])
+    assert response.status_code == 204
+    assert sample_api_key.expiry_date.date() == datetime.utcnow().date()
```
leohemsted: same `freeze_time` comment here rather than asserting the date is the same 

https://github.com/alphagov/notifications-api/pull/808/files#r98639225

app/service/rest.py
```diff
@@ -343,7 +345,63 @@ def deactivate_service(service_id):
...
+    When a service is archived the service is made inactive, templates are archived and api keys are revoked.
+    There is no coming back from this operation.
+    :param service_id:
+    :return:
+    """
+    service = dao_fetch_service_by_id(service_id)
+
+    if not service.active:
+        # assume already inactive, don't change service name
+        return '', 204
```
leohemsted: ignoring the fact that i probably wrote this code originally <_<

tests/app/service/test_suspend_resume_service.py
```diff
@@ -33,25 +35,31 @@ def test_has_not_effect_when_service_is_already_that_state(client, sample_servic
...
-    auth_header = create_authorization_header()
-    response = client.post("/service/{}/archive".format(sample_service.id),
-                           headers=[auth_header])
-    assert response.status_code == 204
-    assert sample_api_key.expiry_date.date() == datetime.utcnow().date()
+    with freeze_time('2001-10-22T11:59:00'):
+        sample_api_key.expiry_date = datetime.utcnow()
+        sample_service.active = False
+        auth_header = create_authorization_header()
+        client.post("/service/{}/resume".format(sample_service.id),
```
leohemsted: dont think you need this post here?

PR #811, opened 2017-01-31
Check service is active

servingUpAces:
- Create job now checks that the service is active before creating job.
- If the service is inactive at the time of sending the notification to the provider, the notification is marked as technical-failure.
- removed deactivate service endpoint, renamed to archive


minglis:
👍

PR #816, opened 2017-02-03
Small refactor to reuse code.

servingUpAces:


quis:
I don’t like the way that `service_is_inactive` – which looks like a simple boolean check – has side effects.

I think it would be better to keep the check in the main code blocks, and make the helper function more specific, eg:
```python
if not service.active:
    technically_fail(notification)
    return
```

PR #819, opened 2017-02-07
Fix: Update history when auto switching provider

imdadahad:
This is a small fix that ensures the provider history table is updated when switching providers.

Originally the priorities were being updated only.

### DIFF
--- a/tests/app/dao/test_provider_details_dao.py
+++ b/tests/app/dao/test_provider_details_dao.py
@@ -122,8 +122,7 @@ def test_get_alternative_sms_provider_returns_expected_provider(notify_db, provi
 
 def test_switch_sms_provider_to_current_provider_does_not_switch(
     restore_provider_details,
-    current_sms_provider,
-    mocker
+    current_sms_provider
 ):
     dao_switch_sms_provider_to_provider_with_identifier(current_sms_provider.identifier)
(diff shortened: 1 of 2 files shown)


imdadahad:
I can..although there are a couple of tests that check the `dao_update_provider_details` update the history. Specifically `test_provider_details_dao.py::test_update_adds_history`.

imdadahad:
I've added a test to check the version has been updated `test_provider_details_dao.py::test_toggle_sms_provider_updates_version`. If the version is updated, it's safe to say that the history has been written. The method that writes to the audit table is also separately tested in `test_update_adds_history`.

tests/app/dao/test_provider_details_dao.py
```diff
@@ -172,3 +168,14 @@ def test_toggle_sms_provider_switches_when_provider_priorities_are_equal(
...
+
+def test_toggle_sms_provider_updates_version(
+    restore_provider_details,
+    current_sms_provider
+):
+    version_before = current_sms_provider.version
+    dao_toggle_sms_provider(current_sms_provider.identifier)
+    version_after = current_sms_provider.version
+
+    assert version_after - version_before == 1
```
servingUpAces: Can we assert the existence of the ProviderDetailsHistory data?

PR #822, opened 2017-02-08
Add a stop celery script into the scripts folder.

minglis:
Stopping celery via upstart issues a TERM signal to the whole process group, which immediately kills the worker threads. This can lead to inconsistentcy in the state of a notification as an API call, for example, can take several seconds consequently the likelihood of the call being lost due to a restart is relatively high.

We need to ensure that we stop cleanly to ensure code deploys and autoscale events do not cause issues.

this script is called as part of a pre-stop hook in upstart, which passes a pid file location to the script. The script will then issues the TERM signal to only the master celery process. which will cleanly shut down the workers and exit.

the script will loop based on the existence of a /proc/{pid} for the celery master process. Once the file is gone the script will exit. Upstart blocks on this script before issuing kill on whatever processes may be left hanging around.

the length of the loop is a maximum 15 minutes - which is the tested max time a task may take. In reality it will exit much quicker.

### DIFF
--- a/scripts/stop_celery.py
+++ b/scripts/stop_celery.py
@@ -57,9 +57,7 @@ def pid_still_running(pid):
 
     Once the process stops this file no longer exists. Slim possibilty of a race condition here.
     """
-    if os.path.exists("/proc/" + pid):
-        return True
-    return False
+    return os.path.exists("/proc/" + pid)
 
 
 if __name__ == "__main__":
@@ -78,6 +76,6 @@ if __name__ == "__main__":
     """
     iteration = 0
     while pid_still_running(celery_pid) and iteration < 180:
-        print("waited for ", iteration * 5, " secs")
+        print("[", celery_pid_file, "] waited for ", iteration * 5, " secs")
         sleep(5)
         iteration += 1
--- a/scripts/stop_celery.py
+++ b/scripts/stop_celery.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+"""
+
+Scipt used to stop celery in AWS environments. This is used from upstart to issue a TERM signal to the master celery process.
+
+This will then allow the worker threads to stop, after completing whatever tasks that are in flight.
+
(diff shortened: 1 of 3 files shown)


scripts/stop_celery.py
```diff
@@ -0,0 +1,76 @@
...
+
+    """
+    Blocking loop to check for the still running process.
+    5 seconds between loops
+    180 loops
+    Maximum block time of 900 seconds (15 minutes)
+    """
+    iteration = 0
+    while pid_still_running(celery_pid) and iteration < 180:
+        print("waited for ", iteration * 5, " secs")
```
leohemsted: think we should say what proc we're waiting for here - won't a whole bunch of these be interleaved for different celery process?

scripts/stop_celery.py
```diff
@@ -0,0 +1,76 @@
...
+    for line in result.stderr.readlines():
+        print(line.rstrip())
+
+def pid_still_running(pid):
+    """
+    uses the proc filesystem to identify if the celery master pid is still around.
+
+    Once the process stops this file no longer exists. Slim possibilty of a race condition here.
+    """
+    if os.path.exists("/proc/" + pid):
```
leohemsted: `return os.path.exists('/proc/' + pid)`

scripts/stop_celery.py
```diff
@@ -0,0 +1,76 @@
...
+
+def issue_term_signal_to_pid(pid):
+    """
+    Issues a TERM signal (15) to the master celery process.
+
+    This method attempts to print out any response from this subprocess call. However this call is generally silent.
+    """
+    result = subprocess.Popen(['kill', '-15', pid], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    for line in result.stdout.readlines():
+        print(line.rstrip())
```
leohemsted: think we should print what we just tried to kill if it gives output

PR #824, opened 2017-02-13
Auto-switch providers on slow delivery of notifications

imdadahad:
This adds a celery task (running every minute) to check for slow delivery of notifications and switch if so.

## How it works

Currently we switch providers automatically when an exception is raised during delivery (either we could not understand the response or they raised an exception). We also want to switch if they are taking a 'long' time to deliver notifications. Based on our functional tests this is deemed to be four minutes. 

The `switch_current_sms_provider_on_slow_delivery` task will run every minute. It will check for any `slow delivery' notifications that occurred in the last ten minutes for the current provider. If more than two occurred, we will switch to the other provider. 

### Avoiding a provider switching loop

Since this will poll in the last ten minutes for any slow delivery notifications, we want to avoid the case where we switching back and forth between providers. Looking in the last ten minutes for slow delivery notifications is not enough, rather we need to look after the point at which we switched to current provider. E.g. If we switched to the current provider on Thursday at 5pm, we look from this point onwards and in the last ten minutes: `thursday 5pm -------> [ten minutes ago ------ current time]`. In this window, we check for slow deliveries.

## How to test

1. Set `FUNCTIONAL_TEST_PROVIDER_SERVICE_ID` and `FUNCTIONAL_TEST_PROVIDER_SMS_TEMPLATE_ID` to the respective ids of your dev service and a sample sms template.

2. Send some notifications with that template and manually override the `updated_at` in the database. This should be set to a date and time more than four minutes after the `sent_at`.

3. Run the celery beat task.

4. Once the above is set, monitor the celery logs to see the task run and the provider switch. It is set to run every minute so you should see it shortly after.  

_Note: This is set to run on production only and is driven by our functional test provider service_


tests/app/conftest.py
```diff
@@ -421,29 +430,34 @@ def sample_notification_with_job(
...
+    reference=None,
+    created_at=None,
+    sent_at=None,
+    billable_units=1,
+    personalisation=None,
+    api_key_id=None,
+    key_type=KEY_TYPE_NORMAL,
+    sent_by=None,
+    client_reference=None,
+    updated_at=None
```
leohemsted: rather than change this function, can you use `create_notification` in `api/tests/db.py`? that way we don't need to faff about with notify_db/notify_db_session in tests

app/celery/scheduled_tasks.py
```diff
@@ -141,3 +149,36 @@ def send_daily_performance_platform_stats():
...
+    functional_test_provider_service_id = current_app.config.get('FUNCTIONAL_TEST_PROVIDER_SERVICE_ID')
+    functional_test_provider_template_id = current_app.config.get('FUNCTIONAL_TEST_PROVIDER_SMS_TEMPLATE_ID')
+
+    if functional_test_provider_service_id and functional_test_provider_template_id:
+        current_provider = get_current_provider('sms')
+        slow_delivery_notifications = get_count_of_slow_delivery_sms_notifications_for_provider(
+            provider=current_provider.identifier,
+            threshold=2,
+            created_at=current_provider.updated_at,
+            sent_at=datetime.utcnow() - timedelta(minutes=10),
```
leohemsted: think we should be `sent_at` rather than `created_at` when filtering with regards to provider - the provider isn't actually referenced in the notification creation process until we send the thing

so i think we should ditch `created_at` from the args, and then have

```python
sent_at=max(datetime.utcnow() - timedelta(minutes=10), current_provider.updated_at)
```

or something like that?

tests/app/conftest.py
```diff
@@ -138,46 +141,52 @@ def sample_service(notify_db,
...
-        'process_type': process_type
-    }
-    if template_type in ['email', 'letter']:
-        data.update({
-            'subject': subject_line
-        })
-    template = Template(**data)
-    dao_create_template(template)
+
+    template = Template.query.filter_by(id=template_id).first()
```
leohemsted: i'm not really a huge fan of this (and in the same vein the changes to `sample_service`) - there are two options that i think i'd prefer:

1. move this check into the `_create_slow_delivery_notification` fn - so that fn now says
```python
template = Template.query.get(template_id)
if not template:
    create_template(...)
```
2. change the config in the tests to set `FUNCTIONAL_TEST_PROVIDER_SMS_TEMPLATE_ID` to `sample_template.id` or similar

no huge preference, think the second one is nicer but not sure how the code'll look. thoughts?

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -13,19 +15,63 @@
...
 )
+from tests.conftest import set_config_values
 from unittest.mock import call, patch, PropertyMock
 
 
+def _create_slow_delivery_notification(notify_db, notify_db_session, provider='mmg'):
+    now = datetime.utcnow()
+    five_minutes_from_now = now + timedelta(minutes=5)
+
+    service = create_sample_service(
```
leohemsted: if you use the `tests/db.py` functions you won't need `notify_db` or `notify_db_session` in these function calls :)

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -244,3 +292,124 @@ def test_send_daily_performance_stats_calls_with_correct_totals(
...
+    }):
+        switch_current_sms_provider_on_slow_delivery()
+
+    assert get_notifications_mock.called is False
+    assert toggle_sms_mock.called is False
+
+
+def test_switch_providers_on_slow_delivery_runs_if_config_set(
+    notify_db,
+    notify_db_session,
```
leohemsted: dont need the notify_db fixtures here

tests/app/conftest.py
imdadahad: Any particular reason you're not a fan of this?

tests/app/conftest.py
imdadahad: Forgot that existed 👍

tests/app/conftest.py
leohemsted: * the use case of "create a template for a test with a specific ID" is limited, as far as I can think, to specifically tests involving either this or the notify templates (eg 2FA, invite email, reset password email) - and the notify 2FA has its own fixture at sms_code_template
* these functions are already immensely complex, i've been trying to cut them down with my `db.py` stuff anyway
* given that we're calling sample_notification with an ID from only one place, and that's specifically where the test is, I don't think we lose anything by my suggestion #1 of grabbing the template in `_create_slow_delivery_notification`, but we gain keeping slightly complex code limited to only be called in the place where it's actually needed

app/dao/notifications_dao.py
```diff
@@ -421,3 +424,32 @@ def get_total_sent_notifications_in_date_range(start_date, end_date, notificatio
     ).scalar()
 
     return result or 0
+
+
+def get_count_of_slow_delivery_sms_notifications_for_provider(
```
imdadahad: So I wanted the function to not only tell me whether there was slow delivery, but also how many culprit notifications there were. Although, I do think a bool func is cleaner along with the separation. Will update.

app/dao/notifications_dao.py
```diff
@@ -421,3 +424,32 @@ def get_total_sent_notifications_in_date_range(start_date, end_date, notificatio
...
+    ).filter(
+        Notification.service_id == service_id,
+        Notification.template_id == template_id,
+        Notification.created_at >= created_at,
+        Notification.sent_at >= sent_at,
+        Notification.status == NOTIFICATION_DELIVERED,
+        Notification.sent_by == provider,
+        (Notification.updated_at - Notification.sent_at) >= delivery_time,
+    ).group_by(
+        Notification.sent_by
```
imdadahad: Yeah no need for grouping anymore based on other changes.

app/celery/scheduled_tasks.py
imdadahad: <-----10 minutes ago-----provider updated (7 mins ago)------now-->

This works well as will also ensure that we only check in a marginal window (from provider updated to now which is 3 mins) if we updated the provider only a couple of minutes ago.

tests/app/db.py
```diff
@@ -72,6 +74,13 @@ def create_notification(
 ):
     if created_at is None:
         created_at = datetime.utcnow()
+
+    if sent_at is None:
+        sent_at = datetime.utcnow()
+
+    if updated_at is None:
+        updated_at = datetime.utcnow()
```
leohemsted: maybe we should instead say here
```python

if status != 'created':
  sent_at = sent_at or datetime.utcnow()
  updated_at = updated_at or datetime.utcnow()
```

i thinkk that covers everything?

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -13,19 +15,56 @@
...
+    create_notification(
+        template=template,
+        status='delivered',
+        sent_by=provider,
+        updated_at=five_minutes_from_now
+    )
+
+
+@pytest.fixture(scope='function')
+def set_provider_updated_at(current_sms_provider):
```
leohemsted: should this fixture depend on the `restore_provider_details` fixture?

tests/app/dao/test_notification_dao.py
```diff
@@ -1422,3 +1425,131 @@ def test_get_total_sent_notifications_for_email_excludes_sms_counts(
...
+    notification_five_minutes_to_deliver = partial(
+        create_notification,
+        template=sample_template,
+        sent_at=now,
+        sent_by='firetext',
+        created_at=now,
+        updated_at=five_minutes_from_now
+    )
+
+    notification_five_minutes_to_deliver(status='created')
```
leohemsted: need to be careful here - this notification would be created with a sent_at, sent_by, updated_at, but still status = `'created'`

tests/app/dao/test_notification_dao.py
```diff
@@ -1422,3 +1425,131 @@ def test_get_total_sent_notifications_for_email_excludes_sms_counts(
...
+        delivery_time=timedelta(minutes=5),
+        service_id=sample_template.service.id,
+        template_id=sample_template.id
+    )
+
+    assert not slow_delivery
+
+
+@freeze_time("2016-01-10 12:00:00.000000")
+def test_slow_provider_delivery_returns_for_delivered_notifications_only(
```
leohemsted: have we thought about temp-failures/permanent failures?

leohemsted: i think we probably don't care about that but lets double check on monday

tests/app/db.py
imdadahad: Fair point. That sounds good 👍

tests/app/dao/test_notification_dao.py
imdadahad: As discussed, something not important at this moment in time. We will probably look at this later as we iterate over this provider switching capability.

tests/app/dao/test_notification_dao.py
imdadahad: Updated `db.py` so we won't have this issue anymore.

tests/app/dao/test_provider_details_dao.py
```diff
@@ -19,7 +19,7 @@
 )
 
 
-def set_primary_sms_provider(identifier='mmg'):
+def set_primary_sms_provider(restore_provider_details, identifier='mmg'):
```
leohemsted: dont think this is meant to be here - this isn't a fixture, and so passing this in isn't doing what you think it should do

also i don't think identifier should have a default value - we don't call this function without specifying the provider

tests/app/dao/test_notification_dao.py
leohemsted: no - still will, because the partial is passing in a specific sent_at and sent_by

imdadahad: Removing the created notification as this test doesn't require it to achieve its goal.

PR #825, opened 2017-02-13
Remove app and py3 packages on codedeploy-ment

imdadahad:
This will remove the app and installed python3 packages on the instance before run the deployment scripts.

## How it works

Currently we recycle our instances when we auto-scale. This means that our application files and python packages are installed atop of the existing files during deployment. This will not however take care of removing files that we deleted in the app or packages we removed. 

This taps into the AWS `BeforeInstall` hook which will run specified commands before the app files are copied. At this stage it makes sense to remove all our existing application files `/home/notify-app/notifications-api` and also remove all installed python3 packages. 

Once these actions are complete, we will run the standard deployment steps to copy over the new application files and install requirements according to `requirements.txt`.

### DIFF
--- a/scripts/aws_clear_instance.sh
+++ b/scripts/aws_clear_instance.sh
@@ -5,6 +5,6 @@ echo "Removing application and dependencies"
 if [ -d "/home/notify-app/notifications-api" ]; then
     # Remove and re-create the directory
     rm -rf /home/notify-app/notifications-api
-    mkdir -vp /home/notify-app/notifications-api
+    mkdir -p /home/notify-app/notifications-api
 fi
 
--- a/scripts/aws_clear_instance.sh
+++ b/scripts/aws_clear_instance.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+echo "Removing application and dependencies"
+
+if [ -d "/home/notify-app/notifications-api" ]; then
+    # Remove and re-create the directory
+    rm -rf /home/notify-app/notifications-api
+    mkdir -vp /home/notify-app/notifications-api
+    # Remove installed py3 packages
+    pip3 freeze | xargs pip3 uninstall -y
+else
+    echo "Directory does not exist, something went wrong!"
+fi
+
--- a/scripts/aws_clear_instance.sh
+++ b/scripts/aws_clear_instance.sh
@@ -6,9 +6,5 @@ if [ -d "/home/notify-app/notifications-api" ]; then
     # Remove and re-create the directory
     rm -rf /home/notify-app/notifications-api
     mkdir -vp /home/notify-app/notifications-api
-    # Remove installed py3 packages
-    pip3 freeze | xargs pip3 uninstall -y
-else
-    echo "Directory does not exist, something went wrong!"
 fi
 
--- a/appspec.yml
+++ b/appspec.yml
@@ -5,6 +5,10 @@ files:
   - destination: /home/notify-app/notifications-api
     source: /
 hooks:
+  BeforeInstall:
+    - location: scripts/aws_clear_instance.sh
+      runas: root
+      timeout: 1000
   AfterInstall:
     - location: scripts/aws_install_dependencies.sh
       runas: root


scripts/aws_clear_instance.sh
```diff
@@ -0,0 +1,14 @@
...
+echo "Removing application and dependencies"
+
+if [ -d "/home/notify-app/notifications-api" ]; then
+    # Remove and re-create the directory
+    rm -rf /home/notify-app/notifications-api
+    mkdir -vp /home/notify-app/notifications-api
+    # Remove installed py3 packages
+    pip3 freeze | xargs pip3 uninstall -y
+else
+    echo "Directory does not exist, something went wrong!"
```
leohemsted: should we exit 1 in this case? will that stop deploy? do we want to stop deploy?

etc

cos this echo will never ever be seen

imdadahad: Good point, based on quick googling looks like codedeploy with fail with any non-zero exit code. I'll add it in.

imdadahad: Looks like we're using that logic too in `scripts::register_with_elb.sh` on line 23 where we have `error_exit "Unable to get this instance's ID; cannot continue."`.

imdadahad: Although in saying all of this I don't think we should exit here (cause codedeploy to fail). 

If the directory doesn't exist, it could be a new instance. We are using auto-scaling so we do recycle instances but let's say we have a new non-recycled instance, it wouldn't make sense to fail as the directory would never exist. 

I think having the else helps for debugging, in the case where we have a recycled instance and we expect the folder to exist but instead this is printed out.

Feel it should stay as is based on the above.

leohemsted: yeah no that's totally reasonable

scripts/aws_clear_instance.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+echo "Removing application and dependencies"
+
+if [ -d "/home/notify-app/notifications-api" ]; then
+    # Remove and re-create the directory
+    rm -rf /home/notify-app/notifications-api
+    mkdir -vp /home/notify-app/notifications-api
```
bandesz: the -v option here seems unnecessary

scripts/aws_clear_instance.sh
```diff
@@ -0,0 +1,14 @@
...
+    # Remove installed py3 packages
+    pip3 freeze | xargs pip3 uninstall -y
```
bandesz: Why do we need this? Will this remove other global packages as well which we install when we build the AMI? Because then nothing will reinstall those.

scripts/aws_clear_instance.sh
bandesz: this whole else part is unnecessary

scripts/aws_clear_instance.sh
leohemsted: that's a good point - setuptools, wheel, etc are all installed automatically right

bandesz: Maybe pip uninstall -r requirements.txt is better? Or should we switch to using a virtualenv?

leohemsted: `pip uninstall -r requirements.txt` won't uninstall, for example, `werkzeug`, a dependency of flask. if flask upgrades its werkzeug version we'll have the old wheel kicking about, so same problem

think virtualenv is a goer tho

bandesz: I wouldn't worry about leaving some dependencies on the node, we plan to recycle them every week eventually.

bandesz: Or will this cause version issues?

bandesz: Also the wheels are shipped with the app so all of them will be deleted when we clear the app directory. Or pip still installs them to the system and only uses the wheelhouse directory as package source?

leohemsted: @bandesz it will cause version issues

if i've got a wheelhouse with requests == 2.3 and requests == 2.4 and say

```
(foo) [leohemsted:~]$ pip install --use-wheel --no-index --find-links=wheelhouse wheelhouse/*
Double requirement given: requests==2.4.0 from file:///Users/leohemsted/wheelhouse/requests-2.4.0-py2.py3-none-any.whl (already in requests==2.3.0 from file:///Users/leohemsted/wheelhouse/requests-2.3.0-py2.py3-none-any.whl, name='requests')
```

leohemsted: disregard that

leohemsted: yeah i *think* we can just remove this `pip uninstall` stuff, and just have the `rm -rf notify directory` command to clear out existing wheels

imdadahad: yeah, having old requirements for a week won't hamper the next story (installing completely from offline wheels). 

Although I'll need to do more testing in that story to make sure all the reqs are installed properly with correct versions.

PR #827, opened 2017-02-14
Set the expiry time on a verify code (2fa) to 10 minutes.

servingUpAces:
When the verify code is wrong or expired increment the failed to login count for the user.
When the verify code is successfully used reset the failed login count to 0.
Added a limit to the number of active verify codes.
I removed the tests for the email verify code, we do not use them anymore. 

Merge https://github.com/alphagov/notifications-admin/pull/1134 after this one.

https://www.pivotaltracker.com/story/show/139184309

### DIFF
--- a/app/dao/users_dao.py
+++ b/app/dao/users_dao.py
@@ -1,8 +1,6 @@
 import random
 from datetime import (datetime, timedelta)
-
 from sqlalchemy import func
-
 from app import db
 from app.models import (User, VerifyCode)
 
@@ -48,7 +46,7 @@ def get_user_code(user, code, code_type):
     # time searching for the correct code.
     codes = VerifyCode.query.filter_by(
         user=user, code_type=code_type).order_by(
-            VerifyCode.created_at.desc())
+        VerifyCode.created_at.desc())
     retval = None
     for x in codes:
         if x.check_code(code):
@@ -86,7 +84,8 @@ def count_user_verify_codes(user):
     query = db.session.query(
         func.count().label('count')
     ).filter(VerifyCode.user == user,
-             VerifyCode.expiry_datetime <= datetime.utcnow()).one()
+             VerifyCode.expiry_datetime > datetime.utcnow(),
+             VerifyCode.code_used.is_(False)).one()
     return query.count
 
 
--- a/app/dao/users_dao.py
+++ b/app/dao/users_dao.py
@@ -82,6 +82,14 @@ def delete_user_verify_codes(user):
     db.session.commit()
 
 
+def count_user_verify_codes(user):
+    query = db.session.query(
+        func.count().label('count')
+    ).filter(VerifyCode.user == user,
+             VerifyCode.expiry_datetime <= datetime.utcnow()).one()
+    return query.count
+
+
 def get_user_by_id(user_id=None):
     if user_id:
         return User.query.filter_by(id=user_id).one()
--- a/app/dao/users_dao.py
+++ b/app/dao/users_dao.py
@@ -35,7 +35,7 @@ def save_model_user(usr, update_dict={}, pwd=None):
 
 def create_user_code(user, code, code_type):
     verify_code = VerifyCode(code_type=code_type,
-                             expiry_datetime=datetime.utcnow() + timedelta(hours=1),
+                             expiry_datetime=datetime.utcnow() + timedelta(minutes=30),
                              user=user)
     verify_code.code = code
     db.session.add(verify_code)
--- a/app/dao/users_dao.py
+++ b/app/dao/users_dao.py
@@ -81,12 +81,12 @@ def delete_user_verify_codes(user):
 
 
 def count_user_verify_codes(user):
-    query = db.session.query(
-        func.count().label('count')
-    ).filter(VerifyCode.user == user,
-             VerifyCode.expiry_datetime > datetime.utcnow(),
-             VerifyCode.code_used.is_(False)).one()
-    return query.count
+    query = VerifyCode.query.filter(
+        VerifyCode.user == user,
+        VerifyCode.expiry_datetime > datetime.utcnow(),
+        VerifyCode.code_used.is_(False)
+    )
+    return query.count()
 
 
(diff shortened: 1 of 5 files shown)


tests/app/user/test_rest_verify.py
```diff
@@ -293,22 +261,17 @@ def test_send_user_code_for_sms_with_optional_to_field(notify_api,
...
-            auth_header = create_authorization_header()
-            resp = client.post(
-                url_for('user.send_user_sms_code', user_id=uuid_),
-                data=data,
-                headers=[('Content-Type', 'application/json'), auth_header])
-            assert resp.status_code == 404
-            assert json.loads(resp.get_data(as_text=True))['message'] == 'No result found'
+def test_send_sms_code_returns_404_for_bad_input_data(client):
+    data = json.dumps({})
+    import uuid
```
leohemsted: import should probably be at the top of the file

tests/app/user/test_rest_verify.py
```diff
@@ -293,22 +261,17 @@ def test_send_user_code_for_sms_with_optional_to_field(notify_api,
...
+    uuid_ = uuid.uuid4()
+    auth_header = create_authorization_header()
+    resp = client.post(
+        url_for('user.send_user_sms_code', user_id=uuid_),
+        data=data,
```
leohemsted: this can just be `json={}` to save the json.dumps line at the top of this test

tests/app/user/test_rest_verify.py
```diff
@@ -15,206 +14,175 @@
...
-                data=data,
-                headers=[('Content-Type', 'application/json'), auth_header])
-            assert resp.status_code == 400
-            json_resp = json.loads(resp.get_data(as_text=True))
-            assert 'Required field missing data' in json_resp['message']['password']
+    data = json.dumps({'bingo': 'bongo'})
+    auth_header = create_authorization_header()
+    resp = client.post(
+        url_for('user.verify_user_password', user_id=sample_user.id),
+        data=data,
```
leohemsted: can remove the data = json.dumps line and just write `json={'bingo': 'bongo'}` :drum:

tests/app/user/test_rest_verify.py
```diff
@@ -15,206 +14,175 @@
...
-            assert sample_user.failed_login_count == 1
+    assert sample_user.failed_login_count == 1
 
-            data = json.dumps({'password': 'password'})
-            auth_header = create_authorization_header()
-            resp = client.post(
-                url_for('user.verify_user_password', user_id=sample_user.id),
-                data=data,
-                headers=[('Content-Type', 'application/json'), auth_header])
+    data = json.dumps({'password': 'password'})
```
leohemsted: can remove the data = json.dumps line and just write `json={...}` here instead

tests/app/user/test_rest_verify.py
```diff
@@ -15,206 +14,175 @@
...
-            resp = client.post(
-                url_for('user.verify_user_password', user_id=sample_user.id),
-                data=data,
-                headers=[('Content-Type', 'application/json'), auth_header])
-            assert resp.status_code == 400
-            json_resp = json.loads(resp.get_data(as_text=True))
-            assert 'Incorrect password' in json_resp['message']['password']
+    resp = client.post(
+        url_for('user.verify_user_password', user_id=sample_user.id),
+        data=data,
```
leohemsted: can remove the data = json.dumps line and just write `json={...}` here instead

tests/app/user/test_rest_verify.py
```diff
@@ -15,206 +14,175 @@
...
+def test_user_verify_code(client,
+                          sample_sms_code):
+    assert not VerifyCode.query.first().code_used
+    data = json.dumps({
+        'code_type': sample_sms_code.code_type,
+        'code': sample_sms_code.txt_code})
+    auth_header = create_authorization_header()
+    resp = client.post(
+        url_for('user.verify_user_code', user_id=sample_sms_code.user.id),
+        data=data,
```
leohemsted: can remove the data = json.dumps line and just write `json={...}` here instead

app/dao/users_dao.py
```diff
@@ -82,6 +82,14 @@ def delete_user_verify_codes(user):
     db.session.commit()
 
 
+def count_user_verify_codes(user):
+    query = db.session.query(
+        func.count().label('count')
+    ).filter(VerifyCode.user == user,
+             VerifyCode.expiry_datetime <= datetime.utcnow()).one()
```
leohemsted: don't we specifically want to only capture verify codes where they _haven't_ expired and also where code_used = False?

otherwise if i log in and out (legitimately) a lot of times earlier in the day, it'll refuse to send

leohemsted: also, while you're looking at this fn, this can be done by just saying `return VerifyCode.query.filter(....).count()`, rather than needing .one and the label and all that gubbins

app/user/rest.py
```diff
@@ -137,6 +138,10 @@ def send_user_sms_code(user_id):
     user_to_send_to = get_user_by_id(user_id=user_id)
     verify_code, errors = request_verify_code_schema.load(request.get_json())
 
+    if count_user_verify_codes(user_to_send_to) >= current_app.config.get('MAX_VERIFY_CODE_COUNT'):
+        # Prevent more than `MAX_VERIFY_CODE_COUNT` active verify codes at a time
+        return jsonify({}), 204
```
leohemsted: think it might be worth throwing a logging statement in here

tests/app/user/test_rest_verify.py
```diff
@@ -249,6 +249,28 @@ def test_send_sms_code_returns_404_for_bad_input_data(client):
     assert json.loads(resp.get_data(as_text=True))['message'] == 'No result found'
 
 
+def test_send_sms_code_returns_204_when_too_many_codes_already_created(client, sample_user):
```
leohemsted: might need to add more tests, if you agree with my assumptions about `expiry_datetime` and `code_used`

PR #829, opened 2017-02-14
Add endpoint for service monthly usage stats by template

imdadahad:
This adds an endpoint to retrieve stats for a service grouped by template per month.

##  How it works

Call the endpoint with a given service id and financial year. It will return a dictionary containing counts of notifications by status for each template per month.

E.g. Calling for 2016 would return in the format of:

```
{
    '2017-03': { 
        "first_template_id":  {
            "name": "sample template one",
            "counts": { "sending": 0, "created": 0, .....}
        },
        "second_template_id": {
            "name": "sample template two",
            "counts": { "sending": 0, "created": 5, .....}
        },
        ...
     }, 
    '2017-02': {}, 
    '2017-01': {},
    ...
}
```

### DIFF
--- a/app/dao/templates_dao.py
+++ b/app/dao/templates_dao.py
@@ -13,8 +13,8 @@ from app.dao.dao_utils import (
 
 @transactional
 @version_class(Template, TemplateHistory)
-def dao_create_template(template, template_id=None):
-    template.id = template_id or uuid.uuid4()  # must be set now so version history model can use same id
+def dao_create_template(template):
+    template.id = uuid.uuid4()  # must be set now so version history model can use same id
     template.archived = False
     db.session.add(template)
 
--- a/app/service/statistics.py
+++ b/app/service/statistics.py
@@ -1,7 +1,7 @@
 import itertools
 from datetime import datetime, timedelta
 
-from app.models import TEMPLATE_TYPES
+from app.models import NOTIFICATION_STATUS_TYPES, TEMPLATE_TYPES
 
 
 def format_statistics(statistics):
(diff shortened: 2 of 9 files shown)


quis:
Format of the JSON looks good, matches https://github.com/alphagov/notifications-api/pull/807

app/service/statistics.py
```diff
@@ -29,6 +29,28 @@ def format_weekly_notification_stats(statistics, service_created_at):
     return week_dict
 
 
+def format_monthly_template_notification_stats(year, rows):
+    dict = {
```
quis: Probably shouldn’t use `dict` as a variable name…

quis:
Trying to build on this now. Think the JSON could be different to save doing work in the admin:

```
{
    '2017-03': [
        {
            "id": "first_template_id"
            "type": "email",
            "name": "sample template one",
            "counts": { "sending": 0, "created": 0, .....}
        },
        {
            "id": "second_template_id",
            "type": "email",
            "name": "sample template two",
            "counts": { "sending": 0, "created": 5, .....}
        },
        ...
     ], 
    '2017-02': {}, 
    '2017-01': {},
    ...
}
```

So that’s:
- `list` instead of `dict` for templates (ideally sorted by templates used)
- adds `type` of template (we show this in the front end)

quis:
Change of plan: shouldn’t do sorting on the API site because it would mean summing the counts. Summing the counts is presentational, and should be handled on the admin side.

app/dao/templates_dao.py
```diff
@@ -13,8 +13,8 @@
 
 @transactional
 @version_class(Template, TemplateHistory)
-def dao_create_template(template):
-    template.id = uuid.uuid4()  # must be set now so version history model can use same id
+def dao_create_template(template, template_id=None):
```
servingUpAces: Why change the method signature for tests?

tests/app/dao/test_services_dao.py
```diff
@@ -655,3 +658,90 @@ def test_dao_resume_service_marks_service_as_active_and_api_keys_are_still_revok
...
+    notify_db_session
+):
+    notification_history = functools.partial(
+        create_notification_history,
+        notify_db,
+        notify_db_session,
+        status='delivered'
+    )
+
+    template_one = create_sample_template(notify_db, notify_db_session, template_id=uuid.uuid4())
```
servingUpAces: This appears to be the only place you are using the new template_id parameter of dao_create_template, but for what purpose? The change does not seem necessary

PR #834, opened 2017-02-16
ensure we reset failed_login_count when appropriate

leohemsted:
in verify_user_password, if succesful we reset the failed_login_count.
now we use failed_login_count for 2FA attempts, we need to make sure we
reset it in other places too, so that people don't get blocked,
especially in the reset-password user journey.

* verify_user_code - if it's succesful, reset the failed_login_count
* update_password - reset failed_login_count because either
  * you're logged in and so it's 0 anyway
  * you're resetting your password via pword reset link, and the old
    count isn't relevant anymore

update_user now resets failed_login_count if password is changed  …
until work is done to stop using PUT /user/{id} on the admin app, this
function also needs to reset failed logins, cos it's used during the
forgotten password flow

best viewed in the separate commits 	ac9739f and 521872c - 	4d4d37b is just whitespace changes

### DIFF
--- a/app/dao/users_dao.py
+++ b/app/dao/users_dao.py
@@ -114,7 +114,6 @@ def reset_failed_login_count(user):
 
 def update_user_password(user, password):
     # reset failed login count - they've just reset their password so should be fine
-    user.failed_login_count = 0
     user.password = password
     user.password_changed_at = datetime.utcnow()
     db.session.add(user)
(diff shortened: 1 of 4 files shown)


leohemsted:
fix squashed out (https://github.com/alphagov/notifications-api/pull/834/files#diff-af9b5392bd33c53f8bf7cc055f8c99a8R294)

leohemsted:
gonna go ahead and merge this so i unblock the deploy pipeline

PR #851, opened 2017-03-03
Store who updated providers

imdadahad:
This stores adds an audit trail for who updated a provider (whether it was an automatic or manual switch). Additionally this adds robustness to switching providers when both providers have equal priorities.

## How it works

Currently there are two ways to update a provider:

1. Either through the API
2. Automatically (on failure or slow notification delivery)

In the former case, the update endpoint has been updated to store who made the change. We expect this to be passed in by the admin (it currently is). Additionally the update endpoint code has been refactored (removed marshmallow) with validation to ensure only few fields are updated.  

In the latter case we are now storing the 'Notify service user' as being the user that updated the provider when an auto-provider switch occurs. This user is created in our database migrations so works out better than having two separate database columns to indicate how a provider was switch (auto vs. manual) and who switched (if manual).

## Retrieving the history of provider updates

An extra endpoint has been added to retrieve historical records of provider updates. This will facilitate viewing an individual provider's history on the admin.

## Adding robustness to provider switching

During development of the above, a bug was spotted where auto-switching did not take place properly in the case where provider priorities are equal. This is because previously we were using `get_current_provider` which doesn't make sense when two active providers have equal priorities. The code has been refactored (along with tests) to ensure this is handled properly.


app/provider_details/rest.py
```diff
@@ -29,19 +29,37 @@ def get_provider_by_id(provider_details_id):
...
-    invalid_keys = {'identifier', 'version', 'updated_at'} & set(key for key in request.get_json().keys())
     if invalid_keys:
         message = "Not permitted to be updated"
         errors = {key: [message] for key in invalid_keys}
         raise InvalidRequest(errors, status_code=400)
 
-    dao_update_provider_details(update_dict)
-    return jsonify(provider_details=provider_details_schema.dump(fetched_provider_details).data), 200
+    provider = get_provider_details_by_id(provider_details_id)
+    req_json = request.get_json()
```
servingUpAces: Small thing, but you could set this variable above line 45 to avoid calling request.get_json() twice.

migrations/versions/0068_add_created_by_to_provider.py
```diff
@@ -0,0 +1,36 @@
...
+    op.create_foreign_key(None, 'provider_details_history', 'users', ['created_by_id'], ['id'])
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(None, 'provider_details_history', type_='foreignkey')
+    op.drop_index(op.f('ix_provider_details_history_created_by_id'), table_name='provider_details_history')
+    op.drop_column('provider_details_history', 'created_by_id')
+    op.drop_constraint(None, 'provider_details', type_='foreignkey')
```
imdadahad: Ah alembic doesn't add constraint names 😞. Just added them in now.

app/provider_details/rest.py
imdadahad: 👍

PR #855, opened 2017-03-10
Add new job status for dvla file processing

servingUpAces:
Add new job status for dvla file processing
`ready to send` == we have created a dvla file in S3, the job id is in the file name
`sent to dvla` == we have created an aggregate dvla file to send for the day and have sent the file via sftp


app/models.py
```diff
@@ -416,6 +416,8 @@ class ProviderDetailsHistory(db.Model, HistoryModel):
 JOB_STATUS_SENDING_LIMITS_EXCEEDED = 'sending limits exceeded'
 JOB_STATUS_SCHEDULED = 'scheduled'
 JOB_STATUS_CANCELLED = 'cancelled'
+JOB_STATUS_READY_TO_SEND = 'ready to send'
```
imdadahad: Why not `ready_to_send`?

servingUpAces: Because all the other statuses have spaces rather than underscores.

PR #856, opened 2017-03-13
Added v2 template get ID endpoint, schema & tests

kentsanggds:
## What

Added v2/template/{id} endpoint with JSON schema and supporting tests.

Details outlined in this document under `Templates / Get Template by ID` section - [Notify V2 API](https://docs.google.com/document/d/1EuCipShwpZKVPQp3N0j7yRKazS34v5Bj5E9epl89iqY/edit#)

## How to review 

execute `py.test tests/app/v2/template/`

### References

[Pivotal Tracker story] (https://www.pivotaltracker.com/n/projects/1443052/stories/141482957)

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -110,9 +110,11 @@ def register_blueprint(application):
 def register_v2_blueprints(application):
     from app.v2.notifications.post_notifications import notification_blueprint as post_notifications
     from app.v2.notifications.get_notifications import notification_blueprint as get_notifications
+    from app.v2.template.get_template import template_blueprint
 
     application.register_blueprint(post_notifications)
     application.register_blueprint(get_notifications)
+    application.register_blueprint(template_blueprint)
 
 
 def init_app(app):
--- a/app/v2/template/get_template.py
+++ b/app/v2/template/get_template.py
@@ -1,6 +1,7 @@
 import uuid
 
 from flask import jsonify, request
+from jsonschema.exceptions import ValidationError
 from werkzeug.exceptions import abort
 
 from app import api_user
@@ -11,21 +12,19 @@ from app.v2.template.template_schemas import get_template_by_id_request
 
 
 @template_blueprint.route("/<template_id>", methods=['GET'])
-@template_blueprint.route("/<template_id>/version/<version>", methods=['GET'])
+@template_blueprint.route("/<template_id>/version/<int:version>", methods=['GET'])
 def get_template_by_id(template_id, version=None):
     try:
-        casted_id = uuid.UUID(template_id)
-
         _data = {}
         _data['id'] = template_id
         if version:
-            _data['version'] = int(version)
+            _data['version'] = version
 
         data = validate(_data, get_template_by_id_request)
     except ValueError or AttributeError:
         abort(404)
 
     template = templates_dao.dao_get_template_by_id_and_service_id(
-        casted_id, api_user.service_id, data.get('version'))
+        template_id, api_user.service_id, data.get('version'))
 
     return jsonify(template.serialize()), 200
--- a/app/v2/template/get_template.py
+++ b/app/v2/template/get_template.py
@@ -1,0 +1,31 @@
+import uuid
+
+from flask import jsonify, request
+from werkzeug.exceptions import abort
+
+from app import api_user
+from app.dao import templates_dao
+from app.schema_validation import validate
+from app.v2.template import template_blueprint
+from app.v2.template.template_schemas import get_template_by_id_request
+
+
+@template_blueprint.route("/<template_id>", methods=['GET'])
+@template_blueprint.route("/<template_id>/version/<version>", methods=['GET'])
+def get_template_by_id(template_id, version=None):
+    try:
+        casted_id = uuid.UUID(template_id)
+
+        _data = {}
+        _data['id'] = template_id
+        if version:
+            _data['version'] = int(version)
+
+        data = validate(_data, get_template_by_id_request)
+    except ValueError or AttributeError:
+        abort(404)
+
+    template = templates_dao.dao_get_template_by_id_and_service_id(
+        casted_id, api_user.service_id, data.get('version'))
+
+    return jsonify(template.serialize()), 200
--- a/app/v2/template/template_schemas.py
+++ b/app/v2/template/template_schemas.py
@@ -29,7 +29,7 @@ get_template_by_id_response = {
         },
         "updated_at": {
             "format": "date-time",
-            "type": "string",
+            "type": ["string", "null"],
             "description": "Date+time updated"
         },
         "created_by": {"type": "string"},
--- a/app/v2/template/template_schemas.py
+++ b/app/v2/template/template_schemas.py
@@ -4,7 +4,7 @@ from app.schema_validation.definitions import uuid
 
 get_template_by_id_request = {
     "$schema": "http://json-schema.org/draft-04/schema#",
-    "description": "schema for query parameters allowed when getting list of notifications",
+    "description": "schema for parameters allowed when getting template by id",
     "type": "object",
     "properties": {
         "id": uuid,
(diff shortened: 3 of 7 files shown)


app/v2/template/get_template.py
```diff
@@ -0,0 +1,19 @@
...
+from app.dao import templates_dao
+from app.schema_validation import validate
+from app.v2.template import template_blueprint
+from app.v2.template.template_schemas import get_template_by_id_schema_response
+
+
+@template_blueprint.route("/<id>", methods=['GET'])
+def get_template_by_id(id):
+    template = templates_dao.dao_get_template_by_id(
+        id, request.args.get('version'))
```
imdadahad: Ideally we want to avoid using a variable name that's a 'built-in', as it can be confusing. Could just replace with `template_id`?

Also would be nice to do some validation here, especially if this will be used as a public endpoint.

app/__init__.py
```diff
@@ -136,7 +142,8 @@ def required_authentication():
 
     @app.before_request
     def record_user_agent():
-        statsd_client.incr("user-agent.{}".format(process_user_agent(request.headers.get('User-Agent', None))))
+        statsd_client.incr(
```
imdadahad: Spacing and line returns look off, think we should undo all of the 'linting' related changes once your editor is configured properly?

app/v2/template/template_schemas.py
```diff
@@ -0,0 +1,29 @@
...
+            "format": "date-time",
+            "type": "string",
+            "description": "Date+time created"
+        },
+        "updated_at": {
+            "format": "date-time",
+            "type": "string",
+            "description": "Date+time updated"
+        },
+        "created_by": {"type": ["string", "null"]},
```
imdadahad: Don't think this should be `null`?

tests/app/v2/template/test_get_template.py
```diff
@@ -0,0 +1,99 @@
...
+LETTER_TYPE = 'letter'
+
+template_types = [EMAIL_TYPE, SMS_TYPE, LETTER_TYPE]
+valid_versions = [None, 1]
+
+
+@pytest.mark.parametrize("tmp_type", template_types)
+@pytest.mark.parametrize("version", valid_versions)
+def test_get_email_template_by_id_returns_200(client, notify_db, notify_db_session, sample_service, tmp_type, version):
+    template = create_sample_template(
```
imdadahad: You can use the `create_template` fixture in `../tests/app/db.py` instead to avoid importing the `notify_db` and `notify_db_session`

tests/app/v2/template/test_get_template.py
```diff
@@ -0,0 +1,99 @@
...
+        "subject": template.subject if tmp_type == 'email' else None
+    }
+
+    assert json_response == expected_response
+
+
+def test_get_template_with_invalid_template_id_returns_404(client, sample_service):
+    auth_header = create_authorization_header(service_id=sample_service.id)
+
+    invalid_template_id = 'invalid_template_id'
```
imdadahad: Do you need this here...given that the variable value is the same as the variable name?

tests/app/v2/template/test_get_template.py
```diff
@@ -0,0 +1,99 @@
...
+
+    assert response.status_code == 404
+    assert response.headers['Content-type'] == 'application/json'
+
+    json_response = json.loads(response.get_data(as_text=True))
+
+    assert json_response == {
+        "errors": [
+            {
+                "error": "NoResultFound",
```
imdadahad: So might be good to add validation (as mentioned further above) so the error messages actually indicate what's wrong. `NoResultFound` shouldn't be the error here, as the param was invalid in the first place and that's what we're testing.

tests/app/v2/template/test_get_template.py
```diff
@@ -0,0 +1,99 @@
...
+                "message": "No result found"
+            }
+        ],
+        "status_code": 404
+    }
+
+
+@pytest.mark.parametrize("tmp_type", template_types)
+def test_get_template_with_invalid_version_returns_404(client, notify_db, notify_db_session, sample_service, tmp_type):
+    template = create_sample_template(
```
imdadahad: `create_template` here too would be better.

app/v2/template/get_template.py
kentsanggds: ok, updated

app/__init__.py
kentsanggds: ok, removed `format on save` setting

app/v2/template/template_schemas.py
kentsanggds: ok, updated the V2 API doc to reflect it being required

tests/app/v2/template/test_get_template.py
kentsanggds: Need to create email templates with subjects, so sample_template is a better fit

app/v2/template/template_schemas.py
kentsanggds: Also added schema for request parameters

tests/app/v2/template/test_get_template.py
kentsanggds: According to Becca, this should be ok as other validation will be caught in schema tests

tests/app/v2/template/test_get_template.py
kentsanggds: ok, updated

app/v2/template/get_template.py
```diff
@@ -0,0 +1,31 @@
...
+        casted_id = uuid.UUID(template_id)
+
+        _data = {}
+        _data['id'] = template_id
+        if version:
+            _data['version'] = int(version)
+
+        data = validate(_data, get_template_by_id_request)
+    except ValueError or AttributeError:
+        abort(404)
```
imdadahad: `raise InvalidRequest(error, status_code=404)` instead of `abort(404)`?

app/v2/template/get_template.py
```diff
@@ -0,0 +1,31 @@
...
+from app.v2.template import template_blueprint
+from app.v2.template.template_schemas import get_template_by_id_request
+
+
+@template_blueprint.route("/<template_id>", methods=['GET'])
+@template_blueprint.route("/<template_id>/version/<version>", methods=['GET'])
+def get_template_by_id(template_id, version=None):
+    try:
+        casted_id = uuid.UUID(template_id)
+
```
imdadahad: We can probably remove this cast/check as the schema will attempt to validate this anyway?

tests/app/v2/template/test_get_template.py
```diff
@@ -0,0 +1,93 @@
...
+from flask import json
+
+from app import DATETIME_FORMAT
+from tests import create_authorization_header
+from tests.app.conftest import sample_template as create_sample_template
+
+EMAIL_TYPE = 'email'
+SMS_TYPE = 'sms'
+LETTER_TYPE = 'letter'
+
```
imdadahad: can use from app.models import TEMPLATE_TYPES as you did above

tests/app/v2/template/test_get_template.py
imdadahad: It has a default to add a subject if it's set to `EMAIL_TYPE`

tests/app/v2/template/test_get_template.py
```diff
@@ -0,0 +1,93 @@
...
+    assert json_response == {
+        "message": "The requested URL was not found on the server.  "
+                   "If you entered the URL manually please check your spelling and try again.",
+        "result": "error"
+    }
+
+
+@pytest.mark.parametrize("tmp_type", template_types)
+def test_get_template_with_invalid_version_returns_404(client, notify_db, notify_db_session, sample_service, tmp_type):
```
imdadahad: `test_get_template_with_non_existent_version_returns_404`?

PR #860, opened 2017-03-15
Create dvla file for a letter job

servingUpAces:
A new task has been created to build a dvla file from the notifications for a given job for letters.
- If all notifications for the job have been persisted to the database we then create a file to save to S3.
Each notification is transformed to a pipe delimited line that matches the DVLA file format, https://github.com/alphagov/notify-ftp-documentation/tree/master/dvla_documentation.
The notifications-ftp app will then aggregate all the files for the day to then send to DVLA to be posted. 
- Else if all the notifications are not persisted to the database the task goes to the retry queue and is retried in 5 minutes for 15 tries. It is likely a job will complete in about 15 minutes so it is unlikely the task will retry more than a few times. 

TODO: move the s3upload method to the notifications-utils package. This is not done at this time because I have been asked to build a security feature. 
TODO: update the job status to `ready to send` after the s3upload is successfully complete.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,10 +1,12 @@
+import botocore
+from boto3 import resource
 from datetime import (datetime)
 
 from flask import current_app
 from notifications_utils.recipients import (
     RecipientCSV
 )
-from notifications_utils.template import SMSMessageTemplate, WithSubjectTemplate
+from notifications_utils.template import SMSMessageTemplate, WithSubjectTemplate, LetterDVLATemplate
 from sqlalchemy.exc import SQLAlchemyError
 from app import (
     create_uuid,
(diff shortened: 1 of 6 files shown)


app/celery/tasks.py
```diff
@@ -248,13 +255,62 @@ def persist_letter(
...
+@notify_celery.task(bind=True, name="build-dvla-file", max_retries=15, default_retry_delay=300)
+@statsd(namespace="tasks")
+def build_dvla_file(self, job_id):
+    if all_notifications_are_created_for_job(job_id):
+        notifications = dao_get_all_notifications_for_job(job_id)
+        file = ""
+        for n in notifications:
+            t = {"content": n.template.content, "subject": n.template.subject}
+            # This unique id is a 7 digits requested by DVLA, not known if this number needs to be sequential.
+            unique_id = int(''.join(map(str, random.sample(range(9), 7))))
```
leohemsted: do we care enough to store this against each notification?

app/dao/jobs_dao.py
```diff
@@ -24,6 +28,25 @@ def dao_get_notification_outcomes_for_job(service_id, job_id):
...
 
+@statsd(namespace="dao")
+def all_notifications_are_created_for_job(job_id):
+    query = db.session.query(func.count(Notification.id), Job.id)\
+        .join(Job)\
+        .filter(Job.id == job_id)\
+        .group_by(Job.id)\
+        .having(func.count(Notification.id) == Job.notification_count).all()
+
+    if query:
```
leohemsted: `return query`

app/celery/tasks.py
servingUpAces: I'm not sure. I will investigate this on the next round.

PR #863, opened 2017-03-17
Added logging for any exception thrown in build_dvla_file

servingUpAces:


app/celery/tasks.py
```diff
@@ -265,22 +265,25 @@ def persist_letter(
...
+                file = file + str(template) + "\n"
+            s3upload(filedata=file,
+                     region=current_app.config['AWS_REGION'],
+                     bucket_name=current_app.config['DVLA_UPLOAD_BUCKET_NAME'],
+                     file_location="{}-dvla-job.text".format(job_id))
+        else:
+            current_app.logger.info("All notifications for job {} are not persisted".format(job_id))
+            self.retry(queue="retry", exc="All notifications for job {} are not persisted".format(job_id))
+    except Exception as e:
+        current_app.logger.exception("build_dvla_file threw exception: {}", e)
```
leohemsted: should call `raise` after

also don't need to use format in this - just `current_app.logger.exception("build_dvla_file threw exception")` should do, since `.exception` specifically then prints stack trace + exc info after

servingUpAces: oops forgot to raise - silly me

leohemsted: sorry, can you make the other change - to how the exception logger is called?

leohemsted: im just super wary of logging statements being called wrong, since if there's an error in a logging statement it masks real errors

PR #867, opened 2017-03-24
Bump utils to allow lists as placeholder values

quis:
Brings in:
- [x] https://github.com/alphagov/notifications-utils/pull/128

Also adds a test to make sure that passing a dictionary as a value in the `personalisation` argument works.

# What it means for users of the API

## With this template:

> Here are some animals:
>
> ((animal))

## …and this API call:
```python
send_email(…, personalisation={
    'animal', ['cat', 'rat', 'gnat']
})
```


quis:
Also bringing this one in as part of this PR:
- [x] alphagov/notifications-utils#134

PR #872, opened 2017-03-29
Fix daterange in service DAO to work with BST

kentsanggds:
## What 

Problem after setting time to BST in get_detailed_services detected by test test_get_detailed_services_for_date_range which was failing. 
After setting the time to a datetime object the query on the DB would search for a precise datetime range rather than a date range.
Fixed in the service DAO by setting the end_date to the end of the day by adding 23 hours, 59 minutes and 59 seconds and filtering on the query based on datetime and not date.

## How to review

execute py.test tests/app/service/test_rest.py::test_get_detailed_services_for_date_range

test has been updated with a range of datetimes for various states

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -31,7 +31,7 @@ from app.models import (
 )
 from app.service.statistics import format_monthly_template_notification_stats
 from app.statsd_decorators import statsd
-from app.utils import get_london_month_from_utc_column
+from app.utils import get_london_month_from_utc_column, get_london_midnight_in_utc
 
 
 def dao_fetch_all_services(only_active=False):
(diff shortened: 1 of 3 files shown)


tests/app/service/test_rest.py
```diff
@@ -1393,19 +1393,25 @@ def test_get_detailed_services_only_includes_todays_notifications(notify_db, not
...
-    create_sample_notification(notify_db, notify_db_session, created_at=datetime.now() - timedelta(days=1))
-    create_sample_notification(notify_db, notify_db_session, created_at=datetime.now())
 
-    start_date = (datetime.utcnow() - timedelta(days=2)).date()
-    end_date = (datetime.utcnow() - timedelta(days=1)).date()
+    with freeze_time(set_time):
+        create_sample_notification(notify_db, notify_db_session, created_at=datetime.now() - timedelta(days=3))
+        create_sample_notification(notify_db, notify_db_session, created_at=datetime.now() - timedelta(days=2))
+        create_sample_notification(notify_db, notify_db_session, created_at=datetime.now() - timedelta(days=1))
+        create_sample_notification(notify_db, notify_db_session, created_at=datetime.now())
```
leohemsted: we should probably make sure all of these are `utcnow`

app/dao/services_dao.py
```diff
@@ -324,14 +324,19 @@ def dao_fetch_todays_stats_for_all_services(include_from_test_key=True):
 
 @statsd(namespace='dao')
 def fetch_stats_by_date_range_for_all_services(start_date, end_date, include_from_test_key=True):
+    if not isinstance(end_date, datetime):
```
leohemsted: can we move the call of `get_london_midnight_in_utc` into this function, so this function is only ever called with naive date objects, which i feel makes most logical sense

PR #877, opened 2017-03-31
Update daily performance job 4am

kentsanggds:
## What 

Problems after change to BST as date passed to performance platform were 1 day out

## How to review

Performance platform stats should be correct for future data pushes

https://www.gov.uk/performance/govuk-notify

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -52,9 +52,7 @@ class Config(object):
     EXPIRE_CACHE_IN_SECONDS = 600
 
     # Performance platform
-    PERFORMANCE_PLATFORM_ENABLED = os.getenv('PERFORMANCE_PLATFORM_ENABLED') == '1'
-    PERFORMANCE_PLATFORM_URL = 'https://www.performance.service.gov.uk/data/govuk-notify/notifications'
-    PERFORMANCE_PLATFORM_TOKEN = os.getenv('PERFORMANCE_PLATFORM_TOKEN')
+    PERFORMANCE_PLATFORM_ENABLED = False
 
     # Logging
     DEBUG = False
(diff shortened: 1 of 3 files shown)


app/config.py
```diff
@@ -52,9 +52,7 @@ class Config(object):
     EXPIRE_CACHE_IN_SECONDS = 600
 
     # Performance platform
-    PERFORMANCE_PLATFORM_ENABLED = os.getenv('PERFORMANCE_PLATFORM_ENABLED') == '1'
-    PERFORMANCE_PLATFORM_URL = 'https://www.performance.service.gov.uk/data/govuk-notify/notifications'
-    PERFORMANCE_PLATFORM_TOKEN = os.getenv('PERFORMANCE_PLATFORM_TOKEN')
+    PERFORMANCE_PLATFORM_ENABLED = False
```
leohemsted: shouldn't be moving these vars, or they won't be defined outside of prod and it may crash

app/config.py
```diff
@@ -244,6 +242,9 @@ class Live(Config):
     FROM_NUMBER = '40604'
     FUNCTIONAL_TEST_PROVIDER_SERVICE_ID = '6c1d81bb-dae2-4ee9-80b0-89a4aae9f649'
     FUNCTIONAL_TEST_PROVIDER_SMS_TEMPLATE_ID = 'ba9e1789-a804-40b8-871f-cc60d4c1286f'
+    PERFORMANCE_PLATFORM_ENABLED = True
+    PERFORMANCE_PLATFORM_URL = 'https://www.performance.service.gov.uk/data/govuk-notify/notifications'
+    PERFORMANCE_PLATFORM_TOKEN = os.getenv('PERFORMANCE_PLATFORM_TOKEN')
```
leohemsted: this won't work on its own - you need to add the relevant line to the `cloudfoundry_config.py` file

PR #879, opened 2017-04-03
Bump memory on prod

minglis:
Bumps the available memory on worker instances on PaaS prod.

Also breaks up the schedule on longer running jobs.


sakisv:
Looks good 👍

PR #885, opened 2017-04-04
Platform admin query

servingUpAces:
Updated the fetch_stats_by_date_range to return test notifications if the start and end date are with in the last 7 days.
- If the date range is with the last 7 days we query Notifications.
- If the date range is outside of the last 7 days we query NotificationHistory.

NotificationHistory does not persist notifications created with a test api key, since January 10, 2017.

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -325,8 +325,7 @@ def dao_fetch_todays_stats_for_all_services(include_from_test_key=True):
 @statsd(namespace='dao')
 def fetch_stats_by_date_range_for_all_services(start_date, end_date, include_from_test_key=True):
     start_date = get_london_midnight_in_utc(start_date)
-    end_date = get_london_midnight_in_utc(end_date)
-    end_date += timedelta(hours=23, minutes=59, seconds=59)
+    end_date = get_london_midnight_in_utc(end_date + timedelta(days=1))
 
     query = db.session.query(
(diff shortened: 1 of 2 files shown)


app/dao/services_dao.py
```diff
@@ -325,27 +325,48 @@ def dao_fetch_todays_stats_for_all_services(include_from_test_key=True):
 @statsd(namespace='dao')
 def fetch_stats_by_date_range_for_all_services(start_date, end_date, include_from_test_key=True):
     start_date = get_london_midnight_in_utc(start_date)
-    end_date = get_london_midnight_in_utc(end_date)
-    end_date += timedelta(hours=23, minutes=59, seconds=59)
+    end_date = get_london_midnight_in_utc(end_date + timedelta(days=1))
+
+    if start_date >= datetime.utcnow() - timedelta(days=7):
```
sakisv: Although not a big fan of it, I think this is a place that duck typing would be useful to eliminate duplicated code, something along the lines of

    if start_date >= datetime.utcnow() - timedelta(days=7):
        NotificationTable = Notification
    else:
        NotificationTable = NotificationHistory

leohemsted: +1 to suggested change - we already do that elsewhere

https://github.com/alphagov/notifications-api/blob/master/app/dao/notifications_dao.py#L92-L95

PR #886, opened 2017-04-05
Revert "Revert "Bump utils to allow lists as placeholder values""

quis:
Reverts alphagov/notifications-api#883

### DIFF
--- a/tests/app/notifications/rest/test_send_notification.py
+++ b/tests/app/notifications/rest/test_send_notification.py
@@ -133,10 +133,39 @@ def test_send_notification_with_placeholders_replaced(notify_api, sample_email_t
             assert response_data['subject'] == 'Jo'
 
 
-def test_send_notification_with_placeholders_replaced_with_list(
+@pytest.mark.parametrize('personalisation, expected_body, expected_subject', [
+    (
+        ['Jo', 'John', 'Josephine'],
+        (
+            'Hello \n\n'
+            '* Jo\n'
+            '* John\n'
+            '* Josephine\n'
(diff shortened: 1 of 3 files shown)


quis:
In order for this to safely be reverted:

## Add handling for data types other than list/string

Brings in:
- [ ] alphagov/notifications-utils#135

Also adds extra tests for:
- the exact issue that we saw in production when #867 was deployed
- what happens when `None` is passed as a placeholder value, because this should never get as far as the relevant bit of utils

quis:
Also now includes more fixes to letter output:
- [x] alphagov/notifications-utils#136
- [ ] alphagov/notifications-utils#131

quis:
Going to merge this because it’s just utils changes that have already been reviewed.

tests/app/notifications/rest/test_send_notification.py
```diff
@@ -133,6 +133,65 @@ def test_send_notification_with_placeholders_replaced(notify_api, sample_email_t
...
+        6,
+        (
+            'Hello 6\n'
+            'This is an email from GOV.\u200BUK'
+        ),
+        '6',
+    ),
+    pytest.mark.xfail((
+        None,
+        ('we consider None equivalent to missing personalisation'),
```
leohemsted: i think rather than this xfailing, it should probably assert that the placeholder's still in. maybe split out to a separate test function if that makes sense.

i mean, this is only xfailing because you're setting the value to something we know it won't be - this is currently a "test as documentation" rather than "does our code continue to work in the same way"

PR #888, opened 2017-04-05
Throw exception on long running client call

minglis:
Added a 60 second timeout to the MMG and fire text clients.
- This is a CONNECT and a READ timeout.
- Gets wrapped in the standard client exception, with a status code of 504, message Gateway timeout.
- Is quiet noisy in logs to allow us to see it
- Ensures we flick across the provider.

To test change the timeout to 0 and it will timeout.

### DIFF
--- a/tests/app/clients/test_firetext.py
+++ b/tests/app/clients/test_firetext.py
@@ -129,7 +129,7 @@ def test_send_sms_override_configured_shortcode_with_sender(mocker, mock_firetex
     assert request_args['from'][0] == 'fromservice'
 
 
-def test_send_sms_raises_if_firetext_rejects_with_timeout(rmock, mock_firetext_client):
+def test_send_sms_raises_if_firetext_rejects_with_connect_timeout(rmock, mock_firetext_client):
     to = content = reference = 'foo'
 
     with pytest.raises(FiretextClientResponseException) as exc:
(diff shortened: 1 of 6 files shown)


tests/app/clients/test_firetext.py
```diff
@@ -126,3 +127,25 @@ def test_send_sms_override_configured_shortcode_with_sender(mocker, mock_firetex
...
+
+    with pytest.raises(FiretextClientResponseException) as exc:
+        rmock.register_uri('POST', 'https://www.firetext.co.uk/api/sendsms/json', exc=ConnectTimeout)
+        mock_firetext_client.send_sms(to, content, reference)
+
+    assert exc.value.status_code == 504
+    assert exc.value.text == 'Gateway Time-out'
+
+
+def test_send_sms_raises_if_firetext_rejects_with_timeout(rmock, mock_firetext_client):
```
leohemsted: needs to be different name to above, or the top one won't run

tests/app/clients/test_mmg.py
```diff
@@ -113,3 +114,25 @@ def test_send_sms_raises_if_mmg_fails_to_return_json(notify_api, mocker):
...
+
+    with pytest.raises(MMGClientResponseException) as exc:
+        rmock.register_uri('POST', 'https://api.mmg.co.uk/json/api.php', exc=ConnectTimeout)
+        mmg_client.send_sms(to, content, reference)
+
+    assert exc.value.status_code == 504
+    assert exc.value.text == 'Gateway Time-out'
+
+
+def test_send_sms_raises_if_firetext_rejects_with_timeout(rmock):
```
leohemsted: need to sort out these test names to

PR #892, opened 2017-04-06
V2 preview subject personalisation

kentsanggds:
## What

Personalisation in v2 preview for the subject line was not correctly handled and the preview schema was expecting personalisation when it is legitimate to not provide it for a preview.

## How to review 

pytest -k test_valid_post_template_returns_200

## Reference 

[V2 API specification](https://docs.google.com/document/d/1EuCipShwpZKVPQp3N0j7yRKazS34v5Bj5E9epl89iqY/edit#heading=h.pdggcx1vfo8x)

### DIFF
--- a/app/v2/template/post_template.py
+++ b/app/v2/template/post_template.py
@@ -29,8 +29,7 @@ def post_template_preview(template_id):
     check_placeholders(template_object)
 
     resp = create_post_template_preview_response(template=template,
-                                                 body=str(template_object),
-                                                 url_root=request.url_root)
+                                                 body=str(template_object))
 
     return jsonify(resp), 200
 
--- a/app/v2/template/post_template.py
+++ b/app/v2/template/post_template.py
@@ -1,8 +1,5 @@
-import uuid
-
 from flask import jsonify, request
 from jsonschema.exceptions import ValidationError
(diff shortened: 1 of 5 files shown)


app/v2/template/post_template.py
```diff
@@ -28,9 +29,11 @@ def post_template_preview(template_id):
 
     check_placeholders(template_object)
 
+    subject = template_object.subject if template.template_type != SMS_TYPE else None
+
     resp = create_post_template_preview_response(template=template,
```
imdadahad: Small point.

It would be nicer if `create_post_template_preview_response` took in `template` and `template_object`. That way the logic for handling subject and returning `str(template_object)` can be abstracted into that function.

app/v2/template/post_template.py
```diff
@@ -28,9 +29,11 @@ def post_template_preview(template_id):
...
-                                                 body=str(template_object),
-                                                 url_root=request.url_root)
+                                                 subject=subject,
```
imdadahad: Small point.

It would be nicer if `create_post_template_preview_response` took in `template` and `template_object`. That way the logic for handling subject and returning `str(template_object)` can be abstracted into that function.

PR #898, opened 2017-04-11
Update letter jobs

kentsanggds:
## What 

Updated the response from send_letter_jobs to return json block rather than plain text.

## How to review 

execute `pytest tests/app/letters/test_send_letter_jobs.py`

### DIFF
--- a/app/letters/send_letter_jobs.py
+++ b/app/letters/send_letter_jobs.py
@@ -1,3 +1,5 @@
+import json
+
 from flask import Blueprint, jsonify
 from flask import request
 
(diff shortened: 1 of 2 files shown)


app/letters/send_letter_jobs.py
```diff
@@ -17,7 +17,7 @@ def send_letter_jobs():
     job_ids = validate(request.get_json(), letter_job_ids)
     notify_celery.send_task(name="send-files-to-dvla", args=(job_ids['job_ids'],), queue="process-ftp")
 
-    return "Task created to send files to DVLA"
+    return jsonify(data={"response": "Task created to send files to DVLA"}), 200
```
minglis: Other places we sometimes return a 201 created in such a case?

minglis: we might be being inconsistant across the app with that though....

PR #901, opened 2017-04-12
Added the random string reference to the letter

minglis:
- uses the reference field on the notifications table to store a 16char random string used to cross reference DVLA letters back to the notification
- used as letter barcode does not have space for a UUID notification id

- [x] Depends on [https://github.com/alphagov/notifications-utils/pull/149](https://github.com/alphagov/notifications-utils/pull/149)

Renamed the numeric_id to notification_reference in utils and changed validation rules to match this

Note also the persist_notification method set "reference" to be "client_reference" which is confusing and they are different things, so fixed this too.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -47,7 +47,7 @@ def post_notification(notification_type):
                                         notification_type=notification_type,
                                         api_key_id=api_user.id,
                                         key_type=api_user.key_type,
-                                        reference=form.get('reference', None),
+                                        client_reference=form.get('reference', None),
                                         simulated=simulated)
     if not simulated:
         queue_name = 'priority' if template.process_type == PRIORITY else None
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,4 +1,6 @@
 import os
+import random
+import string
 import uuid
 
 from flask import Flask, _request_ctx_stack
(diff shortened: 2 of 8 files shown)


minglis:
the existing test_create_dvla_file_contents test in test_tasks is a little confusing, it mocks the template from utils and then checks the args, passed to it. I change the test a bit to check for both references (one from each created notification) but it threw me a bit so whoever looks at this could you look at it too and see what the thoughts are. ta.

app/v2/notifications/post_notifications.py
```diff
@@ -47,7 +47,7 @@ def post_notification(notification_type):
                                         notification_type=notification_type,
                                         api_key_id=api_user.id,
                                         key_type=api_user.key_type,
-                                        reference=form.get('reference', None),
+                                        client_reference=form.get('reference', None),
```
leohemsted: if someone posted an email with a reference, and then we send it to SES, did we overwrite the user's reference with SES's one?

leohemsted: do we want/need/have the ability to migrate old data to client_reference?

minglis: Actually I don't think so. The persist only ever set the client-reference. Even if that was from the "reference" post body field.

So the data would not be overwritten.

minglis: my change adds in the reference method param, and puts the two things back into the correct params.

leohemsted: ah right, just params, not actual db values. gotcha. :ok_man:

PR #902, opened 2017-04-18
Task to set job to error

minglis:
Handle the case where building the DVLA file fails - ie when a file is not in S3 or it's unable to be concatenated.

When this happens we callback and put the JOB into error, but leave the notifications as they were.

We can then resend/debug etc that file.


migrations/versions/0071_add_job_error_state.py
```diff
@@ -0,0 +1,22 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.execute("INSERT INTO JOB_STATUS VALUES('error')")
+
+
+def downgrade():
+    op.execute("DELETE FROM JOB_STATUS WHERE name = 'error'")
```
leohemsted: need to rollback jobs that have status of `error`, or the downgrade will fail

see https://github.com/alphagov/notifications-api/blob/master/migrations/versions/0053_cancelled_job_status.py

tests/app/celery/test_tasks.py
```diff
@@ -1052,3 +1052,14 @@ def test_update_job_to_sent_to_dvla(sample_letter_template, sample_letter_job):
...
     assert 'sent to dvla' == Job.query.filter_by(id=sample_letter_job.id).one().job_status
+
+
+def test_update_dvla_job_to_error(sample_letter_template, sample_letter_job):
+    create_notification(template=sample_letter_template, job=sample_letter_job)
+    create_notification(template=sample_letter_template, job=sample_letter_job)
+    update_dvla_job_to_error(job_id=sample_letter_job.id)
+
+    updated_notifications = Notification.query.all()
+    assert [(n.status == 'created', n.sent_by == 'dvla') for n in updated_notifications]
```
leohemsted: this only asserts that the array is truthy - ie, that it has a non-zero length (and asserting tuples also just checks that they have contents in 

you'll have to unpack this i'm afraid

minglis:
Done all of the things.

- tests pass
- rollback added

PR #903, opened 2017-04-18
Add v2 template get all

kentsanggds:
## What

Added v2 template get all tests, schemas, endpoint which will get all templates of all types or get a specific list of templates matching a template type depending on the query parameter passed in. 

## How to review

Execute `pytest tests/app/v2/templates/test_get_templates.py`

## Reference 

[v2 API Specification](https://docs.google.com/document/d/1EuCipShwpZKVPQp3N0j7yRKazS34v5Bj5E9epl89iqY/edit#heading=h.dbhrumn2rdd0)


app/v2/templates/get_templates.py
```diff
@@ -0,0 +1,19 @@
...
+from app import api_user
+from app.dao import templates_dao
+from app.schema_validation import validate
+from app.v2.templates import v2_templates_blueprint
+from app.v2.templates.templates_schemas import get_all_template_request
+
+
+@v2_templates_blueprint.route("/", methods=['GET'])
+def get_templates():
+    validate(request.args.to_dict(), get_all_template_request)
```
minglis: I'm confused by this - it appears the request can have a template type filter - but this is not used in the query?

tests/app/v2/templates/test_get_templates.py
```diff
@@ -0,0 +1,91 @@
...
+
+    response = client.get(path='/v2/templates/?type={}'.format(tmp_type),
+                          headers=[('Content-Type', 'application/json'), auth_header])
+
+    assert response.status_code == 200
+    assert response.headers['Content-type'] == 'application/json'
+
+    json_response = json.loads(response.get_data(as_text=True))
+
+    assert len(json_response['templates']) == num_templates
```
minglis: Should this test not be checking that it gets back a single template of the given type from a set of three, one of each type?

Links to comment above where the query doesn't restrict on template type.

app/v2/templates/get_templates.py
kentsanggds: Thats true, I'll revisit the code, this is what happens when you get side tracked

tests/app/v2/templates/test_get_templates.py
kentsanggds: ok, will make change

PR #905, opened 2017-04-19
exclude letters from scheduled cleanup tasks

leohemsted:
letters shouldn't be timed out after 72 hours - it can take longer than that just to post them yet alone get delivery receipts. keep them in created/sending forever.

letters should, however, still be deleted after a week

also i refactored the timeout function cos it was gross

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -1,3 +1,4 @@
+import functools
 import pytz
 from datetime import (
     datetime,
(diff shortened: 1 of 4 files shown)


minglis:
Think this is OK.

We need to talk about lifetime of letters - is 7 days long enough?

👍

leohemsted:
yeah i don't think 7 days is long enough, but whether we strip personal info but keep it around, perhaps?

PR #906, opened 2017-04-19
add DVLA organisations to API

leohemsted:
when services are created, they'll have a dvla_org_id of 001, or HM Government. That can be changed later using a regular update call as you would update any other part of the service


quis:
@leohemsted Added the code to pass the organisation ID through in 5720f05

tests/app/dvla_organisation/test_rest.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_get_dvla_organisations(client):
+    auth_header = create_authorization_header()
+
+    response = client.get('/dvla_organisations', headers=[auth_header])
+
+    assert response.status_code == 200
+    dvla_organisations = json.loads(response.get_data(as_text=True))
+    assert dvla_organisations == {'001': 'HM Government', '500': 'Land Registry'}
```
quis: Thumbs up to the format of this response.

imdadahad:
@quis I missed the conversation yesterday on Slack. Any reason why we're setting the `dvla_organisation_id` at the `service` level as opposed to on `service.organisation`?

migrations/versions/0072_add_dvla_orgs.py
```diff
@@ -0,0 +1,60 @@
...
+        op.f('ix_services_history_dvla_organisation_id'),
+        'services_history',
+        ['dvla_organisation_id'],
+        unique=False
+    )
+
+    op.create_foreign_key(None, 'services', 'dvla_organisation', ['dvla_organisation_id'], ['id'])
+
+def downgrade():
+    op.drop_column('services_history', 'dvla_organisation_id')
```
imdadahad: This is pretty minute but any reason why we don't drop the indexes and constraints first? We've done this in other migrations...

leohemsted: they're done automatically when you drop a column, since they are dependent. it's done in other migrations coss they were autogenerated probs

imdadahad: yeah it doesn't auto generate well. Like it does `op.create_foreign_key(None...` in the upgrade and then `op.drop_constraint(None..` in the downgrade, but that obviously causes issues if you try to downgrade. Slipped me up in the past.

leohemsted: yeah - i've tested locally and downgrade seems to behave

imdadahad: yeah here it would as you don't have that corresponding auto-generated `drop_constraint(None` line in the downgrade 👍

tests/app/service/test_rest.py
```diff
@@ -371,41 +371,41 @@ def test_create_service_should_throw_duplicate_key_constraint_for_existing_email
             assert "Duplicate service name '{}'".format(service_name) in json_resp['message']['name']
 
 
-def test_update_service(notify_api, notify_db, sample_service):
+def test_update_service(client, notify_db, sample_service):
```
imdadahad: Do we also need to update the `create_service` test to check that it creates with the default `DVLA_ORG_HM_GOVERNMENT`?

PR #908, opened 2017-04-20
Make API responsible formatting notifications for CSV

imdadahad:
## Summary

Currently on Admin we `call get_all_notifications_for_job` and then format each notification in the result set before outputting as a CSV. This however slowed down performance, especially around `datetime` conversions. 

This moves the formatting over to the API so we gain faster performance as we no longer to need to do these unnecessary `datetime` conversions which were costly:

```python

python -m timeit -n 1000 -v -s "import dateutil.parser; from datetime import datetime; dt = '2017-04-20T11:46:32.175452'" -c "dateutil.parser.parse(dt)"
raw times: 0.0943 0.0964 0.0924
1000 loops, best of 3: 92.4 usec per loop

```
We were also calling `datetime.strftime()` for each notification, which became costly. We no longer do that. Instead we use `time.stftime()` only once because we have access to the `datetime` on the API. This is faster than `datetime.strftime()`:

```python
→ python3 -m timeit -n 100000 -v -s "from datetime import datetime;now = datetime.utcnow()" -c "now.strftime('%A %d %B %Y')"
raw times: 0.514 0.505 0.503
100000 loops, best of 3: 5.03 usec per loop

→ python3 -m timeit -n 100000 -v -s "from datetime import datetime; import time; now = datetime.utcnow()" -c "time.strftime('%A %d %B %Y')"
raw times: 0.205 0.212 0.204
100000 loops, best of 3: 2.04 usec per loop
```


tests/app/test_model.py
```diff
@@ -65,3 +71,96 @@ def test_status_conversion_handles_failed_statuses(initial_statuses, expected_st
...
+        "row_number": 1,
+        "template_name": "Template Name",
+        "template_type": "sms",
+        "created_at": "Friday 01 January 2016 at 11:09",
+        "job_name": "some.csv",
+        "status": "Sending"
+    }
+
+    serialized = notification.serialize_for_csv()
+    assert serialized == expected_resp
```
kentsanggds: Perhaps just do an assertion on the run_number being 1 rather than matching the whole thing in case underlying sample data changes.

tests/app/test_model.py
```diff
@@ -65,3 +71,96 @@ def test_status_conversion_handles_failed_statuses(initial_statuses, expected_st
...
+        "row_number": "",
+        "template_name": "Template Name",
+        "template_type": template_type,
+        "created_at": "Friday 01 January 2016 at 11:09",
+        "job_name": "some.csv",
+        "status": "Sending"
+    }
+
+    serialized = notification.serialize_for_csv()
+    assert serialized == expected_resp
```
kentsanggds: Perhaps just do an assertion on the template_type in case underlying sample changes.

tests/app/test_model.py
```diff
@@ -65,3 +71,96 @@ def test_status_conversion_handles_failed_statuses(initial_statuses, expected_st
...
+        "row_number": 52,
+        "template_name": "Template Name",
+        "template_type": "sms",
+        "created_at": "Saturday 30 January 2016 at 12:39",
+        "job_name": "some.csv",
+        "status": "Phone not accepting messages right now"
+    }
+
+    serialized = notification.serialize_for_csv()
+    assert serialized == expected_resp
```
kentsanggds: Likewise just check that status is correct, maybe parameterise the status for different responses if it's not already be covered by other tests.

tests/app/test_model.py
```diff
@@ -65,3 +71,96 @@ def test_status_conversion_handles_failed_statuses(initial_statuses, expected_st
...
+        "row_number": 101,
+        "template_name": "Template Name",
+        "template_type": "sms",
+        "created_at": "Monday 27 March 2017 at 00:01",
+        "job_name": "some.csv",
+        "status": "Phone number doesn’t exist"
+    }
+
+    serialized = notification.serialize_for_csv()
+    assert serialized == expected_resp
```
kentsanggds: Again just check that created_at is correct.

tests/app/job/test_rest.py
```diff
@@ -715,3 +715,31 @@ def create_10_jobs(db, session, service, template):
...
+    notification = create_notification(
+        notify_db,
+        notify_db_session,
+        job=job,
+        job_row_number=1,
+        created_at=datetime.utcnow(),
+    )
+
+    response = client.get(
+        path='/service/{}/job/{}/notifications?'.format(notification.service.id, job.id, True),
```
imdadahad: We follow this pattern across the codebase. I think it's okay here, key thing is `query_string` clearly identifies the important parameter.

tests/app/test_model.py
imdadahad: Sure

PR #916, opened 2017-04-24
Update sms rate

servingUpAces:
Added a new rates table.
We only need one rate per channel. This reflects that. The provider_rates has been left for now, it is still not being used.
New dao has been added to select the right rate for the given notification_type and date of notification.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -944,3 +944,12 @@ class Event(db.Model):
         nullable=False,
         default=datetime.datetime.utcnow)
     data = db.Column(JSON, nullable=False)
+
+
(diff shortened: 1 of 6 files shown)


app/models.py
```diff
@@ -944,3 +944,12 @@ class Event(db.Model):
         nullable=False,
         default=datetime.datetime.utcnow)
     data = db.Column(JSON, nullable=False)
+
+
+class Rates(db.Model):
```
leohemsted: to be consistent, this should probably be `Rate` (singular)

PR #918, opened 2017-04-25
Add international support flag for providers (migration)

imdadahad:
This adds a `supports_international` flag on the providers and sets MMG as `True` for this.

The story was redefined: we don't need to have another provider, but do need some way of differentiating between `domestic` and `international`.

**Also: I'll update the migration to be up-to with master in a separate commit.**


minglis:
👍

PR #920, opened 2017-04-25
Autoconfirm http subscriptions

sakisv:
Allows for autoconfirm of an SNS subscription


tests/app/notifications/test_notifications_ses_callback.py
```diff
@@ -256,3 +300,7 @@ def ses_hard_bounce_callback():
...
...5f6a988666d92768dd60a747ba6f3beb71854e285d6ad02428b09ceece29417f1f02d609c582afbacc99c583a916b9981dd2728f4ae6fdb82efd087cc3b7849e05798d2d2785c03b0879594eeac82c01f235d0e717736",\n    "TopicArn": "arn:aws:sns:us-west-2:123456789012:MyTopic",\n    "Message": "You have chosen to subscribe to the topic arn:aws:sns:us-west-2:123456789012:MyTopic.\\nTo confirm the subscription, visit the SubscribeURL included in this message.",\n    "SubscribeURL": "https://sns.us-west-2.amazonaws.com/?Action=ConfirmSubscription&TopicArn=arn:aws:sns:us-west-2:123456789012:MyTopic&Token=2336412f37fb687f5d51e6e241d09c805a5a57b30d712f794cc5f6a988666d92768dd60a747ba6f3beb71854e285d6ad02428b09ceece29417f1f02d609c582afbacc99c583a916b9981dd2728f4ae6fdb82efd087cc3b7849e05798d2d2785c03b0879594eeac82c01f235d0e717736",\n    "Timestamp": "2012-04-26T20:45:04.751Z",\n    "SignatureVersion": "1",\n    "Signature": "EXAMPLEpH+DcEwjAPg8O9mY8dReBSwksfg2S7WKQcikcNKWLQjwu6A4VbeS0QHVCkhRS7fUQvi2egU3N858fiTDN6bkkOxYDVrY0Ad8L10Hs3zH81mtnPk5uvvolIC1CXGu43obcgFxeL3khZl8IKvO61GWB6jI9b5+gLPoBc1Q=",\n    "SigningCertURL": "https://sns.us-west-2.amazonaws.com/SimpleNotificationService-f3ecfb7224c7233fe7bb5f59f96de52f.pem"\n}'  # noqa
```
imdadahad: Really not a fan of this, can we not pack this in a good looking json?

app/notifications/notifications_ses_callback.py
```diff
@@ -33,6 +34,11 @@ def process_ses_response():
 
         if 'Type' in ses_request and ses_request['Type'] == 'SubscriptionConfirmation':
             current_app.logger.info("SNS subscription confirmation url: {}".format(ses_request['SubscribeURL']))
```
imdadahad: Think you can do `ses_request.get('Type') == 'SubscriptionConfirmation'`

app/notifications/notifications_ses_callback.py
```diff
@@ -33,6 +34,11 @@ def process_ses_response():
 
         if 'Type' in ses_request and ses_request['Type'] == 'SubscriptionConfirmation':
```
imdadahad: Think you can do ses_request.get('Type') == 'SubscriptionConfirmation' here?

app/notifications/utils.py
```diff
@@ -0,0 +1,8 @@
+import requests
+
+
+def confirm_subscription(confirmation_request):
+    url = confirmation_request['SubscribeURL']
```
imdadahad: What happens if `confirmation_request['SubscribeURL']` doesn't exist?

app/notifications/utils.py
```diff
@@ -0,0 +1,8 @@
...
+    response = requests.get(url)
+    response.raise_for_status()
+    return confirmation_request['TopicArn']
```
imdadahad: Might be worth adding some logging here.

tests/app/notifications/test_notifications_ses_callback.py
servingUpAces: This is the response we get from ses.

sakisv: Not a big fan of this either - it messes my editor 😅 

We could extract all of these to separate json files and `json.dumps` them?

app/notifications/utils.py
sakisv: We log it on the [caller](https://github.com/alphagov/notifications-api/pull/920/files#diff-ff083df1227c89e7812cac1d7c48bb4aR38)

imdadahad: Think that assumes it was successful, I was mostly referring to `response` which may raise. We might want to know what error occurred here.

sakisv: Ah, I see. Yes, good point 👍

tests/app/notifications/test_notifications_ses_callback.py
imdadahad: Think it's okay.

PR #921, opened 2017-04-26
Add intl notification fields

imdadahad:
This adds three new fields the `Notification` model that will be used to differentiate between domestic and international notifications.

All fields are nullable for now but a default has been set so that any new notification created with have `international` set to `False`.

**Note: Will rebase of master once #918 is in.**


minglis:
This is OK when rebased/merged/tests green etc:

👍

PR #923, opened 2017-04-26
Validate International phone numbers

minglis:
- uses new utils methods to validate phone numbers
- defaults to International=True on validation. This ensures the validator works on all numbers
- Then check if the user can send this message to the number internationally if needed.

### DIFF
--- a/app/notifications/rest.py
+++ b/app/notifications/rest.py
@@ -2,8 +2,7 @@ from flask import (
     Blueprint,
     jsonify,
     request,
-    current_app,
-    json
+    current_app
 )
 
 from app import api_user
@@ -14,10 +13,6 @@ from app.dao import (
 )
 from app.models import KEY_TYPE_TEAM, PRIORITY
 from app.models import SMS_TYPE
-from app.notifications.process_client_response import (
-    validate_callback_data,
-    process_sms_client_response
-)
 from app.notifications.process_notifications import (persist_notification,
                                                      send_notification_to_queue,
                                                      simulated_recipient)
@@ -34,6 +29,8 @@ from app.schemas import (
 )
 from app.service.utils import service_allowed_to_send_to
 from app.utils import pagination_links, get_template_instance
+
+from notifications_utils.recipients import get_international_phone_info
 
 notifications = Blueprint('notifications', __name__)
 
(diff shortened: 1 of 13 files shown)


tests/app/notifications/rest/test_send_notification.py
```diff
@@ -1046,3 +1046,81 @@ def test_send_notification_uses_priority_queue_when_template_is_marked_as_priori
 
     assert response.status_code == 201
     mocked.assert_called_once_with([notification_id], queue='priority')
+
+
+def test_should_allow_store_original_number_on_sms_notification(notify_api, sample_template, mocker):
+    with notify_api.test_request_context():
+        with notify_api.test_client() as client:
```
imdadahad: You can use the `client` fixture here?

tests/app/notifications/rest/test_send_notification.py
```diff
@@ -1046,3 +1046,81 @@ def test_send_notification_uses_priority_queue_when_template_is_marked_as_priori
...
+            assert response.status_code == 201
+            assert notification_id
+            notifications = Notification.query.all()
+            assert len(notifications) == 1
+            assert '+(44) 7700-900 855' == notifications[0].to
+
+
+def test_should_not_allow_international_number_on_sms_notification(notify_api, sample_template, mocker):
+    with notify_api.test_request_context():
+        with notify_api.test_client() as client:
```
imdadahad: Same as above

app/notifications/rest.py
```diff
@@ -160,6 +161,16 @@ def get_notification_return_data(notification_id, notification, template):
     return output
 
 
+def _service_can_send_internationally(service, number):
```
minglis: This response code matches that used with team key errors. Interesting to make it a 403 - ie forbidden to do the action, not the request was bad.

Think this is better (403) will put a story in for it.

tests/app/notifications/rest/test_send_notification.py
minglis: and this

tests/app/notifications/rest/test_send_notification.py
minglis: Good call - done this

PR #927, opened 2017-04-27
Persist international notification fields

imdadahad:
This will persist the notification with the international fields to the database. This calls additional validation/formatting in the persist_notification method, given that it will be passed the original recipient entry.

Also, we currently do no validation on send user sms code. For now we rely on the admin validation that only allows UK phone numbers. We'll need to think about supporting verification codes for international numbers in a separate story.


minglis:
👍 - on tests pass

PR #930, opened 2017-04-27
Send to provider for international

imdadahad:
This adds the last piece of the puzzle to _actually_ send international notifications to the providers. 

Additionally this also ensures we do not update international notifications with a status.

This depends on:

- [x] https://github.com/alphagov/notifications-admin/pull/1254


minglis:
@leohemsted this story https://www.pivotaltracker.com/story/show/144348993 is to do the logging you requested.

PR #938, opened 2017-05-03
Update the valid_from date for the rate

servingUpAces:
Update the valid_from date for the rate that is intended to start at the beginning of the financial year.

It was the start of the financial year in BST, needed to convert it to UTC.
Small change to the logic to find the rates.


tests/app/dao/test_notification_usage_dao.py
```diff
@@ -9,14 +9,26 @@
 
 
 def test_get_rates_for_year(notify_db, notify_db_session):
+    set_up_rate(notify_db, datetime(2016, 5, 18), 0.016)
+    set_up_rate(notify_db, datetime(2017, 3, 31, 23), 0.0158)
+    start_date, end_date = get_financial_year(2017)
+    print(start_date)
```
kentsanggds: Don't need the print here

PR #939, opened 2017-05-04
Process SNS request triggered by a DVLA S3 update

imdadahad:
This processes the SNS request that is triggered when DVLA return a response file (S3). This will call a task to process the file if the file is in the correct format as we expect.

## Summary 

DVLA will return a response file containing the list of notification references along with their status (Success/Failed). We've setup SNS to trigger a notification which will be sent to our endpoint. 

This pulls the file from S3 and then prepares it for processing. Currently this doesn't actually do anything with each notification to update. We can only do this once we've setup our migration to move away from status enums.

## SNS Response types

We will receive two possible events from SNS, `Notification` for an S3 update or `SubscriptionConfirmation`. Both will have the `Type`, `MessageId` and `Message` hence their inclusion in the schema. If the Type isn't a `SubscriptionConfirmation`, we can safely assume it's a notification indicating an object has been added to S3 by DVLA.

## Additional Notes

1) Added a decorator to do JSON schema validation `validate_schema`, it's in the same file for now but can be moved out if/when used elsewhere.
2) Autoconfirm now added to the endpoint

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -363,7 +363,8 @@ def get_template_class(template_type):
 @notify_celery.task(bind=True, name='update-letter-notifications-statuses')
 @statsd(namespace="tasks")
 def update_letter_notifications_statuses(self, filename):
-    response_file = s3.get_s3_object('development-notifications-csv-upload', filename).decode('utf-8')
+    bucket_location = '{}-ftp'.format(current_app.config['NOTIFY_EMAIL_DOMAIN'])
+    response_file = s3.get_s3_object(bucket_location, filename).decode('utf-8')
     lines = response_file.splitlines()
     notification_updates = []
 
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -380,7 +380,7 @@ def update_letter_notifications_statuses(self, filename):
     else:
         if notification_updates:
             for update in notification_updates:
-                current_app.logger.error(str(update))
+                current_app.logger.info('DVLA update: {}'.format(str(update)))
                 # TODO: Update notifications with desired status
             return notification_updates
         else:
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -365,16 +365,13 @@ def get_template_class(template_type):
 def update_letter_notifications_statuses(self, filename):
     bucket_location = '{}-ftp'.format(current_app.config['NOTIFY_EMAIL_DOMAIN'])
     response_file = s3.get_s3_object(bucket_location, filename).decode('utf-8')
-    lines = response_file.splitlines()
-    notification_updates = []
 
     try:
         NotificationUpdate = namedtuple('NotificationUpdate', ['reference', 'status', 'page_count', 'cost_threshold'])
-        for line in lines:
-            notification_updates.append(NotificationUpdate(*line.split('|')))
+        notification_updates = [NotificationUpdate(*line.split('|')) for line in response_file.splitlines()]
 
     except TypeError:
-        current_app.logger.exception('DVLA response file has an invalid format')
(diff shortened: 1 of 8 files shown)


sakisv:
Could we have the [autoconfirm](https://github.com/alphagov/notifications-api/blob/master/app/notifications/notifications_ses_callback.py#L37-L39) for this endpoint too? 

(If we start having more needs for this it would be worth exploring a nicer integration so that we don't need to do it explicitly every time)

tests/app/celery/test_tasks.py
```diff
@@ -1071,3 +1076,38 @@ def test_update_dvla_job_to_error(sample_letter_template, sample_letter_job):
...
+
+    assert updates[0].reference == 'ref-foo'
+    assert updates[0].status == 'Sent'
+    assert updates[0].page_count == '1'
+    assert updates[0].cost_threshold == 'Unsorted'
+
+    assert updates[1].reference == 'ref-bar'
+    assert updates[1].status == 'Sent'
+    assert updates[1].page_count == '2'
+    assert updates[1].cost_threshold == 'Sorted'
```
servingUpAces: Should we assert the mock is called?

imdadahad: I reserved that for `test_update_letter_notifications_statuses_calls_with_correct_bucket_location`. I didn't want to duplicate the same call as here I specifically wanted to test that the `update_letter_notifications_statuses` parses and returns the expected list. Thoughts?

imdadahad:
@sakisv I've added the autoconfirm, separated into its own method for now...

app/celery/tasks.py
```diff
@@ -354,3 +358,30 @@ def get_template_class(template_type):
...
+def update_letter_notifications_statuses(self, filename):
+    bucket_location = '{}-ftp'.format(current_app.config['NOTIFY_EMAIL_DOMAIN'])
+    response_file = s3.get_s3_object(bucket_location, filename).decode('utf-8')
+    lines = response_file.splitlines()
+    notification_updates = []
+
+    try:
+        NotificationUpdate = namedtuple('NotificationUpdate', ['reference', 'status', 'page_count', 'cost_threshold'])
+        for line in lines:
+            notification_updates.append(NotificationUpdate(*line.split('|')))
```
imdadahad: Noice!

app/celery/tasks.py
```diff
@@ -354,3 +358,30 @@ def get_template_class(template_type):
...
+
+    except TypeError:
+        current_app.logger.exception('DVLA response file has an invalid format')
```
imdadahad: Good shout

app/celery/tasks.py
```diff
@@ -354,3 +358,30 @@ def get_template_class(template_type):
...
+        raise
```
imdadahad: As discussed, it's fine as is.

tests/app/celery/test_tasks.py
```diff
@@ -1071,3 +1076,38 @@ def test_update_dvla_job_to_error(sample_letter_template, sample_letter_job):
...
+def test_update_letter_notifications_statuses_raises_for_invalid_format(notify_api, mocker):
+    invalid_file = b'ref-foo|Sent|1|Unsorted\nref-bar|Sent|2'
+    mocker.patch('app.celery.tasks.s3.get_s3_object', return_value=invalid_file)
+
+    with pytest.raises(TypeError):
+        update_letter_notifications_statuses(filename='foo.txt')
+
+
+def test_update_letter_notifications_statuses_calls_with_correct_bucket_location(notify_api, mocker):
+    invalid_file = b'ref-foo|Sent|1|Unsorted\nref-bar|Sent|2'
```
imdadahad: 👍

PR #940, opened 2017-05-05
Search for notification by number or email

servingUpAces:
New endpoint to search for notifications given an email address or phone number. 
The query ignores case and spaces.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -456,3 +456,10 @@ def dao_update_notifications_sent_to_dvla(job_id, provider):
         {'status': NOTIFICATION_SENDING, "sent_by": provider, "sent_at": now, "updated_at": now})
 
     return updated_count
+
+
+@statsd(namespace="dao")
+def dao_get_notifications_by_to_field(service_id, search_term):
+    return Notification.query.filter(
+        Notification.service_id == service_id,
+        func.replace(func.lower(Notification.to), " ", "") == search_term.lower().replace(" ", "")).all()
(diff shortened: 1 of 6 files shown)


app/dao/notifications_dao.py
```diff
@@ -456,3 +456,10 @@ def dao_update_notifications_sent_to_dvla(job_id, provider):
         {'status': NOTIFICATION_SENDING, "sent_by": provider, "sent_at": now, "updated_at": now})
 
     return updated_count
+
+
+@statsd(namespace="dao")
+def dao_get_notifications_by_to_field(service_id, search_term):
+    return Notification.query.filter(
+        Notification.service_id == service_id,
+        func.replace(func.lower(Notification.to), " ", "") == search_term.lower().replace(" ", "")).all()
```
quis: What happens when I search for `07900900123`? Or `+1 (555)-000-000`?

servingUpAces: I'll need to add the validate_and_format_recipient method to the request. Good spot. thanks

app/service/rest.py
```diff
@@ -271,6 +271,13 @@ def get_all_notifications_for_service(service_id):
     ), 200
 
 
+@service_blueprint.route('/<uuid:service_id>/notification/<search_term>', methods=['GET'])
```
quis: Should be a query string not URL parameter, eg `?to=`

quis: And should be /notification**s**

app/dao/notifications_dao.py
servingUpAces: We've decided to leave the first iteration of the search to strictly match the result. The only thing it does is remove all spaces and ignore case.

PR #941, opened 2017-05-05
Load service on auth

minglis:
Optimise auth code to store service on the request context on a successful auth.

this removes the need to fetch the service on every subsequent controller method.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -1,7 +1,7 @@
 from flask import request, jsonify, current_app
 from sqlalchemy.orm.exc import NoResultFound
 
-from app import api_user
+from app import api_user, authenticated_service
 from app.dao import services_dao, templates_dao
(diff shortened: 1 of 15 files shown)


app/v2/notifications/post_notifications.py
```diff
@@ -60,23 +58,32 @@ def post_notification(notification_type):
...
+            notification=notification,
+            research_mode=authenticated_service.research_mode,
+            queue=queue_name
+        )
+
     else:
         current_app.logger.info("POST simulated notification for id: {}".format(notification.id))
     if notification_type == SMS_TYPE:
-        sms_sender = service.sms_sender if service.sms_sender else current_app.config.get('FROM_NUMBER')
+        sms_sender = (authenticated_service.sms_sender
```
imdadahad: `authenticated_service.sms_sender or current_app.config.get('FROM_NUMBER')` ?

PR #942, opened 2017-05-09
notification status as table

leohemsted:
Adds a new column, `notification_status`, to the notification object. I know the name's a bit confusing, but I didn't want it to have some meta name that referenced the fact it's from a table.

This PR sets up the status table, and the new column as a nullable, foreign keyed table.

I've renamed the cols on the model to `_status_enum` and `_status_fkey` - they should not be interacted with directly. `Notification.status` now refers to a property that *reads from enum* and *updates both* - this way we immediately start populating new data without breaking old data. Then we can look at how to transfer old data across

### DIFF
--- a/migrations/versions/0081_noti_status_as_enum.py
+++ b/migrations/versions/0081_noti_status_as_enum.py
@@ -15,9 +15,24 @@ import sqlalchemy as sa
 
 
 def upgrade():
-    op.create_table('notification_status',
+    status_table = op.create_table('notification_status_types',
         sa.Column('name', sa.String(), nullable=False),
         sa.PrimaryKeyConstraint('name')
+    )
+    op.bulk_insert(status_table,
+        [
+            {'name': x} for x in {
+                'created',
+                'sending',
+                'delivered',
+                'pending',
+                'failed',
+                'technical-failure',
+                'temporary-failure',
+                'permanent-failure',
+                'sent',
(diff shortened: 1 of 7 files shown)


migrations/versions/0081_noti_status_as_enum.py
```diff
@@ -0,0 +1,33 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.create_table('notification_status',
+        sa.Column('name', sa.String(), nullable=False),
+        sa.PrimaryKeyConstraint('name')
+    )
+
```
leohemsted: i need to add the status values here

leohemsted:
Reason we're creating a new column is because `ALTER table ALTER COLUMN` commands are generally slow on columns that area already populated. On a table with 1.2m rows, to run `ALTER TABLE notifications ALTER COLUMN status SET DATA TYPE varchar` takes 30 seconds. We saw it take 2-3 minutes on live when we previously did this, and during this it acquires an `EXCLUSIVE LOCK`
> (This mode guarantees that the holder is the only transaction accessing the table in any way.)

Surprisingly, it's pretty quick (1.5 seconds) to add the foreign key constraint, but by that point the damage has been done, any selects, updates or inserts to the notification history table will have timed out.

PR #944, opened 2017-05-10
Update notification international

servingUpAces:
Update the script to set the international flag False where the value is null in the  notifications and notification_history, these updates are done in separate loops.

It takes about 1.5 minutes to update 27,000 notifications and 27,000 notification_history. 

The update is a row level lock so will only affect updates to the same row.
This is unlikely as the data being updated should be older than 3 days.
The second scripts updates the table to set international as not null, to make the model.

### DIFF
--- a/migrations/versions/0082_set_international.py
+++ b/migrations/versions/0082_set_international.py
@@ -1,39 +1,44 @@
 """empty message
 
 Revision ID: 0082_set_international
-Revises: 0080_fix_rate_start_date
+Revises: 0081_noti_status_as_enum
 Create Date: 2017-05-05 15:26:34.621670
 
 """
+from datetime import datetime
+from alembic import op
 
 # revision identifiers, used by Alembic.
-from datetime import datetime
-
 revision = '0082_set_international'
-down_revision = '0080_fix_rate_start_date'
-
-from alembic import op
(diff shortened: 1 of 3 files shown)


migrations/versions/0082_set_international.py
```diff
@@ -0,0 +1,45 @@
...
+def upgrade():
+    conn = op.get_bind()
+    start = datetime.utcnow()
+    notification_history = "select id from notification_history where international is null limit 10000"
+
+    results = conn.execute(notification_history)
+    res = results.fetchall()
+
+    while len(res) > 0:
+        conn.execute("update notification_history set international = False where id in ({})".format(
```
minglis: Question. Is it worth having some sort of sleep in here to pause between each 10,000 rows?

servingUpAces: For what reason?

servingUpAces:
It takes less than 3 seconds to update 10,000 rows.
It took 2.5 minutes to update 276,843 notification and 277,858 notification_history rows.
I won't be able to do a proper test until I preview or staging where there is a lot more data. 
Since the international flag has been set for 14 days none of the rows being updated should be getting updated themselves, so the impact should be minimal.

PR #945, opened 2017-05-10
Add go live template (Migration)

imdadahad:
This adds a template in a migration that will be sent automatically when a service goes live. 

_Note: This will also include some personal information that'll be added separately._

### DIFF
--- a/migrations/versions/0082_add_golive_template.py
+++ b/migrations/versions/0082_add_golive_template.py
@@ -1,0 +1,119 @@
+"""empty message
+
+Revision ID: 0082_add_go_live_template
+Revises: 0081_noti_status_as_enum
+Create Date: 2017-05-10 16:06:04.070874
+
+"""
+
+# revision identifiers, used by Alembic.
+from datetime import datetime
+
+from flask import current_app
+
+from alembic import op
+import sqlalchemy as sa
+
+revision = '0082_add_go_live_template'
+down_revision = '0081_noti_status_as_enum'
+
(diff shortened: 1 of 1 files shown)


migrations/versions/0082_add_golive_template.py
```diff
@@ -0,0 +1,116 @@
...
+^Only use this email address for out of hours emergencies. Don’t share this address with people outside of your team.\n\n
+
+We’ll get back to you within 30 minutes and give you hourly updates until the problem’s fixed.\n\n
+
+For non-emergency problems or questions, use our support page and we’ll reply in office hours:
+https://www.notifications.service.gov.uk/support \n\n
+#Escalation for emergency problems\n\n
+
+If we haven’t acknowledged an emergency problem you’ve reported within 30 minutes and you need to know what’s happening, you can escalate to:\n\n
+
```
quis: Something missing here?

quis:
I don’t think you need all the `\n`s if you’re `"""`triple quoting`"""` stuff.

migrations/versions/0082_add_golive_template.py
imdadahad: Ah yes, sorry forgot to update the description.

I'm planning on inserting personal content (phone number etc.) manually. I didn't want those details to be in git history.

quis: Good point. Didn’t think about that.

imdadahad:
Thanks, just removed them.

PR #948, opened 2017-05-11
Async job stats

minglis:
First pass at building offline, aggregate statistics.

Doing jobs first as these are:
(a) Source of lag on the service dashboard page
(b) low impact as not used for billing etc

Two new tasks created to handle counting the job data. 
- record initial send
- record outcome

These increment counters against the job for each notification type, three columns
1) Sent (incremented when we send a message)
2) Failed (incremented on a failed callback)
3) Delivered (incremented on a success callback)

These DAO methods do a sort of hand rolled upsert:
- attempt to update
- if no rows updated attempt to insert
- if insert fails due to constraint (ie row already exists) then try update again

The tasks are only created if the notification is in the correct state:
- create initial task IF NOT IN COMPLETED STATE
- create outcome task IF IN COMPLETED STATE

Also times out the job counts in the same way we do for notifications: bump the failed count to marry up sent with (delivered + failed) after the timeout window.

This story is missing 1 thing:
1) Proper treatment of letter jobs (not sure what the states are and how they map)


app/dao/statistics_dao.py
```diff
@@ -0,0 +1,115 @@
...
+    SMS_TYPE,
+    LETTER_TYPE,
+    NOTIFICATION_STATUS_TYPES_FAILED,
+    NOTIFICATION_DELIVERED,
+    NOTIFICATION_SENT)
+from app.statsd_decorators import statsd
+
+
+@transactional
+def timeout_job_counts(notifications_type, timeout_start):
```
servingUpAces: There are no tests for this update.

app/celery/scheduled_tasks.py
```diff
@@ -180,3 +181,12 @@ def switch_current_sms_provider_on_slow_delivery():
             )
 
             dao_toggle_sms_provider(current_provider.identifier)
+
+
+@notify_celery.task(name='timeout-job-statistics')
+@statsd(namespace="tasks")
+def timeout_job_statistics():
+    updated = dao_timeout_job_statistics(current_app.config.get('SENDING_NOTIFICATIONS_TIMEOUT_PERIOD'))
```
servingUpAces: Why assume that the status of the unaccounted for notifications are failed. Could we do a count of the status of the notifications here instead?

app/dao/statistics_dao.py
minglis: there are, just there was some aliasing going on. The aliasing is removed so tests should be easier to find.

app/celery/scheduled_tasks.py
minglis: Done

migrations/versions/0084_add_created_to_job_stats.py
```diff
@@ -0,0 +1,22 @@
...
+# revision identifiers, used by Alembic.
+revision = '0084_add_created_to_job_stats'
+down_revision = '0083_add_job_stats'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('job_statistics', sa.Column('created_at', sa.DateTime(), nullable=True))
```
minglis: it was added as later in the story as needed for the delete task. At the time I was wondering whether it should be a separate PR so it stayed a separate script. Didn't think it mattered too much.

minglis: rebuilt migrations to have single script

app/dao/statistics_dao.py
```diff
@@ -0,0 +1,147 @@
...
+
+    results = db.session.query(
+        JobStatistics.job_id.label('job_id'),
+        func.count(Notification.status).label('count'),
+        Notification.status.label('status')
+    ).filter(
+        JobStatistics.job_id == Notification.job_id,
+        JobStatistics.created_at < timeout_start,
+        sent != failed + delivered
+    ).group_by(Notification.status, JobStatistics.job_id).order_by(JobStatistics.job_id).all()
```
servingUpAces: Why order by job_id? There is only one job_id in the results.

app/dao/statistics_dao.py
```diff
@@ -0,0 +1,147 @@
...
+        groups.append(list(g))
+
+    for job in groups:
+        sent_count = 0
+        delivered_count = 0
+        failed_count = 0
+        for notification_status in job:
+            if notification_status.status in [NOTIFICATION_DELIVERED, NOTIFICATION_SENT]:
+                delivered_count += notification_status.count
+            else:
```
servingUpAces: Is it possible that there is a pending status in the list?

app/dao/statistics_dao.py
minglis: no there is multiple jobs in this query.

app/dao/statistics_dao.py
minglis: pending is marked as failed as it'd be timed out anyway.

app/dao/statistics_dao.py
servingUpAces: Right sorry - that was silly of me

app/dao/statistics_dao.py
```diff
@@ -0,0 +1,147 @@
...
+    failed = columns(notifications_type, 'failed')
+
+    results = db.session.query(
+        JobStatistics.job_id.label('job_id'),
+        func.count(Notification.status).label('count'),
+        Notification.status.label('status')
+    ).filter(
+        JobStatistics.job_id == Notification.job_id,
+        JobStatistics.created_at < timeout_start,
+        sent != failed + delivered
```
servingUpAces: Could add a Notification.type filter since you are passing in a notification type.

app/dao/statistics_dao.py
```diff
@@ -0,0 +1,147 @@
...
+    groups = []
+    for k, g in groupby(sort, key=lambda result: result.job_id):
+        groups.append(list(g))
+
+    for job in groups:
+        sent_count = 0
+        delivered_count = 0
+        failed_count = 0
+        for notification_status in job:
+            if notification_status.status in [NOTIFICATION_DELIVERED, NOTIFICATION_SENT]:
```
servingUpAces: You can use your new NOTIFICATIONS_STATUS_SUCCESS list.

minglis: Done

app/dao/statistics_dao.py
minglis: yep done - doesn't change outcome but better.

(As columns in the stats table also do type restriction)

PR #951, opened 2017-05-12
Send a welcome email to newly live services

imdadahad:
This sends an email to all active users of a service when we make them live.

## Summary

This sends the email (added in #945 ) to active users of the service when we update them to live. The `send_notification_to_service_users` method can be used to in future should we want to send other types of notifications later on. 

## Testing

You can test this locally by making your service live. You should receive an email (make sure you're not in research mode).


tests/app/service/test_sender.py
```diff
@@ -0,0 +1,110 @@
...
+    create_notify_service(notify_db, notify_db_session)
+    service = create_sample_service(notify_db, notify_db_session, user=sample_user)
+    template = create_template(service, template_type=EMAIL_TYPE)
+    send_notification_to_service_users(
+        service_id=service.id,
+        template_id=template.id,
+        include_user_fields=['name', 'email_address', 'state']
+    )
+
+    assert persist_mock.mock_calls[0][2]['personalisation'] == {
```
kentsanggds: Not sure what [0][2] means, can you make this clearer

tests/app/dao/test_services_dao.py
```diff
@@ -783,3 +785,11 @@ def test_fetch_monthly_historical_template_stats_for_service_separates_templates
     assert len(result.get('2016-04').keys()) == 2
     assert str(template_one.id) in result.get('2016-04').keys()
     assert str(template_two.id) in result.get('2016-04').keys()
+
+
+def test_dao_fetch_active_users_for_service_returns_active_only(sample_service):
```
kentsanggds: perhaps change sample_service to create_service to explicitly show that we are creating 1 active user as well as the pending user

PR #955, opened 2017-05-15
Add service permissions model + migration script

kentsanggds:
Added service permissions type model and service permissions mapping table.

Also refactored some of the code in models.py to reuse constants.


app/models.py
```diff
@@ -183,6 +195,29 @@ class Service(db.Model, Versioned):
...
+class ServicePermissionTypes(db.Model):
+    __tablename__ = 'service_permission_types'
+
+    name = db.Column(db.String(255), primary_key=True)
+
+
+class ServicePermission(db.Model):
+    __tablename__ = "service_permissions"
+
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'),
```
servingUpAces: We should add the relationship in the model as well.

PR #956, opened 2017-05-15
Add service permissions DAO and refactor user service permission mock

kentsanggds:
Created service permissions DAO code to handle - 

- fetching of all service permissions for a given service id
- creating a service permission for a given service id
- removing a service permission for a given service id

and created tests to ensure that the DAO is working as expected.

Also had to rename sample_service_permission to sample_user_service_permission in order to avoid confusion with mocks associated with service_permissions


app/dao/service_permissions_dao.py
```diff
@@ -0,0 +1,26 @@
+from sqlalchemy import exc
+
+from app import db
+from app.dao.dao_utils import transactional
+from app.models import Service, ServicePermission, SERVICE_PERMISSION_TYPES
```
imdadahad: `exc` and `Service` are unused

app/dao/service_permissions_dao.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def dao_fetch_service_permissions(service_id):
+    return ServicePermission.query.filter(
+        ServicePermission.service_id == service_id).all()
+
+
+@transactional
+def dao_add_and_commit_service_permission(service_id, permission):
```
imdadahad: Can we rename this to `dao_create_service_permission`? It's implicit in that we'll add and commit a service permission.

app/dao/service_permissions_dao.py
```diff
@@ -0,0 +1,26 @@
...
+    if permission not in SERVICE_PERMISSION_TYPES:
+        raise ValueError("'{}' not of service permission type: {}".format(permission, SERVICE_PERMISSION_TYPES))
+
```
imdadahad: I think this logic shouldn't be in the DAO method as DAO tends to be database only. Any validation etc. should be at a higher layer, thoughts?

app/dao/service_permissions_dao.py
```diff
@@ -0,0 +1,26 @@
...
+    service_permission = ServicePermission(service_id=service_id, permission=permission)
+
+    db.session.add(service_permission)
+
+
+def dao_remove_service_permission(service_id, permission=None):
+    return ServicePermission.query.filter(
```
imdadahad: Sort of related to above comment about DAO being strictly DB and any logic should be elsewhere. Applying that here think we should only call this method if we know at this point that `permission` exists and we want to remove it. So think we should make `permission` a required param?

tests/app/conftest.py
```diff
@@ -666,11 +666,11 @@ def sample_permission(notify_db,
 
 
 @pytest.fixture(scope='function')
-def sample_service_permission(notify_db,
-                              notify_db_session,
-                              service=None,
-                              user=None,
-                              permission="manage_settings"):
+def sample_user_service_permission(notify_db,
+                                   notify_db_session,
```
imdadahad: Think we have some spaces here, can be removed.

tests/app/dao/test_service_permissions_dao.py
```diff
@@ -0,0 +1,55 @@
...
+    for spt in service_permission_types:
+        create_service_permission(service_id=sample_service.id, permission=spt)
+    service_permissions = dao_fetch_service_permissions(sample_service.id)
+
+    assert len(service_permissions) == len(service_permission_types)
+    assert all(sp.service_id == sample_service.id for sp in service_permissions)
+    assert all(sp.permission in service_permission_types for sp in service_permissions)
+
+
+def test_create_invalid_service_permissions_raises_error(sample_service):
```
imdadahad: `def test_create_invalid_service_permission_raises_error` small typo on permissions.

tests/app/dao/test_service_permissions_dao.py
```diff
@@ -0,0 +1,55 @@
+import pytest
+
+from app.dao.service_permissions_dao import (
+    dao_fetch_service_permissions, dao_remove_service_permission)
+from app.models import (
+    EMAIL_TYPE, SMS_TYPE, LETTER_TYPE, INTERNATIONAL_SMS_TYPE, INCOMING_SMS_TYPE, SERVICE_PERMISSION_TYPES)
```
imdadahad: `create_service` and `INTERNATIONAL_SMS_TYPE` are unused here.

tests/app/dao/test_service_permissions_dao.py
```diff
@@ -0,0 +1,55 @@
...
+    assert len(service_permission) == 1
+    assert all(sp.service_id == sample_service.id for sp in service_permission)
+    assert all(sp.permission in service_permission_type for sp in service_permission)
+
+
+def test_fetch_service_permissions_gets_service_permissions(sample_service):
+    service_permission_types = [LETTER_TYPE, EMAIL_TYPE, SMS_TYPE]
+    for spt in service_permission_types:
+        create_service_permission(service_id=sample_service.id, permission=spt)
+    service_permissions = dao_fetch_service_permissions(sample_service.id)
```
imdadahad: Generally better to avoid for loops in test as its less explicit. Think here it'd be more readable/debuggable to do:

```
create_service_permission(service_id=sample_service.id, permission=LETTER_TYPE)
create_service_permission(service_id=sample_service.id, permission=EMAIL_TYPE)
create_service_permission(service_id=sample_service.id, permission=SMS_TYPE)
```

tests/app/dao/test_service_permissions_dao.py
```diff
@@ -0,0 +1,55 @@
...
+
+    assert "'invalid' not of service permission type: {}".format(str(SERVICE_PERMISSION_TYPES)) in str(e.value)
+
+
+def test_remove_service_permission(sample_service):
+    service_permission_types_to_create = [EMAIL_TYPE, INCOMING_SMS_TYPE]
+    service_permission_type_to_remove = EMAIL_TYPE
+    service_permission_type_remaining = INCOMING_SMS_TYPE
+
+    for spt in service_permission_types_to_create:
```
imdadahad: Same as above with regards to avoiding for loops in test

tests/app/service/test_rest.py
```diff
@@ -961,13 +961,13 @@ def test_remove_user_from_service(notify_api, notify_db, notify_db_session, samp
             assert resp.status_code == 204
 
 
-def test_remove_user_from_service(notify_api, notify_db, notify_db_session, sample_service_permission):
+def test_remove_user_from_service(notify_api, notify_db, notify_db_session, sample_user_service_permission):
```
imdadahad: Why do this test have the same name as the one above?

tests/app/service/test_rest.py
```diff
@@ -941,12 +941,12 @@ def test_add_unknown_user_to_service_returns404(notify_api, notify_db, notify_db
             assert result['message'] == expected_message
 
 
-def test_remove_user_from_service(notify_api, notify_db, notify_db_session, sample_service_permission):
+def test_remove_user_from_service(notify_api, notify_db, notify_db_session, sample_user_service_permission):
```
imdadahad: You can also use the `client` fixture here to get rid of all the `with` statements.

app/dao/service_permissions_dao.py
kentsanggds: ok

app/dao/service_permissions_dao.py
kentsanggds: ok

app/dao/service_permissions_dao.py
kentsanggds: ok, will assume that it's handled by the json schema

app/dao/service_permissions_dao.py
kentsanggds: ok, since we probably don't need to remove all permissions from a service

tests/app/dao/test_service_permissions_dao.py
kentsanggds: ok

tests/app/dao/test_service_permissions_dao.py
kentsanggds: ok, will use INTERNATIONAL_SMS_TYPE somewhere and remove create_service

tests/app/dao/test_service_permissions_dao.py
kentsanggds: ok

tests/app/dao/test_service_permissions_dao.py
kentsanggds: ok

PR #957, opened 2017-05-16
Schedule api notification

servingUpAces:
Added an optional scheduled_for attribute to the post_notification form data.  This will allow the sender to state what time they want the notification to be posted. 
Scheduled_for is expected to be in the format: yyyy-mm-dd hh:mi:ss


minglis:
Don't think scheduled for should have hours and minutes. Should just be hours. We shouldn't imply we can send at 10:07 for example.

migrations/versions/0083_scheduled_notifications.py
```diff
@@ -0,0 +1,31 @@
...
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = '0083_scheduled_notifications'
+down_revision = '0083_add_perm_types_and_svc_perm'
+
+
+def upgrade():
+    op.create_table('scheduled_notifications',
```
minglis: Why do we need a new table?

app/celery/scheduled_tasks.py
```diff
@@ -46,6 +47,18 @@ def run_scheduled_jobs():
         raise
 
 
+@notify_celery.task(name='send-scheduled-notifications')
+@statsd(namespace="tasks")
+def send_scheduled_notifications():
+    try:
+        for notification in dao_get_scheduled_notifications():
```
minglis: This could be gnarly - thats an unbounded action - could be thousands of API calls to SQS. Depending on how many things queued for scheduled sending. If this is for an overnight batch job - it could be 100,000 things ready to go.

PR #966, opened 2017-05-19
[DO NOT DEPLOY] Change default from number to be GOVUK

quis:


quis:
Doing this on the DB instead. Intention is to deprecate the `FROM_NUMBER`.

PR #969, opened 2017-05-22
Refactored service permissisons data model

kentsanggds:
Had to refactor the service_permissions data model in order to get the many-to-many relationship handling properly, using an [association_proxy](http://docs.sqlalchemy.org/en/rel_0_9/orm/extensions/associationproxy.html#simplifying-association-objects).

I have also changed the value of 'incoming_sms' to 'inbound_sms' to reflect the language used in the team.


app/models.py
```diff
@@ -226,6 +216,23 @@ def set_permissions(self):
...
+
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'),
+                           primary_key=True, index=True, nullable=False)
+    permission = db.Column(db.String(255), db.ForeignKey('service_permission_types.name'),
+                           index=True, primary_key=True, nullable=False)
+    service = db.relationship("Service", foreign_keys=[service_id])
+    created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow, nullable=False)
+
+    service_permission_types = db.relationship(
+        Service, backref=backref("permissions", cascade="all, delete-orphan"))
```
imdadahad: Why can't you used `db.backref` here?

imdadahad: What does `delete-orphan` do here exactly? Just want to make sure we know the difference to avoid any unexpected behaviour?

app/models.py
```diff
@@ -226,6 +216,23 @@ def set_permissions(self):
...
 
 
+class ServicePermission(db.Model):
+    __tablename__ = "service_permissions"
+
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'),
+                           primary_key=True, index=True, nullable=False)
+    permission = db.Column(db.String(255), db.ForeignKey('service_permission_types.name'),
+                           index=True, primary_key=True, nullable=False)
+    service = db.relationship("Service", foreign_keys=[service_id])
```
imdadahad: Open question, can this be `db.relationship("Service")`? We do this in `TemplateHistory`

migrations/versions/0085_update_incoming_to_inbound.py
```diff
@@ -0,0 +1,28 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('service_permissions', 'updated_at')
+    op.execute("UPDATE service_permission_types SET name='inbound_sms' WHERE name='incoming_sms'")
+    # ### end Alembic commands ###
```
imdadahad: Might be worth getting rid of these comments, nice to keep migrations clean.

app/models.py
kentsanggds: Can do, instead of importing backref

app/models.py
kentsanggds: ok, think this was added as part of original refactoring

migrations/versions/0085_update_incoming_to_inbound.py
kentsanggds: ok

app/models.py
kentsanggds: This setting ensures that when deleting a service, there are no orphaned service permissions, will write a test to check that this is the case

kentsanggds: Wrote the test and it the types table wasn't affected, also checked whether removing the cascade had any impact and it didn't so I will just remove it to keep the code cleaner and reintroduce if necessary.

PR #970, opened 2017-05-22
Create a server command to run custom db scripts.

servingUpAces:
We want a way to do data updates that are long running where we can commit the update after some number of rows and stop the script running if necessary. 
This pull request adds a server command to do this. The expectation is that a custom db script is added as a method for the server command to call.


app/commands.py
```diff
@@ -60,3 +63,19 @@ def run(self, service_name_prefix=None, user_email_prefix=None):
                     else:
                         delete_user_verify_codes(usr)
                         delete_model_user(usr)
+
+
+class CustomDbScript(Command):
+    def run(self):
+        self.update_notification_international_flag()
```
leohemsted: should look at how to give this an argument, so you call `python application.py custom_db_script <name_of_function>` rather than us having to update this the next time we run stuff

imdadahad:
Just adding onto what leo said, would be good - assuming we want to do more of this in the future - to have a separate file for db migrations. That way we could do `python application.py custom_db_script <db_command>` where the `db_command` would be imported from that file.

app/commands.py
```diff
@@ -60,3 +63,29 @@ def run(self, service_name_prefix=None, user_email_prefix=None):
...
+        result = db.session.execute(subq).fetchall()
+        while len(result) > 0:
+            db.session.execute(update)
+            print('commit 250000 updates at {}'.format(datetime.utcnow()))
+            db.session.commit()
+            result = db.session.execute(subq).fetchall()
+
+        # Now update notification_history
+        subq_history = "select id from notification_history where international is null limit 250000"
+        update_history = "update notification_history set international = False where id in ({})".format(subq)
```
imdadahad: `.format(subq)` -> `.format(subq_history)`?

PR #971, opened 2017-05-22
set sms_sender to be 'GOVUK' if not otherwise specified

leohemsted:
this is a precursor to making the column non-nullable

- [ ] https://github.com/alphagov/notifications-admin/pull/1284
- [ ] https://github.com/alphagov/notifications-api/pull/973


leohemsted:
rolling this into #973 (and partially into the upcoming db upgrade)

leohemsted:
rolling this into #973 (and partially into the upcoming db upgrade)

PR #976, opened 2017-05-22
Remove drop updated_at column from db script

kentsanggds:
Updated the script so that it doesn't drop any columns from the service_permissions table


kentsanggds:
Created by accident

PR #979, opened 2017-05-23
Add normalised_to field to notification

imdadahad:
This adds `normalised_to` field to the `Notification` which will store the number that we normalise to when sending notifications.


imdadahad:
@servingUpAces yeah in #981

PR #981, opened 2017-05-23
Persist normalised recipient to `Notification`

imdadahad:
This persists the normalised recipient to the `Notification` for sms and email.

This paves way for searching for notifications (needed for two-way) as we can query against this. 

This was branched off #979 so will be rebased off `master` once that's merged in.

## Dependencies:

- [x] #979

### DIFF
--- a/tests/app/notifications/test_process_notification.py
+++ b/tests/app/notifications/test_process_notification.py
@@ -348,3 +348,38 @@ def test_persist_scheduled_notification(sample_notification):
     assert len(scheduled_notification) == 1
     assert scheduled_notification[0].notification_id == sample_notification.id
     assert scheduled_notification[0].scheduled_for == datetime.datetime(2017, 5, 12, 13, 15)
+
+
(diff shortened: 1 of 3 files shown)


tests/app/notifications/test_process_notification.py
```diff
@@ -339,3 +339,35 @@ def test_persist_notification_with_international_info_does_not_store_for_email(
...
     assert persisted_notification.rate_multiplier is None
+
+
+@pytest.mark.parametrize('recipient, expected_recipient_normalised', [
+    ('7900900123', '447900900123'),
+    ('+447900   900 123', '447900900123'),
+    ('  07700900222', '447700900222'),
+    ('07700900222', '447700900222'),
+    (' 73122345678', '73122345678'),
+    ('360623400400', '360623400400')
```
quis: Can you add a test with brackets and dashes just to make sure we’re stripping them out?

imdadahad: Good point, just added them in.

PR #984, opened 2017-05-24
Ken update api service permissions handling

kentsanggds:
Enables the API to handle both deprecated service permission `can-send` columns in the `services` table and new permissions in `service_permissions` table.

The new json expected by the API in order to update service permissions will be in the form of - 

```
{
  "permissions": ["email", "sms"]
}
```

The existing flags `can_send_letters` and `can_send_international_sms` will still be supported and will update the service_permissions within this PR.

To add a new permission simply be append a valid permission to the list.
To remove permissions remove it from the list.

This will also update the deprecated flags.

Supported permissions -  `["email", "sms", "letter", "international_sms", "inbound_sms"]`

Duplicate permissions will raise errors.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -241,7 +241,8 @@ class ServicePermission(db.Model):
     service = db.relationship("Service")
     created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow, nullable=False)
 
-    service_permission_types = db.relationship(Service, backref=db.backref("permissions"))
+    service_permission_types = db.relationship(
+        Service, backref=db.backref("permissions", cascade="all, delete-orphan"))
 
     def __repr__(self):
         return '<{} has service permission: {}>'.format(self.service_id, self.permission)
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -25,6 +25,7 @@ from notifications_utils.recipients import (
 
 from app import ma
 from app import models
+from app.models import ServicePermission
 from app.dao.permissions_dao import permission_dao
 from app.utils import get_template_instance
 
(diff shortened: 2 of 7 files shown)


kentsanggds:
Can't get it to fail on my machine, will try it on another machine.

servingUpAces:
I downgraded my db then tried again. It works for me now.

app/models.py
```diff
@@ -241,7 +241,8 @@ class ServicePermission(db.Model):
     service = db.relationship("Service")
     created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow, nullable=False)
 
-    service_permission_types = db.relationship(Service, backref=db.backref("permissions"))
+    service_permission_types = db.relationship(
+        Service, backref=db.backref("permissions", cascade="all, delete-orphan"))
```
servingUpAces: We don't use cascade for other models. I am skeptical of having it in the model. I would be more comfortable if we removed it.

app/schemas.py
```diff
@@ -178,25 +180,116 @@ class ServiceSchema(BaseSchema):
...
     @validates('sms_sender')
     def validate_sms_sender(self, value):
         if value and not re.match(r'^[a-zA-Z0-9\s]+$', value):
             raise ValidationError('Only alphanumeric characters allowed')
 
+    @validates('permissions')
+    def validate_permissions(self, value):
+        permissions = [v.permission for v in value]
+        for p in permissions:
+            if p not in models.SERVICE_PERMISSION_TYPES:
```
servingUpAces: Do we need this if the table foreign key constraint will do this for us?

app/schemas.py
```diff
@@ -178,25 +180,116 @@ class ServiceSchema(BaseSchema):
...
+                raise ValidationError("Invalid Service Permission: '{}'".format(p))
+
+        if len(set(permissions)) != len(permissions):
+            duplicates = list(set([x for x in permissions if permissions.count(x) > 1]))
```
servingUpAces: Again don't we have a constraint in the table to error on duplicates?

app/models.py
kentsanggds: Need the (cascade)[http://docs.sqlalchemy.org/en/latest/orm/cascades.html] for it to work with association_proxy, if the concern is that the permission_service_types permission is removed if there are no parents attached to it, then it doesn't do this and I have a test to ensure that this is the case - test_remove_service_does_not_remove_service_permission_types

app/schemas.py
kentsanggds: I thought it would be good to do it here to avoid hitting the database and raise a nicer error

app/schemas.py
kentsanggds: Same thought as above comment

PR #986, opened 2017-05-24
separate service deserialization from validation

leohemsted:
Marshmallow validates and deserialises - BUT, when it deserialises, it explicitly sets 
`sms_sender=None`, even when you haven't passed sms_sender in. This is problematic, because we wanted to take advantage of sqlalchemy's default value to set sms_sender to `GOVUK` when the actual DB commit happens.

Instead, still use marshmallow for validating, but manually carry out the json deserialisation in the model class.

This fixes a bug that only manifested when the database was upgraded, but the code hadn't updated. :tada:

### DIFF
--- a/tests/app/service/test_rest.py
+++ b/tests/app/service/test_rest.py
@@ -20,7 +20,7 @@ from tests.app.conftest import (
     sample_notification_history as create_notification_history,
     sample_notification_with_job
 )
-from app.models import KEY_TYPE_NORMAL, KEY_TYPE_TEAM, KEY_TYPE_TEST
+from app.models import Service, KEY_TYPE_NORMAL, KEY_TYPE_TEAM, KEY_TYPE_TEST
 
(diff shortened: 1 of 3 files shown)


tests/app/service/test_rest.py
```diff
@@ -216,6 +216,10 @@ def test_create_service(client, sample_user):
     assert json_resp['data']['dvla_organisation'] == '001'
     assert json_resp['data']['sms_sender'] == current_app.config['FROM_NUMBER']
 
+    service_db = Service.query.get(json_resp['data']['id'])
+    assert service_db.name == 'created service'
```
minglis: Can we have a test clause somewhere for that marshmallow fakery in the to_json method on the model object? Or is that covered by other tests?

leohemsted: covered by existing tests - it'll die kicking and screaming

PR #991, opened 2017-05-26
Invite the service to schedule notifications from the API

servingUpAces:
Check service.permissions for the existence of schedule_notifications if the notifications is being created with a scheduled_for param.

We still need a way to add the permission from the front end.

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -6,7 +6,7 @@ from notifications_utils.recipients import (
 )
 
 from app.dao import services_dao
-from app.models import KEY_TYPE_TEST, KEY_TYPE_TEAM, SMS_TYPE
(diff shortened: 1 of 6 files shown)


app/notifications/validators.py
```diff
@@ -92,6 +92,7 @@ def check_sms_content_char_count(content_count):
         raise BadRequestError(message=message)
 
 
-def service_can_schedule_notification(service):
-    # TODO: implement once the service permission works.
-    raise BadRequestError(message="Your service must be invited to schedule notifications via the API.")
+def service_can_schedule_notification(service, scheduled_for):
+    if scheduled_for:
+        if SCHEDULE_NOTIFICATIONS not in [p.permission for p in service.permissions]:
+            raise BadRequestError(message="Your service must be invited to schedule notifications via the API.")
```
minglis: Really pedantic, but can we review the error message?

quis: > Cannot schedule notifications (this feature is invite-only)

minglis:
👍

PR #994, opened 2017-05-26
Ken migrate default service permissions

kentsanggds:
This script should migrate existing flags set into the service permissions and set default service permissions (sms, email) for each service. It doesn't affect existing flags.

As each row has a created_at of '2017-05-26 17:30:00.000000' for this migration, the downgrade is to remove all service permissions with created_at '2017-05-26 17:30:00.000000' set.


migrations/versions/0088_migrate_existing_svc_perms.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+# revision identifiers, used by Alembic.
+revision = '0088_migrate_existing_svc_perms'
+down_revision = '0087_scheduled_notifications'
+
+from alembic import op
+import sqlalchemy as sa
+
+migration_date = '2017-05-26 17:30:00.000000'
```
servingUpAces: Do we want this to be the date all the service_permissions are created?

PR #995, opened 2017-05-30
add inbound sms table

leohemsted:
When we receive inbound SMS from MMG, we receive the following info:

* the message contents
* the number that sent it
* the number it was sent to
* when the provider [mmg] processed it
* the provider reference (mmg internal ID)

We were already logging this to kibana, but we should also persist this data - this PR does that.

We have registered a selection of numbers as inbound endpoints with MMG - so we should only receive messages for those to this endpoint, and it tries to find a single service matching that sender - if there is not exactly one service, we should 400 because we don't know which service to tie this random message to.

For now, it will *not* normalise that inbound number, so the sender ID needs to exactly match what MMG send us when the number is texted - i'll have a think about how easy it is to relax this restriction though.


app/models.py
```diff
@@ -1141,3 +1141,26 @@ def __str__(self):
...
+
+class InboundSms(db.Model):
+    __tablename__ = 'inbound_sms'
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), index=True, nullable=False)
+    service = db.relationship('Service', backref='inbound_sms')
+
+    notify_number = db.Column(db.String, nullable=False)  # the service's number, that the msg was sent to
```
minglis: good names

app/notifications/receive_notifications.py
```diff
@@ -1,17 +1,64 @@
...
+        'ID': 'SOME-MMG-SPECIFIC-ID',
+        'DateRecieved': '2017-05-21+11%3A56%3A11'
+    }
+    """
     post_data = request.get_json()
-    post_data.pop('MSISDN', None)
-    current_app.logger.info("Recieve notification form data: {}".format(post_data))
+    potential_services = dao_fetch_services_by_sms_sender(post_data['Number'])
+
+    if len(potential_services) != 1:
```
minglis: Can we put a DB constraint on this number to prevent this?

app/notifications/receive_notifications.py
```diff
@@ -1,17 +1,64 @@
...
+            status_code=400
+        )
+
+    service = potential_services[0]
+
+    inbound = create_inbound_sms_object(service, post_data)
+
+    current_app.logger.info('{} received inbound SMS with reference {}'.format(service.id, inbound.provider_reference))
+
+    return 'RECEIVED', 200
```
minglis: We use no body 201's elsewhere?

app/notifications/receive_notifications.py
leohemsted: no, because multiple services can, and should be able to, have sms_sender == `GOVUK` or `DWP` or similar

app/notifications/receive_notifications.py
leohemsted: When this endpoint was created in https://github.com/alphagov/notifications-api/commit/77f520acba154bddc1903169171e6157faf9b851, becca said "returns the requested RECEIVED", so i wasn't sure if changing this would impact MMG in any way

servingUpAces: MMG had requested that we return RECEIVED

leohemsted: okay, i guess i need to make sure we still return RECEIVED even if the number isn't associated with a service

PR #998, opened 2017-05-30
Did some work around the delete queues script

minglis:
Wrapped it in docopts so we can choose to list or delete queues. Basic functionality the same.

- list generates a csv
- delete reads the csv and deletes each queue in it.

Use with caution.

### DIFF
--- a/scripts/delete_sqs_queues.py
+++ b/scripts/delete_sqs_queues.py
@@ -4,20 +4,20 @@ Script to manage SQS queues. Can list or delete queues.
 
 Uses boto, so relies on correctly set up AWS access keys and tokens.
 
-In principle use this script to dump details of all queues in a gievn environment, and then 
+In principle use this script to dump details of all queues in a gievn environment, and then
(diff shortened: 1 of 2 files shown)


scripts/delete_sqs_queues.py
```diff
@@ -27,26 +50,28 @@ def get_queue_attributes(queue_name):
...
 
 
-def delete_queue(queue_name):
+def delete_queue(queue_url):
+    print("DELETEING {}".format(queue_url))
     response = client.delete_queue(
-        QueueUrl=queue_name
+        QueueUrl=queue_url
     )
     if response['ResponseMetadata']['HTTPStatusCode'] == 200:
```
imdadahad: I should've added this in as a comment, but apparently according to `boto` docs for `delete_queue` it says that "If the specified queue doesn't exist, Amazon SQS returns a successful response.". Just worth adding in as a comment for future users of the script.

minglis: done

PR #1003, opened 2017-05-31
manually set sms_sender when creating service

leohemsted:
sqlalchemy default doesn't appear to work correctly when there is a difference between the DB schema and the code (ie: during a migration) in this case, lets just set sms_sender ourselves.

we can't write unit tests for this because this only happens when the db is in an inconsistent state :weary:

(i've actually tested this).

Once this is deployed, I'll re-re-re-re-revert the db migration


minglis:
👍

PR #1005, opened 2017-06-01
Ignore one-off messages in job list on dashboard

quis:
Same as how we ignore ‘send yourself a test’ messages (see: d8467bfc3cdca0f7dcd019f93ca20d981ef28760). The dashboard gets clogged up with one off messages otherwise, which affects:
- performance
- users ability to find their jobs


quis:
# Rename ‘One-off message’ to ‘Report’

See alphagov/notifications-admin@7915845

PR #1008, opened 2017-06-01
mmg spell received incorrectly, lets use that

leohemsted:
also made it more resilient


minglis:
👍

But we should see if they can fix their typo.....

PR #1010, opened 2017-06-02
Inbound sms api

leohemsted:
This adds an inbound sms api, with two endpoints - one to get all inbound sms, and one to get a summary (for use on the dashboard)

the inbound sms endpoint is limitable and searchable

also, I added an `admin_request` fixture which takes care of some `client` boilerplate - namely gets rid of the need to sort out headers, and json loading/dumping.


quis:
Endpoints look good to me. Someone else should probs look at the code more closely.

PR #1015, opened 2017-06-05
Return notification content for outbound messages

imdadahad:
This changes `search_for_notification_by_to_field` to return the content of each `Notification`. 

## Summary

Instead of creating a new endpoint for retrieving notification content for outbound messages, this uses the existing endpoint which is used in `get_notifications_for_service`. This returns the notifications in descending order of `created_at`.

### DIFF
--- a/tests/app/service/test_rest.py
+++ b/tests/app/service/test_rest.py
@@ -10,7 +10,12 @@ from freezegun import freeze_time
 
 from app.dao.users_dao import save_model_user
 from app.dao.services_dao import dao_remove_user_from_service
-from app.models import User, Organisation, DVLA_ORG_LAND_REGISTRY, Rate, ServicePermission
+from app.models import (
+    Organisation, Rate, Service, ServicePermission, User,
+    KEY_TYPE_NORMAL, KEY_TYPE_TEAM, KEY_TYPE_TEST,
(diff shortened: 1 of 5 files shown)


tests/app/service/test_rest.py
```diff
@@ -2051,3 +2050,34 @@ def test_search_for_notification_by_to_field_filters_by_statuses(client, notify_
...
+        headers=[create_authorization_header()]
+    )
+    notifications = json.loads(response.get_data(as_text=True))['notifications']
+
+    assert response.status_code == 200
+    assert len(notifications) == 1
+
+    assert notifications[0]['id'] == str(notification.id)
+    assert notifications[0]['to'] == '+447700900855'
+    assert notifications[0]['body'] == 'Hello Foo\nYour thing is due soon'
```
quis: Are we going to get back both `body` and `content`? Or is `body` instead of `content`?

imdadahad: Only the `body` which has the personalisation content subbed in. Did we want anything else?

imdadahad: Just for reference it uses `NotificationWithPersonalisationSchema` in `schemas.py` to do a dump of the `Notification`.

quis: We should keep using the schema in the same way for consistency. 👍 to not change.

PR #1020, opened 2017-06-06
Add private endpoint to get notification by ID

quis:
We need this for the two way stuff in the admin app.

We already have this as a public endpoint, but the admin app can’t use it, because the admin app auths with its own key, not that of the service it’s acting on behalf of.

This endpoint makes sure that a request originating from one service can’t be used to see notifications belonging to another service.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -300,6 +300,19 @@ def get_all_notifications_for_service(service_id):
     ), 200
 
 
+@service_blueprint.route('/<uuid:service_id>/notifications/<uuid:notification_id>', methods=['GET'])
+def get_notification_for_service(service_id, notification_id):
+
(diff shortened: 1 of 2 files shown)


quis:
@imdadahad Doesn’t the `uuid:` type annotation do that?

imdadahad:
Yeah kind of, in that it returns a 404 I think for an invalid `uuid`. Think that's okay technically although not sure if it should be a `400` to indicate that it's a bad request instead. I did that on the other one anyway, happy to approve either way. 

Worth adding a test anyway to verify it returns a `404` or `400` depending on what you do.

quis:
I’m happy to leave it as `404` – could go either way but that is less code. Added a test in 6b5451e
 too.

PR #1021, opened 2017-06-06
Add endpoint to get inbound sms by id

imdadahad:
This adds a new endpoint to get an inbound message by id (to be used by admin).

## Notes:

1. URL: `/service/<id>/inbound-sms/<id>`
2. Added basic validation for now (similar to how we do in v2), will need to add a JSON schema at some point but may possible be an overkill to just validate a single `uuid`...


quis:
If you have `service_id` in the URL this should validate that the message is associated with the service I think (we do this for the outbound endpoint).

quis:
Working locally now 😎

Changes look good 👍

PR #1025, opened 2017-06-07
Remove transformed DVLA files older than 7 days

imdadahad:
This adds a scheduled task (runs at 4:40am) to remove transformed DVLA files older than seven days. We only remove two day's worth like we do in the other removal jobs.

## Summary

We send transformed files to DVLA (that we build) for letter delivery. This removes those files older than seven days but only two day's worth as we do this for the other job files.

Note: Happy for suggestions on how I can make `test_remove_dvla_transformed_files_removes_expected_files` shorter as I opted for readability.

### DIFF
--- a/tests/app/celery/test_scheduled_tasks.py
+++ b/tests/app/celery/test_scheduled_tasks.py
@@ -1,10 +1,12 @@
-import pytest
-
 from datetime import datetime, timedelta
 from functools import partial
+from unittest.mock import call, patch, PropertyMock
 
 from flask import current_app
+
+import pytest
 from freezegun import freeze_time
+
 from app.celery import scheduled_tasks
 from app.celery.scheduled_tasks import (
     delete_email_notifications_older_than_seven_days,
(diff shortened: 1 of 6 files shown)


tests/app/celery/test_scheduled_tasks.py
```diff
@@ -488,3 +492,37 @@ def test_remove_csv_files_filters_by_type(mocker, sample_service):
...
+    just_under_seven_days = seven_days_ago + timedelta(seconds=1)
+    just_over_seven_days = seven_days_ago - timedelta(seconds=1)
+    eight_days_ago = seven_days_ago - timedelta(days=1)
+    nine_days_ago = eight_days_ago - timedelta(days=1)
+    just_under_nine_days = nine_days_ago + timedelta(seconds=1)
+    just_over_nine_days = nine_days_ago - timedelta(seconds=1)
+
+    job(created_at=seven_days_ago)
+    job(created_at=just_under_seven_days)
+    job_to_delete_1 = job(created_at=just_over_seven_days)
```
minglis: Test case for a job that shouldn't be deleted?

imdadahad: Sure thing, added in `test_remove_dvla_transformed_files_does_not_remove_files`

PR #1026, opened 2017-06-08
Bumped version of celery

minglis:
When this deploys it MUST do workers THEN api. We'll need to figure that out.

as per:
http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#step-1-upgrade-to-celery-3-1-25

Part of https://www.pivotaltracker.com/story/show/146435939


minglis:
When this deploys it MUST do workers THEN api. We'll need to figure that out.

PR #1027, opened 2017-06-08
Template query to not limit by days

minglis:
Fix for:

https://www.pivotaltracker.com/story/show/146615775

Note needs further work in:

https://www.pivotaltracker.com/story/show/146850973 to have a longer term solution.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -109,7 +109,9 @@ def dao_get_template_usage(service_id, limit_days=None):
         table = Notification
 
     query_filter = [table.service_id == service_id, table.key_type != KEY_TYPE_TEST]
-    if limit_days is not None:
+
+    # only limit days if it's not seven days, as 7 days == the whole of Notifications table.
+    if limit_days is not None and limit_days != 7:
         query_filter.append(table.created_at >= days_ago(limit_days))
 
     notifications_aggregate_query = db.session.query(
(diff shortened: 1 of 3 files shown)


app/dao/notifications_dao.py
```diff
@@ -103,13 +103,26 @@ def create_notification_statistics_dict(service_id, day):
...
     table = NotificationHistory
 
-    if limit_days and limit_days <= 7:  # can get this data from notifications table
+    if limit_days is not None and limit_days <= 7:
         table = Notification
 
-    query_filter = [table.service_id == service_id, table.key_type != KEY_TYPE_TEST]
-    if limit_days is not None:
+        # only limit days if it's not seven days, as 7 days == the whole of Notifications table.
+        if limit_days != 7:
```
servingUpAces: I think the if statement on line 125 will do this, we could get rid of this one.

app/dao/notifications_dao.py
```diff
@@ -103,13 +103,26 @@ def create_notification_statistics_dict(service_id, day):
...
+
+    elif limit_days is not None:
+        # case where not under 7 days, so using NotificationsHistory so limit allowed
+        query_filter.append(table.created_at >= days_ago(limit_days))
+
+    query_filter.append(table.service_id == service_id)
+    query_filter.append(table.key_type != KEY_TYPE_TEST)
+
+    # only limit days if it's not seven days, as 7 days == the whole of Notifications table.
+    if limit_days is not None and limit_days != 7:
```
servingUpAces: Do we need tests in test_notification_dao to test limit_days=7?

app/dao/notifications_dao.py
minglis: it was reading really weird so I overwrote the code.

app/dao/notifications_dao.py
minglis: think exitsing tests have this case

PR #1030, opened 2017-06-09
Update celery to 4.0.2

minglis:
[WIP] - Need to be careful how we release this.

Updates celery to 4.0.2

To make this work we need to pull the celery config out of the flask config as celery now has lowercase config property names, which Flask config ignores as Flask config ignores all lowercase names.

So:
- Celery config in the celery.py file. Makes sense
- Celery config initialised using Flask Config - this is only the queue prefix at the moment.
- QueueNames object now in __init__.py in celery folder. Keeps it Handy. Not in celery.py due to circular dependencies.
- Celery config needs String Constants from models.py. These can't be imported from there as again cicular dependencies. So ALL string constants from the model moved into definititions.py. This is why so many files affected.

Up for review and comment.


minglis:
This is not quite ready. I'll take a look later and fix some issues with conflicting config requirements in Flask and Celery.

imdadahad:
So given that we have `definitions.py` now, worth making sure all those original vars aren't being imported from `app.model`. Found a case in `tests/app/conftest.py` where `NOTIFICATION_STATUS_TYPES_COMPLETED` is being imported from `app.models`.

Not sure if there's a quick way to do this...

minglis:
There is a problem with celery 4 and boto 2 that causes an issue. This needs a think to figure out what to do.

https://github.com/celery/celery/issues/3221

PR #1031, opened 2017-06-12
Secrets 2 fa

venusbb:
Modify the generation of (5 digits) secret code in users_dao from using random number to using this library:

https://docs.python.org/3/library/secrets.html

secrets.py was put in the notifications/app folder


app/dao/users_dao.py
```diff
@@ -3,15 +3,20 @@
...
         dict.pop(key, None)
 
 
 def create_secret_code():
-    return ''.join(map(str, random.sample(range(9), 5)))
+    L1 = []
+    for i in range(0, 5):
+        L1.append(secrets.randbelow(10))    #return cryptographically strong random number using secrets module
+    L2 = ''.join(map(str, L1))
+    return L2
```
leohemsted: we don't need to add systems.py, that adds lots of extra functionality that we don't need and is a bit overengineered for this change.


Additionally, by using [string.digits](https://docs.python.org/2/library/string.html#string.digits) we can avoid using the range function, and the casting to a string.

```python
>>> ''.join(random.SystemRandom().sample(string.digits, 5))
'01549'
```

app/dao/users_dao.py
```diff
@@ -3,15 +3,20 @@
...
+
```
quis: I think this could be cleaner – what about:

`str(secrets.randbelow(10**5)).zfill(5)`

leohemsted: additional alternative suggestion - `'{:0>5}'.format(random.SystemRandom().randint(0, 99999))`

PR #1032, opened 2017-06-12
Remove DVLA response files older than 7 days

imdadahad:
This adds a task (run daily at 5:10AM) to remove DVLA response files older than 7 days.

## Summary

This works by using `boto` to get all s3 objects in the bucket and then, for each object, checking `Object["LastModified"]` to see if it is older than seven days (similar to how we do with other jobs). If so we remove the file.

## Notes

- `boto` annoyingly returns the root directory an `Object`, e.g. if you wanted all files in `somedir/`, it'll return all the files within as objects but additionally an object for `somedir/`. This has to be removed manually.
- the `Object["LastModified"]` is 'offset-aware' and so needs additionally processing to make sure date range filters work as expected.

### DIFF
--- a/tests/app/celery/test_scheduled_tasks.py
+++ b/tests/app/celery/test_scheduled_tasks.py
@@ -44,9 +44,10 @@ from tests.app.conftest import (
     sample_job as create_sample_job,
     sample_notification_history as create_notification_history,
     create_custom_template,
-    set_config_values
+    datetime_in_past
 )
-from tests.app.aws.test_s3 import single_s3_object_stub, datetime_in_past
+from tests.app.aws.test_s3 import single_s3_object_stub
+from tests.conftest import set_config_values
 
 
 def _create_slow_delivery_notification(provider='mmg'):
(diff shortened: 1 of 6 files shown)


tests/app/celery/test_scheduled_tasks.py
```diff
@@ -89,6 +93,8 @@ def test_should_have_decorated_tasks_functions():
         'switch_current_sms_provider_on_slow_delivery'
     assert delete_inbound_sms_older_than_seven_days.__wrapped__.__name__ == \
         'delete_inbound_sms_older_than_seven_days'
+    assert delete_dvla_response_files_older_than_seven_days.__wrapped__.__name__ == \
+        'delete_dvla_response_files_older_than_seven_days'
```
kentsanggds: Shouldn't this be trying to ensure that the celery task name `delete_dvla_response_files` is linked to the function? 
I think that this is not implemented correctly as I can change the celery task name and the test still passes leading the test to report false positives.

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -488,3 +494,40 @@ def test_remove_csv_files_filters_by_type(mocker, sample_service):
...
+        call(current_app.config['DVLA_RESPONSE_BUCKET_NAME'], single_page_s3_objects[0]["Contents"][0]["Key"]),
+        call(current_app.config['DVLA_RESPONSE_BUCKET_NAME'], single_page_s3_objects[0]["Contents"][1]["Key"])
+    ])
+
+
+def test_delete_dvla_response_files_older_than_seven_days_does_not_remove_files(notify_api, mocker):
+    single_page_s3_objects = [{
+        "Contents": [
+            single_s3_object_stub('bar/foo1.txt', datetime_in_past(days=6)),
+            single_s3_object_stub('bar/foo2.txt', datetime_in_past(days=9)),
```
kentsanggds: Could refactor the test to explicitly show what the values passed in for the days argument mean, as 9 doesn't mean anything until you look at the code and find out that there is a 2 day tolerance from the `older_than` date.

PR #1033, opened 2017-06-13
Job stats for dashboard

servingUpAces:
New endpoints created that use a query that use the job_statistics table rather than the notifications table to aggregate totals. We should see a performance improvement on the dashboard, once integrated into the admin app.

### DIFF
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -8,7 +8,7 @@ from app.dao import days_ago
 from app.models import (
     Job, JobStatistics, Notification, NotificationHistory, Template,
     JOB_STATUS_SCHEDULED, JOB_STATUS_PENDING,
-    EMAIL_TYPE, SMS_TYPE, LETTER_TYPE
+    LETTER_TYPE
 )
 from app.statsd_decorators import statsd
 
@@ -142,3 +142,36 @@ def dao_get_all_letter_jobs():
     return db.session.query(Job).join(Job.template).filter(
         Template.template_type == LETTER_TYPE
     ).order_by(desc(Job.created_at)).all()
+
+
+@statsd(namespace="dao")
+def dao_get_job_statistics_for_job(job_id):
+    query = db.session.query(
+        JobStatistics.job_id,
+        Job.original_file_name,
+        Job.created_at,
+        JobStatistics.sent,
+        JobStatistics.delivered,
+        JobStatistics.failed
+    ).filter(JobStatistics.job_id == job_id).filter(Job.id == JobStatistics.job_id)
+    return query.one()
+
+
+@statsd(namespace="dao")
+def dao_get_job_stats_for_service(service_id, page=1, page_size=50, limit_days=None):
+    query = Job.query.join(
+        JobStatistics, Job.id == JobStatistics.job_id
+    ).filter(
+        Job.service_id == service_id
+    ).add_columns(
+        JobStatistics.job_id,
+        Job.original_file_name,
+        Job.created_at,
+        JobStatistics.sent,
+        JobStatistics.delivered,
+        JobStatistics.failed
+    )
+    if limit_days:
+        query = query.filter(Job.created_at >= days_ago(limit_days))
+    query = query.order_by(Job.created_at.desc())
+    return query.paginate(page=page, per_page=page_size).items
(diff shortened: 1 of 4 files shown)


app/dao/jobs_dao.py
```diff
@@ -142,3 +142,59 @@ def dao_get_all_letter_jobs():
...
+
+
+@statsd(namespace="dao")
+def dao_get_job_statistics_for_job(service_id, job_id):
+    query = Job.query.join(
+        JobStatistics, Job.id == JobStatistics.job_id
+    ).filter(
+        Job.id == job_id
+    ).join(
+        Template, Template.id == Job.template_id and Template.version == Job.template_version
```
leohemsted: Why are we joining onto template? we don't appear to be using any columns from it

app/dao/jobs_dao.py
```diff
@@ -142,3 +142,59 @@ def dao_get_all_letter_jobs():
...
+        Job.template_id,
+        Job.template_version,
+        Job.job_status,
+        Job.service_id,
+        Job.notification_count,
+        JobStatistics.sent,
+        JobStatistics.delivered,
+        JobStatistics.failed
+    ).filter(
+        Job.service_id == service_id
```
leohemsted: can you combine this filter with the other filter?

tests/app/job/test_rest.py
```diff
@@ -762,3 +763,89 @@ def test_get_all_notifications_for_job_returns_csv_format(
...
+
+def test_get_jobs_raises_for_bad_limit_days(client, sample_service):
+    path = '/service/{}/job/job-stats'.format(sample_service.id)
+    auth_header = create_authorization_header()
+    response = client.get(path,
+                          query_string={'limit_days': 'bad_number'},
+                          headers=[auth_header])
+    assert response.status_code == 400
+    assert response.get_data(as_text=True) == '{\n  "message": {\n    "limit_days": [\n      ' \
+                                              '"bad_number is not an integer"\n    ]\n  },\n  "result": "error"\n}'
```
leohemsted: can you json parse the response and assert that the json matches? i don't like this because the dict could re-order at any time based on the implementation of `json.dumps`

tests/app/job/test_rest.py
```diff
@@ -762,3 +763,89 @@ def test_get_all_notifications_for_job_returns_csv_format(
...
+    assert resp_json["service_id"] == str(sample_job.service_id)
+    assert resp_json["requested"] == sample_job.notification_count
+    assert resp_json["sent"] == 0
+    assert resp_json["delivered"] == 0
+    assert resp_json["failed"] == 0
+
+
+def test_get_job_stats_with_invalid_job_id_returns404(client, sample_template):
+    service_id = sample_template.service.id
+    path = '/service/{}/job/job-=stats{}'.format(service_id, "bad-id")
```
leohemsted: think there's a typo in this bad id (unless you intended this)

should it not still be `/service/{}/job/job-stats/{}`

cos this will 404 because of the routing engine not recognising the path, rather than the job id not matching anything in the DB which is what we should be testing

tests/app/job/test_rest.py
```diff
@@ -762,3 +763,89 @@ def test_get_all_notifications_for_job_returns_csv_format(
...
+    auth_header = create_authorization_header()
+    response = client.get(path, headers=[auth_header])
+    assert response.status_code == 404
+    resp_json = json.loads(response.get_data(as_text=True))
+    assert resp_json['result'] == 'error'
+    assert resp_json['message'] == 'No result found'
+
+
+def test_get_job_stats_with_invalid_service_id_returns404(client, sample_job):
+    path = '/service/{}/job/job-=stats{}'.format(uuid.uuid4(), sample_job.id)
```
leohemsted: same URL comment here

app/dao/jobs_dao.py
servingUpAces: I was going to return the template data for the admin app because in one place they get the data in a separate call. I'll take this out for now.

tests/app/job/test_rest.py
```diff
@@ -762,3 +763,89 @@ def test_get_all_notifications_for_job_returns_csv_format(
...
+    auth_header = create_authorization_header()
+    response = client.get(path, headers=[auth_header])
+    assert response.status_code == 404
+    resp_json = json.loads(response.get_data(as_text=True))
+    assert resp_json['result'] == 'error'
+    assert resp_json['message'] == 'No result found'
+
+
+def test_get_job_stats_with_invalid_service_id_returns404(client, sample_job):
+    path = '/service/{}/job/job-stats{}'.format(uuid.uuid4(), sample_job.id)
```
leohemsted: sorry, you're still missing a slash in `job-stats/{}`

PR #1034, opened 2017-06-15
Push inbound sms

servingUpAces:
New table to store the inbound api information for a service. The table is versioned.
There is a new endpoint to create the inbound api and one to update it.


tests/app/dao/test_service_inbound_api_dao.py
```diff
@@ -0,0 +1,116 @@
...
+            assert not x.updated_at
+        elif x.version == 2:
+            assert x.url == "https::/some_service/changed_url"
+            assert x.updated_at
+        else:
+            pytest.fail("version should not exist")
+        assert x.id
+        assert x.service_id == sample_service.id
+        assert x.updated_by_id == sample_service.users[0].id
+        assert get_secret(x.bearer_token) == "some_unique_string"
```
leohemsted: woah this test is pretty long and complex. I try and avoid for loops entirely in tests, but definitely for loops with ifs and elses and pytest.fails inside

i think this would be much nicer by saying

```
assert versioned_results[0].url == ...
assert versioned_results[0].updated_at is None

assert versioned_results[1].url == ...
assert versioned_results[1].updated_at is not None

...
```

tests/app/service/test_rest.py
```diff
@@ -2149,3 +2149,73 @@ def test_search_for_notification_by_to_field_returns_content(
...
+    assert resp_json["id"]
+    assert resp_json["service_id"] == str(sample_service.id)
+    assert resp_json["url"] == "https://some_service/inbound-sms"
+    assert resp_json["bearer_token"] != "some-unique-string"  # returned encrypted
+    assert resp_json["updated_by_id"] == str(sample_service.users[0].id)
+    assert resp_json["created_at"]
+    assert not resp_json["updated_at"]
+
+
+def test_set_service_inbound_api_raises_500_when_service_does_not_exist(client):
```
leohemsted: i think either we should be returning 404, or this test shouldn't exist

app/service/rest.py
```diff
@@ -531,3 +539,32 @@ def get_yearly_monthly_usage(service_id):
         return json.dumps(json_results)
     except TypeError:
         return jsonify(result='error', message='No valid year provided'), 400
+
+
+@service_blueprint.route('/<uuid:service_id>/inbound-api', methods=['POST'])
+def create_service_inbound_api(service_id):
+    data = request.get_json()
+    validate(data, service_inbound_api)
+    data["service_id"] = service_id
```
servingUpAces: It's difficult to test this, if it does timeout, that will timeout the request for the admin app. 
Instead we could add a test button from the admin app.

PR #1036, opened 2017-06-16
Send one off messages from new admin endpoint

leohemsted:
new admin endpoint takes `template_id`, `personalisation`, `to` and `created_by` vals, and, after similar validation to an API call, creates a notification and queues up its sending.

We now track `created_by` on the notification object - although only one-off notifications will have a value for this, since it's not applicable for API, and for jobs it's already stored on `job.created_by`.

### DIFF
--- a/app/service/send_notification.py
+++ b/app/service/send_notification.py
@@ -1,0 +1,61 @@
+from app.config import QueueNames
+from app.notifications.validators import (
+    check_service_over_daily_message_limit,
+    check_sms_content_char_count,
+    validate_and_format_recipient,
+)
+from app.notifications.process_notifications import (
+    create_content_for_notification,
+    persist_notification,
+    send_notification_to_queue,
+)
+from app.models import (
+    KEY_TYPE_NORMAL,
+    PRIORITY,
+    SMS_TYPE,
+)
+from app.dao.services_dao import dao_fetch_service_by_id
+from app.dao.templates_dao import dao_get_template_by_id_and_service_id
+
+
+def send_one_off_notification(service_id, post_data):
+    service = dao_fetch_service_by_id(service_id)
+    template = dao_get_template_by_id_and_service_id(
+        template_id=post_data['template_id'],
+        service_id=service_id
+    )
+
+    personalisation = post_data.get('personalisation', None)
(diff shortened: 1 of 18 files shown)


migrations/versions/0099_notification_created_by.py
```diff
@@ -0,0 +1,24 @@
...
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    op.add_column('notifications', sa.Column('created_by_id', postgresql.UUID(as_uuid=True), nullable=True))
+    op.create_foreign_key(None, 'notifications', 'users', ['created_by_id'], ['id'])
+
+
+def downgrade():
+    op.drop_constraint(None, 'notifications', type_='foreignkey')
```
leohemsted: get rid of this line

migrations/versions/0099_notification_created_by.py
```diff
@@ -0,0 +1,24 @@
...
+# revision identifiers, used by Alembic.
+revision = '0099_notification_created_by'
+down_revision = '0098_service_inbound_api'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    op.add_column('notifications', sa.Column('created_by_id', postgresql.UUID(as_uuid=True), nullable=True))
```
leohemsted: add history table as well

app/service/send_notification.py
```diff
@@ -0,0 +1,76 @@
...
+def send_one_off_notification(service_id, post_data):
+    service = dao_fetch_service_by_id(service_id)
+    template = dao_get_template_by_id_and_service_id(
+        template_id=post_data['template_id'],
+        service_id=service_id
+    )
+
+    personalisation = post_data.get('personalisation', None)
+
+    if template.template_type == SMS_TYPE:
```
servingUpAces: Can we use the validators.validate_template method here?

migrations/versions/0099_notification_created_by.py
servingUpAces: Drop column from history as well

PR #1038, opened 2017-06-20
Push inbound sms

servingUpAces:
New task to send the inbound sms to the service. If the service has not set the url then nothing happens. If the request to the service url returns with 500 or greater the task is retries. The task is created when the SMS provider post the inbound SMS

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -411,8 +411,9 @@ def send_inbound_sms_to_service(self, inbound_sms_id, service_id):
         current_app.logger.exception("Exception raised in send_inbound_sms_to_service for service_id: {} and url: {}. "
                                      "\n{}".format(service_id, inbound_api.url, e))
         if e.response.status_code >= 500:
-            self.retry(queue=QueueNames.RETRY,
-                       exc='Unable to send_inbound_sms_to_service for service_id: {} and url: {}. \n{}'.format(
-                           service_id, inbound_api.url, e))
-    except self.MaxRetriesExceededError:
-        current_app.logger.exception('Retry: send_inbound_sms_to_service has retried the max number of times')
+            try:
+                self.retry(queue=QueueNames.RETRY,
+                           exc='Unable to send_inbound_sms_to_service for service_id: {} and url: {}. \n{}'.format(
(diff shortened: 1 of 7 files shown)


app/celery/tasks.py
```diff
@@ -371,3 +376,43 @@ def process_updates_from_file(response_file):
...
+    try:
+        response.raise_for_status()
+    except HTTPError as e:
+        current_app.logger.exception("Exception raised in send_inbound_sms_to_service for service_id: {} and url: {}. "
+                                     "\n{}".format(service_id, inbound_api.url, e))
+        if e.response.status_code >= 500:
+            self.retry(queue=QueueNames.RETRY,
+                       exc='Unable to send_inbound_sms_to_service for service_id: {} and url: {}. \n{}'.format(
+                           service_id, inbound_api.url, e))
+    except self.MaxRetriesExceededError:
```
leohemsted: this won't catch the retry error - this except will only look at `response.raise_for_status`.

eg try running this:

```python
try:
  raise KeyError('foo')
except KeyError:
  raise ValueError('bar')
except ValueError:
  print('hello')
```

You'll need to wrap the `self.retry` call in a separate try catch for the `MaxRetriesExceededError`

app/notifications/receive_notifications.py
```diff
@@ -30,28 +32,59 @@ def receive_mmg_sms():
...
-    inbound = create_inbound_mmg_sms_object(service, post_data)
+    inbound = create_inbound_sms_object(service,
+                                        content=format_mmg_message(post_data["Message"]),
+                                        from_number=post_data['MSISDN'],
+                                        provider_ref=post_data["ID"],
+                                        date_received=post_data.get('DateRecieved'),
+                                        provider_name="mmg")
 
+    tasks.send_inbound_sms_to_service.apply_async([str(inbound.id), str(service.id)], queue=QueueNames.NOTIFY)
     current_app.logger.info('{} received inbound SMS with reference {}'.format(service.id, inbound.provider_reference))
```
leohemsted: i know you didn't write the firetext one originally but can we duplicate this logging line to the `receive_firetext_sms` function?

PR #1047, opened 2017-06-26
add logs to see the IP address of the inbound sms

venusbb:
Some test codes to inject the inbound sms sender's IP address to the log

### DIFF
--- a/app/notifications/receive_notifications.py
+++ b/app/notifications/receive_notifications.py
@@ -28,6 +28,7 @@ def receive_mmg_sms():
         'DateRecieved': '2017-05-21+11%3A56%3A11'
     }
     """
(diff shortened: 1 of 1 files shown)


app/notifications/receive_notifications.py
```diff
@@ -28,6 +28,7 @@ def receive_mmg_sms():
         'DateRecieved': '2017-05-21+11%3A56%3A11'
     }
     """
+    current_app.logger.info("Inbound sms sender IP information {}".format(request.headers.getlist("X-Forwarded-For")))
```
leohemsted: @yahoopete @servingUpAces are we okay putting this IP address in our logs? Given that it'll just be MMG/firetext, i'm not sure how much that classifies as PII

PR #1048, opened 2017-06-28
Use new service permissions and allow disabling of email/sms

kentsanggds:
Refactored the code to remove references to permission flags in service and use service permissions instead.

Also updated code to ensure that email / sms permission settings are honoured for sending, and creating / updating templates.

If the notification is already scheduled or in a queue then honour sending it.

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -17,6 +17,7 @@ from notifications_utils.template import SMSMessageTemplate
 from app.dao.services_dao import dao_fetch_service_by_id
 from app.dao.service_permissions_dao import dao_fetch_service_permissions
 from app.models import SMS_TYPE, EMAIL_TYPE
+from app.notifications.validators import service_has_permission
 from app.schemas import (template_schema, template_history_schema)
 
 template_blueprint = Blueprint('template', __name__, url_prefix='/service/<uuid:service_id>/template')
@@ -25,7 +26,7 @@ from app.errors import (
     register_errors,
     InvalidRequest
 )
-from app.utils import get_template_instance
+from app.utils import get_template_instance, get_public_notify_type_text
 
 register_errors(template_blueprint)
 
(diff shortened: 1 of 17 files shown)


app/template/rest.py
```diff
@@ -35,10 +36,24 @@ def _content_count_greater_than_limit(content, template_type):
     return template.content_count > current_app.config.get('SMS_CHAR_COUNT_LIMIT')
 
 
+def _has_service_permission(template_type, action, permissions):
```
kentsanggds: error response is slightly different as notifications have a 400 response and templates have a 403 response.

tests/app/notifications/rest/test_send_notification.py
```diff
@@ -1163,3 +1168,34 @@ def test_should_allow_international_number_on_sms_notification(client, notify_db
...
 
     assert response.status_code == 201
+
+
+@pytest.mark.parametrize(
+    'template_factory, to, expected_error', [
+        (sample_template_without_sms_permission, '+447700900986', 'Cannot send text messages'),
+        (sample_template_without_email_permission, 'notify@digital.cabinet-office.gov.uk', 'Cannot send emails')
+    ])
+def test_should_not_allow_notification_if_service_permission_not_set(
```
kentsanggds: ok, will add some for v2 stuff

kentsanggds: Actually, seems like v2 has been missed out in the permissions so will add it, and also noticed need to refactor to handle one-off msgs

app/v2/notifications/post_notifications.py
```diff
@@ -30,8 +32,14 @@ def post_notification(notification_type):
     else:
         form = validate(request.get_json(), post_sms_request)
 
+    if service_has_permission(notification_type, authenticated_service.permissions) is False:
+        raise BadRequestError(message="Cannot send {}".format(
+            get_public_notify_type_text(notification_type, plural=True)))
+
     scheduled_for = form.get("scheduled_for", None)
-    service_can_schedule_notification(authenticated_service, scheduled_for)
+    if scheduled_for:
```
servingUpAces: This has been refactored to read in one line of code to three. It doesn't seem like much but the point of the validators module is to compartmentalise the functions to validate notifications such that when v3 comes along it will be easy to build. Plus it makes the controller code succinct.

kentsanggds: ok, II've abstracted it out but renamed the method in line with other methods in the validators module

PR #1049, opened 2017-06-28
Templates can now be [REDACTED] 

leohemsted:
If you make an update_template post containing `{"redact_personalisation": true, "updated_by_id": "..."}`, then we store a `redact_personalisation` marker in a new database table - TemplateRedacted.

This table just contains that boolean, and who and when updated it.

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -59,12 +59,8 @@ def update_template(service_id, template_id):
     data = request.get_json()
 
     # if redacting, don't update anything else
-    if data.get('redact_personalisation') is True and 'updated_by_id' in data:
-        # we also don't need to check what was passed in redact_personalisation - its presence in the dict is enough.
-        # also, only
-        if not fetched_template.redact_personalisation:
-            dao_redact_template(fetched_template, data['updated_by_id'])
-        return 'null', 200
+    if data.get('redact_personalisation') is True:
+        return redact_template(fetched_template, data)
 
     current_data = dict(template_schema.dump(fetched_template).data.items())
     updated_template = dict(template_schema.dump(fetched_template).data.items())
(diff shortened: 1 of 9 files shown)


app/template/rest.py
```diff
@@ -55,9 +56,17 @@ def create_template(service_id):
 def update_template(service_id, template_id):
     fetched_template = dao_get_template_by_id_and_service_id(template_id=template_id, service_id=service_id)
 
+    data = request.get_json()
+
+    # if redacting, don't update anything else
+    if data.get('redact_personalisation') is True and 'updated_by_id' in data:
```
servingUpAces: If updated_by_id is not set what will happen?

migrations/versions/0102_template_redacted.py
```diff
@@ -0,0 +1,31 @@
...
+        sa.Column('template_id', postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column('redact_personalisation', sa.Boolean(), nullable=False),
+        sa.Column('updated_at', sa.DateTime(), nullable=False),
+        sa.Column('updated_by_id', postgresql.UUID(as_uuid=True), nullable=False),
+        sa.ForeignKeyConstraint(['template_id'], ['templates.id'], ),
+        sa.ForeignKeyConstraint(['updated_by_id'], ['users.id'], ),
+        sa.PrimaryKeyConstraint('template_id')
+    )
+    op.create_index(op.f('ix_template_redacted_updated_by_id'), 'template_redacted', ['updated_by_id'], unique=False)
+
```
servingUpAces: Are we doing the insert template_redacted rows in the next deploy?

app/template/rest.py
```diff
@@ -55,9 +56,17 @@ def create_template(service_id):
 def update_template(service_id, template_id):
```
servingUpAces: Is there a test for updating the template to redacted?

leohemsted: added

app/template/rest.py
leohemsted: added a test - nothing happens. Do not update, do not pass go, do not collect $200

PR #1051, opened 2017-06-29
use created_by instead of updated_by

leohemsted:
to behave in same way as other endpoints


leohemsted:
will merge as part of https://github.com/alphagov/notifications-api/pull/1052

PR #1063, opened 2017-07-04
Prevent inbound SMS store if no inbound SMS permissions

kentsanggds:
This update prevents services that do not have inbound SMS and SMS permissions from storing the inbound sms message.

### DIFF
--- a/app/notifications/receive_notifications.py
+++ b/app/notifications/receive_notifications.py
@@ -9,7 +9,7 @@ from app.celery import tasks
 from app.config import QueueNames
 from app.dao.services_dao import dao_fetch_services_by_sms_sender
 from app.dao.inbound_sms_dao import dao_create_inbound_sms
-from app.models import InboundSms
+from app.models import InboundSms, INBOUND_SMS_TYPE, SMS_TYPE
 from app.errors import register_errors
 from app.utils import convert_bst_to_utc
 
@@ -34,8 +34,8 @@ def receive_mmg_sms():
 
     potential_services = fetch_potential_services(inbound_number, 'mmg')
     if not potential_services:
-        # since this is an issue with our service <-> number mapping, we should still tell MMG that we received
-        # successfully
+        # since this is an issue with our service <-> number mapping, or no inbound_sms service permission
+        # we should still tell MMG that we received it successfully
         return 'RECEIVED', 200
 
     statsd_client.incr('inbound.mmg.successful')
@@ -128,6 +128,13 @@ def fetch_potential_services(inbound_number, provider_name):
         ))
         statsd_client.incr('inbound.{}.failed'.format(provider_name))
         return False
+
+    str_permissions = [p.permission for p in potential_services[0].permissions]
+    if set([INBOUND_SMS_TYPE, SMS_TYPE]).issubset(set(str_permissions)) is False:
+        current_app.logger.error(
+            'Service "{}" does not allow inbound SMS'.format(potential_services[0].id))
+        return False
+
     return potential_services
 
 
--- a/app/notifications/receive_notifications.py
+++ b/app/notifications/receive_notifications.py
@@ -67,7 +67,6 @@ def receive_firetext_sms():
         }), 200
 
     service = potential_services[0]
-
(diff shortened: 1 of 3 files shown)


app/notifications/receive_notifications.py
```diff
@@ -128,6 +127,13 @@ def fetch_potential_services(inbound_number, provider_name):
         ))
         statsd_client.incr('inbound.{}.failed'.format(provider_name))
         return False
+
+    str_permissions = [p.permission for p in potential_services[0].permissions]
+    if set([INBOUND_SMS_TYPE, SMS_TYPE]).issubset(set(str_permissions)) is False:
+        current_app.logger.error(
```
imdadahad: This is even more a reason to have a helper `service_has_permissions` so you won't have code like _here_ this moving forward.

tests/app/notifications/test_receive_notification.py
```diff
@@ -5,35 +5,89 @@
...
+    ([SMS_TYPE]),
+    ([INBOUND_SMS_TYPE]),
+])
+def test_receive_notification_without_permissions_does_not_create_inbound(
+        client, mocker, notify_db, notify_db_session, permissions, provider, headers, data, expected_response):
+    service = sample_service(notify_db, notify_db_session, permissions=permissions)
+    mocker.patch("app.notifications.receive_notifications.dao_fetch_services_by_sms_sender",
+                 return_value=[service])
+    mocked = mocker.patch("app.notifications.receive_notifications.tasks.send_inbound_sms_to_service.apply_async")
+    mocked_logger = mocker.patch("flask.current_app.logger.error")
```
imdadahad: Do you have to mock a logger. Just thinking it's a bit nasty if we have to do this in any test where  code will have `current_app.logger.error`.

tests/app/notifications/test_receive_notification.py
```diff
@@ -5,35 +5,89 @@
...
+                           data=data,
+                           headers=headers)
 
     assert response.status_code == 200
-    assert response.get_data(as_text=True) == 'RECEIVED'
-    inbound_sms_id = InboundSms.query.all()[0].id
-    mocked.assert_called_once_with([str(inbound_sms_id), str(sample_service.id)], queue="notify-internal-tasks")
+    assert response.get_data(as_text=True) == expected_response
+    assert len(InboundSms.query.all()) == 0
+    assert mocked.called is False
```
imdadahad: `mocked.assert_not_called()`

If the mock was renamed to `send_inbound_mock`, would read much better `send_inbound_mock.assert_not_called()`.

app/notifications/receive_notifications.py
kentsanggds: This code isn't used anywhere else in this way at the moment, so not sure if work refactoring?

tests/app/notifications/test_receive_notification.py
kentsanggds: I'm mocking the logger as that's currently the only way to distinguish the fact that the right code path is being hit, otherwise might confuse with detecting unique service linked to the inbound number.

tests/app/notifications/test_receive_notification.py
kentsanggds: ok

app/notifications/receive_notifications.py
imdadahad: Fair enough

tests/app/notifications/test_receive_notification.py
imdadahad: Not sure I understand. From the test name you want to check a service without permissions can't created an inbound. Doesn't this `assert len(InboundSms.query.all()) == 0` cover that case. My point is if you shouldn't have to mock a logger to test a code path.

kentsanggds: But it's also the case if there are no unique services bound to the number, hence I'm trying to make sure that the logger is hit with that error message

imdadahad: This is why the permissions check should be in a helper function, as that way you can mock that instead.

PR #1065, opened 2017-07-04
2/3 Stop updating old Notification status column

imdadahad:
This stops updating the old `Notification. _status_enum` and `NotificationHistory. _status_enum` columns and instead will now only update the newer `_status_fkey` column.

## Notes

- Return new column when calling `.status`
- Remove unncessary tests that were checking `._status_enum` was being populated
- Updates some tests which can cause failure if unordered

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -788,14 +788,14 @@ class Notification(db.Model):
         unique=False,
         nullable=True,
         onupdate=datetime.datetime.utcnow)
-    _status_enum = db.Column('status', NOTIFICATION_STATUS_TYPES_ENUM, index=True, nullable=True, default='created')
-    _status_fkey = db.Column(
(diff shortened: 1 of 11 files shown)


app/models.py
```diff
@@ -819,12 +819,11 @@ class Notification(db.Model):
 
     @hybrid_property
     def status(self):
-        return self._status_enum
+        return self._status_fkey
```
leohemsted: think we can just get rid of this property and rename `_status_fkey` to `status`

leohemsted:
looks like in this sea of reverts we've lost your upgrade script that set status to nullable?

imdadahad:
It was because of additional changes I made to `models.py`. Because of those the nullable has to go in first. Will re-raise that pr and merge in first.

PR #1066, opened 2017-07-04
modified the parsing of return ip address

venusbb:
A minor modification on how the return ip address is parsed as the previous version return the string of the whole route.


app/authentication/auth.py
```diff
@@ -47,7 +47,9 @@ def requires_no_auth():
 def restrict_ip_sms():
     ip = ''
     if request.headers.getlist("X-Forwarded-For"):
-        ip = request.headers.getlist("X-Forwarded-For")[0]
+        ip0 = request.headers.getlist("X-Forwarded-For")
+        ip1 = ip0.split(',')
```
leohemsted: This won't work - getlist will return a list, and you can't call `split` on lists.

I'd recommend that you write some unit tests for this function, it's starting to get complex.

PR #1068, opened 2017-07-04
Add a script to correct out-of-sync notifications

imdadahad:
This fixes an issue where `Notification`'s and `NotificationHistory`'s statuses have become out of sync. See #1067.

## Summary

Currently `[Notification, NotificationHistory]._status_enum` is the source of truth. This script will update `[Notification, NotificationHistory]._status_fkey` with these values. 

This relies on #1067 being merged as this caused the out-of-sync issue in the first place.

## Dependencies

- [ ] #1067 Fix syncing issue with Notification statuses

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -115,3 +115,33 @@ class CustomDbScript(Command):
             print('commit 250000 updates at {}'.format(datetime.utcnow()))
             db.session.commit()
             result_history = db.session.execute(subq_history).fetchall()
+
+    def fix_notification_statuses_not_in_sync(self):
+        """
+        This will be used to correct an issue where Notification._status_enum and NotificationHistory._status_fkey
+        became out of sync. See 979e90a.
+
+        Notification._status_enum is the source of truth so NotificationHistory._status_fkey will be updated with
+        these values.
+        """
+        MAX = 10000
+
+        subq = "SELECT id FROM notifications WHERE cast (status as text) != notification_status".format(MAX)
+        update = "UPDATE notifications SET notification_status = status WHERE id in ({})".format(subq)
+        result = db.session.execute(subq).fetchall()
+
+        while len(result) > 0:
+            db.session.execute(update)
+            print('Committed {} updates at {}'.format(MAX, datetime.utcnow()))
+            db.session.commit()
+            result = db.session.execute(subq).fetchall()
+
+        subq_hist = "SELECT id FROM notification_history WHERE cast (status as text) != notification_status".format(MAX)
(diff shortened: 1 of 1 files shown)


app/commands.py
```diff
@@ -115,3 +115,33 @@ def update_notification_international_flag(self):
...
+        """
+        This will be used to correct an issue where Notification._status_enum and NotificationHistory._status_fkey
+        became out of sync. See 979e90a.
+
+        Notification._status_enum is the source of truth so NotificationHistory._status_fkey will be updated with
+        these values.
+        """
+        MAX = 10000
+
+        subq = "SELECT id FROM notifications WHERE cast (status as text) != notification_status".format(MAX)
```
kentsanggds: Do we need to `format(MAX)` here? as there doesn't appear to be a matching `{}`
Similar change for line 139

app/commands.py
```diff
@@ -115,3 +115,33 @@ def update_notification_international_flag(self):
...
+        update = "UPDATE notifications SET notification_status = status WHERE id in ({})".format(subq)
+        result = db.session.execute(subq).fetchall()
+
+        while len(result) > 0:
+            db.session.execute(update)
+            print('Committed {} updates at {}'.format(MAX, datetime.utcnow()))
```
kentsanggds: `MAX` should probably be `len(result)`?
Similar change for line 145

app/commands.py
```diff
@@ -115,3 +115,33 @@ def update_notification_international_flag(self):
...
+
+    def fix_notification_statuses_not_in_sync(self):
+        """
+        This will be used to correct an issue where Notification._status_enum and NotificationHistory._status_fkey
+        became out of sync. See 979e90a.
+
+        Notification._status_enum is the source of truth so NotificationHistory._status_fkey will be updated with
+        these values.
+        """
+        MAX = 10000
```
kentsanggds: Not sure if `MAX` is needed as not really used in code

imdadahad: You're right, it's needed in the select - just updated.

kentsanggds: What happens when `len(result)` is less and `MAX`? 
Wouldn't the message be misleading?

imdadahad: Why?

The message prints out "Committed x at y" where x is the number of updates and y is the time. MAX is used, like the scripts before, to indicate the MAX number of notifications to process.

Really don't see the issue here.

app/commands.py
imdadahad: 👍

PR #1070, opened 2017-07-05
Add OPG to letter organisations

quis:
Logo was added here: https://github.com/alphagov/notifications-template-preview/pull/19

> OPG logo has already been submitted to DVLA and designated 002

### DIFF
--- a/migrations/versions/0105_opg_letter_org.py
+++ b/migrations/versions/0105_opg_letter_org.py
@@ -1,0 +1,28 @@
+"""empty message
+
+Revision ID: 0105_opg_letter_org
+Revises: 0104_more_letter_orgs
+Create Date: 2017-06-29 12:44:16.815039
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '0105_opg_letter_org'
+down_revision = '0104_more_letter_orgs'
+
+from alembic import op
+import sqlalchemy as sa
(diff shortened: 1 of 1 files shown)


migrations/versions/0105_opg_letter_org.py
```diff
@@ -0,0 +1,28 @@
...
+def upgrade():
+    op.execute("""
+        INSERT INTO dvla_organisation VALUES
+        ('002', 'Office of the Public Guardian')
+    """)
+
+
+def downgrade():
+    # data migration, no downloads
+    pass
```
imdadahad: ?

quis: IDK, I just copy/pasted this from the [previous one](https://github.com/alphagov/notifications-api/blob/master/migrations/versions/0104_more_letter_orgs.py#L27) ¯\\_(ツ)_/¯

imdadahad: Hmm, if we downgrade do we want this to remain in the db?

quis: Could cause 🔥 if you try to downgrade and it’s foreign keyed to a service right?

imdadahad: Yeah unless we account for that in the query but probably an overkill.

PR #1073, opened 2017-07-07
Update organisation dao model rest

kentsanggds:
Updated Organisations model, and DAO to allow creating and updating of Organisations

Update of Organisation model to make logo not nullable
- migration script is included as part of this PR.

Also refactored code in organisation rest to remove marshmallow code and use a json_schema instead.

Part 1 / 4 - https://www.pivotaltracker.com/story/show/148289053

### DIFF
--- a/app/dao/organisations_dao.py
+++ b/app/dao/organisations_dao.py
@@ -1,5 +1,5 @@
 from app import db
-from app.dao.dao_utils import transactional, version_class
+from app.dao.dao_utils import transactional
 from app.models import Organisation
 
 
--- a/app/dao/organisations_dao.py
+++ b/app/dao/organisations_dao.py
@@ -14,8 +14,3 @@ def dao_get_organisation_by_id(org_id):
 @transactional
 def dao_create_organisation(organisation):
     db.session.add(organisation)
-
-
-@transactional
-def dao_update_organisation(organisation):
-    db.session.add(organisation)
--- a/app/dao/organisations_dao.py
+++ b/app/dao/organisations_dao.py
@@ -1,5 +1,5 @@
 from app import db
-from app.dao.dao_utils import transactional
+from app.dao.dao_utils import transactional, version_class
 from app.models import Organisation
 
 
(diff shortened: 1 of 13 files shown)


tests/app/organisation/test_rest.py
```diff
@@ -37,3 +37,52 @@ def test_get_organisation_by_id(notify_api, notify_db, notify_db_session):
...
+
+
+def test_create_organisation_without_logo_raises_error(client, notify_db, notify_db_session):
+    data = {
+        'name': 'test organisation',
+        'colour': '#0000ff',
+    }
+    auth_header = create_authorization_header()
+
+    response = client.post(
```
quis: This could be cleaner if you used the `admin_request` fixture that Leo made. For example:

```python
  '/organisation',
  data={
    'name': 'test organisation',
    'colour': '#0000ff',
  }
  _expected_status=400
)
  
assert response == {…}
```

quis: Same for all these tests.

kentsanggds: ok, sure

app/dao/organisations_dao.py
```diff
@@ -0,0 +1,21 @@
...
+def dao_get_organisations():
+    return Organisation.query.all()
+
+
+def dao_get_organisation_by_id(org_id):
+    return Organisation.query.filter_by(id=org_id).one()
+
+
+@transactional
+def dao_create_organisation(organisation):
```
imdadahad: Hmm, this should take some kwargs like other dao methods do. Otherwise it's just doing the same as `dao_update_organisation` when it probably should be doing more.

app/dao/organisations_dao.py
```diff
@@ -0,0 +1,23 @@
+from app import db
```
servingUpAces: Why was this file renamed?

migrations/versions/0107_change_logo_not_nullable.py
```diff
@@ -0,0 +1,27 @@
+"""empty message
```
servingUpAces: You'll need to rename this file now - there is a new db migration.

kentsanggds: ok

app/dao/organisations_dao.py
kentsanggds: trying to keep it in line with other daos, if it's an issue I can change it back

PR #1079, opened 2017-07-10
retrieve sms ip whitelist from credentials on paas

venusbb:
retrieve ip whitelist from credentials

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -49,7 +49,7 @@ def restrict_ip_sms():
         # X-Forwarded-For looks like "203.0.113.195, 70.41.3.18, 150.172.238.178"
         ip_list = request.headers.get("X-Forwarded-For")
         ip = ip_list.split(',')[0].strip()
-        current_app.logger.info("Inbound sms ip list {}".format(ip_list))
+        current_app.logger.info("Inbound sms ip route list {}".format(ip_list))
 
     if ip in current_app.config.get('ALLOW_IP_INBOUND_SMS'):
         current_app.logger.info("Inbound sms ip addresses {} passed ".format(ip))
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -48,7 +48,6 @@ def restrict_ip_sms():
     if request.headers.get("X-Forwarded-For"):
         # X-Forwarded-For looks like "203.0.113.195, 70.41.3.18, 150.172.238.178"
         ip_route = request.headers.get("X-Forwarded-For")
-        # ip = ip_list.split(',')[0].strip()
         ip_list = ip_route.split(',')
         if len(ip_list) >= 3:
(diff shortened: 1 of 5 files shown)


app/authentication/auth.py
```diff
@@ -47,9 +47,11 @@ def restrict_ip_sms():
...
     if request.headers.get("X-Forwarded-For"):
         # X-Forwarded-For looks like "203.0.113.195, 70.41.3.18, 150.172.238.178"
-        ip_list = request.headers.get("X-Forwarded-For")
-        ip = ip_list.split(',')[0].strip()
-        current_app.logger.info("Inbound sms ip list {}".format(ip_list))
+        ip_route = request.headers.get("X-Forwarded-For")
+        ip_list = ip_route.split(',')
+        if len(ip_list) >= 3:
+            ip = ip_list[len(ip_list) - 3]
+        current_app.logger.info("Inbound sms ip route list {}".format(ip_route))
```
imdadahad: So two things:

1. Might be worth adding a comment here to give some context as to why we're looking for the that specific IP  in the list. Although you explained it to me, just in case anyone else comes across this code they'll need some context.

2. Worth creating a test for this to check it actually works

app/config.py
```diff
@@ -262,7 +262,7 @@ class Config(object):
 
     FREE_SMS_TIER_FRAGMENT_COUNT = 250000
 
-    ALLOW_IP_INBOUND_SMS = []
+    ALLOW_IP_INBOUND_SMS = os.environ.get('SMS_INBOUND_WHITELIST', [])
```
imdadahad: Any reason why we didn't have the same config name as the environment variable?

tests/app/authentication/test_authentication.py
```diff
@@ -313,8 +313,8 @@ def __create_token(service_id):
 def restrict_ip_sms_app():
     app = flask.Flask(__name__)
     app.config['TESTING'] = True
-    app.config['ALLOW_IP_INBOUND_SMS'] = ['134.213.243.188']
-
+    app.config['ALLOW_IP_INBOUND_SMS'] = ['111.111.111.111', '100.100.100.100']
+    # app.config['ALLOW_IP_INBOUND_SMS'] = os.environ['SMS_INBOUND_WHITELIST']
```
imdadahad: Comment to remove

tests/app/test_cloudfoundry_config.py
```diff
@@ -197,3 +198,10 @@ def test_redis_config():
 
     assert os.environ['REDIS_ENABLED'] == '1'
     assert os.environ['REDIS_URL'] == 'redis url'
+
+
+@pytest.mark.usefixtures('os_environ', 'cloudfoundry_environ')
+def test_sms_config():
```
imdadahad: `test_sms_inbound_config` instead?

PR #1084, opened 2017-07-12
Add logging to SMS delivery receipts

servingUpAces:
Added logging to show the entire form posted to us by the SMS client … ……providers.

This can be useful information when debugging what happened to a notificaiton.
Recently there was a discrepancy between the failure type used by each provider for a particular number, this logging would have helped.

### DIFF
--- a/app/notifications/notifications_sms_callback.py
+++ b/app/notifications/notifications_sms_callback.py
@@ -1,4 +1,5 @@
 from flask import Blueprint
+from flask import current_app
 from flask import json
 from flask import request, jsonify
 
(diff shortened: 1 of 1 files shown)


imdadahad:
Yikes, PEP

By the way if you add this `api/.git/hooks/pre-commit`:

```
#!/bin/bash
function display_result {
  RESULT=$1
  EXIT_STATUS=$2
  TEST=$3

  if [ $RESULT -ne 0 ]; then
    echo -e "\033[31m$TEST failed\033[0m"
    exit $EXIT_STATUS
  else
    echo -e "\033[32m$TEST passed\033[0m"
  fi
}

pycodestyle .
display_result $? 1 "Code style check"
```

It'll run the style check on commit, that way you shouldn't be able to get to PR stage with PEP errors.

PR #1085, opened 2017-07-12
Celery 4 (attempt 2)

leohemsted:
Brought in martyn's changes - had to do some hefty conflict resolution. Got rid of his definitions.py change cos it was just a bit too big.

* There are some *gross* problems related to imports. I'd like to sort them out before 
* Priority queue locally seems absolutely fine. sent_at - created_at is ~99% < 1 second. Am seeing prod-like times when running celery 3 (15, 20 seconds etc).

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -49,7 +49,6 @@ def create_app(app_name=None):
     from app.config import configs
 
     notify_environment = os.environ['NOTIFY_ENVIRONMENT']
-
     application.config.from_object(configs[notify_environment])
 
     if app_name:
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -13,7 +13,6 @@ from notifications_utils.clients.redis.redis_client import RedisClient
 from notifications_utils import logging, request_id
 from werkzeug.local import LocalProxy
 
-from app.celery.celery import NotifyCelery
 from app.clients import Clients
 from app.clients.email.aws_ses import AwsSesClient
 from app.clients.sms.firetext import FiretextClient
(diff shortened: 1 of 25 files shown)


app/__init__.py
```diff
@@ -26,6 +25,11 @@
 DATE_FORMAT = "%Y-%m-%d"
 
 db = SQLAlchemy()
+
+# avoid circular import by importing after db
+from app.celery.celery import NotifyCelery
+
```
imdadahad: You can move this back above (as discussed).

app/celery/celery.py
```diff
@@ -1,11 +1,130 @@
...
+
+class CeleryConfig:
+    def __init__(self, config):
+        self.broker_transport_options['queue_name_prefix'] = config['NOTIFICATION_QUEUE_PREFIX']
+        self.broker_url = config.get('BROKER_URL', 'sqs://')
+
+    broker_transport_options = {
+        'region': 'sqs.eu-west-1',
+        'polling_interval': 1,  # 1 second
+        'visibility_timeout': 310,
```
imdadahad: Do we still need this `polling_interval' if now we're now using long polling.

leohemsted: ¯\_(ツ)_/¯

leohemsted: long polling means that SQS will keep the request open until it gets a message - but after it returns, celery still needs to know how long to wait until asking for the next one - this is the `polling_interval` (i think)

imdadahad: 👍

imdadahad:
This needs `pycurl` as I get `The curl client requires the pycurl library` when trying to start celery.

PR #1089, opened 2017-07-19
Consider one-off notifications to be jobs

quis:
When we get back the notifications for a service there is the option to filter out those created by a job.

Previously this excluded one-off notifications, because they created a job. Now that they are created directly by the API they are not getting filtered out.

This commit changes the filter to consider a one-off notification to still be created by a job because:
- it keeps the behaviour the same as before
- it means the API integration page is no longer weirdly showing one-off
  messages

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -339,6 +339,7 @@ def get_notifications_for_service(
 
     if not include_jobs or (key_type and key_type != KEY_TYPE_NORMAL):
         filters.append(Notification.job_id.is_(None))
+        filters.append(Notification.api_key_id.isnot(None))
 
     if key_type is not None:
         filters.append(Notification.key_type == key_type)
--- a/tests/app/dao/test_notification_dao.py
+++ b/tests/app/dao/test_notification_dao.py
@@ -836,12 +836,12 @@ def test_get_notification_by_id(notify_db, notify_db_session, sample_template):
     assert notification_from_db.scheduled_notification.scheduled_for == datetime(2017, 5, 5, 14, 15)
 
 
-def test_get_notifications_by_reference(notify_db, notify_db_session, sample_service):
+def test_get_notifications_by_reference(notify_db, notify_db_session, sample_service, sample_api_key):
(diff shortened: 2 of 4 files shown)


tests/app/dao/test_notification_dao.py
```diff
@@ -1302,15 +1313,20 @@ def test_dao_timeout_notifications_doesnt_affect_letters(sample_letter_template)
...
 
     job = sample_job(notify_db, notify_db_session)
     with_job = sample_notification(
-        notify_db, notify_db_session, created_at=datetime.utcnow(), status="delivered", job=job
+        notify_db, notify_db_session, created_at=datetime.utcnow(), status="delivered", api_key_id=None
     )
     without_job = sample_notification(
-        notify_db, notify_db_session, created_at=datetime.utcnow(), status="delivered"
+        notify_db, notify_db_session, created_at=datetime.utcnow(), status="delivered", api_key_id=sample_api_key.id
     )
```
leohemsted: can you make this have three notifications? a job, an api, and a one-off? with their job/api_key layout as we'd expect to see in real life

app/dao/notifications_dao.py
```diff
@@ -339,6 +339,7 @@ def get_notifications_for_service(
 
     if not include_jobs or (key_type and key_type != KEY_TYPE_NORMAL):
         filters.append(Notification.job_id.is_(None))
+        filters.append(Notification.api_key_id.isnot(None))
```
leohemsted: We now have all four combinations of api key and job id in the system! 🎉 

If you keep `job_id.is_(None)` then this'll exclude letters from the API, since they have a job_id.
So this should just be `api_key_id != None`

|            | api key |  job id | 
|------------|---------|---------| 
| csv        | null    |         | 
| api        |         | null    | 
| letter api |         |         | 
| one off    | null    | null    |

leohemsted:
@quis this has been open for a month now 😬

servingUpAces:
This is too old now. We are going to have a design session about this and write a story for this.

PR #1091, opened 2017-07-19
Month billing table

servingUpAces:
Monthly billing - part 1 
This is still a work in progress but it would be good to get some eyes on it. This commit includes creating and updating a row in the monthly billing table and a method to fetch the results. There is a command to populate the monthly billing for a service and month so we can try it out. The total cost at the moment are wrong, they do not take into account the free allowance - see notes below about adding that to the table. Left to do: create a nightly task to run to update the monthly totals. create an endpoint to return the yearly billing, the current day will need to be calculated on the fly and added to the totals. Add the free allowance into the total costs.


app/dao/provider_rates_dao.py
```diff
@@ -9,3 +9,8 @@ def create_provider_rates(provider_identifier, valid_from, rate):
 
     provider_rates = ProviderRates(provider_id=provider.id, valid_from=valid_from, rate=rate)
     db.session.add(provider_rates)
+
+
+@transactional
+def create_sms_rate(rate):
+    db.session.add(rate)
```
leohemsted: i don't think we really need this function, especially since it's only used in the test fn create_rate. i'd just stick the db.session.add in there

tests/app/dao/test_monthly_billing.py
```diff
@@ -0,0 +1,126 @@
...
+
+def assert_monthly_billing(monthly_billing, year, month, service_id, expected_len, first_row):
+    assert monthly_billing.year == year
+    assert monthly_billing.month == month
+    assert monthly_billing.service_id == service_id
+    assert len(monthly_billing.monthly_totals) == expected_len
+    assert sorted(monthly_billing.monthly_totals[0]) == sorted(first_row)
+
+
+def test_get_service_id():
```
leohemsted: when using the `create_*` functions, an unfortunate downside is that they don't automatically trigger the db cleanup. You'll need to add `notify_db_session` as a parameter to this function. 

The others are fine since they still have `sample_template`, which requires `notify_db_session`

tests/app/dao/test_provider_rates_dao.py
```diff
@@ -15,3 +16,11 @@ def test_create_provider_rates(notify_db, notify_db_session, mmg_provider):
     assert ProviderRates.query.first().rate == rate
     assert ProviderRates.query.first().valid_from == now
     assert ProviderRates.query.first().provider_id == provider.id
+
+
+def test_create_sms_rate():
```
leohemsted: need `notify_db_session` here to ensure test isolation

app/commands.py
```diff
@@ -146,3 +147,18 @@ def fix_notification_statuses_not_in_sync(self):
             print('Committed {} updates at {}'.format(len(result), datetime.utcnow()))
             db.session.commit()
             result = db.session.execute(subq_hist).fetchall()
+
+
+class PopulateMonthlyBilling(Command):
+        option_list = (
+            Option('-s', '-service-id', dest='service_id',
+                   help="Service id to populate monthly billing for"),
```
servingUpAces: The intention of is to start testing it.
After that it might be useful if we feel we need to re-run a month that is not the current month.
But once we are happy this is working we can write one to build all the data for all services. Once we have an idea of how long it will take.

app/dao/monthly_billing_dao.py
```diff
@@ -0,0 +1,57 @@
...
+def _monthly_billing_data_to_json(monthly):
+    # ('April', 6, 1, False, 'sms', 0.014)
+    #  (month, billing_units, rate_multiplier, international, notification_type, rate)
+    # total cost must take into account the free allowance.
+    # might be a good idea to capture free allowance in this table
+    return [{"billing_units": x[1],
+             "rate_multiplier": x[2],
+             "international": x[3],
+             "rate": x[5],
+             "total_cost": (x[1] * x[2]) * x[5]} for x in monthly]
```
servingUpAces: Ah - that's better. I'll add that - thanks

app/commands.py
leohemsted: cool 👍

PR #1094, opened 2017-07-20
Stop Pyup trying to upgrade us to Celery 4

quis:


requirements.txt
```diff
@@ -13,7 +13,7 @@ flask-marshmallow==0.6.2
 Flask-Bcrypt==0.6.2
 credstash==1.8.0
 boto3==1.4.4
-celery==3.1.25
+celery==3.1.25 # pyup <4
```
leohemsted: https://pyup.io/docs/filter/
apologies, think this needs to be 
`celery==3.1.25 # pyup: <4`

PR #1134, opened 2017-07-26
Update the last template usage query to check Notification table

imdadahad:
This removes querying from the `NotificationHistory` to check last usage of a given template.

Querying from the `NotificationHistory` table has caused timeouts in multiple places. Let's remove that from here as it's good enough to know (on the admin) that a given template hasn't been used for more than seven days, as opposed to the actual time after seven days.

## Dependencies

- [x] https://github.com/alphagov/notifications-admin/pull/1389


imdadahad:
Added in 82ac4ae

PR #1136, opened 2017-07-27
Letter api

leohemsted:
🎉 

Not too complicated.

there are three steps to this 
1. Create a job  
    * Starts in status 'ready to send'
    * Created by None, since it's from the API
    * original file name 'letter submitted via api'
2. Create a single notification for that job
    * job_row_number 0 
    * client reference if provided
    * address_line_1 as recipient (other variants like `addressline1` or `address-line-1` are not allowed)
3. Trigger the build_dvla_file task
We know that all the notifications have been created for this job (since we just created them ourselves synchronously), so this will just create the dvla-format file for the job, and upload it to s3.

To support this, this PR also alters the jobs schema to make `created_by_id` nullable (since jobs from the API have no user to be created by. The downgrade is potentially un-runnable, if there are jobs with no created_by - if we need to downgrade a system with data, I'll leave it until then to work out what to do with those jobs.

(remember `?w=1` removes whitespace from git diffs)

TODO:

- [ ] test keys shouldn't send letters - they should not trigger the build-dvla-file task, and should instead just update the job and notification straight to dispatched status


app/dao/jobs_dao.py
```diff
@@ -108,6 +109,8 @@ def dao_get_future_scheduled_job_by_id_and_service_id(job_id, service_id):
 
 
 def dao_create_job(job):
+    if not job.id:
+        job.id = uuid.uuid4()
```
imdadahad: Why not put this in `create_letter_api_job`?

tests/app/notifications/test_process_letter_notifications.py
```diff
@@ -0,0 +1,83 @@
...
+    assert job.notification_count == 1
+    assert job.job_status == JOB_STATUS_READY_TO_SEND
+    assert job.created_by is None
+
+
+def test_create_letter_notification_creates_notification(sample_letter_job, sample_api_key):
+    data = {
+        'personalisation': {
+            'address_line_1': 'The Queen',
+            'address_line_2': 'Buckingham Palace',
```
imdadahad: Have to make sure this one gets delivered!

tests/app/notifications/test_process_letter_notifications.py
```diff
@@ -0,0 +1,83 @@
...
+from app.models import Notification
+from app.notifications.process_letter_notifications import create_letter_api_job
+from app.notifications.process_letter_notifications import create_letter_notification
+from app.v2.errors import InvalidRequest
+
+from tests.app.db import create_service
+from tests.app.db import create_template
+
+
+def test_create_job_rejects_inactive_service(notify_db_session):
```
imdadahad: Test for service without letter permission?

tests/app/v2/notifications/test_post_letter_notifications.py
```diff
@@ -1,17 +1,20 @@
...
 import pytest
 
-from app.models import Job, Notification, SMS_TYPE, EMAIL_TYPE, LETTER_TYPE
+from app.models import EMAIL_TYPE
+from app.models import Job
+from app.models import LETTER_TYPE
+from app.models import Notification
+from app.models import SMS_TYPE
 from app.v2.errors import RateLimitError
+from app.v2.notifications.post_notifications import process_letter_notification
```
imdadahad: `process_letter_notification` unused

app/notifications/process_letter_notifications.py
```diff
@@ -0,0 +1,44 @@
...
+from app.v2.errors import InvalidRequest
+
+
+def create_letter_api_job(template):
+    service = template.service
+    if not service.active:
+        raise InvalidRequest('Service {} is inactive'.format(service.id), 403)
+    if template.archived:
+        raise InvalidRequest('Template {} is deleted'.format(template.id), 400)
+
```
imdadahad: permissions check too?

app/notifications/process_letter_notifications.py
```diff
@@ -0,0 +1,44 @@
...
+    job = Job(
+        original_file_name='letter submitted via api',
+        service=service,
```
imdadahad: _It would be nice_ to give it some unique name, even if it's just a timestamp concatenated with something.

tests/app/notifications/rest/test_send_notification.py
```diff
@@ -572,128 +572,124 @@ def test_should_not_send_sms_if_team_api_key_and_not_a_service_user(notify_api,
...
+    data = {
+        'to': sample_email_template.service.created_by.email_address,
+        'template': sample_email_template.id
+    }
+    api_key = ApiKey(service=sample_email_template.service,
+                     name='team_key',
+                     created_by=sample_email_template.created_by,
+                     key_type=KEY_TYPE_TEAM)
+    save_model_api_key(api_key)
+    auth_header = create_jwt_token(secret=api_key.secret, client_id=str(api_key.service_id))
```
imdadahad: Is there any api key fixture we can use here?

app/dao/jobs_dao.py
leohemsted: mainly because this felt like a reasonable time to provide it, so that anywhere else we use `dao_create_job` is handled as well

app/notifications/process_letter_notifications.py
leohemsted: already happens here https://github.com/alphagov/notifications-api/blob/master/app/v2/notifications/post_notifications.py#L45

app/dao/jobs_dao.py
imdadahad: Yeah just wondering if there's anywhere else currently where this path would be hit, in other words if `create_letter_api_job` is the only place we need it then makes sense to do there?

app/notifications/process_letter_notifications.py
imdadahad: 👍

tests/app/notifications/rest/test_send_notification.py
leohemsted: turns out the `create_authorization_header` creates api keys of the appropriate type if they don't exist, so yes, good thinking

tests/app/notifications/test_process_letter_notifications.py
leohemsted: https://github.com/alphagov/notifications-api/blob/master/tests/app/v2/notifications/test_post_letter_notifications.py#L146

app/notifications/process_letter_notifications.py
leohemsted: we need to give it a name we can easily filter out so that it doesn't show up on the admin 😩

PR #1140, opened 2017-07-31
Remove credstash without breaking celery

leohemsted:
removed aws_run_celery because it's confusing and doesn't need to exist anymore

don't remove `notify_celery` import!

### DIFF
--- a/manifest-delivery-base.yml
+++ b/manifest-delivery-base.yml
@@ -33,7 +33,7 @@ applications:
       NOTIFY_APP_NAME: delivery-worker-research
 
   - name: notify-delivery-worker-sender
-    command: scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=11 -Q send-tasks,send-sms-tasks,send-email-tasks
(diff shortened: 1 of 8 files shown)


manifest-delivery-base.yml
```diff
@@ -16,39 +16,39 @@ memory: 1G
...
 
   - name: notify-delivery-worker-research
-    command: scripts/run_app_paas.sh celery -A aws_run_celery.notify_celery worker --loglevel=INFO --concurrency=5 -Q research-mode-tasks
+    command: scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=5 -Q research-mode-tasks
     env:
       NOTIFY_APP_NAME: delivery-worker-research
 
   - name: notify-delivery-worker-sender
-    command: scripts/run_app_paas.sh celery -A aws_run_celery.notify_celery worker --loglevel=INFO --concurrency=11 -Q send-tasks,send-sms-tasks,send-email-tasks
+    command: scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=11 -Q send-tasks,send-sms-tasks,send-email-tasks
```
imdadahad: We should probably remove `send-tasks` and the `SEND_COMBINED` etc. in the app as we're not using those anymore?

PR #1141, opened 2017-07-31
Fix issue where monthly billing data was not being updated

imdadahad:
This fixes the issue where the monthly billing data was not being updated in some cases when running the db command.

## Issue

`create_or_update_monthly_billing_sms` did not query the `MonthlyBilling` table with the `service_id`. This meant that we would not create a new `MonthlyBilling` entry if one existed already with the same `start date` or `notification_type`.

## Fix

Add the `service_id` in the filter.

### DIFF
--- a/tests/app/dao/test_monthly_billing.py
+++ b/tests/app/dao/test_monthly_billing.py
@@ -1,15 +1,38 @@
-from datetime import datetime
-
+from datetime import datetime, timedelta
 from freezegun import freeze_time
 from freezegun.api import FakeDatetime
 
+from app import db
 from app.dao.monthly_billing_dao import (
     create_or_update_monthly_billing_sms,
+    get_monthly_billing_entry,
     get_monthly_billing_sms,
     get_service_ids_that_need_sms_billing_populated
 )
-from app.models import MonthlyBilling
+from app.models import MonthlyBilling, SMS_TYPE
 from tests.app.db import create_notification, create_rate, create_service, create_template
+
+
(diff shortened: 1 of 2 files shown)


tests/app/dao/test_monthly_billing.py
```diff
@@ -136,3 +159,34 @@ def test_get_service_id(notify_db_session):
...
+def test_get_monthly_billing_entry_filters_by_service(notify_db, notify_db_session):
+    service_1 = create_service(service_name="Service One")
+    service_2 = create_service(service_name="Service Two")
+    now = datetime.utcnow()
+
+    create_sample_monthly_billing_entry(
+        notify_db,
+        notify_db_session,
+        service_1.id,
+        'sms',
```
kentsanggds: This could use the SMS_TYPE

tests/app/dao/test_monthly_billing.py
```diff
@@ -136,3 +159,34 @@ def test_get_service_id(notify_db_session):
...
+
+    create_sample_monthly_billing_entry(
+        notify_db,
+        notify_db_session,
+        service_2.id,
+        'sms',
+        [],
+        now,
+        now + timedelta(days=30)
+    )
```
kentsanggds: I wonder if this can be made clearer by adding the key to describe the value when passing the arg in? 

Also perhaps the `create_sample_monthly_billing_entry` can have defaults set as the only difference between the 2 entries is the service `id`

PR #1149, opened 2017-08-02
Stop trying to upgrade us to Celery Pyup

quis:


quis:
Should close and delete https://github.com/alphagov/notifications-api/pull/1147 once this is merged.

PR #1156, opened 2017-08-04
Add task for ses results queue

sakisv:
This abstracts the code that handles the SES callbacks to allow it to be reused from a celery task.

The tests are also updated to reflect this change, calling the process function directly instead of via the endpoint. 

The endpoint is going to be removed in a subsequent PR, once the subscriptions to it have been removed from AWS.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -44,6 +44,7 @@ from app.models import (
     JOB_STATUS_READY_TO_SEND,
     JOB_STATUS_SENT_TO_DVLA, JOB_STATUS_ERROR)
 from app.notifications.process_notifications import persist_notification
+from app.notifications.notifications_ses_callback import process_ses_response
 from app.service.utils import service_allowed_to_send_to
 from app.statsd_decorators import statsd
 from notifications_utils.s3 import s3upload
@@ -418,3 +419,9 @@ def send_inbound_sms_to_service(self, inbound_sms_id, service_id):
                                service_id, inbound_api.url, e))
             except self.MaxRetriesExceededError:
                 current_app.logger.exception('Retry: send_inbound_sms_to_service has retried the max number of times')
+
+
+@notify_celery.task(bind=True, name='process-ses-result')
+@statsd(namespace="tasks")
+def process_ses_results(self, response):
+    process_ses_response(response)
--- a/app/notifications/notifications_ses_callback.py
+++ b/app/notifications/notifications_ses_callback.py
@@ -28,7 +28,11 @@ register_errors(ses_callback_blueprint)
 
 @ses_callback_blueprint.route('/notifications/email/ses', methods=['POST'])
 def sns_callback_handler():
-    process_ses_response(json.loads(request.data))
+    errors, status, kwargs = process_ses_response(json.loads(request.data))
+    if errors:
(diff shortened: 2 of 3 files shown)


app/notifications/notifications_ses_callback.py
```diff
@@ -96,14 +99,12 @@ def process_ses_response():
 
             create_outcome_notification_statistic_tasks(notification)
 
-            return jsonify(
-                result="success", message="SES callback succeeded"
-            ), 200
+            return [], 200, {'result': "success", 'message': "SES callback succeeded"}
```
leohemsted: so, there are two states here:

1. Success. Neither SES or celery care about any response, but we should make sure the request to SES returns 200.

1. Failure. Neither SES or celery care about any response, but we should make sure we log the reason appropriately. Not sure anyone cares about the response code

So I think with these, we can get away with success returning `None`, and error states returning just the error message

Then, in the celery function, we can say

```python
resp = process_ses_response(...)
if resp:
    logger.error(resp)
```

or something

app/celery/tasks.py
```diff
@@ -418,3 +419,9 @@ def send_inbound_sms_to_service(self, inbound_sms_id, service_id):
                                service_id, inbound_api.url, e))
             except self.MaxRetriesExceededError:
                 current_app.logger.exception('Retry: send_inbound_sms_to_service has retried the max number of times')
+
+
+@notify_celery.task(bind=True, name='process-ses-result')
+@statsd(namespace="tasks")
+def process_ses_results(self, response):
+    process_ses_response(response)
```
leohemsted: this should probably have some kind of retry logic - what happens when things go bad? 

i suppose the nice thing about putting this on the queue instead of posting to the app, is that the app staying up isn't a thing we need to worry about if they go onto the queue (sns retried if we 500'd before).

But if we have a problem (realistically, that problem would be DB connectivity) then it'd be good if this retried a few times.

sakisv:
Just found that we can throttle the delivery rate on an SNS topic:

http://docs.aws.amazon.com/sns/latest/dg/DeliveryPolicies.html#delivery-policy-maximum-receive-rate

I think it makes more sense to do that instead of introducing new/more moving pieces. (I only wish I had found that sooner)

app/notifications/notifications_ses_callback.py
```diff
@@ -27,26 +26,30 @@
...
-            current_app.logger.info("Automatically subscribed to topic: {}".format(subscribed_topic))
-            return jsonify(
-                result="success", message="SES callback succeeded"
-            ), 200
+        # TODO: remove this check once the sns_callback_handler is removed
+        if not isinstance(ses_request, dict):
+            ses_request = json.loads(ses_request)
 
         errors = validate_callback_data(data=ses_request, fields=['Message'], client_name=client_name)
         if errors:
```
richardc0: I assume the errors get logged up the chain somewhere?

app/celery/tasks.py
richardc0: If you use a try block you can retry after a time period.

try
....
except SNSExecption as exc:
...         # Retry in 1 minute.
...         raise self.retry(countdown=60, exc=exc)

PR #1164, opened 2017-08-09
Check service name is unique

servingUpAces:
Created a new service endpoint that checks if the service name or email_from is unique.

Why is this needed?
When a user updates a service name they enter the new name in a form, are then asked to confirm the change by entering their password. Then the API call to update_service is called. If we let the update service API call fail with the integrity constraint it will be awkward for the user.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -596,3 +596,27 @@ def handle_sql_errror(e):
 def create_one_off_notification(service_id):
     resp = send_one_off_notification(service_id, request.get_json())
     return jsonify(resp), 201
+
+
+@service_blueprint.route('/unique', methods=["GET"])
+def is_service_name_unique():
+    name, email_from = check_request_args(request)
+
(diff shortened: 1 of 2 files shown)


app/service/rest.py
```diff
@@ -596,3 +596,27 @@ def handle_sql_errror(e):
...
+
+@service_blueprint.route('/unique', methods=["GET"])
+def is_service_name_unique():
+    name, email_from = check_request_args(request)
+
+    name_exists = Service.query.filter_by(name=name).first()
+    email_from_exists = Service.query.filter_by(email_from=email_from).first()
+    if name_exists or email_from_exists:
+        return jsonify(result=False), 200
+    return jsonify(result=True), 200
```
leohemsted: can get rid of the `if` by doing something like

```
result = not (name_exists or email_from_exists)
return jsonify(result=result), 200
```

PR #1165, opened 2017-08-09
don't capture logs directly from stdout

leohemsted:
previously in run_app_paas.sh, we captured stdout from the app and piped that into the log file. However, this came up with a bunch of problems, mainly:

* exceptions with stack traces often weren't formatted properly, and kibana could not parse them
* celery logs were duplicated - we'd collect both the json logs and the human readable stdout logs.

instead, with the updated utils library, we can use that to log json straight to the appropriate directory directly.

- [x] https://github.com/alphagov/notifications-utils/pull/192


leohemsted:
ran through with sakis and imdad

PR #1177, opened 2017-08-14
Fix order of DB migrations

quis:
More migrations got merged in between me making 0114 and merging the PR that contained 0114.

If we named the files as just 0114.py then this would get flagged as a merge conflict…


quis:
Demo’d to @kentsanggds

PR #1174, opened 2017-08-11
Add inbound number rest

kentsanggds:
## What - 

This is 2/6 in https://www.pivotaltracker.com/story/show/146774373, and adds the REST endpoints for the inbound_numbers DAO.

## How to review -

- pytest tests/app/inbound_numbers passes

PR #1178, opened 2017-08-15
Use inbound number for delivering sms and to show on service settings

kentsanggds:
This is step 3 / 7 - https://www.pivotaltracker.com/story/show/146774373

- Returns the `inbound_number.number` as the `sms_sender` when it is set, otherwise it will return the value in `sms_sender`
- Will send an sms using the `inbound_number.number` when inbound_number is set

NOTE - Only merge this after - 
- [x] - https://github.com/alphagov/notifications-api/pull/1174 is reviewed as smaller PR
- [x] - Scripts to inject new inbound numbers and set the correct service to the inbound number is set in step 4 / 7
---
- End to end tests - curl `receive_mmg_sms` and `receive_firetext_sms` endpoint with -
>- Does not process the inbound number -
>- [x] - inbound_number not set to service
>- [x] - service without inbound_sms permissions
>> service linked to inbound_number
>- [x] - inbound_number not active
>> service linked to inbound_number, service sms_sender set to inbound number
>- [x] - number not in inbound number table
---
>- Processes the inbound number
>- [x] - inbound_number set to service
>- [x] - service with sms_sender set with inbound number
>> not yet linked to inbound_number table, number active


app/dao/inbound_numbers_dao.py
```diff
@@ -4,7 +4,7 @@
 
 
 def dao_get_inbound_numbers():
-    return InboundNumber.query.all()
+    return InboundNumber.query.order_by(InboundNumber.updated_at, InboundNumber.number).all()
```
servingUpAces: I would order the query by one or the other. I would as @pete what order he prefers.

tests/app/inbound_number/test_rest.py
```diff
@@ -0,0 +1,117 @@
...
+        service_id=sample_service.id
+    )
+    assert result['data'] == inbound_number.serialize()
+
+
+def test_rest_get_inbound_number_when_service_is_not_assigned_returns_empty_dict(
+        admin_request, notify_db_session, sample_service):
+    result = admin_request.get(
+        'inbound_number.get_inbound_number_for_service',
+        service_id=sample_service.id
```
servingUpAces: should we expect a status code?

app/dao/services_dao.py
```diff
@@ -65,9 +66,22 @@ def dao_fetch_service_by_id(service_id, only_active=False):
     return query.one()
 
 
+############
+# refactor this when API only uses inbound_numbers and not sms_sender
+############
 def dao_fetch_services_by_sms_sender(sms_sender):
+    inbound_number = InboundNumber.query.filter(
```
servingUpAces: what is the inbound_number is inactive?

tests/app/inbound_number/test_rest.py
kentsanggds: Already checked in the admin_request fixture for 200

app/dao/inbound_numbers_dao.py
kentsanggds: I think updated_at makes most sense in that case

app/dao/services_dao.py
kentsanggds: I'll add a check for an active inbound_number

app/dao/inbound_numbers_dao.py
pete: I appreciate whenever my opinion is requested, but I am unfamiliar with the code and this have no preference in this case. Thank you.

PR #1194, opened 2017-08-21
Exclude test keys when searching by recipient

quis:
The activity page doesn’t show notifications sent with a test key.

However it _does_ when you search by recipient. This is confusing and inconsistent.


imdadahad:
Worth adding a test.

quis:
@imdadahad Done.

PR #1196, opened 2017-08-22
Add scheduled letter job

kentsanggds:
## What -

- Added a scheduled job for letters in `ready to send` state to add ids to a queue for processing by the `notifications-ftp` 
- Refactored the filename for letter-jobs to follow standard naming conventions

- [x] - Need to have logging in place so that any problems with letters can be identified early


app/celery/scheduled_tasks.py
```diff
@@ -302,3 +303,15 @@ def populate_monthly_billing():
     start_date, end_date = get_month_start_and_end_date_in_utc(yesterday_in_bst)
     services = get_service_ids_that_need_billing_populated(start_date=start_date, end_date=end_date)
     [create_or_update_monthly_billing(service_id=s.service_id, billing_month=end_date) for s in services]
+
+
+@notify_celery.task(name="run-letter-jobs")
+@statsd(namespace="tasks")
+def run_letter_jobs():
+    ids = dao_get_letter_jobs_by_status(JOB_STATUS_READY_TO_SEND)
```
imdadahad: Being really nit picky, worth renaming to `job_ids` just so it's clear what we're passing to `send_task`

app/celery/scheduled_tasks.py
```diff
@@ -302,3 +303,15 @@ def populate_monthly_billing():
...
     [create_or_update_monthly_billing(service_id=s.service_id, billing_month=end_date) for s in services]
+
+
+@notify_celery.task(name="run-letter-jobs")
+@statsd(namespace="tasks")
+def run_letter_jobs():
+    job_ids = [job.id for job in dao_get_letter_jobs_by_status(JOB_STATUS_READY_TO_SEND)]
+    notify_celery.send_task(
+        name=QueueNames.DVLA_FILES,
+        args=(job_ids),
```
imdadahad: I think we should differentiate between queue names and tasks names. As it's slightly awkward to call `send_task` with the task name being set as the queue name. 

For now we could just put the task name like we do elsewhere and then look at maybe putting this into a `TaskNames` class?

Also, think we should put this task on the `QueueNames.NOTIFY` as we need to make sure the API doesn't read from the `PROCESS_FTP` queue.

kentsanggds: OK, though I will leave the task on `QueueNames.PROCESS_FTP`
Also added a test to ensure that queue names are not accidentally added to `all_queues()`

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -28,6 +31,7 @@
     timeout_notifications,
     populate_monthly_billing)
 from app.clients.performance_platform.performance_platform_client import PerformancePlatformClient
+from app.config import QueueNames, TaskNames
```
imdadahad: 👍 on `TaskNames`

tests/app/test_config.py
```diff
@@ -57,3 +58,20 @@ def test_load_config_if_cloudfoundry_not_available(monkeypatch, reload_config):
 def test_cloudfoundry_config_has_different_defaults():
     # these should always be set on Sandbox
     assert config.Sandbox.REDIS_ENABLED is False
+
+
+def test_queue_names_all_queues_correct():
+    # Need to ensure that all_queues() only returns queue names used in API
+    queues = QueueNames.all_queues()
+    assert len(queues) == 10
+    assert set([QueueNames.PRIORITY,
```
imdadahad:
```
set([
   QueueNames.PRIORITY,
   ...
])
```

Think that looks better?

kentsanggds: ok, have updated the code

PR #1201, opened 2017-08-23
run tests in multiple threads at once

leohemsted:
previously we didn't do this because the tests all used the same DB (test_notifications_api), however @minglis shared a snippet that simply creates one test db per thread. If you run without using `run_tests` (or `py.test -n #`) then it'll create a db `test_notifications_api_master`


leohemsted:
@sakisv this isn't playing nice with postgres in docker - not sure why. Something to do with the psycopg2 connection being weird i guess /shrug

leohemsted:
It's running much quicker now - ~5.5 minutes instead of ~11minutes, that's good. Lets merge it.

PR #1205, opened 2017-08-23
Perf stats refactor

leohemsted:
With a second performance stats endpoint coming, we needed a refactor - we were previously only dealing with one endpoint and one bearer token, but now we have many, so they should be in a cloudfoundry config.

Also, a lot of refactoring happened, moving stuff out of performance_platform_client that wasn't specifically generic performance platform client stuff, so that's gone in its own file in `app/performance_platform/`

this needs to be deployed at around the same time as: 

- [ ] https://github.com/alphagov/notifications-aws/pull/225
- [ ] https://github.com/alphagov/notifications-credentials/pull/10

it needs to happen after AWS, but before any other api deploys happen

### DIFF
--- a/app/performance_platform/total_sent_notifications.py
+++ b/app/performance_platform/total_sent_notifications.py
@@ -4,26 +4,20 @@ from app import performance_platform_client
 from app.dao.notifications_dao import get_total_sent_notifications_in_date_range
 from app.utils import (
     get_london_midnight_in_utc,
-    get_midnight_for_day_before,
-    convert_utc_to_bst,
+    get_midnight_for_day_before
 )
 
 
-def send_total_notifications_sent_for_day_stats(date, channel, count, period):
-    payload = {
-        '_timestamp': convert_utc_to_bst(date).isoformat(),
-        'service': 'govuk-notify',
-        'channel': channel,
-        'count': count,
-        'dataType': 'notifications',
-        'period': period
-    }
-    performance_platform_client.add_id_to_payload(payload)
-
-    performance_platform_client.send_stats_to_performance_platform(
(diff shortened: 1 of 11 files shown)


app/performance_platform/total_sent_notifications.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def send_total_notifications_sent_for_day_stats(date, channel, count, period):
+    payload = {
+        '_timestamp': convert_utc_to_bst(date).isoformat(),
+        'service': 'govuk-notify',
+        'channel': channel,
+        'count': count,
+        'dataType': 'notifications',
+        'period': period
```
leohemsted: hardcode this to `'day'`

app/performance_platform/total_sent_notifications.py
```diff
@@ -0,0 +1,45 @@
...
+    }
+    performance_platform_client.add_id_to_payload(payload)
```
leohemsted: can we assign `_timestamp` in here

leohemsted: `performance_platform_client.create_payload({'channel': ..., 'count': ..., 'dataType': ...})`

look into this and how it can be generic - remove perf platform specific stuff from this business logic file if possible

manifest-api-base.yml
```diff
@@ -1,4 +1,4 @@
----
+#---
```
leohemsted: kill this comment

manifest-api-base.yml
```diff
@@ -10,6 +10,7 @@ services:
   - firetext
   - hosted-graphite
   - redis
+  - performance-platform
```
leohemsted: should be in manifest-delivery-base.yml as well

PR #1213, opened 2017-08-25
Remove Swagger specification

quis:
This was only ever a spike into what it might look like to document Notify’s API with Swagger (see 7c3d25a87a4c32b8616944ae7aceeb8e29bd70b0). It’s no longer updated, and only talks about version 1 of the public API.

Keeping it around now is just a liability, and gives us additional Pyup upgrades to deal with.

Closes #1207


tests/app/spec/test_rest.py
```diff
@@ -1,18 +0,0 @@
-import flex
```
leohemsted: think we can get rid of `flex` from `requirements_for_test.txt` now too

PR #1222, opened 2017-08-30
Get performance stats

klssmith:
We now send the performance platform both the total number of sms and email notifications sent via the api, and the total number sent within 10 seconds.

Paired with @leohemsted


tests/app/dao/notification_dao/test_notification_dao_performance_platform.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_get_total_notifications_filters_on_date_at_midnight(sample_template):
+    create_notification(sample_template, created_at=datetime(2016, 10, 17, 23, 59, 59))
+    create_notification(sample_template, created_at=datetime(2016, 10, 18, 0, 0))
+    create_notification(sample_template, created_at=datetime(2016, 10, 18, 23, 59, 59))
+    create_notification(sample_template, created_at=datetime(2016, 10, 19, 0, 0))
+    result = dao_get_total_notifications_sent_per_day_for_perfomance_platform(date(2016, 10, 18), date(2016, 10, 19))
+    assert result.messages_total == 2
+
```
imdadahad: Can we combine`test_get_total_notifications_filters_on_date` and `test_get_total_notifications_filters_on_date_at_midnight`, seems like they're just testing that returned notifications are within the date range.

app/dao/notifications_dao.py
```diff
@@ -519,3 +521,37 @@ def set_scheduled_notification_to_processed(notification_id):
...
+    coalesce(sum(CASE WHEN sent_at - created_at <= interval '10 seconds' THEN 1 ELSE 0 END), 0)
+    FROM notifications
+    WHERE
+    created_at > 'START DATE' AND
+    created_at < 'END DATE' AND
+    api_key_id IS NOT NULL AND
+    key_type != 'test' AND
+    notification_type != 'letter';
+    """
+    under_10_secs = Notification.sent_at - Notification.created_at <= timedelta(seconds=10)
```
leohemsted: this function should be using `NotificationHistory` rather than `Notification` everywhere (including in the SQL docstring), because we don't want to count research mode notifications (they don't follow the normal queues so shouldn't affect performance), but there's currently no easy way of telling research mode notifications apart from real notifications, except that research mode notifications don't persist to the `NotificationHistory` table

PR #1223, opened 2017-08-31
Warn devs about changing healthcheck page response

quis:


quis:
Not a functional change – going to merge without deploying.

PR #1226, opened 2017-08-31
Backfill perf stats

leohemsted:
to run this on a server:`cf run-task notify-api "python application.py backfill_processing_time -s='2017-07-01' -e='2017-09-01'"`

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -16,6 +16,8 @@ from app.dao.services_dao import (
 )
 from app.dao.provider_rates_dao import create_provider_rates
 from app.dao.users_dao import (delete_model_user, delete_user_verify_codes)
+from app.utils import get_midnight_for_day_before, get_london_midnight_in_utc
+from app.performance_platform.processing_time import send_processing_time_for_start_and_end
 
 
 class CreateProviderRateCommand(Command):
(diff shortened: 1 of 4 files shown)


app/commands.py
```diff
@@ -167,33 +169,62 @@ def link_inbound_numbers_to_service(self):
...
+        )
+        print("Finished populating data for {} for service id {}".format(month, str(service_id)))
+        print('SMS: {}'.format(sms_res.monthly_totals))
+        print('Email: {}'.format(email_res.monthly_totals))
+
+
+class BackfillProcessingTime(Command):
+    option_list = (
+        Option('-s', '--start_date', dest='start_date', help="Date (%Y-%m-%d) start date inclusive"),
+        Option('-e', '--end_date', dest='end_date', help="Date (%Y-%m-%d) end date inclusive"),
```
imdadahad: It's not inclusive of the end date right

leohemsted: this function is. I call `delta.days + 1` on L218 to specifically make it inclusive.

imdadahad: apologies, missed the commit message.

leohemsted:
good point 👍

PR #1238, opened 2017-09-08
[3/5] Migrate existing data to new service_email_reply_to table

imdadahad:
Branched off #1237 and will be rebased once that's merged in.

This adds a command that will migrate all the existing data to the new `service_email_reply_to` table.

## Summary

As it stands, any service that updates their `reply_to_email_address` will have an 'upserted` entry in 
`service_email_reply_to`. However services that do not, still need an entry in this new table if their `reply_to_email_address` has been set. 

This adds a command that will do that. There are two key rules:

1. If the `reply_to_email_address` is `NULL` for a service, we do not want to create an entry
2. If an entry already exists for a service in `service_email_reply_to`, don't populate the new table as  it's already there

**UPDATE: Rebased off master**

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -228,3 +228,23 @@ class BackfillProcessingTime(Command):
                 process_end_date.isoformat()
             ))
             send_processing_time_for_start_and_end(process_start_date, process_end_date)
+
+
+class PopulateServiceEmailReplyTo(Command):
+
+    def run(self):
+        services_to_update = """
+            INSERT INTO service_email_reply_to(id, service_id, email_address, is_default, created_at)
+            SELECT '{}', id, reply_to_email_address, true, '{}'
+            FROM services
+            WHERE reply_to_email_address IS NOT NULL
+            AND id NOT IN(
+                SELECT service_id
+                FROM service_email_reply_to
+            )
+        """.format(uuid.uuid4(), datetime.utcnow())
+
+        result = db.session.execute(services_to_update)
+        db.session.commit()
+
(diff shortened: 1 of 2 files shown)


app/commands.py
```diff
@@ -228,3 +228,23 @@ def run(self, start_date, end_date):
...
             ))
             send_processing_time_for_start_and_end(process_start_date, process_end_date)
+
+
+class PopulateServiceEmailReplyTo(Command):
+
+    def run(self):
+        services_to_update = """
+            INSERT INTO service_email_reply_to(id, service_id, email_address, is_default, created_at)
+            SELECT '{}', id, reply_to_email_address, true, '{}'
```
servingUpAces: Have you tried the select statement? not in can be slow, just check it's ok

imdadahad: I think it's okay purely because the `service_email_reply_to` will only contain entries for services that have updated their email since #1237, which will only be a couple at most.

servingUpAces: fair enough

app/commands.py
```diff
@@ -228,3 +228,23 @@ def run(self, start_date, end_date):
...
+            FROM services
+            WHERE reply_to_email_address IS NOT NULL
+            AND id NOT IN(
+                SELECT service_id
+                FROM service_email_reply_to
+            )
+        """.format(uuid.uuid4(), datetime.utcnow())
```
servingUpAces: Using uuid.uuid4() here will not work for a insert select statement, because it will use the same uuid for each row.

PR #1239, opened 2017-09-08
Update service_sms_sender for services

servingUpAces:
This PR populates the new table service_sms_sender any time a service is created or a service updates the sms sender.

There should only be one service_sms_sender per service and only for new or updated services.

### DIFF
--- a/app/dao/service_sms_sender_dao.py
+++ b/app/dao/service_sms_sender_dao.py
@@ -1,0 +1,19 @@
+from app import db
+from app.models import ServiceSmsSender
+
+
+def update_service_sms_sender(service, sms_sender):
+    result = db.session.query(
+        ServiceSmsSender
+    ).filter(
+        ServiceSmsSender.service_id == service.id
+    ).update(
+        {'sms_sender': sms_sender}
+    )
+    if result == 0:
+        new_sms_sender = ServiceSmsSender(sms_sender=sms_sender,
+                                          service=service,
+                                          is_default=True
+                                          )
+        db.session.add(new_sms_sender)
+        db.session.commit()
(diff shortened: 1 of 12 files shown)


app/dao/service_sms_sender_dao.py
```diff
@@ -0,0 +1,18 @@
+from app import db
+from app.models import ServiceSmsSender
+
+
+def update_service_sms_sender(service, sms_sender):
```
imdadahad: Do we need transactional here?

tests/app/dao/test_service_sms_sender_dao.py
```diff
@@ -0,0 +1,16 @@
+from app.dao.service_sms_sender_dao import update_service_sms_sender
+from app.models import ServiceSmsSender
+from tests.app.db import create_service
+
+
```
imdadahad: Should we have a test that checks we create a row if one doesn't already exist?

app/dao/service_sms_sender_dao.py
servingUpAces: That's what I wanted to chat about. The update and create service is wrapped in transactional which is why I left it out here. Not sure if this will be called without update but it might. Can they both have transactional?

tests/app/dao/test_service_sms_sender_dao.py
servingUpAces: That check is in the create service. Maybe I  should move this method.

app/dao/service_sms_sender_dao.py
imdadahad: Yeah I think it's okay, it's just if we use it elsewhere that isn't wrapped in transactional. E.g. When we do this in `update_service` on API?

tests/app/dao/test_service_sms_sender_dao.py
imdadahad: Oh, so are we testing it elsewhere?

PR #1241, opened 2017-09-08
Letters are accepted, not created or sending

leohemsted:
`accepted` to be specific.

From there, they don't change at all when we send to DVLA - but may go to `technical failure` if we mess up somewhere

### DIFF
--- a/tests/app/v2/notifications/test_get_notifications.py
+++ b/tests/app/v2/notifications/test_get_notifications.py
@@ -1,6 +1,6 @@
 import datetime
 import pytest
-from flask import json
+from flask import json, url_for
 
 from app import DATETIME_FORMAT
 from tests import create_authorization_header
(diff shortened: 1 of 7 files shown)


tests/app/v2/notifications/test_get_notifications.py
```diff
@@ -498,3 +505,35 @@ def test_get_all_notifications_filter_multiple_query_parameters(client, sample_e
...
+    auth_header = create_authorization_header(service_id=sample_letter_notification.service_id)
+    response = client.get(
+        path=url_for('v2_notifications.get_notifications'),
+        headers=[('Content-Type', 'application/json'), auth_header]
+    )
+
+    json_response = json.loads(response.get_data(as_text=True))
+    assert response.status_code == 200
+
+    for noti in json_response['notifications']:
```
servingUpAces: I'd be tempted to check email notification as well.

leohemsted:
need to check:

* can you filter by accepted when calling `getNotificationById`?
* what do they show up as on the front end?
 - on the job status page?
 - on the api page?

PR #1252, opened 2017-09-13
Inbound sms monitoring update

venusbb:
Using 24 bit mask to monitor IP traffic for inbound sms.

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -44,6 +44,12 @@ def requires_no_auth():
 
 
 def restrict_ip_sms():
+    # Check route of inbound sms (Experimental)
+    # Temporary custom header for route security
+    if request.headers.get("X-Custom-forwarder"):
+        current_app.logger.info("X-Custom-forwarder {}".format(request.headers.get("X-Custom-forwarder")))
+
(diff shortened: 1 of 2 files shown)


app/authentication/auth.py
```diff
@@ -54,21 +60,29 @@ def restrict_ip_sms():
...
-    if request.headers.get("X-Custom-forwarder"):
-        current_app.logger.info("X-Custom-forwarder {}".format(request.headers.get("X-Custom-forwarder")))
+    # IP whitelist
+    allowed_ips = current_app.config.get('SMS_INBOUND_WHITELIST')
+    allowed = False
+
+    for allowed_ip in allowed_ips:
+        p1 = allowed_ip.split('.')
+        if p0[0] == p1[0] and p0[1] == p1[1] and p0[2] == p1[2]:
+            allowed = True
```
leohemsted: i'd rather we use the [ipaddress library](https://docs.python.org/3/library/ipaddress.html) for this.

And then, we should update the `SMS_INBOUND_WHITELIST` variable to contain address masks (eg `192.168.0.0/24`) rather than whole IPs.

PR #1254, opened 2017-09-14
fixing minor problem, logged ip variable name changed

venusbb:


venusbb:
This is a minor fixed to the name of one of the monitored variable.

app/authentication/auth.py
```diff
@@ -80,7 +80,7 @@ def restrict_ip_sms():
                 'message': 'Inbound sms ip address',
                 'log_contents': {
                     'passed': allowed,
-                    'ip_address': ip
+                    'ip_address': inbound_ip
```
leohemsted: can we get rid of the `ip` variable since it appears to just be an unused empty string now

venusbb: Yes

PR #1257, opened 2017-09-15
Platform admin Timeout problem

venusbb:
When viewing platform admin services, we do an sql queries and whether it's trial mode or live services is filtered at the admin views. 

This PR pass the filtered live/trial mode service to the query and reduce query time. 

This is in relation to PR https://github.com/alphagov/notifications-admin/pull/1491

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -358,14 +358,19 @@ def dao_fetch_monthly_historical_stats_for_service(service_id, year):
 
 
 @statsd(namespace='dao')
-def dao_fetch_todays_stats_for_all_services(include_from_test_key=True):
+def dao_fetch_todays_stats_for_all_services(include_from_test_key=True, trial_mode_services=True):
+
+    # service_id = db.session.query(Service.id).filter(Service.restricted == True)
+
     query = db.session.query(
         Notification.notification_type,
         Notification.status,
         Notification.service_id,
         func.count(Notification.id).label('count')
+    ).join(Service
     ).filter(
-        func.date(Notification.created_at) == date.today()
+        func.date(Notification.created_at) == date.today(),
+        Service.restricted == trial_mode_services
     ).group_by(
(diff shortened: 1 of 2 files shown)


app/dao/services_dao.py
```diff
@@ -358,14 +358,20 @@ def dao_fetch_monthly_historical_stats_for_service(service_id, year):
 
 
 @statsd(namespace='dao')
-def dao_fetch_todays_stats_for_all_services(include_from_test_key=True):
+def dao_fetch_todays_stats_for_all_services(include_from_test_key=True, trial_mode_services=True):
+
+    # service_id = db.session.query(Service.id).filter(Service.restricted == True)
```
servingUpAces: No need for this commented line.

app/dao/services_dao.py
```diff
@@ -358,14 +358,20 @@ def dao_fetch_monthly_historical_stats_for_service(service_id, year):
 
 
 @statsd(namespace='dao')
-def dao_fetch_todays_stats_for_all_services(include_from_test_key=True):
+def dao_fetch_todays_stats_for_all_services(include_from_test_key=True, trial_mode_services=True):
```
servingUpAces: I would have thought that live services was the default.

app/dao/services_dao.py
```diff
@@ -358,14 +358,20 @@ def dao_fetch_monthly_historical_stats_for_service(service_id, year):
...
         Notification.notification_type,
         Notification.status,
         Notification.service_id,
         func.count(Notification.id).label('count')
+    ).join(
+        Service
     ).filter(
-        func.date(Notification.created_at) == date.today()
+        func.date(Notification.created_at) == date.today(),
+        Service.restricted == trial_mode_services
```
servingUpAces: We need to check the performance of this query as it is introducing a join to services. 
We have used a subquery to instead of a join to make the query more efficient.

PR #1258, opened 2017-09-15
Add multiple reply to email endpoints

servingUpAces:
This PR introduces new endpoints to insert multiple reply to email addresses for a service and update a single reply to email address for a service.

Note to the reviewer: there is some complexity around making sure there is always at least one default. Any suggestions to reduce the complexity is welcome.

### DIFF
--- a/app/dao/service_email_reply_to_dao.py
+++ b/app/dao/service_email_reply_to_dao.py
@@ -75,10 +75,10 @@ def _get_existing_default(service_id):
         if len(old_default) == 1:
             return old_default[0]
         else:
-            # is this check necessary
-            raise InvalidRequest(
+            raise Exception(
                 "There should only be one default reply to email for each service. Service {} has {}".format(
                     service_id, len(old_default)))
+    return None
 
 
 def _reset_old_default_to_false(old_default):
--- a/app/dao/service_email_reply_to_dao.py
+++ b/app/dao/service_email_reply_to_dao.py
@@ -40,7 +40,7 @@ def dao_update_reply_to_email(reply_to):
 
 
 @transactional
-def add_reply_to_email_address_for_service(service_id, email_address, is_default=True):
+def add_reply_to_email_address_for_service(service_id, email_address, is_default):
     old_default = _get_existing_default(service_id)
     if is_default:
         _reset_old_default_to_false(old_default)
@@ -53,7 +53,7 @@ def add_reply_to_email_address_for_service(service_id, email_address, is_default
 
 
 @transactional
-def update_reply_to_email_address(service_id, reply_to_id, email_address, is_default=True):
+def update_reply_to_email_address(service_id, reply_to_id, email_address, is_default):
     old_default = _get_existing_default(service_id)
     if is_default:
         _reset_old_default_to_false(old_default)
--- a/app/dao/service_email_reply_to_dao.py
+++ b/app/dao/service_email_reply_to_dao.py
@@ -9,7 +9,7 @@ def dao_get_reply_to_by_service_id(service_id):
         ServiceEmailReplyTo
     ).filter(
         ServiceEmailReplyTo.service_id == service_id
-    ).all()
+    ).order_by(ServiceEmailReplyTo.created_at).all()
     return reply_to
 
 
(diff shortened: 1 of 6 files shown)


app/dao/service_email_reply_to_dao.py
```diff
@@ -37,3 +37,57 @@ def dao_create_reply_to_email_address(reply_to_email):
...
+    reply_to_update = ServiceEmailReplyTo.query.get(reply_to_id)
+    reply_to_update.email_address = email_address
+    reply_to_update.is_default = is_default
+    db.session.add(reply_to_update)
+    return reply_to_update
+
+
+def _get_existing_default(service_id):
+    existing_reply_to = dao_get_reply_to_by_service_id(service_id=service_id)
+    if existing_reply_to:
```
leohemsted: i'd like us to explicitly `return None` if this if check fails, since we use the return value of this function elsewhere.

app/service/service_senders_schema.py
```diff
@@ -0,0 +1,11 @@
+add_service_email_reply_to_request = {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "description": "POST service email reply to address",
+    "type": "object",
+    "title": "Add new email reply to address for service",
+    "properties": {
+        "email_address": {"type": "string", "format": "email_address"},
+        "is_default": {"type": "boolean"}
```
richardc0: Cann you mark this as optional?

app/service/service_senders_schema.py
```diff
@@ -0,0 +1,11 @@
...
+    },
+    "required": ["email_address"]
```
leohemsted: would it make it easier if we demand that is_default is always set? i'm a fan of more restrictive inbound schemas in general. Not super fussed tho.

app/service/service_senders_schema.py
leohemsted: jsonschema knows it's optional because it's not in the `required` list below

app/dao/service_email_reply_to_dao.py
```diff
@@ -37,3 +37,57 @@ def dao_create_reply_to_email_address(reply_to_email):
...
+
+    reply_to_update = ServiceEmailReplyTo.query.get(reply_to_id)
+    reply_to_update.email_address = email_address
+    reply_to_update.is_default = is_default
+    db.session.add(reply_to_update)
+    return reply_to_update
+
+
+def _get_existing_default(service_id):
+    existing_reply_to = dao_get_reply_to_by_service_id(service_id=service_id)
```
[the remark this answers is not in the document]
servingUpAces: There are 0..n rows in the service_email_reply_to table, so it is not a problem if the service does not exist in the table.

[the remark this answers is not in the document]
servingUpAces: I could name the method as get_existing_default_or_none?

app/service/service_senders_schema.py
servingUpAces: I think this is ok because we assume that if it is not included in the schema then `is_default`=True

tests/app/service/test_rest.py
```diff
@@ -2198,3 +2198,72 @@ def test_get_email_reply_to_addresses_with_multiple_email_addresses(client, noti
...
+def test_update_service_reply_to_email_address_returns_400_when_no_default(client, sample_service):
+    original_reply_to = create_reply_to_email(service=sample_service, email_address="some@email.com")
+    data = json.dumps({"email_address": "changed@reply.com", "is_default": False})
+    response = client.post('/service/{}/email-reply-to/{}'.format(sample_service.id, original_reply_to.id),
+                           data=data,
+                           headers=[('Content-Type', 'application/json'), create_authorization_header()])
+
+    assert response.status_code == 400
+    json_resp = json.loads(response.get_data(as_text=True))
+    assert json_resp['message'] == 'You must have at least one reply to email address as the default.'
```
leohemsted: can we add a test into this file for what `test_add_service_400s_when_invalid_service_id` cos I think it'll go through all the process until you insert the row, which will throw an integrity error cos of the fkey constraint - good to make sure that gets handled properly

leohemsted: or maybe `404`?

app/service/service_senders_schema.py
servingUpAces: It is more straight forward

app/service/service_senders_schema.py
richardc0: is_default doesn't need to be optional on internal method now?

app/dao/service_email_reply_to_dao.py
```diff
@@ -75,10 +75,10 @@ def _get_existing_default(service_id):
         if len(old_default) == 1:
             return old_default[0]
         else:
-            # is this check necessary
-            raise InvalidRequest(
+            raise Exception(
```
richardc0: Catch a more specific Exception(s)?

app/dao/service_email_reply_to_dao.py
```diff
@@ -37,3 +37,57 @@ def dao_create_reply_to_email_address(reply_to_email):
...
+
+
+def _get_existing_default(service_id):
+    existing_reply_to = dao_get_reply_to_by_service_id(service_id=service_id)
+    if existing_reply_to:
+        old_default = [x for x in existing_reply_to if x.is_default]
+        if len(old_default) == 1:
+            return old_default[0]
+        else:
+            raise Exception(
```
richardc0: Catch a more specific Exception(s)?

PR #1264, opened 2017-09-20
Return placeholders when getting a template

quis:
> Currently when retrieving a template via one of the clients, we do not return the personalisation fields that are required for that template.
>
> This is useful for services who want to perform template validation on their own systems. A service user has also requested this.

– https://www.pivotaltracker.com/story/show/150674476

This commit adds an extra attribute to the JSON response containing an array of the placeholder names. This key is called "personalisation", to match the argument that developers use to pass in the values of placeholders.

### DIFF
--- a/tests/app/v2/template/test_get_template.py
+++ b/tests/app/v2/template/test_get_template.py
@@ -40,9 +40,55 @@ def test_get_template_by_id_returns_200(client, sample_service, tmp_type, expect
         'body': template.content,
         "subject": expected_subject,
         'name': expected_name,
+        'personalisation': [],
     }
 
     assert json_response == expected_response
+
+
+@pytest.mark.parametrize("create_template_args, expected_personalisation", [
+    (
+        {
+            "template_type": SMS_TYPE,
+            "content": "Hello ((placeholder)) ((conditional??yes))",
+        },
+        ["placeholder", "conditional"]
+    ),
+    (
+        {
+            "template_type": EMAIL_TYPE,
+            "subject": "((subject))",
+            "content": "((content))",
+        },
+        ["subject", "content"]
+    ),
+    (
+        {
+            "template_type": LETTER_TYPE,
+            "subject": "((letterSubject))",
+            "content": "((letter_content))",
+        },
+        ["letterSubject", "letter_content", "contact block"]
+    )
+])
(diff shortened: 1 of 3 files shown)


quis:
Needs client updates. The team who’ve asked for this are using Java.

tests/app/v2/template/test_get_template.py
```diff
@@ -40,11 +40,57 @@ def test_get_template_by_id_returns_200(client, sample_service, tmp_type, expect
...
     assert json_response == expected_response
 
 
+@pytest.mark.parametrize("create_template_args, expected_personalisation", [
+    (
+        {
+            "template_type": SMS_TYPE,
+            "content": "Hello ((placeholder)) ((conditional??yes))",
+        },
+        ["placeholder", "conditional"]
```
leohemsted: do we want to identify conditionals separately in the data we return? either as a separate list, or maybe a list of dicts `[{'name': 'placeholder', 'type': 'normal'}, {'name': 'conditional', type': optional'}]`, or maybe a dict of lists `{'normal': ['placeholder'], 'optional': ['conditional']}`?

quis:
## Return metadata about placeholders

In the future, we may want to return additional information about placeholders.  We came up with three possible formats:
1. list of `dict`s, eg `[{'name': 'first name', 'required': True}]`
2. `dict` of `list`s, eg `{'required': ['first name']}`
3. `dict` of `dict`s, eg `{'name': {'required': True}}`

I don’t like 1. because it’s harder to traverse if all you want is the name of the placeholders, and suggests that you could have two placeholders with the same name (which you can’t). I don’t like 2. because it only lets the data be sliced by one dimension (unless the inner lists aren’t exclusive, in which case you’d need to filter duplicates when just listing placeholders).

I think 3. has the two advantages of:
- represents that personalisation is unique, ie you can’t pass back in   two different values for the same key
- is forward compatible, ie we can add many more properties of a  placeholder without breaking anything 

The only real downside of 3. is that it loses the sort order of placeholders. We don’t think this information is that important.

So this commit implements 3.
--

tamasperger:
Java client has been updated - could you please update the checklist in the pivotal tracker?
https://www.pivotaltracker.com/story/show/150674476

PR #1269, opened 2017-09-21
Fix sending emails with no logo

quis:
Code was not expecting logo to be `None`, thereby causing the task to throw an exception, and retry until eventually putting the email in technical error (for services with org branding but no logo).


quis:
Demo’d to @leohemsted

PR #1271, opened 2017-09-21
Add ServiceLetterContact data model and migration script

kentsanggds:
## What - 

Step 1 / 3 from https://www.pivotaltracker.com/story/show/151287398 to create table to store letter contact details for services

- added data model & migration script
- added tests

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1406,6 +1406,8 @@ class ServiceLetterContact(db.Model):
 
     def serialize(self):
         return {
+            'id': str(self.id),
+            'service_id': str(self.service_id),
             'contact_block': self.contact_block,
             'is_default': self.is_default,
             'created_at': self.created_at.strftime(DATETIME_FORMAT),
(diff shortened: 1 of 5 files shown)


app/models.py
```diff
@@ -1381,3 +1389,25 @@ def serialize(self):
...
+    service = db.relationship(Service, backref=db.backref("letter_contacts"))
+
+    contact_block = db.Column(db.Text, nullable=False, index=False, unique=False)
+    is_default = db.Column(db.Boolean, nullable=False, default=True)
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    updated_at = db.Column(db.DateTime, nullable=True, onupdate=datetime.datetime.utcnow)
+
+    def serialize(self):
+        return {
+            'contact_block': self.contact_block,
```
servingUpAces: We'll need id and service_id in the serialize method

PR #1275, opened 2017-09-22
Platform Admin Timeout problem

venusbb:
**Why**
Platform Admin takes too long to load

**Changed**
1. /service/ end point pass live or trial mode as a parameter to limit the number of returned items. 
2. Simplified DetailedServiceSchema

**Story**

**Related PRs**
https://github.com/alphagov/notifications-admin/pull/1506

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -82,6 +82,7 @@ def get_services():
     detailed = request.args.get('detailed') == 'True'
     user_id = request.args.get('user_id', None)
     include_from_test_key = request.args.get('include_from_test_key', 'True') != 'False'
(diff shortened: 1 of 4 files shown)


app/service/rest.py
```diff
@@ -84,6 +84,7 @@ def get_services():
     detailed = request.args.get('detailed') == 'True'
     user_id = request.args.get('user_id', None)
     include_from_test_key = request.args.get('include_from_test_key', 'True') != 'False'
+    trial_mode_services = request.args.get('trial_mode_services')
```
servingUpAces: Add default here

PR #1277, opened 2017-09-25
[2/4] Upsert into ServiceLetterContact table when updating a service

imdadahad:
This upserts into the `ServiceLetterContact` table when updating a service as well as preserving the existing update to `Service.letter_contact_block`

## Summary 

1. This adds an 'upsert' dao method to insert or update in the new table when a service updates their letter contact block

2. This takes some of the code used in #1258 to manage defaults effectively. In short, when adding a `ServiceLetterContact` we want to ensure that there is only ever one entry that is the current default (`.is_default = True`).

3. When updating a service via the endpoint, this will call the DAO method from 1.

Pivotal Story: https://www.pivotaltracker.com/story/show/151287398

## Dependencies

- [x] - https://github.com/alphagov/notifications-admin/pull/1511

### DIFF
--- a/app/dao/service_letter_contact_dao.py
+++ b/app/dao/service_letter_contact_dao.py
@@ -28,7 +28,7 @@ def create_or_update_letter_contact(service_id, contact_block):
         letter_contacts[0].contact_block = contact_block
         dao_update_letter_contact(letter_contacts[0])
     else:
-        # Once we move allowing letter contact blocks, this method will be removed
+        # TODO: Once we move allowing letter contact blocks, this method will be removed
         raise InvalidRequest(
             "Multiple letter contacts were found, this method should not be used.",
             status_code=500
@@ -43,3 +43,65 @@ def dao_create_letter_contact(letter_contact):
 @transactional
 def dao_update_letter_contact(letter_contact):
     db.session.add(letter_contact)
+
+
+@transactional
+def add_letter_contact_for_service(service_id, contact_block, is_default):
+    old_default = _get_existing_default(service_id)
+    if is_default:
+        _reset_old_default_to_false(old_default)
+    else:
+        _raise_when_no_default(old_default)
+
+    new_letter_contact = ServiceLetterContact(
+        service_id=service_id,
+        contact_block=contact_block,
+        is_default=is_default
+    )
+    db.session.add(new_letter_contact)
+    return new_letter_contact
+
+
+@transactional
+def update_letter_contact(service_id, letter_contact_id, contact_block, is_default):
+    old_default = _get_existing_default(service_id)
+    # if we want to make this the default, ensure there are no other existing defaults
+    if is_default:
+        _reset_old_default_to_false(old_default)
+    else:
+        if old_default.id == letter_contact_id:
+            raise InvalidRequest("You must have at least one letter contact as the default.", 400)
+
+    letter_contact_update = ServiceLetterContact.query.get(letter_contact_id)
+    letter_contact_update.contact_block = contact_block
+    letter_contact_update.is_default = is_default
+    db.session.add(letter_contact_update)
+    return letter_contact_update
+
+
+def _get_existing_default(service_id):
+    letter_contacts = dao_get_letter_contacts_by_service_id(service_id=service_id)
+    if letter_contacts:
+        old_default = [x for x in letter_contacts if x.is_default]
+        if len(old_default) == 1:
+            return old_default[0]
+        else:
+            raise Exception(
+                "There should only be one default letter contact for each service. Service {} has {}".format(
+                    service_id,
+                    len(old_default)
+                )
+            )
+    return None
+
+
+def _reset_old_default_to_false(old_default):
+    if old_default:
+        old_default.is_default = False
+        db.session.add(old_default)
+
+
+def _raise_when_no_default(old_default):
+    # check that the update is not updating the only default to false
+    if not old_default:
+        raise InvalidRequest("You must have at least one letter contact as the default.", 400)
--- a/app/dao/service_letter_contact_dao.py
+++ b/app/dao/service_letter_contact_dao.py
@@ -1,0 +1,45 @@
+from app import db
+from app.dao.dao_utils import transactional
+from app.errors import InvalidRequest
+from app.models import ServiceLetterContact
+
+
+def dao_get_letter_contacts_by_service_id(service_id):
+    letter_contacts = db.session.query(
+        ServiceLetterContact
+    ).filter(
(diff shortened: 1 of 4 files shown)


app/dao/service_letter_contact_dao.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _get_existing_default(service_id):
+    letter_contacts = dao_get_letter_contacts_by_service_id(service_id=service_id)
+    if letter_contacts:
+        old_default = [x for x in letter_contacts if x.is_default]
+        if len(old_default) == 1:
+            return old_default[0]
+        else:
+            raise Exception(
```
richardc0: Is there a more specific exception to raise?

app/service/rest.py
```diff
@@ -140,15 +146,19 @@ def update_service(service_id):
...
-    if 'reply_to_email_address' in req_json:
-        create_or_update_email_reply_to(fetched_service.id, req_json['reply_to_email_address'])
-
     current_data = dict(service_schema.dump(fetched_service).data.items())
     current_data.update(request.get_json())
     update_dict = service_schema.load(current_data).data
+    dao_update_service(update_dict)
+
+    if 'reply_to_email_address' in req_json:
+        create_or_update_email_reply_to(fetched_service.id, req_json['reply_to_email_address'])
```
richardc0: Not sure how defensive we want to be but are we guaranteed that req_json['reply_to_email_address'] will contain a value i.e. would you want to pass in an empty string?

tests/app/dao/test_service_letter_contact_dao.py
```diff
@@ -0,0 +1,202 @@
...
+    create_or_update_letter_contact,
+    dao_get_letter_contacts_by_service_id,
+    update_letter_contact
+)
+from app.errors import InvalidRequest
+from app.models import ServiceLetterContact
+from tests.app.db import create_letter_contact, create_service
+
+
+def test_dao_get_letter_contacts_by_service_id(notify_db_session):
```
richardc0: Do we assume service_id is always valid? Is it worth a test with an invalid service_id?

tests/app/dao/test_service_letter_contact_dao.py
```diff
@@ -0,0 +1,202 @@
...
+    service = create_service()
+    create_letter_contact(service, 'London, NW1 2DB')
+    create_or_update_letter_contact(service.id, 'Bristol, BR1 2DB')
+
+    letter_contacts = dao_get_letter_contacts_by_service_id(service.id)
+
+    assert len(letter_contacts) == 1
+
+
+def test_create_or_update_letter_contact_updates_existing_entry(notify_db_session):
```
richardc0: You could delete the above test as its testing the same thing?

tests/app/dao/test_service_letter_contact_dao.py
```diff
@@ -0,0 +1,202 @@
...
+    service = create_service()
+    create_letter_contact(service=service, contact_block='Edinburgh, ED1 1AA')
+    create_letter_contact(service=service, contact_block='Manchester, MA1 2BB', is_default=False)
+
+    with pytest.raises(expected_exception=InvalidRequest) as e:
+        create_or_update_letter_contact(service_id=service.id, contact_block='Swansea, SN1 3CC')
+    assert e.value.message == "Multiple letter contacts were found, this method should not be used."
+
+
+def test_create_or_update_letter_contact_raises_exception_if_multiple_letter_contacts_exist(notify_db_session):
```
richardc0: This is the same code as the above test? Combine the tests in the method name?

tests/app/dao/test_service_letter_contact_dao.py
```diff
@@ -0,0 +1,202 @@
...
+    assert results[0].contact_block == 'Edinburgh, ED1 1AA'
+    assert results[0].is_default
+
+    assert results[1].contact_block == 'Swansea, SN1 3CC'
+    assert not results[1].is_default
+
+
+def test_add_letter_contact_with_no_default_raises_exception(notify_db_session):
+    service = create_service()
+    with pytest.raises(expected_exception=InvalidRequest):
```
richardc0: Do we usually check the exception message too?

tests/app/dao/test_service_letter_contact_dao.py
```diff
@@ -0,0 +1,202 @@
...
+            service_id=service.id,
+            contact_block='Swansea, SN1 3CC',
+            is_default=False
+        )
+
+
+def test_add_letter_contact_when_multiple_defaults_exist_raises_exception(notify_db_session):
+    service = create_service()
+    create_letter_contact(service=service, contact_block='Edinburgh, ED1 1AA')
+    create_letter_contact(service=service, contact_block='Aberdeen, AB12 23X')
```
richardc0: Would it be better to override the first as the default? I think this is the behaviour Becca adopted for multiple senders? Might be worth being consistent? (this might be consistent with somewhere else?

app/dao/service_letter_contact_dao.py
imdadahad: I don't think so but I think this is okay primarily because this code will be deprecated as we start working on adding code to create multiple letter contact blocks.

app/service/rest.py
imdadahad: As of https://github.com/alphagov/notifications-admin/pull/1511, we no longer allow `reply_to_email_address` to be passed with an empty value.

I don't think it's worth the change, simple because we'll be deprecating marshmallow at some point and this validation will be refactored/removed anyway to support adding multiple letter contact blocks.

tests/app/dao/test_service_letter_contact_dao.py
imdadahad: Validation for a correct service id happens in the endpoint. Not sure there'd be much value adding it here.

tests/app/dao/test_service_letter_contact_dao.py
imdadahad: Not necessarily, although it's calling the same function it is asserting that the update took place.

tests/app/dao/test_service_letter_contact_dao.py
imdadahad: This has a different input and checks if an exception is raised.

tests/app/dao/test_service_letter_contact_dao.py
imdadahad: I don't think it matters too much. Key thing is the purpose of the test is clear and the inherent code. I like to think this is pretty concise?

PR #1278, opened 2017-09-25
[3/4] - Populate service_letter_contacts from existing data

imdadahad:
This adds a command to populate the new `service_letter_contacts` from existing data pulled from `Service.letter_contact_block` for each service.

Conditions:
1. Only create a record if `Service.letter_contact_block` actually has something
2. service_letter_contacts hasn't already been populated for the service

## Dependencies

- [x] - https://github.com/alphagov/notifications-api/pull/1277
- [x] - https://github.com/alphagov/notifications-admin/pull/1511

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -285,3 +285,23 @@ class PopulateServiceSmsSender(Command):
         print("Populated sms sender {} services from services".format(second_result.rowcount))
         print("{} services in table".format(services_count_query))
         print("{} service_sms_senders".format(service_sms_sender_count_query))
+
+
+class PopulateServiceLetterContact(Command):
+
+    def run(self):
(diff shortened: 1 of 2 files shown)


app/commands.py
```diff
@@ -248,3 +248,23 @@ def run(self):
         db.session.commit()
 
         print("Populated email reply to adderesses for {}".format(result.rowcount))
+
+
+class PopulateServiceLetterContact(Command):
```
richardc0: Should there be a test for this?

klssmith: This command will only ever be run once in each environment - we'll test it by checking the database before and after running it to make sure it has worked as expected.

PR #1279, opened 2017-09-26
Enable Inbound sms IP blocking

venusbb:
**Why**
Secure inbound sms by IP whitelisting the sms providers' address 

**What's new**
Previously we have monitored the inbound sms traffic. In this update we enable blocking using the whitelist.  Also update the tests to include this additional header that we check. 

**Pivotal story**
146688097

### DIFF
--- a/tests/app/authentication/test_authentication.py
+++ b/tests/app/authentication/test_authentication.py
@@ -338,7 +338,7 @@ def test_allow_valid_ips(restrict_ip_sms_app):
     assert response.status_code == 200
 
 
-@pytest.mark.xfail(reason='Currently not blocking invalid IPs', strict=True)
+# @pytest.mark.xfail(reason='Currently not blocking invalid IPs', strict=True)
 def test_reject_invalid_ips(restrict_ip_sms_app):
     with pytest.raises(AuthError) as exc_info:
         restrict_ip_sms_app.get(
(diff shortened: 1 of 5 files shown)


tests/app/authentication/test_authentication.py
```diff
@@ -338,7 +338,7 @@ def test_allow_valid_ips(restrict_ip_sms_app):
     assert response.status_code == 200
 
 
-@pytest.mark.xfail(reason='Currently not blocking invalid IPs', strict=True)
+# @pytest.mark.xfail(reason='Currently not blocking invalid IPs', strict=True)
```
leohemsted: can you remove these xfail lines rather than commenting them out

tests/app/notifications/rest/test_callbacks.py
```diff
@@ -297,36 +271,19 @@ def test_firetext_callback_should_update_multiple_notification_status_sent(
...
-
-    client.post(
-        path='/notifications/sms/firetext',
-        data='mobile=441234123123&status=0&time=2016-03-10 14:17:00&reference={}'.format(
-            notification3.id
-        ),
-        headers=[('Content-Type', 'application/x-www-form-urlencoded')])
+    firetext_post(client, data)
+    firetext_post(client, data)
+    firetext_post(client, data)
```
leohemsted: this test isn't the same - before it was sending for three separate notifications one after the other, and now it's sending for the first notification three times.

this test has also never had any asserts 😱 

Can you make it send for the three notifications again, and also, assert something relevant?

Or alternatively we can remove it, I'm not sure how much value this test is providing.

PR #1280, opened 2017-09-26
Letter api not by job

leohemsted:
Sorry it's a bit of a beast!

### `POST /v2/notifications/letter/`

* No longer create a job
* If test/research mode, create the notification with status `sending`, and then trigger `update-notifications-to-sent-to-dvla`

### new scheduled task `run-letter-api-notifications`

* Runs at 5:40 (`run-letter-jobs runs` is at 5:30)
* gets all notifications that:
  - are letters
  - that were sent via the api
  - using a normal (live) api key
  - that are in status `created`
* sets those notifications to status `pending` to prevent sending twice if the task is triggered twice
* converts these notifications to dvla pipe file format, and concatenates them into one big file
* upload that to a new s3 bucket, `{}-dvla-letter-api-files` (https://github.com/alphagov/notifications-aws/pull/235), with the filename `{ISO_TIMESTAMP}-dvla-notifications.txt`
* tells the FTP app to send the notifications to dvla (https://github.com/alphagov/notifications-ftp/pull/24)

### update status to success error tasks

* two new tasks - `update-letter-notifications-to-error` and `update-letter-notifications-to-sent`
* they are given a list of notification references, and update status based on `reference in ({list_of_refs})`

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -325,9 +325,9 @@ def run_letter_jobs():
         current_app.logger.info("Queued {} ready letter job ids onto {}".format(len(job_ids), QueueNames.PROCESS_FTP))
 
 
-@notify_celery.task(name="run-letter-notifications")
+@notify_celery.task(name="run-letter-api-notifications")
 @statsd(namespace="tasks")
-def run_letter_notifications():
+def run_letter_api_notifications():
     current_time = datetime.utcnow().isoformat()
 
     notifications = dao_set_created_live_letter_api_notifications_to_pending()
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -335,17 +335,17 @@ def run_letter_notifications():
     if notifications:
         file_contents = create_dvla_file_contents_for_notifications(notifications)
 
-        file_location = '{}-dvla-notifications.txt'.format(current_time)
+        filename = '{}-dvla-notifications.txt'.format(current_time)
         s3upload(
             filedata=file_contents + '\n',
             region=current_app.config['AWS_REGION'],
             bucket_name=current_app.config['DVLA_BUCKETS']['notification'],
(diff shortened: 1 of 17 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -313,9 +316,41 @@ def populate_monthly_billing():
...
+        )
+        current_app.logger.info("Queued {} ready letter job ids onto {}".format(len(job_ids), QueueNames.PROCESS_FTP))
+
+
+@notify_celery.task(name="run-letter-api-notifications")
+@statsd(namespace="tasks")
+def run_letter_api_notifications():
+    current_time = datetime.utcnow().isoformat()
+
+    notifications = dao_set_created_live_letter_api_notifications_to_pending()
```
servingUpAces: What happens if there is an exception after marking these notifications as pending?
Do we need to build an alert?

leohemsted: i naively hope that we have an alert for uncaught exceptions in celery tasks, but will double check with sakis

leohemsted: we don't have an alert at the moment but it is straightforward for @sakisv  to add

app/dao/notifications_dao.py
```diff
@@ -474,6 +474,27 @@ def dao_update_notifications_sent_to_dvla(job_id, provider):
 
 
 @statsd(namespace="dao")
+@transactional
+def dao_update_notifications_by_reference(references, update_dict):
+    now = datetime.utcnow()
+    updated_count = Notification.query.filter(
+        Notification.reference.in_(references)
```
servingUpAces: The in_ does have a limit of how many it can take before breaking. Let's make sure this is under with the max number you will send this function.

leohemsted: tested with 10,000 random references, runs in under a sec including db setup/tear down stuff

PR #1282, opened 2017-09-27
[1/10] Allow API calls to specify the reply address option

richardc0:
- Added notification_to_email_reply_to link table to the model to link notifications and service_email_reply_to so that notifications are associated to a reply to address.
- Created database migration script to apply the changes.

https://www.pivotaltracker.com/story/show/150440099
Task 1/10


quis:
You don’t need to change anything on this pull request, but I’d suggest reading https://github.com/alphagov/styleguides/blob/master/git.md

app/models.py
```diff
@@ -1414,3 +1414,24 @@ def serialize(self):
             'created_at': self.created_at.strftime(DATETIME_FORMAT),
             'updated_at': self.updated_at.strftime(DATETIME_FORMAT) if self.updated_at else None
         }
+
+
+class NotificationSmsSender(db.Model):
```
kentsanggds: Should probably be NotificationEmailReplyTo?

app/models.py
```diff
@@ -1414,3 +1414,24 @@ def serialize(self):
...
+    __tablename__ = "notification_to_email_sender"
```
kentsanggds: Change this to `notification_to_email_reply_to`? Or something like that

migrations/versions/0123_add_noti_to_email_sender.py
```diff
@@ -0,0 +1,31 @@
+"""
+
+Revision ID: 0123_add_noti_to_email_sender
```
kentsanggds: Update the revision ID

migrations/versions/0123_add_noti_to_email_sender.py
```diff
@@ -0,0 +1,31 @@
...
+Revises: 0122_add_service_letter_contact
+Create Date: 2017-09-27 09:42:39.412731
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = '0123_add_noti_to_email_sender'
```
kentsanggds: Update the revision ID

app/models.py
richardc0: Yes updated

app/models.py
richardc0: Ok updated

migrations/versions/0123_add_noti_to_email_reply.py
```diff
@@ -0,0 +1,31 @@
...
+
+def upgrade():
+    op.create_table('notification_to_email_reply_to',
+    sa.Column('notification_id', postgresql.UUID(as_uuid=True), nullable=False),
+    sa.Column('service_email_reply_to_id', postgresql.UUID(as_uuid=True), nullable=False),
+    sa.ForeignKeyConstraint(['notification_id'], ['notifications.id'], ),
+    sa.ForeignKeyConstraint(['service_email_reply_to_id'], ['service_email_reply_to.id'], ),
+    sa.PrimaryKeyConstraint('notification_id', 'service_email_reply_to_id')
+    )
+    op.create_index(op.f('ix_notification_to_email_sender_notification_id'), 'notification_to_email_reply_to', ['notification_id'], unique=False)
```
servingUpAces: We expect there to be only one notification_id in the mapping table. We could make the index unique.

PR #1283, opened 2017-09-27
Unbind the db service before pushing the app

sakisv:
This should cause the app to get new credentials for the database upon
pushing.


leohemsted:
do we want to unbind other services?

(i'm mainly thinking about notify-config, will we have to rebind if we change config values?)

sakisv:
I added `notify-config` and `notify-aws`. Although, in that case we should restage it anyway.

PR #1288, opened 2017-10-02
Return service_sms_sender

servingUpAces:
Created a get_default_sms_sender method, which returns the default sms_sender from the service_sms_sender table rather than service.sms_sender.

One step closer to removing services.sms_sender.

**Pivotal Story:**
https://www.pivotaltracker.com/story/show/150880948

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -257,23 +257,15 @@ class Service(db.Model, Versioned):
 
     def get_default_sms_sender(self):
         default_sms_sender = [x for x in self.service_sms_senders if x.is_default]
-        if len(default_sms_sender) > 1:
-            raise Exception("There should only ever be one default")
         return default_sms_sender[0].sms_sender
 
     def get_default_reply_to_email_address(self):
         default_reply_to = [x for x in self.reply_to_email_addresses if x.is_default]
-        if len(default_reply_to) > 1:
-            raise Exception("There should only ever be one default")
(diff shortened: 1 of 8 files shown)


app/models.py
```diff
@@ -254,7 +254,13 @@ def get_inbound_number(self):
         if self.inbound_number and self.inbound_number.active:
             return self.inbound_number.number
         else:
-            return self.sms_sender
+            return self.get_default_sms_sender()
+
+    def get_default_sms_sender(self):
+        default_sms_sender = [x for x in self.service_sms_senders if x.is_default]
+        if len(default_sms_sender) > 1:
+            raise Exception("There should only ever be one default")
```
richardc0: Raise a more specific exception? I would probably made a custom model exception?

tests/app/test_model.py
```diff
@@ -279,3 +280,8 @@ def test_service_get_default_contact_letter(sample_service):
 # this test will need to be removed after letter_contact_block is dropped
 def test_service_get_default_letter_contact_block_from_service(sample_service):
     assert sample_service.get_default_letter_contact() == sample_service.letter_contact_block
+
+
+def test_service_get_default_sms_sender(notify_db_session):
```
richardc0: Can you test that an exception occurs if you have more that 1 entry?

PR #1291, opened 2017-10-03
[2/10] Allow API calls to specify the reply address option

richardc0:
- Updated notification schema to include email_reply_to_id so that it can be persisted
- Updated the end point in the post_notification to consume the new value in the schema and added a new validator to check the new email_reply_to_id is valid
- Added tests to test the new functionality added to the schema, code in the end point and unit tests for the validator


app/notifications/validators.py
```diff
@@ -132,3 +133,13 @@ def validate_template(template_id, personalisation, service, notification_type):
...
+
+
+def check_service_email_reply_to_id(service_id, reply_to_id):
+    if not (reply_to_id is None):
+        try:
+            reply_to = dao_get_reply_to_by_id(service_id, reply_to_id)
+        except NoResultFound:
+            message = 'reply_to_id does not exist in database'
+            raise BadRequestError(message=message,
+                                  fields=[{'template': message}])
```
servingUpAces: why is template a field in the message?

tests/app/notifications/test_validators.py
```diff
@@ -322,3 +324,26 @@ def test_allows_api_calls_with_international_numbers_if_service_does_allow_int_s
...
+
+def test_check_service_email_reply_to_id_where_reply_to_id_is_none():
+    assert check_service_email_reply_to_id(None, None) is None
+
+
+def test_check_service_email_reply_to_id_where_reply_to_id_is_not_found(sample_service, fake_uuid):
+    with pytest.raises(BadRequestError) as e:
+        check_service_email_reply_to_id(sample_service.id, fake_uuid)
+    assert e.value.status_code == 400
+    assert e.value.message == 'reply_to_id does not exist in database'
```
servingUpAces: The error message should match the name of the form parameter, email_reply_to_id

app/notifications/validators.py
richardc0: Removed

PR #1303, opened 2017-10-09
Purge notifications to email reply to rows older than 7 days

kentsanggds:
When we purge the notifications table we also need to purge any rows in the link table `notifications_email_reply_to` table as we will not be able to remove the notifications with the foreign key constraint.

Reference 

Step 5 - https://www.pivotaltracker.com/story/show/150440099

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -27,6 +27,7 @@ from app.dao.notifications_dao import (
     dao_timeout_notifications,
     is_delivery_slow_for_provider,
     delete_notifications_created_more_than_a_week_ago_by_type,
+    delete_notification_to_email_reply_to_more_than_a_week_ago,
     dao_get_scheduled_notifications,
     set_scheduled_notification_to_processed,
     dao_set_created_live_letter_api_notifications_to_pending,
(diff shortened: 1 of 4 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -119,6 +120,17 @@ def delete_sms_notifications_older_than_seven_days():
...
+        current_app.logger.info(
+            "Delete {} job started {} finished {} deleted {} notification to email reply to mappings".format(
+                'email',
+                start,
+                datetime.utcnow(),
+                deleted_noti_to_reply_to
+            )
+        )
+
+        start = datetime.utcnow()
```
richardc0: Is this needed again there is already a similar line on 122

app/dao/notifications_dao.py
```diff
@@ -26,14 +26,15 @@
...
     NotificationHistory,
     NotificationStatistics,
-    NotificationEmailReplyTo,
-    ServiceEmailReplyTo,
     ScheduledNotification,
+    ServiceEmailReplyTo,
     Template,
     KEY_TYPE_NORMAL,
     KEY_TYPE_TEST,
+    EMAIL_TYPE,
```
richardc0: Could be moved to alphabetical order.

app/dao/notifications_dao.py
```diff
@@ -377,6 +378,22 @@ def delete_notifications_created_more_than_a_week_ago_by_type(notification_type)
...
+    subq = db.session.query(Notification.id).filter(
+        func.date(Notification.created_at) < seven_days_ago
+    ).subquery()
+    deleted = db.session.query(
+        NotificationEmailReplyTo
+    ).filter(
+        NotificationEmailReplyTo.notification_id.in_(subq)
+    ).delete(synchronize_session='fetch')
+    db.session.commit()
+    return deleted
```
richardc0: Might be worth checking the subq length = the deleted length and logging if there is a difference.

or do a synchronize_session='evaluate' check the number match then run the actual delete?

app/dao/notifications_dao.py
kentsanggds: ok

app/dao/notifications_dao.py
kentsanggds: Will add this to `delete_notifications_created_more_than_a_week_ago_by_type` function to limit impact when there is an issue when deleting records

PR #1304, opened 2017-10-10
Added alert when job.notification_count doesn't match total notification for job

kentsanggds:
- Updated the process_started date when we update the job to in progress
- Added log for when a job starts so that we will know when the processing of a job starts with the number of notifications
- Added a celery task to check that the job is finished as a scheduled task that runs every minute, this may be too aggressive, but it only looks at job within a 5 minute window - to avoid reporting the in progress jobs forever.
- Added a `JobIncompleteError` to make it easier to find in the logs
- Next thing to do is to add a task to start processing these jobs.
Reference - https://www.pivotaltracker.com/story/show/151806308

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -371,7 +371,7 @@ def check_job_status():
     and scheduled_at or created_at is older that 30 minutes.
     if any results then
         raise error
-        process the rows in the csv thatgit c are missing (in another task) just do the check here.
+        process the rows in the csv that are missing (in another task) just do the check here.
     """
     thirty_minutes_ago = datetime.utcnow() - timedelta(minutes=30)
     thirty_five_minutes_ago = datetime.utcnow() - timedelta(minutes=35)
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -4,6 +4,7 @@ from datetime import (
 )
 
 from flask import current_app
+from sqlalchemy import or_, and_
 from sqlalchemy.exc import SQLAlchemyError
 from notifications_utils.s3 import s3upload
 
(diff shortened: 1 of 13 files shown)


app/celery/tasks.py
```diff
@@ -85,21 +86,30 @@ def process_job(job_id):
...
 
-    if template.template_type == LETTER_TYPE:
-        if service.research_mode:
-            update_job_to_sent_to_dvla.apply_async([str(job.id)], queue=QueueNames.RESEARCH_MODE)
-        else:
-            build_dvla_file.apply_async([str(job.id)], queue=QueueNames.JOBS)
-            current_app.logger.info("send job {} to build-dvla-file in the {} queue".format(job_id, QueueNames.JOBS))
+    notification_total_in_db = dao_get_total_notifications_for_job_id(job.id)
+
+    if job.notification_count != notification_total_in_db:
```
servingUpAces: This is always going to error. At this point we only have the notifications on a queue to be persisted, but they will not have been persisted yet (at least not all of them).

tests/app/test_config.py
```diff
@@ -74,5 +74,6 @@ def test_queue_names_all_queues_correct():
         QueueNames.STATISTICS,
         QueueNames.JOBS,
         QueueNames.RETRY,
-        QueueNames.NOTIFY
+        QueueNames.NOTIFY,
+        QueueNames.CHECK_JOBS,
```
servingUpAces: No need to create a new queue - we can use periodic here.

app/celery/tasks.py
```diff
@@ -85,6 +86,10 @@ def process_job(job_id):
     TemplateClass = get_template_class(db_template.template_type)
     template = TemplateClass(db_template.__dict__)
 
+    current_app.logger.info("Starting job {} processing {} notifications".format(job_id, job.notification_count))
+
+    check_job_status.apply_async([str(job.id)], queue=QueueNames.JOBS)
```
servingUpAces: Why is this here?

app/celery/scheduled_tasks.py
```diff
@@ -354,3 +357,30 @@ def run_letter_api_notifications():
...
+@notify_celery.task(name='check-job-status')
+@statsd(namespace="tasks")
+def check_job_status():
+    """
+    every x minutes do this check
+    select
+    from jobs
+    where job_status == 'in progress'
+    and template_type in ('sms', 'email')
+    and scheduled_at or created_at is older that 30 minutes.
```
richardc0: typo that -> then

PR #1305, opened 2017-10-11
[#151837054] Add new column free_sms_fragment_limit in the Services table

venusbb:
Prepare the back-end so that we can store different free_sms_fragment_limit values for different organisation types for billing purposes. 

- Add new database column
- Changed schema
- Made sure the existing read and write end points work before the change in the front end

Related Stories
https://www.pivotaltracker.com/story/show/151837054
https://www.pivotaltracker.com/story/show/151176604

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -305,3 +305,27 @@ class PopulateServiceLetterContact(Command):
         db.session.commit()
 
         print("Populated letter contacts for {} services".format(result.rowcount))
+
+
+class PopulateServiceAndServiceHistoryFreeSmsFragmentLimit(Command):
+
+    def run(self):
+        services_to_update = """
+            UPDATE services
+            SET free_sms_fragment_limit = 250000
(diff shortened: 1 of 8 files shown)


environment_test.sh
```diff
@@ -5,6 +5,7 @@ export NOTIFY_ENVIRONMENT='test'
 export ADMIN_CLIENT_SECRET='dev-notify-secret-key'
 export ADMIN_BASE_URL='http://localhost:6012'
 export FROM_NUMBER='from_number'
+export FREE_SMS_TIER_FRAGMENT_COUNT='250000'
```
servingUpAces: why is this necessary? FREE_SMS_TIER_FRAGMENT_COUNT is in config.py

klssmith: It wasn't necessary - now removed.

venusbb:
Add new free_sms_fragment_limit column to reflect organisation types changes

PR #1306, opened 2017-10-11
Add command to populate 'free sms fragment limit'

klssmith:
Added a one-off command to set values of the `free_sms_fragment_limit` in
the `services` table to the default of 250000.

_Note_ - only the last commit has been added - this will be rebased onto master when the branch it is based off is merged.

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -305,3 +305,27 @@ class PopulateServiceLetterContact(Command):
         db.session.commit()
 
         print("Populated letter contacts for {} services".format(result.rowcount))
+
+
+class PopulateServiceAndServiceHistoryFreeSmsFragmentLimit(Command):
+
+    def run(self):
+        services_to_update = """
+            UPDATE services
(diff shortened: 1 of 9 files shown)


app/commands.py
```diff
@@ -305,3 +305,18 @@ def run(self):
         db.session.commit()
 
         print("Populated letter contacts for {} services".format(result.rowcount))
+
+
+class PopulateServiceFreeSmsFragmentLimit(Command):
+
+    def run(self):
+        services_to_update = """
```
servingUpAces: Should we update services_history too?

venusbb:
A default free_sms_fragment_limit parameter is set at the end point /service/{id} where create_service is called. We need to set the default before the dao is called.

app/commands.py
klssmith: Now updated to update both tables

klssmith:
Closing, as this change has now been added here instead https://github.com/alphagov/notifications-api/pull/1305

PR #1310, opened 2017-10-12
Adding extra logging to celery tasks and gunicorn

richardc0:
-Added extra logging to celery and gunicorn tasks, specifically to capture when SIGTERM and SIGINIT are received, so that we can track better when an app restarts and if it has been restarted ungracefully.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -48,7 +48,7 @@ from app.utils import convert_utc_to_bst
 
 @worker_process_shutdown.connect
 def worker_process_shutdown(sender, signal, pid, exitcode):
-    current_app.logger.info('Scheduled tasks worker shutdown:: PID: {} Exitcode: {}'.format(pid, exitcode))
+    current_app.logger.info('Scheduled tasks worker shutdown: PID: {} Exitcode: {}'.format(pid, exitcode))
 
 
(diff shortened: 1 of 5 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -45,6 +46,11 @@
 from app.utils import convert_utc_to_bst
 
 
+@worker_process_shutdown.connect
+def worker_process_shutdown(sender, signal, pid, exitcode):
+    current_app.logger.info('Scheduled tasks worker shutdown:: PID: {} Exitcode: {}'.format(pid, exitcode))
```
sakisv: Typo: `::`

richardc0:
- There needs to be an annotated method in the task and its a single line method in each, so there doesn't seem to be a better way. There doesn't seem to be a server wide hook?
- Not particularly interested in the pid. They are available so thought it would be good to print out. The pids are printed out on start up so we can cross correlate etc.  If you really don't think it serves a purpose we remove it. 

Also the latest version of Celery has a new method worker_shutting_down which determines the shutdown method, warm or cold which could be useful.

app/celery/scheduled_tasks.py
richardc0: Doh! Done

PR #1324, opened 2017-10-17
Remove unique constraint service sms sender

servingUpAces:
This will allow us to store multiple sms senders per service.


servingUpAces:
Add this change with another PR

PR #1326, opened 2017-10-18
[#152067121] Create new table annual_billing

venusbb:
**Why:**
The introduction of different organisation types means different starting free sms allowance. This new table will stores all free_sms_fragment_limit per year per service.

**What**
* New migration script 0126
* New model
* Populate new table with past data

Pivotal story


migrations/versions/0126_add_annual_billing.py
```diff
@@ -0,0 +1,29 @@
...
+    op.create_table('annual_billing',
+    sa.Column('id', postgresql.UUID(as_uuid=True), nullable=False),
+    sa.Column('service_id', postgresql.UUID(as_uuid=True), nullable=False),
+    sa.Column('financial_year_start', sa.Integer(), nullable=False),
+    sa.Column('free_sms_fragment_limit', sa.Integer(), nullable=False),
+    sa.Column('created_at', sa.DateTime(), nullable=True),
+    sa.Column('updated_at', sa.DateTime(), nullable=True),
+    sa.ForeignKeyConstraint(['service_id'], ['services.id'], ),
+    sa.PrimaryKeyConstraint('id'))
+
```
servingUpAces: Should we add a unique constraint for service_id and financial_year_start?

venusbb: Done.

servingUpAces: Can you add the UniqueConstraint('financial_year_start', 'service_id', name='ix_annual_billing_service_id') to the model as well please?
If you don't when you run db migrate it will show up. The model and database schema need to match.

PR #1327, opened 2017-10-18
Add celery tasks- save_sms, save_email, save_letter

klssmith:
Created three new celery tasks:
* `save_sms`          (will replace `send_sms`)
* `save_email`        (will replace `send_email`)
* `save_letter`       (will replace `persist_letter`)

The difference between the new tasks and the tasks they are replacing is
that we no longer pass in the datetime as a parameter.

The code has been changed to use the new tasks, and the tests now run
against the new tasks too. The old tasks will need be removed in a separate
commit.

### DIFF
--- a/tests/app/celery/test_tasks.py
+++ b/tests/app/celery/test_tasks.py
@@ -19,9 +19,9 @@ from app.celery.tasks import (
     create_dvla_file_contents_for_job,
     process_job,
     process_row,
-    send_sms,
-    send_email,
-    persist_letter,
+    save_sms,
+    save_email,
+    save_letter,
     process_incomplete_job,
     process_incomplete_jobs,
     get_template_class,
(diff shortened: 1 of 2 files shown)


tests/app/celery/test_tasks.py
```diff
@@ -727,24 +708,20 @@ def test_should_not_send_email_if_team_key_and_recipient_not_in_team(sample_emai
     team_members = [user.email_address for user in sample_email_template_with_placeholders.service.users]
     assert "my_email@my_email.com" not in team_members
 
-    with freeze_time("2016-01-01 11:09:00.00000"):
```
richardc0: We could keep this test as is, remove the parameter and then check the created time matches the freeze time?

klssmith: As discussed in person, we'll leave time out of this test (because no notifications get created in this test, so we can't add in an assertion to check the creation time). In the tests where we are using `save_email` and the email notifications do get created we mostly include an assertion for the creation time, for example [here](https://github.com/alphagov/notifications-api/blob/0c9a4bce595f4e186c6e0fc4d29eb1439eec4103/tests/app/celery/test_tasks.py#L746) and [here](https://github.com/alphagov/notifications-api/blob/0c9a4bce595f4e186c6e0fc4d29eb1439eec4103/tests/app/celery/test_tasks.py#L817)

klssmith:
@richardc0 We've added new tasks instead of changing the existing ones because we wanted to change the task names anyway (the tasks aren't really sending anything)

PR #1333, opened 2017-10-23
Handle DVLA response file

kentsanggds:
## What

Currently letter responses are not being handled so we have no idea if a letter has been sent, this PR will update letter notifications to a `delivered` or `failed` status based on the DVLA response file.

- status returned in the response file should be either `Sent` or `Failed`
- rather than use an existing `delivered` status in the json response this will be mapped to `received` for requests. Internally we will still use `delivered` status.

## Reference - 

https://www.pivotaltracker.com/story/show/151649162

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -36,6 +36,7 @@ from app.dao.jobs_dao import (
     dao_update_job_status)
 from app.dao.notifications_dao import (
     get_notification_by_id,
+    update_notification_status_by_reference,
     dao_update_notifications_for_job_to_sent_to_dvla,
     dao_update_notifications_by_reference,
     dao_get_last_notification_added_for_job_id)
@@ -46,6 +47,7 @@ from app.dao.templates_dao import dao_get_template_by_id
 from app.models import (
     Job,
     Notification,
+    DVLA_STATUS_SENT,
     EMAIL_TYPE,
     JOB_STATUS_CANCELLED,
     JOB_STATUS_FINISHED,
@@ -55,6 +57,8 @@ from app.models import (
     JOB_STATUS_SENT_TO_DVLA, JOB_STATUS_ERROR,
     KEY_TYPE_NORMAL,
     LETTER_TYPE,
+    NOTIFICATION_DELIVERED,
+    NOTIFICATION_FAILED,
     NOTIFICATION_SENDING,
     NOTIFICATION_TECHNICAL_FAILURE,
     SMS_TYPE,
@@ -443,8 +447,21 @@ def update_letter_notifications_statuses(self, filename):
         raise
     else:
         for update in notification_updates:
-            current_app.logger.info('DVLA update: {}'.format(str(update)))
-            # TODO: Update notifications with desired status
+            status = NOTIFICATION_DELIVERED if update.status == DVLA_STATUS_SENT else NOTIFICATION_FAILED
+            notification = update_notification_status_by_reference(
+                update.reference,
+                status
+            )
+
+            if not notification:
+                msg = "Update letter notification file {filename} failed: notification either not found " \
+                    "or already updated from delivered. Status {status} for notification reference {reference}".format(
+                        filename=filename, status=status, reference=update.reference)
+                current_app.logger.error(msg)
+            else:
+                current_app.logger.info(
+                    'DVLA file: {filename}, notification updated to {status}: {reference}'.format(
+                        filename=filename, status=status, reference=str(update.reference)))
 
 
 def process_updates_from_file(response_file):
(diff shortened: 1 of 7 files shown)


app/celery/tasks.py
```diff
@@ -46,6 +47,7 @@
 from app.models import (
     Job,
     Notification,
+    DVLA_STATUS_SENT,
```
richardc0: There is already JOB_STATUS_SENT_TO_DVLA if this is different might be worth expanding it? I prefer your one so could consider changing the existing one?

app/models.py
```diff
@@ -876,6 +876,9 @@ def check_code(self, cde):
 
 NOTIFICATION_STATUS_LETTER_ACCEPTED = 'accepted'
 NOTIFICATION_STATUS_LETTER_ACCEPTED_PRETTY = 'Accepted'
+NOTIFICATION_STATUS_LETTER_RECEIVED = 'received'
```
richardc0: Should received by capitalised? Or is this defined by DVLA?

app/models.py
```diff
@@ -1044,6 +1056,7 @@ def formatted_status(self):
                 'technical-failure': 'Technical failure',
                 'sending': NOTIFICATION_STATUS_LETTER_ACCEPTED_PRETTY,
                 'created': NOTIFICATION_STATUS_LETTER_ACCEPTED_PRETTY,
+                'delivered': 'Received'
```
richardc0: Use the defined const? NOTIFICATION_STATUS_LETTER_RECEIVED?

tests/app/celery/test_ftp_update_tasks.py
```diff
@@ -7,10 +7,14 @@
 from app.models import (
     Job,
     Notification,
+    NOTIFICATION_DELIVERED,
+    NOTIFICATION_FAILED,
     NOTIFICATION_SENDING,
     NOTIFICATION_CREATED,
-    NOTIFICATION_TECHNICAL_FAILURE
+    NOTIFICATION_TECHNICAL_FAILURE,
+    NOTIFICATION_STATUS_LETTER_RECEIVED
```
richardc0: Could reorder these

tests/app/celery/test_ftp_update_tasks.py
```diff
@@ -91,6 +95,20 @@ def test_update_letter_notifications_statuses_builds_updates_list(notify_api, mo
     assert updates[1].cost_threshold == 'Sorted'
 
 
+def test_update_letter_notifications_statuses_persisted(notify_api, mocker, sample_letter_template):
+    sent_letter = create_notification(sample_letter_template, reference='ref-foo', status=NOTIFICATION_SENDING)
+    failed_letter = create_notification(sample_letter_template, reference='ref-bar', status=NOTIFICATION_SENDING)
+
+    valid_file = '{}|Sent|1|Unsorted\n{}|Failed|2|Sorted'.format(
```
richardc0: asset they are both in sending first?

tests/app/celery/test_ftp_update_tasks.py
```diff
@@ -91,6 +95,20 @@ def test_update_letter_notifications_statuses_builds_updates_list(notify_api, mo
     assert updates[1].cost_threshold == 'Sorted'
 
 
+def test_update_letter_notifications_statuses_persisted(notify_api, mocker, sample_letter_template):
```
richardc0: Add tests for neither Sent or Failed?

app/celery/tasks.py
kentsanggds: I think that's something different, this is handling letters at the notification level, not job

app/models.py
kentsanggds: Don't think we use capitalisation internally for the different statuses

app/models.py
kentsanggds: Defined const is in lower case, for it to be nicer looking it's capitalised here, and since it's only used here I haven't made it a variable

tests/app/celery/test_ftp_update_tasks.py
kentsanggds: OK

tests/app/celery/test_ftp_update_tasks.py
kentsanggds: At the moment if it's not `Sent`, then it's assumed to be Failed and set to `failed` status. 
If we don't get a `Sent` back from DVLA then I'm not sure if there is a better status to set it to other than `failed`

tests/app/celery/test_ftp_update_tasks.py
kentsanggds: Not sure if that proves anything, as they are both explicitly set to sending?

app/celery/tasks.py
richardc0: Maybe a rename to make it clearer? We should probably prefix all the notification statuses and all the job status with something? Or use two enums JobStatus.Sent and NotificationStatus.Sent?

PR #1337, opened 2017-10-24
Add stats test for BST and fix existing stat test

kentsanggds:
## What

`test_fetch_stats_by_date_range_for_all_services_returns_test_notifications` was failing when the test was run between 11pm and 12am during BST as the expected results did not match due to time adjustment when retrieving stats.

- So time was fixed so that the test would not fail
- New test for the BST hour with similar data injection to make sure that the function works.


venusbb:
Are there anywhere in the code other than tests that may be affected?

kentsanggds:
Nothing else has broken within the BST time, otherwise they would have also errored

PR #1340, opened 2017-10-25
Add task for ses results queue

richardc0:
Carrying on from Sakis's PR from August https://github.com/alphagov/notifications-api/pull/1156

In preparation for moving the SNS notification to an SES queue remove the HTTP errors codes and arguments as the method will now be run by a celery task. Also made the callback http method return more generic codes as this will be removed once the SNS has been moved to use a queue. Also added a retry if process_ses_results task fails retry every 5 mins for an hour.

- Updated the process_ses_results task to retry
- Added tests to check for retry
- Removed errors and arguments returned from process_ses_response
- Updated tests


app/celery/tasks.py
```diff
@@ -530,3 +531,12 @@ def process_incomplete_job(job_id):
             process_row(row_number, recipient, personalisation, template, job, job.service)
 
     job_complete(job, job.service, template, resumed=True)
+
+
+@notify_celery.task(bind=True, name="process-ses-result", max_retries=12, default_retry_delay=300000)
```
servingUpAces: Our other task use max_retries = 5 and default_retry_delay = 300. Why is this one different?

richardc0: Ahh I thought it was milliseconds for some reason so I'll update 30000 to 300.
So the other is 25 mins total time which felt like it would survive a blip but not an outage, 12 is an hour which seemed like a reasonable time to get everything back up. If its longer than 60 mins we have bigger issues that delivery receipts. If we are happy with 25 mins I'll make it consistent.

app/notifications/notifications_ses_callback.py
```diff
@@ -27,11 +27,12 @@
...
 def sns_callback_handler():
-    errors, status, kwargs = process_ses_response(json.loads(request.data))
+    errors = process_ses_response(json.loads(request.data))
     if errors:
-        raise InvalidRequest(errors, status)
+        current_app.logger.error(errors)
+        raise InvalidRequest(errors, 400)
 
-    return jsonify(**kwargs), status
+    return 200
```
sakisv: Is this enough to create a valid response object?

app/notifications/notifications_ses_callback.py
```diff
@@ -27,26 +26,31 @@
...
 
 @ses_callback_blueprint.route('/notifications/email/ses', methods=['POST'])
-def process_ses_response():
+def sns_callback_handler():
+    errors = process_ses_response(json.loads(request.data))
+    if errors:
+        current_app.logger.error(errors)
+        raise InvalidRequest(errors, 400)
+
+    return 200
```
sakisv: Is this enough to create a valid response object?

sakisv:

PR #1343, opened 2017-10-26
Improve get services

servingUpAces:
Refactored the get_detailed_services to stop using marshmallow. 
Also removed an extra query to services.
The query has been refactored to use an outer join to services on the notifications or notification_history table.
The expectation is that this change will improve the performance of the trial/live-services pages for platform admins.

### DIFF
--- a/tests/app/dao/test_services_dao.py
+++ b/tests/app/dao/test_services_dao.py
@@ -682,7 +682,7 @@ def test_dao_fetch_todays_stats_for_all_services_includes_all_services(notify_db
     assert stats == sorted(stats, key=lambda x: x.service_id)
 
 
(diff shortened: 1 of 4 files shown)


tests/app/dao/test_services_dao.py
```diff
@@ -682,7 +682,7 @@ def test_dao_fetch_todays_stats_for_all_services_includes_all_services(notify_db
     assert stats == sorted(stats, key=lambda x: x.service_id)
 
 
-def test_dao_fetch_todays_stats_for_all_services_only_includes_today(notify_db):
+def test_dao_fetch_todays_stats_for_all_services_only_includes_today(notify_db, notify_db_session):
```
leohemsted: nasty. good catch 👌

PR #1351, opened 2017-10-27
Add default email reply to address

kentsanggds:
## What

Need another email reply to address which is set to default otherwise it won't show the different reply to email addresses when testing the feature.

## Reference

https://www.pivotaltracker.com/story/show/150440099


kentsanggds:
Probably not needed as will create the ids manually

PR #1352, opened 2017-10-30
Add auth_type column to user and invited user

leohemsted:
the new `auth_type` table contains `sms_auth` and `email_auth`, however, any users or invited users created with old code will default to `sms_auth` until the code is updated. (this is done using a server_default so that it ensures backwards compat.)

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -50,6 +50,11 @@ PRIORITY = 'priority'
 TEMPLATE_PROCESS_TYPE = [NORMAL, PRIORITY]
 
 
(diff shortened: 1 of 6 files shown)


app/models.py
```diff
@@ -50,6 +50,11 @@
 TEMPLATE_PROCESS_TYPE = [NORMAL, PRIORITY]
 
 
+SMS_AUTH_TYPE = 'sms_auth'
```
richardc0: could just use sms and email. Rather than put '_auth' on. the column is called auth_type.

PR #1353, opened 2017-10-30
Retry in only certain scenarios

richardc0:
Instead of retrying if there are genuine errors, only retry if there are errors which are unexpected as otherwise the retries will happen and fail for the same reason e.g. that the message has changed format and will require a code update.

- Updated process_ses_results to only retry if there in an unknown exception
- Update test and assert that there is a retry there is a unknown exception


sakisv:

PR #1354, opened 2017-10-30
SMS sender id for post notifications

servingUpAces:
We have recently add the functionality to add multiple SMS senders for a service. This PR allows a service to send the SMS sender in the POST notification request, that SMS sender is then used as the sender id. At the moment only the default sender is used for a notification.

- Added an optional parameter in the form for POST /v2/notifications/sms and /service/<service_id>/send-notification to pass in the SMS sender id.

- Get the sms sender from the notification_sms_sender mapping table if that does not exist get the default sms sender to pass on to the sms provider.
- Added the notification_to_sms_sender mapping table to the purge notifications query

As part of https://www.pivotaltracker.com/story/show/152106587

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -20,8 +20,8 @@ from app.notifications.process_notifications import (
     persist_notification,
     persist_scheduled_notification,
     send_notification_to_queue,
-    simulated_recipient
-)
+    simulated_recipient,
+    persist_sms_sender_id_for_notification)
 from app.notifications.process_letter_notifications import (
     create_letter_notification
 )
@@ -31,7 +31,8 @@ from app.notifications.validators import (
     check_service_can_schedule_notification,
     check_service_has_permission,
     validate_template,
-    check_service_email_reply_to_id
+    check_service_email_reply_to_id,
+    check_service_sms_sender_id
 )
 from app.schema_validation import validate
 from app.v2.errors import BadRequestError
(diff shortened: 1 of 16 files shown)


app/v2/notifications/post_notifications.py
```diff
@@ -31,7 +31,8 @@
     check_service_can_schedule_notification,
     check_service_has_permission,
     validate_template,
-    check_service_email_reply_to_id
+    check_service_email_reply_to_id,
+    check_service_sms_sender_id
```
venusbb: Should these be done under different notification_type?

app/v2/notifications/post_notifications.py
```diff
@@ -163,6 +164,15 @@ def process_sms_or_email_notification(*, form, notification_type, api_key, templ
     return notification
 
 
+def persist_sender_to_notification_mapping(form, notification):
+    email_reply_to_id = form.get("email_reply_to_id", None)
+    if email_reply_to_id:
```
venusbb: This function is not tested.

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -8,8 +8,8 @@
     ScheduledNotification,
     SCHEDULE_NOTIFICATIONS,
     EMAIL_TYPE,
-    SMS_TYPE
-)
+    SMS_TYPE,
+    NotificationSmsSender)
```
kentsanggds: Can the new import be ordered

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -28,7 +29,7 @@
     create_service,
     create_template,
     create_reply_to_email,
-    create_service_sms_sender)
+    create_service_sms_sender, create_notification)
```
kentsanggds: New import could be on new line

tests/app/notifications/test_validators.py
```diff
@@ -311,9 +311,9 @@ def test_should_not_rate_limit_if_limiting_is_disabled(
 
 @pytest.mark.parametrize('key_type', ['test', 'normal'])
 def test_rejects_api_calls_with_international_numbers_if_service_does_not_allow_int_sms(
-    key_type,
-    notify_db,
-    notify_db_session,
+        key_type,
+        notify_db,
+        notify_db_session,
```
kentsanggds: Extra tab is not necessary here

PR #1342, opened 2017-10-25
Free sms limit by service and year

venusbb:
Continue PR #1326, create schema, dao for the new annual_billing table. Each row of the table is accessed per service per year. 
* created new dao for reading and writing annual_billing table
* created new schema for validating the request data
* updated create_service to add a row of exisint
* created new tests for reading and writing to the new table
* Updated the existing create_service related tests
* A new function get_current_financial_year_start_year to return the current financial year as number. 

**Pivotal story**
https://www.pivotaltracker.com/story/show/152067121

PR #1355, opened 2017-10-30
Added logic to return past and future free_sms_limit_data that dont e…

venusbb:
Create logic in requesting rest endpoint to get free_sms_fragment_limit for a service. 
**At the get end point**
* If an entry doesn't exist and it is a past year, we return the oldest entry that exist.
* If an entry doesn't exist and it is a current or future year, we return the newest entry that exist, and create a new record for that requested year in the db with the allowance from this entry. 
**At the post end point**
* Dao has a new function dao_update_annual_billing_for_current_and_future_years which will update the current year (or year in the parameter) and all the latter years. 

**Change to rest and Dao**
Some of the persistence logic is not pushed to Dao rather than rest. 

**Related PR**
https://github.com/alphagov/notifications-api/pull/1342

**Pivotal Story**
#1326


app/billing/rest.py
```diff
@@ -106,17 +106,34 @@ def get_free_sms_fragment_limit(service_id):
...
 
     if financial_year_start is None:
-        results = dao_get_all_free_sms_fragment_limit(service_id)
-
-        if len(results) == 0:
-            raise InvalidRequest('no annual billing information for this service', status_code=404)
-        return jsonify(data=[row.serialize_free_sms_items() for row in results]), 200
+        # return a list of all entries related to the service
+        sms_list = dao_get_all_free_sms_fragment_limit(service_id)
+        if len(sms_list) == 0:
```
kentsanggds: You can replace this with `if not sms_list:` for clearer code

app/billing/rest.py
```diff
@@ -106,17 +106,34 @@ def get_free_sms_fragment_limit(service_id):
...
+            raise InvalidRequest('no free-sms-fragment-limit list for this service in DB', status_code=404)
```
kentsanggds: Rather than erroring with `this service in DB`, you could mention the service affected, `service xxx` as not sure how `in DB` would help

app/billing/rest.py
```diff
@@ -106,17 +106,34 @@ def get_free_sms_fragment_limit(service_id):
...
+
         if result is None:
-            raise InvalidRequest('no free-sms-fragment-limit-info for this service and year', status_code=404)
+            # An entry does not exist in annual_billing table for that service and year. If it is a past year,
+            # we return the oldest entry.
+            # If it is the current or future years, we create an entry in the db table using the newest record,
+            # and return that number.  If all fails, we return InvalidRequest.
+            sms_list = dao_get_all_free_sms_fragment_limit(service_id)
+            if len(sms_list) == 0:
+                raise InvalidRequest('no free-sms-fragment-limit entry for this service in DB', status_code=404)
```
kentsanggds: Could probably move the raise error before the whole `if... else...` block rather than repeat the code

app/billing/rest.py
```diff
@@ -106,17 +106,34 @@ def get_free_sms_fragment_limit(service_id):
...
+            else:
+                result = sms_list[-1]   # The newest entry
+
+                new_sms = AnnualBilling(service_id=service_id, financial_year_start=financial_year_start,
+                                        free_sms_fragment_limit=result.free_sms_fragment_limit)
+
+                dao_create_or_update_annual_billing_for_year(new_sms)
 
-        return jsonify(data=result.serialize_free_sms_items()), 200
+        return jsonify(result.serialize_free_sms_items()), 200
```
kentsanggds: It's probably better to pass the serialized result through with `data=` as this follows a similar pattern used in other parts of the API so would be consistent.

tests/app/billing/test_billing.py
```diff
@@ -282,8 +282,8 @@ def test_create_free_sms_fragment_limit(client, sample_service):
     json_resp = json.loads(response_get.get_data(as_text=True))
     assert response.status_code == 201
     assert response_get.status_code == 200
-    assert json_resp['data']['financial_year_start'] == 2017
-    assert json_resp['data']['free_sms_fragment_limit'] == 250
+    assert json_resp['financial_year_start'] == 2017
```
kentsanggds: Need to add in the `['data']` block if you choose to return the json block in `['data']`, you could put this block as part of the `json_resp` setting in line 282 rather than adding it here. 
You will also need to refactor the other tests.

tests/app/billing/test_billing.py
```diff
@@ -345,19 +345,18 @@ def test_get_free_sms_fragment_limit_for_all_years(client, sample_service):
...
+        assert json_resp[i]['free_sms_fragment_limit'] == limits[i]
+        assert json_resp[i]['financial_year_start'] == years[i]
 
 
-def test_get_free_sms_fragment_limit_no_year_data_return_404(client, sample_service):
+def test_get_free_sms_fragment_limit_no_service_return_404(client, sample_service):
 
     response_get = client.get(
-        'service/{}/billing/free-sms-fragment-limit?financial_year_start={}'.format(sample_service.id, 1999),
+        'service/{}/billing/free-sms-fragment-limit?financial_year_start={}'.format(uuid.uuid4(), 1999),
```
kentsanggds: You could use the mock `fake_uuid` rather than the uuid function

tests/app/billing/test_billing.py
```diff
@@ -395,5 +394,58 @@ def test_post_free_sms_fragment_limit_current_year(client, sample_service):
...
+def test_get_free_sms_fragment_limit_past_year_get_db_oldest_record(client, sample_service):
+    response = client.get(
+        'service/{}/billing/free-sms-fragment-limit'.format(sample_service.id),
+        headers=[('Content-Type', 'application/json'), create_authorization_header()])
+    sms_list = json.loads(response.get_data(as_text=True))
+    past_year = sms_list[0]['financial_year_start'] - 1
+
+    # create an older entry than the one created in sample service has an entry already
+    annual_billing = {'financial_year_start': past_year, 'free_sms_fragment_limit': 1000}
+    response = client.post('service/{}/billing/free-sms-fragment-limit'.format(sample_service.id),
```
kentsanggds: When seeding data for your test you probably want to do this via the data model or dao directly rather than using rest endpoints, as then the test will be more targeted and efficient.

tests/app/billing/test_billing.py
```diff
@@ -395,5 +394,58 @@ def test_post_free_sms_fragment_limit_current_year(client, sample_service):
...
+
+    response = client.get(
+        'service/{}/billing/free-sms-fragment-limit'.format(sample_service.id),
+        headers=[('Content-Type', 'application/json'), create_authorization_header()])
+    sms_list = json.loads(response.get_data(as_text=True))
+    next_year = sms_list[-1]['financial_year_start'] + 1
+
+    # create an older entry than the one created in sample service has an entry already
+    annual_billing = {'financial_year_start': next_year, 'free_sms_fragment_limit': 1000}
+    response = client.post('service/{}/billing/free-sms-fragment-limit'.format(sample_service.id),
```
kentsanggds: Similar to previous comment, probably better to seed test data using DAO

tests/app/billing/test_billing.py
```diff
@@ -395,5 +394,58 @@ def test_post_free_sms_fragment_limit_current_year(client, sample_service):
...
+                           data=json.dumps(annual_billing),
+                           headers=[('Content-Type', 'application/json'), create_authorization_header()])
+    # request one year ahead that record doesn't exist yet
+    response = client.get(
+        'service/{}/billing/free-sms-fragment-limit?financial_year_start={}'.format(sample_service.id, next_year + 2),
+        headers=[('Content-Type', 'application/json'), create_authorization_header()])
+    json_resp = json.loads(response.get_data(as_text=True))
+
+    # get a list of request and see if the newe
+    response = client.get(
```
kentsanggds: If you are writing a test to make sure that a new record is created for a year it's probably better to use a DAO function / data model query to check that the data is as expected.

tests/app/billing/test_billing.py
venusbb: This should be OK as long as consistence between front and backend

app/billing/rest.py
venusbb: Only get a list if there is no result. Most of the time there should be entry for the year.

app/billing/rest.py
venusbb: Discussed with Becca on this. It is not necessary.

tests/app/billing/test_billing.py
```diff
@@ -282,118 +284,90 @@ def test_create_free_sms_fragment_limit(client, sample_service):
     json_resp = json.loads(response_get.get_data(as_text=True))
     assert response.status_code == 201
     assert response_get.status_code == 200
-    assert json_resp['data']['financial_year_start'] == 2017
-    assert json_resp['data']['free_sms_fragment_limit'] == 250
+    assert json_resp['financial_year_start'] == get_current_financial_year_start_year()
```
kentsanggds: Will this still work for future years? as you are doing a `client.get` on 2017? 
Also it's probably better to use a db session to check the data rather than the request GET

tests/app/billing/test_billing.py
```diff
@@ -282,118 +284,90 @@ def test_create_free_sms_fragment_limit(client, sample_service):
...
-def test_get_free_sms_fragment_limit_for_all_years(client, sample_service):
-    years = [2016, 2017, 2018]
-    limits = [1000, 2000, 3000]
-
-    for i in range(0, len(years)):
-        annual_billing = {'financial_year_start': years[i], 'free_sms_fragment_limit': limits[i]}
-        response = client.post('service/{}/billing/free-sms-fragment-limit'.format(sample_service.id),
-                               data=json.dumps(annual_billing),
-                               headers=[('Content-Type', 'application/json'), create_authorization_header()])
+def test_update_free_sms_fragment_limit(client, sample_service):
```
kentsanggds: Not sure if this test will actually update an existing `free_sms_fragment_limit`, 
if it is set in the default then assert the expected value in the DB before doing the POST, then do a POST which updates that limit and check that it has been updated via a db session

tests/app/billing/test_billing.py
```diff
@@ -282,118 +284,90 @@ def test_create_free_sms_fragment_limit(client, sample_service):
...
 
 
-def test_get_free_sms_fragment_limit_current_year(client, sample_service):
-    response = client.get(
-        'service/{}/billing/free-sms-fragment-limit/current-year'.format(sample_service.id, True),
-        headers=[('Content-Type', 'application/json'), create_authorization_header()])
-    json_resp = json.loads(response.get_data(as_text=True))
+def test_get_free_sms_fragment_limit_past_year_not_exist(client, sample_service):
+    current_year = get_current_financial_year_start_year()
+    create_annual_billing(sample_service.id, 9999, current_year - 1)
```
kentsanggds: Better to put in the keyword arg in the parameter so it's easier to understand what `9999` means.

tests/app/billing/test_billing.py
```diff
@@ -282,118 +284,90 @@ def test_create_free_sms_fragment_limit(client, sample_service):
...
+    create_annual_billing(sample_service.id, 10000, current_year + 1)
```
kentsanggds: Not sure if this adds anything to the test as you are checking that the years before the earliest year picks up the setting from the earliest yet

tests/app/billing/test_billing.py
```diff
@@ -282,118 +284,90 @@ def test_create_free_sms_fragment_limit(client, sample_service):
...
 
-def test_post_free_sms_fragment_limit_current_year(client, sample_service):
 
-    data_new = {'free_sms_fragment_limit': 7777}
-    response = client.post('service/{}/billing/free-sms-fragment-limit'.format(sample_service.id),
-                           data=json.dumps(data_new),
-                           headers=[('Content-Type', 'application/json'), create_authorization_header()])
+def test_get_free_sms_fragment_limit_future_year_not_exist(client, sample_service):
+    current_year = get_current_financial_year_start_year()
+    create_annual_billing(sample_service.id, 9999, current_year - 1)
```
kentsanggds: Same here, not sure if this shows anything, as you just want to check that there is a setting for years beyond the existing setting

app/billing/rest.py
```diff
@@ -97,49 +97,51 @@ def _transform_billing_for_month(billing_for_month):
...
+    result = dao_get_free_sms_fragment_limit_for_year(service_id, financial_year_start)
 
-        if len(results) == 0:
-            raise InvalidRequest('no annual billing information for this service', status_code=404)
-        return jsonify(data=[row.serialize_free_sms_items() for row in results]), 200
-    else:
-        result = dao_get_free_sms_fragment_limit_for_year(service_id, financial_year_start)
-        if result is None:
-            raise InvalidRequest('no free-sms-fragment-limit-info for this service and year', status_code=404)
+    if result is None:
```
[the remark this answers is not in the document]
venusbb: Probably the same.

app/billing/rest.py
```diff
@@ -97,49 +97,51 @@ def _transform_billing_for_month(billing_for_month):
...
+        if not sms_list:
+            raise InvalidRequest('no free-sms-fragment-limit entry for service {} in DB'.format(service_id), 404)
+        else:
+            if financial_year_start is None:
+                financial_year_start = get_current_financial_year_start_year()
+
+            if int(financial_year_start) < get_current_financial_year_start_year():
+                result = sms_list[0]   # The oldest entry
+            else:
+                result = sms_list[-1]  # The newest entry
```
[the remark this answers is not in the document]
venusbb: The return is actually an AnnualBilling object. I will rename that.

app/billing/rest.py
```diff
@@ -97,49 +97,51 @@ def _transform_billing_for_month(billing_for_month):
...
 
     financial_year_start = form.get('financial_year_start')
     free_sms_fragment_limit = form.get('free_sms_fragment_limit')
 
-    result = dao_get_free_sms_fragment_limit_for_year(service_id, financial_year_start)
-
-    if result:
-        result.free_sms_fragment_limit = free_sms_fragment_limit
+    current_year = get_current_financial_year_start_year()
+    if financial_year_start is None or financial_year_start >= current_year:
```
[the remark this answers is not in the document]
venusbb: The same

tests/app/billing/test_billing.py
venusbb: Agree. Changed.

tests/app/billing/test_billing.py
venusbb: agreed. Changed.

tests/app/billing/test_billing.py
venusbb: Agree. Extra assert added

app/billing/rest.py
```diff
@@ -117,14 +117,15 @@ def get_free_sms_fragment_limit(service_id):
                 financial_year_start = get_current_financial_year_start_year()
 
             if int(financial_year_start) < get_current_financial_year_start_year():
-                result = sms_list[0]   # The oldest entry
+                annual_billing = sms_list[0]   # The oldest entry
             else:
-                result = sms_list[-1]  # The newest entry
+                annual_billing = sms_list[-1]  # The newest entry
```
kentsanggds: This could be `latest_free_fragment_limit = sms_list[-1].free_sms_fragment_limit` and then you can pass in this value to the `dao_create_or_update_annual_billing_for_year`, this would make the code here much clearer.

As it looks like you are overwriting the `annual_billing` variable after setting it.

tests/app/billing/test_billing.py
kentsanggds: To make it clearer this could be -

`create_annual_billing(sample_service.id, free_sms_fragment_limit=9999, financial_year_start=current_year - 1)`

as it is much clearer

app/billing/rest.py
venusbb: Think this is OK as no chance with both exist. It's either past or future

PR #1358, opened 2017-10-31
Fix bug processing letters with . as address line

quis:
If an address line started with punctuation it caused a bug where we couldn’t properly map the address to the columns required in the DVLA file.

This bug is fixed by:
- [x] https://github.com/alphagov/notifications-utils/pull/244


servingUpAces:
I did this in a previous PR, closing this one. Thanks @quis

PR #1359, opened 2017-10-31
Update utils with a bug fix for addresses in LetterDVLA template

servingUpAces:
- Update utils with a fix for address formats for letters. 
- Added a command to re-run the build_dvla_file for a given job id.  
There was an instance where the file failed to create because the address in the personalisation had punctuation which caused the address lines to merge into one.
Utils has been updated to fix that problem, this task will allow us to re-run the task so that the notifications can be sent.


sakisv:

PR #1363, opened 2017-11-02
Log successful process of SES receipt

sakisv:


sakisv:
Closing, as we already do that in `process_ses_response`

PR #1364, opened 2017-11-03
Work out SMS message prefixing in a way that is aware of multiple SMS senders

quis:
## Add prefix SMS with service name to service model

If the service is sending messages from GOVUK, then its messages should be prefixed with the service name. Right now this logic is:
- worked out separately in the admin app and API
- isn’t aware of multiple senders

This commit moves the logic to one place (the service model). It does this in a slightly naive way, in that it only looks at the default sender, not the actual sender of the message.

In the future this will go away because we’ll move it to being a setting that’s controlled independently of the service name. But this is the first step towards that.

## Look at model to work out whether to prefix message

`service.sms_sender` has been deprecated; we should be looking at which of the service’s SMS senders is default to work out if the message has been sent from GOVUK or not (and if it has, then prefix the message with the service name). This logic is now in the model, so this commit makes use of it.

The arguments to `SMSMessageTemplate` are _super_ badly named – `sender` isn’t really used as a string, it’s a boolean that effectively means ‘is this a custom sender (`True`) or the platform default (`False`)’. We should rename it once this bug is fixed.

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -48,7 +48,7 @@ def send_sms_to_provider(notification):
         )
         template_model = dao_get_template_by_id(notification.template_id, notification.template_version)
 
-        sender_has_been_customised = (not service.prefix_sms_with_service_name())
+        sender_has_been_customised = (not service.get_prefix_sms_with_service_name())
 
         template = SMSMessageTemplate(
             template_model.__dict__,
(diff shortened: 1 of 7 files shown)


quis:
## Add a column for a service to set SMS prefixing

In future changes, services will be able to control whether their text messages will be prefixed with the name of their service.

This commit:
- adds a column to store the value of that setting
- makes the service model take notice of it, if it were to have a value set

It doesn’t:
- provide a way of setting the value of this column

Currently the column can have three values:
- `None` – ignore it (this is what all current services will start as) and continue to determine whether to prefix messages by looking at the sender
- `True` – always the service name to the start of text messages
- `False` – never add the service name to the start of text messages

In the future we’ll migrate all services to be either `True` or `False`, the `None` will go away and all services will have direct control over the setting.

PR #1371, opened 2017-11-03
Email 2fa api endpoints

leohemsted:
I've added a new template to the GOV.UK Notify service:

image

If you call `POST /<user_id>/email-code`, we'll send out one of those emails with a fancy link to click.

That fancy link contains a signed dict of format `{'user_id': str(user.id), 'secret_code': secret_code}`. `user_id` is self explanatory, `secret_code` is an arbitrary uuid - it is saved in the database in the VerifyCode table

To get it into the VerifyCode table, we had to do some work - specifically, the existing email codes (for new user creation) no longer go into the verify code table. That's okay since https://github.com/alphagov/notifications-admin/pull/1599 fixed that anyway.


app/user/rest.py
```diff
@@ -142,41 +136,75 @@ def verify_user_code(user_id):
...
     use_user_code(code.id)
     return jsonify({}), 204
 
 
 @user_blueprint.route('/<uuid:user_id>/sms-code', methods=['POST'])
 def send_user_sms_code(user_id):
+    data = request.get_json()
+    user_to_send_to = validate_2fa_call(user_id, data, post_send_user_sms_code_schema)
+    if not user_to_send_to:
+        return jsonify({}), 204
```
richardc0: If this fails shouldn't you return an error?

app/user/rest.py
```diff
@@ -142,41 +136,75 @@ def verify_user_code(user_id):
...
+    template = dao_get_template_by_id(current_app.config['SMS_CODE_TEMPLATE_ID'])
+
+    personalisation = {'verify_code': secret_code}
+
+    create_2fa_code(template, mobile, personalisation)
+    return jsonify({}), 204
+
+
+@user_blueprint.route('/<uuid:user_id>/email-code', methods=['POST'])
+def send_user_email_code(user_id):
```
richardc0: This is pretty much the same as the code for SMS, you should consolidate this method and pass in the template and the personalisation

app/user/rest.py
```diff
@@ -142,41 +136,75 @@ def verify_user_code(user_id):
...
+    data = request.get_json()
+    user_to_send_to = validate_2fa_call(user_id, data, post_send_user_email_code_schema)
+    if not user_to_send_to:
+        return jsonify({}), 204
+
+    secret_code = str(uuid.uuid4())
```
richardc0: Why not the same as the SMS code?

app/user/users_schema.py
```diff
@@ -0,0 +1,38 @@
...
+        'code_type': {'type': 'string'},
+    },
+    'required': ['code', 'code_type'],
+    'additionalProperties': False
+}
+
+
+post_send_user_email_code_schema = {
+    '$schema': 'http://json-schema.org/draft-04/schema#',
+    'description': 'POST schema for generating a 2fa email',
```
richardc0: Update description to why its different to the schema definition below.

tests/app/invite/test_invite_rest.py
```diff
@@ -115,6 +115,7 @@ def test_get_all_invited_users_by_service(client, notify_db, notify_db_session,
     for invite in json_resp['data']:
         assert invite['service'] == str(sample_service.id)
         assert invite['from_user'] == str(invite_from.id)
+        assert invite['auth_type'] == SMS_AUTH_TYPE
```
richardc0: Should the test method now be test_get_all_invited_users_via_sms_by_service

tests/app/user/test_rest_verify.py
```diff
@@ -9,11 +9,14 @@
 from flask import url_for, current_app
 from freezegun import freeze_time
 
+from app.dao.users_dao import create_user_code
 from app.dao.services_dao import dao_update_service, dao_fetch_service_by_id
 from app.models import (
     VerifyCode,
     User,
-    Notification
+    Notification,
```
richardc0: Order imports

tests/app/invite/test_invite_rest.py
leohemsted: nah i haven't changed the method, it's just that users default to `SMS_AUTH` - i'm checking to make sure it's returned to the front-end as that info is needed by the admin

app/user/rest.py
leohemsted: hmm, i was just following existing behaviour that threw away the errors object (`verify_code, errors = request_verify_code_schema.load(request.get_json())`) - i imagine that was a dev mistake rather than intentional. will look into it and see if any tests break

leohemsted: it's deliberate - see `test_send_sms_code_returns_204_when_too_many_codes_already_created`

PR #1372, opened 2017-11-03
Make SMS prefix setting non-nullable

quis:
We have now:
- defaulted new services to start with this column set to `true`
- migrated all preexisting services to have either `true` or `false` set for this column

There is no way for a service to switch back from `true`/`false` to `null`.

This means that we can safely enforce a non-nullable constraint on this column now.

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/1370

### DIFF
--- a/migrations/versions/0140_sms_prefix_non_nullable.py
+++ b/migrations/versions/0140_sms_prefix_non_nullable.py
@@ -1,0 +1,46 @@
+"""
+
+Revision ID: 0140_sms_prefix_non_nullable
+Revises: 0139_migrate_sms_allowance_data
+Create Date: 2017-11-07 13:04:04.077142
+
+"""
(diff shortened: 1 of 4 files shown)


quis:
@richardc0 Can you or someone else take another look at this please?

I had to make some changes because the GOV.UK Notify service still relies on  the `sms_sender` column. It doesn’t have a row in the `service_sms_senders` table.

This means the [previous migration](https://github.com/alphagov/notifications-api/blob/e14275c19a81196f5396f89d5f56deaeb2fca028/migrations/versions/0133_set_services_sms_prefix.py) never changed this service’s value for `prefix_sms` from `null`. 

So  this commit also changes its value to `False`, so that the rest of  the migration, of the whole column, is possible.

PR #1373, opened 2017-11-03
Get inbound SMS API

kentsanggds:
## What

Add endpoints for API to get `inbound_sms` information

- dates for the `inbound_sms` `created_at` and `provider_date` had to be refactored in order to pass the schema response check
- paging added for response json, `next` element will show the url value to use to fetch the next page of results, limit is set to `API_PAGE_SIZE` which is currently default to 250 in `config.py`

## Reference

https://www.pivotaltracker.com/story/show/147024273

### DIFF
--- a/app/dao/inbound_sms_dao.py
+++ b/app/dao/inbound_sms_dao.py
@@ -34,7 +34,6 @@ def dao_get_inbound_sms_for_service(service_id, limit=None, user_number=None):
 
 def dao_get_paginated_inbound_sms_for_service(
     service_id,
-    user_number=None,
     older_than=None,
     page_size=None
 ):
@@ -47,9 +46,6 @@ def dao_get_paginated_inbound_sms_for_service(
         older_than_created_at = db.session.query(
             InboundSms.created_at).filter(InboundSms.id == older_than).as_scalar()
         filters.append(InboundSms.created_at < older_than_created_at)
-
-    if user_number:
-        filters.append(InboundSms.user_number == user_number)
 
     query = InboundSms.query.filter(*filters)
 
--- a/app/dao/inbound_sms_dao.py
+++ b/app/dao/inbound_sms_dao.py
@@ -2,7 +2,8 @@ from datetime import (
     timedelta,
     datetime
 )
-
+from flask import current_app
+from sqlalchemy import desc
 
 from app import db
 from app.dao.dao_utils import transactional
@@ -29,6 +30,34 @@ def dao_get_inbound_sms_for_service(service_id, limit=None, user_number=None):
         q = q.limit(limit)
 
     return q.all()
+
+
+def dao_get_paginated_inbound_sms_for_service(
+    service_id,
+    user_number=None,
+    older_than=None,
+    page=1,
+    page_size=None
+):
+    if page_size is None:
+        page_size = current_app.config['PAGE_SIZE']
+
+    filters = [InboundSms.service_id == service_id]
+
+    if older_than is not None:
+        older_than_created_at = db.session.query(
+            InboundSms.created_at).filter(InboundSms.id == older_than).as_scalar()
+        filters.append(InboundSms.created_at < older_than_created_at)
+
+    if user_number:
+        filters.append(InboundSms.user_number == user_number)
+
+    query = InboundSms.query.filter(*filters)
+
+    return query.order_by(desc(InboundSms.created_at)).paginate(
+        page=page,
+        per_page=page_size
+    ).items
 
 
 def dao_count_inbound_sms_for_service(service_id):
--- a/app/dao/inbound_sms_dao.py
+++ b/app/dao/inbound_sms_dao.py
@@ -36,7 +36,6 @@ def dao_get_paginated_inbound_sms_for_service(
     service_id,
     user_number=None,
     older_than=None,
-    page=1,
     page_size=None
 ):
     if page_size is None:
@@ -44,7 +43,7 @@ def dao_get_paginated_inbound_sms_for_service(
 
     filters = [InboundSms.service_id == service_id]
 
-    if older_than is not None:
+    if older_than:
         older_than_created_at = db.session.query(
             InboundSms.created_at).filter(InboundSms.id == older_than).as_scalar()
         filters.append(InboundSms.created_at < older_than_created_at)
@@ -55,7 +54,6 @@ def dao_get_paginated_inbound_sms_for_service(
     query = InboundSms.query.filter(*filters)
 
     return query.order_by(desc(InboundSms.created_at)).paginate(
-        page=page,
         per_page=page_size
     ).items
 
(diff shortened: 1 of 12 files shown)


app/v2/inbound_sms/get_inbound_sms.py
```diff
@@ -0,0 +1,36 @@
...
+
+    inbound_sms = inbound_sms_dao.dao_get_inbound_sms_for_service(
+        authenticated_service.id, user_number=user_number
+    )
+
+    return jsonify(inbound_sms_list=[i.serialize() for i in inbound_sms]), 200
+
+
+@v2_inbound_sms_blueprint.route("", methods=['GET'])
+def get_all_inbound_sms():
```
servingUpAces: We may need to paginate this request.

kentsanggds: Sure, will update

app/v2/inbound_sms/get_inbound_sms.py
```diff
@@ -0,0 +1,83 @@
...
+from notifications_utils.recipients import InvalidPhoneError
+
+from app import authenticated_service
+from app.dao import inbound_sms_dao
+from app.v2.errors import BadRequestError
+from app.v2.inbound_sms import v2_inbound_sms_blueprint
+
+
+@v2_inbound_sms_blueprint.route("/<user_number>", methods=['GET'])
+def get_inbound_sms_by_number(user_number):
```
servingUpAces: Why is user_number a path param and not a query param?

app/v2/inbound_sms/get_inbound_sms.py
```diff
@@ -0,0 +1,83 @@
...
+    _data = request.args.to_dict(flat=False)
+
+    # flat=False makes everything a list, but we only ever allow one value for "older_than"
+    if 'older_than' in _data:
```
servingUpAces: Is it easier to just use request.args.get('older_than')?

app/v2/inbound_sms/get_inbound_sms.py
```diff
@@ -0,0 +1,83 @@
...
+        received_text_messages=[i.serialize() for i in paginated_inbound_sms],
+        links=_build_links(
+            paginated_inbound_sms,
+            endpoint='get_inbound_sms_by_number',
+            user_number=user_number
+        )
+    ), 200
+
+
+@v2_inbound_sms_blueprint.route("", methods=['GET'])
```
servingUpAces: If user_number was a query param we would only need one route.

app/v2/inbound_sms/get_inbound_sms.py
```diff
@@ -0,0 +1,83 @@
+import uuid
```
servingUpAces: Unused import

app/v2/inbound_sms/get_inbound_sms.py
```diff
@@ -0,0 +1,83 @@
...
+
+from flask import jsonify, request, url_for, current_app
+from sqlalchemy.orm.exc import NoResultFound
```
servingUpAces: Unused imports

tests/app/v2/inbound_sms/test_get_inbound_sms.py
```diff
@@ -0,0 +1,251 @@
+import datetime
```
servingUpAces: check for unused imports, please

app/v2/inbound_sms/get_inbound_sms.py
kentsanggds: it's following a similar pattern to getting notification by id. Could convert it to a query param if that is preferable as I guess it is optional and generally path params are used for required params.

app/v2/inbound_sms/get_inbound_sms.py
kentsanggds: sure, was just copy and pasting from another function

app/v2/inbound_sms/get_inbound_sms.py
kentsanggds: ok

tests/app/v2/inbound_sms/test_get_inbound_sms.py
kentsanggds: will do

app/v2/inbound_sms/get_inbound_sms.py
servingUpAces: It is preferable, plus we don't want the user_number showing up in our logs.

app/v2/inbound_sms/get_inbound_sms.py
servingUpAces: Fair enough - I think the other function could just use the get as well

app/dao/inbound_sms_dao.py
```diff
@@ -31,6 +32,32 @@ def dao_get_inbound_sms_for_service(service_id, limit=None, user_number=None):
     return q.all()
 
 
+def dao_get_paginated_inbound_sms_for_service(
+    service_id,
+    user_number=None,
```
servingUpAces: user_number is never used for this method - should we get rid of it.

app/dao/inbound_sms_dao.py
```diff
@@ -31,6 +32,32 @@ def dao_get_inbound_sms_for_service(service_id, limit=None, user_number=None):
     return q.all()
 
 
+def dao_get_paginated_inbound_sms_for_service(
```
servingUpAces: We should test this method

app/v2/inbound_sms/inbound_sms_schemas.py
```diff
@@ -0,0 +1,77 @@
...
+    "description": "GET inbound sms schema response",
+    "type": "object",
+    "title": "GET response v2/inbound_sms",
+    "properties": {
+        "provider_date": {
+            "format": "date-time",
+            "type": "string",
+            "description": "Date+time sent by provider"
+        },
+        "provider_reference": {"type": ["string", "null"]},
```
servingUpAces: I doubt the provider reference is important to the service

app/v2/inbound_sms/inbound_sms_schemas.py
```diff
@@ -0,0 +1,77 @@
...
+        "user_number": {"type": "string"},
+        "created_at": {
```
servingUpAces: Let's just uses either created_at or provider_date, having both may be confusing for the service.

app/dao/inbound_sms_dao.py
kentsanggds: ok

app/dao/inbound_sms_dao.py
kentsanggds: ok

app/v2/inbound_sms/inbound_sms_schemas.py
kentsanggds: ok

app/v2/inbound_sms/inbound_sms_schemas.py
kentsanggds: ok, created_at makes more sense

kentsanggds: will we ever need the provider info to be serialized? as I could just update the `serialize` function to strip those out otherwise will create another serialize function for api calls.

PR #1376, opened 2017-11-07
Remove computed SMS prefix

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-admin/pull/1619

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -209,7 +209,6 @@ class ServiceSchema(BaseSchema):
     reply_to_email_address = fields.Method(method_name="get_reply_to_email_address")
     sms_sender = fields.Method(method_name="get_sms_sender")
     letter_contact_block = fields.Method(method_name="get_letter_contact")
-    prefix_sms_with_service_name = fields.Method(method_name="get_prefix_sms_with_service_name")
 
     def service_permissions(self, service):
         return [p.permission for p in service.permissions]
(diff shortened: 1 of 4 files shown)


app/schemas.py
```diff
@@ -209,7 +209,7 @@ class ServiceSchema(BaseSchema):
     reply_to_email_address = fields.Method(method_name="get_reply_to_email_address")
     sms_sender = fields.Method(method_name="get_sms_sender")
     letter_contact_block = fields.Method(method_name="get_letter_contact")
-    prefix_sms_with_service_name = fields.Method(method_name="get_prefix_sms_with_service_name")
+    prefix_sms = field_for(models.Service, 'prefix_sms')
```
servingUpAces: the prefix_sms will show up on the schema without this line.

migrations/versions/0140_sms_prefix_non_nullable.py
```diff
@@ -0,0 +1,46 @@
+"""
```
servingUpAces: You'll need update this - there is a 0140 already

app/schemas.py
quis: Squished out.

migrations/versions/0140_sms_prefix_non_nullable.py
quis: Was only showing up because this was branched off the other branch. There’s no migration in this PR.

PR #1377, opened 2017-11-07
Added scheduled task to get stats for template usage

richardc0:
Currently the 'See templates used by month' report is timing out for some services due to the query time of the data from notification_history. This is a stats table which will hold the data and will be updated by a scheduled celery task once a day. This data will be combined with the 'live' data from notifications tables (which will be considerably less) to form the data of the new report.

- Added database migration script
- Added Celery task to run nightly
- Added dao code to retrieve the data from notification_history
- Added dao code to add and update stats


migrations/versions/0134_stats_template_usage.py
```diff
@@ -0,0 +1,65 @@
...
+    sa.Column('month', sa.Integer(), nullable=False),
+    sa.Column('year', sa.Integer(), nullable=False),
+    sa.Column('count', sa.Integer(), nullable=False),
+    sa.ForeignKeyConstraint(['template_id'], ['templates.id'], ),
+    sa.PrimaryKeyConstraint('template_id', 'month', 'year')
+    )
+    op.create_index(op.f('ix_stats_template_usage_by_month_month'), 'stats_template_usage_by_month', ['month'], unique=False)
+    op.create_index(op.f('ix_stats_template_usage_by_month_template_id'), 'stats_template_usage_by_month', ['template_id'], unique=False)
+    op.create_index(op.f('ix_stats_template_usage_by_month_year'), 'stats_template_usage_by_month', ['year'], unique=False)
+    op.drop_table('notification_statistics')
```
servingUpAces: I have the drop table for notificaiton_statistics in another PR I am working on now.

migrations/versions/0134_stats_template_usage.py
```diff
@@ -0,0 +1,65 @@
...
+    op.alter_column('notification_history', 'international',
```
servingUpAces: We can't update the model with this as of yet, we need to update the data. There is a server command to do this - but it just ended up on the bottom of the list and hasn't been done as of yet.

migrations/versions/0134_stats_template_usage.py
```diff
@@ -0,0 +1,65 @@
...
+               existing_type=sa.BOOLEAN(),
+               nullable=False)
+    op.alter_column('notifications', 'international',
```
servingUpAces: We can probably do this - but it's not really part of your story. may want to do it later.

app/config.py
```diff
@@ -241,6 +241,11 @@ class Config(object):
             'task': 'check-job-status',
             'schedule': crontab(),
             'options': {'queue': QueueNames.PERIODIC}
+        },
+        'daily-stats-template_usage_by_month': {
+            'task': 'daily-stats-template_usage_by_month',
+            'schedule': crontab(hour=00, minute=50),
```
leohemsted: minor thing - i'd prefer this to be `0` rather than `00` - due to poor decisions, in python 2 a `0` prefix referred to an octal number (it's now `0o123`), but in python3 it's now a syntax error - try typing `01` into a python shell for example. `00` works for some reason /shrug. So I'd prefer this to be just `0`, to dissuade people from trying to prefix everything with 0

app/utils.py
```diff
@@ -75,6 +75,22 @@ def get_london_month_from_utc_column(column):
...
+def get_london_year_from_utc_column(column):
+    """
+     Where queries need to count notifications by month it needs to be
+     the month in BST (British Summer Time).
+     The database stores all timestamps as UTC without the timezone.
+      - First set the timezone on created_at to UTC
+      - then convert the timezone to BST (or Europe/London)
+      - lastly truncate the datetime to month with which we can group
+        queries
+    """
```
leohemsted: docstring still talking about month.

Also, I don't think we need to muck about with timezones here, because the year switch over is always in GMT +0. Think this can just be `func.date_trunc('month', column)` (and then there's a discussion about whether it needs its own function in utils, but i'll leave that decision with you

migrations/versions/0135_stats_template_usage.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+    op.create_index(op.f('ix_stats_template_usage_by_month_month'), 'stats_template_usage_by_month', ['month'], unique=False)
+    op.create_index(op.f('ix_stats_template_usage_by_month_template_id'), 'stats_template_usage_by_month', ['template_id'], unique=False)
+    op.create_index(op.f('ix_stats_template_usage_by_month_year'), 'stats_template_usage_by_month', ['year'], unique=False)
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('notification_statistics',
```
leohemsted: did you mean to drop the table in the upgrade script?

tests/app/celery/test_csv_files/email.csv
```diff
@@ -0,0 +1,2 @@
+email_address
```
leohemsted: you've duplicated the `/test_csv_files/` folder into `tests/app/celery/`

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -834,3 +838,108 @@ def test_check_job_status_task_raises_job_incomplete_error_for_multiple_jobs(moc
...
+            assert result.month == 10
+            assert result.year == 2017
+            assert result.count == 1
+        elif result.template_id == template_two.id:
+            assert result.template_id == template_two.id
+            assert result.month == 4
+            assert result.year == 2016
+            assert result.count == 2
+        else:
+            raise AssertionError()
```
leohemsted: i'd rather you put an `order_by` on the query (just by month or something maybe?), and then just assert 
```
assert results[0].template_id == ...
assert results[0].month == ...
```
and so on. Logic in tests is the bane of my existence

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -834,3 +838,108 @@ def test_check_job_status_task_raises_job_incomplete_error_for_multiple_jobs(moc
...
+
+def test_daily_stats_template_usage_my_month_no_data():
+    daily_stats_template_usage_my_month()
+
+    results = db.session.query(StatsTemplateUsageByMonth).all()
+
+    assert len(results) == 0
+
+
+def test_daily_stats_template_usage_my_month_multiple_runs(notify_db, notify_db_session):
```
leohemsted: my => by

leohemsted: and the other test names too

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -834,3 +838,108 @@ def test_check_job_status_task_raises_job_incomplete_error_for_multiple_jobs(moc
...
+                assert result.count == 1
+            elif result.month == 9:
+                assert result.template_id == template_three.id
+                assert result.month == 9
+                assert result.year == 2017
+                assert result.count == 1
+            else:
+                raise AssertionError()
+        else:
+            raise AssertionError()
```
leohemsted: same comment about unrolling the loop here too

tests/app/dao/test_services_dao.py
```diff
@@ -1004,3 +1005,38 @@ def _assert_service_permissions(service_permissions, expected):
...
+            assert result.month.month == 10
+            assert result.year.year == 2017
+            assert result.count == 1
+        elif result.template_id == template_two.id:
+            assert result.template_id == template_two.id
+            assert result.month.month == 4
+            assert result.year.year == 2016
+            assert result.count == 2
+        else:
+            raise AssertionError()
```
leohemsted: unrolling the loop's harder here I guess because the template id sorting is non deterministic. 🤔

leohemsted: could maybe say `results = sorted(dao_fetch_monthly_historical_stats_by_template(), key=lambda x: x.month, x. year)`

not sure how you feel about that

app/celery/scheduled_tasks.py
```diff
@@ -402,3 +406,17 @@ def check_job_status():
...
+@notify_celery.task(name='daily-stats-template_usage_by_month')
+@statsd(namespace="tasks")
+def daily_stats_template_usage_my_month():
+    results = dao_fetch_monthly_historical_stats_by_template()
+
+    for result in results:
+        insert_or_update_stats_for_template(
+            result.template_id,
+            result.month.month,
+            result.year.year,
```
leohemsted: why do you have to do `result.month.month` here? what does `result.month` return, and can we change the dao fn so we don't need to do this?

app/dao/services_dao.py
```diff
@@ -520,3 +520,25 @@ def dao_fetch_active_users_for_service(service_id):
...
+    month = get_london_month_from_utc_column(NotificationHistory.created_at)
+    year = get_london_year_from_utc_column(NotificationHistory.created_at)
+    end_date = datetime.combine(date.today(), time.min)
+
+    return db.session.query(
+        NotificationHistory.template_id,
+        month.label('month'),
+        year.label('year'),
+        func.count().label('count')
+    ).filter(
```
servingUpAces: Do we need to filter 'test' notifications? I know we weren't before but we are filtering those from the dashboard

PR #1379, opened 2017-11-09
remove check to financial_year_start in schema as it can be None

venusbb:
A small pull request that remove the schema check of financial_year_start as this parameter is allowed to be None


quis:
Does this need a test?

venusbb:
I don't think it is affecting anything. Just skipping the check on this parameter.

PR #1380, opened 2017-11-09
Remove unused methods for sms sender

servingUpAces:
Service.sms_sender is not longer used in the application, instead there is a table to store multiple sms_senders per service service_sms_senders. 
This PR stops populating the service.sms_sender when a service is created. 
Also remove some of the inbound_number routes that are not needed and related tests.


tests/app/db.py
```diff
@@ -77,18 +76,84 @@ def create_service(
...
 
-    if do_create_inbound_number and INBOUND_SMS_TYPE in service_permissions:
-        create_inbound_number(number=sms_sender, service_id=service.id)
+    service.active = active
+    service.research_mode = research_mode
+
+    return service
+
+
+def create_service_with_inbound_number(
```
leohemsted: can you make these functions have *args and **kwargs that they pass through to `create_service` to reduce code duplication? eg:


```
def create_service_with_inbound_number(inbound_number='123456', *args, **kwargs):
    service = create_service(*args, **kwargs)
    create_inbound_number(inbound_number, service.id)
    [...]
    return service
```

leohemsted: can you make these functions have *args and **kwargs that they pass through to `create_service` to reduce code duplication? eg:


```
def create_service_with_inbound_number(inbound_number='123456', *args, **kwargs):
    service = create_service(*args, **kwargs)
    create_inbound_number(inbound_number, service.id)
    [...]
    return service
```

PR #1383, opened 2017-11-10
User mobile optional

leohemsted:
You can set both at the same time if you want.

If not, it currently gives a message "Mobile number must be set if auth_type is set to sms_auth". idk if we want to tweak that to make it a front-end display ready error message.

Also removed `PUT /user/<id>` as it's unused on the front-end

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -105,6 +105,26 @@ class UserSchema(BaseSchema):
             "_password", "verify_codes")
         strict = True
 
+    @validates('name')
+    def validate_name(self, value):
+        if not value:
+            raise ValidationError('Invalid name')
+
+    @validates('email_address')
+    def validate_email_address(self, value):
+        try:
+            validate_email_address(value)
+        except InvalidEmailError as e:
+            raise ValidationError(str(e))
+
+    @validates('mobile_number')
+    def validate_mobile_number(self, value):
+        try:
+            if value is not None:
+                validate_phone_number(value, international=True)
+        except InvalidPhoneError as error:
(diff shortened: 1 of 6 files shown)


app/user/rest.py
```diff
@@ -63,23 +77,6 @@ def create_user():
...
-            raise InvalidRequest(errors, status_code=400)
-        else:
-            reset_failed_login_count(user_to_update)
-    save_model_user(user_to_update, update_dict=update_dct, pwd=pwd)
-    return jsonify(data=user_schema.dump(user_to_update).data), 200
-
-
 @user_blueprint.route('/<uuid:user_id>', methods=['POST'])
 def update_user_attribute(user_id):
     user_to_update = get_user_by_id(user_id=user_id)
```
servingUpAces: Can we throw an error if the update dict includes password, please?

app/schemas.py
```diff
@@ -105,6 +105,26 @@ class Meta:
...
 
+    @validates('name')
+    def validate_name(self, value):
+        if not value:
+            raise ValidationError('Invalid name')
+
+    @validates('email_address')
+    def validate_email_address(self, value):
+        try:
+            validate_email_address(value)
```
richardc0: It feels like validate email should really return a true or false and then raise an exception rather than catch and rethrow.

leohemsted: this is a marshmallow pattern - we have a specific error handler in `app/errors.py` that deals with `ValidationError`s

leohemsted: oh wait i misread. Either way, `validate_email_address` is a utils fn and way out of scope of this PR

PR #1384, opened 2017-11-10
Fix notifications template history relationship

allait:
### Replace Notifications.template_history with Notifications.template

Each notification instance is connected to a specific version of the template that was used to generate notification content. Template versions are stored in TemplateHistory, so that's the table we need to use for Notifications.template relationship.

Currently, using Notifications.template returns the latest template version, which leads to incorrect content being returned by the API if the template has been updated.

This change removes Notifications.template_history attribute and replaces template relationship to refer to TemplateHistory instance based on the notification's template ID and version.

### Avoid assigning notification.template when creating test objects

`Notification.template` changed from being a Template relationship to TemplateHistory. When a relationship attribute is being assigned, SQLAlchemy checks that the assigned value type matches the relationship type. Since most tests at the moment create a notification using a Template instance this check fails.

Rewriting the tests to use TemplateHistory objects would require changes to the majority of tests. Instead, when creating a notification objects we assign the foreign key attributes directly. This skips the SQLAlchemy type check, but we still get the constraint check on the foreign keys, so a matching TemplateHistory object needs to exist in the database.

###  Add TemplateHistory.redact_personalisation attribute

TemplateHistory objects need to be connected to the template's TemplateRedacted record. This requires setting up a new SQLAlchemy relationship to allow accessing redact_personalisation from TemplateHistory instances.

We can avoid creating a foreign key in this case by setting up an explicit `primaryjoin` expression. Since TemplateHistory.id is created from Template.id and TemplateRedacted.template_id is already a foreign key to Template.id the foreign key should always be valid even without a DB constraint.

### Add a migration to replace notifications_template foreign key

Removes notifications.template_id foreign key and replaces it with a composite foreign key constraint to TemplateHistory using the existing notification columns for template ID and template version.

Foreign key constraint is created as NOT VALID to avoid locking the notifications table while postgres verifies that existing records don't break the constraint. From postgres docs:

> If the constraint is marked NOT VALID, the potentially-lengthy initial check to verify that all rows in the table satisfy the constraint is skipped. The constraint will still be enforced against subsequent inserts or updates (that is, they'll fail unless there is a matching row in the referenced table, in the case of foreign keys; and they'll fail unless the new row matches the specified check constraints). But the database will not assume that the constraint holds for all rows in the table, until it is validated by using the VALIDATE CONSTRAINT option.

VALIDATE CONSTRAINT will be issued as a separate migration in a follow-up PR.


richardc0:
This should have probably been four PRs!

migrations/versions/0136_notification_template_hist.py
```diff
@@ -0,0 +1,36 @@
+"""
+
+Revision ID: 0136_notification_template_hist
+Revises: 0135_stats_template_usage
+Create Date: 2017-11-08 10:15:07.039227
+
+"""
+from alembic import op
+
+revision = '0136_notification_template_hist'
```
richardc0: Watch out for conflicts Venus has a 0136 too.

allait:
@richardc0 I'm not sure what you'd split in this case. The change itself is 4 lines, the rest is fixing the failing tests.

PR #1389, opened 2017-11-14
migrate data from service table to annual billing

venusbb:
This step is part of the organisation type changes in this pivotal story: 

in this step, we 
- check all the free_sms_fragment_limit is not null. If it is, fill with default of 250000. 
- Check if all service has at least one row in annual_billing table. If there is none, create one for the current year. 
- migrate the existing free_sms_fragment_limit data from service table to annual billing table.


quis:
I’d prefer someone who is more familiar with SQL than me to look at this. The approach seems sound, but there are some non-trivial queries going on there…

quis:
Also, looking on prod there are 1405 rows in the services table and 2122 in the annual billing table – are you sure we haven’t already done step 2?

migrations/versions/0136_migrate_sms_allowance_data.py
```diff
@@ -0,0 +1,49 @@
...
+    update_sms_allowance = """
+        UPDATE annual_billing SET free_sms_fragment_limit = services.free_sms_fragment_limit
+        FROM services
+        WHERE annual_billing.service_id = services.id
+    """
+    op.execute(update_sms_allowance)
+
+
+def downgrade():
+    print('There is no action for downgrading to the previous version.')
```
richardc0: Are you sure this is correct? I would add a comment here saying why.

Although you have updated columns and inserted rows?  What if I wanted to downgrade 10 versions would this still work?

venusbb: It will still work as only data is affected.

venusbb:
@quis we have done step 1 and 2 already on the actual database, but step 1 and step 2 here will ensure:-
1. If there are any gaps (created) because we have not moved to use the new end points, will be filled in. 
2. For bootstrap and other users, make sure the database is ready to be used.

venusbb:
generate UUID from MD5
https://stackoverflow.com/questions/12505158/generating-a-uuid-in-postgres-for-insert-statement

PR #1390, opened 2017-11-14
Added new endpoint to get the new template stats

richardc0:
Added a new endpoint which combines the usage of the stats table and the data from the notifications tables, instead of using all the data from the notification_history table. This should speed up the query times and improve the page performance.

- Updated to make the stats create and update function transactional as it actually wasn't committing the data to the table
- Added the get from the stats table
- Add a a method to combine the two results
- Added the endpoint

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -27,7 +27,6 @@ from app.models import (
     Service,
     ServicePermission,
     ServiceSmsSender,
-    StatsTemplateUsageByMonth,
     Template,
     TemplateHistory,
     TemplateRedacted,
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -547,20 +547,19 @@ def dao_fetch_monthly_historical_stats_by_template():
     ).all()
 
 
-@transactional
 @statsd(namespace="dao")
-def dao_fetch_monthly_historical_usage_by_template_for_service(service_id):
+def dao_fetch_monthly_historical_usage_by_template_for_service(service_id, year):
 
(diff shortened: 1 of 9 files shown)


app/dao/services_dao.py
```diff
@@ -540,5 +542,61 @@ def dao_fetch_monthly_historical_stats_by_template():
         month,
         year
     ).order_by(
-        NotificationHistory.template_id
+        year,
+        month
     ).all()
+
+
+@transactional
```
servingUpAces: No need for transactional for read only queries

app/service/rest.py
```diff
@@ -517,6 +519,15 @@ def resume_service(service_id):
     return '', 204
 
 
+@service_blueprint.route('/<uuid:service_id>/notifications/templates_usage/monthly', methods=['GET'])
+def get_monthly_template_usage(service_id):
+    try:
+        data = dao_fetch_monthly_historical_usage_by_template_for_service(service_id)
+        return jsonify(stats=[i.serialize() for i in data]), 200
```
servingUpAces: Do we need the template name in the response?

app/dao/services_dao.py
```diff
@@ -12,31 +12,33 @@
...
     Notification,
+    NotificationHistory,
     Permission,
-    User,
-    InvitedUser,
+    ProviderStatistics,
     Service,
     ServicePermission,
+    ServiceSmsSender,
+    StatsTemplateUsageByMonth,
```
servingUpAces: Unused import

PR #1391, opened 2017-11-14
Add proxy header check

sakisv:
This adds a check for all incoming requests to ensure they are coming from our proxies.

Depends on https://github.com/alphagov/notifications-utils/pull/250


venusbb:
How do you ensure that all the routes will be working as normal after deploying this? Is there a test you can run on preview?

sakisv:
Good point. We could introduce a temporary feature flag that would be checked on utils, which will allow us to only log the result without actually blocking the requests. That will allow us to inspect the logs after a day or so and enable it with more confidence.

@leohemsted @allait thoughts?

leohemsted:
i like the idea of a feature flag in general I think. `current_app.config['CHECK_PROXY_HEADER']` or something

sakisv:
Something like [this](https://github.com/alphagov/notifications-utils/pull/253)?

PR #1392, opened 2017-11-14
Bridging API to update annual_billing table

venusbb:
This PR is part of the organisation type changes in this pivotal story:

In this step, we create bridging code to update the annual_billing table from the service end point, before the admin code is deployed and the data is migrated.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -98,6 +98,7 @@ from app.schemas import (
     detailed_service_schema
 )
 from app.utils import pagination_links
+from app.billing.rest import update_free_sms_fragment_limit_data
 
 service_blueprint = Blueprint('service', __name__)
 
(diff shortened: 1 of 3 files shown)


app/service/rest.py
```diff
@@ -202,6 +203,10 @@ def update_service(service_id):
     if 'letter_contact_block' in req_json:
         create_or_update_letter_contact(fetched_service.id, req_json['letter_contact_block'])
 
+    # bridging code between frontend is deployed and data has not been migrated yet. Can only update current year
+    if 'free_sms_fragment_limit' in req_json:
```
richardc0: Do you need to add a test for this?

venusbb: A test has been tested

PR #1395, opened 2017-11-15
Retry unknown exceptions for send_inbound_sms_to_service task

servingUpAces:
If the post to a url for the inbound sms message to the service throwns an unknown exception - or RequestException, the base exception for request, the error will be logged and retried.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -14,7 +14,8 @@ from notifications_utils.template import (
 )
 from requests import (
     HTTPError,
-    request
+    request,
+    RequestException
 )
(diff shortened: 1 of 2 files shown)


venusbb:
Why is this change needed?

leohemsted:
@venusbb we catch errors where the server returned a response that was 500+. However, we weren't previously catching errors where the connection completely failed - for example, because the SSL handshake failed. The actual `request.post` call would raise an exception (rather than `response.raise_for_status`, and it wouldn't have a response object with a status code to interrogate.

PR #1399, opened 2017-11-16
 Added a template_type to the rest call and Aggregation Bug Fix

richardc0:
The check for aggregation was too broad and hence was adding together totals based on template_id and not the unique combination of template id, month and year. The template type is needed for the UI .

- Added test to test for the failure
- Added check and a test to for template_id, month and year matches
- Celery process name did not match the task
- Added type to the rest call

### DIFF
--- a/tests/app/dao/test_services_dao.py
+++ b/tests/app/dao/test_services_dao.py
@@ -1189,9 +1189,9 @@ def test_dao_fetch_monthly_historical_usage_by_template_for_service_does_add_old
         status='delivered'
     )
 
-    template_one = create_sample_template(notify_db, notify_db_session, template_name='1')
-    template_two = create_sample_template(notify_db, notify_db_session, template_name='2')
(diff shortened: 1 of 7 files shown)


tests/app/dao/test_services_dao.py
```diff
@@ -1292,3 +1312,71 @@ def test_dao_fetch_monthly_historical_usage_by_template_for_service_get_this_yea
...
+    assert result[0].month == 10
+    assert result[0].year == 2017
+    assert result[0].count == 1
+
+    assert result[1].template_id == template_one.id
+    assert result[1].name == template_one.name
+    assert result[1].template_type == template_one.template_type
+    assert result[1].month == 11
+    assert result[1].year == 2017
+    assert result[1].count == 1
```
servingUpAces: Do we want to test this with templates of different types? It looks like we are only testing with SMS templates

PR #1400, opened 2017-11-16
Allow include jobs request param on v2 get all notifications

kentsanggds:
## What

In order to get the functional tests to work with the v2 endpoint we need to allow it to handle the `include_jobs` argument to return all notifications including ones created by jobs.

## Reference 

https://www.pivotaltracker.com/story/show/152925357

### DIFF
--- a/tests/app/v2/notifications/test_get_notifications.py
+++ b/tests/app/v2/notifications/test_get_notifications.py
@@ -266,7 +266,8 @@ def test_get_notification_doesnt_have_delivery_estimate_for_non_letters(
     assert 'estimated_delivery' not in json.loads(response.get_data(as_text=True))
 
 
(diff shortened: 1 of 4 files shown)


tests/app/v2/notifications/test_get_notifications.py
```diff
@@ -295,6 +295,34 @@ def test_get_all_notifications_returns_200(client, sample_template):
     assert not json_response['notifications'][0]['scheduled_for']
 
 
+def test_get_all_notifications_with_include_jobs_arg_returns_200(
```
servingUpAces: Do we have a test that checks that notifications created by a job are not included when the param does not exist?

kentsanggds: ok, I see what you mean, can add it

PR #1402, opened 2017-11-17
Add a script to fix migration ordering conflicts

allait:
screen shot 2017-11-17 at 14 09 53

When generating a new migration we give it a number that increments the latest existing migration on master. This means that when there are multiple PRs open containing a migration and one of them gets merged the others need to be updated to move their migration files to apply on top of the recently merged one.

This requires renaming the file and changing migration references for both the migration revision and down_revision.

If a PR introduced more than 1 migration they all need to be updated one after another since each one needs to be renamed.

This adds a script to simplify this process. `./scripts/fix_migrations.py` will check for any branch points If it finds exactly one (which should be the common case), it asks which migration should be moved and renames / updates references to move the selected branch on top of the other one.

It won't resolve any conflicts within migrations themselves (eg if both branches modified the same column) and it won't try to resolve cases with more than 1 branch.

### DIFF
--- a/scripts/fix_migrations.py
+++ b/scripts/fix_migrations.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+import os
+import sys
+
+from alembic.script import ScriptDirectory
+
+sys.path.append('.')
+
+
+def get_branch_points(migrations):
+    return [m for m in migrations.walk_revisions() if m.is_branch_point]
+
+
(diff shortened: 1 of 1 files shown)


venusbb:
Thanks. This is great! Would it be reasonable to include this in run_tests.sh?

leohemsted:
@venusbb this doesn't handle merging - just offers prompts. So we can't have it run on run_tests.sh since that's what runs on jenkins

PR #1404, opened 2017-11-17
Inbound sms pre-implementation test 1

venusbb:
This PR verifies that authorisation token is sent as claimed by the provider. 

Pivotal story: https://www.pivotaltracker.com/story/show/152964252

### DIFF
--- a/app/notifications/receive_notifications.py
+++ b/app/notifications/receive_notifications.py
@@ -56,6 +56,10 @@ def receive_mmg_sms():
 def receive_firetext_sms():
     post_data = request.form
 
+    # This is a pre-implementation test code to validate the provider is sending through what they claim.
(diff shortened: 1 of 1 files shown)


app/notifications/receive_notifications.py
```diff
@@ -56,6 +56,10 @@ def receive_mmg_sms():
 def receive_firetext_sms():
     post_data = request.form
 
+    # This is a pre-implementation test code to validate the provider is sending through what they claim.
+    auth = request.authorization
+    current_app.logger.info("Inbound sms username: {}".format(auth.username))
```
allait: I think `request.authorization` can be `None` when the header is empty, so `auth.username` will raise an AttributeError if we don't wrap this with `if auth: ...`

PR #1406, opened 2017-11-20
Template usage stats updates for FY usage and bug fix

richardc0:
Return financial year not calendar year as the report on the admin UI is FY only and ensure double precision values are not returned from queries as the admin UI convert the number of the month to a word and this requires an int, however the 'extract' database function was returning a double precision float.

- Updated stats_template_usage_by_month_dao.py to return the results for financial year not calendar, as the report os for FY only and hence only the FY data is required
- Updated services_dao.py to ensure double precision values are converted to an int as the 'exact' function returns double precision from the database query, as the admin code requires the value for month to be an int
- Added tests for the above


venusbb:
Should verify it works with BST and UTC.

PR #1409, opened 2017-11-20
Add FIRETEXT_INBOUND_SMS_AUTH config variable and auth check

allait:
Checks authentication header value on inbound SMS requests from
Firetext against a list of allowed API keys set in the application
config.

At the moment, we're only logging the attempts without aborting the
requests. Once this is rolled out to production and we've checked
the logs we'll switch on the aborts and add the tests for 401 and 403
responses.

Pivotal: https://www.pivotaltracker.com/story/show/152964252


venusbb:
Are we confident enough to block in the first instance already? Or should we just pass all the traffic and log the outcome and do xfail in the tests?

PR #1412, opened 2017-11-21
 Template usage always aggregating today's stats 

richardc0:
- Added a check to ensure that the current date falls in between the financial year for the year supplied by the method, so that the todays stats will only be appended in that situation.
- The integration tests did test for zero return. Added a method to test for a year which has no data and also tweaked the existing tests to ensure they are testing the year more fully.

Pivotal: https://www.pivotaltracker.com/story/show/151679396

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -562,47 +562,50 @@ def dao_fetch_monthly_historical_usage_by_template_for_service(service_id, year)
         stats.append(stat)
 
     month = get_london_month_from_utc_column(Notification.created_at)
-    year = func.date_trunc("year", Notification.created_at)
+    year_func = func.date_trunc("year", Notification.created_at)
     start_date = datetime.combine(date.today(), time.min)
 
-    today_results = db.session.query(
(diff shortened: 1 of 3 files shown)


app/dao/services_dao.py
```diff
@@ -562,47 +562,50 @@ def dao_fetch_monthly_historical_usage_by_template_for_service(service_id, year)
...
+
+            if add_to_stats:
+                new_stat = type("", (), {})()
+                new_stat.template_id = today_result.template_id
+                new_stat.template_type = today_result.template_type
+                new_stat.name = today_result.name
+                new_stat.month = int(today_result.month)
+                new_stat.year = int(today_result.year)
+                new_stat.count = today_result.count
+                stats.append(new_stat)
```
venusbb: What is the else (fy_start < datetime.now() < fy_end) here?

richardc0: Is the current date later that the 01/04/<year specified> and before the 31/03/<year specified> if so then add todays stats, else its a previous year so don't add todays stats

PR #1415, opened 2017-11-22
Allow letter templates to select the default contact block

allait:
Paired with @klssmith.

### Verify that attribute exists on *History model when versioning objects

When createing a history instance of the updated object `create_history` sets attributes using `setattr`. Since SQLAlchemy model instances are Python objects they don't prevent new attributes being created by setattr, which means that if history models are missing some of the columns the attributes will still be assigned, but their values will not be persisted by SQLAlchemy since database columns for them do not exist.

To avoid this, we check that the attribute is defined on the `history_cls` and raise an error if it isn't.

### Move common Template/TemplateHistory attributes to a base class

This allows us to avoid duplication between Template and TemplateHistory classes and makes it easier to ensure that all columns are copied to the TemplateHistory objects.

### Add template.service_letter_contact_id and reply_to wrapper property

Adds a relationship between Template models and service letter contact blocks.

Depending on template type, we can have a reference to either a letter contact block, email reply-to address or SMS sender record. This means that in order to enforce foreign key constraints we need to define three separate foreign key columns on the template model.

To hide this implementation detail and make it easier to access the sender/reply-to information we define a wrapper property that returns the value from the correct column.

The relationship and the property are only defined for letter templates at the moment.

The setter raises an error when trying to assign a reply_to value for non-letter templates. The exception isn't raised if the value being assigned is `None` since it can get assigned by marshmallow schemas and as it matches the value returned for other template types it doesn't need to be written anywhere.

### Add reply_to fields to template schemas

We're hiding the `service_letter_contact_id` column since it should only be readable and writable using the `.reply_to` wrapper.

Schemas are defined using `fields.Method` since the fields are represented by a property on the Template model that requires template type to be set. When creating a template, if `reply_to` is defined using `fields.String` it gets assigned at the same time as `template_type` (or the order of assignments is not defined), so the schema loader attempts to set `.reply_to` on a Template object with a `None` `template_type`, which raises an exception.

Using `fields.Method` seems to delay `.reply_to` assignment until the Template object is created, which means it already has a valid type.

### DIFF
--- a/tests/app/template/test_rest_history.py
+++ b/tests/app/template/test_rest_history.py
@@ -3,6 +3,7 @@ from datetime import (datetime, date)
 from flask import url_for
 from app.dao.templates_dao import dao_update_template
 from tests import create_authorization_header
+from tests.app.db import create_letter_contact
 
 
 def test_template_history_version(notify_api, sample_user, sample_template):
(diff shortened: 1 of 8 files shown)


tests/app/template/test_rest_history.py
```diff
@@ -93,3 +94,21 @@ def test_all_versions_of_template(notify_api, sample_template):
             assert json_resp['data'][1]['content'] == newer_content
             assert json_resp['data'][1]['updated_at']
             assert json_resp['data'][2]['content'] == old_content
+
+
+def test_update_template_reply_to_updates_history(client, sample_letter_template):
+    auth_header = create_authorization_header()
```
servingUpAces: The client libraries will need an update

PR #1420, opened 2017-11-23
List templates in alphabetical order

quis:
# Screenshots

Before | After 
--- | ---
image | image

# Ordering templates by created at isn’t working great

Currently templates are ordered by the newest created first. This made sense when, after creating a new template, you were landed on the page that listed all the templates. In other words, you needed to see confirmation of the thing that you’ve just done.

Since https://github.com/alphagov/notifications-admin/pull/1195 (March 2017) you get landed on the page for just that template. So you always see the template you’ve just created, no matter how the list of templates is ordered. So we are free to change the order of the templates.

Ordering by time created is not great, because it gives users no control over which templates appear first. For example, our research reported this from one team:

> One frustration they have is that when you add a new template it automatically goes to the top of the list. To get round this, whenever they add a new template they delete all of the existing ones and start again because they want to keep their templates in numerical order. They'd like to be able to control the order of templates in the list.

# Ordering by alphabetical is better

We _could_ give people some sort of drag-and-drop template ordering thing. But this feels like overkill. I think that alphabetical order is better because:

## It’s discoverable

Anyone who wants to know how a list is ordered can quickly tell just by looking at it

## It’s universal

Everyone knows how alphabetical ordering works

## It’s familiar

This is how people documents on their computer are ordered; there’s no new UI to learn

## It’s the behaviour that users are exhibiting already

From the same service as above:
  > They number their templates 1,2a, 2b, 3a etc

So this commit changes the ordering from newest created first to alphabetical.

# Background reading

Previous changes to template order and navigation:
- https://github.com/alphagov/notifications-admin/pull/1163
- https://github.com/alphagov/notifications-admin/pull/1195
- https://github.com/alphagov/notifications-admin/pull/1330

### DIFF
--- a/app/dao/templates_dao.py
+++ b/app/dao/templates_dao.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 import uuid
 
-from sqlalchemy import desc
+from sqlalchemy import asc, desc
 from sqlalchemy.sql.expression import bindparam
 
 from app import db
@@ -65,14 +65,16 @@ def dao_get_all_templates_for_service(service_id, template_type=None):
             template_type=template_type,
             archived=False
         ).order_by(
-            desc(Template.created_at)
+            asc(Template.name),
+            asc(Template.template_type),
         ).all()
 
     return Template.query.filter_by(
         service_id=service_id,
         archived=False
     ).order_by(
-        desc(Template.created_at)
+        asc(Template.name),
+        asc(Template.template_type),
     ).all()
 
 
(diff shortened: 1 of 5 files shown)


quis:
We’ve looked at the list of templates for 5 of the services with the most templates, and the difference it would make if we changed the sorting to alphabetical. We’ve decided that the benefits of this change outweigh the disruption that it might cause.

---

Also of interest is the number of templates per live service. Many more than when we last looked:

image

quis:
I also had to add sorting by template type, so that the order is deterministic and the tests pass.

But I think this makes sense anyway, for when you have, for example, an email template and a text message template with the same name. If you have two text message templates with
the same name you’re on your own.

PR #1421, opened 2017-11-23
celery.task.retry exc param should be a throwable.

leohemsted:
This causes an issue when it hits the max retry limit, and tries to throw your exception to let you deal with it - at this point it was moaning because we pass in a string

if it's not defined, and we're inside an exception block celery grabs the exc itself.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -323,8 +323,9 @@ def build_dvla_file(self, job_id):
             )
             dao_update_job_status(job_id, JOB_STATUS_READY_TO_SEND)
         else:
-            current_app.logger.info("All notifications for job {} are not persisted".format(job_id))
-            self.retry(queue=QueueNames.RETRY, exc="All notifications for job {} are not persisted".format(job_id))
+            msg = "All notifications for job {} are not persisted".format(job_id)
+            current_app.logger.info(msg)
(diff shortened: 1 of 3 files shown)


app/celery/tasks.py
```diff
@@ -323,8 +323,9 @@ def build_dvla_file(self, job_id):
             )
             dao_update_job_status(job_id, JOB_STATUS_READY_TO_SEND)
         else:
-            current_app.logger.info("All notifications for job {} are not persisted".format(job_id))
-            self.retry(queue=QueueNames.RETRY, exc="All notifications for job {} are not persisted".format(job_id))
+            msg = "All notifications for job {} are not persisted".format(job_id)
+            current_app.logger.info(msg)
+            self.retry(queue=QueueNames.RETRY, exc=Exception(msg))
     except Exception as e:
         # ? should this retry?
```
servingUpAces: We should implement some retry code here. When would it throw an exception? Database throws exception, s3upload throws an exception. Both seem like reasons to retry.

leohemsted: self.retry throws the `exc` parameter. 🤔

leohemsted: yeah so the exception block should retry - might refactor out the retry on line 328 then.

leohemsted:
jenkins is passing - see https://jenkins.notify.tools/job/notifications-api/2147/

PR #1423, opened 2017-11-23
fix 500s when inbound msgs sent from alphanumerics rather than normal…

leohemsted:
… phone numbers

- [ ] https://github.com/alphagov/notifications-utils/pull/257


servingUpAces:
Did we check that a short code or alphanumeric will display properly on the admin app?

leohemsted:
@servingUpAces yup - https://github.com/alphagov/notifications-admin/pull/1661

PR #1424, opened 2017-11-23
Use model constructors when creating history instances

allait:
`create_history` used to initialise an empty model instance and assign attributes to it one by one. Assigned attributes are collected by iterating over internal state of the modified record and stored in a dictionary, so the order in which they are set on the history instance is undefined.

This works fine for most model instances, but recent changes to the Template models require `template_type` to be set before `reply_to`, otherwise Template doesn't know which column to use to persist the `reply_to` value.

This means that we need to order the attributes when they're being assigned. We achieved this by defining a constructor for Template objects that assigns `template_type` first.

In order to make it work with `create_history` we replace individual `setattr` calls with a single call to the Template constructor which passes the entire data dictionary as keyword arguments.

This works because all persisted attributes are defined as columns and SQLAlchemy accepts all column values as constructor arguments.

### DIFF
--- a/tests/app/template/test_rest.py
+++ b/tests/app/template/test_rest.py
@@ -565,6 +565,29 @@ def test_update_set_process_type_on_template(client, sample_template):
     assert template.process_type == 'priority'
 
 
+def test_create_a_template_with_reply_to(admin_request, sample_user):
+    service = create_service(service_permissions=['letter'])
+    letter_contact = create_letter_contact(service, "Edinburgh, ED1 1AA")
+    data = {
+        'name': 'my template',
+        'subject': 'subject',
+        'template_type': 'letter',
+        'content': 'template <b>content</b>',
+        'service': str(service.id),
+        'created_by': str(sample_user.id),
+        'reply_to': str(letter_contact.id),
+    }
+
+    json_resp = admin_request.post('template.create_template', service_id=service.id, _data=data, _expected_status=201)
+
+    assert json_resp['data']['template_type'] == 'letter'
+    assert json_resp['data']['reply_to'] == str(letter_contact.id)
+
+    template = Template.query.get(json_resp['data']['id'])
+    from app.schemas import template_schema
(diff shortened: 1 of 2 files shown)


tests/app/template/test_rest.py
```diff
@@ -616,6 +616,37 @@ def test_update_set_process_type_on_template(client, sample_template):
...
+    auth_header = create_authorization_header()
+
+    response = client.post(
+        '/service/{}/template'.format(service.id),
+        headers=[('Content-Type', 'application/json'), auth_header],
+        data=data
+    )
+    assert response.status_code == 201, response.get_data(as_text=True)
+
+    json_resp = json.loads(response.get_data(as_text=True))
```
leohemsted: if you use the `admin_request` fixture, it does a lot of the gubbins around setting up a request - json handling, headers, URL creation, etc.

```python
json_resp = admin_request.post('template.template', service_id=service.id, _data=data, _expected_status=201)
```

tests/app/template/test_rest.py
```diff
@@ -616,6 +616,37 @@ def test_update_set_process_type_on_template(client, sample_template):
...
+    assert json_resp['data']['template_type'] == 'letter'
+    assert json_resp['data']['reply_to'] == str(letter_contact.id)
+
+    template = Template.query.get(json_resp['data']['id'])
+    from app.schemas import template_schema
```
leohemsted: can this go up the top with the rest of the imports?

allait: I originally copied it from the test above and it looks like all tests delay this import like this.
It turns out 59 tests fail if this is moved to module level. My guess is marshmallow stores some db/session reference somewhere and if that happens before the session fixtures are applied (eg during test module import) it gets stuck with a `DummySession`.

I'm not really sure what the implications of this are.

leohemsted: 😩

PR #1425, opened 2017-11-23
Abort unauthenticated requests for Firetext inbound SMS

allait:
Switches on authentication checks for Firetext inbound SMS callbacks.

This should only be released once Firetext callback URLs have been
updated with authentication details.

### DIFF
--- a/app/notifications/receive_notifications.py
+++ b/app/notifications/receive_notifications.py
@@ -60,10 +60,10 @@ def receive_firetext_sms():
     auth = request.authorization
     if not auth:
         current_app.logger.warning("Inbound sms no auth header")
-        # abort(401)
+        abort(401)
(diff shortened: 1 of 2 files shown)


app/notifications/receive_notifications.py
```diff
@@ -60,10 +60,10 @@ def receive_firetext_sms():
     auth = request.authorization
     if not auth:
         current_app.logger.warning("Inbound sms no auth header")
-        # abort(401)
+        abort(401)
```
servingUpAces: When both providers have authorization, will we migrate this to be handled by the auth similar to the other endpoint validations?

allait: Yeah, we can move it `before_request` if they end up with a similar authentication method.

PR #1438, opened 2017-11-28
Use reply to in send to provider

servingUpAces:
In the previous commit, we recorded the text value of the sender or reply_to_text in the notification. 
In this PR, the value of the reply_to_text from the notification will be sent to the provider who is sending the email, text or letter.


tests/app/delivery/test_send_to_providers.py
```diff
@@ -816,7 +822,7 @@ def test_send_email_to_provider_should_format_reply_to_email_address(
     mocker.patch('app.aws_ses_client.send_email', return_value='reference')
     mocker.patch('app.delivery.send_to_providers.create_initial_notification_statistic_tasks')
 
-    db_notification = create_notification(template=sample_email_template)
+    db_notification = create_notification(template=sample_email_template, reply_to_text="test@test.com\t")
```
leohemsted: is there supposed to be a tab at the end of this reply to?

PR #1439, opened 2017-11-28
Flake8

leohemsted:
We should be following The GDS Way, and using flake8 everywhere. For more information, see here: https://gds-way.cloudapps.digital/manuals/programming-languages/python/linting.html#flake8

Flake8 has a all of pycodestyle's checks, and more. So much more!

So much more, that I haven't enabled it for the tests folder yet - there was way too much to fix there in one PR. Mostly unused variables. If we merge this, we *need* to commit to getting rid of those at some point.

Most of this PR is fairly benign stuff - imports etc. A few serious things though - several tests with the same name that just weren't being run, for example.


app/service/sender.py
```diff
@@ -7,7 +7,9 @@
 from app.notifications.process_notifications import persist_notification, send_notification_to_queue
 
 
-def send_notification_to_service_users(service_id, template_id, personalisation={}, include_user_fields=[]):
+def send_notification_to_service_users(service_id, template_id, personalisation=None, include_user_fields=None):
+    personalisation = personalisation or {}
+    include_user_fields = include_user_fields or []
```
sakisv: Did it actually catch this as PEP violation?

leohemsted: nah, i've just got a hawklike vision for detail

PR #1440, opened 2017-11-28
Migration to create new database table service_callback_api

venusbb:
Create a new table for storing the URL and bearer tokens for sending the receipts for services. 
- create new table service_callback_api, model and migration script
- refractor the previous migration script filename (duplicated)
Pivotal: https://www.pivotaltracker.com/story/show/153087419


app/models.py
```diff
@@ -466,6 +466,40 @@ def serialize(self):
...
+    @property
+    def bearer_token(self):
+        if self._bearer_token:
+            return encryption.decrypt(self._bearer_token)
+        return None
+
+    @bearer_token.setter
+    def bearer_token(self, bearer_token):
+        if bearer_token:
+            self._bearer_token = encryption.encrypt(str(bearer_token))
```
leohemsted: I don't think this should be encrypted in the database - we need to read it out in plain text anyway to add it to the outbound message

so i think we can just have this as a normal field rather than a property

venusbb: It is not right to store someone's secret in the database in plain text. We should mask this. We can always change this if we go for a more secure algorithm in the future.

servingUpAces: This is the same technique used for api keys, bearer tokens for the inbound api, and personalisation.

venusbb: After discussion, we will keep the coherence of the database encode method and refractor the 'encrypt' 'decrypt' keywords in a separate PR.

PR #1457, opened 2017-12-01
Bye bye free fragments

leohemsted:
👋 

You still pass them in when you create a service - because the list of how many each type of org (local, nhs, central) get is stored on admin. But we don't update.

This needs to be merged before https://github.com/alphagov/notifications-api/pull/1458

- [x] https://github.com/alphagov/notifications-admin/pull/1682

### DIFF
--- a/app/billing/rest.py
+++ b/app/billing/rest.py
@@ -14,7 +14,7 @@ from app.utils import convert_utc_to_bst
 from app.dao.annual_billing_dao import (dao_get_free_sms_fragment_limit_for_year,
                                         dao_get_all_free_sms_fragment_limit,
                                         dao_create_or_update_annual_billing_for_year,
-                                        dao_update_annual_billing_for_current_and_future_years)
+                                        dao_update_annual_billing_for_future_years)
 from app.billing.billing_schemas import create_or_update_free_sms_fragment_limit_schema
 from app.errors import InvalidRequest
 from app.schema_validation import validate
@@ -155,7 +155,7 @@ def update_free_sms_fragment_limit_data(service_id, free_sms_fragment_limit, fin
     # if we're trying to update historical data, don't touch other rows.
     # Otherwise, make sure that future years will get the new updated value.
     if financial_year_start >= current_year:
-        dao_update_annual_billing_for_current_and_future_years(
+        dao_update_annual_billing_for_future_years(
             service_id,
             free_sms_fragment_limit,
             financial_year_start
--- a/app/billing/rest.py
+++ b/app/billing/rest.py
@@ -117,6 +117,7 @@ def get_free_sms_fragment_limit(service_id):
                 financial_year_start = get_current_financial_year_start_year()
 
             if int(financial_year_start) < get_current_financial_year_start_year():
+                # return the earliest historical entry
                 annual_billing = sms_list[0]   # The oldest entry
             else:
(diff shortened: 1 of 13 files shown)


app/service/rest.py
```diff
@@ -154,19 +153,24 @@ def get_service_by_id(service_id):
 @service_blueprint.route('', methods=['POST'])
 def create_service():
     data = request.get_json()
-    if not data.get('user_id', None):
-        errors = {'user_id': ['Missing data for required field.']}
+    errors = {
+        required_field: ['Missing data for required field.']
+        for required_field in ['user_id', 'free_sms_fragment_limit']
+        if not data.get(required_field, None)
+    }
```
venusbb: data.get(required_field) without 'None' in the second argument seems to achieve the same thing.

app/service/rest.py
```diff
@@ -181,10 +185,6 @@ def update_service(service_id):
     update_dict = service_schema.load(current_data).data
     dao_update_service(update_dict)
 
-    # bridging code between frontend is deployed and data has not been migrated yet. Can only update current year
-    if 'free_sms_fragment_limit' in req_json:
-        update_free_sms_fragment_limit_data(fetched_service.id, req_json['free_sms_fragment_limit'])
-
     if service_going_live:
```
venusbb: We need this. Updating free_sms_fragment)limit still go through the service_api_client in admin.

tests/app/service/test_rest.py
```diff
@@ -133,16 +132,8 @@ def test_get_service_by_id(admin_request, sample_service):
...
-
-
-def test_get_service_by_id_returns_free_sms_limit(admin_request, sample_service):
-    json_resp = admin_request.get('service.get_service_by_id', service_id=sample_service.id)
-    assert json_resp['data']['free_sms_fragment_limit'] == current_app.config['FREE_SMS_TIER_FRAGMENT_COUNT']
-
-
-def test_get_detailed_service_by_id_returns_free_sms_limit(admin_request, sample_service):
-    json_resp = admin_request.get('service.get_service_by_id', service_id=sample_service.id, detailed=True)
-    assert json_resp['data']['free_sms_fragment_limit'] == current_app.config['FREE_SMS_TIER_FRAGMENT_COUNT']
```
venusbb: We need this. Updating free_sms_fragment)limit still go through the service_api_client in admin.

leohemsted:
@venusbb I'd rather do the admin work before merging this, rather than keeping those changes.

app/billing/rest.py
```diff
@@ -141,10 +142,21 @@ def create_or_update_free_sms_fragment_limit(service_id):
...
+        service_id,
+        free_sms_fragment_limit,
+        financial_year_start
+    )
+    # if we're trying to update historical data, don't touch other rows.
+    # Otherwise, make sure that future years will get the new updated value.
+    if financial_year_start >= current_year:
+        dao_update_annual_billing_for_current_and_future_years(
+            service_id,
+            free_sms_fragment_limit,
```
venusbb: This would work, but if the financial_year_start >= current_year, it will do the updating twice. 

We could possibly change dao_update_annual_billing_for_current_and_future_years to dao_create_and_update_annual_billing_for_current_and_future_years, so

if financial_year_start>=current_year
       # create and update multiple rows
otherwise
      # just update one row

PR #1459, opened 2017-12-01
Send delivery status back to services for email, letter and sms

venusbb:
Continue from this PR: https://github.com/alphagov/notifications-api/pull/1447
When a callback receipts/status update is received from the providers for email, letter and sms, we check if there is an entry in service_callback_api, and use the url and bearer token set to send the delivery statuses back to the services. 
In this PR, 
* Added send_delivery_status_to_service task worker in service_callback_tasks file
* Rename callback_tasks.py to process_ses_receipts_tasks.py
* Use send_delivery_status_to_service in sms, ses and letter callbacks

**TODO:** consider to use a different queue to send delivery status receipts 

Pivotal: https://www.pivotaltracker.com/story/show/153087419


app/celery/tasks.py
```diff
@@ -472,6 +477,9 @@ def update_letter_notifications_statuses(self, filename):
                 current_app.logger.info(
                     'DVLA file: {filename}, notification updated to {status}: {reference}'.format(
                         filename=filename, status=status, reference=str(update.reference)))
+                notifications = dao_get_notifications_by_reference(references=[update.reference])
```
servingUpAces: We can check if the service has a callback url first, that way we do not need to create a task on the queue.

app/celery/tasks.py
```diff
@@ -472,6 +477,9 @@ def update_letter_notifications_statuses(self, filename):
...
+                for notification in notifications:
+                    send_delivery_status_to_service.apply_async([notification.id], queue=QueueNames.NOTIFY)
```
servingUpAces: Need to wrap notification_id with str(notification_id) otherwise there will be a serialisation error

app/notifications/notifications_ses_callback.py
```diff
@@ -75,7 +77,7 @@ def process_ses_response(ses_request):
                 )
 
             create_outcome_notification_statistic_tasks(notification)
-
+            send_delivery_status_to_service.apply_async([notification.id], queue=QueueNames.NOTIFY)
```
servingUpAces: Need to wrap notification_id with str(notification_id) otherwise there will be a serialisation error

app/notifications/process_client_response.py
```diff
@@ -83,5 +84,7 @@ def process_sms_client_response(status, reference, client_name):
 
     create_outcome_notification_statistic_tasks(notification)
 
+    send_delivery_status_to_service.apply_async([notification.id], queue=QueueNames.NOTIFY)
```
servingUpAces: Need to wrap notification_id with str(notification_id) otherwise there will be a serialisation error

app/celery/service_callback_tasks.py
```diff
@@ -0,0 +1,71 @@
...
+from app.config import QueueNames
+
+
+@notify_celery.task(bind=True, name="send-delivery-status", max_retries=5, default_retry_delay=300)
+@statsd(namespace="tasks")
+def send_delivery_status_to_service(self, notification_id):
+    # TODO: do we need to do rate limit this?
+    notification = get_notification_by_id(notification_id)
+    service_callback_api = get_service_callback_api_for_service(service_id=notification.service_id)
+    if not service_callback_api:
```
servingUpAces: We should probably check for the service_callback_api before creating this task.

app/celery/tasks.py
```diff
@@ -391,6 +393,9 @@ def update_letter_notifications_to_error(self, notification_references):
     )
 
     current_app.logger.info("Updated {} letter notifications to technical-failure".format(updated_count))
+    notifications = dao_get_notifications_by_reference(references=notification_references)
```
servingUpAces: We should probably check for the service_callback_api before creating this task.

app/celery/tasks.py
```diff
@@ -391,6 +393,9 @@ def update_letter_notifications_to_error(self, notification_references):
...
+    for notification in notifications:
+        send_delivery_status_to_service.apply_async([notification.id], queue=QueueNames.NOTIFY)
```
servingUpAces: You will need to wrap notification with str(notification_id)

app/dao/notifications_dao.py
```diff
@@ -459,6 +459,13 @@ def dao_get_notifications_by_to_field(service_id, search_term, statuses=None):
     return results
 
 
+@statsd(namespace="dao")
+def dao_get_notifications_by_reference(references):
```
servingUpAces: Perhaps we should call this dao_get_notifications_by_references, to indicate we are passing in a list of references.

app/notifications/process_client_response.py
```diff
@@ -83,5 +84,7 @@ def process_sms_client_response(status, reference, client_name):
 
     create_outcome_notification_statistic_tasks(notification)
```
servingUpAces: We could check for the existence of service_callback_api before creating the task.

app/celery/service_callback_tasks.py
venusbb: Agree. I will leave this check here as we need to get service_callback_api anyway. But will add the check before putting the task on queue.

app/celery/tasks.py
venusbb: changed in 3594fc.

app/celery/tasks.py
venusbb: changed in 3594fc.

app/celery/tasks.py
venusbb: changed in 3594fc.

app/celery/tasks.py
venusbb: changed in 3594fc.

app/dao/notifications_dao.py
venusbb: changed in 3594fc.

app/notifications/notifications_ses_callback.py
venusbb: changed in 3594fc.

app/notifications/process_client_response.py
venusbb: changed in 3594fc.

app/notifications/process_client_response.py
venusbb: changed in 3594fc.

tests/app/celery/test_process_ses_receipts_tasks.py
```diff
@@ -18,15 +18,15 @@ def test_process_ses_results(sample_email_template):
```
servingUpAces: Do we need a test that if the service doesn't have a url the task is not created?

PR #1462, opened 2017-12-04
Add letters pdf s3 bucket

kentsanggds:
## What 

Add function to enable uploading of letters pdfs to an s3 bucket

## Reference 

https://www.pivotaltracker.com/story/show/151511579


app/aws/s3.py
```diff
@@ -76,3 +79,23 @@ def remove_transformed_dvla_file(job_id):
     file_location = '{}-dvla-job.text'.format(job_id)
     obj = get_s3_object(bucket_name, file_location)
     return obj.delete()
+
+
+def upload_letters_pdf(reference, crown, filedata):
+    now = datetime.utcnow()
+    upload_file_name = LETTERS_PDF_FILE_LOCATION_STRUCTURE.format(
+        folder=now.date().isoformat(),
```
servingUpAces: Why are we using date as the folder name? Not saying its wrong.

kentsanggds: Just thought it would be easier to download a folder for zipping up rather than something tied to a service.

PR #1464, opened 2017-12-05
Add crown column to services

servingUpAces:
A service can be crown or non-crown, this adds an attribute to services to identify which they are. We need to know if a service is crown or non-crown because the letter rates are different. 

This PR add crown to the services table and populates the column. When a service is created or updated crown is set based on organisation type, however, it is possible that a service does not conform to the decision. It has been decided that if the value of crown needs to be updated that it will be done from the db. This is because some changes will happen for the organisation reporting epic, and it will be done at that time.


venusbb:
If “non crown” == “not central”, why do we need a separate column to store this?

servingUpAces:
crown == central is just the default. But there are cases where this isn't true. And a service can be the opposite of this.

PR #1469, opened 2017-12-06
Revert changes to staging for performance tests and increased rate limit

richardc0:
- Reverted the Gunicorn worker number to 5 (this should be investigated further on a well baselined system to compare)
- Enabled REDIS
- Increased the rate limit to 400 req/sec as using early testing yesterday 450+ was being achieved so brings it more inline with what can be acheived

### DIFF
--- a/manifest-api-staging.yml
+++ b/manifest-api-staging.yml
@@ -1,7 +1,6 @@
 ---
 
 inherit: manifest-api-base.yml
-command: scripts/run_app_paas.sh gunicorn -c /home/vcap/app/gunicorn_config.py --error-logfile /home/vcap/logs/gunicorn_error.log -w 6 -b 0.0.0.0:$PORT application
 services:
   - notify-aws
   - notify-config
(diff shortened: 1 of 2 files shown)


manifest-api-staging.yml
```diff
@@ -1,7 +1,6 @@
 ---
 
 inherit: manifest-api-base.yml
-command: scripts/run_app_paas.sh gunicorn -c /home/vcap/app/gunicorn_config.py --error-logfile /home/vcap/logs/gunicorn_error.log -w 6 -b 0.0.0.0:$PORT application
 services:
```
venusbb: Shouldn't you change the worker number rather than deleting the line here?

venusbb:
It was 210000 before the change what do you mean by "using early testing yesterday 450+"? I can see the new change to 240000 is to support 400 request/sec.

richardc0:
The 21000 was an arbitrary amount (a little more than the ~280 req/sec we need to achieve 1 million an hour). The new amount just allows us to test the system with everything turned on (rate limit etc) at that higher rate. (Current load + new load) < 450 req/sec, as that is what SES is set to and I didn't want to top out as the functional tests might not work etc.

manifest-api-staging.yml
servingUpAces: This is overriding the command line in the manifest-api-base.yml file. Deleting this line means that staging will use the default.

PR #1473, opened 2017-12-07
log auth errors from the v2 api

leohemsted:
(will help with tracking down client problems)

We currently don't log anything - but it'll be useful to see if services are getting lots of invalid system clock, or key expired, etc etc.

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -92,7 +92,7 @@ def requires_auth():
         g.service_id = api_key.service_id
         _request_ctx_stack.top.authenticated_service = service
         _request_ctx_stack.top.api_user = api_key
-        current_app.logger.info('Succesful login for service {} with api key {}, using client'.format(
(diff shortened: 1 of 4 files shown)


leohemsted:
@servingUpAces i've changed it - now it logs the service id and api key id where possible, to help us figure out what's going on when we're investigating unusal API usage patterns

a bit of a bigger change as i wanted to update some of the tests to make sure this'd work

leohemsted:
example log messages:

```
"API AuthError, client: NOTIFY-API-PYTHON-CLIENT/4.7.0 error: AuthError({'token': ['Invalid token: signature, api token not found']}, 403, service_id=3e1ed7ea-8a05-4b4e-93ec-d7bebfea6cae, api_key_id=None)
API authorised for service 3e1ed7ea-8a05-4b4e-93ec-d7bebfea6cae with api key 604807b2-8ec8-4ed2-abc1-507e5169cfdb, using client NOTIFY-API-PYTHON-CLIENT/4.7.0
```

PR #1476, opened 2017-12-07
modified dao_timeout_notifications

venusbb:
As part of the pivotal story of https://www.pivotaltracker.com/story/show/153087419

We should also send delivery status from calling scheduled timeout tasks as it is one of the final status update place.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -58,6 +58,8 @@ from app.celery.tasks import (
 from app.config import QueueNames, TaskNames
 from app.utils import convert_utc_to_bst
 from app.v2.errors import JobIncompleteError
+from app.dao.service_callback_api_dao import get_service_callback_api_for_service
+from app.celery.service_callback_tasks import send_delivery_status_to_service
 
 
 @worker_process_shutdown.connect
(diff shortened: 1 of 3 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -189,10 +191,18 @@ def delete_invitations():
...
-    updated = dao_timeout_notifications(current_app.config.get('SENDING_NOTIFICATIONS_TIMEOUT_PERIOD'))
-    if updated:
+    notifications = dao_timeout_notifications(current_app.config.get('SENDING_NOTIFICATIONS_TIMEOUT_PERIOD'))
+
+    if notifications:
+        for notification in notifications:
+            # queue callback task only if the service_callback_api exists
+            service_callback_api = get_service_callback_api_for_service(service_id=notification.service_id)
+
+            if service_callback_api:
```
richardc0: Is this normal operation i.e. if a service hasn't got it configured then service_callback_api is None? Or is this defensive as in something failed?

If the latter add a log statement etc

venusbb: Yes. That's correct. We don't want to put a task on the queue if they haven't got it configure.

PR #1479, opened 2017-12-08
Add letters pdf task & api

kentsanggds:
## What

- Added a task to create a cmyk pdf and store this in the letters-pdf bucket on s3.
- Only allow this for services that have `letters_as_pdf` set
- Ensure that for these services the `build_dvla_file` is not called so that DVLA only get pdfs or csv files to process.
- Ensure that letter notifications API calls the create letters pdf task when in `letters_as_pdf` service
- Make sure that only notifications not in `letters_as_pdf` services are processed for api calls that are building dvla files.

## How to manually test

- From admin settings, switch a service to send letters as pdf
- then from a letter template upload a csv file
  - you should be able to see the newly created CMYK files on the s3 bucket for that environment
  - or from celery logs you should be able to see something like this -
  `PDF Letter [noti id] reference [reference] created at [datetime], [pdf size] bytes`
  `Uploading letters [pdf file location] to [environment]-letters-pdf`

- From an API call - 
`PYTHONPATH=. python ./utils/make_api_call.py http://localhost:6011 [test key for service with letters_as_pdf] create --type=letter --template=[letter template id] --personalisation="{\"address_line_1\":\"someone\", \"address_line_2\":\"London\", \"postcode\":\"N1\"}" --reference="test letters"`

## Reference 

Step 4/5 https://www.pivotaltracker.com/story/show/151511579

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -5,7 +5,6 @@ from requests import (
 )
 
 from botocore.exceptions import ClientError as BotoClientError
-from sqlalchemy.orm.exc import NoResultFound
 
 from app import notify_celery
 from app.aws import s3
@@ -18,9 +17,7 @@ from app.statsd_decorators import statsd
 @statsd(namespace="tasks")
 def create_letters_pdf(self, notification_id):
     try:
-        notification = get_notification_by_id(notification_id)
-        if not notification:
-            raise NoResultFound()
+        notification = get_notification_by_id(notification_id, _raise=True)
 
         pdf_data = get_letters_pdf(
             notification.template,
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -30,7 +30,7 @@ def create_letters_pdf(self, notification_id):
         )
         current_app.logger.info("PDF Letter {} reference {} created at {}, {} bytes".format(
             notification.id, notification.reference, notification.created_at, len(pdf_data)))
-        s3.upload_letters_pdf(reference=notification.reference, crown=True, filedata=pdf_data)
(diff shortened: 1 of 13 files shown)


app/dao/notifications_dao.py
```diff
@@ -541,14 +543,28 @@ def dao_set_created_live_letter_api_notifications_to_pending():
...
+
     notifications = db.session.query(
         Notification
     ).filter(
         Notification.notification_type == LETTER_TYPE,
         Notification.status == NOTIFICATION_CREATED,
         Notification.key_type == KEY_TYPE_NORMAL,
-        Notification.api_key != None  # noqa
+        Notification.api_key != None,  # noqa
+        Notification.service_id.in_(services_without_letters_as_pdf)
```
servingUpAces: Might be worth checking the performance of this. You may want to get the list of service ids outside of this query rather than a subquery.

app/celery/letters_pdf_tasks.py
```diff
@@ -0,0 +1,68 @@
...
+from app.dao.notifications_dao import get_notification_by_id, update_notification_status_by_id
+from app.statsd_decorators import statsd
+
+
+@notify_celery.task(bind=True, name="create-letters-pdf", max_retries=15, default_retry_delay=300)
+@statsd(namespace="tasks")
+def create_letters_pdf(self, notification_id):
+    try:
+        notification = get_notification_by_id(notification_id)
+        if not notification:
```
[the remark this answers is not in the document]
kentsanggds: ok, will update

app/dao/notifications_dao.py
kentsanggds: The join appears to be marginally faster than the sub query, so will implement that.

PR #1485, opened 2017-12-12
put pdfs in tomorrow's dvla bucket after 17:30

leohemsted:
So if someone sends a letter in the evening, it gets picked up the
next day


kentsanggds:
We probably need to consider BST as we are now putting files to the next day folder that occur after 17:30.

leohemsted:
@kentsanggds the task triggers at 17:30UTC. We probably need to think about what that means during summer when it triggers at 18:30BST, but don't need to worry about BST here

PR #1488, opened 2017-12-13
Update billable units for letters pdf task

kentsanggds:
## What 

As we have access to the page count from the template preview service print.pdf endpoint, we should be able to update the billable units using this information after uploading the file to the s3 bucket. 

A billable letter unit is counted per sheet which consists of 2 pages, so to calculate it I use this calculation - 

`billable_units = celiing( page count / 2 )`

| page count | billable_units |
|------------|----------------|
| 1          | 1              |
| 2          | 1              |
| 3          | 2              |
| 4          | 2              |

## Reference

Step 5/5 - https://www.pivotaltracker.com/story/show/151511579

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 from flask import current_app
 import math
 from requests import (
@@ -14,7 +13,7 @@ from app.config import QueueNames
 from app.dao.notifications_dao import (
     get_notification_by_id,
     update_notification_status_by_id,
-    dao_update_notifications_by_reference
+    dao_update_notification
 )
 from app.statsd_decorators import statsd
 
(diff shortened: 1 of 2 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -28,6 +34,24 @@ def create_letters_pdf(self, notification_id):
...
             notification.id, notification.reference, notification.created_at, len(pdf_data)))
         s3.upload_letters_pdf(reference=notification.reference, crown=notification.service.crown, filedata=pdf_data)
+
+        updated_count = dao_update_notifications_by_reference(
+            references=[notification.reference],
+            update_dict={
+                "billable_units": billable_units,
+                "updated_at": datetime.utcnow()
+            }
+        )
```
leohemsted: i'd just call the normal dao_update_notification method here since you've still got the notification object


```
notification.billable_units = billable_units
```

then we won't need the error logging below either

tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -59,6 +88,28 @@ def test_create_letters_pdf_calls_upload_letters_pdf(mocker, sample_letter_notif
...
+def test_create_letters_pdf_sets_billable_units(mocker, sample_letter_notification):
+    mocker.patch('app.celery.letters_pdf_tasks.get_letters_pdf', return_value=(b'\x00\x01', 1))
+    mocker.patch('app.celery.tasks.s3.upload_letters_pdf')
+
+    create_letters_pdf(sample_letter_notification.id)
+    noti = Notification.query.filter(Notification.reference == sample_letter_notification.reference).one()
+    assert noti.billable_units == 1
+
+
+def test_create_letters_pdf_handles_update_failure(mocker, sample_letter_notification):
```
leohemsted: won't need this test either

PR #1491, opened 2017-12-14
if apps crash on startup, then fail deploy process

leohemsted:
we saw an issue where the app started, then immediately crashed due to a setup error. However, jenkins had already returned positively, and the deploy continued.

cf-deploy should fail if the app doesn't start up.

We do this by looking through the cloudfoundry events, and aborting if there are any `app.crash` events for the new GUID.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -281,6 +281,9 @@ cf-deploy: ## Deploys the app to Cloud Foundry
 	# sleep for 10 seconds to try and make sure that all worker threads (either web api or celery) have finished before we delete
 	# when we delete the DB is unbound from the app, which can cause "permission denied for relation" psycopg2 errors.
 	sleep 10
+
+	# get the new GUID, and find all crash events for that. If there were any crashes we will abort the deploy.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -281,6 +281,9 @@ cf-deploy: ## Deploys the app to Cloud Foundry
 	# sleep for 10 seconds to try and make sure that all worker threads (either web api or celery) have finished before we delete
 	# when we delete the DB is unbound from the app, which can cause "permission denied for relation" psycopg2 errors.
 	sleep 10
+
+	# get the new GUID, and find all crash events for that. If there were any crashes we will abort the deploy.
+	[ $(cf curl "/v2/events?q=type:app.crash&q=actee:$$(cf app --guid notify-delivery-worker-receipts)" | jq ".total_results") -eq 0 ]
```
sakisv: You need `$$` in the beginning otherwise it fails cause it will try to interpret it as `make` variable.

PR #1492, opened 2017-12-14
Add MMG_INBOUND_SMS_AUTH config variable and auth check

klssmith:
Checks authentication header value on inbound SMS requests from
MMG  against a list of allowed API keys set in the application
config.

At the moment, we're only logging the attempts without aborting the
requests. Once this is rolled out to production and we've checked
the logs we'll switch on the aborts and add the tests for 401 and 403
responses.

This work has already been done for Firetext in a previous PR:
https://github.com/alphagov/notifications-api/pull/1409

Paired with @venusbb 

To be merged after deploying new env variables

### DIFF
--- a/app/notifications/receive_notifications.py
+++ b/app/notifications/receive_notifications.py
@@ -30,6 +30,15 @@ def receive_mmg_sms():
     """
     post_data = request.get_json()
 
+    auth = request.authorization
+
+    if not auth:
+        current_app.logger.warning("Inbound sms (MMG) no auth header")
+        # abort(401)
+    elif auth.username != 'mmg_co_2017' or auth.password not in current_app.config['MMG_INBOUND_SMS_AUTH']:
+        current_app.logger.warning("Inbound sms (MMG) incorrect username ({}) or password".format(auth.username))
+        # abort(403)
+
     inbound_number = strip_leading_forty_four(post_data['Number'])
 
     service = fetch_potential_service(inbound_number, 'mmg')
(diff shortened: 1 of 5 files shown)


klssmith:
This will need the right credentials to be added before it can be merged.

app/notifications/receive_notifications.py
```diff
@@ -30,6 +30,15 @@ def receive_mmg_sms():
     """
     post_data = request.get_json()
 
+    auth = request.authorization
+
+    if not auth:
+        current_app.logger.warning("Inbound sms (MMG) no auth header")
+        # abort(401)
+    elif auth.username != 'mmg_co_2017' or auth.password not in current_app.config['MMG_INBOUND_SMS_AUTH']:
```
richardc0: Should the username be an environment variable too? Its giving away half of the credentials.

tests/app/notifications/test_receive_notification.py
```diff
@@ -37,14 +37,21 @@ def firetext_post(client, data, auth=True, password='testkey'):
...
         data=json.dumps(data),
-        headers=[
-            ('Content-Type', 'application/json'),
-            ('X-Forwarded-For', '203.0.113.195, 70.41.3.18, 150.172.238.178')
-        ])
+        headers=headers
+    )
 
 
 def test_receive_notification_returns_received_to_mmg(client, mocker, sample_service_full_permissions):
```
richardc0: The if and else blocks are not really being tested can you mock the logger and check they are being called?

app/notifications/receive_notifications.py
venusbb: Will do in the PR that enables blocking.

tests/app/notifications/test_receive_notification.py
venusbb: Only wants to observe the blocking at the moment. If we take away the commented blocking codes and the skipped statement, the tests would pass and the correct if/else is being called.

PR #1506, opened 2017-12-18
Remove the delivery receipt callback for letters

klssmith:
This has been removed because the services only see one status.

Pivotal story: https://www.pivotaltracker.com/story/show/153674962

Paired with @allait


venusbb:
Would it be better to add a feature flag to disable this for now? Are we sure we won't need to send delivery status for letters at all?

servingUpAces:
No we will not need it. A feature flag is not necessary.

PR #1507, opened 2017-12-18
Add run letter pdfs

kentsanggds:
## What

Added a scheduled task which will run at 17:50 in order to kick off processing for zipping letter pdfs.

## Reference 

Step 1 / 5 - https://www.pivotaltracker.com/story/show/151511632

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -1,4 +1,5 @@
 from datetime import (
+    date,
     datetime,
     timedelta
 )
@@ -33,6 +34,7 @@ from app.dao.notifications_dao import (
     dao_timeout_notifications,
     is_delivery_slow_for_provider,
     delete_notifications_created_more_than_a_week_ago_by_type,
+    dao_get_count_of_letters_to_process_for_date,
     dao_get_scheduled_notifications,
     set_scheduled_notification_to_processed,
     dao_set_created_live_letter_api_notifications_to_pending,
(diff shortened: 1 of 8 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -355,6 +357,20 @@ def run_letter_jobs():
         current_app.logger.info("Queued {} ready letter job ids onto {}".format(len(job_ids), QueueNames.PROCESS_FTP))
 
 
+@notify_celery.task(name="run-letter-pdfs")
+@statsd(namespace="tasks")
+def run_letter_pdfs():
```
servingUpAces: Can we give this a better name? What is it actually doing?

manifest-delivery-base.yml
```diff
@@ -51,7 +51,7 @@ applications:
       NOTIFY_APP_NAME: delivery-worker-priority
 
   - name: notify-delivery-worker
-    command: scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=11 -Q job-tasks,retry-tasks,notify-internal-tasks,create-letters-pdf-tasks
+    command: scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=11 -Q job-tasks,retry-tasks,notify-internal-tasks,create-letters-pdf-tasks,letter-tasks
```
venusbb: Should we be doing with delivery-worker-periodic worker?

venusbb:
As an option, we could store the meta data (filename and size) in DB for the day (a period) rather than iterate them from the meta data from S3.

venusbb:
The "collate-letter-pdfs-for-day" function do 3 things
1. Collect meta-data (file sizes and names) from S3 for all the files that belong to today
2. Group them in chunk of 1G file size to be zipped up
3. send them to letter queue to be processed by FTP worker. 

It was discussed that because the master record is in the notifications table. If there is any problem (lost files, etc), we would like to be able to be able to perform the checking logic in API. And we should be able to trigger this task independently from the scheduler for a certain day.

PR #1510, opened 2017-12-19
add collate-letter-pdfs task

leohemsted:
add collate-letter-pdfs task (name pending). This retrieves a list of
letter pdf files (just the metadata, not the actual data) from s3, and
loops through them, calling the ftp task zip-and-send-letter-pdfs. It
groups them up by adding them to lists while counting the total
filesize, if it gets over a certain filesize (currently set to 500mb)
it breaks at that chunk, sends off that list of files to the ftp app,
and then starts building up a new list.

DVLA have a hard 2gb limit on how big the zip files we can send is -
however we're going to be limited by the amount of memory on the ftp
app well before we get around to handling 2gb of pdf data - so the
limit is 500mb for now. We'll adjust it after we see how ftp performs.

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -97,7 +97,7 @@ def collate_letter_pdfs_for_day(date):
         )
         notify_celery.send_task(
             name=TaskNames.ZIP_AND_SEND_LETTER_PDFS,
-            kwargs={'filenames': filenames},
+            kwargs={'filenames_to_zip': filenames},
             queue=QueueNames.PROCESS_FTP
         )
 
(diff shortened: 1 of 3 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -79,3 +79,45 @@ def get_letters_pdf(template, contact_block, org_id, values):
...
+
+
+@notify_celery.task(name='collate-letter-pdfs-for-day')
+def collate_letter_pdfs_for_day(date):
+    letter_pdfs = s3.get_s3_bucket_objects(
+        current_app.config['LETTERS_PDF_BUCKET_NAME'],
+        subfolder=date
+    )
+    for letters in group_letters(letter_pdfs):
+        filenames = [letter['Key'] for letter in letters]
```
kentsanggds: Just tried it and because we are creating zip files in the same folder on FTP it's probably a good idea to make sure that only files with `.PDF` extensions should be in the array otherwise other files like zip files could be added to the zip archive. Could be argued that it will only be generated once and so it won't have the zip file in the folder but if we were to reprocess a day then the zip file might be present.

app/celery/letters_pdf_tasks.py
```diff
@@ -79,3 +79,45 @@ def get_letters_pdf(template, contact_block, org_id, values):
...
+        current_app.logger.info(
+            'Calling task zip-and-send-letter-pdfs for {} pdfs of total size {:,} bytes'.format(
+                len(filenames),
+                sum(letter['Size'] for letter in letters)
+            )
+        )
+        notify_celery.send_task(
+            name=TaskNames.ZIP_AND_SEND_LETTER_PDFS,
+            kwargs={'filenames': filenames},
```
kentsanggds: The name I've given in FTP is actually `filenames_to_zip`, feel free to update FTP to `filenames` in line with the API instead in order to get it working with it.

PR #1514, opened 2017-12-20
Added SQLALCHEMY settings to staging for db connections

richardc0:
Updated SQL Alchemy settings to implement in part optimistic disconnect handling. On staging there is little traffic and as such connections can become stale. Whilst this is less apparent on production planned upgrade times are likely to be in periods of low activity and hence are more likely to the become stale. It is hoped that there will be less errors thrown after a page is visited if the following settings are used as connections are being recycled.

There is no direct correlation between SQL alchecmy and postgres but interpreting the postgres values  (https://www.postgresql.org/docs/9.5/static/runtime-config-connection.html) => (tcp_keepalives_idle = 300 + tcp_keepalives_interval = 30) x tcp_keepalives_count is 2 so that would be 11 mins before its considered dead? Is how I read it. Hence the below values are halving that.

More info on SQL Alchemy settings here: http://flask-sqlalchemy.pocoo.org/2.3/config/#configuration-keys
- Updated SQLALCHEMY_POOL_TIMEOUT timeout idle connections after 30 secs
- Updated SQLALCHEMY_POOL_RECYCLE to recycle the connection every 5 mins
   See guide on optimistic disconnect handling - using the pool recycle as a way to manage this:
   http://docs.sqlalchemy.org/en/latest/core/pooling.html#disconnect-handling-optimistic

Example: 

Current workflow: Database is moved to failover / upgraded -> DNS updated -> old connection becomes stale -> page visited -> db operation fails -> reconnect

New workflow: Database is moved to failover / upgraded -> DNS updated -> old connection becomes stale -> connection recycled -> page visited


leohemsted:
I don't really see what this (especially `SQLALCHEMY_POOL_RECYCLE`) will solve - I'm not concerned about periodic workers failing an hour later. I am concerned about `notify-delivery-worker-sender` and `notify-api`, both of which have far more frequent activity than either of these values.

Also I'd like some documentation, either in a comment, commit message, or PR message, about what these numbers do and what we did to arrive at these values (and what they default to for postgres)

richardc0:
Yep should have added more docs, so added to the PR. In my defence we'd discussed at a meeting and the things to change.

Basically this is a change to staging (only) to test another upgrade. If the database is down, its down but this should hopefully reduce the amount of stale connections and reduce the amount of errors e.g.  SSL connection has been closed unexpectedly [SQL: 'SELECT version_num FROM alembic_version']

PR #1532, opened 2017-12-28
Clean makefile

sakisv:
Ready for review, but don't merge until alphagov/notifications-aws#292 has been merged/deployed.

EDIT:
And of course jenkins is going to fail until then


Makefile
```diff
@@ -220,7 +139,7 @@ test-with-docker: prepare-docker-build-image create-docker-test-db ## Run tests
 		${DOCKER_BUILDER_IMAGE_NAME} \
 		gosu hostuser make test
 
-.PHONY: test-with-docker
+.PHONY: create-docker-test-db
```
leohemsted: 🤔

sakisv: Probably a copy/paste leftover. `.PHONY` is used to override the default behaviour of `make` which is to match targets to files.

leohemsted: would it have been running both commands when we call test-with-docker?

sakisv: No, it would just look for a file called `create-docker-test-db` when we called `make create-docker-test-db`, before executing the commands.

PR #1536, opened 2018-01-02
Group letters into a max number of files for sending to DVLA

klssmith:
Grouping the letters into a maximum number of files is necessary because
the SQS task needs to be under a certain size. We also compress the task
when sending.


kentsanggds:
Could we have a test that deals with a mix of max file size and max file counts?

PR #1539, opened 2018-01-03
Update template schema to return the contact block

servingUpAces:
- Add reply_to in the template schema
- Added a boolean column to templates called is_letter_contact_blank. 
If is_letter_contact_blank then the user has set the letter contact block to be blank on purpose
ELSE IF is_letter_contact_blank is false THEN use the template default
IF template default is blank THEN the service_letter_contact is blank use the service default


app/template/rest.py
```diff
@@ -120,7 +120,7 @@ def get_template_by_id_and_service_id(service_id, template_id):
 def preview_template_by_id_and_service_id(service_id, template_id):
     fetched_template = dao_get_template_by_id_and_service_id(template_id=template_id, service_id=service_id)
     data = template_schema.dump(fetched_template).data
-
+    print(data)
```
richardc0: Remove

tests/app/template/test_rest.py
```diff
@@ -623,6 +623,7 @@ def test_get_template_reply_to(client, sample_letter_template):
 
     assert 'service_letter_contact_id' not in json_resp['data']
     assert json_resp['data']['reply_to'] == letter_contact.contact_block
+    assert not json_resp['data']['is_letter_contact_blank']
```
richardc0: Should you add a test to check for handling the error when is_letter_contact_blank = True and service_letter_contact_id is not Null

PR #1543, opened 2018-01-05
Replaces user-provided services with manifest environment variables

allait:
### Add a script to generate PaaS manifest with environment variables

`./scripts/generate_manifest.py` takes a path to a PaaS manifest file and a list of variable files and prints a single CloudFoundry manifest.

The generated manifest replaces all `inherit` keys by loading the data from parent manifests. This allows us to pipe the script output directly to CF CLI, without saving it to disk, which minimises the risk of it being accidentally included in the deployment artefact. The combined manifest might differ from the results produced by CF CLI itself, so the original manifest shouldn't normally be used on its own.

After combining the manifests the script will load and parse all listed variable files and add them to the manifest's `env` by merging the files together in the order they were listed (so in case of any key conflicts the latest file overwrites previous values), upper-casing keys and processing any list or dictionary values with `json.dumps`, so that they can be set as environment variables.

This gives us a full list of environment variables that were previously parsed from the CloudFoundry user services data.

### Remove cloudfoundry config parsing logic

By replacing user-provided services with manifest environment variables we avoid the need to set the application environment variables from the service data.

Most of the variable names already match the service JSON keys, but we need to rename the ones that don't (eg MMG and Firetext `api_key`) this is done in a separate credentials PR.

### Only update existing manifest variables when adding env secrets

Changes generate manifest script to parse variables file as YAML and only add variables to the manifest if they're already listed in the `env` section.

This allows us to use a single variables file for all applications and avoid duplicating secrets across multiple files while adding only the relevant secrets to the application manifest.

### Replace manifest services with lists of env variables

Removes bindings to user provided services for new application deployments and sets the expected environment variables to `null`.

Listing variables in the manifest allows us to only inject secrets used by the app. `null` is not a valid value for a variable in a CloudFoundry manifest, so if any of the keys don't have a match in the credentials file `cf push` will return an error and stop the deployment.

### Use generated manifest when deploying apps to PaaS

Adds a new `make generate-manifest` target which is called by deployment steps.


manifest-api-preview.yml
```diff
@@ -3,14 +3,7 @@
 inherit: manifest-api-base.yml
 
 services:
-  - notify-aws
-  - notify-config
   - notify-db
```
venusbb: notify-db service is already in manifest-notify-base.yml. Do we still need it here?

allait: The lists aren't merged, so whenever we define `services` in a manifest we need to list all services again

PR #1549, opened 2018-01-08
Change for template default address

servingUpAces:
If template has a default address use that address ELSE the address is blank.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -635,8 +635,6 @@ class TemplateBase(db.Model):
         if self.template_type == LETTER_TYPE:
             if self.service_letter_contact_id is not None:
                 return self.service_letter_contact.contact_block
-            else:
-                return self.service.get_default_letter_contact()
         elif self.template_type == EMAIL_TYPE:
             return self.service.get_default_reply_to_email_address()
         elif self.template_type == SMS_TYPE:
--- a/app/models.py
+++ b/app/models.py
@@ -633,8 +633,7 @@ class TemplateBase(db.Model):
 
     def get_reply_to_text(self):
         if self.template_type == LETTER_TYPE:
-            if self.service_letter_contact_id is not None:
-                return self.service_letter_contact.contact_block
+            return self.service_letter_contact.contact_block if self.service_letter_contact else None
         elif self.template_type == EMAIL_TYPE:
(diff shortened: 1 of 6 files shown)


app/models.py
```diff
@@ -635,8 +635,6 @@ def get_reply_to_text(self):
         if self.template_type == LETTER_TYPE:
```
leohemsted: think this'd be a bit cleaner if it were combined in to one line. It's a minor thing, but it'd mean that we'd actually hit the `return None` at the end of the if/elif branches, rather than implicitly returning None.
`if self.template_type == LETTER_TYPE and self.service_letter_contact_id is not None:`

tests/app/template/test_rest.py
```diff
@@ -611,20 +611,33 @@ def test_create_a_template_with_foreign_service_reply_to(admin_request, sample_u
...
                       headers=[auth_header])
 
     assert resp.status_code == 200, resp.get_data(as_text=True)
     json_resp = json.loads(resp.get_data(as_text=True))
 
     assert 'service_letter_contact_id' not in json_resp['data']
-    assert json_resp['data']['reply_to'] == str(letter_contact.id)
-    assert json_resp['data']['reply_to_text'] == letter_contact.contact_block
+    assert json_resp['data']['reply_to'] == reply_to_id
+    assert json_resp['data']['reply_to_text'] == template_default
```
leohemsted: so we're never using the service default? not even when there's no `reply_to` specified, but there is a service_default in the db?

tests/app/template/test_rest.py
```diff
@@ -611,20 +611,33 @@ def test_create_a_template_with_foreign_service_reply_to(admin_request, sample_u
     )
 
 
-def test_get_template_reply_to(client, sample_letter_template):
+@pytest.mark.parametrize('template_default, service_default',
+                         [('template address', 'service address'),
+                          (None, 'service address'),
+                          ('template address', None),
+                          ])
```
leohemsted: do we ever get a case where a service has no letter block at all? (ie: `(None, None)`)

tests/app/template/test_rest.py
servingUpAces: They need to set the address on the template other wise it will be blank.

PR #1554, opened 2018-01-09
Get rate limit from service.rate_limit column (not config)

klssmith:
PR #1550 added the rate_limit column to the Service table.

This PR removes the rate limits from the config and uses rate_limit from
the Service model instead. Rate limits are still separated into 'team',
'normal' and 'test', but these values are the same for a service.

Pivotal story https://www.pivotaltracker.com/story/show/153992529

Paired with @leohemsted

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -250,6 +250,7 @@ class Service(db.Model, Versioned):
         nullable=True,
     )
     crown = db.Column(db.Boolean, index=False, nullable=False, default=True)
+    rate_limit = db.Column(db.Integer, index=False, nullable=False, default=3000)
 
     association_proxy('permissions', 'service_permission_types')
 
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -26,8 +26,8 @@ from app.dao.service_letter_contact_dao import dao_get_letter_contact_by_id
 def check_service_over_api_rate_limit(service, api_key):
     if current_app.config['API_RATE_LIMIT_ENABLED'] and current_app.config['REDIS_ENABLED']:
         cache_key = rate_limit_cache_key(service.id, api_key.key_type)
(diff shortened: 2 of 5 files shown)


app/notifications/validators.py
```diff
@@ -26,8 +26,8 @@
 def check_service_over_api_rate_limit(service, api_key):
     if current_app.config['API_RATE_LIMIT_ENABLED'] and current_app.config['REDIS_ENABLED']:
         cache_key = rate_limit_cache_key(service.id, api_key.key_type)
-        rate_limit = current_app.config['API_KEY_LIMITS'][api_key.key_type]['limit']
-        interval = current_app.config['API_KEY_LIMITS'][api_key.key_type]['interval']
+        rate_limit = service.rate_limit
+        interval = 60
```
richardc0: Should this really be static? It s nice to be able to change it?

app/models.py
```diff
@@ -250,6 +250,7 @@ class Service(db.Model, Versioned):
         nullable=True,
     )
     crown = db.Column(db.Boolean, index=False, nullable=False, default=True)
+    rate_limit = db.Column(db.Integer, index=False, nullable=False, default=3000)
```
richardc0: Does this need a database migration script too?

app/notifications/validators.py
leohemsted: We decided it'd be better not to change it - it would make it confusing if some services are per minute and other services are per second, for example. What's the need to change it per env/service/etc?

i'm open to us documenting it

app/models.py
leohemsted: already merged https://github.com/alphagov/notifications-api/pull/1550

app/notifications/validators.py
servingUpAces: Shall we deploy this and iterate it if need be?

PR #1555, opened 2018-01-09
Drop rate limit database default

klssmith:
The default for the rate_limit column in the services and
services_history model is now set in the model, so we can remove the
default from the database.

Pivotal story: https://www.pivotaltracker.com/story/show/153992529

Paired with @leohemsted


klssmith:
This has been branched off PR #1554, and needs that PR to be merged and deployed first

PR #1556, opened 2018-01-09
Fix the template history when updating the reply_to for a template.

servingUpAces:
It seems selecting the service_letter_contact in the validation method was causing SQLAlchemy to persist the object. When the dao was called to save the object nothing was different so we didn't persist the history object.

It may be time to take another look at how we version. :(

### DIFF
--- a/app/dao/templates_dao.py
+++ b/app/dao/templates_dao.py
@@ -34,6 +34,36 @@ def dao_update_template(template):
 
 
 @transactional
+def dao_update_template_reply_to(template_id, reply_to):
+    Template.query.filter_by(id=template_id).update(
+        {"service_letter_contact_id": reply_to,
+         "updated_at": datetime.utcnow(),
+         "version": Template.version + 1,
+         }
+    )
+    template = Template.query.filter_by(id=template_id).one()
+
+    history = TemplateHistory(**
+                              {
+                                  "id": template.id,
+                                  "name": template.name,
+                                  "template_type": template.template_type,
+                                  "created_at": template.created_at,
+                                  "updated_at": template.updated_at,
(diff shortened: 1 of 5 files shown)


tests/app/dao/test_templates_dao.py
```diff
@@ -100,15 +101,71 @@ def test_update_template_reply_to(sample_service, sample_user):
...
+    assert updated.reply_to == letter_contact_2.id
+    assert updated.version == 2
+    assert updated.updated_at
+
+    updated_history = TemplateHistory.query.filter_by(id=created.id, version=2).one()
+    assert updated_history.service_letter_contact_id == letter_contact_2.id
+    assert updated_history.updated_at == updated_history.updated_at
+
+
+def test_dao_update_tempalte_reply_to_some_to_none(sample_service, sample_user):
```
richardc0: Typo tempalte -> template

app/dao/templates_dao.py
```diff
@@ -33,6 +33,36 @@ def dao_update_template(template):
...
+                                  "service_id": template.service_id,
+                                  "subject": template.subject,
+                                  "created_by_id": template.created_by_id,
+                                  "version": template.version,
+                                  "archived": template.archived,
+                                  "process_type": template.process_type,
+                                  "service_letter_contact_id": template.service_letter_contact_id
+                              })
+    db.session.add(history)
+    return history
```
richardc0: Technically do you want to return template here? Although they are very similar?

I know you use it in the test but its the template being updated?

servingUpAces: I can return either - makes more sense to use template. I'll update

PR #1567, opened 2018-01-12
Raise alert every day to compare ack files with every zip file

venusbb:
Pivotal story: https://www.pivotaltracker.com/story/show/152965020

DVLA sends an acknowledgement file everyday if we send them a zip file. This file contains a list of all the zip files they received for the day. We process this file every day at 23:00 (this time may be adjusted in the future). If there is/are zip files that is not acknowledged, we raise a 500 exception, which contains a list of zip files. 

**Get a list of S3 key name based on a defined suffix**
This function is used to get a list of '.zip' files and '.ack.txt' files from S3 buckets

**Only process letter callback if it is a status update file**
update_letter_notifications_statuses is only called if the filename contains '.rs.txt', which indicates it is a response file, not an acknowledgement file. 

**Define new type of error NoAckFileReceived to be sent to cloudwatch**

**Schedule checking task**
At 23:00 everyday, API checks the received acknowledgement file agains the list of zip files we have. Raise an alert if there are missing zip file acknowledgement

**Build s3 test**
These are unit tests that can be run on the local machine to process s3 buckets and folders.

### DIFF
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -107,3 +107,19 @@ def upload_letters_pdf(reference, crown, filedata):
 
     current_app.logger.info("Uploading letters PDF {} to {}".format(
         upload_file_name, current_app.config['LETTERS_PDF_BUCKET_NAME']))
+
+
+def get_list_of_files_by_suffix(bucket_name, subfolder='', suffix=''):
+    s3_client = client('s3', current_app.config['AWS_REGION'])
+    paginator = s3_client.get_paginator('list_objects_v2')
+
+    page_iterator = paginator.paginate(
+        Bucket=bucket_name,
+        Prefix=subfolder
+    )
+
+    for page in page_iterator:
+        for obj in page['Contents']:
+            key = obj['Key']
+            if key.endswith(suffix):
+                yield key
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -447,3 +447,22 @@ def daily_stats_template_usage_by_month():
                 result.year,
                 result.count
             )
+
+
+@notify_celery.task(name='raise-alert-if-no-letter-ack-file')
+@statsd(namespace="tasks")
+def raise_alert_if_no_letter_ack_file():
+    """
+    Get all files sent "today"
+    list_of_zip_files => get file names s3.get_s3_bucket_objects() look in the folder with todays date
+
+    list_of_ack_files => Get all files sent today with name containing ack.txt from a diff bucket
+
+    For filename in list_of_zip_files:
+        for ack_file in list_of_ack_files if name= file.strip("NOTIFY.", ".ZIP")
+        IF no ack_file
+            raise NoAckFileReceived(status=500, message="No ack file received for {filename}")
+
+    """
+
+    pass
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -59,7 +59,7 @@ from app.celery.tasks import (
 )
 from app.config import QueueNames, TaskNames
 from app.utils import convert_utc_to_bst
-from app.v2.errors import JobIncompleteError
+from app.v2.errors import JobIncompleteError, NoAckFileReceived
 from app.dao.service_callback_api_dao import get_service_callback_api_for_service
 from app.celery.service_callback_tasks import send_delivery_status_to_service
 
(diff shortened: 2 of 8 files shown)


scripts/test_s3.py
```diff
@@ -0,0 +1,66 @@
...
+
+# Tests for boto3 and s3, can only perform locally against the Tools aws account and have permissions to access S3.
+# The tests are based on the above folders and files already uploaded to S3 Tools aws account (If these are removed or
+# renamed, the tests won't pass.
+
+
+def test_get_zip_files():
+    zip_file_list = []
+    for key in get_list_of_files_by_suffix(bucket_name=zip_bucket_name, subfolder=zip_sub_folder, suffix='.ZIP'):
+        print('File: ' + key)
```
servingUpAces: remove print

scripts/test_s3.py
```diff
@@ -0,0 +1,66 @@
...
+        zip_file_list.append(key)
+    assert zip_file_name in zip_file_list
+
+
+def test_get_ack_files():
+    ack_file_list = []
+    for key in get_list_of_files_by_suffix(bucket_name=ack_bucket_name, subfolder=ack_subfolder, suffix='.ACK.txt'):
+        print('File: ' + key)
```
servingUpAces: another print

app/aws/s3.py
```diff
@@ -107,3 +107,19 @@ def upload_letters_pdf(reference, crown, filedata):
...
+
+
+def get_list_of_files_by_suffix(bucket_name, subfolder='', suffix=''):
+    s3_client = client('s3', current_app.config['AWS_REGION'])
+    paginator = s3_client.get_paginator('list_objects_v2')
+
+    page_iterator = paginator.paginate(
+        Bucket=bucket_name,
+        Prefix=subfolder
+    )
```
servingUpAces: We can add a parameter here to only get the files added since "today"

venusbb: There is no filter for returned list by date. The field StartAfter takes in the file name and returns the files after that particular file. 
I have added filtering in the function itself to process the files and only return files that are modified after the date, using the meta data of the files returned from paginate.

app/celery/scheduled_tasks.py
```diff
@@ -452,29 +453,39 @@ def daily_stats_template_usage_by_month():
...
+
     for key in s3.get_list_of_files_by_suffix(bucket_name=current_app.config['LETTERS_PDF_BUCKET_NAME'],
-                                              subfolder=datetime.utcnow().strftime('%Y-%m-%d'), suffix='.ZIP'):
+                                              subfolder=datetime.utcnow().strftime('%Y-%m-%d'),
+                                              suffix='.zip'):
         zip_file_list.append(key)
 
     # get acknowledgement file
     ack_file_list = []
+    yesterday = datetime.now(tz=pytz.utc) - timedelta(days=1)
```
servingUpAces: you could just use datetime.utcnow()

tests/app/aws/test_s3.py
```diff
@@ -173,3 +174,37 @@ def test_upload_letters_pdf_puts_in_tomorrows_bucket_after_half_five(notify_api,
...
+        {
+            "Contents": [
+                single_s3_object_stub('bar/foo1.rs.txt', datetime_in_past(2, 0)),
+            ]
+        }
+    ]
+    paginator_mock.return_value.get_paginator.return_value.paginate.return_value = multiple_pages_s3_object
+    if (days_before):
+        key = get_list_of_files_by_suffix('foo-bucket', subfolder='bar', suffix=suffix_str,
+                                          last_modified=datetime.now(tz=pytz.utc) - timedelta(days=days_before))
```
servingUpAces: use datetime.utc()

app/aws/s3.py
```diff
@@ -107,3 +107,20 @@ def upload_letters_pdf(reference, crown, filedata):
 
     current_app.logger.info("Uploading letters PDF {} to {}".format(
         upload_file_name, current_app.config['LETTERS_PDF_BUCKET_NAME']))
+
+
+def get_list_of_files_by_suffix(bucket_name, subfolder='', suffix='', last_modified=None):
```
vmogoreanu: last_modified parameter name is not clear (before-after-equal?) - I would have called it modified_after

app/aws/s3.py
```diff
@@ -107,3 +107,20 @@ def upload_letters_pdf(reference, crown, filedata):
...
+
+    page_iterator = paginator.paginate(
+        Bucket=bucket_name,
+        Prefix=subfolder
+    )
+
+    for page in page_iterator:
+        for obj in page['Contents']:
+            key = obj['Key'].lower()
+            if key.endswith(suffix.lower()):
```
vmogoreanu: as it suffix - shouldn't it be startswith, not endswith?

tests/app/aws/test_s3.py
venusbb: The datetime returned by aws is comparable with datetime.now(tz=pytz.utc) format, not datetime.utc()

app/celery/scheduled_tasks.py
```diff
@@ -447,3 +448,44 @@ def daily_stats_template_usage_by_month():
...
+                for zf in zip_file_list:
+                    if s[0].lower() in zf.lower():
+                        zip_file_list.remove(zf)
+                    else:
+                        zip_not_today.append(s[0])
+
+    if zip_file_list:
+        raise NoAckFileReceived(message=zip_file_list)
+
+    if zip_not_today:
```
vmogoreanu: this will not be reached if NoAckFileReceived is raised - should it be?

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -1026,3 +1027,46 @@ def test_dao_fetch_monthly_historical_stats_by_template_null_template_id_not_cou
...
 
     assert len(result) == 1
+
+
+def mock_s3_get_list_match(bucket_name, subfolder='', suffix='', lastModified=None):
+
+    if subfolder == '2018-01-11':
+        return ['NOTIFY.20180111175007.ZIP', 'NOTIFY.20180111175008.ZIP']
+    if subfolder == 'root/dispatch':
+        return ['root/dispatch/NOTIFY.20180111175733.ACK.txt']
```
vmogoreanu: I think we should also test multiple ACK files scenario

PR #1578, opened 2018-01-17
Fix a typo error on task argument

venusbb:
A small task to fix a typo error in schedule_task to compare zip files
Modified unit test to be more robust
Related to this PR
https://github.com/alphagov/notifications-api/pull/1567

### DIFF
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -119,8 +119,12 @@ def get_list_of_files_by_suffix(bucket_name, subfolder='', suffix='', last_modif
     )
 
     for page in page_iterator:
-        for obj in page['Contents']:
-            key = obj['Key'].lower()
-            if key.endswith(suffix.lower()):
-                if not last_modified or obj['LastModified'] >= last_modified:
-                    yield key
+        try:
+            for obj in page['Contents']:
+                key = obj['Key'].lower()
+                if key.endswith(suffix.lower()):
+                    if not last_modified or obj['LastModified'] >= last_modified:
+                        yield key
+        except KeyError:
+            # Not content for today
+            pass
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -119,12 +119,8 @@ def get_list_of_files_by_suffix(bucket_name, subfolder='', suffix='', last_modif
     )
 
     for page in page_iterator:
-        try:
-            for obj in page['Contents']:
-                key = obj['Key'].lower()
-                if key.endswith(suffix.lower()):
-                    if not last_modified or obj['LastModified'] >= last_modified:
-                        yield key
-        except KeyError:
-            # Not content for today
-            pass
+        for obj in page.get('Contents', []):
+            key = obj['Key'].lower()
+            if key.endswith(suffix.lower()):
+                if not last_modified or obj['LastModified'] >= last_modified:
+                    yield key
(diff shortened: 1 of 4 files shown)


app/aws/s3.py
```diff
@@ -119,8 +119,12 @@ def get_list_of_files_by_suffix(bucket_name, subfolder='', suffix='', last_modif
     )
 
     for page in page_iterator:
-        for obj in page['Contents']:
-            key = obj['Key'].lower()
-            if key.endswith(suffix.lower()):
-                if not last_modified or obj['LastModified'] >= last_modified:
-                    yield key
+        try:
```
richardc0: Can you query without exceptioning?

Exceptions are for exceptional circumstances this seems like a normal event therefore should be handled by logic?

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -1054,7 +1054,13 @@ def test_letter_not_raise_alert_if_ack_files_match_zip_list(mocker, notify_db):
 
     letter_raise_alert_if_no_ack_file_for_zip()
 
+    yesterday = datetime.utcnow() - timedelta(days=1)
```
richardc0: Tests look good but there are five paths through the task

1. Goof path
2. Exception is thrown
3. page['Contents'] is empty
4. key.endswith(suffix.lower()) is false
5. not last_modified or obj['LastModified'] < last_modified:

Should you add more tests?

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -1054,7 +1054,13 @@ def test_letter_not_raise_alert_if_ack_files_match_zip_list(mocker, notify_db):
...
+    subfoldername = datetime.utcnow().strftime('%Y-%m-%d')
     assert mock_file_list.call_count == 2
+    assert mock_file_list.call_args_list == [
+        call(bucket_name='test-letters-pdf', subfolder=subfoldername, suffix='.zip'),
+        call(bucket_name='test.notify.com-ftp', subfolder='root/dispatch', suffix='.ACK.txt', last_modified=yesterday),
```
kentsanggds: Could use the actual config name to be clearer - `current_app.config['LETTERS_PDF_BUCKET_NAME']` and `current_app.config['DVLA_RESPONSE_BUCKET_NAME']`

app/aws/s3.py
venusbb: Changed to the query Leo suggested.

tests/app/celery/test_scheduled_tasks.py
venusbb: 1, 4, 5 are tested in unit tests test_get_list_of_files_by_suffix in test_s3.py
2 doesn't apply with different query. 
Added an extra test for 3 in test_s3.py

tests/app/celery/test_scheduled_tasks.py
venusbb: Changed

tests/app/aws/test_s3.py
```diff
@@ -208,3 +208,18 @@ def test_get_list_of_files_by_suffix(notify_api, mocker, suffix_str, days_before
...
+        {
+            "other_content": [
+                'some_values',
+            ]
+        }
+    ]
+    paginator_mock.return_value.get_paginator.return_value.paginate.return_value = multiple_pages_s3_object
+    key = get_list_of_files_by_suffix('foo-bucket', subfolder='bar', suffix='.pdf')
+
+    assert sum(1 for x in key) == 0
```
leohemsted: or even just `assert not key`

PR #1579, opened 2018-01-17
Add a celery scheduled task to alert if letters are still sending

allait:
### Add utils DeskproClient and configuration variables

Deskpro client is used to create tickets from celery alerting tasks (eg alerts for missing ack or response files from DVLA).

### Add a celery scheduled task to alert if letters are still sending

We should receive a response file from DVLA by 4pm the next working day (next Monday for letters created on Friday, Saturday or Sunday).

Response file triggers a task to update the letters status from 'sending' to either 'failed' or 'delivered', at which point there should be no letter notifications in the 'sending' state for that day.

To catch any errors in the process (eg a missing response file from DVLA) we add a scheduled task that checks letter notifications for previous day (or Friday when run on Monday) and raises a Deskpro ticket if it finds any in a 'sending' state. The task runs at 4:30pm, which should give the response file processing task enough time to finish if the file was uploaded at 4pm.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -17,7 +17,7 @@ from app.dao.services_dao import (
 )
 from app.dao.stats_template_usage_by_month_dao import insert_or_update_stats_for_template
 from app.performance_platform import total_sent_notifications, processing_time
-from app import performance_platform_client
+from app import performance_platform_client, deskpro_client
 from app.dao.date_util import get_month_start_and_end_date_in_utc
 from app.dao.inbound_sms_dao import delete_inbound_sms_created_more_than_a_week_ago
 from app.dao.invited_user_dao import delete_invitations_created_more_than_two_days_ago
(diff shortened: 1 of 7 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -332,6 +334,38 @@ def delete_dvla_response_files_older_than_seven_days():
         raise
 
 
+@notify_celery.task(name="raise-alert-if-letter-notifications-still-sending")
+@statsd(namespace="tasks")
+def raise_alert_if_letter_notifications_still_sending():
+    today = datetime.utcnow().date()
+
+    # Do nothing on the weekend
+    if today.isoweekday() in [6, 7]:
```
venusbb: typo 'isoweekday'?

app/celery/scheduled_tasks.py
```diff
@@ -332,6 +334,38 @@ def delete_dvla_response_files_older_than_seven_days():
...
+        offset_days = 3
+    else:
+        offset_days = 1
+
+    still_sending = Notification.query.filter(
+        Notification.notification_type == LETTER_TYPE,
+        Notification.status == NOTIFICATION_SENDING,
+        Notification.created_at >= today - timedelta(days=offset_days),
+        Notification.created_at < today
+    ).count()
```
venusbb: How about notifications that are created after 5:40pm (the cut-off time) yesterday?

allait: I've asked @servingUpAces about this and she suggested changing `created_at` to `sent_at` since that's set when the file is sent to DVLA, so it will be the correct date both for letters created on the day and for letters created after 5:30pm the day before. I've updated the task and the tests to do this instead.

PR #1588, opened 2018-01-18
Process letter acknowledge file with new S3 zip file meta file structure

venusbb:
This work is a continuation from 
https://github.com/alphagov/notifications-api/pull/1567  and
https://github.com/alphagov/notifications-api/pull/1578
https://github.com/alphagov/notifications-ftp/pull/42

We are going to stop uploading the daily zip files created for sending to DVLA to S3. It means that we need to persists the zip information somewhere else. 
FTP is going to create a sub-folder inside today's subfolder called zips_sent. In this folder, it is a list of .ZIP.TXT that contains the name of the zip files and meta data of the pdf sent with that zip file. 

This PR changes the processing of DVLA acknowledgement file against this new sub directory. 
- Compare the DVLA acknoledgement file against all the zip files in this sub-directory
- If there is missing zip files in the acknowledgement file, we will raise a 500 error as well as a Deskpro ticket
- If there are zip files in the acknowledgement file that do not exist in today's sub-folder, we print the discrepancy in the log. 
- Added a unit tests for when get_list_of_files_by_suffix returns None

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -487,36 +487,32 @@ def daily_stats_template_usage_by_month():
 @statsd(namespace="tasks")
 def letter_raise_alert_if_no_ack_file_for_zip():
     # get a list of zip files since yesterday
-    zip_file_list = []
+    zip_file_set = set()
 
     for key in s3.get_list_of_files_by_suffix(bucket_name=current_app.config['LETTERS_PDF_BUCKET_NAME'],
                                               subfolder=datetime.utcnow().strftime('%Y-%m-%d') + '/zips_sent',
                                               suffix='.TXT'):
-        zip_file_list.append(key.upper().rstrip('.TXT'))
+        zip_file_set.add(key.upper().rstrip('.TXT'))
 
     # get acknowledgement file
-    ack_file_list = []
+    ack_file_set = set()
     # yesterday = datetime.now(tz=pytz.utc) - timedelta(days=1)
     yesterday = datetime.utcnow() - timedelta(days=1)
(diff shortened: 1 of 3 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -511,12 +511,18 @@ def letter_raise_alert_if_no_ack_file_for_zip():
             for zip_file in content.split('\n'):    # each line
                 s = zip_file.split('|')
                 for zf in zip_file_list:
-                    if s[0].lower() in zf.lower():
+                    if s[0].upper() in zf:
                         zip_file_list.remove(zf)
```
leohemsted: you shouldn't mutate state of a list while traversing it - see https://unspecified.wordpress.com/2009/02/12/thou-shalt-not-modify-a-list-during-iteration/

what's probably better, is to create a `zip_file_set`, and an `ack_file_set`, and then just assert that they are equal

leohemsted:
```
zip_file_set = {
    key.upper().rstrip('.TXT') 
    for key in s3.get_list_of_files_by_suffix(
        bucket_name=current_app.config['LETTERS_PDF_BUCKET_NAME'],
        subfolder=datetime.utcnow().strftime('%Y-%m-%d') + '/zips_sent',
        suffix='.TXT'
    )
}

ack_file_set = set()

for key in ack_file_list....
    ...
          ack_file_set.add(s[0].upper())

if ack_file_set != zip_file_set:
    logger.error('Ack file set: {}, Zip file set: {}'.format(ack_file_set, zip_file_set))
    deskpro_client.create_ticket(message='ack file doesn't match what we sent')```

PR #1594, opened 2018-01-22
Fix grammar in message sent to Deskpro

quis:


venusbb:
New PR changes the wordings of the ticket.

PR #1599, opened 2018-01-23
Store PDF letters in S3 bucket for research mode services

klssmith:
We now upload letters as PDF to S3 when a service is in research mode. These letters are uploaded to a 'research' sub-directory of the S3 bucket we normally use.

We also ensure that letter personalisation is always serialized in the same way.

Pivotal story: https://www.pivotaltracker.com/story/show/151511723

### DIFF
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -81,7 +81,7 @@ def remove_transformed_dvla_file(job_id):
     return obj.delete()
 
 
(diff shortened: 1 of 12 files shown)


kentsanggds:
No longer needed as we decided to cut out creating PDFs and collating notifications for a zip file as part of the functional test.

PR #1600, opened 2018-01-24
Turn letters on by default for new services

quis:
Letters is a mature enough feature now – and one that we’ve been talking about offering for long enough – that we shouldn’t make people dig around in the settings.

I think we’d want to wait a bit longer/indefinitely before deciding to turn it on for existing services across the platform.

---

This takes letters from 4 to 5:

image

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -37,7 +37,8 @@ from app.models import (
     KEY_TYPE_TEST,
     NOTIFICATION_STATUS_TYPES,
     SMS_TYPE,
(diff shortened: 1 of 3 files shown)


servingUpAces:
This is great - but I wonder if we want to do this once we turn on letters_as_pdf and this permission is deprecated.

quis:
@servingUpAces You mean once everyone is moved to `letters_as_pdf`? Think it’s fine to do this before or after (but not at the same time, because merge conflicts 🙀 )

PR #1608, opened 2018-01-26
PR for discussion

servingUpAces:
While trying to add the `download csv` link for the notifications activity page, I found that the rendering of the marshmallow schema was causing performance issues.

Why? Because the schema will make a query for every unique row to all the tables that have a foreign key. For example for every unique template there is a query to template.
Options to fix this problem:
 - update the schema to exclude anything you do not need
 - create a new schema to include only the columns you care about
This seems easy but the problem is also deciding when to join to the table in the query and when to let the schema do it. Another problem is that the get_all_notificaitons_for_services is used on many different pages in the admin app. Each has a different data set.

Approach:
 - create a unique endpoint and schema for each endpoint.
 - ensure the query is joining only to the tables that it needs to.

I have opted for creating the new schema in json schema for two reasons:
 - it doesn't use a framework to do the serialisation
 - can easily control what columns to include, meaning there is higher visibility on what extra queries are being made.
 - we are only returning the data that the calling app needs

Complications:
 - this means we need a new endpoint for each call from the admin app, requiring some refactoring in the admin.
 - leads to more code - however, it should be clearer.

The dao function could probably be simplified, I have not really changed that.

There are more tests to write.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -355,6 +355,7 @@ def get_all_notifications_for_service_csv(service_id):
     limit_days = data.get('limit_days')
     include_jobs = data.get('include_jobs', True)
     include_from_test_key = data.get('include_from_test_key', False)
+    include_created_by_user = data.get('include_created_by_user', False)
 
     pagination = notifications_dao.get_notifications_for_service(
         service_id,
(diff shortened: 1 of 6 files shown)


app/service/rest.py
```diff
@@ -344,6 +345,45 @@ def get_all_notifications_for_service(service_id):
     ), 200
 
 
+@service_blueprint.route('/<uuid:service_id>/notifications/csv', methods=['GET'])
+def get_all_notifications_for_service_csv(service_id):
```
servingUpAces: Obviously this needs to be refactored so that all the code is reused rather than copied.

PR #1609, opened 2018-01-26
Refactor letter research processing for letter pdfs

kentsanggds:
## What 

In order to run letter functional tests we need to improve the way that API handles letters in research mode. I've also had to fix some bugs that I picked up during the functional tests.

- Fixed the way that the personalisation for letters is serialized so that it can serialize regardless of how it is stored
- Refactored the code so that the business logic resides in the letter task handling rather than s3 handling
- fixed ACK file handling as it was reporting random errors due to random order of sets
- added a create fake letter response file task in research mode tasks which will also trigger the dvla callback in development environment
- updated handling of letter jobs for letter pdfs to call the create response file task in preview and development environments
- updated handling of letter apis for letter pdfs to call the create response file task in preview and development environments

## How to test locally

- set a service to research mode
- Jobs 
  - create a letter template
  - set the service as sending letters as pdf
  - upload a recipients csv for letters

- API
  - create a test API key from that service and use in 
  - using the python client repo

```
PYTHONPATH=. python ./utils/make_api_call.py http://localhost:6011 research-API-TEST-KEY create --type=letter --template=[letter template id] --personalisation="{\"address_line_1\":\"someone\", \"address_line_2\":\"London\", \"postcode\":\"N1\"}"
```

- logs should have something like `DVLA file: NOTIFY.20180126142426.RSP.TXT, notification updated to delivered`

## Reference 

https://www.pivotaltracker.com/story/show/154597494

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -6,6 +6,7 @@ from notifications_utils.recipients import try_validate_and_format_phone_number
 
 from app import api_user, authenticated_service
 from app.config import QueueNames
+from app.dao.notifications_dao import update_notification_status_by_reference
 from app.models import (
     SMS_TYPE,
     EMAIL_TYPE,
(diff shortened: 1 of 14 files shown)


app/v2/notifications/post_notifications.py
```diff
@@ -191,10 +194,19 @@ def process_letter_notification(*, letter_data, api_key, template, reply_to_text
...
-            [str(notification.id)],
-            queue=QueueNames.CREATE_LETTERS_PDF
-        )
+        if should_send:
+            create_letters_pdf.apply_async(
+                [str(notification.id)],
+                queue=QueueNames.CREATE_LETTERS_PDF
+            )
+        elif api_key.service.research_mode and \
+                current_app.config['NOTIFY_ENVIRONMENT'] in ['preview', 'development']:
```
leohemsted: can we use `()` instead of `\` continuation pls

PR #1613, opened 2018-01-29
fix 500 when receiving delivery receipt from some international nums

leohemsted:
if the international_billing_rates.yml has `dlr: null`, that means we don't know what delivery receipts they provide - they might not provide any. So if we do get an update, we don't know for sure that the message was actually delivered - lets not update it.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -134,7 +134,8 @@ def _decide_permanent_temporary_failure(current_status, status):
 
 
 def country_records_delivery(phone_prefix):
-    return INTERNATIONAL_BILLING_RATES[phone_prefix]['attributes']['dlr'].lower() == 'yes'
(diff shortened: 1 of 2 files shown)


leohemsted:
@richardc0 this was a deliberate choice - they're not in "sending" before this, they're in "sent" (as in "sent internationally" - because we can't be 100% sure what delivery receipts from sudan mean.

For example, SMS sent to USA we get "Carrier DLR" for - which means that the phone provider returns a delivery receipt to us acknowledging that they own that number, rather than "yes this sms was sent to a real person's phone".

We got a 500 because we didn't expect to receive any delivery receipt from sudan - I've no idea on the quality of sudanese delivery receipts, and I don't want to say that one sudanese SMS was delivered successfully, if the next sudanese SMS we might not get anything for. And then it might turn out that neither of them were delivered.

PR #1624, opened 2018-02-05
Email branding endpoints

leohemsted:
- [x] https://github.com/alphagov/notifications-api/pull/1621

This is the big bang PR that changes us to using email branding.

To recap, after #1621 is merged, we will be adding to the mapping table and also updating organisation on the service - but not using the new mapping table.

This PR:

* changes all references of organisation (except those mentioned below) to email_branding.
* makes the send email function read from service.email_branding, rather than from service.organisation

Exceptions/notes:

* passing in either `email-branding` OR `organisation` to update_service will still update email branding
* both `/email-branding` and `/organisation` hit the same code
* service endpoints still return organisation as well as email branding in the json


app/__init__.py
```diff
@@ -159,8 +159,10 @@ def register_blueprint(application):
     provider_details_blueprint.before_request(requires_admin_auth)
     application.register_blueprint(provider_details_blueprint, url_prefix='/provider-details')
 
-    organisation_blueprint.before_request(requires_admin_auth)
-    application.register_blueprint(organisation_blueprint, url_prefix='/organisation')
+    email_branding_blueprint.before_request(requires_admin_auth)
+    application.register_blueprint(email_branding_blueprint, url_prefix='/email_branding')
```
gov-cjwaszczuk: Should be email-branding

PR #1627, opened 2018-02-06
Use eventlet workers for gunicorn

sakisv:
Cap their parallel connections to 1/4 of the default to play it safe
while we keep an eye on our resources utilisation


manifest-api-base.yml
```diff
@@ -1,7 +1,7 @@
 ---
 
 buildpack: python_buildpack
-command: scripts/run_app_paas.sh gunicorn -c /home/vcap/app/gunicorn_config.py --error-logfile /home/vcap/logs/gunicorn_error.log -w 5 -b 0.0.0.0:$PORT application
+command: scripts/run_app_paas.sh gunicorn -c /home/vcap/app/gunicorn_config.py -k eventlet --worker-connections 256 --error-logfile /home/vcap/logs/gunicorn_error.log -w 5 -b 0.0.0.0:$PORT application
```
leohemsted: minor things - can you split cmd out into multiple lines like we've done in admin

https://github.com/alphagov/notifications-admin/blob/master/manifest-base.yml

sakisv: I'd prefer if we moved all the config args to `gunicorn_config.py` so that we have only one place that gunicorn is configured.

leohemsted: im fine with that if you want to

sakisv: I want. And I did 😄

PR #1635, opened 2018-02-08
Remove failed as a possible status

servingUpAces:
We are not handling the case of an unknown status code sent by the SMS provider. This PR attempts to fix that. `failed` is a deprecated failure type.

- If the SMS client sends a status code that we do not recognize raise a ClientException and set the notification status to technical-failure
- Simplified the code in process_client_response, using a simple map.

### DIFF
--- a/app/notifications/process_client_response.py
+++ b/app/notifications/process_client_response.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from flask import current_app
 
 from app import statsd_client
+from app.clients import ClientException
 from app.dao import notifications_dao
 from app.clients.sms.firetext import get_firetext_responses
 from app.clients.sms.mmg import get_mmg_responses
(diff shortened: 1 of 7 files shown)


app/notifications/process_client_response.py
```diff
@@ -49,32 +50,28 @@ def process_sms_client_response(status, reference, client_name):
...
-        msg = "{} callback failed: status {} not found.".format(client_name, status)
-        return success, msg
+        process_for_status(notification_status='technical-failure', client_name=client_name, reference=reference)
+        raise ClientException("{} callback failed: status {} not found.".format(client_name, status))
 
-    notification_status = response_dict['notification_status']
-    notification_status_message = response_dict['message']
-    notification_success = response_dict['success']
+    success = process_for_status(notification_status=notification_status, client_name=client_name, reference=reference)
+    return success, errors
```
venusbb: There would never be any errors to return.

venusbb:
There are no tests on process_firetext_response and process_mmg_response

venusbb:
If there are errors, we return 400. Just want to check if this is intended. 
```
    if errors:
```

PR #1639, opened 2018-02-08
Create new organisation model

klssmith:
Added the new Organisation model.

Also added in the following endpoints for notifications-admin to use:
* Get all organisations
* Get an organisation by id
* Create an organisation
* Update an organisation by id

[Pivotal story](https://www.pivotaltracker.com/story/show/153989542)


app/dao/organisation_dao.py
```diff
@@ -0,0 +1,25 @@
...
+    return Organisation.query.filter_by(id=organisation_id).one()
+
+
+@transactional
+def dao_create_organisation(organisation):
+    db.session.add(organisation)
+
+
+@transactional
+def dao_update_organisation(organisation, **kwargs):
```
servingUpAces: I would prefer if we had a function to update_organisation_name(organisation_id, update_dict) then do an update statement.

app/organisation/rest.py
```diff
@@ -0,0 +1,57 @@
...
+    return jsonify(organisation.serialize()), 201
+
+
+@organisation_blueprint.route('/<uuid:organisation_id>', methods=['POST'])
+def update_organisation(organisation_id):
+    data = request.get_json()
+
+    validate(data, post_update_organisation_schema)
+
+    fetched_organisation = dao_get_organisation_by_id(organisation_id)
```
servingUpAces: If we change the update method as I suggest we don't need to do this query, the filter happens on the update query.

klssmith: @servingUpAces I've now made the changes to this file and the DAO in the latest commit

PR #1640, opened 2018-02-08
Fixed typo in log call so that the name of the task is returned

richardc0:
Fixed typo in call, should be self.name not Task.name.

Task.name always returned None.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -60,8 +60,6 @@ class TaskNames(object):
 
 
 class Config(object):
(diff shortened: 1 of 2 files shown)


app/config.py
```diff
@@ -60,6 +60,8 @@ class TaskNames(object):
 
 
 class Config(object):
+    NOTIFY_LOG_LEVEL = 'DEBUG'
```
leohemsted: 🖨 🔥

PR #1649, opened 2018-02-12
Link service to organisation

kentsanggds:
## What - 

Link a service to an organisation, when calling the endpoint and the service already has an organisation it will remove the existing relationship to the organisation and add the new one in as there can only be 1 organisation per service.

- additionally added in an endpoint to get all services for an organisation which will be used in another story - https://www.pivotaltracker.com/story/show/154768116

## How to review - end to end

Use this fork for admin - https://github.com/gov-cjwaszczuk/notifications-admin/tree/link-service-to-organisation

Log in and create an organisation, then select a service and assign it to the organisation

It should also be possible to change the organisation if more than 1 is available

## Reference 

https://www.pivotaltracker.com/story/show/153989801


kentsanggds:
Sure, updating to throw 400s instead and added tests to check that this is the case

PR #1651, opened 2018-02-12
Switch service callback workers to use eventlet pool implementation

allait:
### Create a single app context for each celery task run

Celery tasks require an active app context in order to use app logger, database or app configuration values. Since there's no built-in support we create this context by using a custom celery task class NotifyTask.

However, NotifyTask was using `current_app`, which itself is only available within an app context so the code was pushing an initial context in run_celery.py.

This works with the default prefork pool implementation, but raises a "working outside of application context" error with an eventlet pool since that initial application context is local to a thread and eventlet celery worker pool will create multiple green threads.

To avoid this, we bind NotifyTask to the app variable with a closure and use that variable instead of `current_app` to create the context for executing the task. This means we don't need to push an additional context when starting up celery workers and avoids any issues caused by shared initial app context being lost when spawning additional worker threads.

### Release DB connection before executing service API callback

Flask-SQLAlchemy sets up a connection pool with 5 connections and will create up to 10 additional connections if all the pool ones are in use.

If all the connections in the pool and all overflow connections are in use SQLAlchemy will block new DB sessions for the duration of the pool timeout (which we set to 30s). If a new connection becomes available during that time it will be used for the blocked session, otherwise a TimeoutError is raised.

By default db.session is deleted with the related context object (so when the request is finished or app context is discarded).

This effectively limits the number of concurrent requests/tasks with multithreaded gunicorn/celery workers to the maximum DB connection pool size. Most of the time these limits are fine since the API requests are relatively quick and are mainly interacting with the database anyway.

Service callbacks however have to make an HTTP request to a third party. If these requests start taking a long time and the number of threads is larger than the number of DB connections then remaining threads will start blocking and potentially failing if it takes more than 30s to acquire a connection. For example if a 100 threads start running tasks that take 20s each with a max DB connection pool size of 10 then first 10 threads will acquire a connection right away, next 10 tasks will block for 20 seconds before the initial connections are released and all other tasks will raise a TimeoutError after 30 seconds.

To avoid this, we perform all database operations at the beginning of the task and then explicitly close the DB session before sending the HTTP request to the service callback URL. Closing the session ends the transaction and frees up the connection, making it available for other tasks. Making calls to the DB after calling `close` will acquire a new connection. This means that tasks are still limited to running at most 15 queries at the same time, but can have a lot more concurrent HTTP requests in progress.

### Switch service callback workers to use eventlet pool implementation

Service callbacks are I/O bound and can take a long time if the callback URL takes a long time to respond. This is a perfect use case for an eventlet worker pool since it allows spawning multiple green threads (1000 in the proposed configuration) to execute HTTP requests concurrently without a significant increase in CPU load or memory usage needed for adding additional worker processes.


leohemsted:
reverted eventlets in https://github.com/alphagov/notifications-api/pull/1754 😬

PR #1656, opened 2018-02-13
Allow services to add punctuation to or change the case of their name

klssmith:
Changed the `/service/unique` endpoint to optionally accept the
`service_id` parameter. It now doesn't matter if a user tries to change
the capitalization or add punctuation to their own service name. But
there should still be an error if a user tries to change the punctuation
or capitalization of another service.

`service_id` needs to be allowed to be `None` until notifications-admin is
updated to always pass it in.

[Pivotal story](https://www.pivotaltracker.com/story/show/155127026)


klssmith:
@kentsanggds 👍 Now updated the test names

PR #1661, opened 2018-02-14
Automatically set environment vars before tests

quis:
Sometimes you just wanna run some tests directly using the `pytest` command. But you’re in a new shell, and have forgotten to do `source environment_test.sh`. The screen fills with red, and your day just got a little bit worse.

This commit will stop this from ever happening again, by making the setting environment variables part of running Pytest. It does this with a plugin called pytest-env<sup>1</sup>.

pytest.ini is the standard way of configuring pytest. Creating this file where it didn’t exist before changes the behaviour of pytest, in that it will now look for tests in the same directory as the file, rather than defaulting to the `tests/` directory. So we also have to explicitly configure pytest<sup>2</sup> to tell it that it should only look in this directory. Otherwise it gets lost in the weeds of `node_modules`.

This also removes the ability to override the value of `SQLALCHEMY_DATABASE_URI` because:
- it’s not possible with this plugin
- the two different values are not needed (the username and password can be provided in all contexts)

---

1. https://github.com/MobileDynasty/pytest-env
2. https://docs.pytest.org/en/latest/customize.html#confval-testpaths


pytest.ini
```diff
@@ -0,0 +1,20 @@
+[pytest]
+testpaths = tests
+env =
+    SQLALCHEMY_DATABASE_URI=postgresql://postgres:postgres@localhost/test_notification_api
```
leohemsted: this is `localhost` locally, but on jenkins it is `postgres`

quis: Hopefully this is fixed with `D:SQLALCHEMY_DATABASE_URI=postgresql://@localhost/test_notification_api`, which only sets the variable if it’s not already in the environment.

PR #1663, opened 2018-02-14
Stop pytest on 10th failing test

quis:
If a PR is going to fail because tests aren’t passing then you:
- should know about it as quick as possible
- shouldn’t waste precious Jenkins CPU running subsequent tests

This commit adds the `-maxfail=10` flag to pytest, which stops the test run once an unacceptable number of tests have failed.

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -31,5 +31,5 @@ flake8 .
 display_result $? 1 "Code style check"
 
 # run with four concurrent threads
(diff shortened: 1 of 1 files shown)


leohemsted:
not sure i agree with this. lemme think it over during lunch

quis:
@leohemsted as discussed:

# faa7675 Allow up to 10 tests to fail before stopping 

This is a sensible compromise between 1 test and ALL THE TESTS.

Uses the `maxfail` option as documented here:
https://docs.pytest.org/en/latest/usage.html#stopping-after-the-first-or-n-failures

PR #1665, opened 2018-02-14
Reduce logging level

servingUpAces:
If the organisation name that is being inserted or updated is not unique we just want to return a 400 to the admin app.
Updated the code so that we are not logging.exception, this is because a cloud watch alert is set to the support team. This type of error is not something we need to investigate.


app/organisation/rest.py
```diff
@@ -29,11 +29,7 @@ def handle_integrity_error(exc):
     Handle integrity errors caused by the unique contraint on ix_organisation_name
     """
     if 'ix_organisation_name' in str(exc):
-        current_app.logger.exception('Unique constraint ix_organisation_name triggered')
-        return jsonify(
-            result='error',
-            message='Organisation name already exists'
-        ), 400
+        raise InvalidRequest('Organisation name already exists', 400)
```
kentsanggds: We can't raise error within error handlers, @leohemsted suggests that we do something like this - 

```
    if 'ix_organisation_name' in str(exc):
        return jsonify(
            result='error',
            message='Organisation name already exists'
        ), 400

    current_app.logger.exception(exc)
    return jsonify(result='error', message="Internal server error"), 500
```

Also in `dao_utils.py` it's logging errors in the `transactional` decorator, we might want to ignore the `IntegrityError` there too

PR #1670, opened 2018-02-15
Improve error message for wrong template ID

quis:
If you try to send a notification from a template which doesn’t exist, or there’s a typo in the ID then the error you get isn’t useful enough that you can figure out what to do to fix it.

This commit makes the error message more helpful.


leohemsted:
think you need to be modifying `app/v2/errors.py`

PR #1681, opened 2018-02-19
make sure tests always run in test db

leohemsted:
previously, if you had no db set it would work, but if you had the live db set in the env as SQLALCHEMY_DATABASE_URI then it'd try and use that, and error out.

no-one runs on weird setups so we can just manually set it if it's not docker.


leohemsted:
@allait this look bon?

PR #1685, opened 2018-02-20
Reporting: send database update tasks to report queues

venusbb:
This is test codes for the reporting design as documented here: 
https://docs.google.com/document/d/1ZGOLXfdZNapnHr-r33litZrOt-belNkvxm4KHASSyis/edit?usp=sharing

In this PR, we send updating tasks with data format as json to a report queue when there is an update for services, templates or notifications, similar to how a trigger works. These tasks are picked up by the new report apps, that will update the new reporting database. 

We only want to test it on the Preview and local Development environment at the moment.


quis:
The detail in the ADR is really nice. But it’s only available to people who can login to Google Docs. 

It would be really nice if we could get some of the rationale for a change into our commit messages, because:

> While a link can add some extra context for people reviewing a pull-request, the commit message should stand on its own. There's no guarantee that the link will continue to work in the future when someone is looking through the commit history to understand why a change was made.

_https://github.com/alphagov/styleguides/blob/bd959db34a27bfd8fd6b6f80f1b385638f6306c3/git.md#commit-messages_

app/dao/notifications_dao.py
```diff
@@ -116,6 +116,9 @@ def dao_create_notification(notification):
     if _should_record_notification_in_history_table(notification):
         db.session.add(NotificationHistory.from_original(notification))
 
+    if current_app.config['SEND_REPORTS']:
```
richardc0: Not sure the dao is the best place to put adding messages to the queue, as they are usually reserved for database actions only.

Is there a business object / method of notification which would be more suitable i.e. the method that calls this?

app/dao/notifications_dao.py
```diff
@@ -199,6 +202,9 @@ def dao_update_notification(notification):
         db.session.add(notification_history)
     db.session.commit()
 
+    if current_app.config['SEND_REPORTS']:
```
richardc0: As with the other dao comment

app/dao/services_dao.py
```diff
@@ -180,12 +181,18 @@ def dao_create_service(service, user, service_id=None, service_permissions=None)
     insert_service_sms_sender(service, current_app.config['FROM_NUMBER'])
     db.session.add(service)
 
+    if current_app.config['SEND_REPORTS']:
```
richardc0: As above

app/dao/templates_dao.py
```diff
@@ -26,12 +28,19 @@ def dao_create_template(template):
 
     db.session.add(template)
 
+    # Send data to Reports queue to update Report DB
+    if current_app.config['SEND_REPORTS']:
```
richardc0: As above

app/models.py
```diff
@@ -775,6 +788,20 @@ def get_link(self):
             _external=True
         )
 
+    def serialize_for_reports(self):
+        serialized = {
```
richardc0: Is there any reason why this does not follow the format above? 

For consistently use the same style i.e. return { your dict } or return serilized

app/reports/report_tasks.py
```diff
@@ -0,0 +1,29 @@
+from app import notify_celery
```
richardc0: Might be good to have a report worker in future?

app/reports/report_tasks.py
```diff
@@ -0,0 +1,29 @@
...
+from app.config import QueueNames
```
richardc0: Will these tasks run on all preview, staging and prod?

Any way you can restrict them so they use your config item.

servingUpAces:
Closing this PR. We are going to try a different tactic

PR #1691, opened 2018-02-21
send letters for services without letters_for_pdf

leohemsted:
also make sure that the service isn't in research mode

### DIFF
--- a/tests/app/dao/notification_dao/test_notification_dao.py
+++ b/tests/app/dao/notification_dao/test_notification_dao.py
@@ -35,7 +35,6 @@ from app.dao.notifications_dao import (
     dao_get_notifications_by_references
 )
 from app.dao.services_dao import dao_update_service
-from app.dao.service_permissions_dao import dao_add_service_permission
 from app.models import (
     Job,
     Notification,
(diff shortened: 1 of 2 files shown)


leohemsted:
we'll need to manually replay the letters from the 19th

tests/app/dao/notification_dao/test_notification_dao.py
```diff
@@ -2093,28 +2088,23 @@ def test_dao_get_count_of_letters_to_process_for_today_without_notis_does_not_ra
 
 
 @freeze_time("2017-12-18 17:50")
-def test_dao_get_count_of_letters_to_process_for_date_ignores_service_not_letters_as_pdf(
-        sample_letter_template, sample_template):
-    # not expected
-    create_notification(template=sample_letter_template, created_at='2017-12-18 17:29:00')
+def test_dao_get_count_of_letters_to_process_for_date_ignores_service_not_letters_as_pdf(sample_letter_template):
```
richardc0: Is this test name still relevant or should it renamed for a test for research mode or should there be two different tests?

leohemsted: oops i thought i renamed that test 😬yeah it should just be renamed. letters_as_pdf is no longer relevant at all

PR #1692, opened 2018-02-21
don't return or accept sms_sender in update service anymore

leohemsted:
it's no longer a field on the service

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -205,21 +205,17 @@ class ServiceSchema(BaseSchema):
     email_branding = field_for(models.Service, 'email_branding')
     organisation = field_for(models.Service, 'organisation')
     override_flag = False
-    reply_to_email_address = fields.Method(method_name="get_reply_to_email_address")
     letter_contact_block = fields.Method(method_name="get_letter_contact")
 
     def service_permissions(self, service):
         return [p.permission for p in service.permissions]
-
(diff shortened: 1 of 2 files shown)


venusbb:
Delete here in admin and check if there are other references
https://github.com/alphagov/notifications-admin/blob/master/app/notify_client/service_api_client.py#L99

app/schemas.py
```diff
@@ -216,9 +214,6 @@ def service_permissions(self, service):
     def get_reply_to_email_address(self, service):
```
servingUpAces: Does the same apply to reply_to_email?

app/schemas.py
```diff
@@ -216,9 +214,6 @@ def service_permissions(self, service):
...
         return service.get_default_reply_to_email_address()
 
-    def get_sms_sender(self, service):
-        return service.get_default_sms_sender()
-
     def get_letter_contact(self, service):
```
servingUpAces: Does the same apply to letter_contact?

leohemsted: used in admin here - https://github.com/alphagov/notifications-admin/blob/master/app/main/views/service_settings.py#L663

app/schemas.py
leohemsted: yup you're right - removed

PR #1693, opened 2018-02-21
Increase DB connection pool size to 10

sakisv:
This is to address some errors we saw yesterday such as:

`sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10
reached, connection timed out, timeout 30`

Related flask-sqlalchemy docs:
http://flask-sqlalchemy.pocoo.org/2.3/config/#configuration-keys


app/config.py
```diff
@@ -127,6 +127,7 @@ class Config(object):
     SQLALCHEMY_COMMIT_ON_TEARDOWN = False
     SQLALCHEMY_RECORD_QUERIES = True
     SQLALCHEMY_TRACK_MODIFICATIONS = True
+    SQLALCHEMY_POOL_SIZE = os.environ.get('SQLALCHEMY_POOL_SIZE', 5)
```
richardc0: I think we should set SQLALCHEMY_MAX_OVERFLOW as well?

If the overflow is 10 and the pool size if 10 we are technically no better off? Its just more efficient as the overflow connections don't stay open whereas the pool connection will.

Maybe set it to SQLALCHEMY_MAX_OVERFLOW = 15?

sakisv: Why are we not better off with the default overflow of 10? 

Before we would go to 15 connections per worker per instance (5 permanent + 10 temporary) and now we will go to 20 connections per worker per instance (10 permanent + 10 temporary).

We have double the amount of permanent connections and we keep the same headroom for extra connections if necessary. Am I missing something?

PR #1695, opened 2018-02-21
Organisation user endpoints

servingUpAces:
New endpoints to support inviting and adding a user to an organisation.

These endpoints support the admin app to invite someone to become a team member of an organisation. That person will receive an email to inviting them to click a link to accept the invitation to the the organisation. They will be prompted to login or register with Notify. 

From the organisation dashboard you can see who is a team member. 

https://www.pivotaltracker.com/story/show/153990168


migrations/versions/0167_add_org_invite_template.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+from datetime import datetime
+
+from alembic import op
+from flask import current_app
+
+
+revision = '0167_add_org_invite_template'
+down_revision = '0166_add_org_user_stuff'
+
```
venusbb: Need to change this to 0168

app/organisation/accept_organisation_invite.py
```diff
@@ -0,0 +1,30 @@
...
+from app.errors import register_errors, InvalidRequest
+
+accept_organisation_invite_blueprint = Blueprint(
+    'accept_organisation_invite', __name__,
+    url_prefix='/organisation-invitation')
+
+register_errors(accept_organisation_invite_blueprint)
+
+
+@accept_organisation_invite_blueprint.route("/<token>", methods=['GET'])
```
venusbb: We can combine the two accept invite blueprint with the "any" parameter
@accept_invite_blueprint.route("/<any:accept_invite, accept_organisation_invite: level>/<token>", methods=['GET'])
def accept_invitation(level, token):

then "level" will be available to be used in the function.

app/dao/invited_org_user_dao.py
```diff
@@ -0,0 +1,29 @@
...
+
+def get_invited_org_users_for_organisation(organisation_id):
+    return InvitedOrganisationUser.query.filter_by(organisation_id=organisation_id).all()
+
+
+def delete_org_invitations_created_more_than_two_days_ago():
+    deleted = db.session.query(InvitedOrganisationUser).filter(
+        InvitedOrganisationUser.created_at <= datetime.utcnow() - timedelta(days=2)
+    ).delete()
+    db.session.commit()
```
venusbb: The time is not hard-coded in accept invite, why here?
```def accept_organisation_invitation(token):
+    max_age_seconds = 60 * 60 * 24 * current_app.config['INVITATION_EXPIRATION_DAYS']```

File: app/organisation/accept_organisation_invite.py (hunk 2 above)
Comment (venusbb) [reply app/organisation/accept_organisation_invite.py]: Similar applies to the invite rests.

File: app/dao/organisation_dao.py [hunk 4]
```diff
@@ -42,3 +46,23 @@ def dao_add_service_to_organisation(service, organisation_id):
...
     organisation.services.append(service)
+
+
+def dao_get_invited_organisation_user(user_id):
+    return InvitedOrganisationUser.query.filter_by(id=user_id).one()
+
+
+def dao_get_users_for_organisation(organisation_id):
+    return User.query.filter(
+        User.user_to_organisation.any(id=organisation_id),
```
leohemsted: (this becomes `User.organisations.any(id=organisation_id)` if the backref changes)

app/organisation/accept_organisation_invite.py
venusbb: You will end up with maintaining two flows, which pretty much does the same thing.

app/models.py
```diff
@@ -1451,6 +1449,16 @@ class InvitedOrganisationUser(db.Model):
...
 
+    def serialize(self):
+        return {
+            'id': str(self.id),
+            'email_address': self.email_address,
+            'invited_by': str(self.invited_by_id),
+            'organisation': str(self.organisation_id),
+            'created_at': self.created_at.strftime(DATETIME_FORMAT),
+            'status': self.status
+        }
```
venusbb: do we need 'organisation_id' in here?

app/dao/invited_org_user_dao.py
servingUpAces: It shouldn't be

tests/app/invite/test_invite_rest.py
```diff
@@ -161,40 +160,6 @@ def test_get_invited_users_by_service_with_no_invites(client, notify_db, notify_
...
-
-    auth_header = create_authorization_header()
-
-    response = client.get(
-        url,
-        headers=[('Content-Type', 'application/json'), auth_header]
-    )
-    assert response.status_code == 404
-
-
```
venusbb: These two tests are removed because...?

leohemsted: the unused endpoint they tested was removed

app/models.py
servingUpAces: yes

app/dao/invited_org_user_dao.py
servingUpAces: I think it's ok - the method is called delete_org_invitations_created_more_than_two_days_ago.
It saves getting the config from current_app.

PR #1698, opened 2018-02-22
Added a new exception type for DVLAException

servingUpAces:
The Notify team needs to investigate when a notification is marked as failed.
We will process the whole file and mark the notifications with the appropriate status, if any are failed an exception is raised.
The exception will trigger a cloud watch error for the team to investigate.

### DIFF
--- a/tests/app/celery/test_ftp_update_tasks.py
+++ b/tests/app/celery/test_ftp_update_tasks.py
@@ -5,6 +5,7 @@ import pytest
 from freezegun import freeze_time
 from flask import current_app
 
+from app.exceptions import DVLAException
 from app.models import (
     Job,
     Notification,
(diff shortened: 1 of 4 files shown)


tests/app/celery/test_ftp_update_tasks.py
```diff
@@ -65,8 +66,23 @@ def test_update_letter_notifications_statuses_raises_for_invalid_format(notify_a
     invalid_file = 'ref-foo|Sent|1|Unsorted\nref-bar|Sent|2'
     mocker.patch('app.celery.tasks.s3.get_s3_file', return_value=invalid_file)
 
-    with pytest.raises(TypeError):
+    with pytest.raises(DVLAException) as e:
```
leohemsted: 👌

PR #1703, opened 2018-02-23
Post pre-compiled letter PDF notification endpoint

allait:
### Add a hidden column to templates

Allows hiding templates from the templates list in the admin app and related API responses.

This is used for 'internal' templates that we create for notifications that wouldn't have a template otherwise (eg pre-compiled PDF letters)

### Add a DB migration to create Templates.hidden column

Creates the column as nullable, sets the value to false for all existing templates and template versions and then applies a not-nullable constraint.

All future Templates are created with `False` as the default set in SQLAlchemy.

### Add a view function for pre-compiled PDF letters

Adds a separate view function that is registered under the same route as existing letter POST notification.

### Upload pre-compiled letter PDF to S3

Pre-compiled letter endpoint uploads PDF contents to S3 directly instead of creating a letter task to generate PDF using template preview.

This moves some of the utility functions used by existing letter celery tasks to app.letters.utils, so that they can be reused by the API endpoint.

### Don't return hidden templates in API service template responses

Removes hidden templates from the service templates list and returns 404 when searching for a hidden template or template versions by ID.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -224,6 +224,12 @@ def process_letter_notification(*, letter_data, api_key, template, reply_to_text
     if not api_key.service.research_mode and api_key.service.restricted and api_key.key_type != KEY_TYPE_TEST:
         raise BadRequestError(message='Cannot send letters when service is in trial mode', status_code=403)
 
+    if precompiled:
+        try:
+            letter_content = base64.b64decode(letter_data['content'])
+        except ValueError:
+            raise BadRequestError(message='Cannot decode letter content (invalid base64 encoding)', status_code=400)
+
     should_send = not (api_key.service.research_mode or api_key.key_type == KEY_TYPE_TEST)
 
     # if we don't want to actually send the letter, then start it off in SENDING so we don't pick it up
@@ -236,7 +242,7 @@ def process_letter_notification(*, letter_data, api_key, template, reply_to_text
 
     if should_send:
         if precompiled:
-            upload_letter_pdf(notification, base64.b64decode(letter_data['content']))
+            upload_letter_pdf(notification, letter_content)
         else:
             create_letters_pdf.apply_async(
                 [str(notification.id)],
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -1,3 +1,4 @@
+import base64
 import functools
 
 from flask import request, jsonify, current_app, abort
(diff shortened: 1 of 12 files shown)


migrations/versions/0168_hidden_templates.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def upgrade():
+    op.add_column('templates', sa.Column('hidden', sa.Boolean(), nullable=True))
+    op.add_column('templates_history', sa.Column('hidden', sa.Boolean(), nullable=True))
+
+    op.execute('UPDATE templates SET hidden=false')
+    op.execute('UPDATE templates_history SET hidden=false')
+
+    op.alter_column('templates', 'hidden', nullable=True)
```
richardc0: nullable=False?

app/v2/notifications/post_notifications.py
```diff
@@ -222,3 +282,27 @@ def get_reply_to_text(notification_type, form, template):
         reply_to = template.get_reply_to_text()
 
     return reply_to
+
+
+def get_precompiled_letter_template(service_id):
+    template = Template.query.filter_by(
```
richardc0: Could make this into a dao method as I'll need something for determining if a letter is precompiled or not

app/v2/notifications/post_notifications.py
```diff
@@ -222,3 +282,27 @@ def get_reply_to_text(notification_type, form, template):
...
+        service_id=service_id,
+        template_type=LETTER_TYPE,
+        hidden=True
+    ).first()
```
richardc0: One? We should only ever have one and if we don't something is wrong?

BUT it shouldn't break the system

app/v2/notifications/post_notifications.py
```diff
@@ -53,6 +60,50 @@
...
+
+    template = get_precompiled_letter_template(authenticated_service.id)
+
+    form['personalisation'] = {
+        'address_line_1': form['reference']
+    }
+
+    reply_to = get_reply_to_text(LETTER_TYPE, form, template)
+
+    notification = process_letter_notification(
```
richardc0: Should do we need to do any other file validation?

I know we are assuming good path but anything easy to catch? File size maybe? I'm checking for file type on the client but double up here?

migrations/versions/0168_hidden_templates.py
allait: Oops! Fixed now.

PR #1717, opened 2018-02-27
Support celery multi

sakisv:
### Add a start script for running celery multi

The existing script would not work with `celery multi` as it was trying
to put it in the background and the get its pid.~

`celery multi` creates a number of worker processes and stores their
PIDs in files named celeryN.pid, where N the index number of the worker
(starting at 1).

### Make the sender-worker use 4 celery workers and 2G of RAM


sakisv:
We need to test this before it gets merged.

scripts/run_multi_worker_app_paas.sh
```diff
@@ -0,0 +1,123 @@
...
+  fi
+
+  if [ -z "${CW_APP_NAME}" ]; then
+    CW_APP_NAME=${NOTIFY_APP_NAME}
+  fi
+}
+
+function configure_aws_logs {
+  # create files so that aws logs agent doesn't complain
+  touch /home/vcap/logs/gunicorn_error.log
```
richardc0: /home/vcap/logs/ is used a few times - could make it a const so its easier to change

scripts/run_multi_worker_app_paas.sh
```diff
@@ -0,0 +1,123 @@
...
+file = /home/vcap/logs/gunicorn_error.log
+log_group_name = paas-${CW_APP_NAME}-gunicorn
+log_stream_name = {hostname}
+EOF
+}
+
+# For every PID, check if it's still running
+# if it is, send the sigterm
+function on_exit {
+  n=0
```
richardc0: Could use a better descriptor? wait_time? total_wait_time?

scripts/run_multi_worker_app_paas.sh
```diff
@@ -0,0 +1,123 @@
...
+      if [[ "$n" -ge "$TERMINATE_TIMEOUT" ]]; then
+        echo "Timeout reached, killing process with pid ${APP_PID}"
+        kill -9 ${APP_PID} || true
+        continue
+      else
+        echo "Timeout not reached yet, checking " ${APP_PID}
+        # else, if process is still running send SIGTERM
+        if [[ $(kill -0 ${APP_PID} 2&>/dev/null) ]]; then
+          echo "Terminating celery process with pid ${APP_PID}"
+          kill ${APP_PID} || true
```
richardc0: Is it better be specific sigterm -15?

scripts/run_multi_worker_app_paas.sh
sakisv: I was just copying from the `run_app_paas.sh`, but ok, I'll change it in both.

scripts/run_multi_worker_app_paas.sh
sakisv: I don't think so, I'd only specify a signal if it's not the default.

PR #1719, opened 2018-02-28
Send one-off messages as fast as possible

quis:
The vast majority of messages that are being sent one-off are time sensitive. A typical example is a caseworker on the phone who sends a message at the end of the call. They normally wait until the message has been delivered, so all the time they’re waiting is time when they can’t be helping someone else.

What we don’t want to happen is for the messages they’re sending to get stuck behind a big lump of GOV.UK Subscription emails or passport reminder texts. I think the best way to do this is shift them onto the priority queue.

We’re currently seeing queue sizes of up to 5,000 on the ‘normal’ queues; I don’t think there’s any risk of this change making the priority queue more heavily-laden than this. Especially since the
traffic patterns of users sending one-off messages won’t be spiky.

I also split up the parametrize definition in the tests a bit because it was getting really hard to follow.

### DIFF
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -118,6 +118,13 @@ def send_notification_to_queue(notification, research_mode, queue=None):
     if research_mode or notification.key_type == KEY_TYPE_TEST:
         queue = QueueNames.RESEARCH_MODE
 
+    if all((
+        not queue,
+        notification.api_key_id is None,
+        notification.job_id is None,
+    )):
+        queue = QueueNames.PRIORITY
+
     if notification.notification_type == SMS_TYPE:
         if not queue:
             queue = QueueNames.SEND_SMS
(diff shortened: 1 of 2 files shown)


app/notifications/process_notifications.py
```diff
@@ -118,6 +118,13 @@ def send_notification_to_queue(notification, research_mode, queue=None):
     if research_mode or notification.key_type == KEY_TYPE_TEST:
         queue = QueueNames.RESEARCH_MODE
 
+    if all((
```
servingUpAces: Could we not just send queue_name=priority from app.service.send_notification.send_one_off_notification?

quis: Oh yeah, would be super easy to change this: https://github.com/alphagov/notifications-api/blob/d584ca49498ac28d89460d73115610d856cbb722/app/service/send_notification.py#L78

quis:
Closing in favour of https://github.com/alphagov/notifications-api/pull/1722

PR #1721, opened 2018-02-28
Remove old letter flow

servingUpAces:
Now that we only send pdf files to the letter provider we no longer need the code to build the text files. This PR removes all the code used in that flow. 

Next will be to remove letters_as_pdf as a service permission.


app/celery/tasks.py
```diff
@@ -335,29 +329,6 @@ def save_letter(
...
-            self.retry(queue=QueueNames.RETRY)
-    # specifically don't catch celery.retry errors
-    except (SQLAlchemyError, BotoClientError):
-        current_app.logger.exception("build_dvla_file threw exception")
-        self.retry(queue=QueueNames.RETRY)
-
-
 @notify_celery.task(bind=True, name='update-letter-job-to-sent')
 @statsd(namespace="tasks")
 def update_job_to_sent_to_dvla(self, job_id):
```
kentsanggds: Is this also deprecated as letter pdfs are processed by date rather than job

servingUpAces: good spot

PR #1723, opened 2018-02-28
Remove priority flag from example templates

quis:
Since we send all one off messages as priority [now](https://github.com/alphagov/notifications-api/pull/1722), we don’t need to explicitly mark the ‘Example text message template’ template as
being priority.

This commit retroactively removes the priority flag from these templates, so that they can’t be reused to send other messages as priority. I don’t think this is a real problem now, but worth cleaning this up.

This query will hit approximately 827 templates.

Depends on:
- [x] https://github.com/alphagov/notifications-admin/pull/1915 (otherwise we might create some new, unwanted priority templates in the mean time)

### DIFF
--- a/migrations/versions/0172_deprioritise_examples.py
+++ b/migrations/versions/0172_deprioritise_examples.py
@@ -1,0 +1,31 @@
+"""
+
+Revision ID: 0172_deprioritise_examples
+Revises: 0171_add_org_invite_template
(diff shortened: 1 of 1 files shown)


quis:
@servingUpAces I accidentally set all the templates to `PRIORITY` instead of `NORMAL`. Whoops. Have fixed and squashed now.

PR #1729, opened 2018-03-01
Have workers also look at the priority queue

quis:
This is to try to work around a Celery problem where the worker doesn’t look at the queue often enough because it’s often empty.

Want to deploy this and see if it makes a difference to the performance of priority messages.


quis:
Going to try to fix this with Celery 4 in a few weeks.

PR #1732, opened 2018-03-02
Return `is_precompiled_letter` field as part of json for notification by id

kentsanggds:
## What 

`is_precompiled_letter` field is required by Admin app so that it can distinguish between created and precompiled letters

## Reference 

https://www.pivotaltracker.com/story/show/155323846


app/dao/services_dao.py
```diff
@@ -571,6 +574,8 @@ def dao_fetch_monthly_historical_usage_by_template_for_service(service_id, year)
                 new_stat.month = int(today_result.month)
                 new_stat.year = int(today_result.year)
                 new_stat.count = today_result.count
+                new_stat.hidden = today_result.hidden
                 stats.append(new_stat)
 
+    print(stats[2].__dict__)
```
richardc0: Remove

app/letters/utils.py
```diff
@@ -79,4 +80,8 @@ def get_letter_pdf(notification):
 
 
 def is_precompiled_letter(template):
-    return template.hidden and template.name == current_app.config['PRECOMPILED_TEMPLATE_NAME']
```
servingUpAces: I think this belongs in the data model, if it's a property it should be included in the marshmallow schema without a change.

kentsanggds: I guess it could go in to the data model for the Template class just as a getter.
Will refactor and push through after Richards bug fix is deployed

PR #1734, opened 2018-03-02
Send count of sent letters to performance platform

quis:
Now we’ve been sending real letters for quite a while it would be nice to show how many.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -221,6 +221,7 @@ def send_total_sent_notifications_to_performance_platform():
     count_dict = total_sent_notifications.get_total_sent_notifications_yesterday()
     email_sent_count = count_dict.get('email').get('count')
     sms_sent_count = count_dict.get('sms').get('count')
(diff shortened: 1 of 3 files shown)


quis:
Tested with this curl command:
```shell
curl -X POST \
-d '{"_id": "MjAxOC0wMy0wMVQwMDowMDowMGdvdnVrLW5vdGlmeWxldHRlcm5vdGlmaWNhdGlvbnNkYXk=", "_timestamp": "2018-03-01T00:00:00", "service": "govuk-notify", "dataType": "notifications", "period": "day", "count": 1, "channel":"letter"}' \
-H 'Content-type: application/json' -H 'Authorization: Bearer <token>' https://www.performance.service.gov.uk/data/govuk-notify/notifications
```

quis:
Showing our one letter:

image

leohemsted:
do we want to back-populate this data?

quis:
Yeah, looks like we can do it with this: https://github.com/alphagov/notifications-api/blob/f248234cfcf0e70cc5990a47579d68bff13ca6dc/app/commands.py#L212-L215

PR #1737, opened 2018-03-05
Updated API to handle pre-compiled pdfs

richardc0:
This PR is to added the precompiled flow to the template endpoint so that the API can determine if a notification is a precompiled PDF and in which case should get the pdf file from S3 and then generate the png from the template preview endpoint precompiled-preview.png rather than the preview.png endpoint. If the notification is a standard letter the flow should not change.

* Added a method to letter/utils.py to get the PDF document from the S3 bucket
* Added the logic to return the pdf or to produce a png of the pdf
* Added is_precompiled_letter method to letter/utils.py
* Added tests for letter/utils.py
* Added tests for the rest endpoint
* Moved the Precompiled name to a central location
* Added hidden field to the test method to create a template


tests/app/letters/test_letter_utils.py
```diff
@@ -0,0 +1,29 @@
...
+        reference=sample_notification.reference
+    ).upper()
+
+
+def test_get_bucket_prefix_for_notification_invalid_notification():
+    with pytest.raises(AttributeError):
+        get_bucket_prefix_for_notification(None)
+
+
+def test_is_precompiled_letter_false(sample_letter_template):
```
kentsanggds: Might also want to check conditions where it is hidden but doesn't have the correct template name and where it does have the correct template name but is not hidden

tests/app/template/test_rest.py
```diff
@@ -874,3 +875,215 @@ def test_preview_letter_template_by_id_template_preview_500(
...
+                               subject='Pre-compiled PDF',
+                               hidden=True)
+
+    notification = create_notification(template)
+
+    with set_config_values(notify_api, {
+        'TEMPLATE_PREVIEW_API_HOST': 'http://localhost/notifications-template-preview',
+        'TEMPLATE_PREVIEW_API_KEY': 'test-key'
+    }):
+        import requests_mock
```
kentsanggds: Perhaps move the imports to the top?

app/template/rest.py
```diff
@@ -194,29 +199,61 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
...
-        'template': template_for_letter_print,
-        'values': notification.personalisation,
-        'dvla_org_id': service.dvla_organisation_id,
-    }
+            pdf_file = get_letter_pdf(notification)
 
-    resp = requests_post(
-        '{}/preview.{}{}'.format(
+        except botocore.exceptions.ClientError:
+            current_app.logger.info
```
leohemsted: Not logging anything!

app/template/rest.py
```diff
@@ -225,5 +262,4 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
             'Error generating preview for {}'.format(notification_id), status_code=500
```
leohemsted: also not logging anything!

PR #1742, opened 2018-03-06
Renamed 'Pre-compiled PDF' to 'Provided as PDF'

richardc0:
Initially we used the term 'Pre-compiled PDF' but after a time demo to
the rest of the team we decided on 'Provided as PDF'. This updates the
config to use the term and updates any file in the code base which uses
a static string to use the configuration item so that it is easier to
maintain in future and static strings don't get hidden away.

* Updated the Config with the new name
* Updated endpoint to use config item
* Updated static strings in tests to use the config item

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -642,7 +642,7 @@ class TemplateProcessTypes(db.Model):
     name = db.Column(db.String(255), primary_key=True)
 
 
-PRECOMPILED_TEMPLATE_NAME = 'Pre-compiled PDF'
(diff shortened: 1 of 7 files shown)


quis:
Did we get @thomchambers input on this?

richardc0:
@quis I gave Thom a demo he said he'd speak to you. Let me know if anything needs to change.

kentsanggds:
Closed as changes were placed on the Admin side instead after chat with CHS.

PR #1745, opened 2018-03-06
Match on partial email addresses and phone numbers in search

quis:
Users expect the search to work on partial email addresses ‘similar to Gov.Pay’. We can’t have a situation where Pay are doing something better than us 😜

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -446,7 +446,7 @@ def dao_get_notifications_by_to_field(service_id, search_term, statuses=None):
 
     filters = [
         Notification.service_id == service_id,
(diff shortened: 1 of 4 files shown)


quis:
Also: 

# Do more normalisation for better partial matching

Phone numbers sometimes contain stuff we normalise out. This matches perfectly if we have a full phone number, because we can normalise the thing we’re searching for in the same way as the search term.

With partial search terms we can’t do this completely, because we can’t work out if ‘123’ is part of a UK number, an international number, the start of the phone number, the last 3 digits, etc.

What we can do is remove some stuff that we can know will cause partial search terms to not match:
- leading pluses
- leading `0`s
- any brackets
- any spaces

PR #1746, opened 2018-03-06
Add endpoint to check for unique organisation names

gov-cjwaszczuk:
Currently, when changing a service name, a request is made to check if the name is unique before the update request is made. This is so we can give the user good errors before they continue through the flow to verify their password and get hit by a 'name already in use' error anyway.

This PR contains the endpoint to check the organisation name choice is unique, and keep the same standard of error formatting for the user.

**Notes**
- `email_from` has been removed from the organisation expected request params. Unlike a service, it's not necessary for an organisation. In place of email_from, a check will ensure that it is not possible to enter a service name that already exists, albeit in a different case. 
- Also possible for an organisation to bypass the case check if the current identical name happens to belong to the same org id.

### DIFF
--- a/app/organisation/rest.py
+++ b/app/organisation/rest.py
@@ -107,3 +107,26 @@ def get_organisation_users(organisation_id):
 
     result = user_schema.dump(org_users, many=True)
     return jsonify(data=result.data)
+
+
+@organisation_blueprint.route('/unique', methods=["GET"])
+def is_organisation_name_unique():
+    organisation_id, name = check_request_args(request)
+
+    name_exists = Organisation.query.filter(Organisation.name.ilike(name)).first()
+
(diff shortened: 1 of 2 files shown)


klssmith:
When we check service name uniqueness we check the `email_from` as well as whether the service name already exists in the database. This stops you from having a service called `my service` and being allowed to create a second service called `MY SERVICE`.

Since we're not using `email_from` for organisations, I think we'll have to change  or add in some more checks on the organisation name uniqueness. Otherwise you would be able to have an org called `Test Organisation` and be allowed to created another organisation called `test organisation` which we probably don't want.

gov-cjwaszczuk:
Makes sense, I'll take a look over it and add some sort of check.

PR #1747, opened 2018-03-06
Fixed bug only where only data being passed to TemplatePreview when JSON is required

richardc0:
Some template preview endpoints take data and some take json. After a refactor only 'data' was being passed and hence the call where json was passed was producing an error because the content header is automatically being set based on the parameter.

* Added some logic to handle if the post requires data or json
* Added extra logging to display the error with more detail
* Added tests to ensure the correct data type is being passed to the post request

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -266,7 +266,7 @@ def _get_png_preview(url, data, notification_id):
                 resp.content
             ))
         raise InvalidRequest(
-            'Error generating preview letter for {} \nStatus code: {}\n{}'.format(
+            'Error generating preview letter for {}\nStatus code: {}\n{}'.format(
                 notification_id,
                 resp.status_code,
                 resp.content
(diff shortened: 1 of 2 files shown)


app/template/rest.py
```diff
@@ -253,13 +254,23 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
 def _get_png_preview(url, data, notification_id):
     resp = requests_post(
         url,
-        data=data,
+        json=data,
```
leohemsted: this now won't work for the binary data in the precompiled endpoint 😩

app/template/rest.py
```diff
@@ -205,8 +205,9 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
 
             pdf_file = get_letter_pdf(notification)
 
-        except botocore.exceptions.ClientError:
-            current_app.logger.info
+        except botocore.exceptions.ClientError as e:
+            current_app.logger.exception(
+                'Error getting letter file from S3 notification id {}'.format(notification_id), e)
```
leohemsted: `exception` log doesn't need the exception - it pulls it automatically out of `sys.exc_info`. So we just want `current_app.logger.exception('My error msg {}'.format(my_args))`

PR #1752, opened 2018-03-08
Handle malformed invite tokens

gov-cjwaszczuk:
Currently, If a malformed invite token hits the API, the server will return with a 500. This PR handles the bad tokens and returns a 400 with an error message, the same as used in malformed email login tokens.

### DIFF
--- a/app/accept_invite/rest.py
+++ b/app/accept_invite/rest.py
@@ -4,7 +4,7 @@ from flask import (
     current_app
 )
 
-from itsdangerous import SignatureExpired
+from itsdangerous import SignatureExpired, BadData
 
 from notifications_utils.url_safe_token import check_token
 
@@ -37,6 +37,9 @@ def validate_invitation_token(invitation_type, token):
         errors = {'invitation':
                   ['Your invitation to GOV.UK Notify has expired. '
                    'Please ask the person that invited you to send you another one']}
+        raise InvalidRequest(errors, status_code=400)
+    except BadData:
+        errors = {'invitation': 'Something’s wrong with this link. Make sure you’ve copied the whole thing.'}
         raise InvalidRequest(errors, status_code=400)
 
     if invitation_type == 'service':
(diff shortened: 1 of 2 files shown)


app/accept_invite/rest.py
```diff
@@ -38,6 +38,9 @@ def validate_invitation_token(invitation_type, token):
                   ['Your invitation to GOV.UK Notify has expired. '
                    'Please ask the person that invited you to send you another one']}
         raise InvalidRequest(errors, status_code=400)
+    except BadData:
+        errors = {'invitation': 'There’s something wrong with the link you’ve used.'}
```
quis: Why are people hitting this error? Is it because they haven’t copy/pasted the whole URL?

gov-cjwaszczuk: yeah

quis: The error message should help them fix that then…

app/accept_invite/rest.py
```diff
@@ -38,6 +38,9 @@ def validate_invitation_token(invitation_type, token):
                   ['Your invitation to GOV.UK Notify has expired. '
                    'Please ask the person that invited you to send you another one']}
         raise InvalidRequest(errors, status_code=400)
+    except BadData:
+        errors = {'invitation': 'Something’s wrong with this link. Make sure you’ve copied the whole thing.'}
```
leohemsted: can we achieve this by just changing `BadSignature` to `BadData` in `app/__init__.py`? (and updating the error message there.

That means we don't have multiple messages for all the different tokens - which is a list that i think looks something like this:

* '/invitation/'
* '/organisation-invitation/'
* '/new-password/'
* '/verify-email/'
* '/user-profile/email/confirm/'
* '/email-auth/'

leohemsted: oh wait this is on api

gov-cjwaszczuk: yeah I was sufficiently confused by the comment :P

leohemsted: so is this needed anymore? https://github.com/alphagov/notifications-admin/blob/master/app/__init__.py#L522-L526

PR #1754, opened 2018-03-08
Revert service callback eventlets

leohemsted:
We have seen problems with the service callback workers due to the db connection pool being exhausted. When the worker picks up the task, it makes a db query to get the notification, a query to get the callback url, and then closes the session before it makes the 3rd party request.

However, even closing the session before the (potentially lengthy) web request wasn't enough - we've seen significant amounts of `sqlalchemy.exc.TimeoutError`s.

This reverts commit 2dfbd93


leohemsted:
Medium-long term, we should move the service callback task to take in all the data it needs through celery arguments, then the task doesn't need any db access at all, and can move back to eventlets

```
notification.id
notification.client_reference
notification.to
notification.status
notification.created_at
notification.updated_at
notification.sent_at
notification.notification_type
service_callback_api.url
service_callback_api.bearer_token
```

PR #1755, opened 2018-03-08
Refactor service callback task

servingUpAces:
When we sent the service the status callback for a notification, we have all the information we need.

Which means we can remove the need to request the data from the database.
In order for the PR to be backwards compatible I have added an optional parameter "encrypted_status_update".
If this is not None then the new code is called.

The next PR will send the encrypted data to this task.

A final PR will remove the code that uses the database to get the notification and service callback api.

A new server command to replay service status updates.

### DIFF
--- a/app/celery/service_callback_tasks.py
+++ b/app/celery/service_callback_tasks.py
@@ -73,7 +73,9 @@ def send_delivery_status_to_service(self, notification_id,
                     self.retry(queue=QueueNames.RETRY)
                 except self.MaxRetriesExceededError:
                     current_app.logger.exception(
-                        'Retry: send_delivery_status_to_service has retried the max num of times')
+                        """Retry: send_delivery_status_to_service has retried the max num of times
+                         for notification: {}""".format(notification_id)
+                    )
 
 
 def process_update_with_notification_id(self, notification_id):
@@ -135,4 +137,7 @@ def process_update_with_notification_id(self, notification_id):
         try:
             self.retry(queue=QueueNames.RETRY)
         except self.MaxRetriesExceededError:
-            current_app.logger.exception('Retry: send_delivery_status_to_service has retried the max num of times')
+            current_app.logger.exception(
+                """Retry: send_delivery_status_to_service has retried the max num of times
+                 for notification: {}""".format(notification_id)
+            )
--- a/app/celery/service_callback_tasks.py
+++ b/app/celery/service_callback_tasks.py
@@ -6,6 +6,7 @@ from app import (
     db,
     DATETIME_FORMAT,
     notify_celery,
+    encryption
 )
 from app.dao.notifications_dao import (
     get_notification_by_id,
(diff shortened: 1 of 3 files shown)


app/celery/service_callback_tasks.py
```diff
@@ -23,10 +24,61 @@
...
+                    status_update['service_callback_api_url'],
+                    e
+                )
+            )
+            if not isinstance(e, HTTPError) or e.response.status_code >= 500:
+                try:
+                    self.retry(queue=QueueNames.RETRY)
+                except self.MaxRetriesExceededError:
+                    current_app.logger.exception(
+                        'Retry: send_delivery_status_to_service has retried the max num of times')
```
richardc0: Will you get information of the notification_id etc here?

leohemsted: you've missed out the unhandled exception logic that i added yesterday, it would be good to keep that in.

app/commands.py
```diff
@@ -311,5 +315,55 @@ def insert_inbound_numbers_from_file(file_name):
...
+            errors.append("Reference: {} was not found in notifications.".format(ref))
+
+    for e in errors:
+        print(e)
+    if errors:
+        raise Exception("Some notifications for the given references were not found")
+
+    for n in notifications:
+        data = {
+            "notification_id": str(n.id),
```
leohemsted: don't need this - we aren't using it in the send_delivery_status task as we pass through the id separately

app/celery/service_callback_tasks.py
```diff
@@ -23,10 +24,61 @@
...
+                                    encrypted_status_update=None
+                                    ):
+    if not encrypted_status_update:
+        process_update_with_notification_id(self, notification_id=notification_id)
+    else:
+        try:
+            status_update = encryption.decrypt(encrypted_status_update)
+
+            data = {
+                "id": str(notification_id),
```
leohemsted: notification_id is already a string (celery doesn't accept UUIDs as args)

app/commands.py
```diff
@@ -311,5 +315,55 @@ def insert_inbound_numbers_from_file(file_name):
...
+            "notification_sent_at": n.sent_at.strftime(DATETIME_FORMAT),
+            "notification_type": n.notification_type,
+            "service_callback_api_url": callback_api.url,
+            "service_callback_api_bearer_token": callback_api.bearer_token,
+        }
+        encrypted_status_update = encryption.encrypt(data)
+        send_delivery_status_to_service.apply_async([str(n.id), encrypted_status_update],
+                                                    queue=QueueNames.CALLBACKS)
+
+    print("Replay service status for service: {}. Sent {} notification status updates".format(service_id,
```
richardc0: I assume this is just for us? Technically some might have failed but we would know yet?

app/commands.py
```diff
@@ -311,5 +315,55 @@ def insert_inbound_numbers_from_file(file_name):
...
+    callback_api = get_service_callback_api_for_service(service_id=service_id)
+    if not callback_api:
+        print("Callback api was not found for service: {}".format(service_id))
+        return
+
+    errors = []
+    notifications = []
+    file = open(file_name)
+
+    for ref in file:
```
richardc0: is ref notificiation id then => notification_id

app/commands.py
```diff
@@ -311,5 +315,55 @@ def insert_inbound_numbers_from_file(file_name):
...
+        try:
+            notification = Notification.query.filter_by(client_reference=ref.strip()).one()
+            notifications.append(notification)
+        except NoResultFound as e:
+            errors.append("Reference: {} was not found in notifications.".format(ref))
+
+    for e in errors:
+        print(e)
+    if errors:
+        raise Exception("Some notifications for the given references were not found")
```
richardc0: Do you want to exception here or carry on?

We just want to know which ones we didn't find?

tests/app/celery/test_service_callback_tasks.py
```diff
@@ -7,35 +7,141 @@
...
 
 
 @pytest.mark.parametrize("notification_type",
                          ["email", "letter", "sms"])
-def test_send_delivery_status_to_service_post_https_request_to_service(notify_db,
-                                                                       notify_db_session,
-                                                                       notification_type):
-    user = create_user()
-    service = create_sample_service(notify_db, notify_db_session, user=user, restricted=True)
+def test_send_delivery_status_to_service_post_https_request_to_service_with_encrypted_data(
```
richardc0: Add a test for a NoResultFound on notification id?

app/commands.py
```diff
@@ -311,5 +315,55 @@ def insert_inbound_numbers_from_file(file_name):
...
+            "notification_status": n.status,
+            "notification_created_at": n.created_at.strftime(DATETIME_FORMAT),
+            "notification_updated_at": n.updated_at.strftime(DATETIME_FORMAT),
+            "notification_sent_at": n.sent_at.strftime(DATETIME_FORMAT),
+            "notification_type": n.notification_type,
+            "service_callback_api_url": callback_api.url,
+            "service_callback_api_bearer_token": callback_api.bearer_token,
+        }
+        encrypted_status_update = encryption.encrypt(data)
+        send_delivery_status_to_service.apply_async([str(n.id), encrypted_status_update],
```
richardc0: Re. message below, could you do it synchronous and catch the exception so you have a better idea?

app/commands.py
```diff
@@ -311,5 +315,55 @@ def insert_inbound_numbers_from_file(file_name):
...
+def replay_service_callbacks(file_name, service_id):
+    print("Start send service callbacks for service: ", service_id)
+    callback_api = get_service_callback_api_for_service(service_id=service_id)
+    if not callback_api:
+        print("Callback api was not found for service: {}".format(service_id))
+        return
+
+    errors = []
+    notifications = []
+    file = open(file_name)
```
leohemsted: always use `with` statements to open files, and try not to name variables `file` as it's obscuring a built-in

tests/app/celery/test_service_callback_tasks.py
```diff
@@ -7,35 +7,141 @@
...
+    assert request_mock.request_history[0].url == callback_api.url
+    assert request_mock.request_history[0].method == 'POST'
+    assert request_mock.request_history[0].text == json.dumps(mock_data)
+    assert request_mock.request_history[0].headers["Content-type"] == "application/json"
+    assert request_mock.request_history[0].headers["Authorization"] == "Bearer {}".format(callback_api.bearer_token)
+
+
+@pytest.mark.parametrize("notification_type",
+                         ["email", "letter", "sms"])
+def test_send_delivery_status_to_service_retries_if_request_returns_500_with_encrypted_data(
```
leohemsted: should probs add a new test to make sure it retries if the DB errors too.

app/commands.py
servingUpAces: It's the client reference

app/commands.py
servingUpAces: I want a whole file to be valid here.

tests/app/celery/test_service_callback_tasks.py
servingUpAces: There are no hits to the db

leohemsted: duhhhh 👍

PR #1756, opened 2018-03-09
Updated the notification template endpoint to extract the pdf page

richardc0:
Updated the notification template endpoint to extract the pdf page before sending it to template preview. This stops the whole pdf file being sent to template preview for each page which is really inefficient on network traffic and memory usage.

* Added logic to the endpoint to extract the specific page requested
* Updated tests to add a mock for the new call to utils
* Added a new test case for exceptions in the PDF extraction process

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -1,11 +1,14 @@
 import base64
+from io import BytesIO
 
 import botocore
+from PyPDF2.utils import PdfReadError
 from flask import (
     Blueprint,
     current_app,
     jsonify,
     request)
+from notifications_utils.pdf import extract_page_from_pdf
 from requests import post as requests_post
 
 from app.dao.notifications_dao import get_notification_by_id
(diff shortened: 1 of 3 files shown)


app/template/rest.py
```diff
@@ -211,9 +214,21 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
         content = base64.b64encode(pdf_file).decode('utf-8')
 
         if file_type == 'png':
-            url = '{}/precompiled-preview.png{}'.format(
-                current_app.config['TEMPLATE_PREVIEW_API_HOST'],
-                '?page={}'.format(page) if page else ''
+
+            try:
+                page_number = page if page else "0"
```
kentsanggds: I wonder if casting `page` var on line 196 to an int would be better than the casting logic here as it should still work with the format logic for the created letter path

app/template/rest.py
```diff
@@ -211,9 +214,21 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
...
+                pdf_page = extract_page_from_pdf(BytesIO(pdf_file), int(page_number) - 1)
+                content = base64.b64encode(pdf_page).decode('utf-8')
+            except PdfReadError:
+                current_app.logger.exception(
```
kentsanggds: Do we need this `current_app.logger.exception` here, as the error handler in `errors.py` should already log exceptions.

PR #1758, opened 2018-03-12
Filter revoked api keys older than 7 days

gov-cjwaszczuk:
Currently there is no filter on the api_keys query. In certain cases, it clutters the admin panel with redundant api keys. This PR filters out any revoked API keys that are older than 7 days.

### DIFF
--- a/tests/app/dao/test_api_key_dao.py
+++ b/tests/app/dao/test_api_key_dao.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import pytest
 from sqlalchemy.exc import IntegrityError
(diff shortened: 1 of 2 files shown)


tests/app/dao/test_api_key_dao.py
```diff
@@ -95,3 +95,16 @@ def test_save_api_key_should_not_create_new_service_history(sample_service):
...
+    expired_api_key = ApiKey(**{'service': sample_service,
+                                'name': sample_service.name,
+                                'created_by': sample_service.created_by,
+                                'key_type': KEY_TYPE_NORMAL,
+                                'expiry_date': datetime.utcnow() - timedelta(days=8)})
+    save_model_api_key(expired_api_key)
+
+    all_api_keys = get_model_api_keys(service_id=sample_service.id)
+
+    assert len(all_api_keys) == 0
```
leohemsted: can you add a second key with expiry date of 6 days ago and make sure that's in the returned list?

PR #1761, opened 2018-03-12
Add new virus scan statuses

klssmith:
Added the following new notification statuses:
* `pending-virus-check`
* `virus-scan-failed`

If we decide to remove these statuses in future, we will need to replace
them with a different status in the `notifications` and
`notification_history` tables where they are referenced, so
`pending-virus-check` will be replaced with `sending`, and `virus-scan-failed`
will be replaced with `permanent-failure`.

[Pivotal story](https://www.pivotaltracker.com/story/show/155323461)


tests/app/v2/notifications/test_get_notifications.py
```diff
@@ -422,7 +422,8 @@ def test_get_all_notifications_filter_by_status_invalid_status(client, sample_no
     assert json_response['status_code'] == 400
```
richardc0: Should we add a new test which sets the status to pending-virus-check and virus-scan-failed (or add them to a parametized test) to check they don't do things that are not expected

i.e. A virus-scan-failed might appears on get message status but pending-virus-check wouldn't get sent etc?

app/models.py
```diff
@@ -1044,6 +1046,8 @@ def check_code(self, cde):
     NOTIFICATION_TECHNICAL_FAILURE,
     NOTIFICATION_TEMPORARY_FAILURE,
     NOTIFICATION_PERMANENT_FAILURE,
+    NOTIFICATION_PENDING_VIRUS_CHECK,
+    NOTIFICATION_VIRUS_SCAN_FAILED,
```
leohemsted: should `VIRUS_SCAN_FAILED` also be put in the `NOTIFICATION_STATUS_TYPES_FAILED` list on line 1009?

klssmith: 👍 I've added it in to the list

kentsanggds:
After further thought, the status for a notification in `pending_virus_check` should probably be `created` during a downgrade as this is what the expected state will be processed for letters processed in `collate_letter_pdfs_for_day`. Story has been updated with correct state.

PR #1763, opened 2018-03-12
Small updates to 'update-letter-notifications-statuses' task

klssmith:
Simplified the code in the `update-letter-notifications-statuses` task to get the `billing_day` for the `daily_sorted_letter` table to use today's date instead of getting the date from the response file name.

Moved the check for temporary failures to a `finally` block, so that this is executed even if the response file contains a status other than `Sorted` or `Unsorted`.

This is an update to [this Pivotal story.](https://www.pivotaltracker.com/story/show/154395015)


klssmith:
Closing and replacing with https://github.com/alphagov/notifications-api/pull/1765.

PR #1765, opened 2018-03-13
Ensure update letter notification statuses task handles temporary failure

klssmith:
Two small updates to the `update-letter-notifications-statuses` task -

* We ensure that temporary failures are always handled, regardless of whether 
the response file we receive contains unknown Sorted statuses or not by putting
this code in a `finally` block.

* When getting the billing date from the name of the letter response file we now split
on `-` instead of `.` since the files we receive use `-` in the filename now.

This is an update to [this Pivotal story.](https://www.pivotaltracker.com/story/show/154395015)

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -420,22 +420,23 @@ def update_letter_notifications_statuses(self, filename):
             update_letter_notification(filename, temporary_failures, update)
             sorted_letter_counts[update.cost_threshold] += 1
 
-        if temporary_failures:
-            # This will alert Notify that DVLA was unable to deliver the letters, we need to investigate
-            message = "DVLA response file: {filename} has failed letters with notification.reference {failures}".format(
-                filename=filename, failures=temporary_failures)
-            raise DVLAException(message)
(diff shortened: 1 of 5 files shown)


app/celery/tasks.py
```diff
@@ -420,22 +420,23 @@ def update_letter_notifications_statuses(self, filename):
...
+            persist_daily_sorted_letter_counts(billing_date, sorted_letter_counts)
+        finally:
+            if temporary_failures:
+                # This will alert Notify that DVLA was unable to deliver the letters, we need to investigate
+                message = "DVLA response file: {filename} has failed letters with notification.reference {failures}" \
+                    .format(filename=filename, failures=temporary_failures)
+                raise DVLAException(message)
 
 
 def get_billing_date_in_bst_from_filename(filename):
```
servingUpAces: do we need to split the filename on "-"

klssmith: Yes, we definitely do - I had forgotten to add that bit back in when I changed the PR.

PR #1766, opened 2018-03-13
User orgs and services

leohemsted:
New endpoint that returns a user's organisations and services. It groups services by organisation - provided that the user knows about that organisation. It only returns active services/orgs.

### DIFF
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -31,7 +31,7 @@ from app.notifications.process_notifications import (
 )
 from app.schemas import (
     email_data_request_schema,
-    user_schema,
+    create_user_schema,
     permission_schema,
     user_update_schema_load_json,
     user_update_password_schema_load_json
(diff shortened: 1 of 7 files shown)


app/user/rest.py
```diff
@@ -392,7 +393,45 @@ def update_password(user_id):
...
         raise InvalidRequest(errors, status_code=400)
     update_user_password(user, pwd)
-    return jsonify(data=user_schema.dump(user).data), 200
+    return jsonify(data=user.serialize()), 200
+
+
+@user_blueprint.route('/<uuid:user_id>/organisations-and-services', methods=['GET'])
+def get_organisations_and_services_for_user(user_id):
+    user = get_user_and_accounts(user_id)
+    data = {
```
servingUpAces: It might be nice to pull this into a method.

Maybe it belongs in a schema or a method on the user?

PR #1769, opened 2018-03-14
Escape special characters in search by recipient

quis:
SQLAlchemy handles escaping anything that could allow a SQL injection attack. But it doesn’t escape the characters used for wildcard searching. This is the reason we’re able to do `.like('%example%')` at all.

But we shouldn’t be letting our users search with wildcard characters, so we need to escape them. Which is what this commit does.

Documentation: https://www.postgresql.org/docs/9.0/static/functions-matching.html#FUNCTIONS-LIKE

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -455,10 +455,10 @@ def dao_get_notifications_by_to_field(service_id, search_term, notification_type
     else:
         raise InvalidRequest("Only email and SMS can use search by recipient", 400)
 
-    for special_character in {'_', '%', '/'}:
+    for special_character in ('\\', '_', '%', '/'):
         normalised = normalised.replace(
             special_character,
-            '\\{}'.format(special_character)
+            '\{}'.format(special_character)
         )
 
     filters = [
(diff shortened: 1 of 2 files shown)


app/dao/notifications_dao.py
```diff
@@ -455,6 +455,12 @@ def dao_get_notifications_by_to_field(service_id, search_term, notification_type
     else:
         raise InvalidRequest("Only email and SMS can use search by recipient", 400)
 
+    for special_character in {'_', '%', '/'}:
+        normalised = normalised.replace(
+            special_character,
+            '\\{}'.format(special_character)
```
leohemsted: should this be a raw string? eg `r'\\_'` - or python will convert `\\` into a single backslash

quis: Single backslash is what I want – maybe would be clearer as `r'\{}'`?

quis: 9103ca5

leohemsted: ah i think i got myself confused by reading something from the postgres mailing list from 2000.

`r` is marginally clearer  but not enough for me to mind

PR #1772, opened 2018-03-14
Upload precompiled letter to test letters bucket for test keys

kentsanggds:
## What

Currently no pdfs are uploaded to s3 so they can't be previewed in Admin, so in order to show preview of precompiled letters sent using a test key we need to put them into another bucket otherwise they could get processed as real letters.

- also some test code has been moved to a more logical file

## Reference 

https://www.pivotaltracker.com/story/show/155974923


kentsanggds:
Not needed as part of https://github.com/alphagov/notifications-api/pull/1773

PR #1773, opened 2018-03-14
Preview precompiled letters sent using an API test key

kentsanggds:
## What 
To break down this PR, you can have a look at this first - 

https://github.com/alphagov/notifications-api/pull/1772

This PR will access the PDF from the correct bucket for a preview image sent using a test API key

## Dependencies 

This is a branch off https://github.com/alphagov/notifications-api/tree/ken-upload-test_key-precompiled-to-test_letters-bucket, so should be merged after that branch has merged or can be merged in place of the merging of the parent branch.

## Reference 

https://www.pivotaltracker.com/story/show/155974923

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -321,6 +321,7 @@ class Development(Config):
 
     CSV_UPLOAD_BUCKET_NAME = 'development-notifications-csv-upload'
     LETTERS_PDF_BUCKET_NAME = 'development-letters-pdf'
+    TEST_LETTERS_BUCKET_NAME = 'development-test-letters'
     DVLA_RESPONSE_BUCKET_NAME = 'notify.tools-ftp'
 
     ADMIN_CLIENT_SECRET = 'dev-notify-secret-key'
@@ -357,6 +358,7 @@ class Test(Development):
 
     CSV_UPLOAD_BUCKET_NAME = 'test-notifications-csv-upload'
     LETTERS_PDF_BUCKET_NAME = 'test-letters-pdf'
+    TEST_LETTERS_BUCKET_NAME = 'test-test-letters'
     DVLA_RESPONSE_BUCKET_NAME = 'test.notify.com-ftp'
 
     # this is overriden in jenkins and on cloudfoundry
@@ -384,6 +386,7 @@ class Preview(Config):
     NOTIFY_ENVIRONMENT = 'preview'
     CSV_UPLOAD_BUCKET_NAME = 'preview-notifications-csv-upload'
     LETTERS_PDF_BUCKET_NAME = 'preview-letters-pdf'
+    TEST_LETTERS_BUCKET_NAME = 'preview-test-letters'
     DVLA_RESPONSE_BUCKET_NAME = 'notify.works-ftp'
     FROM_NUMBER = 'preview'
     API_RATE_LIMIT_ENABLED = True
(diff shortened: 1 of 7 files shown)


app/config.py
```diff
@@ -395,6 +398,7 @@ class Staging(Config):
     NOTIFY_ENVIRONMENT = 'staging'
     CSV_UPLOAD_BUCKET_NAME = 'staging-notify-csv-upload'
     LETTERS_PDF_BUCKET_NAME = 'staging-letters-pdf'
+    TEST_LETTERS_BUCKET_NAME = 'staging-test-letters'
```
richardc0: We should probably prefix all of these with the environment at some point

app/letters/utils.py
```diff
@@ -5,24 +5,32 @@
...
+PRECOMPILED_BUCKET_PREFIX = '{folder}NOTIFY.{reference}'
 
 
-def get_letter_pdf_filename(reference, crown):
-    now = datetime.utcnow()
+def get_folder_name(_now, is_test_letter):
+    if is_test_letter:
+        folder_name = ''
+    else:
+        print_datetime = _now
```
richardc0: Should we use this as an opportunity to use created_at instead of now()?

It would be less prone to error and would mean we haven't got to search the bucket but use the created_at date. Get file would be just obj = s3.Object(bucket_name=bucket_name, key=item.key)

Technically a notification could be created at <deadline>:999 and this be ran on <deadline>:001 which would cause the folder name to be different for the get_letter_pdf function

tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -28,25 +28,6 @@ def test_should_have_decorated_tasks_functions():
     assert create_letters_pdf.__wrapped__.__name__ == 'create_letters_pdf'
 
 
-@pytest.mark.parametrize('crown_flag,expected_crown_text', [
```
richardc0: Why were these deleted? They seem like valid tests (may need updating to determine if it was a test or not test though)

tests/app/letters/test_letter_utils.py
```diff
@@ -13,6 +33,70 @@ def test_get_bucket_prefix_for_notification_valid_notification(sample_notificati
     ).upper()
 
 
+@freeze_time(FROZEN_DATE_TIME)
```
richardc0: I think that the freeze time masks the issue with using a transient date

app/config.py
kentsanggds: Not sure what you mean here, they are already prefixed with the environment?

app/letters/utils.py
kentsanggds: Wanted to keep this PR as small as possible, that other issue could be fixed in another PR

tests/app/celery/test_letters_pdf_tasks.py
kentsanggds: They weren't deleted, just moved to a more logical place to `test_letter_utils.py`

tests/app/letters/test_letter_utils.py
kentsanggds: I would introduce a test specific for that, and maybe fix it in another PR

app/letters/utils.py
kentsanggds: Had another thought on this, if the letter gets processed but then we create the pdf after the cut off time, there's a potential that it might never gets processed if it's using the `created_at` time to indicate which subfolder it will be placed in. Think we need a bit more discussion around this.

PR #1780, opened 2018-03-19
Send service callback if sent at is none

servingUpAces:
## Context
When a notification is timed out in the scheduled task that may happen because the notification has not been sent.
Which means the sent_at date for the notification could be empty causing the service callback to fail.

## What has changed in this PR
- Allow code to work if notification.sent_at or updated_at is None
- Update calls to send_delivery_status_to_service to send the data encrypted so that the task does not need to use the db.
- Throw an exception whenever we updated a notification to technical failure. If this is happening we want to know about it.

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -33,7 +33,7 @@ def deliver_sms(self, notification_id):
             )
             self.retry(queue=QueueNames.RETRY)
         except self.MaxRetriesExceededError:
-            message = "RETRY FAILED: task send_sms_to_provider failed for notification {}. " \
+            message = "RETRY FAILED: Max retries reached. The task send_sms_to_provider failed for notification {}. " \
                       "Notification has been updated to technical-failure".format(notification_id)
             update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
             raise NotificationTechnicalFailureException(message)
@@ -57,7 +57,8 @@ def deliver_email(self, notification_id):
             )
             self.retry(queue=QueueNames.RETRY)
         except self.MaxRetriesExceededError:
-            message = "RETRY FAILED: task send_email_to_provider failed for notification {}. " \
+            message = "RETRY FAILED: Max retries reached. " \
+                      "The task send_email_to_provider failed for notification {}. " \
                       "Notification has been updated to technical-failure".format(notification_id)
             update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
             raise NotificationTechnicalFailureException(message)
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -9,6 +9,8 @@ from app.config import QueueNames
 from app.dao import notifications_dao
 from app.dao.notifications_dao import update_notification_status_by_id
 from app.delivery import send_to_providers
+from app.exceptions import NotificationTechnicalFailureException
+from app.models import NOTIFICATION_TECHNICAL_FAILURE
 
 
 @worker_process_shutdown.connect
@@ -31,10 +33,10 @@ def deliver_sms(self, notification_id):
             )
             self.retry(queue=QueueNames.RETRY)
         except self.MaxRetriesExceededError:
-            current_app.logger.exception(
-                "RETRY FAILED: task send_sms_to_provider failed for notification {}".format(notification_id),
-            )
-            update_notification_status_by_id(notification_id, 'technical-failure')
+            message = "RETRY FAILED: task send_sms_to_provider failed for notification {}. " \
+                      "Notification has been updated to technical-failure".format(notification_id)
+            update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
+            raise NotificationTechnicalFailureException(message)
 
 
 @notify_celery.task(bind=True, name="deliver_email", max_retries=48, default_retry_delay=300)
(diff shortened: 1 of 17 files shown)


app/celery/provider_tasks.py
```diff
@@ -31,10 +33,10 @@ def deliver_sms(self, notification_id):
             )
             self.retry(queue=QueueNames.RETRY)
         except self.MaxRetriesExceededError:
-            current_app.logger.exception(
-                "RETRY FAILED: task send_sms_to_provider failed for notification {}".format(notification_id),
-            )
-            update_notification_status_by_id(notification_id, 'technical-failure')
+            message = "RETRY FAILED: task send_sms_to_provider failed for notification {}. " \
```
richardc0: Should this message message that its a Max Retries Failure not just a Retry Failure?

app/celery/provider_tasks.py
```diff
@@ -55,7 +57,7 @@ def deliver_email(self, notification_id):
             )
             self.retry(queue=QueueNames.RETRY)
         except self.MaxRetriesExceededError:
-            current_app.logger.error(
-                "RETRY FAILED: task send_email_to_provider failed for notification {}".format(notification_id)
-            )
-            update_notification_status_by_id(notification_id, 'technical-failure')
+            message = "RETRY FAILED: task send_email_to_provider failed for notification {}. " \
```
richardc0: Same for this one?

app/notifications/notifications_ses_callback.py
```diff
@@ -93,8 +96,10 @@ def remove_emails_from_bounce(bounce_dict):
...
 
-def _check_and_queue_callback_task(notification_id, service_id):
+def _check_and_queue_callback_task(notification):
     # queue callback task only if the service_callback_api exists
-    service_callback_api = get_service_callback_api_for_service(service_id=service_id)
+    service_callback_api = get_service_callback_api_for_service(service_id=notification.service_id)
     if service_callback_api:
-        send_delivery_status_to_service.apply_async([str(notification_id)], queue=QueueNames.CALLBACKS)
+        encrypted_notification = create_encrypted_callback_data(notification, service_callback_api)
+        send_delivery_status_to_service.apply_async([str(notification.id), encrypted_notification],
```
venusbb: Do we still need notification.id here?

app/celery/service_callback_tasks.py
```diff
@@ -141,3 +141,21 @@ def process_update_with_notification_id(self, notification_id):
                 """Retry: send_delivery_status_to_service has retried the max num of times
                  for notification: {}""".format(notification_id)
             )
+
+
+def create_encrypted_callback_data(notification, service_callback_api):
+    from app import DATETIME_FORMAT, encryption
+    data = {
```
richardc0: Should this be moved into the notifications model - its basically serialising?

app/notifications/notifications_ses_callback.py
venusbb: You also need to change the definition of send_delivery_status_to_service here:
https://github.com/alphagov/notifications-api/blob/master/app/celery/service_callback_tasks.py#L27

app/notifications/process_client_response.py
```diff
@@ -83,7 +85,9 @@ def _process_for_status(notification_status, client_name, reference):
     service_callback_api = get_service_callback_api_for_service(service_id=notification.service_id)
 
     if service_callback_api:
-        send_delivery_status_to_service.apply_async([str(notification.id)], queue=QueueNames.CALLBACKS)
+        encrypted_notification = create_encrypted_callback_data(notification, service_callback_api)
+        send_delivery_status_to_service.apply_async([str(notification.id), encrypted_notification],
+                                                    queue=QueueNames.CALLBACKS)
```
venusbb: This does the same thing as _check_and_queue_callback_task(notification) in notifications_ses_callback. Should we combine them?

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -180,8 +182,10 @@ def test_update_status_of_notifications_after_timeout(notify_api, sample_templat
             status='pending',
             created_at=datetime.utcnow() - timedelta(
                 seconds=current_app.config.get('SENDING_NOTIFICATIONS_TIMEOUT_PERIOD') + 10))
-        timeout_notifications()
-
+        with pytest.raises(NotificationTechnicalFailureException) as e:
+            timeout_notifications()
+        print(e.value.message)
```
richardc0: Remove

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -210,15 +214,17 @@ def test_should_not_update_status_of_letter_notifications(client, sample_letter_
 
 
 def test_timeout_notifications_sends_status_update_to_service(client, sample_template, mocker):
-    create_service_callback_api(service=sample_template.service)
+    callback_api = create_service_callback_api(service=sample_template.service)
```
richardc0: Should add a test for timeout_notifications where there is a technical_failure_notifications and verify that there is an exception thrown

app/notifications/notifications_ses_callback.py
servingUpAces: I need it for backward compatibility.

app/notifications/process_client_response.py
servingUpAces: Perhaps but that will take 2 deploys to remove it. I thought since it doesn't harm anything nor change the functionality I might leave it.

app/celery/provider_tasks.py
servingUpAces: Perhaps - but if the message is marked as technical failure than it has retried the max number of times.  (I'll add max to the message)

tests/app/celery/test_scheduled_tasks.py
servingUpAces: done

tests/app/celery/test_scheduled_tasks.py
servingUpAces: I think that's covered in test_provider_tasks.test_should_go_into_technical_error_if_exceeds_retries_on_deliver_sms_task and test_provider_tasks.test_should_go_into_technical_error_if_exceeds_retries_on_deliver_email_task

app/celery/service_callback_tasks.py
servingUpAces: Probably not, because the service_callback_api stuff is in there too.

PR #1783, opened 2018-03-20
Send task to antivirus app and process antivirus callbacks

kentsanggds:
## What

Adds on to this PR - https://github.com/alphagov/notifications-api/pull/1768

Sends a task to the antivirus app for precompiled letters and handles the callback to either move the file from the scan to process bucket or delete the pdf if it fails the virus scan.

## Dependencies

- [x] https://github.com/alphagov/notifications-aws/pull/337 needs to be applied to all environments to get the moving of files between buckets to work
- [x] needs the antivirus app - https://github.com/alphagov/notifications-antivirus deployed
- [ ] https://github.com/alphagov/notifications-api/pull/1768 can be reviewed first to keep the PR short, any changes to that PR should be merged in otherwise this PR can be reviewed instead.

## Reference 

https://www.pivotaltracker.com/story/show/155323461


app/celery/letters_pdf_tasks.py
```diff
@@ -144,3 +152,42 @@ def letter_in_created_state(filename):
...
+            "There should only be one letter notification for each reference. Found {} notifications".format(
+                updated_count
+            )
+        )
+
+
+@notify_celery.task(name='process-virus-scan-failed')
+def process_virus_scan_failed(filename):
+    current_app.logger.info('Virus scan failed: {}'.format(filename))
+    delete_pdf_from_letters_scan_bucket(filename)
```
leohemsted: i'd rather do this delete last - just so it doesn't run if something else unexpected happens?

app/letters/utils.py
```diff
@@ -74,6 +85,33 @@ def upload_letter_pdf(notification, pdf_data, is_test_letter=False):
...
+def move_scanned_pdf_to_letters_pdf_bucket(filename):
+    source_bucket_name = current_app.config['LETTERS_SCAN_BUCKET_NAME']
+    target_bucket_name = current_app.config['LETTERS_PDF_BUCKET_NAME']
+
+    s3 = boto3.resource('s3')
+    copy_source = {'Bucket': source_bucket_name, 'Key': filename}
+    target_filename = get_folder_name(datetime.utcnow()) + filename
+    target_bucket = s3.Bucket(target_bucket_name)
+    obj = target_bucket.Object(target_filename)
+    obj.copy(copy_source, ExtraArgs={'ServerSideEncryption': 'AES256'})
```
leohemsted: does this move across tags? if so, does the expiry time get refreshed or stay the same?

kentsanggds: Tags get moved across 
The expiry time will get reset when moved to the new bucket, so a pdf that has been in the scan bucket for 3 days will be moved to the process bucket and will have the expiration time reset to 7 days.

PR #1790, opened 2018-03-22
Update inbound sms method

gov-cjwaszczuk:
Adjusted the old inbound sms endpoint to allow for a page to also be given, this gives the flexibility to create a paginated inbox page without also changing the conversation page.

### DIFF
--- a/app/dao/inbound_sms_dao.py
+++ b/app/dao/inbound_sms_dao.py
@@ -32,7 +32,23 @@ def dao_get_inbound_sms_for_service(service_id, limit=None, user_number=None):
     return q.all()
 
 
-def dao_get_paginated_inbound_sms_for_service(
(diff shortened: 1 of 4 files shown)


app/dao/inbound_sms_dao.py
```diff
@@ -32,7 +32,26 @@ def dao_get_inbound_sms_for_service(service_id, limit=None, user_number=None):
...
+        InboundSms.service_id == service_id
+    ).order_by(
+        InboundSms.created_at.desc()
+    )
+
+    if user_number:
+        q = q.filter(InboundSms.user_number == user_number)
+
+    if limit:
+        q = q.limit(limit)
```
leohemsted: we don't want a limit variable if we're using paginate.

PR #1797, opened 2018-03-23
Add antivirus check on precompiled letters sent with test key

kentsanggds:
## What 

Before previewing precompiled PDFs sent with a test key we should run it pass the antivirus scan.

- precompiled PDFs sent by test key uploaded to scan bucket
- set status to `virus-scan-failed` for pdfs failing virus scan rather than `permanent-failure`
- Make call to AV app for precompiled letters sent via a test key, and set notification status to 'pending-virus-scan`
 - Successful scans should set the notification to `delivered`
- Failed scans should be set to `virus-scan-failed` and deleted from the scan bucket

## Reference 

https://www.pivotaltracker.com/story/show/155856605

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -17,16 +17,22 @@ from app.dao.notifications_dao import (
     get_notification_by_id,
     update_notification_status_by_id,
     dao_update_notification,
+    dao_get_notification_by_reference,
     dao_get_notifications_by_references,
     dao_update_notifications_by_reference,
 )
 from app.letters.utils import (
     delete_pdf_from_letters_scan_bucket,
     get_reference_from_filename,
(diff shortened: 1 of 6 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -157,24 +163,26 @@ def letter_in_created_state(filename):
...
+    )
+    update_letter_pdf_status(
+        reference,
+        NOTIFICATION_DELIVERED if is_test_key else NOTIFICATION_CREATED
+    )
 
 
 @notify_celery.task(name='process-virus-scan-failed')
 def process_virus_scan_failed(filename):
     current_app.logger.error('Virus scan failed: {}'.format(filename))
```
richardc0: I think we want this to be an exception in the first instance

kentsanggds: ok

PR #1799, opened 2018-03-23
Update Python version to match PaaS

quis:
We use Python 3.5.4 in production.


sakisv:
Merging this to fix failing jobs.

PR #1800, opened 2018-03-23
Replay emails and sms that are still in created

servingUpAces:
If a sms or email has not been sent after 4 hours and 15 minutes then put it on the delivery queue.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -266,6 +266,11 @@ class Config(object):
             'task': 'daily-stats-template-usage-by-month',
             'schedule': crontab(hour=0, minute=5),
             'options': {'queue': QueueNames.PERIODIC}
+        },
+        'replay_created_notifications': {
+            'task': 'replay_created_notifications',
+            'schedule': crontab(minute=15),
+            'options': {'queue': QueueNames.PERIODIC}
         }
     }
     CELERY_QUEUES = []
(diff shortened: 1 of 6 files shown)


app/config.py
```diff
@@ -279,6 +279,9 @@ class Config(object):
 
     SENDING_NOTIFICATIONS_TIMEOUT_PERIOD = 259200  # 3 days
 
+    RESEND_CREATED_NOTIFICATIONS_OLDER_THAN = (60 * 60 * 4) + (60 * 15)
```
richardc0: Can you add a comment on the maths 4 hours + 15 mins?

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -1191,3 +1192,33 @@ def test_letter_not_raise_alert_if_no_files_do_not_cause_error(mocker, notify_db
...
+                                    status='created')
+    # notifications that are not to be resent
+    create_notification(template=sms_template, created_at=datetime.utcnow() - timedelta(seconds=older_than),
+                        status='sending')
+    create_notification(template=email_template, created_at=datetime.utcnow() - timedelta(seconds=older_than),
+                        status='delivered')
+    create_notification(template=sms_template, created_at=datetime.utcnow(),
+                        status='created')
+    create_notification(template=email_template, created_at=datetime.utcnow(),
+                        status='created')
```
leohemsted: can you add a letter notification into this not resent list?

servingUpAces: I need to do something different for letters - that is coming in a separate PR.

PR #1801, opened 2018-03-23
New command to create-pdf-letter task for a given notification id.

servingUpAces:
After a notificaiton is created we create a task to create the pdf and save it to S3,
if for some reason that task does not run we are left with notifications that are not sent.
This should not happen, but if it does we have a way to continue sending the letter.


app/commands.py
```diff
@@ -316,6 +317,14 @@ def insert_inbound_numbers_from_file(file_name):
     file.close()
 
 
+@notify_command(name='replay-create-pdf-letters')
+@click.option('-n', '--notification_id', required=True,
+              help="Notification id of the letter that needs the create_letters_pdf task replayed")
```
leohemsted: is it worth having `type=click.UUID` here to do validation?

note, would then need to cast back to a str in the apply_async.

servingUpAces: done

PR #1802, opened 2018-03-23
Added a new task to handle any error cases with the anti-virus

richardc0:
Added a new task to handle any error cases with the anti-virus application. If the Anti-virus app fails due to s3 errors or ClamAV so does not scan (even after retries) the file at all an error needs to be raised and the notification set to technical-failure.

Files should be moved to a 'folder' a separate one for ERROR and FAILURE.

* Added new letter task to process the error
* Added a new method to letter utils.py to move a file into an error or failure folder based on the input
* Added tests to test the task and the utils.py method

### DIFF
--- a/app/letters/utils.py
+++ b/app/letters/utils.py
@@ -104,10 +104,7 @@ def move_scanned_pdf_to_test_or_live_pdf_bucket(source_filename, is_test_letter=
 def move_failed_pdf(source_filename, scan_error_type):
     scan_bucket = current_app.config['LETTERS_SCAN_BUCKET_NAME']
 
-    if scan_error_type == ScanErrorType.ERROR:
-        target_filename = 'ERROR/' + source_filename
-    elif scan_error_type == ScanErrorType.FAILURE:
-        target_filename = 'FAILURE/' + source_filename
+    target_filename = ('ERROR/' if scan_error_type == ScanErrorType.ERROR else 'FAILURE/') + source_filename
 
     _move_s3_object(scan_bucket, source_filename, scan_bucket, target_filename)
 
--- a/app/letters/utils.py
+++ b/app/letters/utils.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+from enum import Enum
 
 import boto3
 from flask import current_app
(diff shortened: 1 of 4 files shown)


app/letters/utils.py
```diff
@@ -117,6 +123,46 @@ def delete_pdf_from_letters_scan_bucket(filename):
     current_app.logger.info("Deleted letter PDF: {}/{}".format(bucket_name, filename))
 
 
+def move_failed_pdf(filename, scan_error_type):
```
kentsanggds: Code in the function should be refactored so we are not duplicating code, could even update the existing `move_scanned_pdf_to_test_or_live_pdf_bucket` function to pass an additional argument for the `scan_error_type`

app/letters/utils.py
```diff
@@ -117,6 +123,46 @@ def delete_pdf_from_letters_scan_bucket(filename):
...
+    # in the destination bucket the expiration time will be reset to 7 days left to expire
+    obj.copy(copy_source, ExtraArgs={'ServerSideEncryption': 'AES256'})
+
+    s3.Object(bucket_name, filename).delete()
+
+    current_app.logger.info("Moved letter PDF: {}/{} to {}/{}".format(
+        bucket_name, filename, bucket_name, target_filename))
+
+
+def move_pdf_from_letters_scan_bucket(filename, scan_error_type):
```
kentsanggds: I don't think this function is being used anymore?

app/letters/utils.py
```diff
@@ -85,36 +91,25 @@ def upload_letter_pdf(notification, pdf_data):
...
-def delete_pdf_from_letters_scan_bucket(filename):
-    bucket_name = current_app.config['LETTERS_SCAN_BUCKET_NAME']
+def move_failed_pdf(source_filename, scan_error_type):
+    scan_bucket = current_app.config['LETTERS_SCAN_BUCKET_NAME']
 
-    s3 = boto3.resource('s3')
-    s3.Object(bucket_name, filename).delete()
+    if scan_error_type == ScanErrorType.ERROR:
+        target_filename = 'ERROR/' + source_filename
+    elif scan_error_type == ScanErrorType.FAILURE:
```
kentsanggds: Should this be an else? What happens if the `ScanErrorType` is not matched? 
You can also write like -

```
target_filename = ('ERROR/' if ScanErrorType.ERROR else 'FAILURE/') + source_filename
```

PR #1806, opened 2018-03-26
Update logging for provider tasks.

servingUpAces:
Move the info message before the fetch.
Include the exception in the log message.

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -22,15 +22,15 @@ def worker_process_shutdown(sender, signal, pid, exitcode):
 @statsd(namespace="tasks")
 def deliver_sms(self, notification_id):
     try:
+        current_app.logger.info("Start sending SMS for notification id: {}".format(notification_id))
         notification = notifications_dao.get_notification_by_id(notification_id)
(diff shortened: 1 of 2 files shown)


app/celery/provider_tasks.py
```diff
@@ -22,15 +22,15 @@ def worker_process_shutdown(sender, signal, pid, exitcode):
...
         notification = notifications_dao.get_notification_by_id(notification_id)
         if not notification:
             raise NoResultFound()
-        current_app.logger.info("Start sending SMS for notification id: {}".format(notification_id))
         send_to_providers.send_sms_to_provider(notification)
     except Exception as e:
         try:
             current_app.logger.exception(
-                "SMS notification delivery for id: {} failed".format(notification_id)
+                "SMS notification delivery for id: {} failed".format(notification_id), e
```
leohemsted: Any reason to include the exception in the log message here? - by calling `logger.exception` the exception's already included (along with the stack trace) in exc_info

servingUpAces: Ah - the exception was from the provider, it just didn't have enough information.

PR #1807, opened 2018-03-27
Billing database tweak and BST bug fix

venusbb:
Pivotal story:  https://www.pivotaltracker.com/story/show/155511547

# create-nightly-billing
* now runs from n-3 to n-1 days (total 3 days) inclusively where n=today 
* now do migration day by day with commit for each day
* The day is defined with BST time rather than UTC time

# commands.py
* use bst time to define a day
* process specified start day to end day exclusive of the end day - to allow the command to run for just 1 day. 
* Additional print informations

# DB Migration:
* Reinsert data based into dm_datetime using 24:00:00 to defend utc_day_end so that there is  no gap in time for each bst_date
* Define primary key also using service_id column for defining a unique record

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -10,10 +10,11 @@ from app.models import (Notification,
                         LETTER_TYPE, SMS_TYPE)
 from app import db
 from sqlalchemy import func, desc, case
-from app.dao.dao_utils import transactional
 from notifications_utils.statsd_decorators import statsd
 from app import notify_celery
 from flask import current_app
+from app.utils import convert_bst_to_utc
+from dateutil import parser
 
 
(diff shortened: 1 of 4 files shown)


app/celery/reporting_tasks.py
```diff
@@ -28,91 +29,94 @@ def get_rate(non_letter_rates, letter_rates, notification_type, date, crown=None
...
-        Notification.international,
-        Service.crown
-    ).join(
-        Service
-    ).order_by(
-        'day_created'
-    ).all()
+    for i in range(0, 3):
+        process_day = day_start - timedelta(days=i)
+        ds = convert_bst_to_utc(parser.parse("{:%Y-%m-%d}".format(process_day) + ' 00:00:00'))
```
allait: You could use datetime.combine to make this a bit simpler:
```python
from datetime import datetime, time

ds = convert_bst_to_utc(datetime.combine(process_date, time.min))
```

PR #1811, opened 2018-03-28
Add http health-check for the api app

sakisv:
Also rename the 'elb' argument to 'simple' to be more descriptive.

The healthcheck is going to issue requests to `/_status?simple=true`
just to check that gunicorn is up and running - we don't need to go all
the way to the database, as this would have adverse impact when api is
under load.

### Depends on: #1809


sakisv:
Nope

PR #1817, opened 2018-04-03
Logging refactor to make debugging Antivirus scans easier

richardc0:
Before, the filename needed to be known. Added the notification id to the logging message so that
the notification can be traced through the logging system by knowing the notification id, making it easier to debug. Also changed to raise an exception so that alerts are generated. This way we should get an email to say that there has been an error.

### DIFF
--- a/app/errors.py
+++ b/app/errors.py
@@ -10,6 +10,12 @@ from jsonschema import ValidationError as JsonSchemaValidationError
 from app.authentication.auth import AuthError
 
 
+class VirusScanError(Exception):
+    def __init__(self, message):
+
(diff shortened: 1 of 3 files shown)


app/errors.py
```diff
@@ -10,6 +10,12 @@
 from app.authentication.auth import AuthError
 
 
+class VirusScanError(Exception):
+    def __init__(self, message):
+
+        super().__init__(message)
```
leohemsted: don't even need this init. can just be 
```
class VirusScanError(Exception):
    pass
```

PR #1820, opened 2018-04-03
Fix research mode preview of precompiled letters

kentsanggds:
## What

When sending precompiled letters using a test key in research mode letter previews were not working.

- additionally also refactored the code so it will not be triggered in `staging` or `live` environments

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -265,12 +265,6 @@ def process_letter_notification(*, letter_data, api_key, template, reply_to_text
                 [str(notification.id)],
                 queue=QueueNames.CREATE_LETTERS_PDF
             )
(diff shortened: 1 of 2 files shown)


app/v2/notifications/post_notifications.py
```diff
@@ -281,6 +275,14 @@ def process_letter_notification(*, letter_data, api_key, template, reply_to_text
                 kwargs={'filename': filename},
                 queue=QueueNames.ANTIVIRUS,
             )
+        elif (
+            api_key.service.research_mode and
+            current_app.config['NOTIFY_ENVIRONMENT'] not in ['staging', 'live']
```
servingUpAces: If you change this, you need to add 'production' to the list. But I am wondering why you made this change? I find it easier to read conditions that are written in the positive.

PR #1822, opened 2018-04-04
Added new paginated inbound endpoint

gov-cjwaszczuk:
Updated paginated inbound sms message endpoint to get the most recent message for 50 users ( rather than the most recent 50 messages). 

Also removed page and limit from old `get_inbound_sms_for_service` method because they were never used.

### DIFF
--- a/app/dao/inbound_sms_dao.py
+++ b/app/dao/inbound_sms_dao.py
@@ -4,7 +4,8 @@ from datetime import (
 )
 from flask import current_app
 from notifications_utils.statsd_decorators import statsd
-from sqlalchemy import desc
+from sqlalchemy import desc, and_
+from sqlalchemy.orm import aliased
 
 from app import db
 from app.dao.dao_utils import transactional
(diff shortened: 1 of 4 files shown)


app/dao/inbound_sms_dao.py
```diff
@@ -93,3 +78,48 @@ def dao_get_inbound_sms_by_id(service_id, inbound_id):
...
         service_id=service_id
     ).one()
+
+
+def dao_get_paginated_most_recent_inbound_sms_by_user_number_for_service(
+    service_id,
+    page
+):
+    """
+    This query starts from inbound_sms and joins on to itself to find the most recent row for each user_number
```
servingUpAces: I'm happy that this query is doing the right thing. I confirmed the results by comparing it with this query:
SELECT distinct(user_number), notify_number, max(created_at)                                                                                                     from inbound_sms                                                                                                                                                                                                  
WHERE service_id = :service_id 
GROUP BY user_number, notify_number 
ORDER BY max(created_at) desc
LIMIT :page_size OFFSET :page

gov-cjwaszczuk:
yep

PR #1823, opened 2018-04-04
Post notification with document upload

allait:
### Add a document download client

Allows uploading documents to the Document Download API. The client is configured with an API host and auth token. There's no need for a flag to disable the client in the test environments at the moment since the upload is only triggered by a specific payload which would only be sent with an explicit goal of using document download.

### Upload files from personalisation data to document download

Adds support for a new personalisation value type: file upload.

File uploads are represented as a dictionary with a "file" key and a base64-encoded file data as the key's value:

``` personalisation={ 'field1': {'file': '<base64-encoded file contents>'} } ```

Post notification endpoint checks the request personalisation data looking for the file uploads in personalisation data. If any are found and the service has permissions to upload documents the files are sent to document download API and personalisation values are replaced with the URLs returned in the document download response.

A fake document URL is returned for simulated notifications, no documents are stored in Document Download.

Multiple files can be uploaded for one notification by providing a file upload in more than one personalisation field.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -138,7 +138,7 @@ def post_notification(notification_type):
             reply_to_text=reply_to
         )
     else:
-        notification, personalisation = process_sms_or_email_notification(
+        notification = process_sms_or_email_notification(
             form=form,
             notification_type=notification_type,
             api_key=api_user,
@@ -147,7 +147,7 @@ def post_notification(notification_type):
             reply_to_text=reply_to
         )
 
-        template_with_content.values = personalisation
+        template_with_content.values = notification.personalisation
 
     if notification_type == SMS_TYPE:
         create_resp_partial = functools.partial(
@@ -216,7 +216,7 @@ def process_sms_or_email_notification(*, form, notification_type, api_key, templ
         else:
             current_app.logger.debug("POST simulated notification for id: {}".format(notification.id))
 
-    return notification, personalisation
+    return notification
 
 
 def process_document_uploads(personalisation_data, service, simulated=False):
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: app/v2/notifications/post_notifications.py [hunk 1]
```diff
@@ -136,7 +137,7 @@ def post_notification(notification_type):
             reply_to_text=reply_to
         )
     else:
-        notification = process_sms_or_email_notification(
+        notification, personalisation = process_sms_or_email_notification(
```
leohemsted: i'm not a huge fan of returning tuples from functions - is there any problem with setting `template_with_content.values = notification.personalisation` further down? the transformed personalisation should be populated onto the notification object right?

tests/app/clients/test_document_download.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_upload_document(document_download):
+    with requests_mock.Mocker() as request_mock:
+        request_mock.post('https://document-download/services/service-id/documents', json={
+            'document': {'url': 'https://document-download/services/service-id/documents/uploaded-url'}
+        }, status_code=201)
+
+        resp = document_download.upload_document('service-id', 'abababab')
+
+    assert resp == 'https://document-download/services/service-id/documents/uploaded-url'
```
leohemsted: is it worth checking that the auth header is put on properly?

app/v2/notifications/post_notifications.py
allait: Yep, good point, haven't thought of using notification attribute

tests/app/clients/test_document_download.py
allait: I've added a request mock header check

PR #1824, opened 2018-04-05
make sure redis command filters dates properly

leohemsted:
also write the tests that i should have written a while ago for this


app/commands.py
```diff
@@ -504,14 +504,21 @@ def populate_redis_template_usage(service_id, day):
     Recalculate and replace the stats in redis for a day.
     To be used if redis data is lost for some reason.
     """
+    # the day variable is set to midnight of that day
     assert current_app.config['REDIS_ENABLED']
```
richardc0: Do we actually want to assert here in app/commands?

app/commands.py
```diff
@@ -504,14 +504,21 @@ def populate_redis_template_usage(service_id, day):
     Recalculate and replace the stats in redis for a day.
     To be used if redis data is lost for some reason.
     """
+    # the day variable is set to midnight of that day
```
richardc0: Does this comment go with the REDIS_ENABLED line? probably move it a couple down?

tests/app/commands/test_populate_redis.py
```diff
@@ -0,0 +1,69 @@
...
+from app.commands import populate_redis_template_usage
+
+from tests.conftest import set_config
+from tests.app.db import create_notification, create_template, create_service
+
+
+def test_populate_redis_template_usage_does_nothing_if_redis_disabled(mocker, notify_api, sample_service):
+    mock_redis = mocker.patch('app.commands.redis_store')
+    with set_config(notify_api, 'REDIS_ENABLED', False):
+        with pytest.raises(AssertionError):
```
richardc0: Should we be throwing an exception here. 

Redis being turned off is not exceptional circumstances? If it is RedisConfigError?

tests/app/commands/test_populate_redis.py
```diff
@@ -0,0 +1,69 @@
...
+            populate_redis_template_usage.callback.__wrapped__(sample_service.id, datetime.utcnow())
+
+    assert not mock_redis.called
+
+
+def test_populate_redis_template_usage_does_nothing_if_no_data(mocker, notify_api, sample_service):
```
richardc0: Maybe rename? It does do something because redis is being called?

does not store data if there is nothing to store?

tests/app/commands/test_populate_redis.py
```diff
@@ -0,0 +1,69 @@
...
+    mock_redis = mocker.patch('app.commands.redis_store')
+    with set_config(notify_api, 'REDIS_ENABLED', True):
+        populate_redis_template_usage.callback.__wrapped__(sample_service.id, datetime.utcnow())
+
+    assert not mock_redis.called
```
richardc0: Test its called with the right parameters because this could be called is there is data?

leohemsted: https://www.songkick.com/concerts/33124154-65daysofstatic-at-queen-elizabeth-hall-southbank-centre?utm_source=8123&utm_medium=partner&utm_content=d16deb28b60332f7aed0a29297a35950&utm_campaign=entity

leohemsted: no - this test has no data in the DB, so redis isn't being called.

That said, I do think this isn't the right call. i should make sure none of redis's child objects (the actual set fns) are being called.

tests/app/commands/test_populate_redis.py
leohemsted: it's turned off locally.

app/commands.py
leohemsted: changed to if statement and `sys.exit(1)`

PR #1833, opened 2018-04-13
Read template usage stats from new redis keys

leohemsted:
## Reading template usage stats from new redis tables

The new flow is as follows:

For each day:
* get stats from redis
* if there's nothing in redis, get stats from the db
  - if there was data in the DB (even a count of 0) put that in redis so we don't need to access it again

Merge stats for each day and return those.

### Lots of cleanup

* Cleaned up filtering.
  - You now _need_ to pass a limit_days in to the REST endpoint. (The front-end always passes in limit_days=7
  - the dao function now only accepts a single day. Lots of testing and refactoring to make this timezone aware too.
* No longer ever query the NotificationHistory table from this endpoint
* made the `days_ago` function return a BST midnight rather than a timezone-unaware date - a couple of tweaks to other functions to use this. Also moved it to utils next to the other date functions
* i hate timezones now

### Lots of tests

* re-wrote all the tests as they're pretty much completely different now
* moved the notification dao ones into their own file to break up the mega test file
* look at all those timezone tests!


app/utils.py
```diff
@@ -90,3 +93,25 @@ def get_public_notify_type_text(notify_type, plural=False):
         notify_type_text = 'text message'
 
     return '{}{}'.format(notify_type_text, 's' if plural else '')
+
+
+def days_ago(number_of_days):
```
richardc0: So would 0 give me 00:00 today?

richardc0: Suggest a rename because days ago implies the date?

midnight_n_days_ago

tests/app/dao/notification_dao/test_notification_dao_template_usage.py
```diff
@@ -0,0 +1,167 @@
...
+
+def test_should_by_able_to_get_template_count(sample_template, sample_email_template):
+    create_notification(sample_template)
+    create_notification(sample_template)
+    create_notification(sample_template)
+    create_notification(sample_email_template)
+    create_notification(sample_email_template)
+
+    results = dao_get_template_usage(sample_template.service_id, date.today())
+    assert results[0].name == 'Email Template Name'
```
richardc0: could use results[0].name == sample_email_template.name etc

tests/app/dao/notification_dao/test_notification_dao_template_usage.py
```diff
@@ -0,0 +1,167 @@
...
+    assert results.id == team_key.id
+
+
+def test_last_template_usage_should_be_able_to_get_no_template_usage_history_if_no_notifications_using_template(
+        sample_template):
+    results = dao_get_last_template_usage(sample_template.id, 'sms')
+    assert not results
+
+
+def test_should_by_able_to_get_template_count(sample_template, sample_email_template):
```
richardc0: freeze this?

Technically the test could run and create some notifications in a previous day

tests/app/dao/notification_dao/test_notification_dao_template_usage.py
```diff
@@ -0,0 +1,167 @@
...
+    assert results[0].name == 'Email Template Name'
+    assert results[0].template_type == 'email'
+    assert results[0].count == 2
+
+    assert results[1].name == 'Template Name'
+    assert results[1].template_type == 'sms'
+    assert results[1].count == 3
+
+
+def test_template_usage_should_ignore_test_keys(
```
richardc0: freezetime

tests/app/dao/notification_dao/test_notification_dao_template_usage.py
```diff
@@ -0,0 +1,167 @@
...
+    assert res3[1].count == 0
+
+
+def test_template_usage_should_by_able_to_get_zero_count_from_notifications_history_if_no_rows(sample_service):
+    results = dao_get_template_usage(sample_service.id, date.today())
+    assert len(results) == 0
+
+
+def test_template_usage_should_by_able_to_get_zero_count_from_notifications_history_if_no_service():
+    results = dao_get_template_usage(str(uuid.uuid4()), date.today())
```
richardc0: Should this actually be an exception? Can this case happen regularly?

tests/app/dao/notification_dao/test_notification_dao_template_usage.py
```diff
@@ -0,0 +1,167 @@
...
+    assert len(results) == 0
+
+
+@freeze_time('2017-06-10T12:00:00')
+def test_template_usage_should_by_able_to_get_template_count_for_specific_day(sample_template):
```
richardc0: How about a new test on the bst boundary?

tests/app/test_utils.py
```diff
@@ -44,3 +49,34 @@ def test_convert_bst_to_utc():
...
+
+    # summer
+    ('2018-06-05 10:00', datetime(2018, 6, 3, 23, 0))
+])
+def test_days_ago(current_time, expected_datetime):
+    with freeze_time(current_time):
+        assert days_ago(1) == expected_datetime
+
+
+def test_last_n_days():
```
richardc0: Should we test last_n_days = 0 and -1?

tests/app/test_utils.py
```diff
@@ -44,3 +49,34 @@ def test_convert_bst_to_utc():
...
+
+    # bst switchover at 1am 25th
+    ('2018-03-25 10:00', datetime(2018, 3, 24, 0, 0)),
+    ('2018-03-26 10:00', datetime(2018, 3, 25, 0, 0)),
+    ('2018-03-27 10:00', datetime(2018, 3, 25, 23, 0)),
+
+    # summer
+    ('2018-06-05 10:00', datetime(2018, 6, 3, 23, 0))
+])
+def test_days_ago(current_time, expected_datetime):
```
richardc0: How about a time on the bst boundary?

app/template_statistics/rest.py
```diff
@@ -70,31 +58,45 @@ def get_template_statistics_for_template_id(service_id, template_id):
...
-    #     # "<SERVICE_ID>-template-usage-{YYYY-MM-DD}"
-    #     key = cache_key_for_service_templates_used_per_day(service_id, limit_days)
-    #     try:
-    #         template_stats_by_id += Counter(redis_store.get_all_from_hash(key, raise_exception=True))
-    #     except RedisException:
-    #         # TODO: ????
-    #
-    # # TODO: streamline db query and avoid weird unions if possible.
-    # return dao_get_templates_for_cache(template_stats_by_id.items())
+def get_template_statistics_for_last_n_days(service_id, limit_days):
```
richardc0: No tests for this method?

app/template_statistics/rest.py
```diff
@@ -27,31 +29,17 @@
 
 @template_statistics.route('')
 def get_template_statistics_for_service_by_day(service_id):
```
richardc0: As there is new logic here should there be some more tests added?

app/utils.py
leohemsted: yes. added a test

tests/app/dao/notification_dao/test_notification_dao_template_usage.py
leohemsted: I don't expect it to ever happen, but not sure it's worth putting extra logic in to raise an exception. i  view this test as more tests-as-documentation than risk management or anything.

tests/app/test_utils.py
leohemsted: that's already there

app/template_statistics/rest.py
leohemsted: i viewed them as being part of of the `get_template_statistics_for_service_by_day` tests

app/template_statistics/rest.py
leohemsted: there have been more tests added. lots of them. all of https://github.com/alphagov/notifications-api/blob/a166a16c096b5570c4fe546b05a3153adbd6ad97/tests/app/template_statistics/test_rest.py#L30-L246 are tests for this function

leohemsted:
@richardc0 addressed all the comments i think

PR #1838, opened 2018-04-24
Allow reply to addresses to be made inactive

klssmith:
It should be possible to delete the three types of reply-to addresses:
* `service_email_reply_to`
* `service_sms_sender`
* `service_letter_contact`

It isn't possible to hard delete letter contacts because they are referenced by the Template table, so to keep the ways things work consistent between all notification types, we mark the reply-to addresses as inactive instead.

#### Updated DAO functions to return reply-to addresses
The DAO functions to return one and all reply-to addresses now only return the active ones.

#### Added DAO functions that delete the reply-to addresses
Added a DAO function for each notification type which sets the reply-to address to inactive. There are various edge cases that need to be accounted for:
- You shouldn't be able to delete a default reply-to address
- For SMS, you shouldn't be able to delete an inbound number
- In addition to having a default letter contact block for a service, letters can also have default letter contact block for a template - template defaults should not be able to be deleted

### Pivotal story
https://www.pivotaltracker.com/story/show/156476018

### DIFF
--- a/app/exceptions.py
+++ b/app/exceptions.py
@@ -6,3 +6,7 @@ class DVLAException(Exception):
 class NotificationTechnicalFailureException(Exception):
     def __init__(self, message):
         self.message = message
+
+
+class ValidationError(Exception):
+    pass
--- a/tests/app/dao/test_service_email_reply_to_dao.py
+++ b/tests/app/dao/test_service_email_reply_to_dao.py
@@ -25,6 +25,23 @@ def test_dao_get_reply_to_by_service_id(notify_db_session):
     assert second_reply_to == results[2]
 
 
+def test_dao_get_reply_to_by_service_id_only_returns_active_reply_tos(notify_db_session):
+    service = create_service()
+    create_reply_to_email(service=service, email_address='something@email.com')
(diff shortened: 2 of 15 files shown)


app/exceptions.py
```diff
@@ -6,3 +6,7 @@ def __init__(self, message):
 class NotificationTechnicalFailureException(Exception):
     def __init__(self, message):
         self.message = message
+
+
+class ValidationError(Exception):
```
servingUpAces: When I first saw this I thought it was the jsonschema ValidationError. Can we give it a different name?

tests/app/dao/test_service_email_reply_to_dao.py
```diff
@@ -151,6 +173,33 @@ def test_update_reply_to_email_address_raises_exception_if_single_reply_to_and_s
                                       is_default=False)
 
 
+def test_set_reply_to_inactive_deletes_reply_to(sample_service):
```
servingUpAces: Strictly speaking the reply_to is inactive and not deleted.

klssmith:
Closing this and replacing with https://github.com/alphagov/notifications-api/pull/1847

PR #1839, opened 2018-04-24
Adjust get_service_stats to get counts for a maximum of 7 days ago

gov-cjwaszczuk:
### Summary
Previously, the dashboard totals on the main page and activity page were gathering counts based off all the notifications in the table. This PR filters the amount of notifications gathered so that the count is only based off the previous 7 days, inclusive of all hours on the 7th day due to only checking on day rather than hour. 
### Changes
``` dao_fetch_stats_for_service ``` has an extra filter to filter out notifications with a created at older than 7 days from today.
### Tests 
- Test to ensure the notifications are correctly counted, notifcations with createdAts set to:
  - Current time
  - 7 days ago
  - 7 days 2 hours ago (To ensure it checks for the entire day) 
  - 8 days ago

### DIFF
--- a/tests/app/dao/test_services_dao.py
+++ b/tests/app/dao/test_services_dao.py
@@ -625,6 +625,29 @@ def test_fetch_stats_for_today_only_includes_today(notify_db, notify_db_session,
     assert stats['created'] == 1
 
 
+def test_fetch_stats_should_not_gather_notifications_older_than_7_days(notify_db, notify_db_session, sample_template):
+
+    # 8 days ago
+    create_notification(notify_db, None, to_field='1', status='delivered', created_at='2001-01-01T12:00:00')
+
+    # 7 days ago, 2hours ago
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/app/dao/test_services_dao.py [hunk 1]
```diff
@@ -625,6 +625,31 @@ def test_fetch_stats_for_today_only_includes_today(notify_db, notify_db_session,
     assert stats['created'] == 1
 
 
+def test_fetch_stats_should_not_gather_notifications_older_than_7_days(notify_db, notify_db_session, sample_template):
+    with freeze_time('2001-01-01T12:00:00'):
+        # 8 days ago
+        create_notification(notify_db, None, to_field='1', status='delivered')
```
servingUpAces: If you use the function that is aliased create_notification_db you can pass in a created_at date and not need to use freeze_time

PR #1844, opened 2018-04-25
zendesk instead of deskpro

leohemsted:
(when letters are still stuck in sending)


app/celery/scheduled_tasks.py
```diff
@@ -513,25 +511,29 @@ def letter_raise_alert_if_no_ack_file_for_zip():
...
 
-    deskpro_message = "Letter ack file does not contains all zip files sent. " \
-                      "Missing ack for zip files: {}, " \
-                      "pdf bucket: {}, subfolder: {}, " \
-                      "ack bucket: {}".format(str(sorted(zip_file_set - ack_content_set)),
-                                              current_app.config['LETTERS_PDF_BUCKET_NAME'],
-                                              datetime.utcnow().strftime('%Y-%m-%d') + '/zips_sent',
-                                              current_app.config['DVLA_RESPONSE_BUCKET_NAME'])
+    message = (
+        "Letter ack file does not contains all zip files sent. "
```
allait: `does not contain` while we're here

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -646,14 +646,14 @@ def test_alert_if_letter_notifications_still_sending(sample_letter_template, moc
...
+    mock_create_ticket = mocker.patch("app.celery.scheduled_tasks.zendesk_client.create_ticket")
 
     raise_alert_if_letter_notifications_still_sending()
 
-    mock_celery.assert_called_once_with(
+    mock_create_ticket.assert_called_once_with(
         subject="[test] Letters still sending",
         message="There are 1 letters in the 'sending' state from Monday 15 January",
-        ticket_type='alert'
+        ticket_type='incident'
```
allait: It feels like the `zendesk_client.TYPE_INCIDENT` abstraction is leaking here. I think you could use the constant here as well.

PR #1845, opened 2018-04-25
Use ft billing for usage page

servingUpAces:
Refactor code in reporting_tasks to create re-usable methods. Most of the changes in this PR are for that refactor and the addition of tests for the new methods. 

New query to get billing data for a month, since we are missing todays data there is some code to get the deltas and update/insert data in ft_billing. This new code is only being called in tests. The new endpoint is still coming, but this PR is large enough to get a review and push through as is.

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -1,9 +1,16 @@
 from datetime import datetime, timedelta
-from sqlalchemy import func
+from sqlalchemy import func, case, desc, extract
 
 from app import db
 from app.dao.date_util import get_month_start_and_end_date_in_utc, get_financial_year
-from app.models import FactBilling
+from app.models import (
+    FactBilling, Notification, Service, NOTIFICATION_CREATED, NOTIFICATION_TECHNICAL_FAILURE,
+    KEY_TYPE_TEST,
+    LETTER_TYPE,
+    SMS_TYPE,
+    Rate,
+    LetterRate
+)
 from app.utils import convert_utc_to_bst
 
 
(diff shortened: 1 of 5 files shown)


app/dao/fact_billing_dao.py
```diff
@@ -0,0 +1,165 @@
...
+        FactBilling.service_id,
+        FactBilling.rate,
+        FactBilling.rate_multiplier,
+        FactBilling.international
+    ).filter(
+        FactBilling.service_id == service_id,
+        FactBilling.bst_date >= year_start_date,
+        FactBilling.bst_date <= year_end_date
+    ).group_by(
+        extract('month', FactBilling.bst_date),
```
allait: I think you can use the label name from the `.query` call here - `"Month"`.

app/dao/fact_billing_dao.py
```diff
@@ -0,0 +1,165 @@
...
+    KEY_TYPE_TEST,
+    LETTER_TYPE,
+    SMS_TYPE,
+    Rate,
+    LetterRate
+)
+from app.utils import convert_utc_to_bst, convert_bst_to_utc
+
+
+def fetch_annual_billing_for_year(service_id, year):
```
allait: `fetch_monthly_billing_for_year`?

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -0,0 +1,233 @@
...
+def test_fetch_billing_data_for_today_includes_data_with_the_right_status(notify_db_session):
+    service = create_service()
+    template = create_template(service=service, template_type="email")
+    for status in ['delivered', 'sending', 'temporary-failure', 'created', 'technical-failure']:
+        create_notification(template=template, status=status)
+
+    today = convert_utc_to_bst(datetime.utcnow())
+    results = fetch_billing_data_for_day(today)
+    assert len(results) == 1
+    assert results[0].notifications_sent == 3
```
allait: This feels a bit indirect since we're not checking the statuses that end up in the result. Maybe split this in two parts - one tat checks the count after creating all successful statuses and one that checks for 0 after creating all filtered ones.

PR #1848, opened 2018-04-26
Added a command to rebuild the data for a given service_id and date

servingUpAces:
This will allow us to rebuild the data if there is an issue with the data that is older than the past 3 days.

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -17,6 +17,7 @@ from app.celery.scheduled_tasks import send_total_sent_notifications_to_performa
 from app.celery.service_callback_tasks import send_delivery_status_to_service
 from app.celery.letters_pdf_tasks import create_letters_pdf
 from app.config import QueueNames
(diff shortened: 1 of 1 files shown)


app/commands.py
```diff
@@ -545,3 +547,18 @@ def populate_redis_template_usage(service_id, day):
             current_app.config['EXPIRE_CACHE_EIGHT_DAYS'],
             raise_exception=True
         )
+
+
+@notify_command(name='rebuild-ft-billing-for-month-and-service')
+@click.option('-s', '--service_id', required=True, type=click.UUID)
+@click.option('-d', '--day', help="The date to recalculate, as YYYY-MM-DD", required=True,
+              type=click_dt(format='%Y-%m-%d'))
+def rebuild_ft_billing_for_month_and_service(service_id, day):
```
richardc0: Date or day? and might be worth adding a comment with an example?

servingUpAces: It recalculates a whole day (using BST)

PR #1855, opened 2018-05-01
Only get inbound messages that are a maximum of 7 days old

gov-cjwaszczuk:
### Summary
Inbound messages were showing if they were older than 7 days, dependant on celery beat to remove them, this PR changes the way that the stats are retrieved, filtering anything older than 7 days within the query.
### Changes
``` dao_get_inbound_sms_for_service ``` has an extra filter to filter out notifications with a created at older than 7 days from today.
``` dao_count_inbound_sms_for_service ``` has an extra filter to filter out notifications with a created at older than 7 days from today.
``` dao_get_paginated_most_recent_inbound_sms_by_user_number_for_service ``` has an extra filter to filter out notifications with a created at older than 7 days from today.
### Tests 
- Fixed tests that weren't reliant on time to determine counts
- Tested 3 methods to ensure that inbound_sms messages outside of the 7 day period are not gathered
### My concerns
- ``` test_most_recent_inbound_sms_only_returns_values_within_7_days ``` Returns the date in the opposite order I think?

### DIFF
--- a/app/dao/inbound_sms_dao.py
+++ b/app/dao/inbound_sms_dao.py
@@ -1,6 +1,7 @@
 from datetime import (
     timedelta,
-    datetime
+    datetime,
+    date
 )
 from flask import current_app
 from notifications_utils.statsd_decorators import statsd
@@ -10,6 +11,7 @@ from sqlalchemy.orm import aliased
 from app import db
 from app.dao.dao_utils import transactional
 from app.models import InboundSms
+from app.utils import get_london_midnight_in_utc
 
 
 @transactional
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: app/dao/inbound_sms_dao.py [hunk 1]
```diff
@@ -18,8 +20,10 @@ def dao_create_inbound_sms(inbound_sms):
 
 
 def dao_get_inbound_sms_for_service(service_id, limit=None, user_number=None):
+    start_date = get_london_midnight_in_utc(date.today() - timedelta(days=6))
     q = InboundSms.query.filter(
-        InboundSms.service_id == service_id
+        InboundSms.service_id == service_id,
+        func.date(InboundSms.created_at) >= start_date
```
leohemsted: don't call `func.date` here (or anywhere else in this PR). The `start_date` might be, for example, `2018-05-01T23:00:00` (midnight tonight), and sms created both at 12:00 and 23:30 (today and tomorrow respectively) will both have a date of `2018-05-01T00:00:00`.

You should be able to write a test that proves this as well.

leohemsted: see https://github.com/alphagov/notifications-api/pull/1858 for details of what to change

PR #1863, opened 2018-05-04
Use ft billing for usage

servingUpAces:
This PR adds a new endpoint to get yearly usage by month from the ft_billing table. 

If usage is requested for the current year, the deltas added to ft_billing before executing the query to fetch the data. 

There is a command to compare the results of ft_billing to monthly_billing that will give us confidence before switching to the new endpoint on the admin app.

### DIFF
--- a/app/billing/rest.py
+++ b/app/billing/rest.py
@@ -3,6 +3,7 @@ import json
 
 from flask import Blueprint, jsonify, request
 
+from app.dao.fact_billing_dao import fetch_monthly_billing_for_year
 from app.dao.monthly_billing_dao import (
     get_billing_data_for_financial_year,
     get_monthly_billing_by_notification_type
@@ -28,6 +29,16 @@ billing_blueprint = Blueprint(
 
 
 register_errors(billing_blueprint)
+
+
+@billing_blueprint.route('/ft-monthly-usage')
+def get_yearly_usage_by_monthy_from_ft_billing(service_id):
+    try:
+        year = int(request.args.get('year'))
+        results = fetch_monthly_billing_for_year(service_id=service_id, year=year)
+        serialize_ft_billing(results)
+    except TypeError:
+        return jsonify(result='error', message='No valid year provided'), 400
 
 
 @billing_blueprint.route('/monthly-usage')
@@ -192,3 +203,21 @@ def update_free_sms_fragment_limit_data(service_id, free_sms_fragment_limit, fin
             free_sms_fragment_limit,
             financial_year_start
         )
+
+
+def serialize_ft_billing(data):
+    results = []
+
+    for d in data:
+        j = {
+            "Month": d.month,
+            "service_id": d.service_id,
+            "notifications_type": d.notification_type,
+            "notifications_sent": d.notifications_sent,
+            "billable_units": d.billable_units,
+            "rate": d.rate,
+            "rate_multiplier": d.rate_multiplier,
+            "international": d.international,
+        }
+        results.append(j)
+    return results
--- a/app/billing/rest.py
+++ b/app/billing/rest.py
@@ -1,25 +1,30 @@
-from datetime import datetime
 import json
+from datetime import datetime
 
 from flask import Blueprint, jsonify, request
 
+from app.billing.billing_schemas import (
+    create_or_update_free_sms_fragment_limit_schema,
+    serialize_ft_billing_remove_emails
+)
+from app.dao.annual_billing_dao import (
+    dao_get_free_sms_fragment_limit_for_year,
+    dao_get_all_free_sms_fragment_limit,
+    dao_create_or_update_annual_billing_for_year,
+    dao_update_annual_billing_for_future_years
+)
+from app.dao.date_util import get_current_financial_year_start_year
+from app.dao.date_util import get_months_for_financial_year
 from app.dao.fact_billing_dao import fetch_monthly_billing_for_year
 from app.dao.monthly_billing_dao import (
     get_billing_data_for_financial_year,
     get_monthly_billing_by_notification_type
 )
-from app.dao.date_util import get_months_for_financial_year
+from app.errors import InvalidRequest
 from app.errors import register_errors
 from app.models import SMS_TYPE, EMAIL_TYPE, LETTER_TYPE
-from app.utils import convert_utc_to_bst
-from app.dao.annual_billing_dao import (dao_get_free_sms_fragment_limit_for_year,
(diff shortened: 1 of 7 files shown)


app/billing/rest.py
```diff
@@ -192,3 +204,17 @@ def update_free_sms_fragment_limit_data(service_id, free_sms_fragment_limit, fin
             free_sms_fragment_limit,
             financial_year_start
         )
+
+
+def serialize_ft_billing(data):
+    results = []
+    no_emails = [x for x in data if x.notification_type != 'email']
+    for d in no_emails:
```
richardc0: Consider rename x, d and j to be more meaningful
x= notification
j = result
d = email?

app/billing/rest.py
```diff
@@ -192,3 +204,17 @@ def update_free_sms_fragment_limit_data(service_id, free_sms_fragment_limit, fin
             free_sms_fragment_limit,
             financial_year_start
         )
+
+
+def serialize_ft_billing(data):
+    results = []
+    no_emails = [x for x in data if x.notification_type != 'email']
```
richardc0: Consider updating to notifications?

app/billing/rest.py
```diff
@@ -192,3 +204,17 @@ def update_free_sms_fragment_limit_data(service_id, free_sms_fragment_limit, fin
             free_sms_fragment_limit,
             financial_year_start
         )
+
+
+def serialize_ft_billing(data):
```
richardc0: Is this the best place for this? I'd usually but it in a business model but we don't have one?
The other place would be the DAO but that also feels wrong?

app/commands.py
```diff
@@ -562,3 +564,39 @@ def rebuild_ft_billing_for_month_and_service(service_id, day):
...
+                ft_billing_response = get_yearly_usage_by_monthly_from_ft_billing(service_id)
+            compare_monthly_billing_to_ft_billing(ft_billing_response, monthly_billing_response)
+    else:
+        with current_app.test_request_context(
+                path='/service/{}/billing/monthly-usage?year={}'.format(service_id, year)):
+            monthly_billing_response = get_yearly_usage_by_month(service_id)
+        with current_app.test_request_context(
+                path='/service/{}/billing/ft-monthly-usage?year={}'.format(service_id, year)):
+            ft_billing_response = get_yearly_usage_by_monthly_from_ft_billing(service_id)
+        compare_ft_billing_to_monthly_billing(ft_billing_response, monthly_billing_response)
```
richardc0: This should be compare_monthly_billing_to_ft_billing?

Is that path being tested?

tests/app/billing/test_billing.py
```diff
@@ -409,3 +410,148 @@ def test_update_free_sms_fragment_limit_data(client, sample_service):
...
+
+
+def test_get_yearly_usage_by_monthly_from_ft_billing(client, notify_db_session):
+    service = create_service()
+    sms_template = create_template(service=service, template_type="sms")
+    email_template = create_template(service=service, template_type="email")
+    letter_template = create_template(service=service, template_type="letter")
+    for month in range(1, 13):
+        mon = str(month).zfill(2)
+        days_in_month = {1: 32, 2: 30, 3: 32, 4: 31, 5: 32, 6: 31, 7: 32, 8: 32, 9: 31, 10: 32, 11: 31, 12: 32}
```
richardc0: Days in January 32? Technicallt 2 can be 28 and 29

Could you use monthrange(2012, 2) instead here?

tests/app/billing/test_billing.py
```diff
@@ -409,3 +410,148 @@ def test_update_free_sms_fragment_limit_data(client, sample_service):
...
+                             }
+
+    for k in keys:
+        assert ft_sms[0][k] == expected_sms_april[k]
+        assert ft_letters[0][k] == expected_letter_april[k]
+    assert len(ft_email) == 0
+
+
+def test_compare_ft_billing_to_monthly_billing(client, notify_db_session):
+    service = create_service()
```
richardc0: The set-up is similar to the above can you consolidate?

app/commands.py
```diff
@@ -562,3 +564,39 @@ def rebuild_ft_billing_for_month_and_service(service_id, day):
     transit_data = fetch_billing_data_for_day(process_day=day, service_id=service_id)
     for data in transit_data:
         update_fact_billing(data, day)
+
+
+@notify_command(name='compare-ft-billing-to-monthly-billing')
+@click.option('-y', '--year', required=True)
+@click.option('-s', '--service_id', required=False, type=click.UUID)
+def compare_ft_billing_to_monthly_billing(year, service_id=None):
```
richardc0: Should there be unit tests for this?

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -230,7 +248,48 @@ def test_fetch_annual_billing_for_year_adds_data_for_today(notify_db_session):
...
+                              service=service,
+                              template=letter_template,
+                              notification_type='letter',
+                              rate=0.33)
+
+    results = fetch_monthly_billing_for_year(service.id, 2016)
+    # returns 3 rows, per month, returns financial year april to december
+    # Orders by Month
+
+    assert len(results) == 27
```
richardc0: 3 rows per month should be 36?

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -230,7 +248,48 @@ def test_fetch_annual_billing_for_year_adds_data_for_today(notify_db_session):
...
+    assert str(results[0].month) == "2016-04-01 00:00:00+01:00"
+    assert results[0].notification_type == 'email'
+    assert str(results[1].month) == "2016-04-01 00:00:00+01:00"
```
richardc0: Is ther a check you can do on the value?

app/billing/rest.py
servingUpAces: I moved it to billing_schemas.py

app/commands.py
servingUpAces: Most of the commands are not tested because they are one offs. There is a unit test to compare monthly_billing to ft_billing - which I built this from.

servingUpAces: The commands should go away after they have fulfilled their purpose

tests/app/dao/test_ft_billing_dao.py
servingUpAces: Ah - that's confusing, the data is only populated until December so there is 9 months in the test

PR #1867, opened 2018-05-10
Update ft_billing table to use integer data types

servingUpAces:
We found that the reporting task failed twice because of integrity constraints.

This was because the rate_multiplier was being added as 1 and 1.0 which was not resolving to the same.
This updates the table to use Integrer.
Also changed the logging for the task.

Updated the command so that we can rebuild ft_billing for all services.

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -1,5 +1,6 @@
 from datetime import datetime, timedelta
 
+from flask import current_app
 from notifications_utils.statsd_decorators import statsd
 
 from app import notify_celery
@@ -24,3 +25,6 @@ def create_nightly_billing(day_start=None):
 
         for data in transit_data:
             update_fact_billing(data, process_day)
+
+        current_app.logger.info(
+            "create-nightly-billing task complete. {} rows updated for day: {}".format(len(transit_data), process_day))
--- a/migrations/versions/0189_ft_billing_data_type.py
+++ b/migrations/versions/0189_ft_billing_data_type.py
@@ -1,7 +1,7 @@
 """
 
 Revision ID: 0189_ft_billing_data_type
-Revises: 0187_another_letter_org
+Revises: 0188_add_ft_notification_status
 Create Date: 2018-05-10 14:57:52.589773
 
 """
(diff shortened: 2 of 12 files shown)


app/celery/reporting_tasks.py
```diff
@@ -24,3 +25,5 @@ def create_nightly_billing(day_start=None):
 
         for data in transit_data:
             update_fact_billing(data, process_day)
+
+    current_app.logger.info("create-nightly-billing task complete. {} rows updated".format(len(transit_data)))
```
klssmith: Is this only logging the rows that were updated for the last day processed?

servingUpAces:
Yes - and I forgot the command.py changes.

app/celery/reporting_tasks.py
servingUpAces: Ah - yes you are right. I'm not really sure what logging we need here.

migrations/versions/0189_ft_billing_data_type.py
```diff
@@ -0,0 +1,32 @@
+"""
+
+Revision ID: 0189_ft_billing_data_type
+Revises: 0187_another_letter_org
+Create Date: 2018-05-10 14:57:52.589773
```
servingUpAces: I except this will have to change for the 0188 script that is pending

app/celery/reporting_tasks.py
servingUpAces: Ideas?

leohemsted: i'd probably move the log line in the for loop so you get three log messages - one for each day.

tests/app/billing/test_billing.py
```diff
@@ -542,14 +556,68 @@ def test_compare_ft_billing_to_monthly_billing(client, notify_db_session):
...
+    assert json_response[0]['rate'] == 0
+    assert json_response[0]['letter_total'] == 0
+    assert json_response[1]['notification_type'] == 'letter'
+    assert json_response[1]['billing_units'] == 275
+    assert json_response[1]['rate'] == 0.33
+    assert json_response[1]['letter_total'] == 90.75
+    assert json_response[2]['notification_type'] == 'sms'
+    assert json_response[2]['billing_units'] == 825
+    assert json_response[2]['rate'] == 0.0162
+    assert json_response[0]['letter_total'] == 0
```
leohemsted: this last one should be `json_response[2]`

PR #1871, opened 2018-05-11
Revert "Update gunicorn to 19.8.1"

leohemsted:
Reverts alphagov/notifications-api#1854

We've seen several notify-api instances crash since we upgraded gunicorn - wondering if it's related


leohemsted:
it was merged on admin 11 days ago tho without trouble 🤔 https://github.com/alphagov/notifications-admin/pull/2052

PR #1878, opened 2018-05-17
Fix ft billing migration

servingUpAces:
The rates were not being populated properly.


app/commands.py
```diff
@@ -465,20 +435,19 @@ def migrate_data_to_ft_billing(start_date, end_date):
...
                         and s.crown = l.crown and n.notification_type='letter'), 0) as letter_rate,
                         coalesce(n.international, false) as international,
                         n.billable_units,
                         1 as notifications_sent
                     from public.notification_history n
-                    left join templates t on t.id = n.template_id
-                    left join dm_datetime da on n.created_at>= da.utc_daytime_start
-                        and n.created_at < da.utc_daytime_end
                     left join services s on s.id = n.service_id
                     where n.notification_status!='technical-failure'
```
leohemsted: can remove this tech failure check now that we're matching exact statuses below

servingUpAces: of course - doh

PR #1895, opened 2018-05-31
Added Service contact_link column

richardc0:
This is going to be used for for the document download citizen landing page, a service will add a contact link e.g. https://customerservicecontactnumber.uk/dwp/ which will allow the
user to contact the sending department if there is an error or any issues with the download. This link will be shown along with the Service name.

* Added the contact link to the model
* Added db migration script to add the column to the database

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -342,6 +342,7 @@ class Service(db.Model, Versioned):
     )
     crown = db.Column(db.Boolean, index=False, nullable=False, default=True)
     rate_limit = db.Column(db.Integer, index=False, nullable=False, default=3000)
+    contact_link = db.Column(db.String(255), nullable=True, unique=False)
 
     organisation = db.relationship(
         'Organisation',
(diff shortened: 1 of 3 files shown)


app/models.py
```diff
@@ -342,6 +342,7 @@ class Service(db.Model, Versioned):
     )
     crown = db.Column(db.Boolean, index=False, nullable=False, default=True)
     rate_limit = db.Column(db.Integer, index=False, nullable=False, default=3000)
+    contact_link = db.Column(db.String(255), nullable=True, unique=False)
```
servingUpAces: Will it be possible to have more than one contact link? For example is it not possible to have more than one template that includes a document download link. 
Just wondering if this should be in a mapping table. Maybe I'm over thinking it.

richardc0: When I discussed it with Chris H it was a Service contact link i.e. 'Check your state pension' would have a contact page. I think in MVP terms this is right - per template feels like an advanced feature we should user research or test etc.

PR #1900, opened 2018-06-07
Add some useful information to the log.

servingUpAces:
Fix unit test to actually test the right thing

### DIFF
--- a/app/notifications/notifications_ses_callback.py
+++ b/app/notifications/notifications_ses_callback.py
@@ -95,8 +95,8 @@ def process_ses_response(ses_request):
 
 
 def determine_notification_bounce_type(notification_type, ses_message):
-    remove_emails_from_bounce(ses_message['bounce'])
-    current_app.logger.info('SES bounce dict: {}'.format(ses_message['bounce']))
+    remove_emails_from_bounce(ses_message)
+    current_app.logger.info('SES bounce dict: {}'.format(ses_message))
     if ses_message['bounce']['bounceType'] == 'Permanent':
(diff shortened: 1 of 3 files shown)


app/notifications/notifications_ses_callback.py
```diff
@@ -105,8 +105,12 @@ def determine_notification_bounce_type(notification_type, ses_message):
 
 
 def remove_emails_from_bounce(bounce_dict):
-    for recip in bounce_dict['bouncedRecipients']:
-        recip.pop('emailAddress')
+    if bounce_dict.get('mail').get('headers', None):
+        bounce_dict['mail'].pop('headers')
+    if bounce_dict.get('mail').get('commonHeaders'):
```
sakisv: I'd write that as 

```
if bounce_dict['mail'].get('headers', None):
    ....
```

So that we're more explicit that `mail` _must_ be there. In any case if it is not, this will fail anyway.

app/notifications/notifications_ses_callback.py
```diff
@@ -131,6 +135,10 @@ def handle_complaint(ses_message):
 
 
 def remove_emails_from_complaint(complaint_dict):
+    if complaint_dict.get('headers', None):
```
sakisv: `None` is not necessary, `.get()` defaults to that anyway.

app/notifications/notifications_ses_callback.py
servingUpAces: If will fail either way, but I'll change it so it anyway to be clearer

PR #1903, opened 2018-06-11
Enable syslog drain on all environments

sakisv:
To be merged after alphagov/notifications-aws#387 and once #1902 has been tested for a few days.

This puts the `logit-ssl-syslog-drain` service in the base manifest for both api and delivery boxes so it will be enabled for all environments.


sakisv:
@richardc0 yes we should. The less code we have the better.

PR #1905, opened 2018-06-13
Search for emails or texts, depending on term

quis:
The caseworking view is going to have a page which displays emails and text messages combined together.

In order for the search to work on this page the user needs to be able to search for an email or a text message. This commit makes it guess what to search for when the `notification_type` isn’t known (basically by saying ‘if the search term is only digits they’re probably looking searching by phone number’).

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -1,4 +1,5 @@
 import functools
+import string
 from datetime import (
     datetime,
     timedelta,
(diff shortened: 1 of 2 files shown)


app/dao/notifications_dao.py
```diff
@@ -597,3 +601,20 @@ def notifications_not_yet_sent(should_be_sending_after_seconds, notification_typ
...
+def guess_notification_type(search_term):
+    try:
+        return (
+            SMS_TYPE,
+            try_validate_and_format_phone_number(''.join(
+                str(int(character))
+                if character not in {'(', ')', ' ', '-', '+'} else ''
+                for character in search_term
+            )).lstrip('+0'),
+        )
```
leohemsted: if we're already failing on letters, and stripping non numeric characters here, and we explicitly allow partials, what does `try_and_validate` give us?

leohemsted: actually, would `search_for_email = set(search_term) & set(string.ascii_letters + '@')` catch the actual likely search terms well enough? if they search for something weird like `!"£$%^&*(` it would match a phone num, though i think that's fine as who knows what they mean then

quis: Yeah, that set magic works just as well. I’ve changed it to that, squashed it into ad423c4 and refactored a bit to not do the normalisation in two different places.

PR #1910, opened 2018-06-19
Add endpoint to show number of complaints

klssmith:
Added an endpoint, `/count-by-date-range`, which returns the total number of complaints for all services. This takes an optional start and end date to filter by, and returns the total number of complaints for today if the start and end date are not given.

[Pivotal story](https://www.pivotaltracker.com/story/show/157969699)


app/complaint/complaint_schema.py
```diff
@@ -0,0 +1,11 @@
+
+complaint_count_request = {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "description": "complaint count request schema",
+    "type": "object",
+    "title": "Complaint count request",
+    "properties": {
+        "start_date": {"type": ["string", "null"], "format": "date"},
```
servingUpAces: Is date a default format type in jsonschema?

klssmith: Yes - there's a test here in case anything to do with the date schema validation changes

https://github.com/alphagov/notifications-api/blob/531a7a5aaa29c0f5eb422cb523b5b171ac2a56e7/tests/app/complaint/test_complaint_rest.py#L55-L64

PR #1918, opened 2018-06-25
Wrap exception to return a sensible message to the client

servingUpAces:
We are getting 500 errors when a POST notification has characters that can not be decoded or the json is malformed.

The exception is wrapped and a sensible error message is returned to the client.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -108,7 +108,8 @@ def post_notification(notification_type):
     try:
         request_json = request.get_json()
     except werkzeug.exceptions.BadRequest as e:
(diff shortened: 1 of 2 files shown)


quis:
What does the error message say?

servingUpAces:
The error message will be the "description" of the exception. 
For example: {error: BadRequestError: message: "'utf-8' codec can't decode byte 0xf8 in position 182: invalid start byte"}

servingUpAces:
In conjunction with this we should take a look at the java client and see can either catch this early or fix the encoding of the string. Trouble is we haven't figured out how to reproduce the unicode error

servingUpAces:
There is a story, https://www.pivotaltracker.com/story/show/158361458

PR #1919, opened 2018-06-26
New endpoint for the updated platform admin stats page

klssmith:
Added in a new endpoint and DAO function to provide the data for the new
platform admin statistics page. The new DAO method gets different data from
the Notifications / NotificationHistory table and also groups it differently.

The old endpoint has not been deleted yet to allow the numbers on the
old and new pages to be compared.

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -453,34 +453,6 @@ def fetch_aggregate_stats_by_date_range_for_all_services(start_date, end_date, i
     return query.all()
 
 
-@statsd(namespace='dao')
-def fetch_new_aggregate_stats_by_date_range_for_all_services(start_date, end_date):
-    start_date = get_london_midnight_in_utc(start_date)
-    end_date = get_london_midnight_in_utc(end_date + timedelta(days=1))
-    table = NotificationHistory
-
-    if start_date >= datetime.utcnow() - timedelta(days=7):
-        table = Notification
-
-    query = db.session.query(
-        table.notification_type,
-        table.status,
-        table.key_type,
-        func.count(table.id).label('count')
-    ).filter(
-        table.created_at >= start_date,
-        table.created_at < end_date
-    ).group_by(
-        table.notification_type,
-        table.key_type,
-        table.status
-    ).order_by(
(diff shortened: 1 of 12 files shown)


app/dao/services_dao.py
```diff
@@ -453,6 +453,34 @@ def fetch_aggregate_stats_by_date_range_for_all_services(start_date, end_date, i
...
+@statsd(namespace='dao')
+def fetch_new_aggregate_stats_by_date_range_for_all_services(start_date, end_date):
+    start_date = get_london_midnight_in_utc(start_date)
+    end_date = get_london_midnight_in_utc(end_date + timedelta(days=1))
+    table = NotificationHistory
+
+    if start_date >= datetime.utcnow() - timedelta(days=7):
+        table = Notification
+
+    query = db.session.query(
```
servingUpAces: Can we use the new FT_NOTIFICATION_STATUS table for this?
We would need to do get "todays" data from notifications, but for other data it could be FT_NOTIFICATION_STATUS.
But it's ok to save that for another PR

app/dao/services_dao.py
```diff
@@ -453,6 +453,34 @@ def fetch_aggregate_stats_by_date_range_for_all_services(start_date, end_date, i
     return query.all()
 
 
+@statsd(namespace='dao')
+def fetch_new_aggregate_stats_by_date_range_for_all_services(start_date, end_date):
```
servingUpAces: Why is this in the service_dao?

app/service/rest.py
```diff
@@ -131,6 +132,19 @@ def get_platform_stats():
     return result
 
 
+@service_blueprint.route('/platform-stats-new', methods=['GET'])
+def get_new_platform_stats():
```
servingUpAces: I know the old route was in the service.rest but should we have a new route for platform admin stuff?

klssmith:
@servingUpAces I've moved the DAO function to the Notifications DAO and created a new blueprint, `platform-stats`. I also noticed that the test notifications weren't always showing when they should, so added a commit for that, and a commit to validate the date format.

PR #1924, opened 2018-07-02
Added a method to reply letters that are in the error folder on S3.

servingUpAces:
You can replay one at a time by sending in the file name. Or all the messages in the error folder.

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -26,7 +26,9 @@ from app.letters.utils import (
     get_reference_from_filename,
     move_scanned_pdf_to_test_or_live_pdf_bucket,
     upload_letter_pdf,
-    move_failed_pdf, ScanErrorType)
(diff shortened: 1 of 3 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -219,3 +224,29 @@ def update_letter_pdf_status(reference, status):
...
+        notify_celery.send_task(
+            name=TaskNames.SCAN_FILE,
+            kwargs={'filename': filename},
+            queue=QueueNames.ANTIVIRUS,
+        )
+    else:
+        error_files = get_file_names_from_error_bucket()
+        for item in error_files:
+            moved_file_name = item.key.split('/')[1]
+            current_app.logger.info("Calling scan_file for: {}".format(moved_file_name))
```
quis: This and the next 7 lines could be replaced with `replay_letters_in_error(moved_file_name)`

PR #1937, opened 2018-07-10
Create ServiceDataRetention

servingUpAces:
Create a table SERVICE_DATA_RETENTION to store the number of days the notification is to be kept for.


migrations/versions/0202_service_data_retention.py
```diff
@@ -0,0 +1,35 @@
...
+                              postgresql.ENUM(name='notification_type', create_type=False),
+                              nullable=False),
+                    sa.Column('days_of_retention', sa.Integer(), nullable=False),
+                    sa.Column('created_at', sa.DateTime(), nullable=False),
+                    sa.Column('updated_at', sa.DateTime(), nullable=True),
+                    sa.ForeignKeyConstraint(['service_id'], ['services.id'], ),
+                    sa.PrimaryKeyConstraint('id')
+                    )
+    op.create_index(op.f('ix_service_data_retention_service_id'), 'service_data_retention', ['service_id'],
+                    unique=False)
```
leohemsted: you should add a uniqueness constraint on `service_id` and `notification_Type`

http://alembic.zzzcomputing.com/en/latest/ops.html#alembic.operations.Operations.create_unique_constraint

app/models.py
```diff
@@ -1843,3 +1843,15 @@ def serialize(self):
...
+class ServiceDataRetention(db.Model):
+    __tablename__ = 'service_data_retention'
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), unique=False, index=True, nullable=False)
+    service = db.relationship(Service, backref=db.backref('service_data_retention'))
+    notification_type = db.Column(notification_types, nullable=False)
+    days_of_retention = db.Column(db.Integer, nullable=False)
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    updated_at = db.Column(db.DateTime, nullable=True, default=datetime.datetime.utcnow)
```
klssmith: Should the `updated_at` column be `onupdate=datetime.datetime.utcnow` instead of `default=datetime.datetime.utcnow` or do they do the same thing?

servingUpAces: It doesn't change the column on the table, but it will populate the value on update if not provided, so it's a good idea to change it to onupdate

PR #1940, opened 2018-07-10
Sort out the dashboard day limits

leohemsted:
Three things that take dates on the dashboard. They should all return between seven and eight days. On a monday morning, they should show all of the previous week, including last monday, from monday 00:00.

### Big numbers

A couple of months ago they were adjusted down to 6-7 days by accident, to line up with template preview. Corrected that, and updated tests.

### Template usage

Since the shift to redis, only returned 6-7 days. Corrected that, and updated tests. I thought the `limit_days` variable might have confused things, so changed it to read `whole_days`, because we always want seven whole days of data (plus change)

### Jobs

Always showed 7-8 days. Updated the test to confirm this.

### DIFF
--- a/app/template_statistics/rest.py
+++ b/app/template_statistics/rest.py
@@ -29,17 +29,18 @@ register_errors(template_statistics)
 
 @template_statistics.route('')
 def get_template_statistics_for_service_by_day(service_id):
+    whole_days = request.args.get('whole_days', request.args.get('limit_days', ''))
     try:
-        limit_days = int(request.args.get('limit_days', ''))
+        whole_days = int(whole_days)
     except ValueError:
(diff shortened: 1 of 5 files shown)


app/template_statistics/rest.py
```diff
@@ -29,17 +29,18 @@
 
 @template_statistics.route('')
 def get_template_statistics_for_service_by_day(service_id):
+    whole_days = request.args.get('whole_days', request.args.get('limit_days', ''))
```
allait: Is the plan to drop this fallback once admin is released or are there other things using it?

leohemsted: nah - will drop this fallback. This endpoint is only used by the service dashboard page

PR #1944, opened 2018-07-12
create fake letter response files with variables timestamps 

leohemsted:
when a test letter is created on dev or preview, we upload a file to the dvla ftp response bucket, to test that our integration with s3 works. s3 triggers an sns notification, which we pick up, and then we download the file and mark the letters it mentions as delivered.

However, if two tests run at the same time, they'll create the same file on s3. One will just overwrite the next, and the first letter will never move into delivered - this was causing functional tests to intermittently fail.

This commit makes the test letter task check if the file exists - if it does, it moves back one second and tries again. It tries this thirty times before giving up.


app/celery/research_mode_tasks.py
```diff
@@ -123,7 +124,18 @@ def firetext_callback(notification_id, to):
 def create_fake_letter_response_file(self, reference):
     now = datetime.utcnow()
     dvla_response_data = '{}|Sent|0|Sorted'.format(reference)
-    upload_file_name = 'NOTIFY-{}-RSP.TXT'.format(now.strftime('%Y%m%d%H%M%S'))
+
+    # try and find a filename that hasn't been taken yet - going back in time 60 seconds
+    for i in range(30):
```
allait: I think from what we've seen functional tests can hit this task a few times in a second. So if more than one task hits this loop they'll be trying the same numbers, which means we'll be repeating the same check and there could be a race condition between the check and the file creation.

Maybe we could shuffle the array to try slots in the last 30 seconds in random order.

app/celery/research_mode_tasks.py
```diff
@@ -123,7 +124,18 @@ def firetext_callback(notification_id, to):
 def create_fake_letter_response_file(self, reference):
     now = datetime.utcnow()
     dvla_response_data = '{}|Sent|0|Sorted'.format(reference)
-    upload_file_name = 'NOTIFY-{}-RSP.TXT'.format(now.strftime('%Y%m%d%H%M%S'))
+
+    # try and find a filename that hasn't been taken yet - going back in time 60 seconds
```
allait: 30 seconds?

PR #1945, opened 2018-07-13
Flexible data retention

servingUpAces:
Some of our services need to have the data around for longer than a day. 

In this PR, new endpoints have been added to manage the number of day to retain the data for each change. 
Endpoints created to:
- create service data retention
- modify service data retention
- get all data retention for service
- get data retention for service by id of data retention

### DIFF
--- a/app/dao/service_data_retention_dao.py
+++ b/app/dao/service_data_retention_dao.py
@@ -1,0 +1,34 @@
+from datetime import datetime
+
+from app import db
+from app.dao.dao_utils import transactional
+from app.models import ServiceDataRetention
+
+
+def get_service_data_retention(service_id):
+    data_retention = ServiceDataRetention.query.filter_by(service_id=service_id).all()
+    return data_retention
+
+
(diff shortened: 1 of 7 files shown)


app/dao/service_data_retention_dao.py
```diff
@@ -0,0 +1,44 @@
+from datetime import datetime
+
+from app import db
+from app.dao.dao_utils import transactional
+from app.models import ServiceDataRetention
+
+
+def fetch_service_data_retention_by_id(service_id, data_retention_id):
```
richardc0: If you have the data_retention_id do you need the service_id?

servingUpAces: The service id is only there as a safety check, that the service is requesting the data for it's own data_retention_id.

servingUpAces: We have done this in other queries by id. Just to ensure the service is getting it's own data and not data for a different service

PR #1958, opened 2018-07-17
Add callback_type column to service_callback_api table

CrystalPea:
This is a first part of work for the below ticket.

Ticket: https://www.pivotaltracker.com/story/show/159003296


servingUpAces:
Looks like we need to update pyYaml

allait:
@servingUpAces pyyaml has pulled 4.1 release from PyPI because they've run into some issues with wheel builds. There's no stable 4.x release available at the moment.

The CVE pyup complains about is their design decision that had was reported as an issue back in 2013. `yaml.load` supports custom types by default, which allows arbitrary code execution when an untrusted file is loaded. I think we only use pyyaml with yaml files we write ourselves so it doesn't affect us. We should still upgrade, but only once a stable release is available.

We'll update the service callback query in a follow up PR. We need to migrate existing callbacks to add a type to them for the new query to work and we can only do this when the new records are being created with a type (so after this PR is released).

PR #1971, opened 2018-07-19
Add and call task to send complaints on to service callback APIs

CrystalPea:
Pivotal ticket: https://www.pivotaltracker.com/n/projects/1443052

Part of the work for this story has already been done here:
Add callback_type column to service_callback_api table: https://github.com/alphagov/notifications-api/pull/1958
Create a migration to assign callback_type "delivery_status" to existing service callback api's: https://github.com/alphagov/notifications-api/pull/1964
Add callback_type to service callback history records: https://github.com/alphagov/notifications-api/pull/1966

### DIFF
--- a/app/celery/service_callback_tasks.py
+++ b/app/celery/service_callback_tasks.py
@@ -23,7 +23,7 @@ def send_delivery_status_to_service(
     status_update = encryption.decrypt(encrypted_status_update)
 
     data = {
-        "notification_id": str(notification_id),
+        "id": str(notification_id),
         "reference": status_update['notification_client_reference'],
         "to": status_update['notification_to'],
         "status": status_update['notification_status'],
@@ -64,7 +64,7 @@ def send_complaint_to_service(self, complaint_data):
 
 
 def _send_data_to_service_callback_api(self, data, service_callback_url, token, function_name):
-    notification_id = data["notification_id"]
+    notification_id = (data["notification_id"] if "notification_id" in data else data["id"])
     try:
         response = request(
             method="POST",
--- a/app/celery/service_callback_tasks.py
+++ b/app/celery/service_callback_tasks.py
@@ -17,44 +17,78 @@ from app.config import QueueNames
 
 @notify_celery.task(bind=True, name="send-delivery-status", max_retries=5, default_retry_delay=300)
 @statsd(namespace="tasks")
-def send_delivery_status_to_service(self, notification_id,
-                                    encrypted_status_update
-                                    ):
+def send_delivery_status_to_service(
+    self, notification_id, encrypted_status_update
+):
+    status_update = encryption.decrypt(encrypted_status_update)
+
+    data = {
+        "notification_id": str(notification_id),
+        "reference": status_update['notification_client_reference'],
+        "to": status_update['notification_to'],
+        "status": status_update['notification_status'],
+        "created_at": status_update['notification_created_at'],
+        "completed_at": status_update['notification_updated_at'],
+        "sent_at": status_update['notification_sent_at'],
+        "notification_type": status_update['notification_type']
+    }
+    _send_data_to_service_callback_api(
+        self,
+        data,
+        status_update['service_callback_api_url'],
+        status_update['service_callback_api_bearer_token'],
+        'send_delivery_status_to_service'
+    )
+
+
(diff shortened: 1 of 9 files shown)


app/celery/service_callback_tasks.py
```diff
@@ -17,44 +17,78 @@
...
-                                    ):
-    try:
-        status_update = encryption.decrypt(encrypted_status_update)
+def send_delivery_status_to_service(
+    self, notification_id, encrypted_status_update
+):
+    status_update = encryption.decrypt(encrypted_status_update)
+
+    data = {
+        "notification_id": str(notification_id),
```
servingUpAces: Changing id to notification_id could affect the clients using this data. I don't think we want to change the name of the key.

app/notifications/notifications_ses_callback.py
```diff
@@ -105,7 +110,7 @@ def determine_notification_bounce_type(notification_type, ses_message):
 
 
 def handle_complaint(ses_message):
-    remove_emails_from_complaint(ses_message)
+    recipient_email = remove_emails_from_complaint(ses_message)[0]
```
servingUpAces: Please confirm this works for a real ses message. You can simulate the response using complaint@simulator.amazonses.com email address.

app/celery/service_callback_tasks.py
```diff
@@ -17,44 +17,78 @@
...
-            "to": status_update['notification_to'],
-            "status": status_update['notification_status'],
-            "created_at": status_update['notification_created_at'],
-            "completed_at": status_update['notification_updated_at'],
-            "sent_at": status_update['notification_sent_at'],
-            "notification_type": status_update['notification_type']
-        }
 
+def _send_data_to_service_callback_api(self, data, service_callback_url, token, function_name):
+    notification_id = data["notification_id"]
```
allait: `complaint_id` would be probably more useful than `notification_id` for complaint callback logs. Having to pass through all the variables for log message is a bit unwieldy, but we could do something like `object_id = data.get('id') or data.get('complaint_id')` here 🤷‍♂️

Alternatively, we could set `'id': str(complaint.id)` in complaint data and then the main key is the same in both payloads.

servingUpAces: We would need to send the notification_id to the client. It's more relevant than the complaint id is to them, as some clients do not set the client_reference

allait: We do send both to the service, I think this variable is only used for the logs and it feels like it'd be easier for us to lookup complaint data by `complaint_id` rather than `notification_id`

app/celery/service_callback_tasks.py
CrystalPea: Good point, I will change it back.

app/notifications/notifications_ses_callback.py
CrystalPea: Good call to check it. I simulated the ses complaint response, and checked the logs, and saw the structure of the response after we removed the recipient email, but it did not help me check if the code we wrote would work for it. To do that I needed to see how the recipient email address is stored in the ses response, but our code specifically gets rid of this field before logging it.

Luckily for us, [Amazon handily shares the structure of its ses messages](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html#mail-object)

Based on that I think the most accurate field to send on to the service team is  `"complainedRecipients"`. I have updated the code accordingly.

Also, from the structure of that field value, quoted below, it looks like there might be more than one recipient who complained about the notification. Hence, I though it would be good to send on the whole value (a list of dictionaries) to the service team. But then maybe it is not how it works and they simply made this structure to account for some edge cases and i should strip it to the email string? What do you think?

```
   "complainedRecipients":[
      {
         "emailAddress":"recipient1@example.com"
      }
   ],
```

CrystalPea: I just realised we do not have to account for more than one recipient because Notify sends notifications to one recipient per request if I remember correctly?

CrystalPea: Reverted :)

app/celery/service_callback_tasks.py
CrystalPea: ok, decision made, we are leaving it as is 👍

PR #1981, opened 2018-07-24
Remove monthly_billing table and all references to it

CrystalPea:
Pivotal ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/159164066

There will also be an Admin pull request that needs to be merged before this one.


migrations/versions/0208_remove_monthly_billing_.py
```diff
@@ -0,0 +1,38 @@
+"""
```
leohemsted: this shouldto go into a separate PR. Migrations run before code change, so we need to remove all code references to the table before we can delete it.

app/billing/rest.py
```diff
@@ -60,41 +55,6 @@ def get_yearly_billing_usage_summary_from_ft_billing(service_id):
     return jsonify(data)
 
 
-@billing_blueprint.route('/monthly-usage')
```
leohemsted: do we want to, as part of this story, migrate the new routes to have these old endpoint routes? i'm not a huge fan of leaving `/ft-...` as part of the route, it's just a technical implementation detail.

if so, then this PR could make those endpoints accessible on both the old and the new routes (with a future PR removing the ft-monthly-usage and ft-yearly-usage-summary routes after admin is changed)
if not, then this PR is fine.

migrations/versions/0208_remove_monthly_billing_.py
CrystalPea: Ahh, good point, I didn't think about it. I will move the migration file into a separate PR 👍

app/billing/rest.py
CrystalPea: I could do it no problem 👍

PR #1987, opened 2018-07-25
Fix unique constraint on ServiceCallbackApi

servingUpAces:
The unique constraint on SeviceCallbackApi was on service_id only.
Now that we have 2 types of api callbacks the constraint to be on service_id + callback_type.


servingUpAces:
It sure should! I forgot to add it

PR #1989, opened 2018-07-25
Email branding name and text

CrystalPea:
This is only first part of the below story, changing the model and creating the migration file.

We still need then to update references to that model later: methods for that model, JSON schemas etc.

Pivotal ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/159300417


migrations/versions/0209_email_branding_update.py
```diff
@@ -0,0 +1,27 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('email_branding', sa.Column('text', sa.String(length=255), nullable=True))
+    op.execute('UPDATE email_branding SET text = name')
+    op.execute('UPDATE email_branding SET name = NULL')
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('email_branding', 'text')
```
servingUpAces: You need to move the data from the text column back to the name column before dropping the text column.

CrystalPea: Good call, thank you!

CrystalPea:
@servingUpAces 
`Should we add something to the app/dao/test_email_branding_dao.py test?`

What would you test for? For now there are no field-specific tests in there.

Also, so far I did not add anything to `email_branding_dao`. Should I add a query to get email branding by name maybe? And then we could have a test for that new method, too :).

Also, is there anything else you can think of that should be in this pull request and isn't? Like:
- should I for  add text field to email_branding fixtures for tests?
- should I already update `get_html_email_options` in `app/delivery/send_to_providers.py' to include both name and text for email_branding?

PR #2022, opened 2018-07-31
Handle zero as a letter page count in billing.

servingUpAces:
Last night we had some letter notifications that were in permanent failure and had page count = 0.

This causes the create-nighlty-billing task to fail.
This will make sure that 0 is returned in this case.


leohemsted:
will this end up with `0` rate rows in the usage table? will the admin page handle it?

PR #2024, opened 2018-07-31
Celery dev null

sakisv:
#### Redirect stderr of workers to /dev/null

Celery prints its own stuff to stderr in a non-json format.

These get picked up by syslog drain and are shipped to
logstash together with the normal json-formatted logs resulting in
duplicate information and reducing the usability of the logs.

### DIFF
--- a/tests/test_manifest_delivery_base.py
+++ b/tests/test_manifest_delivery_base.py
@@ -13,11 +13,8 @@ def test_queue_names_set_in_manifest_delivery_base_correctly():
             start_of_queue_arg = command.find(search)
             if start_of_queue_arg > 0:
                 start_of_queue_names = start_of_queue_arg + len(search)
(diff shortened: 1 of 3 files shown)


tests/test_manifest_delivery_base.py
```diff
@@ -14,7 +14,10 @@ def test_queue_names_set_in_manifest_delivery_base_correctly():
             if start_of_queue_arg > 0:
                 start_of_queue_names = start_of_queue_arg + len(search)
                 queues = command[start_of_queue_names:].split(',')
```
leohemsted:
```
end_of_queue_names = command.find('2>') if '2>' in command else 0
watched_queues += set(q.strip() for q in command[start_of_queue_names:end_of_queue_names].split(','))
```

PR #2031, opened 2018-08-06
Use days of retention when deleting notifications in the nightly job.

servingUpAces:
Update method to delete notifications that are a week old, to look at the days of retention set for the service.

If the service does not have the days of retention set, then use 7 days.
Added a method to get days of retention for a service and notificaiton type

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -22,6 +22,8 @@ from sqlalchemy.sql import functions
 from notifications_utils.international_billing_rates import INTERNATIONAL_BILLING_RATES
 
 from app import db, create_uuid
+from app.aws.s3 import get_s3_object_by_prefix
+from app.letters.utils import LETTERS_PDF_FILE_LOCATION_STRUCTURE
 from app.utils import midnight_n_days_ago, escape_special_characters
 from app.errors import InvalidRequest
 from app.models import (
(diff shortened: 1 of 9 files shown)


sakisv:
@govuk-notify-jenkins test this please

app/dao/notifications_dao.py
```diff
@@ -317,21 +319,43 @@ def delete_notifications_created_more_than_a_week_ago_by_type(notification_type)
...
+                                                  Notification.notification_type == notification_type,
+                                                  Notification.service_id.notin_(
+                                                      services_with_data_retention))
+    _delete_letters_from_s3(notification_type=notification_type, query=query)
+    deleted = query.delete(synchronize_session='fetch')
     return deleted
 
 
+def _delete_letters_from_s3(notification_type, query):
+    if notification_type == LETTER_TYPE:
```
allait: This is minor, but I'd probably move this condition to the task function. Then you won't need the `notification_type` argument and it's clear right away that this doesn't do anything for other notification types.

app/dao/notifications_dao.py
```diff
@@ -317,21 +319,43 @@ def delete_notifications_created_more_than_a_week_ago_by_type(notification_type)
...
+        letters_to_delete_from_s3 = query.all()
+        for letter in letters_to_delete_from_s3:
+            bucket_name = current_app.config['LETTERS_PDF_BUCKET_NAME']
+            sent_at = str(letter.sent_at.date())
```
allait: Does send date always match the "scheduled" folder date (e.g. could the FTP task fail to run on the date or what if a letter fails to send and ends up with technical failure)?

servingUpAces: I'm not sure what we'd do for technical failures, it's likely the sent_at date is not set. 
If the letter is not in the sent_at folder, not sure where we'd find it.
The 90 day S3 expiration will kick in.

PR #2042, opened 2018-08-13
Get service data retention by notification type

allait:
### Add limit_days argument to notification statistics endpoint

Allows getting notification counts for a given number of days to support services with custom data retention periods (admin dashboard page should still display counts for the last 7 days, while the notifications page displays all stored notifications).

### Add an endpoint to fetch service data retention by type

Admin app needs to get the service data retention for the specified notification type, so to avoid iterating through the list of all existing service data retention settings we restore the endpoint to get the individual data retention period.

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -246,9 +246,9 @@ def delete_service_and_all_associated_db_objects(service):
 
 
 @statsd(namespace="dao")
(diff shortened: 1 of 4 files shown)


allait:
@servingUpAces I've added some tests with variable limit_days to make sure it's not being hard-coded somewhere else.

PR #2050, opened 2018-08-21
Add notification status for returned letters

allait:
We need to update letter notifications with a new status when DVLA gives us a list of references for returned letters.

This adds the new status to the models and the DB.

DVLA call this 'returned mail', so I'm using it as the status name since it seems less ambiguous than 'returned'.


leohemsted:
minor thing but i'd rather you called this `returned-letter` or something to keep the nomenclature consistent

allait:
@leohemsted @servingUpAces renamed to `returned-letter`.

allait:
Merged this as part of #2075

PR #2051, opened 2018-08-23
Add unit tests for the Users DAO create_secret_code function

pwright08:
I noticed there weren't any unit tests for the Users DAO create_secret_code function (used for generating 2FA codes I believe), so I thought I would add some. Feel free to suggest others that may be useful.

`make test` passed locally.


quis:
@govuk-notify-jenkins test this please

PR #2054, opened 2018-08-23
Add brand_type to EmailBranding.

servingUpAces:
It makes more sense to put the brand_type with EmailBranding rather than in Service.
Next step is to add the new type to the form in admin app.


servingUpAces:
That's not a feature of Notify - the branding is for the service

PR #2057, opened 2018-08-24
Sanitise precompiled pdfs after virus scan

leohemsted:
Notify antivirus, on success, calls the process_virus_scan_passed task.

### Previously, this task would:

* update status to created (or delivered for test keys)
* copy the file from the scan bucket to either the live or test bucket based on the results
* delete the old file in the scan bucket

### We want it to:

* download file from scan bucket
* sanitise PDF using new template-preview functionality
* if sanitise failed, set to new status "validation-failed" and save the pdf somewhere.
* send new pdf to live/test bucket
* update status to created (or delivered for test keys)
* delete the original file in the scan bucket

 ### This PR does some of that:

* download file from scan bucket
* sanitise PDF using new template-preview functionality
* if sanitise failed, just log.
* send OLD pdf to live/test bucket
* update status to created (or delivered for test keys)
* delete the original file in the scan bucket

So if sanitising fails, we won't fall over and not deliver the letter, we'll just log a message for now. If sanitise throws an unexpected error (as opposed to a 400), we'll retry up to fifteen times (the same as when creating a new letter). I've added the code for using the sanitised pdf, but it's commented out for now

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -9,6 +9,7 @@ from requests import (
 )
 
 from notifications_utils.statsd_decorators import statsd
+from notifications_utils.s3 import s3upload
 
 from app import notify_celery
 from app.aws import s3
@@ -24,9 +25,11 @@ from app.dao.notifications_dao import (
 from app.errors import VirusScanError
 from app.letters.utils import (
     get_reference_from_filename,
-    move_scanned_pdf_to_test_or_live_pdf_bucket,
+    get_folder_name,
     upload_letter_pdf,
-    move_failed_pdf, ScanErrorType, move_error_pdf_to_scan_bucket,
+    move_failed_pdf,
+    ScanErrorType,
+    move_error_pdf_to_scan_bucket,
     get_file_names_from_error_bucket
 )
 from app.models import (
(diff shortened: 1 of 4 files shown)


leohemsted:
It's worth noting that template preview specifically avoids returning a 400 on error, for now. And this specifically avoids rejecting a notification on 400, for now. So it's twice-protected.

For now, we'll just look at logs. It'll require changes in both template preview and api to make this complete and production ready

app/celery/letters_pdf_tasks.py
```diff
@@ -163,22 +167,83 @@ def letter_in_created_state(filename):
...
     is_test_key = notification.key_type == KEY_TYPE_TEST
-    move_scanned_pdf_to_test_or_live_pdf_bucket(
+
+    scan_pdf_object = s3.get_s3_object(current_app.config['LETTERS_SCAN_BUCKET_NAME'], filename)
+    old_pdf = scan_pdf_object.get()['Body'].read()
+
+    new_pdf = _sanitise_precomiled_pdf(self, notification, old_pdf)
+
+    if not new_pdf:
+        current_app.logger.info('Invalid precompiled pdf received {} ({})'.format(notification.id, filename))
```
allait: I think it's worth switching 400s on template-preview so we start getting these log messages. The ones from template-preview itself don't seem to have anything that would help us find the original PDF.

leohemsted: that's a v good point. I'll make a PR

leohemsted: i have made a PR https://github.com/alphagov/notifications-template-preview/pull/193

PR #2073, opened 2018-08-30
Drop service branding column 

quis:
_This is branched off https://github.com/alphagov/notifications-api/pull/2072 and will need rebasing before merging._

We’re not using this column any more, so let’s clean up after ourselves.


quis:
Rebase done ✅

PR #2077, opened 2018-09-03
Add unique constraint to email branding domain

quis:
Brandings with a domain set should be considered canonical. It doesn’t make sense to have the same domain set on multiple different email brands – you can’t tell which one to use.


klssmith:
Should the `EmailBranding` model also have `unique=True` added to the `domain` column?

quis:
@klssmith Yes it should. Changed and squashed into f095fa1.

PR #2081, opened 2018-09-04
Add domain constraint 2

quis:
Reverts https://github.com/alphagov/notifications-api/pull/2080

---

Also stops updating email branding domains with empty string values.

Columns with multiple `null`s can have a uniqueness constraint. Columns with multiple empty string values are not considered unique.

So this PR also:
- removes any duplicate empty string values
- casts empty strings to null string any time these columns are updated


quis:
Added error handler so duplicate domain doesn’t cause a `500`:

image

PR #2092, opened 2018-09-14
Add letter_class to Service

servingUpAces:
In order to support sending letters by first class we need a way to know how the service wants us to send the letter.

To start with this will be an attribute on the service, at the time the notification is created it will look at Service.letter_class to decide what class to use for the letter.

This PR adds Service.letter class as a nullable column.
Updated the create_service and update_service method to default the value to second.

Subsequent PRs will add the check constraint to ensure we only get first or second in the letter_class column and make that column nullable.
This can't be done all at once because it will cause an error if someone inserts or updates a service during the deploy.


tests/app/db.py
```diff
@@ -660,3 +662,44 @@ def create_service_data_retention(
         days_of_retention=days_of_retention
     )
     return data_retention
+
+
+def create_sample_inbound_numbers(service=None):
+    if not service:
+        service = create_service(service_name='sample service 2')
+    service_2 = create_service(service_name='service 2 for inbound')
```
leohemsted: i'm not super sure i'm happy about this function randomly creating two arbitrary services - only one of which specifically says inbound. But I appreciate you're just copying the existing behaviour of the `sample_inbound_numbers` fixture. not sure what to do here (if anything)

tests/app/db.py
```diff
@@ -660,3 +662,44 @@ def create_service_data_retention(
...
+    inbound_numbers = list()
+    inbound_numbers.append(create_inbound_number(number='1', provider='mmg'))
+    inbound_numbers.append(create_inbound_number(number='2', provider='mmg', active=False, service_id=service.id))
+    inbound_numbers.append(create_inbound_number(number='3', provider='firetext',
+                                                 service_id=service_2.id))
+    return inbound_numbers
+
+
+def create_code(code_type, usr=None, code=None):
```
leohemsted: I'm not sure we need this function - we're calling it twice - passing in a `usr` and a `code`. I think it would be easier to read just calling `create_user_code` from the calling function and not have this function until we know there's an need for it

PR #2096, opened 2018-09-17
Introduce new letter rates, valid from 01.10.2018

CrystalPea:
Work done:
- insert new rates into the letter_rates table
- use post_class to filter the rates, for now hard-coded as second class
- update existing tests
- write a test for filtering by post_class
- in config, move inbound MMG credentials for local environment from Test class to Development class
- TODO: tests for filtering rates by start date

Pivotal ticket: https://www.pivotaltracker.com/story/show/160229653

### DIFF
--- a/migrations/versions/0226_new_letter_rates.py
+++ b/migrations/versions/0226_new_letter_rates.py
@@ -1,0 +1,53 @@
+"""empty message
+
+Revision ID: 0226_new_letter_rates
+Revises: 0225_another_letter_org
+
+"""
+
+revision = '0226_new_letter_rates'
+down_revision = '0225_another_letter_org'
+
+import uuid
+from datetime import datetime
+from alembic import op
+
+
+start = datetime(2018, 10, 1, 0, 0, tzinfo=timezone.utc)
+
+NEW_RATES = [
+    (uuid.uuid4(), start, 1, 0.30, True, 'second'),
+    (uuid.uuid4(), start, 1, 0.30, False, 'second'),
+    (uuid.uuid4(), start, 2, 0.35, True, 'second'),
+    (uuid.uuid4(), start, 2, 0.35, False, 'second'),
+    (uuid.uuid4(), start, 3, 0.40, True, 'second'),
+    (uuid.uuid4(), start, 3, 0.40, False, 'second'),
+    (uuid.uuid4(), start, 4, 0.45, True, 'second'),
+    (uuid.uuid4(), start, 4, 0.45, False, 'second'),
+    (uuid.uuid4(), start, 5, 0.50, True, 'second'),
+    (uuid.uuid4(), start, 5, 0.50, False, 'second'),
+    (uuid.uuid4(), start, 1, 0.56, True, 'first'),
+    (uuid.uuid4(), start, 1, 0.56, False, 'first'),
+    (uuid.uuid4(), start, 2, 0.61, True, 'first'),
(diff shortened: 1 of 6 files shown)


migrations/versions/0226_new_letter_rates.py
```diff
@@ -0,0 +1,54 @@
...
+revision = '0226_new_letter_rates'
+down_revision = '0225_another_letter_org'
+
+import uuid
+import pytz
+from datetime import datetime
+from alembic import op
+
+
+start = datetime(2018, 10, 1, 0, 0, tzinfo=pytz.utc)
```
klssmith: I think we can get away without adding in timezone info - we don't do this elsewhere, we just keep the database in UTC.

migrations/versions/0226_new_letter_rates.py
```diff
@@ -0,0 +1,54 @@
...
+
+NEW_RATES = [
+    (uuid.uuid4(), start, 1, 0.30, True, 'second'),
```
klssmith: We only need to add in the rows for 1st class letters, and the rows for 2nd class letters where the rate is going to change. For example, the rate for 1 page letters that are crown and 2nd class will not change, so we don't need to add a new row.

klssmith:
The `letter_rates` table has an `end_date` column - I would expect that the rows where we are changing the rate would need to have an `end_date` added. But then from taking a brief look at the existing code, I can't spot anywhere where we currently check the `end_date` to see which rate to use. We seem to determine which rate to used based on the `start_date` - we should check what's currently going on here and how getting a rate should work.

There was a change made (by me, sorry!) a while ago to use the real `letter_rates` table data in the tests. Unfortunately this means that it's difficult to test what happens if you have a rate which has ended in the `letter_rates` table (since we don't currently have any rates that have ended) - maybe it would be helpful to reverse that change so that we can test things thoroughly.

migrations/versions/0226_new_letter_rates.py
servingUpAces: That's right, the database is timezone agnostic. It doesn't persist which timezone you use. This means we need persist the notification as UTC. October first is still in BST, start will be persisted as 2018-10-01 00:00:00, however we actually need it to start 2018-09-30 23:00:00. In short if you create `start = datetime(2018, 09, 30, 23, 0)` that should be fine.

CrystalPea: ok, cheers, will do :)

migrations/versions/0226_new_letter_rates.py
CrystalPea: Ok, there was one rate that did not change, I deleted it from the new migration so it will still be covered by the rate set earlier ^_^

migrations/versions/0229_new_letter_rates.py
```diff
@@ -0,0 +1,52 @@
...
+
+import uuid
+from datetime import datetime
+from alembic import op
+
+
+start = datetime(2018, 9, 30, 23, 0)
+
+NEW_RATES = [
+    (uuid.uuid4(), start, 1, 0.30, False, 'second'),
```
servingUpAces: We don't need new rate for `second` class, let's remove these please. By the way if you were to insert new rates for second we would need to expire the old second class rates, it doesn't make sense to have to valid rates.

CrystalPea: Ok, I will do that :)

PR #2131, opened 2018-10-03
Celery 4

leohemsted:
Check commits for more info, especially regarding `pycurl` and how rubbish it is


Makefile
```diff
@@ -67,6 +67,7 @@ production: ## Set environment to production
 .PHONY: dependencies
 dependencies: venv ## Install build dependencies
 	mkdir -p ${PIP_ACCEL_CACHE}
+	. venv/bin/activate && make install-pycurl
```
allait: I think if we're going to rely on the make target instead of running `pip install` we should sort out the venv/virtualenvwrapper stuff. Right now it looks like this will only play well with a `venv/` folder, which it will try to create each time it's run. Instead, ideally it would recognize when it's already being run inside a venv and not override it.

leohemsted: hmmm i can do this - but this means that every build is going to take a super long time until we sort out the whole dockerfile to not mount the volume, which i don't really want to do as part of this PR just cos it's quite separate. Thoughts?

PR #2139, opened 2018-10-08
[do not merge] Change API get_notifications endpoint to return jobs

klssmith:
Updated the v2/notifications endpoint to always include notifications sent via a job.

[Pivotal story](https://www.pivotaltracker.com/story/show/160757291)


klssmith:
We've decided not to include jobs in the api response

PR #2142, opened 2018-10-09
Revert "Celery 4"

leohemsted:
Reverts alphagov/notifications-api#2131

On preview, testing showed that during low traffic, tasks stay on the queue significantly shorter. We're talking going from 3-7 seconds, down to 0.2 seconds. Great stuff 👍 

However, on staging, with performance tests doing a baseline (4 concurrent requests) or higher load (20 concurrent requests), data was pretty bad.

Long polling (controlled by the `wait_time_seconds` setting in BROKER_TRANSPORT_OPTIONS) did not affect this.

This was a sendSmsAndEmail test with 10 users (of each type), with a ramp of 100 secs, then a 60 second run, then a 60 second ramp down (our default perf test settings). Higher percentages are better.

|      name    | count |      under 3 secs      |      under 5 secs      |     under 10 secs      |     under 30 secs      |     under 60 secs     |
|--------------|-------|------------------------|------------------------|------------------------|------------------------|-----------------------|
|celery 3      |  4113 | 32% | 44% | 58% | 100% | 100%|
|c4 long poll  |  4561 | 5% | 8% | 18% | 70% | 100%|
|c4 short poll |  4511 | 6% | 9% | 16% | 88% | 100%|

some things to try:

* switch out sqs for redis?
* read the debug logs and figure out where the time is ?
* get a good minimal reproducible test case? possibly even one not involving notify app?
* the stuff martin was mucking around with queue names?


leohemsted:
Tried switching out redis - doesn't play nicely with paas redis. Possibly due to paas forcing you to `rediss` (redis over ssl), which isn't supported by a released version of kombu (though has been merged to master). Found that the API didn't appear to be pushing tasks, and the delivery worker crashed on startup - neither gave any useful logs when doing so. 🤷‍♂️

PR #2171, opened 2018-10-18
Update page count after antivirus scan

servingUpAces:
Seems silly to get the page count before the virus scan, so this just moves the code around a bit.


app/celery/letters_pdf_tasks.py
```diff
@@ -207,19 +211,38 @@ def process_virus_scan_passed(self, filename):
...
     current_app.logger.info('notification id {} ({}) sanitised and ready to send'.format(notification.id, filename))
 
+    # temporarily upload original pdf while testing sanitise flow.
     _upload_pdf_to_test_or_live_pdf_bucket(
-        new_pdf,
+        old_pdf,  # TODO: change to new_pdf
         filename,
         is_test_letter=is_test_key)
 
+    billable_units = _get_page_count(notification, old_pdf)
```
leohemsted: i think we should put this above the sanitise command - if the file is invalid, then there's no point making the rest request to template preview

app/celery/letters_pdf_tasks.py
```diff
@@ -207,19 +211,38 @@ def process_virus_scan_passed(self, filename):
...
+    try:
+        pages = pdf_page_count(io.BytesIO(old_pdf))
+        pages_per_sheet = 2
+        billable_units = math.ceil(pages / pages_per_sheet)
+        return billable_units
+    except PdfReadError as e:
+        current_app.logger.exception(msg='Invalid PDF received for notification_id: {}'.format(notification.id))
+        update_letter_pdf_status(
+            reference=notification.reference,
+            status=NOTIFICATION_PERMANENT_FAILURE
```
leohemsted: hmm should we use `NOTIFICATION_VALIDATION_FAILED` for this?

PR #2182, opened 2018-10-23
Added inbound_sms_id to log output

servingUpAces:
the extra information may help when trying to help debugging an issue.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -274,10 +274,10 @@ def save_email(self,
 @notify_celery.task(bind=True, name="save-letter", max_retries=5, default_retry_delay=300)
 @statsd(namespace="tasks")
 def save_letter(
-    self,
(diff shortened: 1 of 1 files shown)


app/celery/tasks.py
```diff
@@ -536,7 +536,8 @@ def send_inbound_sms_to_service(self, inbound_sms_id, service_id):
         response.raise_for_status()
     except RequestException as e:
         current_app.logger.warning(
-            "send_inbound_sms_to_service request failed for service_id: {} and url: {}. exc: {}".format(
+            "send_inbound_sms_to_service failed for service_id: {} for inbound_sms_id: {} and url: {}. exc: {}".format(
+                inbound_sms_id,
```
klssmith: `service_id` should be before `inbound_sms_id` in the arguments to `format()`.

PR #2193, opened 2018-10-29
Add new failed stats to statistic counts

servingUpAces:
Add validation-failed and virus-scan-failed in the failed counts for stats

### DIFF
--- a/tests/app/service/test_statistics.py
+++ b/tests/app/service/test_statistics.py
@@ -35,7 +35,9 @@ NewStatsRow = collections.namedtuple('row', ('notification_type', 'status', 'key
         StatsRow('email', 'technical-failure', 1),
         StatsRow('email', 'temporary-failure', 1),
         StatsRow('email', 'permanent-failure', 1),
-    ], [4, 0, 4], [0, 0, 0], [0, 0, 0]),
+        StatsRow('email', 'validation-failed', 1),
+        StatsRow('email', 'virus-scan-failed', 1),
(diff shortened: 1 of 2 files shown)


tests/app/service/test_statistics.py
```diff
@@ -35,7 +35,9 @@
         StatsRow('email', 'technical-failure', 1),
         StatsRow('email', 'temporary-failure', 1),
         StatsRow('email', 'permanent-failure', 1),
-    ], [4, 0, 4], [0, 0, 0], [0, 0, 0]),
+        StatsRow('email', 'validation-failed', 1),
```
quis: Think this would be easier to understand if you changed the type to `letter` here (since emails can’t get into this state):
```suggestion
        StatsRow('letter', 'validation-failed', 1),
```

tests/app/service/test_statistics.py
```diff
@@ -35,7 +35,9 @@
...
+        StatsRow('email', 'virus-scan-failed', 1),
```
quis: …and here too:
```suggestion
        StatsRow('letter', 'virus-scan-failed', 1),
```

tests/app/service/test_statistics.py
```diff
@@ -35,7 +35,9 @@
...
+    ], [6, 0, 6], [0, 0, 0], [0, 0, 0]),
```
quis: then this line would have counts per channel:
```suggestion
    ], [4, 0, 4], [0, 0, 0], [1, 0, 1]),
```

PR #2201, opened 2018-10-31
Allow the admin app to send one-off letters

quis:
This pull request modifies the code paths the admin app uses to send one off emails and text messages to also accept letters.

It seems to work running locally, but I could do with some help to:
- check it’s doing all the required tasks/database updates
- write some good tests (all I’ve done is modify the existing ones so they pass)

### DIFF
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -12,6 +12,7 @@ from notifications_utils.recipients import (
 
 from app import redis_store
 from app.celery import provider_tasks
+from app.celery.letters_pdf_tasks import create_letters_pdf
 from app.config import QueueNames
 
 from app.models import (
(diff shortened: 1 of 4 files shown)


app/notifications/process_notifications.py
```diff
@@ -149,9 +152,18 @@ def send_notification_to_queue(notification, research_mode, queue=None):
...
+            queue = QueueNames.CREATE_LETTERS_PDF
+        deliver_task = create_letters_pdf
 
     try:
         deliver_task.apply_async([str(notification.id)], queue=queue)
+        if notification.notification_type == LETTER_TYPE:
+            update_notification_status_by_reference(
+                notification.reference,
+                NOTIFICATION_DELIVERED,
+            )
```
leohemsted: `delivered` is saved for when we've sent to dvla. The letter should stay in `created` (the zip_and_send_to_dvla task only processes letters that are in created)

app/notifications/rest.py
```diff
@@ -88,13 +88,18 @@ def get_all_notifications():
 @notifications.route('/notifications/<string:notification_type>', methods=['POST'])
```
leohemsted: this is the old v1 public endpoint - we shouldn't be touching this cos it needs an api key and everything

Instead, the flow should use this function (which is accessed via the route `/<uuid:service_id>/send-notification`):

https://github.com/alphagov/notifications-api/blob/master/app/service/send_notification.py#L37

quis:
I’ve updated this to make sure that:
- one-off letters are processed by the correct task, from the correct queue
- one-off letters sent from a service in research mode don’t get put on a queue and go straight to `delivered` (because we don’t want to send them for real)

PR #2210, opened 2018-11-05
 Get and use sender_id from S3 metadata

klssmith:
The `save_email` and `save_sms` jobs were updated previously to take an
optional `sender_id` and to use this if it was available. This commit
now gets the `sender_id` from the S3 metadata if it exists and passes it
through the the tasks which save the job notifications. This means SMS
and emails sent through jobs can use a specified `sender_id` instead of
the default.

[Pivotal story](https://www.pivotaltracker.com/story/show/161176428)

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -112,7 +112,7 @@ def process_job(job_id, sender_id=None):
             template_type=template.template_type,
             placeholders=template.placeholders
     ).rows:
-        process_row(row, template, job, service)
+        process_row(row, template, job, service, sender_id=sender_id)
 
     job_complete(job, start=start)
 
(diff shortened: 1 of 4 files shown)


app/job/rest.py
```diff
@@ -150,8 +150,11 @@ def create_job(service_id):
 
     dao_create_job(job)
 
+    sender_id = data.get('sender_id')
+    sender_id = None if sender_id == 'None' else sender_id
```
allait: This can be removed if admin doesn't set `sender_id` on S3 metadata when it's `None`

app/celery/tasks.py
```diff
@@ -151,12 +153,17 @@ def process_row(row, template, job, service):
 
     send_fn = send_fns[template_type]
 
+    sender_dict = {}
```
allait: I'd call this `task_kwargs`

PR #2212, opened 2018-11-07
Add parent id for create template

servingUpAces:
- Using jsonschema for create_template. Updated jsonschema to Draft7, this allowed a conditional validation on subject, if template_type == 'email' or 'letter' then subject is required.
This version is backward compatible with Draft4.
When creating TemplateRedacted, I've built the dict depending on if the created_by or created_by_id exists.

- Update model and controller to handle parent_folder_id when creating a template. If the parent_folder_id then check if the folder exists and is for the same service. If it is add the folder to the template model object, the relationship will be persisted when the template is saved. If the folder does not exist or is for a different service, then return a ResultNotFound error.

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -12,6 +12,7 @@ from notifications_utils import SMS_CHAR_COUNT_LIMIT
 from notifications_utils.pdf import extract_page_from_pdf
 from notifications_utils.template import SMSMessageTemplate
 from requests import post as requests_post
+from sqlalchemy.orm.exc import NoResultFound
 
 from app.dao.notifications_dao import get_notification_by_id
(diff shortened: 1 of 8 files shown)


app/template/rest.py
```diff
@@ -46,12 +48,24 @@ def _content_count_greater_than_limit(content, template_type):
     return template.content_count > SMS_CHAR_COUNT_LIMIT
 
 
+def validate_parent_folder(template_json):
+    if template_json.get("parent_folder_id"):
+        return TemplateFolder.query.filter_by(
```
allait: I've updated the dao method in https://github.com/alphagov/notifications-api/pull/2213 to filter based on `service_id`, so you could use that here

PR #2217, opened 2018-11-08
Remove relationship to folder when archiving a template

klssmith:
When a template is archived, it should no longer belong to any folder.

If we don’t do this it will make it very hard to delete folders later
(because folders can only be deleted if they have no templates or folders
inside them).

We originally tried to check if the link between a template and folder
should be removed with

`if template.archived and template.folder:`

instead of using `if template.archived:`. However, this caused issues
because checking `template.folder` flushes the session. Since the
session is no longer dirty, the versioning decorator doesn't work as
expected and doesn't create a new row in `TemplateHistory`.

### DIFF
--- a/app/dao/templates_dao.py
+++ b/app/dao/templates_dao.py
@@ -38,6 +38,9 @@ def dao_create_template(template):
 @transactional
 @version_class(Template, TemplateHistory)
 def dao_update_template(template):
+    if template.archived:
+        template.folder = None
+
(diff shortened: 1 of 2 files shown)


klssmith:
@allait I've updated the commit message - can you have a quick look to check that it makes sense and I've not forgotten any details?

allait:
@klssmith yeah, looks great :+1:

PR #2218, opened 2018-11-08
Move folders and templates to other folders

leohemsted:
new endpoints:
/services/<service_id>/move-to-folder
/services/<service_id>/move-to-folder/<target_template_folder_id>

* takes in a dict containing lists of `templates` and `folders` uuids.
* sets parent of templates and folders to the folder specified in the
  URL. Or None, if there was no id specified.
* if any template or folder has a differen service id, then the whole
  update fails
* if any folder is an ancestor of the target folder, then the whole
  update fails (as that would cause a cyclical folder structure).
* the whole function is wrapped in a single `transactional` decorator,
  so in case of error nothing will be saved.

### DIFF
--- a/app/template_folder/rest.py
+++ b/app/template_folder/rest.py
@@ -111,7 +111,7 @@ def move_to_template_folder(service_id, target_template_folder_id=None):
         try:
             template_folder = dao_get_template_folder_by_id_and_service_id(template_folder_id, service_id)
         except NoResultFound:
-            msg = 'Could not move to folder: No folder found with id {} for service {}'.format(
+            msg = 'No folder found with id {} for service {}'.format(
                 template_folder_id,
                 service_id
             )
--- a/app/template_folder/rest.py
+++ b/app/template_folder/rest.py
@@ -1,4 +1,4 @@
-from flask import Blueprint, jsonify, request, abort, current_app
+from flask import Blueprint, jsonify, request, current_app
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
 
@@ -145,4 +145,4 @@ def move_to_template_folder(service_id, target_template_folder_id=None):
             ))
         else:
             template.folder = target_template_folder
-    return '', 200
+    return '', 204
--- a/app/template_folder/rest.py
+++ b/app/template_folder/rest.py
@@ -116,13 +116,7 @@ def move_to_template_folder(service_id, target_template_folder_id=None):
                 service_id
             )
             raise InvalidRequest(msg, status_code=400)
-
-        if target_template_folder and template_folder.is_parent_of(target_template_folder):
-            msg = 'Could not move to folder: {} is an ancestor of target folder {}'.format(
-                template_folder_id,
-                target_template_folder_id
-            )
-            raise InvalidRequest(msg, status_code=400)
+        _validate_folder_move(target_template_folder, target_template_folder_id, template_folder, template_folder_id)
 
         template_folder.parent = target_template_folder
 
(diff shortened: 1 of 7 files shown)


app/template_folder/rest.py
```diff
@@ -89,3 +92,57 @@ def delete_template_folder(service_id, template_folder_id):
     dao_delete_template_folder(template_folder)
 
     return '', 204
+
+
+@template_folder_blueprint.route('/move-to-folder', methods=['POST'])
+@template_folder_blueprint.route('/move-to-folder/<uuid:target_template_folder_id>', methods=['POST'])
```
allait: It doesn't solve everything, but I think this would be a bit more RESTful if we think of this as updating folder contents instead of a "move" action. So instead of `/move-to-folder/<uuid>` we'd have `POST /template-folder/<uuid>/contents`. This is still a bit awkward without any ID for "top-level", but I think `/template-folder/contents` could work.

One of the benefits is that we can add a `GET` endpoint to retrieve everything in the folder if we need it in the future and model this as a proper resource instead of having a number of unconnected remote procedures.

tests/app/template_folder/test_template_folder_rest.py
```diff
@@ -188,3 +188,160 @@ def test_delete_template_folder_fails_if_folder_contains_templates(admin_request
...
+
+    # versions are all unchanged
+    assert t1.version == 1
+    assert t2.version == 1
+    assert t3.version == 1
+    assert t4.version == 1
+
+
+def test_move_to_folder_moves_folders_and_templates_to_top_level_if_no_target(admin_request, sample_service):
+    f1 = create_template_folder(sample_service, name='f1')
```
CrystalPea: I don't understand: we  want to test moving this folder out from another folder and to the top level, right? But when we create it, isn't it already at the top level?

tests/app/template_folder/test_template_folder_rest.py
```diff
@@ -188,3 +188,160 @@ def test_delete_template_folder_fails_if_folder_contains_templates(admin_request
...
+
+    assert f1.parent is None
+    assert f2.parent is None  # unchanged
+
+    assert t1.folder is None  # moved out of f1, even though f1 is also being moved
+    assert t2.folder == f1  # stays in f1, though f1 has moved
+    assert t3.folder == f2  # unchanged
+
+
+def test_move_to_folder_rejects_folder_from_other_service(admin_request, notify_db_session):
```
CrystalPea: Here (and in other tests where we want status 400) we could also check if the correct message appears, since we have several. It should be 
`'Could not move to folder: No template found with id {} for service {}'`

But I guess that's not essential :)

tests/app/template_folder/test_template_folder_rest.py
CrystalPea: I don't understand: we want to test moving this folder out from another folder and to the top level, right? But when we create it, isn't it already at the top level?

app/template_folder/rest.py
```diff
@@ -89,3 +92,57 @@ def delete_template_folder(service_id, template_folder_id):
...
+            template_folder = dao_get_template_folder_by_id_and_service_id(template_folder_id, service_id)
+        except NoResultFound:
+            msg = 'Could not move to folder: No folder found with id {} for service {}'.format(
+                template_folder_id,
+                service_id
+            )
+            current_app.logger.error(msg)
+            raise InvalidRequest(msg, status_code=400)
+
+        if target_template_folder and template_folder.is_parent_of(target_template_folder):
```
allait: I don't know if this belongs in `is_parent_of` or as a separate check, but we should also check that we're not making the folder its own parent (so `target_template_folder_id != template_folder_id`)

tests/app/template_folder/test_template_folder_rest.py
servingUpAces: It's a good thing you suggested that, one of the tests was getting a 400, but for a different reason than what the test was suppose to be testing

app/template_folder/rest.py
```diff
@@ -89,3 +92,54 @@ def delete_template_folder(service_id, template_folder_id):
...
+            template = dao_get_template_by_id_and_service_id(template_id, service_id)
+        except NoResultFound:
+            msg = 'Could not move to folder: No template found with id {} for service {}'.format(
+                template_id,
+                service_id
+            )
+            raise InvalidRequest(msg, status_code=400)
+
+        if template.archived:
+            current_app.logger.error('Could not move to folder: Template {} is archived. (Skipping)'.format(
```
allait: This probably should be a `.warning` or `.info`

tests/app/template_folder/test_template_folder_rest.py
CrystalPea: Ahh, good catch *high five* :)

PR #2225, opened 2018-11-12
Change name of update method for template folder

CrystalPea:
It was initialy called "rename" which does not comply with
RESTful CRUD (create, update, read, delete) naming practice.
We remove the 'rename' operation in favour of template folder
resource update endpoint as it allows us to extend it with other
attributes.

### DIFF
--- a/app/template_folder/template_folder_schema.py
+++ b/app/template_folder/template_folder_schema.py
@@ -11,9 +11,9 @@ post_create_template_folder_schema = {
     "required": ["name", "parent_id"]
 }
 
-post_rename_template_folder_schema = {
+post_update_template_folder_schema = {
     "$schema": "http://json-schema.org/draft-04/schema#",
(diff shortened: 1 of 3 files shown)


app/template_folder/template_folder_schema.py
```diff
@@ -11,9 +11,9 @@
     "required": ["name", "parent_id"]
 }
 
-post_rename_template_folder_schema = {
+post_update_template_folder_schema = {
     "$schema": "http://json-schema.org/draft-04/schema#",
-    "description": "POST schema for renaming template_folder",
+    "description": "POST schema for updating template_folder",
```
leohemsted: if we're making this generic, `name` shouldn't be required

allait: I think so, since it's the only field at the moment and otherwise it's an empty request

PR #2230, opened 2018-11-16
Delete run letter jobs

servingUpAces:
When we first built letters you could only send them via a CSV upload, initially we needed a way to send those files to dvla per job.
We since stopped using this page. So let's delete it!


tests/app/v2/notifications/test_post_letter_notifications.py
```diff
@@ -480,7 +475,7 @@ def test_post_precompiled_letter_notification_returns_201(client, notify_user, m
 
     assert response.status_code == 201, response.get_data(as_text=True)
 
-    s3mock.assert_called_once_with(ANY, b'letter-content', precompiled=True)
+    # s3mock.assert_called_once_with(ANY, b'letter-content', precompiled=True)
```
leohemsted: how come this one's commented out?

servingUpAces: Not sure - it's back in now

PR #2244, opened 2018-11-30
Add reference to one off letters

quis:
Letters should always have a reference, because that’s what DVLA use to tell us when they’ve sent a letter.

If a letter has a reference of `None` then DVLA say they’ve sent a letter with a reference of `'None'`. This means we can never reconcile the letter, which means it stays in `created`, which means it never
gets billed.

We don’t think this has affected any real letters yet, just ones that we’ve sent as tests.

### DIFF
--- a/app/service/send_notification.py
+++ b/app/service/send_notification.py
@@ -1,5 +1,6 @@
 from sqlalchemy.orm.exc import NoResultFound
 
+from app import create_random_identifier
 from app.config import QueueNames
 from app.dao.notifications_dao import _update_notification_status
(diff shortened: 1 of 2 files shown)


app/service/send_notification.py
```diff
@@ -37,6 +38,12 @@ def validate_created_by(service, created_by_id):
         raise BadRequestError(message=message)
 
 
+def create_one_off_reference(template_type):
+    if template_type != LETTER_TYPE:
```
servingUpAces: It doesn't make a difference, but I prefer to write/read if statement as positive.

PR #2250, opened 2018-12-05
Update switch providers on slow delivery method and query

CrystalPea:
Work done:
- update the `is_delivery_slow_for_provider` query. It now: finds all notifications with status "delivered" or "sending" for current provider (and key not a test key) for the last 10 minutes; looks how many of those are slow, and if it is above the threshold of slow delivery for provider;
- update `switch_current_sms_provider_on_slow_delivery` task to use the changed query with correct arguments, with slow notification being delivered after more than 4 minutes and threshold being 10%. The task only runs the query if the provider has not been switched in the last 10 minutes
- rewrite the tests so they test the new flow
- comment out actual switching providers bit for now so we can monitor how the function behaves

Pivotal ticket: https://www.pivotaltracker.com/story/show/162170074


app/celery/scheduled_tasks.py
```diff
@@ -256,28 +256,25 @@ def switch_current_sms_provider_on_slow_delivery():
     Switch providers if there are at least two slow delivery notifications (more than four minutes)
     in the last ten minutes. Search from the time we last switched to the current provider.
     """
-    functional_test_provider_service_id = current_app.config.get('FUNCTIONAL_TEST_PROVIDER_SERVICE_ID')
-    functional_test_provider_template_id = current_app.config.get('FUNCTIONAL_TEST_PROVIDER_SMS_TEMPLATE_ID')
```
leohemsted: these can now be removed from config.py entirely right

CrystalPea: Good find, I removed those from config and the tests now.

PR #2264, opened 2018-12-14
Add choose_postage service permission and add postage to Template

CrystalPea:
A first PR for this story, it prepares our database for enabling users to choose postage at a template level.

Pivotal ticket: https://www.pivotaltracker.com/story/show/160231772


app/models.py
```diff
@@ -762,6 +764,14 @@ def __init__(self, **kwargs):
     archived = db.Column(db.Boolean, nullable=False, default=False)
     hidden = db.Column(db.Boolean, nullable=False, default=False)
     subject = db.Column(db.Text)
+    postage = db.Column(db.String, nullable=True)
+    CheckConstraint("""
+        CASE WHEN template_type = 'letter' THEN
+            postage in ('first', 'second', null)
```
leohemsted: this won't work like you would expect, because null doesn't equal null. (see `nan poisoning`). The only operation you should do with null is `is` or `is not`

```
notification_api=# select null is null;
 ?column?
----------
 t
(1 row)

notification_api=# select null in (null);
 ?column?
----------

(1 row)
```

app/models.py
```diff
@@ -762,6 +764,14 @@ def __init__(self, **kwargs):
     archived = db.Column(db.Boolean, nullable=False, default=False)
     hidden = db.Column(db.Boolean, nullable=False, default=False)
     subject = db.Column(db.Text)
+    postage = db.Column(db.String, nullable=True)
+    CheckConstraint("""
```
leohemsted: Constraints defined in python are only relevant if we construct the database from python (which we don't do). It's good to have it here so we can refer to it without having to dig out migration scripts, but you'll need to also call alter table in the migration script

CrystalPea: ahh, cool, good to know, I will look how  this was done when postage was added to Notification 👍

app/models.py
servingUpAces: Does this need to be a check constraint or just a typed column?

CrystalPea: I'm not sure, I followed pattern from class NotificationBase where postage column has a check constraint...

PR #2269, opened 2018-12-21
Choose postage on template

CrystalPea:
Work done:
- Don't allow to set postage per template if no service permission
- Choose postage when persisting a notification
- Update v2 template schema to include postage
- Include postage in checking if template changed

Pivotal ticket: https://www.pivotaltracker.com/story/show/160231772

### DIFF
--- a/tests/app/template/test_rest.py
+++ b/tests/app/template/test_rest.py
@@ -17,6 +17,7 @@ from app.models import (
     EMAIL_TYPE,
     LETTER_TYPE,
     SMS_TYPE,
+    CHOOSE_POSTAGE,
     Template,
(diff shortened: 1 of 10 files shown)


tests/app/template/test_rest.py
```diff
@@ -210,6 +210,34 @@ def test_should_raise_error_on_create_if_no_permission(
     assert json_resp['message'] == expected_error
```
servingUpAces: Can we have a test for creating a template when service has CHOOSE_POSTAGE permission and postage is set for template?

servingUpAces: And a similar test for updating a template

PR #2272, opened 2018-12-28
log more info when we receive multiple delivery callbacks

leohemsted:
Previously, we logged a warning containing the notification reference and new status. However it wasn't a great message - this new one includes the notification id, the old status, the time difference and more.

This also separates out logs for callbacks for notifications we don't know (error level) and duplicates (info level).

### DIFF
--- a/app/notifications/process_client_response.py
+++ b/app/notifications/process_client_response.py
@@ -81,10 +81,6 @@ def _process_for_status(notification_status, client_name, provider_reference):
         sent_by=client_name.lower()
     )
     if not notification:
-        current_app.logger.warning("{} callback failed: notification {} either not found or already updated "
-                                   "from sending. Status {}".format(client_name,
-                                                                    provider_reference,
-                                                                    notification_status))
         return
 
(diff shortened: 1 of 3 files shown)


app/notifications/process_client_response.py
```diff
@@ -81,10 +81,13 @@ def _process_for_status(notification_status, client_name, provider_reference):
         sent_by=client_name.lower()
     )
     if not notification:
-        current_app.logger.warning("{} callback failed: notification {} either not found or already updated "
-                                   "from sending. Status {}".format(client_name,
-                                                                    provider_reference,
-                                                                    notification_status))
+        current_app.logger.debug(
+            "{} callback failed: notification {} either not found or already updated from sending. Status {}".format(
```
CrystalPea: After separate logger message was made for duplicate updates, does this part of the message still make sense?  '...or already updated from sending'

leohemsted: yeah we don't need this any more i think

PR #2274, opened 2019-01-02
fix multi worker exit script

leohemsted:
previously the script would:

try and SIGTERM each celery process every second for the 9 second
timeout, and then SIGKILL every second after, with no upper bound.

This commit changes this to:
* SIGTERM each process once.
* Wait nine seconds (checking if the pid files are still present each
  second)
* SIGKILL any remaining processes once.
* exit

### DIFF
--- a/scripts/run_multi_worker_app_paas.sh
+++ b/scripts/run_multi_worker_app_paas.sh
@@ -39,13 +39,15 @@ log_stream_name = {hostname}
 EOF
 }
 
-# For every PID, check if it's still running
-# if it is, send the sigterm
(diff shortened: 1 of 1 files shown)


scripts/run_multi_worker_app_paas.sh
```diff
@@ -39,42 +39,30 @@ log_stream_name = {hostname}
 EOF
 }
 
-# For every PID, check if it's still running
-# if it is, send the sigterm
+# For every PID, check if it's still running. if it is, send the sigterm. then wait 9 seconds before sending sigkill.w
```
sakisv: Nitpicking: trailing "w"

PR #2277, opened 2019-01-04
Use Notification's serialize method for .get_all_notifications_for_service endpoint

CrystalPea:
This PR supersedes the call to marshmallow schema with a call to Notification's serialize method. This is because:
1. we are trying to go away from using the schemas
2. we want to adjust letter notification statuses and Notification's serialize does it for us

As a part of this work a `for_service` flag has been added to serialize method so unnecessary fields are not returned when not needed.

Pivotal ticket: https://www.pivotaltracker.com/story/show/162950719

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1446,7 +1446,7 @@ class Notification(db.Model):
 
         return serialized
 
-    def serialize(self):
+    def serialize(self, for_service=False):
         template_dict = {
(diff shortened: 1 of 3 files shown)


CrystalPea:
We decided to do a more shallow fix for now, as this one was overcomplicating things.

PR #2292, opened 2019-01-10
Remove cancelled from requested statuses in service statistics

CrystalPea:
Pivotal ticket: https://www.pivotaltracker.com/story/show/162310324

Connected admin PR: https://github.com/alphagov/notifications-admin/pull/2646

### DIFF
--- a/app/service/statistics.py
+++ b/app/service/statistics.py
@@ -84,7 +84,8 @@ def create_zeroed_stats_dicts():
 
 
 def _update_statuses_from_row(update_dict, row):
-    update_dict['requested'] += row.count
+    if row.status != 'cancelled':
+        update_dict['requested'] += row.count
     if row.status in ('delivered', 'sent'):
         update_dict['delivered'] += row.count
     elif row.status in (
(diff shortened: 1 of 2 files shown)


app/service/statistics.py
```diff
@@ -84,7 +84,8 @@ def create_zeroed_stats_dicts():
 
 
 def _update_statuses_from_row(update_dict, row):
-    update_dict['requested'] += row.count
+    if row.status is not 'cancelled':
```
klssmith: This line needs to be `!= 'cancelled` instead of `is not 'cancelled'` in order to show the right numbers in Admin

CrystalPea: Hahah, I had a one-line job and I blundered, that's cute xD. Now I read up and I even know why:
"The keyword `is` relates to two objects being the exact same object, whereas `==` and `!=` deal with two objects being equivalent but not necessarily the same object"

Thanks for thorough testing 💖.

klssmith: Confusingly the test was making it look like the version with `is not 'cancelled'` _was_ working since Python is caching the 'cancelled' string object and using the exact same object in the test and the function 🤯

CrystalPea: Wow! 😮

PR #2293, opened 2019-01-10
Choose postage on POST request for precompiled letters

CrystalPea:
This PR enables adding optional 'postage' attribute to a POST request to 'post_precompiled_letter_notification' endpoint and sets notification postage based on this attribute if it's present. The attribute value needs to be either 'first' or 'second'.

Pivotal ticket: https://www.pivotaltracker.com/story/show/160231957

### DIFF
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -75,7 +75,8 @@ def persist_notification(
     created_by_id=None,
     status=NOTIFICATION_CREATED,
     reply_to_text=None,
-    billable_units=None
+    billable_units=None,
+    postage=None
 ):
     notification_created_at = created_at or datetime.utcnow()
     if not notification_id:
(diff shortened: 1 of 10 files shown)


app/notifications/process_notifications.py
```diff
@@ -112,11 +113,14 @@ def persist_notification(
...
-        template = dao_get_template_by_id(template_id, template_version)
-        if service.has_permission(CHOOSE_POSTAGE) and template.postage:
-            notification.postage = template.postage
+        if postage:
+            notification.postage = postage
         else:
-            notification.postage = service.postage
+            template = dao_get_template_by_id(template_id, template_version)
+            if service.has_permission(CHOOSE_POSTAGE) and template.postage:
+                notification.postage = template.postage
```
servingUpAces: The template.postage should be passed into this method, we already have the template before we call this method.  That will save us a db call.

CrystalPea: I can do it, but then I would have to add that argument in many places that call `persist_notification`, and check that all those places have template object:

screen shot 2019-01-11 at 14 40 18

I didn't introduce this call to db for this particular story, I just changed indentation.

CrystalPea: Anyway, let me know.

servingUpAces: If we can avoid a call to the db we should.

CrystalPea: ok, this has been done now 👍

PR #2298, opened 2019-01-11
Use ft_notification_status for monthly template usage

servingUpAces:
The template monthly usage page is using stats_template_usage_by_month table. The nightly job recalculates the template monthly usage in this table every night which takes about 17 minutes. 
This PR stops using stats_template_usage_by_month, using ft_notification_status instead. 
Another PR will follow to remove the nightly job and table.


app/dao/fact_notification_status_dao.py
```diff
@@ -291,3 +291,81 @@ def fetch_stats_for_all_services_by_date_range(start_date, end_date, include_fro
...
+            all_stats_table.c.template_type,
+            all_stats_table.c.month,
+            all_stats_table.c.year,
+        ).order_by(
+            all_stats_table.c.year,
+            all_stats_table.c.month,
+            all_stats_table.c.name
+        )
+    else:
+        query = stats
```
leohemsted: if the range doesn't cover today, then the query won't be ordered. is this okay?

tests/app/dao/test_fact_notification_status_dao.py
```diff
@@ -338,3 +339,53 @@ def test_fetch_stats_for_all_services_by_date_range(notify_db_session):
...
+                                  count=2)
+    create_ft_notification_status(bst_date=date(2018, 2, 1),
+                                  service=sample_service,
+                                  template=template_two,
+                                  count=4)
+    create_ft_notification_status(bst_date=date(2018, 3, 1),
+                                  service=sample_service,
+                                  template=template_three,
+                                  count=5)
+    create_notification(template=template_one)
```
leohemsted: can we add a notification from yesterday for one of the other templates so that we can make sure that one isn't picked up?

app/dao/fact_notification_status_dao.py
servingUpAces: No - good point, I'll change that, and add a test

app/service/rest.py
```diff
@@ -579,10 +579,16 @@ def resume_service(service_id):
...
-        data = dao_fetch_monthly_historical_usage_by_template_for_service(
-            service_id,
-            int(request.args.get('year', 'NaN'))
+        start_date, end_date = get_financial_year(int(request.args.get('year', 'NaN')))
+        data = fetch_monthly_template_usage_for_service(
+            start_date=start_date,
+            end_date=end_date,
+            service_id=service_id
         )
+        # data = dao_fetch_monthly_historical_usage_by_template_for_service(
```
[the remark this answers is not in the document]
servingUpAces: I just forgot to delete it

PR #2307, opened 2019-01-15
Handle celery PIDs more reliably

sakisv:
This addresses some problems that existed in the previous approach:

1. There was a race condition that could occur between the time we were
looking for the existence of the .pid files and actually reading them.

2. If for some reason the .pid file was left behind after a process had
died, the script would never know because we do:

```
    kill -s ${1} ${APP_PID} || true
```


sakisv:
This has been tested on preview and can be merged.

PR #2311, opened 2019-01-18
Update gsi email addresses

servingUpAces:
New server command to remove `.gsi` from the email address of all users for a given service.


app/commands.py
```diff
@@ -634,3 +634,29 @@ def update_jobs_archived_flag(start_date, end_date):
...
+                          WHERE s.id = :service_id
+                            AND u.email_address ilike ('%.gsi%')
+    """
+    results = db.session.execute(users_to_update, {'service_id': service_id})
+    print("Updating {} users.".format(results.rowcount))
+
+    for user in results:
+        print(user)
+
+        update_stmt = """UPDATE users set email_address = replace(email_address, '.gsi', ''), updated_at = now()
```
quis: Should be case-insensitive to handle `GSI.GOV.UK`

quis: Also might be safer to replace `.gsi.gov.uk` with `.gov.uk`

servingUpAces: Ah right, the select was but I missed it on the update.

app/commands.py
```diff
@@ -634,3 +634,30 @@ def update_jobs_archived_flag(start_date, end_date):
...
+                           JOIN user_to_service us on (u.id = us.user_id)
+                           JOIN services s on (s.id = us.service_id)
+                          WHERE s.id = :service_id
+                            AND u.email_address ilike ('%.gsi.gov.uk%')
+    """
+    results = db.session.execute(users_to_update, {'service_id': service_id})
+    print("Updating {} users.".format(results.rowcount))
+
+    for user in results:
+        print(user)
```
allait: I think the logs from cf-task might get shipped to Logit, so I'd try to avoid printing the email address and user name here

klssmith: Updated in 510e6bd to just print the user id

PR #2318, opened 2019-01-23
Letter branding

servingUpAces:
# Why are we doing this
At the moment if we need to add a logo for letters, we need to create a png file from a svg file, upload both files to S3 (in all environments), add a row to dvla_oganisations (in a data migration), then update the service with the new logo. To get rid of this manual process we are going to create a similar process as email branding. More details in the story: https://www.pivotaltracker.com/story/show/163230257 for more information.

# What has been done so far:
- LetterBranding model has been added. The domain field of the LetterBranding will need to be updated manually by a platform admin user once the story is complete. 
- ServiceLetterBranding model has been added to map a service to a letter brand. 
- When a new service is created add the ServiceLetterBranding row if `service_domain` exists in request then fetch the brand for that domain and add the association. 
- "letter_branding" existing in the update_services request, a letter branding relationship is created or updated. 
- Where service.dvla_organisation.filename is used return the letter_branding if not null else dvla_organisation. 

# Deployment plan
Because this story is not populating the tables we can deploy this now.
A full deployment plan will be added to the story. https://www.pivotaltracker.com/story/show/163230257


app/dao/letter_branding_dao.py
```diff
@@ -0,0 +1,18 @@
...
+
+def get_letter_branding_or_platform_default(domain=None):
+    letter_branding = None
+    if domain:
+        letter_branding = LetterBranding.query.filter(
+            LetterBranding.domain == domain
+        ).first()
+    if not letter_branding:
+        letter_branding = LetterBranding.query.filter(
+            LetterBranding.platform_default == True  # noqa
```
CrystalPea: this feels a bit odd. If I understand correctly, we only have one platform default letter branding that is always the same, right? If this is the case, do we have to query the whole table to find it?

I mean I guess it's not an expensive query since the table is tiny, so maybe it's ok...

servingUpAces: It's either that or have something in config. This seems to make sense, and it is a small table.

CrystalPea: ok 👍

app/letter_branding/letter_branding_rest.py
```diff
@@ -0,0 +1,84 @@
...
+    Handle integrity errors caused by the unique constraint
+    """
+    if 'domain' in str(exc):
+        return jsonify(
+            result='error',
+            message={'name': ["Duplicate domain '{}'".format(
+                exc.params.get('domain')
+            )]}
+        ), 400
+    if 'name' in str(exc):
```
leohemsted: i'm going to have a look at what `str(exc)` actually returns because this check is pretty wide-reaching and scares me a little

leohemsted: update: changed this (it was erroneously catching `filename` under the `name` param for example)

PR #2319, opened 2019-01-24
[wip] Use supervisord instead of homerolled supervision

tlwr:
Uses supervisord to supervise celery and awslogs

If any of the processes are stopped it will restart them

If they are stopped > 3 times in < 15s then supervisord will send an
event to kill_supervisor.py which will kill supervisord

### DIFF
--- a/kill_supervisor.py
+++ b/kill_supervisor.py
@@ -1,0 +1,26 @@
+import os
+import sys
+
+def write_stdout(s):
+    sys.stdout.write(s)
+    sys.stdout.flush()
+
+def write_stderr(s):
+    sys.stderr.write(s)
+    sys.stderr.flush()
+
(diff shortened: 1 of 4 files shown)


tlwr:
This was more of a spike, I got it to work on PaaS using the current buildpack though

tlwr:
The make task fails even though it works in PaaS

The best thing to do IMO would be to use https://github.com/ochinchina/supervisord but this is out of scope

Closing

mnaberez:
> supervisor currently doesn't support python 3 (which is what we're running on PaaS) so we can't add it as a requirement. 

[Supervisor 4.0.0](https://pypi.org/project/supervisor/) (PyPI package; supports Python 3.4+ and 2.7)

PR #2342, opened 2019-02-06
Set postage on basis of postage argument or template postage

CrystalPea:
This PR will make choosing postage live for all our users.

To be released together with this PR: https://github.com/alphagov/notifications-admin/pull/2735

Part of this story: https://www.pivotaltracker.com/story/show/162959391


app/v2/notifications/post_notifications.py
```diff
@@ -355,6 +355,7 @@ def get_precompiled_letter_template(service_id):
         hidden=True,
         subject='Pre-compiled PDF',
         content='',
+        postage='second'
```
allait: Let's use SECOND_CLASS here so that the defaults are easier to find

CrystalPea: done 👍

PR #2344, opened 2019-02-07
Return rendered HTML when previewing a template

quis:
If you’re trying to show what a Notify email will look like in your caseworking system all the API gives you at the moment is raw markdown (with the placeholders replaced).

This isn’t that useful if your caseworkers have no idea what markdown is. If we also give teams the HTML then they can embed this in their systems, and the people using those systems will be able to see how headings, bulleted lists, etc. look.

***

The first users of this are working in Java. They’ve said they’re happy to update the client to add this new field.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -3,7 +3,7 @@ from datetime import datetime, timedelta
 import pytz
 from flask import url_for
 from sqlalchemy import func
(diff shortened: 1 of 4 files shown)


quis:
@leohemsted Good call, added in 50074563

PR #2349, opened 2019-02-12
Don't fallback to dvla_organisation if letter branding unset

leohemsted:
The template preview app now accepts a null value for the `filename` parameter. If a service doesn't have a letter branding option set,  previously we defaulted to their dvla_organisation (probably HM  Government). Now, we pass through None, so that we generate letters without any logo or branding.

Tested by checking file on s3.

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -205,7 +205,7 @@ class ServiceSchema(BaseSchema):
     created_by = field_for(models.Service, 'created_by', required=True)
     organisation_type = field_for(models.Service, 'organisation_type')
     dvla_organisation = field_for(models.Service, 'dvla_organisation')
-    letter_logo_filename = fields.Method(serialize='get_letter_logo_filename')
+    letter_logo_filename = fields.Method(dump_only=True, serialize='get_letter_logo_filename')
     permissions = fields.Method("service_permissions")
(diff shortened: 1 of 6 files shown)


app/schemas.py
```diff
@@ -205,16 +205,15 @@ class ServiceSchema(BaseSchema):
...
     permissions = fields.Method("service_permissions")
     email_branding = field_for(models.Service, 'email_branding')
     organisation = field_for(models.Service, 'organisation')
     override_flag = False
     letter_contact_block = fields.Method(serialize="get_letter_contact")
 
     def get_letter_logo_filename(self, service):
-        return service.letter_branding.filename if service.letter_branding\
-            else service.dvla_organisation.filename
+        return service.letter_branding and service.letter_branding.filename
```
CrystalPea: does this method return a boolean now? If yes, should it maybe be called differently?

servingUpAces: The method returns `None` If no letter_branding has been set or the filename if it has been set.

CrystalPea: ahhh, that's really cool, I didn't know that `and` statements could return a string :o

CrystalPea: I found an explanation:
https://stackoverflow.com/questions/4477850/python-and-or-operators-return-value

today I've learnt 😮

PR #2352, opened 2019-02-13
 remove dvla_organisation from database

leohemsted:
downgrade re-populates all static data but leaves all services with 001 (HM Government) branding

- [x] #2349 
- [x] #2351


migrations/versions/0258_remove_dvla_organisation.py
```diff
@@ -0,0 +1,84 @@
...
+Revision ID: 0258_remove_dvla_organisation
+Revises: 0257_letter_branding_migration
+Create Date: 2019-02-12 17:39:02.517571
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = '0258_remove_dvla_organisation'
```
servingUpAces: There is another 0258 migration -- doh!

PR #2353, opened 2019-02-13
Add retries for SES callbacks for recent notifications

allait:
We've seen errors caused by what we suspect is a race condition when
SES callback processing tries to look up the notification before the
sender worker has saved notification reference from the SES POST
response to the database.

This adds a retry for SES callback task if the notification was not
found and the message is less than 10 minutes old and removes the
error log message for notifications older than 3 days (since they
might no longer exist in the notifications table and would've been
marked as failure by then either way).

In order to be able to call retry and silence the error log based on
notification time this change inlines `process_ses_response` and
`update_notification_by_reference` functions into the celery task.
It also removes a lot of defensive error-handling that doesn't appear
to have been triggered in the last few months (for things like missing
keys in SES callback data).

### DIFF
--- a/app/celery/process_ses_receipts_tasks.py
+++ b/app/celery/process_ses_receipts_tasks.py
@@ -1,19 +1,80 @@
-from flask import current_app
+from datetime import datetime, timedelta
+
+import iso8601
+from flask import current_app, json
 from notifications_utils.statsd_decorators import statsd
+from sqlalchemy.orm.exc import NoResultFound
 
-from app import notify_celery
+from app import notify_celery, statsd_client
 from app.config import QueueNames
+from app.clients.email.aws_ses import get_aws_responses
+from app.dao import notifications_dao
+from app.models import NOTIFICATION_SENDING, NOTIFICATION_PENDING
 
-from app.notifications.notifications_ses_callback import process_ses_response
+from app.notifications.notifications_ses_callback import (
+    determine_notification_bounce_type,
+    handle_complaint,
+    _check_and_queue_complaint_callback_task,
+    _check_and_queue_callback_task,
+)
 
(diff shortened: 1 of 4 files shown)


app/celery/process_ses_receipts_tasks.py
```diff
@@ -1,19 +1,80 @@
...
+
+        aws_response_dict = get_aws_responses(notification_type)
+
+        notification_status = aws_response_dict['notification_status']
+        reference = ses_message['mail']['messageId']
+
+        try:
+            notification = notifications_dao.dao_get_notification_by_reference(reference)
+        except NoResultFound:
+            message_time = iso8601.parse_date(ses_message['mail']['timestamp']).replace(tzinfo=None)
```
servingUpAces: Is it guaranteed that the timestamp is in UTC?

allait: All the examples I've seen have a zero offset, so I think so. SES documentation doesn't mention it, but I think it might be implied.

tests/app/celery/test_process_ses_receipts_tasks.py
```diff
@@ -1,40 +1,38 @@
...
 from app.notifications.notifications_ses_callback import remove_emails_from_complaint, remove_emails_from_bounce
 
 from tests.app.db import (
-    create_notification, ses_complaint_callback,
-    ses_notification_callback,
+    create_notification,
+    ses_complaint_callback,
+    create_service_callback_api,
 )
+from tests.app.conftest import sample_notification as create_sample_notification
```
leohemsted: we want to remove these, not add them. Does `tests.app.db.create_notification` not work for you?

PR #2355, opened 2019-02-13
Add fields to record a service’s estimated volumes

quis:
When a service go live we ask people for their estimated sending volumes. At the moment we only put this in the ticket, and store it in a spreadsheet.

This means that a service can
- say they want to go live
- say they are sending 100,000 emails per year
- not have created any email templates
- still see ‘create templates’ as ‘completed’ in the go live checklist

If we store this data against the service we can collect it earlier, and then use it to determine automatically what kind of templates the user needs to create before their go live checklist can be considered complete.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -361,6 +361,9 @@ class Service(db.Model, Versioned):
     crown = db.Column(db.Boolean, index=False, nullable=False, default=True)
     rate_limit = db.Column(db.Integer, index=False, nullable=False, default=3000)
     contact_link = db.Column(db.String(255), nullable=True, unique=False)
+    volume_sms = db.Column(db.String(255), nullable=True, unique=False)
+    volume_email = db.Column(db.String(255), nullable=True, unique=False)
+    volume_letter = db.Column(db.String(255), nullable=True, unique=False)
 
     organisation = db.relationship(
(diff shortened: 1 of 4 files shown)


migrations/versions/0259_service_volumes.py
```diff
@@ -0,0 +1,30 @@
...
+
+TABLES_AND_CHANNELS = product(
+    ('services', 'services_history'),
+    ('volume_{}'.format(channel) for channel in ('email', 'letter', 'sms')),
+)
+
+
+def upgrade():
+    for table, channel in TABLES_AND_CHANNELS:
+        op.add_column(table, sa.Column(channel, sa.String(length=255), nullable=True))
```
servingUpAces: Why a string?

servingUpAces: There is already a 0259 migration :(

quis: @servingUpAces changed to integer in 68ed2c5

leohemsted: @quis sorry there's now a 260 migration too

quis: Bumped version in 7ca62ce

PR #2356, opened 2019-02-14
Add folder permissions permission

allait:
[#163755302](https://www.pivotaltracker.com/story/show/163755302)

Adds a new temporary service permission for enabling folder permission
editing.


migrations/versions/0260_add_folder_permissions_perm.py
```diff
@@ -0,0 +1,21 @@
+"""
+
+Revision ID: 0260_add_folder_permissions_perm
```
quis: Needs to be `262` now 😥

allait: Rebased and reordered

PR #2375, opened 2019-02-26
Add user folder permissions

klssmith:
* Added a `user_folder_permissions` table which contains the `service_id`, `user_id` and `template_folder_id`. Also changed the `user_to_service` mapping table into a model, `ServiceUser`, since we're often only interested in the users for a given service.
* Allow user folder permissions to be updated
* Include list of `users_with_permission` in all folders response
* Delete user folder permissions when a user is removed from a service

[Pivotal story](https://www.pivotaltracker.com/story/show/164050120)

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -754,7 +754,8 @@ class TemplateFolder(db.Model):
             'id': self.id,
             'name': self.name,
             'parent_id': self.parent_id,
-            'service_id': self.service_id
+            'service_id': self.service_id,
+            'users_with_permission': self.get_users_with_permission()
         }
 
     def is_parent_of(self, other):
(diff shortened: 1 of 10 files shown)


app/models.py
```diff
@@ -743,6 +758,20 @@ def is_parent_of(self, other):
             other = other.parent
         return False
 
+    def get_users_with_permission(self):
+        service_users = self.users
+        users_with_permission = [str(service_user.user_id) for service_user in service_users]
+
+        return users_with_permission
+
+    users = db.relationship(
```
allait: I'd move this up to where the previous relationship was defined. I think we normally try to keep field definitions above method ones.

PR #2377, opened 2019-02-26
stub out antivirus in dev

leohemsted:
antivirus is sometimes tough to get running locally - now in dev antivirus is skipped unless `ANTIVIRUS_ENABLED=1` is set on the command line. on all other environments it is always enabled.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -108,6 +108,9 @@ class Config(object):
     CRONITOR_ENABLED = False
     CRONITOR_KEYS = json.loads(os.environ.get('CRONITOR_KEYS', '{}'))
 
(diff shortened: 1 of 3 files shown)


app/config.py
```diff
@@ -354,6 +357,8 @@ class Development(Config):
     STATSD_PORT = 1000
     STATSD_PREFIX = "stats-prefix"
 
+    ANTIVIRUS_ENABLED = os.getenv('ANTIVIRUS_ENABLED') == 1
```
allait: `.getenv` would always return a string, so this should compare to `'1'`

PR #2385, opened 2019-03-06
Raise exception if history can’t be written

quis:
This is fiendishly difficult problem to discover on your own.

It’s caused when, during the creation of a row in the database, you run a query on the same table, or a table that joins to the table you’re inserting into. What I think is happening is that the database is forced to flush the session before running the query in order to maintain consistency.

This means that the session is clean by the time the history stuff comes to do its work, so there’s nothing for it to copy into the history table, and it silently fails to record history.

Hopefully raising an exception will:
- prevent this from failing silently
- save whoever comes across this issue in the future a whole load of time

### DIFF
--- a/app/dao/dao_utils.py
+++ b/app/dao/dao_utils.py
@@ -1,5 +1,5 @@
 import itertools
-from functools import wraps, partial
+from functools import wraps
 
 from app import db
 from app.history_meta import create_history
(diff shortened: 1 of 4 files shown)


quis:
Can’t work out why the tests are failing 😿

quis:
# Do all version table writes in one commit

This way of stacking the version decorators does not work as expected:
```
@version_class(A)
def foo(…):

What you would expect to happen is that each decorator causes a history row to be written for its respective model object.

What actually happens is that the first decorator adds history records
to the database session, but then causes the database session to commit.
This means that subsequent uses of this decorator find a clean session,
and therefore no changes to copy to their respective history tables.

This commit changes the intended use of the decorator so that it is only
used once per function, and accepts multiple definitions of what to
record history for. This way it can record everything that needs to go
into the history before doing anything that would risk flushing the
session.

**Comment by quis** [comment]
# Do all version table writes in one commit

This way of stacking versionning decorators does not work as expected:

```python
@version_class(A)
@version_class(B)
def foo():
  pass
```

What you would expect to happen is that each decorator causes a history row to be written for its respective model object.

What actually happens is that the first decorator adds history records to the database session, but then causes the database session to commit. This means that subsequent uses of this decorator find a clean session, and therefore no changes to copy to their respective history tables.

This commit changes the intended use of the decorator so that it is only used once per function, and accepts multiple definitions of what to record history for. This way it can record everything that needs to go into the history before doing anything that would risk flushing the session.

PR #2396, opened 2019-03-14
fix constraint name check in service error handler

leohemsted:
couldn't think of a good way to test it because the problem was it was suppressing actual 500 internal server errors

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -118,7 +118,10 @@ def handle_integrity_error(exc):
     """
     Handle integrity errors caused by the unique constraint on ix_organisation_name
     """
-    if 'services_name_key' or 'services_email_from_key' in str(exc):
+    if any(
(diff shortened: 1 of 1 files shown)


app/service/rest.py
```diff
@@ -118,7 +118,10 @@ def handle_integrity_error(exc):
     """
     Handle integrity errors caused by the unique constraint on ix_organisation_name
     """
-    if 'services_name_key' or 'services_email_from_key' in str(exc):
+    if any(
+        'duplicate key value violates unique constraint "{}"'.format(constraint) in str(exc)
+        for constraint in {'services_name_key', 'services_email_from_key'}
```
allait: Why is this a set?

leohemsted: i'm not super opinionated but i think its a reasonable habit to get into to use sets rather than lists for unordered distinct sequences

haven't really thought too much about it in this case though 🤷‍♂️

PR #2399, opened 2019-03-15
Reduce the number of files sent to the zip-and-send-letter-pdfs task

servingUpAces:
The ftp application is struggling, running out of CPU. This is attempt to help with that.


leohemsted:
👍

PR #2419, opened 2019-03-26
Split task for updating letters

servingUpAces:
In this PR the action to update letter statuses and count the daily sorted counts has been separated. 
There is also a new command to go through each file sent by DVLA and record the daily sorted counts.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -408,39 +408,49 @@ def get_template_class(template_type):
 @notify_celery.task(bind=True, name='update-letter-notifications-statuses')
 @statsd(namespace="tasks")
 def update_letter_notifications_statuses(self, filename):
+    notification_updates = parse_dvla_file(filename)
+
+    temporary_failures = []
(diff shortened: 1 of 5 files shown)


app/celery/tasks.py
```diff
@@ -408,39 +408,49 @@ def get_template_class(template_type):
 @notify_celery.task(bind=True, name='update-letter-notifications-statuses')
 @statsd(namespace="tasks")
 def update_letter_notifications_statuses(self, filename):
+    notification_updates = parse_dvla_file(filename)
+
+    temporary_failures = []
+
+    for update in notification_updates:
+        check_billable_units(update)
+        update_letter_notification(filename, temporary_failures, update)
```
CrystalPea: I think I don't understand something... is `temporary_failures` list somehow updated through that function call?

servingUpAces: yes - see what happens in `update_letter_notification`

PR #2421, opened 2019-03-27
make inbound sms page respect data retention

leohemsted:
also, it should default to last 7 days, not last 6 days. also change count_inbound_sms to have the days passed in, so that it's more explicit at the endpoint that we only return 7 days regardless of your service's data retention

### DIFF
--- a/tests/app/inbound_sms/test_rest.py
+++ b/tests/app/inbound_sms/test_rest.py
@@ -320,3 +320,17 @@ def test_get_most_recent_inbound_sms_for_service_respects_data_retention(
         '2017-04-06T12:00:00.000000Z',
         '2017-04-05T12:00:00.000000Z',
     ]
+
+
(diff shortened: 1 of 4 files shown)


tests/app/inbound_sms/test_rest.py
```diff
@@ -297,3 +297,26 @@ def test_get_most_recent_inbound_sms_for_service(
...
     assert len(response['data']) == expected_rows
     assert response['has_next'] == has_next_link
+
+
+@freeze_time('Monday 10th April 2017 12:00')
+def test_get_most_recent_inbound_sms_for_service_respects_data_retention(
+    admin_request,
+    sample_service
+):
+    create_service_data_retention(sample_service.id, 'sms', 5)
```
servingUpAces: Do we need a test that we get data that's older than a week?

PR #2424, opened 2019-03-29
Use ft_notification_status instead of notification_history

servingUpAces:
Update the nightly task that send performance platform statistics to use ft_notification_status rather than notification_history.

The previous query was including all notifications regardless of notification_status. This is okay because the performance platform is reporting how many notifications Notify is dealing with. 

I also need to remove the query that's no longer being used.

### DIFF
--- a/tests/app/dao/test_fact_notification_status_dao.py
+++ b/tests/app/dao/test_fact_notification_status_dao.py
@@ -538,11 +538,11 @@ def test_get_total_sent_notifications_for_day_and_type_returns_right_notificatio
         notification_type, count, sample_template, sample_email_template, sample_letter_template
 ):
     create_ft_notification_status(bst_date="2019-03-27", service=sample_template.service, template=sample_template,
-                                  count=count)
+                                  count=3)
     create_ft_notification_status(bst_date="2019-03-27", service=sample_email_template.service,
-                                  template=sample_email_template, count=count)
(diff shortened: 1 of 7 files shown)


tests/app/dao/test_fact_notification_status_dao.py
```diff
@@ -526,3 +528,48 @@ def test_fetch_monthly_template_usage_for_service_does_not_include_test_notifica
...
+                          ("letter", 7)])
+def test_get_total_sent_notifications_for_day_and_type_returns_right_notification_type(
+        notification_type, count, sample_template, sample_email_template, sample_letter_template
+):
+    create_ft_notification_status(bst_date="2019-03-27", service=sample_template.service, template=sample_template,
+                                  count=count)
+    create_ft_notification_status(bst_date="2019-03-27", service=sample_email_template.service,
+                                  template=sample_email_template, count=count)
+    create_ft_notification_status(bst_date="2019-03-27", service=sample_letter_template.service,
+                                  template=sample_letter_template, count=count)
```
leohemsted: i think these three lines should hard code the count to 3/5/7. as it currently reads, you're just setting everything to count=`x` and asserting that `x` things are returned each time

servingUpAces: Doh! I think that's what I meant to do, good catch.

PR #2426, opened 2019-04-01
Fix BST date bug for platform admin summary page.

servingUpAces:
Added test for that.


quis:
You’ll need to run this command for the missing days: https://github.com/alphagov/notifications-api/blob/master/app/commands.py#L188

tests/app/dao/test_fact_notification_status_dao.py
```diff
@@ -319,6 +320,42 @@ def test_fetch_notification_status_totals_for_all_services(
...
+    create_notification(email_template, created_at=datetime(2018, 4, 21, 14, 0, 0), status='delivered')
+
+    results = sorted(
+        fetch_notification_status_totals_for_all_services(
+            start_date=date(2018, 4, 21), end_date=date(2018, 4, 21)),
+        key=lambda x: (x.notification_type, x.status)
+    )
+
+    assert len(results) == 3
+    print(results)
```
klssmith: This `print` can be deleted

PR #2428, opened 2019-04-02
Give existing users all folder permissions

CrystalPea:
This will be done just before we turn on folder permissions to all services to ensure that all users start with all the permissions.

Pivotal ticket: https://www.pivotaltracker.com/story/show/164051002


quis:
(rebased to put this back in order)

quis:
(rebased again to update migration version)

PR #2433, opened 2019-04-03
fix create_nightly_billing date bug

leohemsted:
the `get_rate` function tries to compare the parameter `date` to a datetime field in the DB.

However we're now passing a `datetime.date` object in instead, so were getting errors. Tests used this fn as a boundary so didn't flag this up.

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -209,10 +209,8 @@ def fetch_billing_data_for_day(process_day, service_id=None):
 
 
 def get_rates_for_billing():
-    non_letter_rates = [(r.notification_type, r.valid_from, r.rate) for r in
-                        Rate.query.order_by(desc(Rate.valid_from)).all()]
-    letter_rates = [(r.start_date, r.crown, r.sheet_count, r.rate, r.post_class) for r in
-                    LetterRate.query.order_by(desc(LetterRate.start_date)).all()]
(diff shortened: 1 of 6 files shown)


tests/app/db.py
```diff
@@ -403,6 +405,23 @@ def create_rate(start_date, value, notification_type):
     return rate
 
 
+def create_letter_rate(start_date=None, end_date=None, crown=True, sheet_count=1, rate=0.33, post_class='second'):
```
leohemsted: not sure if this is the right approach - we aren't currently deleting things from letter rates in tests

leohemsted: decision: this is the right approach. we should update tests to delete letterrates between runs

app/dao/fact_billing_dao.py
```diff
@@ -209,10 +210,8 @@ def fetch_billing_data_for_day(process_day, service_id=None):
 
 
 def get_rates_for_billing():
-    non_letter_rates = [(r.notification_type, r.valid_from, r.rate) for r in
-                        Rate.query.order_by(desc(Rate.valid_from)).all()]
-    letter_rates = [(r.start_date, r.crown, r.sheet_count, r.rate, r.post_class) for r in
-                    LetterRate.query.order_by(desc(LetterRate.start_date)).all()]
+    non_letter_rates = Rate.query.order_by(desc(Rate.valid_from)).all()
```
servingUpAces: 👍

PR #2434, opened 2019-04-04
Add a method to guess organisation from a domain

quis:
We need this in the admin app while we still have pages that:
- talk about the data sharing and financial agreement
- but aren’t within a service (so can’t look at the service’s organisation)

This is a `GET`, but it deliberately won’t work if you pass it an email address, in order not to put personally identifying information in our logs.

### DIFF
--- a/app/organisation/rest.py
+++ b/app/organisation/rest.py
@@ -1,10 +1,11 @@
-from flask import Blueprint, jsonify, request, current_app
+from flask import abort, Blueprint, jsonify, request, current_app
 from sqlalchemy.exc import IntegrityError
 
 from app.dao.organisation_dao import (
     dao_create_organisation,
     dao_get_organisations,
     dao_get_organisation_by_id,
+    dao_get_organisation_by_email_address,
     dao_get_organisation_services,
     dao_update_organisation,
     dao_add_service_to_organisation,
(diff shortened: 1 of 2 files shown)


app/organisation/rest.py
```diff
@@ -53,6 +54,24 @@ def get_organisation_by_id(organisation_id):
...
 
+@organisation_blueprint.route('/by-domain', methods=['GET'])
+def get_organisation_by_domain():
+
+    domain = request.args.get('domain')
+
+    if not domain or '@' in domain:
+        abort(400)
+
+    organisation = dao_get_organisation_by_email_address(
```
servingUpAces: `dao_get_organisation_by_email_address` raises an exception if no org is found. So you may either change the query to first() rather than one() or deal with exception.
It seems possible that there could be more than one organisation that matches the domain. For example what happens if the domain = `gov.uk`

tests/app/organisation/test_rest.py
```diff
@@ -85,6 +85,41 @@ def test_get_organisation_by_id_returns_domains(admin_request, notify_db_session
     }
 
 
+@pytest.mark.parametrize('domain, expected_status', (
```
servingUpAces: You might consider a test when no organisation is found.

app/organisation/rest.py
quis: Do you mean this line?
https://github.com/alphagov/notifications-api/blob/c654bc1e5ee5ba946aef67d1b2cf86a2ebf05b19/app/dao/organisation_dao.py#L39

It will always return a result because every domain has an organisation.

If the domain doesn’t exist at all then it will finish the loop and return `None`:
https://github.com/alphagov/notifications-api/blob/c654bc1e5ee5ba946aef67d1b2cf86a2ebf05b19/app/dao/organisation_dao.py#L41

quis: For the case of more than one domain matching, it will match on the longest one first, because of the sort order:
https://github.com/alphagov/notifications-api/blob/c654bc1e5ee5ba946aef67d1b2cf86a2ebf05b19/app/dao/organisation_dao.py#L33

This means that if
- your email address ends in `cabinet-office.gov.uk`
- we have in the database `cabinet-office.gov.uk`, `office.gov.uk` and `gov.uk`

…it will find `cabinet-office.gov.uk` first.

tests/app/organisation/test_rest.py
quis: That’s the `404` I’m testing here:
https://github.com/alphagov/notifications-api/pull/2434/files#diff-3525e2f6804d98e0a7cbf330f454721cR91

app/organisation/rest.py
servingUpAces: Ok

PR #2435, opened 2019-04-04
Populate org data from file

servingUpAces:


app/commands.py
```diff
@@ -675,3 +677,48 @@ def replay_daily_sorted_count_files(file_extension):
...
+    # [3] argeement_signed:: TRUE | FALSE
+    # [4] domains:: comma separated list of domains related to the organisation
+
+    # The expectation is that the organisation, organisation_to_service
+    # and user_to_organisation will be cleared before running this command.
+    # Ignoring duplicates allows us to run the command again with the same file or same file with new rows.
+    with open(file_name, 'r') as f:
+        for line in itertools.islice(f, 1, None):
+            columns = line.split('|')
+            print(columns)
```
quis: I’d add some checks to make sure the data is what we expect:
```suggestion
            print(columns)
            assert columns[0]
            assert columns[1].lower in {'central', 'local', 'nhs'}
            assert columns[2] in {'TRUE', 'FALSE', ''}
            assert columns[3] in {'TRUE', 'FALSE', ''}
```

servingUpAces: I don't think it's necessary - the file works as is.

PR #2437, opened 2019-04-04
Precompiled request preview with overlay when validation fails

CrystalPea:
Has to go after this template-preview PR: https://github.com/alphagov/notifications-template-preview/pull/315

Part of this story: https://www.pivotaltracker.com/story/show/164798738

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -240,11 +240,10 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
 
         if overlay:
             path = '/precompiled/overlay.{}'.format(file_type)
-            query_string = '?invert=1'
             content = pdf_file
         elif file_type == 'png':
-            path = '/precompiled-preview.png'
             query_string = '?hide_notify=true' if page_number == '1' else ''
+            path = '/precompiled-preview.png' + query_string
         else:
             path = None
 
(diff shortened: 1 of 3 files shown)


app/template/rest.py
```diff
@@ -249,10 +249,18 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
...
-            url = '{}/precompiled-preview.png{}'.format(
-                current_app.config['TEMPLATE_PREVIEW_API_HOST'],
-                '?hide_notify=true' if page_number == '1' else ''
-            )
+            if request.args.get('overlay'):
+                content = pdf_page
+                url = '{}/precompiled/overlay.png{}{}'.format(
+                    current_app.config['TEMPLATE_PREVIEW_API_HOST'],
+                    '?invert=1',
+                    '&hide_notify=true' if page_number == '1' else ''
```
leohemsted: it doesn't look like overlay.png uses `hide_notify`, so that can be removed (and the invert bit inlined in the orig format string).

if we do intend to make that change at some point soon, then we could use `urllib.parse.urlencode` to create query strings from a dictionary and clean this function up a bit.

We could even go the whole hog and properly split it up and construct a `ParseResult` object

CrystalPea: I got rid of tag hiding arg, the rest sounds interesting although cryptic to me, I'd love to explore it together on Wednesday if you are up for it :)

PR #2439, opened 2019-04-05
Fix boolean columns for organisation migration

servingUpAces:
Update the migration so that if the boolean columns are set properly. 

'0' == False
'1' == True
'' == None


app/commands.py
```diff
@@ -700,10 +700,11 @@ def populate_organisations_from_file(file_name):
             data = {
                 'name': columns[0],
                 'active': True,
-                'agreement_signed': True if columns[3] == 'TRUE' else False,
-                'crown': True if columns[2] == 'TRUE' else False,
+                'agreement_signed': columns[3] == '1',
```
quis: There are three rows in the spreadsheet that have this cell blank. It should map those to a value of `None`, not `False`.

PR #2442, opened 2019-04-05
Set default branding for NHS services

quis:
The NHS is a special case because it’s not one organisation, but it does have one consistent brand. So anyone working for an NHS organisation should have their default branding set when they create a service, even if we know nothing about their specific organisation.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -122,3 +122,9 @@ def escape_special_characters(string):
             r'\{}'.format(special_character)
         )
     return string
+
+
+def email_address_is_nhs(email_address):
+    return email_address.lower().endswith((
+        '@nhs.uk', '@nhs.net', '.nhs.uk', '.nhs.net',
+    ))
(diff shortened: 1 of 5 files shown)


app/utils.py
```diff
@@ -122,3 +122,9 @@ def escape_special_characters(string):
             r'\{}'.format(special_character)
         )
     return string
+
+
+def email_address_is_nhs(email_address):
+    return email_address.lower().endswith((
+        '@nhs.uk', '@nhs.net', '.nhs.uk', '.nhs.net',
```
servingUpAces: Why can't this be ('nhs.uk',  'nhs.net')?

quis: Because that would also match `fake-nhs.net`. Although I guess it doesn’t really matter here because we’re just setting branding not ensuring the person has a government email address.

quis: Although thinking about it some more if we have a function called `email_address_is_nhs` then doing the more stringent check is better, in case anyone in the future was going to make use of the function elsewhere.

servingUpAces: fair enough

PR #2456, opened 2019-04-10
Remove multi app manifests

leohemsted:
Now, create them on a per-app basis. See commits for more details - we're using jinja2, which we use to generate manifests in template-preview, document-download and antivirus

To test, you should be able to run `CF_SPACE=staging CF_APP=notify-api make generate-manifest` and compare it to that run from master.

Some formatting differences will be present, but mostly it'll be semantically the same. Let me know if you're confused by anything.


Makefile
```diff
@@ -216,8 +216,11 @@ generate-manifest:
 	$(if $(shell which gpg2), $(eval export GPG=gpg2), $(eval export GPG=gpg))
 	$(if ${GPG_PASSPHRASE_TXT}, $(eval export DECRYPT_CMD=echo -n $$$${GPG_PASSPHRASE_TXT} | ${GPG} --quiet --batch --passphrase-fd 0 --pinentry-mode loopback -d), $(eval export DECRYPT_CMD=${GPG} --quiet --batch -d))
 
-	@./scripts/generate_manifest.py ${CF_MANIFEST_FILE} \
-	    <(${DECRYPT_CMD} ${NOTIFY_CREDENTIALS}/credentials/${CF_SPACE}/paas/environment-variables.gpg)
+	@jinja2 --strict manifest.yml.j2 \
```
allait: We should add `jinja2-cli[yaml]` to our dev requirements

manifest.yml.j2
```diff
@@ -0,0 +1,95 @@
...
+    health-check-type: none
+    no-route: true
+    {% endif %}
+
+    services:
+      - notify-db
+      - logit-ssl-syslog-drain
+
+    env:
+      NOTIFY_APP_NAME: {{ app['NOTIFY_APP_NAME'] }}
```
allait: `app.get('NOTIFY_APP_NAME', CF_APP.replace('notify-', ''))`?

manifest.yml.j2
```diff
@@ -0,0 +1,95 @@
...
+  'notify-delivery-worker': {'NOTIFY_APP_NAME': 'delivery-worker'},
+  'notify-delivery-worker-internal': {'NOTIFY_APP_NAME': 'delivery-worker-internal'},
+  'notify-delivery-worker-receipts': {'NOTIFY_APP_NAME': 'delivery-worker-receipts'},
+  'notify-delivery-worker-service-callbacks': {'NOTIFY_APP_NAME': 'delivery-worker-service-callbacks', 'disk_quota': '2G'},
+} -%}
+
+{%- set env_defaults = {
+  'preview': {'instances': 1},
+  'staging': {'instances': 2},
+  'production': {'instances': 2}
```
allait: I think this needs to be passed in as a variable set to the current instance number. Otherwise `v3-apply-manifest` will immediately scale down to the given number, which we don't want to happen in production

leohemsted: decision: this can be the autoscaler's responsibility [as it is outside of deployments]. Remove instances from the manifest entirely

PR #2465, opened 2019-04-16
Go live data recorded for service

servingUpAces:
Adding go_live_user and go_live_at columns to Services table. 

When the `request to go live` link is clicked, we will record the user that requested to go live. 
When the service is marked live we will record that date. 

Include the new columns in the return for get service.

### DIFF
--- a/migrations/versions/0287_add_go_live_user.py
+++ b/migrations/versions/0287_add_go_live_user.py
@@ -1,0 +1,29 @@
+"""
+
+Revision ID: 0287_add_go_live_user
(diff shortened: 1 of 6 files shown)


migrations/versions/0287_add_go_live_user.py
```diff
@@ -0,0 +1,29 @@
+"""
+
+Revision ID: 0287_add_go_live_user
```
klssmith: These numbers will need to be updated

servingUpAces: Fixed

PR #2467, opened 2019-04-24
Populate go live user and date

servingUpAces:
New command to populate the go live user and date from the beta partners spreadsheet.

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -811,15 +811,21 @@ def populate_service_volume_intentions(file_name):
 def populate_go_live(file_name):
     # 0 - count, 1- Link, 2- Service ID, 3- DEPT, 4- Service Name, 5- Main contact,
     # 6- Contact detail, 7-MOU, 8- LIVE date, 9- SMS, 10 - Email, 11 - Letters, 12 -CRM, 13 - Blue badge
-
+    import csv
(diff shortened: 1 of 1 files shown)


app/commands.py
```diff
@@ -804,3 +804,32 @@ def populate_service_volume_intentions(file_name):
...
+
+@notify_command(name='populate-go-live')
+@click.option('-f', '--file_name', required=True, help='CSV file containing live service data')
+def populate_go_live(file_name):
+    # 0 - count, 1- Link, 2- Service ID, 3- DEPT, 4- Service Name, 5- Main contact,
+    # 6- Contact detail, 7-MOU, 8- LIVE date, 9- SMS, 10 - Email, 11 - Letters, 12 -CRM, 13 - Blue badge
+
+    print("Populate go live user and date")
+    with open(file_name, 'r') as f:
+        for line in itertools.islice(f, 1, None):
```
quis: Would be better to use the Python’s CSV parsing code here, to deal with stuff like cells containing commas. See how it’s used here: https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/recipients.py#L146-L150

servingUpAces: Ok - done

PR #2468, opened 2019-04-24
Local statsd exporter

allait:
### Add a make target to download statsd_exporter binary

Downloads a linux binary from Github and puts it in the scripts folder This should add it to the package CI uploads to CF during deploy                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                .

Unfortunately, since out .cfignore is a symlink to .gitignore we can't add the file to be ignored by git, so it shows up in untracked if `cf-deploy` is run locally.

### Make statsd host configurable with an env variable

Setting `STATSD_HOST` for an env variable allows us to switch to a local statsd_exporter on a per-app basis.

This also changes `STATSD_ENABLED` to be on when `STATSD_HOST` is set, avoiding the need to set it separately.

### Add a `local_statsd` configuration to PaaS manifest template

Running `statsd_exporter` alongside the app process allows us to get StatsD metrics pushed by workers to Prometheus.

This requires adding a route to the worker instances and binding the RE prometheus discovery service. So this approach won't work for API and admin since they already have `gunicorn` bound to the `$PORT`.

Since we're not ready to switch all apps to Prometheus metrics at once and we don't currently have a way to push statsd metrics to multiple destination we're using a configuration setting in the manifest template to switch individual workers in specific environments.

`local_statsd` contains a list of environments where the app should use local `statsd_exporter` for pushing statsd metrics instead of HostedGraphite.

### Add statsd_exporter to app PaaS startup scripts

`statsd_exporter` is only started if `STATSD_HOST` is set to `localhost`.

### Add statsd exporter metric mapping configuration file

`statsd_mapping.yml` sets all timers to default to a Prometheus histogram metric type and renames all metrics to start with `notifications_`.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -281,9 +281,9 @@ class Config(object):
     NOTIFICATIONS_ALERT = 5  # five mins
     FROM_NUMBER = 'development'
 
-    STATSD_ENABLED = False
-    STATSD_HOST = "statsd.hostedgraphite.com"
+    STATSD_HOST = os.getenv('STATSD_HOST')
     STATSD_PORT = 8125
+    STATSD_ENABLED = bool(STATSD_HOST)
 
     SENDING_NOTIFICATIONS_TIMEOUT_PERIOD = 259200  # 3 days
 
@@ -356,11 +356,6 @@ class Development(Config):
     SQLALCHEMY_DATABASE_URI = 'postgresql://localhost/notification_api'
     REDIS_URL = 'redis://localhost:6379/0'
 
-    STATSD_ENABLED = False
-    STATSD_HOST = "localhost"
-    STATSD_PORT = 1000
-    STATSD_PREFIX = "stats-prefix"
-
     ANTIVIRUS_ENABLED = os.getenv('ANTIVIRUS_ENABLED') == '1'
 
     for queue in QueueNames.all_queues():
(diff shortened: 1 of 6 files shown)


app/config.py
```diff
@@ -81,9 +81,6 @@ class Config(object):
     # Firetext simluation key
     LOADTESTING_API_KEY = os.getenv("LOADTESTING_API_KEY")
 
-    # Hosted graphite statsd prefix
-    STATSD_PREFIX = os.getenv('STATSD_PREFIX')
```
leohemsted: i think we still need this for apps we haven't switched over to prometheus? We're still including it in the manifest and it looks like it's still being used in https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/clients/statsd/statsd_client.py#L20

allait: Oops, yeah, it absolutely should still be there, good catch!

PR #2482, opened 2019-04-25
Bump utils to bring in changes to RecipientCSV rows

klssmith:
Bumped utils to version 31.2.5, which changes when the rows of a RecipientCSV get created.

Switched to using `.get_rows()` from RecipientCSV (a generator) instead of the `.rows` property (which builds a list of the rows in memory).

Details about the utils change: https://github.com/alphagov/notifications-utils/pull/602

[Pivotal story](https://www.pivotaltracker.com/story/show/165021246)


klssmith:
That would be good - I'll see if I can find a way of monitoring it

PR #2483, opened 2019-04-25
Add get_live_services_data endpoint and related query

CrystalPea:
Adds get_live_services_data endpoint to services rest and a related query to services_dao.

The query gets data about live services from multiple tables (Service, User, Organisation, FactBilling) and then serialises it so it can be pulled by admin app to create a csv report for our platform admins.

Needed for this story: https://www.pivotaltracker.com/story/show/164829877

Needs to be deployed before this PR: https://github.com/alphagov/notifications-admin/pull/2934

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -127,6 +127,8 @@ def dao_fetch_live_services_data():
         Service.volume_email,
         Service.volume_letter,
         this_year_ft_billing.c.notification_type
+    ).order_by(
+        asc(Service.go_live_at)
     ).all()
     results = []
     for row in data:
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -7,6 +7,7 @@ from sqlalchemy.orm import joinedload
 from flask import current_app
 
 from app import db
+from app.dao.date_util import get_current_financial_year
 from app.dao.dao_utils import (
     transactional,
     version_class,
@@ -75,6 +76,11 @@ def dao_count_live_services():
 
 
 def dao_fetch_live_services_data():
+    year_start_date, year_end_date = get_current_financial_year()
+    this_year_ft_billing = FactBilling.query.filter(
+        FactBilling.bst_date >= year_start_date,
+        FactBilling.bst_date <= year_end_date,
+    ).subquery()
     data = db.session.query(
         Service.id,
         Organisation.name.label("organisation_name"),
@@ -89,18 +95,18 @@ def dao_fetch_live_services_data():
         Service.volume_email,
         Service.volume_letter,
         case([
-            (FactBilling.notification_type == 'email', func.sum(FactBilling.notifications_sent))
+            (this_year_ft_billing.c.notification_type == 'email', func.sum(this_year_ft_billing.c.notifications_sent))
         ], else_=0).label("email_totals"),
         case([
-            (FactBilling.notification_type == 'sms', func.sum(FactBilling.notifications_sent))
+            (this_year_ft_billing.c.notification_type == 'sms', func.sum(this_year_ft_billing.c.notifications_sent))
         ], else_=0).label("sms_totals"),
         case([
-            (FactBilling.notification_type == 'letter', func.sum(FactBilling.notifications_sent))
+            (this_year_ft_billing.c.notification_type == 'letter', func.sum(this_year_ft_billing.c.notifications_sent))
         ], else_=0).label("letter_totals"),
     ).outerjoin(
         Service.organisation
     ).outerjoin(
-        FactBilling, Service.id == FactBilling.service_id
+        this_year_ft_billing, Service.id == this_year_ft_billing.c.service_id
     ).outerjoin(
         User, Service.go_live_user_id == User.id
     ).group_by(
@@ -116,7 +122,7 @@ def dao_fetch_live_services_data():
         Service.volume_sms,
         Service.volume_email,
         Service.volume_letter,
-        FactBilling.notification_type
+        this_year_ft_billing.c.notification_type
     ).all()
     results = []
     for row in data:
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -87,6 +87,7 @@ def dao_fetch_live_services_data():
         Service.name.label("service_name"),
         Service.consent_to_research,
         Service.go_live_user_id,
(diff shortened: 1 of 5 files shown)


app/dao/services_dao.py
```diff
@@ -72,6 +75,89 @@ def dao_count_live_services():
...
+        case([
+            (this_year_ft_billing.c.notification_type == 'email', func.sum(this_year_ft_billing.c.notifications_sent))
+        ], else_=0).label("email_totals"),
+        case([
+            (this_year_ft_billing.c.notification_type == 'sms', func.sum(this_year_ft_billing.c.notifications_sent))
+        ], else_=0).label("sms_totals"),
+        case([
+            (this_year_ft_billing.c.notification_type == 'letter', func.sum(this_year_ft_billing.c.notifications_sent))
+        ], else_=0).label("letter_totals"),
+    ).outerjoin(
```
servingUpAces: Do we need to filter the services for live (`count_as_live` = True)?

tests/app/dao/test_services_dao.py
```diff
@@ -380,6 +384,34 @@ def test_get_all_user_services_should_return_empty_list_if_no_services_for_user(
...
+    org = create_organisation()
+    service = create_service(go_live_user=sample_user)
+    template = create_template(service=service)
+    create_service(service_name='second', go_live_user=sample_user)
+    template2 = create_template(service=service, template_type='email')
+    dao_add_service_to_organisation(service=service, organisation_id=org.id)
+    create_ft_billing(bst_date='2019-04-20', notification_type='sms', template=template, service=service)
+    create_ft_billing(bst_date='2019-04-21', notification_type='sms', template=template, service=service)
+    create_ft_billing(bst_date='2018-04-20', notification_type='sms', template=template, service=service)
+    create_ft_billing(bst_date='2019-04-20', notification_type='email', template=template2,
```
servingUpAces: Should expand this test to have more than one service in ft_billing and letters

app/dao/services_dao.py
```diff
@@ -72,6 +75,89 @@ def dao_count_live_services():
...
+        Service.organisation
```
servingUpAces: Do we need to filter by `bst_date`?

app/dao/services_dao.py
CrystalPea: yeah I think so, at least it is part of the story ticket.

app/dao/services_dao.py
CrystalPea: We need to get totals for current financial year, so I think yes :)

tests/app/dao/test_services_dao.py
CrystalPea: ok 👍

CrystalPea: I added more data to the test now, so we have:
- one service with all kinds of records in ft_billing
- one service with letter records in ft_billing
- one service without ft billing records or go_live_user to see that it still appears

app/dao/services_dao.py
CrystalPea: ok I added that as a filter and added a related check to the test ☺️

app/dao/services_dao.py
```diff
@@ -72,6 +75,93 @@ def dao_count_live_services():
     ).count()
 
 
+def dao_fetch_live_services_data():
+    year_start_date, year_end_date = get_current_financial_year()
+    this_year_ft_billing = FactBilling.query.filter(
```
servingUpAces: This can just be a filter on the query rather than a subquery

tests/app/dao/test_services_dao.py
```diff
@@ -380,6 +384,44 @@ def test_get_all_user_services_should_return_empty_list_if_no_services_for_user(
     assert len(dao_fetch_all_services_by_user(user.id)) == 0
 
 
+@freeze_time('2019-04-23T10:00:00')
```
servingUpAces: We could have a test to ensure the date filter is working.

tests/app/dao/test_services_dao.py
```diff
@@ -380,6 +384,44 @@ def test_get_all_user_services_should_return_empty_list_if_no_services_for_user(
...
+    service_2 = create_service(service_name='second', go_live_user=sample_user)
+    create_service(service_name='third')
+    create_service(service_name='not_live', count_as_live=False)
+    template2 = create_template(service=service, template_type='email')
+    template_letter_1 = create_template(service=service, template_type='letter')
+    template_letter_2 = create_template(service=service_2, template_type='letter')
+    dao_add_service_to_organisation(service=service, organisation_id=org.id)
+    create_ft_billing(bst_date='2019-04-20', notification_type='sms', template=template, service=service)
+    create_ft_billing(bst_date='2019-04-21', notification_type='sms', template=template, service=service)
+    create_ft_billing(bst_date='2018-04-20', notification_type='sms', template=template, service=service)
```
CrystalPea: @servingUpAces this ft_billing is dated at 2018 so it does not appear in the final result, is this sufficient as a test?

tests/app/dao/test_services_dao.py
CrystalPea: one of ft_biling instances in the test is dated at 2018 and does not appear in the results, so it kind of is tested, or do you mean like a separate test?

app/dao/services_dao.py
CrystalPea: but then services without ft_billing records in the dates would not show up in the results at all, right? Or at least this is what happened when I tried to apply it as a filter first...

tests/app/dao/test_services_dao.py
CrystalPea: I added comments to this test so it's clearer what happens there :)

app/dao/services_dao.py
```diff
@@ -72,6 +75,93 @@ def dao_count_live_services():
...
+        Service.go_live_user_id,
+        User.name,
+        User.email_address,
+        User.mobile_number,
+        Service.go_live_at,
+        Service.volume_sms,
+        Service.volume_email,
+        Service.volume_letter,
+        this_year_ft_billing.c.notification_type
+    ).all()
```
servingUpAces: Do we need to order the results? Maybe by go_live_at?

CrystalPea: That's a good idea 👍

PR #2484, opened 2019-04-26
Change get_providers endpoint to return no of SMS sent by each provider

klssmith:
In addition to the existing provider data, we also want return the number of billable units (multiplied by the rate multiplier) that each SMS provider sent this month. This will be used on the platform admin providers page.

Since we can no longer get all the information we need from the provider details schema, this makes a new DAO function to get all the data for the endpoint.

[Pivotal story](https://www.pivotaltracker.com/story/show/165382068)

### DIFF
--- a/app/provider_details/rest.py
+++ b/app/provider_details/rest.py
@@ -2,9 +2,9 @@ from flask import Blueprint, jsonify, request
 
 from app.schemas import provider_details_schema, provider_details_history_schema
 from app.dao.provider_details_dao import (
-    get_provider_details,
     get_provider_details_by_id,
     dao_update_provider_details,
+    dao_get_provider_stats,
     dao_get_provider_versions
 )
 from app.dao.users_dao import get_user_by_id
@@ -19,8 +19,23 @@ register_errors(provider_details)
 
 @provider_details.route('', methods=['GET'])
 def get_providers():
-    data = provider_details_schema.dump(get_provider_details(), many=True).data
-    return jsonify(provider_details=data)
+    data = dao_get_provider_stats()
+
+    provider_details = [
+        {'id': row.id,
+         'display_name': row.display_name,
+         'identifier': row.identifier,
+         'priority': row.priority,
+         'notification_type': row.notification_type,
+         'active': row.active,
+         'updated_at': row.updated_at,
+         'supports_international': row.supports_international,
+         'created_by_name': row.created_by_name,
+         'current_month_billable_sms': row.current_month_billable_sms}
+        for row in data
+    ]
+
+    return jsonify(provider_details=provider_details)
 
 
 @provider_details.route('/<uuid:provider_details_id>', methods=['GET'])
--- a/app/dao/provider_details_dao.py
+++ b/app/dao/provider_details_dao.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 
-from sqlalchemy import asc, desc
+from notifications_utils.timezones import convert_utc_to_bst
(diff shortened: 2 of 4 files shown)


app/provider_details/rest.py
```diff
@@ -23,6 +24,14 @@ def get_providers():
     return jsonify(provider_details=data)
 
 
+@provider_details.route('/sms-statistics', methods=['GET'])
```
servingUpAces: Does this need to be a separate API call? Or can we update /provider-details to include the volume percentage?

klssmith: Using the existing api call is better - I've updated this PR and my admin one so we're not adding a new endpoint

app/dao/fact_billing_dao.py
```diff
@@ -255,6 +255,22 @@ def get_rate(
...
 
 
+def get_sms_sent_by_provider_for_current_month():
+    # this query does not include the current day since the task to populate ft_billing runs overnight
+    current_bst_datetime = convert_utc_to_bst(datetime.utcnow())
+    first_day_of_the_month = current_bst_datetime.date().replace(day=1)
+
+    return db.session.query(
+        FactBilling.provider,
+        func.sum(FactBilling.billable_units).label('billable_units_sent')
```
servingUpAces: This should include rate_multiplier, `func.sum(FactBilling.billable_units * rate_multiplier).label('billable_units_sent')`

app/dao/fact_billing_dao.py
```diff
@@ -255,6 +255,22 @@ def get_rate(
...
+    ).filter(
+        FactBilling.notification_type == 'sms',
```
servingUpAces: Best to use the SMS_TYPE from model.py

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -474,3 +475,24 @@ def test_delete_billing_data(notify_db_session):
...
+
+
+@freeze_time('2018-06-28 12:00')
+def test_get_sms_sent_by_provider_for_current_month(notify_db_session):
+    service_1 = create_service(service_name='1')
+    service_2 = create_service(service_name='2')
+    sms_template_1 = create_template(service_1, 'sms')
+    sms_template_2 = create_template(service_2, 'sms')
+
+    create_ft_billing('2017-06-05', 'sms', sms_template_2, service_1, provider='firetext', billable_unit=4)
```
servingUpAces: Could add a row with a rate_multiplier > 1

app/dao/fact_billing_dao.py
klssmith: 👍 changed to `SMS_TYPE`

app/dao/fact_billing_dao.py
klssmith: I've updated this to include rate_multiplier (and the test too to have a row in ft_billing with a different rate_multiplier)

app/provider_details/rest.py
```diff
@@ -19,8 +20,17 @@
 
 @provider_details.route('', methods=['GET'])
 def get_providers():
-    data = provider_details_schema.dump(get_provider_details(), many=True).data
-    return jsonify(provider_details=data)
+    provider_details = provider_details_schema.dump(get_provider_details(), many=True).data
```
servingUpAces: It could be more efficient to do this in one query. So add provider_details to the the stats query.

klssmith: @servingUpAces I've changed this so that it now uses one new DAO query to get all the data

PR #2486, opened 2019-04-30
Change distribution of queues among notify delivery workers

CrystalPea:
This is so that retry-tasks queue, which can have quite a lot of
load, has its own worker, and other queues are paired with queues
that flow similarly:
- letter-tasks with create-letters-pdf-tasks
- job-tasks with database-tasks

Pivotal ticket: https://www.pivotaltracker.com/story/show/165021462
Connected AWS PR: https://github.com/alphagov/notifications-aws/pull/512
Connected autoscaler PR: https://github.com/alphagov/notifications-paas-autoscaler/pull/51


manifest.yml.j2
```diff
@@ -8,12 +8,13 @@
...
 
   'notify-delivery-celery-beat': {'memory': '128M'},
-  'notify-delivery-worker-database': {},
+  'notify-delivery-worker-database-and-job-tasks': {},
   'notify-delivery-worker-research': {'local_statsd': ['preview', 'staging']},
   'notify-delivery-worker-sender': {'disk_quota': '2G', 'memory': '3G', 'local_statsd': ['preview', 'staging']},
   'notify-delivery-worker-periodic': {},
   'notify-delivery-worker-priority': {},
-  'notify-delivery-worker': {},
+  'notify-delivery-worker-letters-tasks': {},
```
leohemsted: i think it would be good to not use the word "tasks" - my thinking is that if we deliberately disassociate the worker from the queue name, it indicates to people that it's not just for "letters-tasks" (but also for create-pdf-tasks)

i'm not as convinced on this as on the other comment though. 

then there's the retry worker which literally only looks at the tasks

i guess the word "tasks" is boilerplate here (it doesn't provide any extra context) that the other apps don't use.

PR #2487, opened 2019-05-01
Insert/update NotificationHistory just before we delete from Notifications

servingUpAces:
We want to stop inserting and updating NotificationHistory each time we insert/update Notification.

This PR adds a function to upsert (insert or update if exists) NotificationHistory all the rows from Notification that we are about to delete in the nightly task. This will happen just before the delete function. Since it is a upsert query the function can be called more than once.
This should allow us remove all the insert/updates to NotificationHistory.

However, there is a concern that this will double the length of time the tasks take. So do we do these upserts in a separate task or in the same one?

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -19,6 +19,7 @@ from sqlalchemy import (desc, func, asc)
 from sqlalchemy.orm import joinedload
 from sqlalchemy.sql import functions
 from sqlalchemy.sql.expression import case
+from sqlalchemy.dialects.postgresql import insert
 from werkzeug.datastructures import MultiDict
 
 from app import db, create_uuid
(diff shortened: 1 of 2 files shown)


app/dao/notifications_dao.py
```diff
@@ -361,6 +364,54 @@ def _delete_notifications(
...
+        Notification.phone_prefix,
+        Notification.rate_multiplier,
+        Notification.created_by_id,
+        Notification.postage
+    ).filter(
+        Notification.notification_type == notification_type,
+        Notification.service_id == service_id,
+        Notification.created_at < date_to_delete_from,
+        Notification.key_type != KEY_TYPE_TEST
+    ).all()
```
leohemsted: rather than calling .all (and invoking big data serialisation and deserialisation costs), does an `insert into notification_history from (select ... from notifications)` query work?

https://docs.sqlalchemy.org/en/13/core/dml.html?highlight=from_select#sqlalchemy.sql.expression.Insert.from_select

servingUpAces: It doesn't look like it can handle the `on_conflict_do_update`

PR #2488, opened 2019-05-02
Only show live services in live services query

CrystalPea:
Earlier filter was insufficient so now additional filters checking if service
is not in trial mode and if it's active have been added.

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -111,7 +111,9 @@ def dao_fetch_live_services_data():
     ).outerjoin(
         User, Service.go_live_user_id == User.id
     ).filter(
-        Service.count_as_live == True  # noqa
+        Service.count_as_live.is_(True),
+        Service.active.is_(True),
+        Service.restricted.is_(False),
     ).group_by(
         Service.id,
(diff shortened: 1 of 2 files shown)


app/dao/services_dao.py
```diff
@@ -111,7 +111,9 @@ def dao_fetch_live_services_data():
     ).outerjoin(
         User, Service.go_live_user_id == User.id
     ).filter(
-        Service.count_as_live == True  # noqa
+        Service.count_as_live == True,  # noqa
```
quis: You can avoid `# noqa` by using `.is_`:
```suggestion
        Service.count_as_live.is_(True),
```

CrystalPea: ahh, sweet, thank you 💖

CrystalPea: done 👍

quis: Still has the `# noqa` comment 😳

CrystalPea:
```
git commit --amend
git push -f origin only_show_live_services
```

shhhh.... it never happened 😯

PR #2490, opened 2019-05-03
Add organisation type to live services report

CrystalPea:
Story ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/165791164

### DIFF
--- a/tests/app/db.py
+++ b/tests/app/db.py
@@ -541,10 +541,11 @@ def create_domain(domain, organisation_id):
     return domain
 
 
-def create_organisation(name='test_org_1', active=True):
+def create_organisation(name='test_org_1', active=True, organisation_type=None):
     data = {
         'name': name,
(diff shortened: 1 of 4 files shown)


tests/app/db.py
```diff
@@ -541,10 +541,11 @@ def create_domain(domain, organisation_id):
     return domain
 
 
-def create_organisation(name='test_org_1', active=True):
+def create_organisation(name='test_org_1', active=True, type=None):
     data = {
         'name': name,
-        'active': active
+        'active': active,
+        'organisation_type': type
```
quis: It’s best not to use `type` as a variable name because it’s also the name of [a built in function](https://docs.python.org/3/library/functions.html#type).

CrystalPea: good point, I changed that now 👍

PR #2493, opened 2019-05-08
Ensure notifications have billable units and provider set if sending fails

klssmith:
If we try to send an SMS to the provider and the provider throws an exception
(because they return a 503 status code) the notification should retry. But if
we get a callback from the provider before the notification has been retried, the
notification will have no billable units or provider set.

To avoid this, we now set billable_units and provider even if there has been
an exception from our provider.

### DIFF
--- a/tests/app/delivery/test_send_to_providers.py
+++ b/tests/app/delivery/test_send_to_providers.py
@@ -562,6 +562,25 @@ def test_should_update_billable_units_according_to_research_mode_and_key_type(
     assert sample_notification.billable_units == billable_units
 
 
+def test_should_set_notification_billable_units_and_provider_if_sending_to_provider_fails(
+    notify_db,
+    sample_service,
+    sample_notification,
+    mocker,
+):
+    mocker.patch('app.mmg_client.send_sms', side_effect=Exception())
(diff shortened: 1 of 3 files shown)


tests/app/delivery/test_send_to_providers.py
```diff
@@ -562,6 +562,25 @@ def test_should_update_billable_units_according_to_research_mode_and_key_type(
...
+    mocker.patch('app.delivery.send_to_providers.dao_toggle_sms_provider')
+
+    sample_notification.billable_units = 0
+    assert sample_notification.sent_by is None
+
+    with pytest.raises(Exception):
+        send_to_providers.send_sms_to_provider(sample_notification)
+
+    assert sample_notification.billable_units == 1
+    assert sample_notification.sent_by == 'mmg'
```
servingUpAces: Can we pop in a quick assert that the mock to toggle the provider is called. I don't see another test for that.

klssmith: Just found we did have a (failing) test for that in another file, so I've combined the two tests and made sure it passes in https://github.com/alphagov/notifications-api/pull/2493/commits/ce24b13654acc6a1609a5f3866a69bd6f5982b63

PR #2496, opened 2019-05-08
Refactor update notifications

servingUpAces:
Refactor send_to_providers. 
- Remove some redundant code for research mode.
- The international parameter in update_notification_to_sending is not needed.
- Update unit tests and removed duplicates

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -90,9 +90,7 @@ def send_email_to_provider(notification):
         return
     if notification.status == 'created':
         provider = provider_to_use(EMAIL_TYPE, notification.id)
-        current_app.logger.debug(
-            "Starting sending EMAIL {} to provider at {}".format(notification.id, datetime.utcnow())
-        )
+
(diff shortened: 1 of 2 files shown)


app/delivery/send_to_providers.py
```diff
@@ -42,9 +40,7 @@ def send_sms_to_provider(notification):
 
     if notification.status == 'created':
         provider = provider_to_use(SMS_TYPE, notification.id, notification.international)
-        current_app.logger.debug(
-            "Starting sending SMS {} to provider at {}".format(notification.id, datetime.utcnow())
-        )
```
CrystalPea: did we decide we don't need those logs anymore? Or are they sent from elsewhere?

servingUpAces: Debug statements don't show up in prod logs. Other log messages in the journey give us this information.

CrystalPea: Ahh, ok, makes sense 👍

PR #2501, opened 2019-05-13
Don't unbind the non-existent notify-api-db-migration service

issyl0:
- Things were failing to deploy because it was trying to unbind a
  service that didn't exist (they were removed in the upgrade to
  `cflinuxfs3` and come up on deploy).

Questions:
- The temporary app for the migration should go away at the end of the deploy, how do we achieve this?
- Can we integrate this with the task below it, "check status", so that if the migration succeeds it deletes the app?
- There doesn't seem to be a `flask db downgrade` step for rolling back the migration?


issyl0:
Oops, this does less stuff than #2498.

PR #2505, opened 2019-05-15
add routes for all apps

leohemsted:
If not specified, healthcheck is set to `port` by cloudfoundry - we've seen some issues with upgrading the deployment from v2 to v3 when using port - it adds apps to load balancer when they're not ready, which can result in 404s. by setting healthcheck to http it'll wait for the `/status` endpoint to return 200, which will wait for flask to get everything up and running properly


leohemsted:
see https://github.com/alphagov/notifications-paas-autoscaler/commit/80c428836891368efdd770561f0f40279dc2d36a for why we're adding routes to all apps, including delivery workers (which shouldn't need a route really)

PR #2525, opened 2019-06-03
Don't insert notification history

servingUpAces:
This PR is the final step to make NotificationHistory truly an archive table. 

After this NotificationHistory is only inserted when the Notification is deleted. A join to NotificationHistory on id is added to the delete query to insure we have the history row before deleting the Notification.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -344,7 +344,8 @@ def _delete_notifications(
     ).filter(
         Notification.notification_type == notification_type,
         Notification.service_id == service_id,
-        Notification.created_at < date_to_delete_from
+        Notification.created_at < date_to_delete_from,
+        NotificationHistory.id == Notification.id
     ).limit(query_limit).subquery()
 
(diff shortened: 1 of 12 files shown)


app/dao/notifications_dao.py
```diff
@@ -438,17 +434,14 @@ def _timeout_notifications(current_statuses, new_status, timeout_start, updated_
...
-        )
-    # return a list of q = notification_ids in Notification table for sending delivery receipts
+    Notification.query.filter(
+        Notification.created_at < timeout_start,
+        Notification.status.in_(current_statuses),
+        Notification.notification_type != LETTER_TYPE
+    ).update(
+        {'status': new_status, 'updated_at': updated_at},
+        synchronize_session=False
+    )
```
leohemsted: not one for right now cos this change is big enough, but in the future we can use `update ... returning ...` to only execute the query once instead of twice

https://docs.sqlalchemy.org/en/13/dialects/postgresql.html#insert-update-returning

PR #2537, opened 2019-06-11
Add rows in template_redacted where they are missing and help remember to insert them in the future

CrystalPea:
- Create template_redacted rows for templates created by migration, so that we can edit those templates

- Add commented out code to past migrations so when somebody is copying those migrations they remember about template_redacted insert

- Add missing statements to downgrades for past migrations

### DIFF
--- a/migrations/versions/0171_add_org_invite_template.py
+++ b/migrations/versions/0171_add_org_invite_template.py
@@ -20,12 +20,12 @@ template_id = '203566f0-d835-47c5-aa06-932439c86573'
 
 def upgrade():
     template_insert = """
-        INSERT INTO templates (id, name, template_type, created_at, content, archived, service_id, subject, 
+        INSERT INTO templates (id, name, template_type, created_at, content, archived, service_id, subject,
         created_by_id, version, process_type, hidden)
         VALUES ('{}', '{}', '{}', '{}', '{}', False, '{}', '{}', '{}', 1, '{}', false)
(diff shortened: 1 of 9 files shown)


migrations/versions/0171_add_org_invite_template.py
```diff
@@ -72,11 +72,24 @@ def upgrade():
...
     # clean up constraints on org_to_service - service_id-org_id constraint is redundant
     op.drop_constraint('organisation_to_service_service_id_organisation_id_key', 'organisation_to_service', type_='unique')
 
 
 def downgrade():
+    op.execute("DELETE FROM notifications WHERE template_id = '{}'".format(template_id))
+    op.execute("DELETE FROM notification_history WHERE template_id = '{}'".format(template_id))
     op.execute("DELETE FROM templates_history WHERE id = '{}'".format(template_id))
     op.execute("DELETE FROM templates WHERE id = '{}'".format(template_id))
+    op.execute("DELETE FROM template_redacted WHERE template_id = '{}'".format(template_id))
```
servingUpAces: FYI: If the delete from template_redacted is before templates it will fail because of the foreign key constraint

CrystalPea: ahh yes, I tested it now and I have to put statement deleting from template_redacted before statement deleting from templates 👍. I will do that now ☺️

PR #2538, opened 2019-06-11
Update to unit tests

servingUpAces:
- Adding test where post_notification can't send to a member outside of the team or whitelist.

- Updating some tests to be pytest4 compliant.


tests/app/test_model.py
```diff
@@ -181,9 +181,9 @@ def test_notification_subject_fills_in_placeholders_for_email(sample_email_templ
     assert noti.subject == 'hello'
 
 
-def test_notification_subject_fills_in_placeholders_for_letter(sample_letter_template):
-    noti = create_notification(sample_letter_template, personalisation={'name': 'hello'})
-    noti.template.subject = '((name))'
+def test_notification_subject_fills_in_placeholders_for_letter(sample_service):
+    template = create_template(service=sample_service, template_type='email', subject='((name))')
```
klssmith: `template_type` here should be 'letter'

klssmith:
Looks good - there are a few places where we don't need to pass in `notify_db` and `notify_db_session` (all the tests using `service_factory` or `sample_service` as a fixture)

tests/app/test_model.py
servingUpAces: Ah good point - or I could parametrise that and test both email and letter

PR #2543, opened 2019-06-13
Cancel letter job

CrystalPea:
In this PR:
- create a rest endpoint cancel_letter_job
- check if letter job can be cancelled
- if yes, cancel letter job and return how many notifications have been cancelled
- if not, return a 'forbidden' error and a message explaining why the letter job can't be cancelled

Others in this series :)
utils: https://github.com/alphagov/notifications-utils/pull/621
admin: https://github.com/alphagov/notifications-admin/pull/3026

### DIFF
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -12,16 +12,18 @@ from sqlalchemy import (
 
 from app import db
 from app.dao.dao_utils import transactional
+from app.dao.templates_dao import dao_get_template_by_id
 from app.utils import midnight_n_days_ago
+
 from app.models import (
     Job,
+    JOB_STATUS_FINISHED,
     JOB_STATUS_PENDING,
     JOB_STATUS_SCHEDULED,
     LETTER_TYPE,
     Notification,
     Template,
     ServiceDataRetention,
-    NOTIFICATION_SENDING,
     NOTIFICATION_CREATED,
     NOTIFICATION_CANCELLED,
     JOB_STATUS_CANCELLED
@@ -169,8 +171,11 @@ def dao_cancel_letter_job(job):
 
 
 def can_cancel_letter_job(job):
-    # assert is a letter job
-    # assert job status == finished???
+    template = dao_get_template_by_id(job.template_id)
+    if template.template_type != LETTER_TYPE:
+        return False
+    if job.job_status != JOB_STATUS_FINISHED:
+        return False
     # Notifications are not in pending-virus-check
     count_notifications = Notification.query.filter(
         Notification.job_id == job.id,
@@ -179,6 +184,3 @@ def can_cancel_letter_job(job):
     if count_notifications != job.notification_count:
         return False
     return letter_can_be_cancelled(NOTIFICATION_CREATED, job.created_at)
-
-
-
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -2,7 +2,7 @@ import uuid
 from datetime import datetime, timedelta
 
 from flask import current_app
-from notifications_utils.letter_timings import letter_can_be_cancelled
+from notifications_utils.letter_timings import letter_can_be_cancelled, CANCELLABLE_JOB_LETTER_STATUSES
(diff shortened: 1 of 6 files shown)


app/dao/jobs_dao.py
```diff
@@ -148,16 +155,35 @@ def dao_get_jobs_older_than_data_retention(notification_types):
...
+        Notification.job_id == job.id
     ).all()
+    count_notifications = len(notifications)
+    if job.job_status != JOB_STATUS_FINISHED or count_notifications != job.notification_count:
+        return "This job is still being processed. Wait a couple of minutes and try again."
+    count_cancellable_notifications = len([
+        n for n in notifications if n.status in CANCELLABLE_JOB_LETTER_STATUSES
+    ])
+    if count_cancellable_notifications != job.notification_count or not letter_can_be_cancelled(
+        NOTIFICATION_CREATED, job.created_at
```
servingUpAces: We should also consider if the job has been scheduled

CrystalPea: good point, for scheduled jobs we should go through `cancel_job` endpoint, right? But then we shouldn't show the "cancel letter job" link in admin to avoid confusion, I will make sure this doesn't happen 👍

app/dao/jobs_dao.py
```diff
@@ -148,16 +155,35 @@ def dao_get_jobs_older_than_data_retention(notification_types):
...
+    notifications = Notification.query.filter(
+        Notification.job_id == job.id
     ).all()
+    count_notifications = len(notifications)
+    if job.job_status != JOB_STATUS_FINISHED or count_notifications != job.notification_count:
+        return "This job is still being processed. Wait a couple of minutes and try again."
+    count_cancellable_notifications = len([
+        n for n in notifications if n.status in CANCELLABLE_JOB_LETTER_STATUSES
+    ])
+    if count_cancellable_notifications != job.notification_count or not letter_can_be_cancelled(
```
servingUpAces: I think this if statement needs to be `if count_cancellable_notifications == job.notification_count or not letter_can_be_cancelled`
If possible the job is `finished` but not all the notifications have been created yet and it is before the cut off.

tests/app/job/test_rest.py
```diff
@@ -71,6 +71,105 @@ def test_cant_cancel_normal_job(client, sample_job, mocker):
     assert mock_update.call_count == 0
 
 
+@freeze_time('2019-06-13 13:00')
+def test_dao_cancel_letter_job_updates_notifications_and_job_to_cancelled(sample_letter_template, admin_request):
```
servingUpAces: remove `dao` from method name - we are testing the endpoint.

app/dao/jobs_dao.py
servingUpAces: Actually, the if stmt could just be `if not letter_can_be_cancelled`

app/dao/jobs_dao.py
```diff
@@ -148,16 +155,35 @@ def dao_get_jobs_older_than_data_retention(notification_types):
...
-        Job.template
-    ).filter(
-        Template.template_type == LETTER_TYPE,
-        # test letter jobs (or from research mode services) are created with a different filename,
-        # exclude them so we don't see them on the send to CSV
-        Job.original_file_name != LETTER_TEST_API_FILENAME
-    ).order_by(
-        desc(Job.created_at)
+@transactional
+def dao_cancel_letter_job(job):
```
servingUpAces: There aren't any tests for this method

app/dao/jobs_dao.py
```diff
@@ -148,16 +155,35 @@ def dao_get_jobs_older_than_data_retention(notification_types):
...
+        Notification.job_id == job.id
+    ).update({'status': NOTIFICATION_CANCELLED,
+              'updated_at': datetime.utcnow(),
+              'billable_units': 0})
+    job.job_status = JOB_STATUS_CANCELLED
+    dao_update_job(job)
+    return number_of_notifications_cancelled
+
+
+def can_letter_job_be_cancelled(job):
```
servingUpAces: There aren't any tests for this method

PR #2548, opened 2019-06-26
tell people when they've signed an MOU

leohemsted:
When someone signs an MOU, we should:

* tell the user that clicked the button
* tell notify
* if the user clicked on behalf of someone else, tell that person as well


migrations/versions/0298_add_mou_signed_receipt.py
```diff
@@ -0,0 +1,146 @@
...
+                    current_app.config['NOTIFY_SERVICE_ID'],
+                    template.get('subject'),
+                    current_app.config['NOTIFY_USER_ID'],
+                    'normal'
+                )
+            )
+
+        op.execute(
+            """
+            INSERT INTO template_redacted
```
servingUpAces: 👍

migrations/versions/0298_add_mou_signed_receipt.py
```diff
@@ -0,0 +1,146 @@
...
+        'content_lines': [
+            'Hi ((signed_by_name)),',
+            '',
+            '((org_name)) has accepted the GOV.​UK Notify data sharing and financial agreement. ',
+            '',
+            'If you need another copy of the agreement you can download it here: ((mou_link))',
+            '',
+            'If you need to add Cabinet Office as a supplier, here are the details you need:',
+            '',
+            'TO BE ADDED MANUALLY',
```
servingUpAces: Does this need to be updated?

leohemsted: yes, but through the admin interface so that we don't do anything. ah. hmm.

i'll split out the migration into a separate PR so we can do that before the automatic sending happens

servingUpAces: I see the templates are in prod - is someone fixing the content?

leohemsted: pete fixed the content just now

PR #2551, opened 2019-06-28
Fix 0 billable units on provider callback

klssmith:
We occasionally get an SMS with 0 `billable_units` if the `delivery-sender-worker` is stopped in the middle of processing a notification - we have to fix these manually. This change checks the billable units when we get the response from our SMS provider and sets the correct billable units if it's 0.

### DIFF
--- a/tests/app/notifications/test_process_client_response.py
+++ b/tests/app/notifications/test_process_client_response.py
@@ -96,7 +96,7 @@ def test_process_sms_response_does_not_send_status_update_for_pending(sample_not
     send_mock.assert_not_called()
 
 
-def test_process_sms_updates_sent_by_with_client_name_if_not_in_noti(notify_db, sample_notification):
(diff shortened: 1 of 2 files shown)


tests/app/notifications/test_process_client_response.py
```diff
@@ -105,7 +105,16 @@ def test_process_sms_updates_sent_by_with_client_name_if_not_in_noti(notify_db,
     assert sample_notification.sent_by == 'mmg'
 
 
-def test_process_sms_does_not_update_sent_by_if_already_set(mocker, notify_db, sample_notification):
+def test_process_sms_updates_billable_units_if_zero(sample_notification):
+    sample_notification.billable_units = 0
```
servingUpAces: Should we use the create_notification() method here for pytest4 compatibility?

klssmith: This should be pytest 4 compatible - I think it's only incompatible if you call a fixture like a function by doing something like `sample_notification()`

PR #2552, opened 2019-07-03
Add extra fields to create/update organisation

quis:
The admin app is passing these in as of https://github.com/alphagov/notifications-admin/pull/3038


quis:
@servingUpAces yes, we should do this when we expand the list of `organisation_types` to be bigger than the current 3 options (happening soon).

PR #2553, opened 2019-07-04
 Set org_type of a service when adding/updating org org type

klssmith:
The organisation_type of a service should match the organisation_type of the service's organisation (if there is one). This changes `dao_update_organisation` and `dao_add_service_to_organisation` to set the `organisation_type` of any services when adding / updating an org.

Also removes some unused fixtures and deletes some unused code.

Part of [this story](https://www.pivotaltracker.com/story/show/166293397)

### DIFF
--- a/tests/app/dao/test_organisation_dao.py
+++ b/tests/app/dao/test_organisation_dao.py
@@ -4,6 +4,7 @@ import uuid
 import pytest
 from sqlalchemy.exc import IntegrityError, SQLAlchemyError
 
+from app import db
 from app.dao.organisation_dao import (
     dao_get_organisations,
     dao_get_organisation_by_email_address,
@@ -16,7 +17,7 @@ from app.dao.organisation_dao import (
     dao_get_users_for_organisation,
     dao_add_user_to_organisation
 )
-from app.models import Organisation
+from app.models import Organisation, Service
 
 from tests.app.db import (
     create_domain,
@@ -29,7 +30,6 @@ from tests.app.db import (
 
 
 def test_get_organisations_gets_all_organisations_alphabetically_with_active_organisations_first(
-        notify_db,
         notify_db_session
 ):
     m_active_org = create_organisation(name='m_active_organisation')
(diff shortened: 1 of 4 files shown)


tests/app/dao/test_organisation_dao.py
```diff
@@ -119,17 +119,53 @@ def test_update_organisation_domains_lowercases(
     assert {domain.domain for domain in organisation.domains} == expected_domains
 
 
-def test_add_service_to_organisation(notify_db, notify_db_session, sample_service, sample_organisation):
+def test_update_organisation_does_not_update_the_service_org_type_if_org_type_is_not_provided(
+    sample_service,
+    sample_organisation,
+):
+    sample_service.organisation_type = 'local'
```
CrystalPea: for pytest 4 compatibility, it's better to use `create_service()` and `create_organisation()` if you want to update them in the test, because updating fixtures in the test is deprecated in pytest 4 as far as I am aware.

klssmith: Changing the properties of a fixture is fine in Pytest 4 as long as the fixture is a parameter of the test function. It's calling fixtures like a function that you're no longer allowed to do. (I've tried out my changes in Pytest 4 to double check and there are no warnings)

CrystalPea: ahhh, ok, thanks for educating me 💖

tests/app/dao/test_organisation_dao.py
```diff
@@ -119,17 +119,57 @@ def test_update_organisation_domains_lowercases(
...
+
+    sample_organisation.services.append(sample_service)
+    db.session.commit()
+
+    assert sample_organisation.name == 'sample organisation'
+
+    dao_update_organisation(sample_organisation.id, name='updated org name')
+
+    assert sample_organisation.name == 'updated org name'
+    assert sample_service.organisation_type == 'local'
```
servingUpAces: It's probably a good idea to add a check that `ServiceHistory` is updated as well.

klssmith: The service doesn't get updated in this test, but I've added versioning to the `dao_add_service_to_organisation` function and a test for `ServiceHistory` in `test_add_service_to_organisation`

PR #2557, opened 2019-07-10
Introduce new organisation types and organisation_types db table

CrystalPea:
#### In this PR:
- introduce new organisation types
- introduce organisation types table, with default for crown setting and allowance
- set service's crown to org's crown when service is joining the organisation

#### TODO when both this API PR and below Admin PR are deployed:
- migrate 'nhs' organisation types in Organisation and Services tables to be either 'nhs_local' or 'nhs_central' depending on free_sms_fragment_limit for services in that organisation
- make a constraint so all org types for orgs and services must be from those listed in organisation_types table
- delete generic `'nhs'` org type from allowed organisation types list in Models

Admin PR: https://github.com/alphagov/notifications-admin/pull/3057


migrations/versions/0299_org_types_table_.py
```diff
@@ -0,0 +1,43 @@
...
+        sa.Column('is_crown', sa.Boolean, nullable=True),
+        sa.Column('annual_free_sms_fragment_limit', sa.BigInteger, nullable=False)
+
+    )
+
+    op.bulk_insert(
+        organisation_types_table,
+        [
+            {'name': x, 'is_crown': y, 'annual_free_sms_fragment_limit': z} for x, y, z in [
+                ["Central government", None, 250000],
```
quis: Can we store something machine-readable in the database and let the admin app deal with the human readable label? I really don’t want to have to deal with database migrations to change these values. Would suggest:
- `central`
- `local`
- `nhs_central`
- `nhs_local`
- `emergency_services`
- `school_or_college`
- `other`

migrations/versions/0299_org_types_table.py
```diff
@@ -0,0 +1,47 @@
...
+            ]
+        ]
+    )
+    op.alter_column('services', 'crown', nullable=True)
+    op.alter_column('services_history', 'crown', nullable=True)
+
+
+def downgrade():
+    op.execute('DROP TABLE organisation_types')
+    op.alter_column('services', 'crown', nullable=False)
```
servingUpAces: This downgrade will fail if crown has been set to null.

CrystalPea: cool, I made away with those lines :)

app/dao/services_dao.py
```diff
@@ -282,7 +284,12 @@ def dao_create_service(
     service.id = service_id or uuid.uuid4()  # must be set now so version history model can use same id
     service.active = True
     service.research_mode = False
-    service.crown = service.organisation_type == 'central'
+    if organisation:
+        service.crown = organisation.crown
+    elif service.organisation_type in CROWN_ORGANISATION_TYPES:
+        service.crown = True
+    elif service.organisation_type in (NON_CROWN_ORGANISATION_TYPES + ['nhs']):
```
servingUpAces: why not just add nhs to the list?

migrations/versions/0299_org_types_table.py
```diff
@@ -0,0 +1,45 @@
...
+        sa.Column('is_crown', sa.Boolean, nullable=True),
+        sa.Column('annual_free_sms_fragment_limit', sa.BigInteger, nullable=False)
+
+    )
+
+    op.bulk_insert(
+        organisation_types_table,
+        [
+            {'name': x, 'is_crown': y, 'annual_free_sms_fragment_limit': z} for x, y, z in [
+                ["central", None, 250000],
```
servingUpAces: do we need to add nhs?

app/dao/services_dao.py
CrystalPea: I don't think we do, since we don't have a constraint on the table at this point, and in next PR I will migrate all 'nhs' organisation_type organisations to be either 'nhs_local' or 'nhs_central' before introducing the constraint.

migrations/versions/0299_org_types_table.py
CrystalPea: I don't think we do, since we don't have a constraint on the table at this point, and in next PR I will migrate all 'nhs' organisation_type organisations to be either 'nhs_local' or 'nhs_central' before introducing the constraint.

app/dao/services_dao.py
CrystalPea: We could temporarily add it, that's true (in this PR, which is a draft right now, we will be getting rid of generic 'nhs' type: https://github.com/alphagov/notifications-api/pull/2565 )

migrations/versions/0299_org_types_table.py
CrystalPea: Actually you are right that this is cleaner code, I added it.

app/dao/services_dao.py
CrystalPea: Actually you are right that this is cleaner code, I added it.

PR #2564, opened 2019-07-19
Add endpoint with data for report for the engagement team

klssmith:
This adds a dao function and an endpoint to get the data needed for a new report for the engagement team. This report will be added to `/platform-admin/reports` in notifications-admin.

[Pivotal story](https://www.pivotaltracker.com/story/show/166495005)

### DIFF
--- a/app/dao/fact_notification_status_dao.py
+++ b/app/dao/fact_notification_status_dao.py
@@ -2,7 +2,7 @@ from datetime import datetime, timedelta, time
 
 from flask import current_app
 from notifications_utils.timezones import convert_bst_to_utc
-from sqlalchemy import func
+from sqlalchemy import case, func, Date
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.sql.expression import literal, extract
 from sqlalchemy.types import DateTime, Integer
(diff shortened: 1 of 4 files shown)


app/dao/fact_notification_status_dao.py
```diff
@@ -440,3 +448,67 @@ def get_total_sent_notifications_for_day_and_type(day, notification_type):
...
+        func.sum(case(
+            [
+                (FactNotificationStatus.notification_status == NOTIFICATION_SENT,
+                 FactNotificationStatus.notification_count)
+            ],
+            else_=0)).label('count_sent'),
+    ).join(
+        Service, FactNotificationStatus.service_id == Service.id
+    ).filter(
+        FactNotificationStatus.notification_status != NOTIFICATION_CREATED,
```
CrystalPea: why we don't count created notifications? Is it because they can be still cancelled?

klssmith: I'm not sure what the report gets used for - just trying to copy the query the engagement team already gets

servingUpAces: The engagement team are not interested in the created notifications. To be fair this report is usually asked for after the end of the month, so there should really be anything still in created.

PR #2565, opened 2019-07-19
Migrate org types

CrystalPea:
### In this PR:
- Migrate all orgs and services onto new organisation types
- Remove all mentions of generic `'nhs'` organisation type.

Part of this story: 
https://www.pivotaltracker.com/story/show/166666407

### DIFF
--- a/migrations/versions/0300_migrate_org_types.py
+++ b/migrations/versions/0300_migrate_org_types.py
@@ -1,116 +1,75 @@
+import os
+
 """
 
 Revision ID: 0300_migrate_org_types
 Revises: 0299_org_types_table
-Create Date: 2019-07-19 11:13:41.286472
+Create Date: 2019-07-24 16:18:27.467361
 
 """
 from alembic import op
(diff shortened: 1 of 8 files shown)


CrystalPea:
@servingUpAces I will check

migrations/versions/0300_migrate_org_types.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+revision = '0300_migrate_org_types'
+down_revision = '0299_org_types_table'
+
+environment = os.environ['NOTIFY_ENVIRONMENT']
+
+
+def upgrade():
+    if environment != "production":
```
servingUpAces: You need to include 'live' as well

CrystalPea: Good call, I've done that now :)

CrystalPea:
@servingUpAces 'nhs' was not added to organisation_types table, it was only in a list in models that the valdiation was done against. This PR removes it completely form API code.

PR #2568, opened 2019-07-29
Add email and letter branding to the add organisation command 

servingUpAces:
Updated the `populate-organisations-from-file` command to include a default email and letter branding column. The name of brand is used to look up the brand id and set it in the Organisation model. Also works if the columns are empty.


app/commands.py
```diff
@@ -713,12 +717,23 @@ def boolean_or_none(field):
         for line in itertools.islice(f, 1, None):
             columns = line.split('|')
             print(columns)
+            email_branding = None
+            email_branding_column = columns[5]
```
klssmith: Could do `columns[5].strip()` like you've done with letter branding below just in case the column contains whitespace

PR #2571, opened 2019-07-31
REP-340: Bump utils and refreeze

andy-paine:
- Includes new statsd client (see change in requirements)


leohemsted:
superceded by #2572

PR #2577, opened 2019-08-07
 Retry deliver_sms task immediately if sending fails

klssmith:
If the `deliver_sms` catches an exception when trying to send an SMS, we want the first retry to happen immediately (because we will have switched providers), then every retry after that to happen at the standard intervals.

[Pivotal story](https://www.pivotaltracker.com/story/show/166575127)

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -26,7 +26,10 @@ def deliver_sms(self, notification_id):
             current_app.logger.exception(
                 "SMS notification delivery for id: {} failed".format(notification_id)
             )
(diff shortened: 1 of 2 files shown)


pytest.ini
```diff
@@ -6,4 +6,4 @@ env =
     LOADTESTING_API_KEY=loadtesting
     FIRETEXT_API_KEY=Firetext
     NOTIFICATION_QUEUE_PREFIX=testing
-addopts = -n 4 -v -p no:warnings
```
servingUpAces: Did you mean to commit this?

servingUpAces: Never mind - I get it now :)

PR #2586, opened 2019-08-13
Add organisation_id to Service.

servingUpAces:
This is the second commit in the series to add organisation_id to Service.
- Data migration to update services.organisation_id from data in organisation_to_service
 (The rollback will lose any updates to organisation unless the script is updated to set organistion_to_service from service.organisation_id )
- Update Service.organisation relationship to a ForeignKey relationship to Organisation.
- Update Organisation.services to a backref relationship to Service.

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -293,7 +293,7 @@ def dao_create_service(
     insert_service_sms_sender(service, current_app.config['FROM_NUMBER'])
 
     if organisation:
-        service.organisation = organisation
+        # service.organisation = organisation
         service.organisation_id = organisation.id
         service.organisation_type = organisation.organisation_type
         if organisation.email_branding:
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -293,7 +293,6 @@ def dao_create_service(
     insert_service_sms_sender(service, current_app.config['FROM_NUMBER'])
 
     if organisation:
-        # service.organisation = organisation
         service.organisation_id = organisation.id
         service.organisation_type = organisation.organisation_type
         if organisation.email_branding:
--- a/app/models.py
+++ b/app/models.py
@@ -349,12 +349,6 @@ class Organisation(db.Model):
     active = db.Column(db.Boolean, nullable=False, default=True)
     created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
     updated_at = db.Column(db.DateTime, nullable=True, onupdate=datetime.datetime.utcnow)
-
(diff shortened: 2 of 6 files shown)


app/dao/services_dao.py
```diff
@@ -293,7 +293,7 @@ def dao_create_service(
     insert_service_sms_sender(service, current_app.config['FROM_NUMBER'])
 
     if organisation:
-        service.organisation = organisation
+        # service.organisation = organisation
```
leohemsted: guess this should be removed rather than commented out?

migrations/versions/0303_populate_services_org_id.py
```diff
@@ -0,0 +1,59 @@
...
+            UPDATE services
+               SET organisation_id = :organisation_id
+            WHERE id = :service_id
+        """
+        conn.execute(text(sql), service_id=str(x.service_id), organisation_id=str(x.organisation_id))
+        history_sql = """
+            UPDATE services_history
+               SET organisation_id = :organisation_id
+            WHERE id = :service_id
+              AND version = (select max(version) from services_history sh2 where id = services_history.id);
```
CrystalPea: what `sh2` means here?

app/models.py
```diff
@@ -352,8 +352,8 @@ class Organisation(db.Model):
 
     services = db.relationship(
         'Service',
-        secondary='organisation_to_service',
-        uselist=True)
+        uselist=True,
+        backref=db.backref('services'))
```
leohemsted: to be consistent with other models, i don't think we should have this relationship here, instead just define it on service model only.

app/models.py
```diff
@@ -480,11 +480,7 @@ class Service(db.Model, Versioned):
     go_live_at = db.Column(db.DateTime, nullable=True)
 
     organisation_id = db.Column(UUID(as_uuid=True), db.ForeignKey('organisation.id'), index=True, nullable=True)
-    organisation = db.relationship(
-        'Organisation',
-        secondary=organisation_to_service,
-        uselist=False,
-        single_parent=True)
+    organisation = db.relationship('Organisation', foreign_keys=[organisation_id])
```
leohemsted: If we stick `backref='services'` here we don't need to define the relationship on the Organisation model above

also the foreign_keys field is only needed when there's multiple columns with the same foreign key (eg go_live_user_id vs created_by_id)

migrations/versions/0303_populate_services_org_id.py
servingUpAces: It's an alias for the table - so when I say services.history it knows I'm talking about the outside table

PR #2591, opened 2019-08-15
Remove the organisation_to_service table.

servingUpAces:
This table is no longer used or referenced in the code.
We can remove this mapping table now that the organisation to service relationship is handled by the foreign key in Services.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -70,7 +70,6 @@ def create_app(application):
     application.config.from_object(configs[notify_environment])
 
     application.config['NOTIFY_APP_NAME'] = application.name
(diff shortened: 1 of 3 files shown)


app/__init__.py
```diff
@@ -70,7 +70,7 @@ def create_app(application):
     application.config.from_object(configs[notify_environment])
 
     application.config['NOTIFY_APP_NAME'] = application.name
-
+    print(application.config['SQLALCHEMY_DATABASE_URI'])
```
klssmith: There's an accidental `print` here

servingUpAces:
I didn't write the populate query on the downgrade because the table has not been in use for almost a week, so it is very unlikely we would need it.

PR #2593, opened 2019-08-15
split up reporting tasks in to separate tasks per day

leohemsted:
to try and speed up overall time by parallelising the tasks.

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -33,13 +33,35 @@ def create_nightly_billing(day_start=None):
     for i in range(0, 4):
         process_day = day_start - timedelta(days=i)
 
-        transit_data = fetch_billing_data_for_day(process_day=process_day)
+        create_nightly_billing_for_day.apply_async(
+            kwargs={'process_day': process_day.isoformat()},
+            queue=QueueNames.REPORTING
+        )
 
-        for data in transit_data:
-            update_fact_billing(data, process_day)
 
-        current_app.logger.info(
-            "create-nightly-billing task complete. {} rows updated for day: {}".format(len(transit_data), process_day))
+@notify_celery.task(name="create-nightly-billing-for-day")
+@statsd(namespace="tasks")
+def create_nightly_billing_for_day(process_day):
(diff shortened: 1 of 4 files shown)


app/config.py
```diff
@@ -219,9 +219,24 @@ class Config(object):
...
+            'options': {'queue': QueueNames.PERIODIC}
+        },
+        'delete-email-notifications': {
+            'task': 'delete-email-notifications',
+            'schedule': crontab(hour=4, minute=45),  # after 'create-nightly-notification-status'
+            'options': {'queue': QueueNames.PERIODIC}
+        },
+        'delete-letter-notifications': {
+            'task': 'delete-letter-notifications',
+            'schedule': crontab(hour=1, minute=20),  # after 'create-nightly-notification-status'
```
servingUpAces: Do you want this around 4?

app/config.py
```diff
@@ -219,9 +219,24 @@ class Config(object):
         },
         'create-nightly-notification-status': {
             'task': 'create-nightly-notification-status',
-            'schedule': crontab(hour=0, minute=30),  # after 'timeout-sending-notifications'
+            'schedule': crontab(hour=4, minute=15),  # after 'timeout-sending-notifications'
```
servingUpAces: Should this be earlier?

app/celery/reporting_tasks.py
```diff
@@ -33,13 +28,35 @@ def create_nightly_billing(day_start=None):
...
+@notify_celery.task(name="create-nightly-billing-for-day")
+@statsd(namespace="tasks")
+def create_nightly_billing_for_day(process_day):
+    process_day = datetime.strptime(process_day, "%Y-%m-%d").date()
+
+    start = datetime.utcnow()
+    transit_data = fetch_billing_data_for_day(process_day=process_day)
+    end = datetime.utcnow()
+
+    current_app.logger.info('create-nightly-billing-for-day {} fetched in {} seconds'.format(
```
servingUpAces: Weird we have this log message - we could just add statsd annotation to the query methods.

PR #2597, opened 2019-08-21
Use retention for reporting tasks

leohemsted:
previously we checked notifications table, and if the results were zero, checked the notification history table to see if there's data in there. When we know that data isn't in notifications, we're still checking. These queries take half a second per service, and we're doing at least ten for each of the five thousand services we have in notify. Most of these services have no data in either table for any given day, and we can reduce the amount of queries we do by only checking one table.

Check the data retention for a service, and then if the date is older than the retention, get from history table.

NOTE: This requires that the delete tasks haven't run yet for the day! If your retention is three days, this will look in the Notification table for data from three days ago - expecting that shortly after the task finishes, we'll delete that data.


app/utils.py
```diff
@@ -112,3 +113,26 @@ def email_address_is_nhs(email_address):
...
         '@nhs.uk', '@nhs.net', '.nhs.uk', '.nhs.net',
     ))
+
+
+def get_notification_table_to_use(service, notification_type, process_day, has_delete_task_run=False):
+    """
+    Work out what table will contain notification data for a service by looking up their data retention.
+
+    Make sure that when you run this you think about whether the delete task has run for that day! If it's run, the
+    notifications from that day will have moved to NotificationHistory. By default we assume it hasn't run, since
```
servingUpAces: since what?

leohemsted: i had a think and the default value felt dangerous (as in, easy to ignore and forget about). so i've just made it non-optional and changed this comment.

app/dao/fact_billing_dao.py
```diff
@@ -303,29 +305,23 @@ def fetch_billing_data_for_day(process_day, service_id=None):
...
-        service_ids = [x.id for x in Service.query.all()]
+        services = Service.query.all()
     else:
-        service_ids = [service_id]
-    for id_of_service in service_ids:
+        services = [Service.query.get(service_id)]
+
+    for service in services:
         for notification_type in (SMS_TYPE, EMAIL_TYPE, LETTER_TYPE):
+            table = get_notification_table_to_use(service, notification_type, process_day, has_delete_task_run=False)
```
servingUpAces: The `has_delete_task_run` is always `False`. Do we need it?

app/dao/fact_notification_status_dao.py
```diff
@@ -111,7 +108,7 @@ def update_fact_notification_status(data, process_day):
             notification_count=row.notification_count,
         )
         db.session.connection().execute(stmt)
-        db.session.commit()
+    db.session.commit()
```
servingUpAces: Should this method be annotated with `@transactional`?

app/models.py
```diff
@@ -2044,7 +2045,13 @@ class ServiceDataRetention(db.Model):
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), unique=False, index=True, nullable=False)
-    service = db.relationship(Service, backref=db.backref('service_data_retention'))
+    service = db.relationship(
+        Service,
+        backref=db.backref(
+            'data_retention',
+            collection_class=attribute_mapped_collection('notification_type')
```
servingUpAces: This is nice

tests/app/test_utils.py
```diff
@@ -51,3 +55,43 @@ def test_get_midnight_for_day_before_returns_expected_date(date, expected_date):
...
+@freeze_time('2019-06-09 23:30')
+def test_get_notification_table_to_use_respects_daylight_savings_time(sample_service):
+    # current time is 12:30am on 10th july in BST
+    assert get_notification_table_to_use(sample_service, 'sms', date(2019, 6, 1), False) == NotificationHistory
+    assert get_notification_table_to_use(sample_service, 'sms', date(2019, 6, 2), False) == Notification
+
+
+@freeze_time('2019-01-10 00:30')
+def test_get_notification_table_to_use_checks_service_data_retention(sample_service):
+    create_service_data_retention(sample_service, 'email', days_of_retention=3)
```
servingUpAces: Do we need a test for a service with data retention longer than the default 7 days?

app/dao/fact_billing_dao.py
leohemsted: we don't _need_it but i liked this because that way whenever you call this function you have to manually define the value for `has_delete_task_run`, making sure you've thought about it for your function.

tests/app/test_utils.py
leohemsted: added to that test

app/dao/fact_billing_dao.py
servingUpAces: fair enough

PR #2599, opened 2019-08-23
Billing for all services

servingUpAces:
New billing report for all services. 

There is a command to populate the missing annual billing - the report will miss wrong for those missing rows. You'll need to run the `flask command populate-annual-billing -y 2019`

After the task is done the count of active services should match the count of services in annual billing for 2019.
`select count(*) from services where active=true`
equals
`select count(*) from annual_billing where financial_year_start=2019`

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -278,8 +278,6 @@ def populate_annual_billing(year):
                                                      free_sms_fragment_limit=free_allowance[0],
                                                      financial_year_start=int(year))
 
-    db.session.commit()
-
 
 @notify_command(name='list-routes')
 def list_routes():
--- a/app/commands.py
+++ b/app/commands.py
@@ -21,6 +21,7 @@ from app.celery.nightly_tasks import send_total_sent_notifications_to_performanc
 from app.celery.service_callback_tasks import send_delivery_status_to_service
 from app.celery.letters_pdf_tasks import create_letters_pdf
 from app.config import QueueNames
+from app.dao.annual_billing_dao import dao_create_or_update_annual_billing_for_year
 from app.dao.fact_billing_dao import (
     delete_billing_data_for_service_for_day,
     fetch_billing_data_for_day,
@@ -248,29 +249,38 @@ def backfill_processing_time(start_date, end_date):
         send_processing_time_for_start_and_end(process_start_date, process_end_date)
 
 
-@notify_command()
-def populate_annual_billing():
+@notify_command(name='latest_annual_billing')
+@click.option('-y', '--year', required=True,
+              help="""The year to populate the annual billing data for, i.e. 2019""")
+def populate_annual_billing(year):
     """
-    add annual_billing for 2016, 2017 and 2018.
+    add annual_billing for given year.
     """
-    financial_year = [2016, 2017, 2018]
+    sql = """
+        Select id from services where active = true
+        except
+        select service_id
+        from annual_billing
+        where financial_year_start = :year
+    """
+    services_without_annual_billing = db.session.execute(sql, {"year": int(year)})
+    for row in services_without_annual_billing:
+        latest_annual_billing = """
+            Select free_sms_fragment_limit
+            from annual_billing
+            where service_id = :service_id
+            order by financial_year_start limit 1
+        """
+        print(row.id)
+        free_allowance_rows = db.session.execute(latest_annual_billing, {"service_id": row.id})
+        free_allowance = [x[0]for x in free_allowance_rows]
+        print(free_allowance)
+        print("create free limit of {} for service: {}".format(free_allowance[0], row.id))
+        dao_create_or_update_annual_billing_for_year(service_id=row.id,
+                                                     free_sms_fragment_limit=free_allowance[0],
+                                                     financial_year_start=int(year))
 
-    for fy in financial_year:
-        populate_data = """
-        INSERT INTO annual_billing(id, service_id, free_sms_fragment_limit, financial_year_start,
-                created_at, updated_at)
-            SELECT uuid_in(md5(random()::text || now()::text)::cstring), id, 250000, {}, '{}', '{}'
-            FROM services
-            WHERE id NOT IN(
-                SELECT service_id
-                FROM annual_billing
-                WHERE financial_year_start={})
-        """.format(fy, datetime.utcnow(), datetime.utcnow(), fy)
-
-        services_result1 = db.session.execute(populate_data)
-        db.session.commit()
-
-        print("Populated annual billing {} for {} services".format(fy, services_result1.rowcount))
+    db.session.commit()
 
 
 @notify_command(name='list-routes')
--- a/app/commands.py
+++ b/app/commands.py
@@ -249,8 +249,8 @@ def backfill_processing_time(start_date, end_date):
         send_processing_time_for_start_and_end(process_start_date, process_end_date)
 
 
-@notify_command(name='latest_annual_billing')
-@click.option('-y', '--year', required=True,
+@notify_command(name='populate-annual-billing')
+@click.option('-y', '--year', required=True, type=int,
               help="""The year to populate the annual billing data for, i.e. 2019""")
 def populate_annual_billing(year):
     """
@@ -263,18 +263,16 @@ def populate_annual_billing(year):
         from annual_billing
         where financial_year_start = :year
     """
-    services_without_annual_billing = db.session.execute(sql, {"year": int(year)})
+    services_without_annual_billing = db.session.execute(sql, {"year": year})
     for row in services_without_annual_billing:
         latest_annual_billing = """
             Select free_sms_fragment_limit
             from annual_billing
             where service_id = :service_id
-            order by financial_year_start limit 1
+            order by financial_year_start desc limit 1
         """
-        print(row.id)
         free_allowance_rows = db.session.execute(latest_annual_billing, {"service_id": row.id})
         free_allowance = [x[0]for x in free_allowance_rows]
-        print(free_allowance)
         print("create free limit of {} for service: {}".format(free_allowance[0], row.id))
         dao_create_or_update_annual_billing_for_year(service_id=row.id,
                                                      free_sms_fragment_limit=free_allowance[0],
--- a/app/dao/date_util.py
+++ b/app/dao/date_util.py
@@ -63,8 +63,9 @@ def get_current_financial_year_start_year():
     return financial_year_start
 
 
-def which_financial_year(start_date):
-    if start_date <= get_april_fools(int(start_date.strftime('%Y'))):
-        return int(start_date.strftime('%Y')) - 1
+def get_financial_year_for_datetime(start_date):
+    year = int(start_date.strftime('%Y'))
+    if start_date < get_april_fools(year):
+        return year - 1
     else:
-        return int(start_date.strftime('%Y'))
+        return year
--- a/app/dao/date_util.py
+++ b/app/dao/date_util.py
@@ -61,3 +61,10 @@ def get_current_financial_year_start_year():
     if now < start_date:
         financial_year_start = financial_year_start - 1
     return financial_year_start
+
+
+def which_financial_year(start_date):
+    if start_date <= get_april_fools(int(start_date.strftime('%Y'))):
+        return int(start_date.strftime('%Y')) - 1
+    else:
+        return int(start_date.strftime('%Y'))
--- a/app/dao/date_util.py
+++ b/app/dao/date_util.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, date, time
 
 from notifications_utils.timezones import convert_bst_to_utc
 import pytz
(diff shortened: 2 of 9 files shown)


app/commands.py
```diff
@@ -248,29 +249,38 @@ def backfill_processing_time(start_date, end_date):
         send_processing_time_for_start_and_end(process_start_date, process_end_date)
 
 
-@notify_command()
-def populate_annual_billing():
+@notify_command(name='latest_annual_billing')
```
leohemsted: assume this should be `name='populate_annual_billing'`

app/commands.py
```diff
@@ -248,29 +249,38 @@ def backfill_processing_time(start_date, end_date):
...
-
-        services_result1 = db.session.execute(populate_data)
-        db.session.commit()
+    sql = """
+        Select id from services where active = true
+        except
+        select service_id
+        from annual_billing
+        where financial_year_start = :year
+    """
```
leohemsted: i don't really mind if this doesnt get changed, but does this do anything different to a more normal join?

```select service_id
from annual_billing
join services on service_id = services.id
where service.active and financial_year_start = :year```

File: app/dao/date_util.py [hunk 3]
```diff
@@ -61,3 +61,10 @@ def get_current_financial_year_start_year():
     if now < start_date:
         financial_year_start = financial_year_start - 1
     return financial_year_start
+
+
+def which_financial_year(start_date):
```
leohemsted: i don't think this file should be in the dao subfolder 🤔

app/platform_stats/rest.py
```diff
@@ -27,3 +32,75 @@ def get_platform_stats():
...
+        raise InvalidRequest(message="Input must be a date in the format: YYYY-MM-DD", status_code=400)
+    if end_date < start_date:
+        raise InvalidRequest(message="Start date must be before end date", status_code=400)
+    if 4 <= int(start_date.strftime("%m")) <= 12:
+        year_start, year_end = get_financial_year(year=int(start_date.strftime("%Y")))
+    else:
+        year_start, year_end = get_financial_year(year=int(start_date.strftime("%Y")) - 1)
+    year_start = year_start.date()
+    year_end = year_end.date()
+    if year_start <= start_date <= year_end and year_start <= end_date <= year_end:
```
leohemsted: i think here you can use your new function to make this much simpler

```python
if which_financial_year(start_date) != which_financial_year(end_date):
    raise InvalidRequest [...]
```

app/dao/date_util.py
```diff
@@ -61,3 +61,10 @@ def get_current_financial_year_start_year():
...
+    if start_date <= get_april_fools(int(start_date.strftime('%Y'))):
```
leohemsted: this should be `<`, not `<=`.

this assert currently fails, and shouldn't

`assert which_financial_year(datetime(2019, 3, 31, 23, 0)) == 2019`

app/dao/date_util.py
leohemsted: also i feel like this function should probably be called `get_financial_year_for_datetime` just to be explicit (especially about how it expects a datetime, not a date)

app/dao/fact_billing_dao.py
```diff
@@ -19,11 +24,155 @@
...
+    start_of_year = convert_utc_to_bst(financial_year_start(billing_year))
+    query = db.session.query(
+        FactBilling.service_id.label("service_id"),
+        AnnualBilling.free_sms_fragment_limit,
+        func.sum(FactBilling.billable_units * FactBilling.rate_multiplier).label('billable_units'),
+        func.greatest((AnnualBilling.free_sms_fragment_limit -
+                       func.sum(FactBilling.billable_units * FactBilling.rate_multiplier)
+                       ).cast(Integer), 0).label('sms_remainder')
+    ).filter(
+        FactBilling.service_id == Service.id,
```
leohemsted: why is this filter here? the `Service` model isn't used in this query

app/dao/fact_billing_dao.py
```diff
@@ -19,11 +24,155 @@
...
+        FactBilling.service_id == AnnualBilling.service_id,
+        FactBilling.bst_date >= start_of_year,
+        FactBilling.bst_date < start_date,
+        FactBilling.notification_type == SMS_TYPE,
+        AnnualBilling.financial_year_start == billing_year,
+    ).group_by(
+        FactBilling.service_id,
+        AnnualBilling.free_sms_fragment_limit,
+    )
+    return query
```
leohemsted: this should probably `return query.subquery()` to make it clearer what this is. i think it's easy to lose the `.subquery()` in the beginning of fetch_sms_billing_for_all_services bit.

I guess this has been done to make testing the function easier, I might have a quick read in to sqlalchemy to see if it's possible to get around this some way

app/dao/fact_billing_dao.py
```diff
@@ -19,11 +24,155 @@
...
+    print(billing_year)
+    start_of_year = convert_utc_to_bst(financial_year_start(billing_year))
+    query = db.session.query(
+        FactBilling.service_id.label("service_id"),
+        AnnualBilling.free_sms_fragment_limit,
+        func.sum(FactBilling.billable_units * FactBilling.rate_multiplier).label('billable_units'),
+        func.greatest((AnnualBilling.free_sms_fragment_limit -
+                       func.sum(FactBilling.billable_units * FactBilling.rate_multiplier)
+                       ).cast(Integer), 0).label('sms_remainder')
+    ).filter(
```
leohemsted: this query needs a `.join(AnnualBilling)`

inner join i assume?

app/dao/fact_billing_dao.py
```diff
@@ -19,11 +24,155 @@
...
+
+    # ASSUMPTION: AnnualBilling has been populated for year.
+    billing_year = which_financial_year(start_date)
+    free_allowance_remainder = fetch_sms_free_allowance_remainder(start_date).subquery()
+    sms_billable_units = func.sum(FactBilling.billable_units * FactBilling.rate_multiplier)
+    sms_remainder = func.coalesce(free_allowance_remainder.c.sms_remainder, AnnualBilling.free_sms_fragment_limit)
+    chargeable_sms = case([(sms_remainder == 0, sms_billable_units),
+                           (sms_billable_units - sms_remainder <= 0, 0),
+                           (sms_billable_units - sms_remainder > 0,
+                            sms_billable_units - sms_remainder)], else_=0)
```
leohemsted: i think this can just be `func.max(sms_billable_units - sms_remainder, 0)`

app/dao/fact_billing_dao.py
```diff
@@ -19,11 +24,155 @@
...
+    sms_cost = case([(sms_remainder == 0, sms_billable_units * FactBilling.rate),
+                     (sms_billable_units - sms_remainder <= 0, 0),
+                     (sms_billable_units - sms_remainder > 0, (sms_billable_units - sms_remainder) * FactBilling.rate)
+                     ], else_=0)
```
leohemsted: i think this can just be `func.max((sms_billable_units - sms_remainder) * FactBilling.rate, 0)`

app/dao/fact_billing_dao.py
```diff
@@ -19,11 +24,155 @@
...
+        sms_remainder.label("sms_remainder"),
+        sms_billable_units.label('sms_billable_units'),
+        chargeable_sms.label("chargeable_billable_sms"),
+        sms_cost.label('sms_cost'),
+    ).join(
+        Service.annual_billing,
+    ).outerjoin(
+        free_allowance_remainder, Service.id == free_allowance_remainder.c.service_id
+    ).outerjoin(
+        Organisation, Service.organisation_id == Organisation.id
```
leohemsted: this query isn't joining right - the `from` bit of the raw sql comes from the first column selected - in this case, organisation name. If you don't want to do that, you can use `.select_from` to select another table.

The tables involved:

Service
Organisation
free_allowance_remainder
FactBilling
AnnualBilling

I think `Service` is probably the best candidate for the select_from, but it depends on what we want to outer join and what we want to inner join.

Any table that isn't explicitly joined (but is referenced eg as a column to select, or to filter on) is cross joined (!)

app/dao/fact_billing_dao.py
```diff
@@ -19,11 +24,155 @@
...
+
+def fetch_letter_costs_for_all_services(start_date, end_date):
+    query = db.session.query(
+        Organisation.name.label("organisation_name"),
+        Organisation.id.label("organisation_id"),
+        Service.name.label("service_name"),
+        FactBilling.service_id.label("service_id"),
+        func.sum(FactBilling.notifications_sent * FactBilling.rate).label("letter_cost")
+    ).outerjoin(
+        Organisation, Service.organisation_id == Organisation.id
```
leohemsted: similar join comment. This table uses Organisation, Service and FactBilling so those should be joined.

app/dao/fact_billing_dao.py
```diff
@@ -19,11 +24,155 @@
...
+        Organisation.name.label("organisation_name"),
+        Organisation.id.label("organisation_id"),
+        Service.name.label("service_name"),
+        FactBilling.service_id.label("service_id"),
+        FactBilling.billable_units.label("sheet_count"),
+        FactBilling.rate.label("letter_rate"),
+        FactBilling.postage.label("postage"),
+        func.sum(FactBilling.notifications_sent).label("letters_sent"),
+    ).outerjoin(
+        Organisation, Service.organisation_id == Organisation.id
```
leohemsted: similar join comment. This table uses Organisation, Service and FactBilling so those should be joined.

app/platform_stats/rest.py
```diff
@@ -27,3 +32,75 @@ def get_platform_stats():
...
+            "organisation_id": str(s.organisation_id) if s.organisation_id else "",
+            "organisation_name": s.organisation_name or "",
+            "service_id": str(s.service_id),
+            "service_name": s.service_name,
+            "sms_cost": float(s.sms_cost),
+            "sms_fragments": s.chargeable_billable_sms,
+            "letter_cost": 0,
+            "letter_breakdown": ""
+        }
+        combined[str(s.service_id)] = entry
```
leohemsted: now that we're no longer jsonifying the dict, we don't need to cast the uuids to strings to use as keys any more

app/commands.py
```diff
@@ -248,29 +249,38 @@ def backfill_processing_time(start_date, end_date):
...
+    services_without_annual_billing = db.session.execute(sql, {"year": int(year)})
+    for row in services_without_annual_billing:
+        latest_annual_billing = """
+            Select free_sms_fragment_limit
+            from annual_billing
+            where service_id = :service_id
+            order by financial_year_start limit 1
```
klssmith: Should this be order by financial_year_start `desc` in order to get the latest annual billing data?

app/dao/date_util.py
leohemsted: update: not moving the file, dont want to make the diff of this PR too big. i did rename this function though, although it actually works fine with dates.

app/commands.py
leohemsted: wait disregard that i see this is selecting the inverse of the join - those services that dont have annual billing. thats a bit grosser query `select s.id from services s left outer join annual_billing ab on ab.service_id = s.id and ab.financial_year_start = :year where ab.id is null` which i dont think is as nice as this.

app/dao/fact_billing_dao.py
leohemsted: (removed annualbilling completely because we can get the allownace from free_allowance_remainder)

app/dao/fact_billing_dao.py
```diff
@@ -19,11 +24,164 @@
...
+        Service.id.label("service_id"),
+        func.sum(FactBilling.notifications_sent * FactBilling.rate).label("letter_cost")
+    ).select_from(
+        Service
+    ).outerjoin(
+        Service.organisation
+    ).join(
+        FactBilling, FactBilling.service_id == Service.id,
+    ).filter(
+        FactBilling.service_id == Service.id,
```
klssmith: Is this line needed since it's in the join clause?

app/commands.py
```diff
@@ -248,29 +249,36 @@ def backfill_processing_time(start_date, end_date):
...
+        """
+        free_allowance_rows = db.session.execute(latest_annual_billing, {"service_id": row.id})
+        free_allowance = [x[0]for x in free_allowance_rows]
+        print("create free limit of {} for service: {}".format(free_allowance[0], row.id))
+        dao_create_or_update_annual_billing_for_year(service_id=row.id,
+                                                     free_sms_fragment_limit=free_allowance[0],
+                                                     financial_year_start=int(year))
 
-        print("Populated annual billing {} for {} services".format(fy, services_result1.rowcount))
+    db.session.commit()
```
klssmith: The dao function is wrapped in `@transactional` so the commit isn't needed

klssmith:
Are we going to need to remember to run the `populate-annual-billing` command at the start of each new financial year in order for this query to return the right data?

leohemsted:
> Are we going to need to remember to run the `populate-annual-billing` command at the start of each new financial year in order for this query to return the right data?

yeah. i wonder if it's worth just running it at the beginning of this query - the select is quick, and the update will only need to run once a year. new services have that year's annualbilling inserted iirc

klssmith:
Yep, that would work. It will mean people need to clean up their local data before accessing the report if they have services without an annual billing entry (I had a few) but production will be fine and it would just be a one-off thing.

PR #2600, opened 2019-08-27
Add ‘GP’ as an org type

quis:
Although their allowances are the same as what we call `nhs_local` it makes more sense to store them separately because:
- we already present them as two separate choices to the user
- we may want to handle them differently in the future, eg in terms of what branding choices are available to them

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -42,6 +42,7 @@ from app.models import (
     EMAIL_TYPE,
     INTERNATIONAL_SMS_TYPE,
     KEY_TYPE_TEST,
+    NHS_ORGANISATION_TYPES,
     NON_CROWN_ORGANISATION_TYPES,
     SMS_TYPE,
     LETTER_TYPE,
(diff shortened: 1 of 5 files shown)


app/models.py
```diff
@@ -326,11 +326,11 @@ class Domain(db.Model):
 
 
 ORGANISATION_TYPES = [
-    "central", "local", "nhs_central", "nhs_local", "emergency_service", "school_or_college", "other",
+    "central", "local", "nhs_central", "nhs_local", "gp", "emergency_service", "school_or_college", "other",
```
CrystalPea: hmm... I feel it might be more useful to call the new org type `nhs_gp`, then if we look for all things 'nhs' it will come up. What do you think?

quis: @CrystalPea that’s a good suggestions, I‘ve changed it in 758fbdb

app/dao/services_dao.py
```diff
@@ -306,7 +307,7 @@ def dao_create_service(
         if organisation.letter_branding and not service.letter_branding:
             service.letter_branding = organisation.letter_branding
 
-    elif service.organisation_type in ['nhs_central', 'nhs_local'] or email_address_is_nhs(user.email_address):
+    elif service.organisation_type in NHS_ORGANISATION_TYPES or email_address_is_nhs(user.email_address):
```
CrystalPea: do we test automatically giving NHS branding somewhere? 🤔

quis: Yeah, I added an extra test case here: https://github.com/alphagov/notifications-api/pull/2600/files#diff-2be09a40f09ecb7eb899c53704ae81efR147

PR #2604, opened 2019-09-04
Handle both new and old response type from template preview's sanitise method

CrystalPea:
We are changing response format in template preview ( here: https://github.com/alphagov/notifications-template-preview/pull/367 ). We want API to handle both new and old response type for a smooth deploy of those changes.

Story ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/168057359

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -1,5 +1,6 @@
 import io
 import math
+import base64
 from datetime import datetime
 from uuid import UUID
 from hashlib import sha512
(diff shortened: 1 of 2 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -213,8 +215,14 @@ def process_virus_scan_passed(self, filename):
         _move_invalid_letter_and_update_status(notification, filename, scan_pdf_object)
         return
 
-    new_pdf = _sanitise_precompiled_pdf(self, notification, old_pdf)
-
+    sanitise_response = _sanitise_precompiled_pdf(self, notification, old_pdf)
+    if not sanitise_response:
```
servingUpAces: sanitise_response is never empty so this wouldn't set `new_pdf`. Do we have a test for this?

CrystalPea: It's empty when validation has failed, then `_sanitise_precompiled_pdf()` returns None.

PR #2608, opened 2019-09-10
Copy original file to redaction_failure folder when redaction fails

CrystalPea:
This is done so that we can see why no matches are found sometimes during the redaction procedure

Needs this PR to work: https://github.com/alphagov/notifications-template-preview/pull/369

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -31,6 +31,7 @@ from app.dao.notifications_dao import (
 )
 from app.errors import VirusScanError
 from app.letters.utils import (
+    copy_redaction_failed_pdf,
     get_reference_from_filename,
     get_folder_name,
     upload_letter_pdf,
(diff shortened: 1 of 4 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -219,10 +220,18 @@ def process_virus_scan_passed(self, filename):
...
         new_pdf = None
     else:
+        sanitise_response = sanitise_response.json()
         try:
-            new_pdf = base64.b64decode(sanitise_response.json()["file"].encode())
+            new_pdf = base64.b64decode(sanitise_response["file"].encode())
         except JSONDecodeError:
             new_pdf = sanitise_response.content
+
+        if "redaction_failed_message" in sanitise_response and sanitise_response["redaction_failed_message"]:
```
leohemsted: you can make this marginally cleaner by a) using `.get` and b) using a tmp variable. eg:

```python
msg = sanitise_response.get('redaction_failed_message')
if msg:
    logger(msg)
```

leohemsted: also can we get a unit test for this code flow? 🙏

PR #2610, opened 2019-09-11
add api endpoint to get pdf for letter

leohemsted:
a new endpoint (`/v2/notifications/<id>/pdf`) that returns a PDF for a letter notification

If the notification's pdf is available, it returns that, as a raw file. If the pdf is not available, it returns 400.

The pdf won't be available if the notification's status is:

* pending-virus-scan
* virus-scan-failed

The pdf will be retrieved from the correct s3 bucket based on its type. To make sure the pdf is available, we now create PDFs for templated test pdfs too (previously we just marked them as delivered).

### DIFF
--- a/app/v2/notifications/get_notifications.py
+++ b/app/v2/notifications/get_notifications.py
@@ -6,7 +6,7 @@ from app import api_user, authenticated_service
 from app.dao import notifications_dao
 from app.letters.utils import get_letter_pdf
 from app.schema_validation import validate
-from app.v2.errors import BadRequestError
+from app.v2.errors import BadRequestError, PDFNotReadyError
 from app.v2.notifications import v2_notification_blueprint
 from app.v2.notifications.notification_schemas import get_notifications_request, notification_by_id
 from app.models import (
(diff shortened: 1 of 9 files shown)


app/v2/notifications/get_notifications.py
```diff
@@ -13,10 +24,45 @@ def get_notification_by_id(notification_id):
...
+        notification_id, authenticated_service.id, _raise=True
+    )
+
+    if notification.notification_type != LETTER_TYPE:
+        raise BadRequestError(message="Notification is not a letter", status_code=400)
+
+    if notification.status in {
+        NOTIFICATION_PENDING_VIRUS_CHECK,
+        NOTIFICATION_VIRUS_SCAN_FAILED,
+        NOTIFICATION_TECHNICAL_FAILURE,
```
servingUpAces: Is there a situation where the letter is in tech-failure but can still retrieve the letter?

leohemsted: i thought best to play it safe, i don't know where the letter is (it could be in any bucket), or why it tech failured (could be something wrong with the letter for example). It might not even have passed virus scan.

PR #2611, opened 2019-09-16
Use created at for sending folder

servingUpAces:
### Refactor the code that figures out what folder and filename to use for the letter pdf files.

- Now we consistently use the created_at date, so we can always get the right file location and name.

- The previous updates to this code were trying to solve the problem if a pdf being created at 17:29, but not ready to upload until 17:31 after the antivirus and validation check.
But in those cases we would have trouble finding the file.

- Change the variable name to make a little more sense.

### DIFF
--- a/app/letters/utils.py
+++ b/app/letters/utils.py
@@ -25,8 +25,8 @@ LETTERS_PDF_FILE_LOCATION_STRUCTURE = \
 PRECOMPILED_BUCKET_PREFIX = '{folder}NOTIFY.{reference}'
 
 
-def get_folder_name(_now, is_test_or_scan_letter=False):
-    if is_test_or_scan_letter:
+def get_folder_name(_now, dont_use_sending_date=False):
+    if dont_use_sending_date:
         folder_name = ''
     else:
         print_datetime = convert_utc_to_bst(_now)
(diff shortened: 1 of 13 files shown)


app/letters/utils.py
```diff
@@ -25,8 +25,8 @@ class ScanErrorType(Enum):
 PRECOMPILED_BUCKET_PREFIX = '{folder}NOTIFY.{reference}'
 
 
-def get_folder_name(_now, is_test_or_scan_letter=False):
-    if is_test_or_scan_letter:
+def get_folder_name(_now, dont_use_sending_date=False):
```
leohemsted: i'd much rather this variable was a positive rather than a negative (eg `use_sending_date=True`), that's how most booleans are used.

app/letters/utils.py
```diff
@@ -36,19 +36,16 @@ def get_folder_name(_now, is_test_or_scan_letter=False):
     return folder_name
 
 
-def get_letter_pdf_filename(reference, crown, is_scan_letter=False, postage=SECOND_CLASS):
-    now = datetime.utcnow()
-
+def get_letter_pdf_filename(reference, crown, sending_date, dont_use_sending_date=False, postage=SECOND_CLASS):
```
leohemsted: same comment about `dont_use_sending_date` vs `use_sending_date`

app/letters/utils.py
```diff
@@ -61,12 +58,7 @@ def get_bucket_name_and_prefix_for_notification(notification):
         bucket_name = current_app.config['TEST_LETTERS_BUCKET_NAME']
     else:
         bucket_name = current_app.config['LETTERS_PDF_BUCKET_NAME']
-        if notification.sent_at:
-            folder = "{}/".format(notification.sent_at.date())
-        elif notification.updated_at:
-            folder = get_folder_name(notification.updated_at, False)
-        else:
-            folder = get_folder_name(notification.created_at, False)
+        folder = get_folder_name(notification.created_at, False)
```
leohemsted: it would be nice to use a keyword param here rather than having a context-less boolean being passed in (so `get_folder_name(notification.created_at, dont_use_sending_date=False)`)

if you want, we can force that in the function definition by including a `*` param before. `def get_folder_name(_now, *, dont_use_sending_date)` will force you to make dont_use_sending_date a keyword param

servingUpAces: oh I didn't know you could do that - nice

app/letters/utils.py
leohemsted: decision: lets not worry about this in this PR

app/letters/utils.py
leohemsted: decision: lets not worry about this in this PR

PR #2616, opened 2019-09-19
Replace negative contractions in the API

karlchillmaid:
Some users find negative contractions harder to understand. We use them a lot on Notify.

The latest [guidance from GOV.UK](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#contractions) says:

>Avoid negative contractions like can’t and don’t. Many users find them harder to read, or misread them as the opposite of what they say. Use cannot, instead of can’t.
>Avoid should’ve, could’ve, would’ve, they’ve too. These can also be hard to read.

This PR replaces:

- can't with cannot
- don't with do not
- won't with will not
- isn't with is not
- didn't with did not
- couldn't with could not
- doesn't with does not
- aren't with are not
- hasn't with has not
- haven't with have not

For more information, read: [Using contractions could be making your writing inaccessible](https://medium.com/@joanne.schofield/using-contractions-could-be-making-your-writing-inaccessible-4034bb3cd76a).


servingUpAces:
After a chat with the rest of the dev team - we've decided that this change is too risky. It is possible that there are teams that rely on the content of the error message. Therefore we need to treat this change as a breaking change. Typically that would mean we need to version the api.

PR #2617, opened 2019-09-24
Return count of notifications in the database for a job

CrystalPea:
When we cancel a job, we need to check if all notifications are already in the database.
So far, we were querying for serialised notification objects in the database and counting them in admin app, which runs into pagination problems for large jobs, and could time out for very large jobs.

Admin part: https://github.com/alphagov/notifications-admin/pull/3130
Bug: https://www.pivotaltracker.com/story/show/168734905

### DIFF
--- a/app/job/rest.py
+++ b/app/job/rest.py
@@ -110,7 +110,7 @@ def get_all_notifications_for_service_job(service_id, job_id):
 
 @job_blueprint.route('/<job_id>/notification_count', methods=['GET'])
 def get_notification_count_for_job_id(service_id, job_id):
-    count = dao_get_notification_count_for_job_id(job_id)
+    count = dao_get_notification_count_for_job_id(service_id=service_id, job_id=job_id)
     return jsonify(
         count=count
     ), 200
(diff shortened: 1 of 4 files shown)


app/job/rest.py
```diff
@@ -108,6 +108,14 @@ def get_all_notifications_for_service_job(service_id, job_id):
     ), 200
 
 
+@job_blueprint.route('/<job_id>/notification_count', methods=['GET'])
+def get_notification_count_for_job_id(service_id, job_id):
+    count = dao_get_notification_count_for_job_id(job_id)
```
servingUpAces: we should make sure the service id is valid, the easiest way to do this is to include service_id in the query.

servingUpAces: If the job_id does not exist the query still returns 0 - is that expected behaviour?

CrystalPea: Thanks, I pushed those changes now ☺️

servingUpAces: The api will return 0 if the service or job is not found. That's not really expected behaviour. I've made an update - as I thought it's easier to show than tell.

CrystalPea: @servingUpAces thanks, only I cannot see the update

CrystalPea: Ahh, I see, makes sense 👍

PR #2626, opened 2019-10-08
Delete dm_datetime database table

idavidmcdonald:
No longer needed

See https://github.com/alphagov/notifications-api/pull/2625 which has already gone out to remove the related sqlalchemy model first

### DIFF
--- a/migrations/versions/0307_delete_dm_datetime.py
+++ b/migrations/versions/0307_delete_dm_datetime.py
@@ -1,0 +1,43 @@
+"""
+
+Revision ID: 0307_delete_dm_datetime
+Revises: 0306_letter_rates_price_rise
+Create Date: 2019-10-08 10:57:54.824807
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = '0307_delete_dm_datetime'
+down_revision = '0306_letter_rates_price_rise'
+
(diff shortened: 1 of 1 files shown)


migrations/versions/307_delete_dm_datetime.py
```diff
@@ -0,0 +1,43 @@
+"""
+
+Revision ID: 307
```
klssmith: The revision ID needs to match the file name here and on line 12. (Because we compare the latest migration file name to the revision ID in the database to see if we need to run test the migrations when deploying)

leohemsted: filename should be `0307` so that it's alphabetical in the file directory.


Also, the revision in code should be the same as the filename - we compare the alembic version in the database (defined by `revision`) to the files present on disk when deciding whether to apply database migrations during deploys. see https://github.com/alphagov/notifications-api/blob/master/scripts/check_if_new_migration.py. we could probably avoid this quirk by calling `flask db heads`, but still, it's good to have a descriptive name thats easy to search for on disk.

idavidmcdonald: Thanks, believe I've addressed your comment

idavidmcdonald: Thanks, believe this is addressed now

PR #2632, opened 2019-10-14
scale up api to 25 instances before deploy on production

leohemsted:
deploys take up to five minutes, during which notify-paas-autoscaler can't scale the app. We saw 502s due to a large volume of traffic coming in during that time, and we couldn't react cos we were deploying.

scale up to 25 instances, the autoscaler won't be able to downscale until after the deploy has finished.


CrystalPea:
Sorry if this is a silly question, but how does the system know that it should only hard-code instances to 25 on deploy, but should be flexible once the app is deployed?

PR #2634, opened 2019-10-15
set startup timeout to 10 minutes

leohemsted:
5 minutes isn't long enough to scale api at load, leading to failed deploys

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ cf-deploy: ## Deploys the app to Cloud Foundry
 	cf v3-cancel-zdt-push ${CF_APP} || true
 
 	cf v3-apply-manifest ${CF_APP} -f <(make -s generate-manifest)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -191,7 +191,7 @@ cf-deploy: ## Deploys the app to Cloud Foundry
 	cf v3-cancel-zdt-push ${CF_APP} || true
 
 	cf v3-apply-manifest ${CF_APP} -f <(make -s generate-manifest)
-	cf v3-zdt-push ${CF_APP} --wait-for-deploy-complete  # fails after 5 mins if deploy doesn't work
+	CF_STARTUP_TIMEOUT=10 cf v3-zdt-push ${CF_APP} --wait-for-deploy-complete  # fails after 5 mins if deploy doesn't work
```
idavidmcdonald: is the comment at the end of this line now incorrect?

leohemsted: yeah :(. should have caught that. blame andy he told me to merge it

PR #2636, opened 2019-10-17
Change to the _status endpoint

servingUpAces:
The _status endpoint returns data about the deployed build which includes the dB version. The dB is queried to get the alembic_version, this is a quick query but is a query none the less.
I'm not convinced that this endpoint needs to use a dB connection to return this data.
I also renamed the attributes of the json to stop referring to travis.

Sound like a good idea?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ production: ## Set environment to production
 
 .PHONY: generate-version-file
 generate-version-file: ## Generates the app version file
-	@echo -e "__travis_commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"\n__travis_job_number__ = \"${BUILD_NUMBER}\"\n__travis_job_url__ = \"${BUILD_URL}\"" > ${APP_VERSION_FILE}
+	@echo -e "__git_commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"\n__jenkins_job_number__ = \"${BUILD_NUMBER}\"\n__jenkins_job_url__ = \"${BUILD_URL}\"" > ${APP_VERSION_FILE}
 
 .PHONY: build-paas-artifact
 build-paas-artifact:  ## Build the deploy artifact for PaaS
--- a/app/status/healthcheck.py
+++ b/app/status/healthcheck.py
@@ -4,7 +4,10 @@ from flask import (
     request
 )
 
-from app import version
+from app import (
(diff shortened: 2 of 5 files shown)


app/status/healthcheck.py
```diff
@@ -19,10 +19,10 @@ def show_status():
     else:
         return jsonify(
             status="ok",  # This should be considered part of the public API
-            travis_commit=version.__travis_commit__,
-            travis_build_number=version.__travis_job_number__,
-            build_time=version.__time__,
-            db_version=get_db_version()), 200
```
klssmith: We are using the db_version in this file to see if we need to test the migrations when deploying api https://github.com/alphagov/notifications-api/blob/0340927a6e0eb7a5e20bce6e569744f4f1895502/scripts/check_if_new_migration.py

servingUpAces: Ah shoot... We might be able to use a different strategy for that. Good catch

Makefile
```diff
@@ -52,7 +52,7 @@ production: ## Set environment to production
 
 .PHONY: generate-version-file
 generate-version-file: ## Generates the app version file
-	@echo -e "__travis_commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"\n__travis_job_number__ = \"${BUILD_NUMBER}\"\n__travis_job_url__ = \"${BUILD_URL}\"" > ${APP_VERSION_FILE}
+	@echo -e "__git_commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"\n__jenkins_job_number__ = \"${BUILD_NUMBER}\"\n__jenkins_job_url__ = \"${BUILD_URL}\"" > ${APP_VERSION_FILE}
```
idavidmcdonald: `jenkins` might need to be changed in a few weeks time. Fine with this going in but could be worth changing it to something more generic like `ci_job_number` etc. I'll leave for you to decide

tests/app/status/test_status.py
```diff
@@ -10,9 +10,8 @@ def test_get_status_all_ok(client, notify_db_session, path):
     assert response.status_code == 200
     resp_json = json.loads(response.get_data(as_text=True))
     assert resp_json['status'] == 'ok'
-    assert resp_json['db_version']
```
idavidmcdonald: Might be nice to add a real small unit test for our endpoint please

PR #2639, opened 2019-10-23
remove loadtesting from the database

leohemsted:
requires https://github.com/alphagov/notifications-api/pull/2638

### DIFF
--- a/migrations/versions/0308_delete_loadtesting_provider.py
+++ b/migrations/versions/0308_delete_loadtesting_provider.py
@@ -1,0 +1,39 @@
+"""
+Remove loadtesting provider
+
+Revision ID: 0308_delete_loadtesting_provider
+Revises: 0307_delete_dm_datetime
+Create Date: 2019-10-22 17:30
+
+"""
+import uuid
+from alembic import op
+from sqlalchemy.sql import text
+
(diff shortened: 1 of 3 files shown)


migrations/versions/0308_delete_loadtesting_provider.py
```diff
@@ -0,0 +1,28 @@
...
+def upgrade():
+    conn = op.get_bind()
+    conn.execute("DELETE FROM provider_details WHERE identifier = 'loadtesting'")
+    conn.execute("DELETE FROM provider_details_history WHERE identifier = 'loadtesting'")
+
+def downgrade():
+    conn = op.get_bind()
+    conn.execute(
+        """
+        INSERT INTO provider_details (id, display_name, identifier, priority, notification_type, active, updated_at, version, created_by_id, supports_international) VALUES
```
idavidmcdonald: Is this line valid SQL? No defined values on the end.

Feels unlikely that we would downgrade but just checking

leohemsted: no ill fix that

leohemsted: i have fixed that

PR #2642, opened 2019-10-23
Add request_ids to some Celery tasks

klssmith:
**Requirement bumps**
Bumped Utils to bring in changes to logging the request_id (https://github.com/alphagov/notifications-utils/pull/658). Freezing requirements updated a few other dependencies too, including Werkzeug from version 0.15.5 to 0.16.0, which required a small change to an import (https://github.com/pallets/werkzeug/blob/master/CHANGES.rst#version-0160).

**Passing the `request_id` to tasks if available**
We want to pass the `request_id` to Celery tasks if the task is called from an HTTP request, so that we can log it. This change overwrites `apply_async` to add the `request_id` to the kwargs if available. When we call the task, we then add the `request_id` to `g` on Flask's application context.

Tasks called from `send_task` won't have a `request_id` for now, and this change only affects tasks called from HTTP requests (not from other tasks or from Celery Beat).

### DIFF
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -2,7 +2,8 @@ import time
 
 from celery import Celery, Task
 from celery.signals import worker_process_shutdown
-from flask import current_app
+from flask import current_app, g, request
+from flask.ctx import has_request_context
 
 
 @worker_process_shutdown.connect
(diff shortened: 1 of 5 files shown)


klssmith:
The tests will fail for this until the Utils master build is fixed

PR #2643, opened 2019-10-25
Set postage

idavidmcdonald:
https://www.pivotaltracker.com/story/show/167534014

- Require postage key for API calls to send an uploaded PDF letter
- Set postage in the DB as sent from API call
- Use JSON schema validation for requests to this endpoint

This will lay the groundwork for https://github.com/alphagov/notifications-admin/pull/3144 to be finished off

I've tested this against 
- https://github.com/alphagov/notifications-admin/pull/3144
- notifications-admin master

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -93,6 +93,7 @@ from app.models import (
 from app.notifications.process_notifications import persist_notification, send_notification_to_queue
 from app.schema_validation import validate
 from app.service import statistics
+from app.service.send_pdf_letter_schema import send_pdf_letter_request
 from app.service.service_data_retention_schema import (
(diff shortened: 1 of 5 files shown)


idavidmcdonald:
So looks like it is in use whenever we mark a json schema field with `"format": "postage"` ie. https://github.com/alphagov/notifications-api/blob/276a9a38283ef1c1e1e3eda14db5b74206fbf232/app/v2/notifications/notification_schemas.py#L243

Probably makes sense for me to use that too when validating postage

idavidmcdonald:
@servingUpAces I've added one more commit to change and use the postage format. Let me know what you think please

PR #2649, opened 2019-10-31
Upgrade to Pytest 5 😅

klssmith:
Changed the code to be Pytest 5 compatible by
- stopping fixtures being called like functions (almost all of the changes in this PR)
- changing how we check for error messages of exceptions

`conftest.py` can be tidied up as a separate PR - the fixtures there don't need any parameters (since they can't be called as functions anymore). There are also some fixtures in the file which aren't being used anywhere, so can be deleted.

### DIFF
--- a/requirements_for_test.txt
+++ b/requirements_for_test.txt
@@ -1,7 +1,7 @@
 -r requirements.txt
 flake8==3.7.7
 moto==1.3.8
-pytest==3.10.1  # pyup: <4
+pytest==5.2.2
 pytest-env==0.6.2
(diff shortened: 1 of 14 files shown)


requirements_for_test.txt
```diff
@@ -1,7 +1,7 @@
 -r requirements.txt
 flake8==3.7.7
 moto==1.3.8
-pytest==3.10.1  # pyup: <4
+pytest==5.2.2
```
leohemsted: 🙌🙌🙌🙌🙌🙌🙌🙌🙌

PR #2652, opened 2019-11-05
Provide feedback on incorrect Content-Type

bambrp:
When calling the API with a Content-Type that isn't `application/json` (even if there is a JSON body), calls to `request.get_json()` return None, and this isn't caught very well- you get a ValidationError with "None is not of type object" as the message.

This PR catches the bad API call and returns a more helpful error message.

### To test
Bring up the API and attempt to call endpoint with a Content-Type of, say `application/x-www-form-urlencoded`, with or without JSON in the body. With the PR applied, you should still get a ValidationError, but with a message reading "Request body arguments not JSON or incorrect Content-Type header supplied."

### DIFF
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -56,6 +56,8 @@ def validate_schema_date_with_hour(instance):
 
 
 def validate(json_to_validate, schema):
+    if json_to_validate == None:
+        raise ValidationError("Request body arguments not JSON or incorrect Content-Type header supplied.")
     validator = Draft7Validator(schema, format_checker=format_checker)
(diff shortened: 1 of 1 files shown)


leohemsted:
hi @bambrp, thanks for the PR! It seems like a sensible change, though I'll have to have a think about if that's the best place in the code to put it (gut feel: it probably is). I'll also run the error messages past our content designer and we'll need to add a unit test for this, but otherwise, thanks for the suggestion!

servingUpAces:
Thank you for you suggestion. I'm going to close the PR in favour of https://github.com/alphagov/notifications-api/pull/2654, which includes some unit tests.

PR #2653, opened 2019-11-05
Check for missing rows

servingUpAces:
Adding a scheduled task to processing missing rows from job 
# Why
Sometimes a job finishes but has missed a row in the middle. It is a mystery why this is happening, it could be that the task to save the notifications has been dropped.
So until we solve the missing let's find missing rows and process them.

# What
A new scheduled task has been added to find any "finished" jobs that do not have enough notifications created. If there are missing notifications the job processes those rows for the job.

A unique key constraint has been added to Notifications to ensure that the row is not added twice. Any index or constraint can affect performance, but this unique constraint should not affect it enough for us to notice.

Added cron for check_for_missing_rows_in_completed_jobs. Run the task every 10 minutes. Does that seem reasonable? Maybe that is too often.


app/config.py
```diff
@@ -198,6 +198,11 @@ class Config(object):
             'schedule': crontab(),
             'options': {'queue': QueueNames.PERIODIC}
         },
+        'check-for-missing-rows-in-completed-jobs': {
+            'task': 'check-for-missing-rows-in-completed-jobs',
+            'schedule': crontab(minute='*/10'),
```
CrystalPea: I think that if it is not heavy on our resources, checking every 10 minutes should be fine. Some services use Notify to send emergency messages, so it is good if they are all processed as soon as possible I think.

tests/app/dao/test_jobs_dao.py
```diff
@@ -409,3 +413,105 @@ def test_can_letter_job_be_cancelled_returns_false_and_error_message_if_notifica
     result, errors = can_letter_job_be_cancelled(job)
     assert not result
     assert errors == "We are still processing these letters, please try again in a minute."
+
+
+def test_find_jobs_with_missing_rows(sample_email_template):
```
CrystalPea: in this test we could also create a job that is finished and not missing any rows, and then show that `find_jobs_with_missing_rows` is only finding the job with missing rows, but not the "healthy" job.

servingUpAces: Good shout! there's a bug with the group by

PR #2654, opened 2019-11-06
Add validation for missing content-type header

servingUpAces:
We require the POST data to be in JSON format and that the content-type header be set to application/json.
This PR does that validation and returns a decent message. Previously the error message was a None type error which is not helpful to anyone.

Thank you to @bambrp for bring this to our attention. I've closed your PR, https://github.com/alphagov/notifications-api/pull/2652, in favour of this one as I have included some unit tests and updated the post_template endpoint because it had to be handled specially since the data can be empty for that request.

### DIFF
--- a/app/v2/template/post_template.py
+++ b/app/v2/template/post_template.py
@@ -6,17 +6,21 @@ from app.schema_validation import validate
 from app.utils import get_template_instance
 from app.v2.errors import BadRequestError
 from app.v2.template import v2_template_blueprint
-from app.v2.template.template_schemas import post_template_preview_request, create_post_template_preview_response
+from app.v2.template.template_schemas import (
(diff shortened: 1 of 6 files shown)


app/v2/template/post_template.py
```diff
@@ -6,14 +6,21 @@
...
+    create_post_template_preview_response
+)
+from app.v2.utils import get_valid_json
 
 
 @v2_template_blueprint.route("/<template_id>/preview", methods=['POST'])
 def post_template_preview(template_id):
-    _data = request.get_json()
-    if _data is None:
+    # If the payload is is empty if there is no personalisation in the template.
```
idavidmcdonald: Typo in here if you are picky

PR #2657, opened 2019-11-08
Validate against messages with no content

CrystalPea:
### In this PR:
Validate against a message with no content for one off messages.

Needs to go in after this is merged: https://github.com/alphagov/notifications-utils/pull/671

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -123,7 +123,7 @@ def check_sms_content_char_count(content_count):
 
 def check_notification_content_is_not_empty(template_with_content):
     if template_with_content.is_message_empty():
-        message = 'This message is empty.'
+        message = 'Your message is empty.'
         raise BadRequestError(message=message)
 
 
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -121,6 +121,12 @@ def check_sms_content_char_count(content_count):
         raise BadRequestError(message=message)
 
 
+def check_if_notification_content_is_not_empty(template_with_content):
+    if len(template_with_content.__str__()) == 0:
+        message = 'This message is empty.'
+        raise BadRequestError(message=message)
+
+
 def validate_template(template_id, personalisation, service, notification_type):
     try:
         template = templates_dao.dao_get_template_by_id_and_service_id(
@@ -137,6 +143,7 @@ def validate_template(template_id, personalisation, service, notification_type):
     template_with_content = create_content_for_notification(template, personalisation)
     if template.template_type == SMS_TYPE:
         check_sms_content_char_count(template_with_content.content_count)
+    check_if_notification_content_is_not_empty(template_with_content)
     return template, template_with_content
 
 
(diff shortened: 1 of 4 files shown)


app/notifications/validators.py
```diff
@@ -121,6 +121,12 @@ def check_sms_content_char_count(content_count):
         raise BadRequestError(message=message)
 
 
+def check_if_notification_has_any_content(content_count):
+    if content_count == 0:
+        message = 'This message is empty.'
```
CrystalPea: @karlchillmaid I wrote this message, so you might want to take a look :)

app/notifications/validators.py
```diff
@@ -137,6 +143,7 @@ def validate_template(template_id, personalisation, service, notification_type):
     template_with_content = create_content_for_notification(template, personalisation)
     if template.template_type == SMS_TYPE:
         check_sms_content_char_count(template_with_content.content_count)
+    check_if_notification_has_any_content(len(template_with_content.__str__()))
```
idavidmcdonald: Based on the name of this method, I'd expect to pass it a notification, rather than the length of the notification already. It feels less useful as a standalone method if it is only checking if a number is bigger than 0. I'd suggest changing it so that you call it like:

`check_if_notification_has_any_content(template_with_content)`

and then the method is responsible for calling `__str__` and `len`.

You may also consider changing the name of the method to `check_notification_content_is_not_empty` or `check_notification_has_content`, or `check_notification_has_non_empty_content` as I don't think the word `any` is particularly useful and this method name could be a little more descriptive. 

I think I'd also like to see a `content_count` method added for emails and letters for consistency so actually it might look something more like:

```
def check_notification_content_is_not_empty(template_with_content):
    char_count =template_with_content.content_count()
    if char_count == 0:
        message = 'This message is empty.'
```

app/celery/tasks.py
```diff
@@ -306,6 +327,15 @@ def save_letter(
     service = dao_fetch_service_by_id(service_id)
     template = dao_get_template_by_id(notification['template'], version=notification['template_version'])
 
+    personalisation = notification.get('personalisation')
+    try:
+        validate_template(template.id, personalisation, service, template.template_type)
+    except BadRequestError as e:
+        current_app.logger.debug(
+            "SMS {} failed: {}".format(notification_id, e.message)
```
idavidmcdonald: This says the word SMS in but the function it's running in is called `save_letter`. Assume this will be changed.

app/celery/tasks.py
```diff
@@ -254,6 +266,15 @@ def save_email(self,
...
```
idavidmcdonald: This says the word SMS in but the function it's running in is called save_email. Assume this will be changed later.

app/notifications/validators.py
```diff
@@ -123,7 +123,7 @@ def check_sms_content_char_count(content_count):
 
 def check_if_notification_content_is_not_empty(template_with_content):
     if template_with_content.is_message_empty():
-        message = 'This message is empty.'
+        message = 'Your message is empty.'
```
karlchillmaid: Content looks 👍

PR #2658, opened 2019-11-14
Alert if a letter doesn't make it past created status

servingUpAces:
Add an alert when a letter is created but doesn't have a file in S3 for sending. We can tell this is the case because there is no updated_at and billable units are still 0.

It is safe to run the app.celery.letters_pdf_tasks.create_letters_pdf task again even if the letter is already in the S3 bucket. Which also means we could shorten the time we wait before check the letter. 

I've reduced the amount of time that we wait before replaying the task to send or create pdf for the notification. This is possible because the tasks can be run twice without harm.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -191,7 +191,7 @@ def replay_created_notifications():
         for n in notifications_to_resend:
             send_notification_to_queue(notification=n, research_mode=n.service.research_mode)
 
-    # if the letter has not be send after 4 hours + 15 minutes, then create a zendesk ticket
+    # if the letter has not be send after an hour, then create a zendesk ticket
     letters = letters_missing_from_sending_bucket(resend_created_notifications_older_than)
 
     if len(letters) > 0:
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -30,6 +30,7 @@ from app.dao.notifications_dao import (
     notifications_not_yet_sent,
     dao_precompiled_letters_still_pending_virus_check,
     dao_old_letters_with_created_status,
+    letters_missing_from_sending_bucket
 )
 from app.dao.provider_details_dao import (
     get_current_provider,
@@ -188,6 +189,25 @@ def replay_created_notifications():
 
         for n in notifications_to_resend:
             send_notification_to_queue(notification=n, research_mode=n.service.research_mode)
+
+    # if the letter has not be send after 4 hours + 15 minutes, then create a zendesk ticket
+    letters = letters_missing_from_sending_bucket(resend_created_notifications_older_than)
+
+    if len(letters) > 0:
+        msg = "{} letters were created four hours and 15 minutes ago, " \
+              "but do not have an updated_at timestamp or billable units. " \
+              "It is likely you need to run the " \
+              "app.celery.letters_pdf_tasks.create_letters_pdf task again with " \
+              "the notification id.\n {}".format(len(letters),
+                                                 [x.id for x in letters])
+        current_app.logger.info(msg)
+        if current_app.config['NOTIFY_ENVIRONMENT'] in ['live', 'production', 'test']:
+            zendesk_client.create_ticket(
+                subject="[{}] Letters still in created status might be missing from S3".format(
+                    current_app.config['NOTIFY_ENVIRONMENT']),
+                message=msg,
+                ticket_type=zendesk_client.TYPE_INCIDENT
+            )
 
 
 @notify_celery.task(name='check-precompiled-letter-state')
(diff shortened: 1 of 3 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -189,6 +190,25 @@ def replay_created_notifications():
         for n in notifications_to_resend:
             send_notification_to_queue(notification=n, research_mode=n.service.research_mode)
 
+    # if the letter has not be send after 4 hours + 15 minutes, then create a zendesk ticket
+    letters = letters_missing_from_sending_bucket(resend_created_notifications_older_than)
+
+    if len(letters) > 0:
+        msg = "{} letters were created four hours and 15 minutes ago, " \
+              "but do not have an updated_at timestamp or billable units. " \
+              "It is likely you need to run the " \
```
idavidmcdonald: I'm liking that this ticket has how to action included on it. Would be nice to make the suggested action a little bit more precise if possible. I'm guessing something along the lines of:

- First check if the letter does exist in the sending bucket (optionally saying how to do this if not obvious or if not clear what the name of the bucket is)
- What to do in either case when the letter does or does not exist in the bucket.

I'm fine with it just being a zendesk ticket at the moment, as long as we keep an eye on if we can swap to automate it later to just rerun the task if we are happy that is what should happen.

app/dao/notifications_dao.py
```diff
@@ -698,12 +699,29 @@ def dao_old_letters_with_created_status():
...
+        Notification.created_at
+    ).all()
+    return notifications
+
+
+def letters_missing_from_sending_bucket(seconds_to_subtract):
+    older_than_date = datetime.utcnow() - timedelta(seconds=seconds_to_subtract)
+
+    notifications = Notification.query.filter(
+        Notification.updated_at == None,  # noqa
```
idavidmcdonald: Optional: As your commit message says, it is letters that have not been updated and have no billable units that are missing. Therefore, just to make this very clear in the code it would be nice to have a small code comment saying this. You could also order the query arguments so those two arguments are next to each other and stand out (.e.g above mentioned comment or a line break)

app/dao/notifications_dao.py
```diff
@@ -698,12 +699,29 @@ def dao_old_letters_with_created_status():
     last_processing_deadline = yesterday_bst.replace(hour=17, minute=30, second=0, microsecond=0)
 
     notifications = Notification.query.filter(
-        Notification.updated_at < convert_bst_to_utc(last_processing_deadline),
+        Notification.created_at < convert_bst_to_utc(last_processing_deadline),
```
idavidmcdonald: Is this fixing a different bug not mentioned in the PR? Just trying to figure out how it is related

servingUpAces: We changed this to be `updated_at` to eliminate false positives, for example when a letter is created at 17:29, but doesn't get updated and in the sent bucket until 17:31 (this triggers the alert but the letter ended up in tomorrows bucket).
However, changing the check meant we were not getting alerted when we should. We decided that it's better to get a couple alerts that turn out to be ok then never get the alert.

klssmith:
If we're now automatically fixing letters that are still `created` by uploading them to S3 again, is there any risk that we could end with a letter being uploaded again to a bucket for a different day and then getting sent twice?

servingUpAces:
Yes that is a risk, which is why I want to shorten the length of time we wait before triggering this alert. 
Or... we need to check for the letter in the other bucket.

app/celery/scheduled_tasks.py
```diff
@@ -189,6 +191,20 @@ def replay_created_notifications():
         for n in notifications_to_resend:
             send_notification_to_queue(notification=n, research_mode=n.service.research_mode)
 
+    # if the letter has not be send after 4 hours + 15 minutes, then create a zendesk ticket
```
idavidmcdonald: This comment no longer matches the 1 hour time frame

app/celery/scheduled_tasks.py
```diff
@@ -173,8 +175,8 @@ def check_job_status():
 @notify_celery.task(name='replay-created-notifications')
 @statsd(namespace="tasks")
 def replay_created_notifications():
-    # if the notification has not be send after 4 hours + 15 minutes, then try to resend.
-    resend_created_notifications_older_than = (60 * 60 * 4) + (60 * 15)
+    # if the notification has not be send after 1 hour, then try to resend.
```
idavidmcdonald: We are OK to resend emails and SMS after an hour then I take it?

servingUpAces: Yes - the task to send email/SMS can be called twice since it checks the notifications status. You shouldn't get the email/SMS sent twice.

PR #2659, opened 2019-11-14
Get sender id from metadata 

servingUpAces:
# What

When a job is created the user can select a sender id (if there is more than one). When the sender id is selected it is added to the metadata of the file. However, there are three other cases were we process data from the CSV: 
 - scheduled jobs
 - a job that has bee restarted (if an app that is processing a job is stopped before the job is done, then we have a check that will look for this and restart the job)
 - there is an odd occurrence where a row from the file is skipped, we have a check for this and will process the missing row.  
For each of these cases the sender id is not used. 

This PR uses the sender id from the CSV file meta data. This will make the need for passing the sender_id into the `process_job` task redundant and can be removed in the next PR.

### DIFF
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -37,6 +37,11 @@ def get_job_location(service_id, job_id):
     )
 
 
+def get_job_and_metadata_from_s3(service_id, job_id):
+    obj = get_s3_object(*get_job_location(service_id, job_id))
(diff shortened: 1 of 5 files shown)


servingUpAces:
@idavidmcdonald - the `check_job_status` task creates a task `app.celery.tasks.process_incomplete_jobs`, and the `app.celery.tasks.process_incomplete_job` method has the fix in it.

PR #2662, opened 2019-11-18
Utils bump

leohemsted:
most of the bumps come from https://github.com/alphagov/notifications-api/pull/2648, except for marshmallow which broke some stuff so i left it out

### DIFF
--- a/app/notifications/rest.py
+++ b/app/notifications/rest.py
@@ -200,7 +200,7 @@ def create_template_object_for_notification(template, personalisation):
 
     if (
         template_object.template_type == SMS_TYPE and
-        template_object.content_count > SMS_CHAR_COUNT_LIMIT
+        template_object.is_message_too_long()
     ):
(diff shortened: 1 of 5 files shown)


app/notifications/rest.py
```diff
@@ -198,10 +198,7 @@ def create_template_object_for_notification(template, personalisation):
         errors = {'template': [message]}
         raise InvalidRequest(errors, status_code=400)
 
-    if (
-        template_object.template_type == SMS_TYPE and
-        template_object.content_count > SMS_CHAR_COUNT_LIMIT
-    ):
+    if template_object.is_message_too_long():
```
idavidmcdonald: Optional: I'd be tempted to keep the check of the template type for SMS given that the content in the if statement only applies for SMS messages. Whilst nothing would break now, this feels a bit more defensive if we changed `is_message_too_long` for emails/letters and did not spot this.

leohemsted: yeah i think you're right. we'll presumably take a look and refactor these anyway if we end up adding a check for the others (it's feasible we might add a pdf page count check at this point)

PR #2667, opened 2019-12-04
Warn team about high failure rates

CrystalPea:
Log an exception and send a zendesk ticket when, in the last 24 hours:
- there were services with failure rate over 25% for their sms with at least 100 messages sent
- there were services that sent at least 100 sms messages to tv numbers

Story ticket: https://www.pivotaltracker.com/n/projects/1443052

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -45,6 +45,8 @@ from app.models import (
 from app.notifications.process_notifications import send_notification_to_queue
 from app.v2.errors import JobIncompleteError
 
+from app.service.utils import get_services_with_high_failure_rates
+
 
 @notify_celery.task(name="run-scheduled-jobs")
 @statsd(namespace="tasks")
@@ -253,3 +255,28 @@ def check_for_missing_rows_in_completed_jobs():
                     current_app.logger.info(
                         "Processing missing row: {} for job: {}".format(row_to_process.missing_row, job.id))
                     process_row(row, template, job, job.service, sender_id=sender_id)
+
+
+@notify_celery.task(name='check-for-services-with-high-failure-rates-or-sending-to-tv-numbers')
+@statsd(namespace="tasks")
+def check_for_services_with_high_failure_rates_or_sending_to_tv_numbers():
+    services_with_failures = get_services_with_high_failure_rates()
+    # services_sending_to_tv_numbers = dao_find_services_sending_to_tv_numbers(number=100)
+
+    if services_with_failures:
+        message = "{} service(s) have had high permanent-failure rates for sms messages in last 24 hours: ".format(
+            len(services_with_failures)
+        )
+        for service in services_with_failures:
+            message += "service id: {} failure rate: {}, ".format(service["id"], service["permanent_failure_rate"])
+
+        current_app.logger.exception(message)
+
+        if current_app.config['NOTIFY_ENVIRONMENT'] in ['live', 'production', 'test']:
+            zendesk_client.create_ticket(
+                subject="[{}] High failure rates for sms spotted for services".format(
+                    current_app.config['NOTIFY_ENVIRONMENT']
+                ),
+                message=message,
+                ticket_type=zendesk_client.TYPE_INCIDENT
+            )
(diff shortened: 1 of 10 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -253,3 +256,42 @@ def check_for_missing_rows_in_completed_jobs():
...
+        for service in services_sending_to_tv_numbers:
+            message += "service id: {}, count of sms to tv numbers: {},\n".format(
+                service.service_id, service.notification_count
+            )
+
+    if services_with_failures or services_sending_to_tv_numbers:
+        current_app.logger.exception(message)
+
+        if current_app.config['NOTIFY_ENVIRONMENT'] in ['live', 'production', 'test']:
+            message += "\nThings to do: contact service? revoke their key?"
```
CrystalPea: Explain what TV numbers are:
These are numbers reserved for the use in TV and drama. The range of TV numbers are 07700 900000 to 900999. When an SMS notification is sent to such a number, we get a 'permanent-failure' status back and we are billed for this notification.

idavidmcdonald: Thanks. That first sentence would be useful to put in the TODO section or in the runbook so a developer receiving the ticket knows how to deal with it

CrystalPea: yeah, I wrote it to myself here only to remember about including it :). As it only transpired in a spoken conversation.

CrystalPea:
@servingUpAces I refactored this code as per our conversations now, what do you think? 

@idavidmcdonald I created a section for this ticket in the Manual and linked it from the ticket.

idavidmcdonald:
@CrystalPea Excellent runbook entry :)

tests/app/dao/test_services_dao.py
```diff
@@ -1099,3 +1071,79 @@ def create_email_sms_letter_template():
     template_two = create_template(service=service, template_name='2', template_type='sms')
     template_three = create_template(service=service, template_name='3', template_type='letter')
     return template_one, template_three, template_two
+
+
+@freeze_time("2019-12-02 12:00:00.000000")
```
servingUpAces: Do we need to freeze_time for this test?

tests/app/dao/test_services_dao.py
```diff
@@ -1099,3 +1071,79 @@ def create_email_sms_letter_template():
...
+def test_dao_find_services_sending_to_tv_numbers(notify_db_session, fake_uuid):
+    service_1 = create_service(service_name="Service 1", service_id=fake_uuid)
```
servingUpAces: Do we need to use fake_uuid?

tests/app/dao/test_services_dao.py
```diff
@@ -1099,3 +1071,79 @@ def create_email_sms_letter_template():
...
+
+    service_6 = create_service(service_name="Service 6")  # notifications too old are excluded
+    with freeze_time("2019-11-30 15:00:00.000000"):
+        template_6 = create_template(service_6)
+        for x in range(0, 5):
+            create_notification(template_6, normalised_to=tv_number, status="permanent-failure")
+
+    service_2 = create_service(service_name="Service 2")  # below threshold is excluded
+    template_2 = create_template(service_2)
+    create_notification(template_2, normalised_to=tv_number, status="permanent-failure")
```
servingUpAces: Should we create some `test` notifications?

tests/app/dao/test_services_dao.py
CrystalPea: I think we do, to show that most notifications are in good date range, while notifications for service_6 are just a bit too old and do not make it to results.

tests/app/dao/test_services_dao.py
CrystalPea: We need an uuid, as we assert on it below, to check the result we get is for the right service.

tests/app/dao/test_services_dao.py
CrystalPea: Good idea, I will add those.

tests/app/dao/test_services_dao.py
servingUpAces: You can just use the id from service_1.id

PR #2669, opened 2019-12-05
split create_nightly_notification_status by notification type

leohemsted:
Separate this refactoring PR from the actual logic change that is coming soon™ (to make it rebuild for letters more often). Just splitting up the task by notification_type - it now requires type all the time. This is fine to merge provided that the task isn't running as it deploys, but given that it runs nightly at 4am or something that's probably fine

This also allows us to rebuild historical data by type, rather than for all three notification types.

And it makes the command just call through to the task, rather than re-implementing it.

### DIFF
--- a/tests/app/celery/test_reporting_tasks.py
+++ b/tests/app/celery/test_reporting_tasks.py
@@ -1,9 +1,11 @@
+import itertools
 from datetime import datetime, timedelta, date
 from decimal import Decimal
 
 import pytest
 from freezegun import freeze_time
 
+from app.config import QueueNames
 from app.celery.reporting_tasks import (
     create_nightly_billing,
     create_nightly_notification_status,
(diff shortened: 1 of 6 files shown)


tests/app/celery/test_reporting_tasks.py
```diff
@@ -56,9 +58,12 @@ def test_create_nightly_notification_status_triggers_tasks_for_days(notify_api,
     mock_celery = mocker.patch('app.celery.reporting_tasks.create_nightly_notification_status_for_day')
     create_nightly_notification_status(day_start)
 
-    assert mock_celery.apply_async.call_count == 4
-    for i in range(4):
-        assert mock_celery.apply_async.call_args_list[i][1]['kwargs'] == {'process_day': expected_kwargs[i]}
+    assert mock_celery.apply_async.call_count == 4 * 3  # four days, three notification types
```
idavidmcdonald: Nice comment

leohemsted: wait till you see my next PR where this got more complicated 😅

PR #2670, opened 2019-12-05
Uploads endpoint

servingUpAces:
We need a page on the admin app to show both uploaded csv files and uploaded letters (excluded uploaded letters from the api). 

This endpoint combines the data for jobs and uploaded letters in one resultset for the admin app. 
The admin PR is here: https://github.com/alphagov/notifications-admin/pull/3219

### DIFF
--- a/tests/app/dao/test_uploads_dao.py
+++ b/tests/app/dao/test_uploads_dao.py
@@ -1,0 +1,168 @@
+from datetime import datetime, timedelta
+
+from app.dao.uploads_dao import dao_get_uploads_by_service_id
+from app.models import LETTER_TYPE, JOB_STATUS_IN_PROGRESS
+from tests.app.db import create_job, create_service, create_template, create_notification
+
+
+def create_uploaded_letter(letter_template, service, status='created', created_at=None):
+    return create_notification(
+        template=letter_template,
+        to_field="file-name",
+        status=status,
+        reference="dvla-reference",
+        client_reference="file-name",
+        one_off=True,
+        created_by_id=service.users[0].id,
+        created_at=created_at
+    )
+
+
+def create_precompiled_template(service):
+    return create_template(
+        service,
+        template_type=LETTER_TYPE,
+        template_name='Pre-compiled PDF',
+        subject='Pre-compiled PDF',
+        content="",
+        hidden=True,
+        postage="second",
+    )
+
+
+def test_get_uploads_for_service(sample_template):
+    job = create_job(sample_template, processing_started=datetime.utcnow())
+    letter_template = create_precompiled_template(sample_template.service)
+    letter = create_uploaded_letter(letter_template, sample_template.service)
+
+    other_service = create_service(service_name="other service")
+    other_template = create_template(service=other_service)
+    other_job = create_job(other_template, processing_started=datetime.utcnow())
+    other_letter_template = create_precompiled_template(other_service)
+    other_letter = create_uploaded_letter(other_letter_template, other_service)
+
+    uploads_from_db = dao_get_uploads_by_service_id(job.service_id).items
+    other_uploads_from_db = dao_get_uploads_by_service_id(other_job.service_id).items
+
+    assert len(uploads_from_db) == 2
+
+    assert uploads_from_db[0] == (letter.id, letter.client_reference, 1, letter.created_at,
+                                  None, letter.created_at, letter.status, "letter")
+    assert uploads_from_db[1] == (job.id, job.original_file_name, job.notification_count, job.created_at,
+                                  job.scheduled_for, job.processing_started, job.job_status, "job")
+
+    assert len(other_uploads_from_db) == 2
+    assert other_uploads_from_db[0] == (other_letter.id,
+                                        other_letter.client_reference,
+                                        1,
+                                        other_letter.created_at,
+                                        None,
+                                        other_letter.created_at,
+                                        other_letter.status,
+                                        "letter")
+    assert other_uploads_from_db[1] == (other_job.id,
+                                        other_job.original_file_name,
+                                        other_job.notification_count,
+                                        other_job.created_at,
+                                        other_job.scheduled_for,
+                                        other_job.processing_started,
+                                        other_job.job_status, "job")
+
+    assert uploads_from_db[0] != other_uploads_from_db[0]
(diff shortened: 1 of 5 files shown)


tests/app/dao/test_uploads_dao.py
```diff
@@ -0,0 +1,168 @@
...
+        status=status,
+        reference="dvla-reference",
+        client_reference="file-name",
+        one_off=True,
+        created_by_id=service.users[0].id,
+        created_at=created_at
+    )
+
+
+def create_precompiled_template(service):
```
idavidmcdonald: Optional: We could make these two helper methods more consistent in their naming e.g.
`create_uploaded_letter` and `create_uploaded_letter_template`
or
`create_precompiled_letter` and `create_precompiled_template`

tests/app/dao/test_uploads_dao.py
```diff
@@ -0,0 +1,168 @@
...
+    assert len(results) == 2
+    assert results[0].id == upload_1.id
+    assert results[1].id == upload_2.id
+
+
+def test_get_uploads_orders_by_processing_started_and_created_at_desc(sample_template):
+    letter_template = create_precompiled_template(sample_template.service)
+
+    days_ago = datetime.utcnow() - timedelta(days=4)
+    upload_0 = create_uploaded_letter(letter_template, service=letter_template.service)
```
idavidmcdonald: Should bump the indexes of these

tests/app/upload/test_rest.py
```diff
@@ -0,0 +1,180 @@
+from datetime import datetime, timedelta, date
+
+from freezegun import freeze_time
+
+from app.models import LETTER_TYPE, JOB_STATUS_FINISHED, JOB_STATUS_PENDING
+from tests.app.db import create_job, create_notification, create_template, create_ft_notification_status
+from tests.conftest import set_config
+
+
+def create_uploaded_letter(letter_template, service, status='created', created_at=None):
```
idavidmcdonald: These two helper methods are duplicated in the DAO tests. Would be good to remove the duplication

tests/app/upload/test_rest.py
```diff
@@ -0,0 +1,180 @@
...
+        hidden=True,
+        postage="second",
+    )
+
+
+def test_get_uploads(admin_request, sample_template):
+    letter_template = create_precompiled_template(sample_template.service)
+
+    upload_1 = create_uploaded_letter(letter_template, sample_template.service, status='delivered',
+                                      created_at=datetime.utcnow() - timedelta(minutes=4))
```
idavidmcdonald: Optional: Technically we don't need the `timedelta` part on all of our timestamps in this test. They would still come out in the same order anyway.

tests/app/upload/test_rest.py
```diff
@@ -0,0 +1,180 @@
...
+    for _ in range(0, 3):
+        create_notification(template=sample_template, job=job_2, status='created')
+
+    job_3 = create_job(sample_template, processing_started=now)
+    for _ in range(0, 4):
+        create_notification(template=sample_template, job=job_3, status='sending')
+    letter_template = create_precompiled_template(sample_template.service)
+    letter_1 = create_uploaded_letter(letter_template, sample_template.service, status='delivered',
+                                      created_at=datetime.utcnow() - timedelta(days=3))
+    letter_2 = create_uploaded_letter(letter_template, sample_template.service, status='delivered',
```
idavidmcdonald: Optional: I don't think we are getting any benefit by having `letter_2` in this test. We are already testing a delivered letter by testing the statistics for `letter_1`. Also removing it might help make the set up of this test a bit more readable, a few more line breaks between groups of related lines might also help.

tests/app/upload/test_rest.py
```diff
@@ -0,0 +1,180 @@
...
+    assert resp_json[2]['id'] == str(job_2.id)
+    assert resp_json[2]['statistics'] == [{'status': 'created', 'count': 3}]
+    assert resp_json[3]['id'] == str(letter_2.id)
+    assert resp_json[3]['statistics'] == [{'status': 'delivered', 'count': 1}]
+    assert resp_json[4]['id'] == str(letter_1.id)
+    assert resp_json[4]['statistics'] == [{'status': 'delivered', 'count': 1}]
+
+
+def test_get_uploads_should_paginate(admin_request, sample_template):
+    for _ in range(0, 10):
```
idavidmcdonald: Optional: You have a couple of `range` statements here that take `0` as a starting argument which is not needed as this will be the case by default i.e. `range(0,10) == range(10)`

tests/app/dao/test_uploads_dao.py
```diff
@@ -0,0 +1,168 @@
...
+    assert results[0].id == upload_0.id
+    assert results[1].id == upload_1.id
+    assert results[2].id == upload_2.id
+    assert results[3].id == upload_3.id
+
+
+def test_get_uploads_is_paginated(sample_template):
+    letter_template = create_precompiled_template(sample_template.service)
+
+    upload_1 = create_uploaded_letter(letter_template, sample_template.service, status='delivered',
```
idavidmcdonald: Optional: Your set up could be simpler here, we don't need many different types of uploads. Your test set up for `test_get_uploads_should_paginate` is much simpler and you could do similar here if you wanted

tests/app/upload/test_rest.py
```diff
@@ -0,0 +1,180 @@
...
+    # it's the 10th today, so 3 days should include all of 7th, 8th, 9th, and some of 10th.
+    just_three_days_ago = datetime(2017, 6, 6, 22, 59, 59)
+    not_quite_three_days_ago = just_three_days_ago + timedelta(seconds=1)
+
+    job_1 = create_job(sample_template, created_at=just_three_days_ago, processing_started=just_three_days_ago)
+    job_2 = create_job(sample_template, created_at=just_three_days_ago, processing_started=not_quite_three_days_ago)
+    # is old but hasn't started yet (probably a scheduled job). We don't have any stats for this job yet.
+    job_3 = create_job(sample_template, created_at=just_three_days_ago, processing_started=None)
+
+    # some notifications created more than three days ago, some created after the midnight cutoff
```
idavidmcdonald: Good use of comments in this test

PR #2671, opened 2019-12-05
process letters for 10 days when updating ft_notification_status

leohemsted:
sms and emails have a very predictable 72 hour lifecycle. letters, on the other hand, have ridiculously complex lifecycles - they might not get sent because it's a weekend, they might not get sent because they're second class and are only processed on alternate days, they might not get sent because a different letter in the same batch had an error that we didn't know about. Either way, it's apparent that four days is definitely not enough time to guarantee that letters have gone from sending to delivered.

Extend the amount of days we process for letters to 10 days. Keep emails and sms down at 4 to keep run-times shorter

We're deliberately not thinking about returned letters here at all.

- [x] https://github.com/alphagov/notifications-api/pull/2669 (this is branched off that PR)

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -68,15 +68,23 @@ def create_nightly_billing_for_day(process_day):
 @cronitor("create-nightly-notification-status")
 @statsd(namespace="tasks")
 def create_nightly_notification_status():
-    day_start = convert_utc_to_bst(datetime.utcnow()).date() - timedelta(days=1)
+    yesterday = convert_utc_to_bst(datetime.utcnow()).date() - timedelta(days=1)
 
-    for i in range(0, 4):
-        process_day = day_start - timedelta(days=i)
-        for notification_type in [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]:
+    # email and sms
+    for i in range(4):
+        process_day = yesterday - timedelta(days=i)
+        for notification_type in [SMS_TYPE, EMAIL_TYPE]:
             create_nightly_notification_status_for_day.apply_async(
                 kwargs={'process_day': process_day.isoformat(), 'notification_type': notification_type},
                 queue=QueueNames.REPORTING
             )
+    # letters get modified for a longer time period than sms and email, so we need to reprocess for more days
+    for i in range(10):
+        process_day = yesterday - timedelta(days=i)
+        create_nightly_notification_status_for_day.apply_async(
+            kwargs={'process_day': process_day.isoformat(), 'notification_type': LETTER_TYPE},
+            queue=QueueNames.REPORTING
+        )
 
 
 @notify_celery.task(name="create-nightly-notification-status-for-day")
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -67,14 +67,9 @@ def create_nightly_billing_for_day(process_day):
 @notify_celery.task(name="create-nightly-notification-status")
 @cronitor("create-nightly-notification-status")
 @statsd(namespace="tasks")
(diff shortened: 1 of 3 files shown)


app/celery/reporting_tasks.py
```diff
@@ -67,21 +67,23 @@ def create_nightly_billing_for_day(process_day):
...
-def create_nightly_notification_status(day_start=None):
-    # day_start is a datetime.date() object. e.g.
-    # 4 days of data counting back from day_start is consolidated
-    if day_start is None:
-        day_start = convert_utc_to_bst(datetime.utcnow()).date() - timedelta(days=1)
-    else:
-        # When calling the task its a string in the format of "YYYY-MM-DD"
-        day_start = datetime.strptime(day_start, "%Y-%m-%d").date()
+def create_nightly_notification_status():
+    most_recent_day_to_process = convert_utc_to_bst(datetime.utcnow()).date() - timedelta(days=1)
```
idavidmcdonald: Is `most_recent_day_to_process` equivalent to yesterday? Or is it more nuanced than that?

app/celery/reporting_tasks.py
```diff
@@ -67,21 +67,23 @@ def create_nightly_billing_for_day(process_day):
...
+
     for i in range(0, 4):
-        process_day = day_start - timedelta(days=i)
+        process_day = most_recent_day_to_process - timedelta(days=i)
         for notification_type in [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]:
```
idavidmcdonald: Very optional: You could exclude letters from this loop and do `range(10)` on the second loop instead. Means behaviour for letters isn't spread across two loops

app/commands.py
```diff
@@ -505,7 +505,7 @@ def migrate_data_to_ft_notification_status(start_date, end_date, notification_ty
 
     start_date = start_date.date()
     end_date = end_date.date()
-    for day_diff in range((end_date - start_date).days):
+    for day_diff in range((end_date - start_date).days + 1):
```
idavidmcdonald: Is there a test you can add/change to confirm this change to an inclusive end date has worked?

app/celery/reporting_tasks.py
leohemsted: no, you're right, it should always be yesterday now that i've got rid of the `day_start` argument. i'll change that

app/commands.py
leohemsted: sadly not :(

app/celery/reporting_tasks.py
leohemsted: yeah i think that's a marginally nicer way of laying out the function 👍. squashed in

tests/app/celery/test_reporting_tasks.py
```diff
@@ -50,21 +50,31 @@ def test_create_nightly_billing_triggers_tasks_for_days(notify_api, mocker, day_
...
-    ('2019-07-21', ['2019-07-21', '2019-07-20', '2019-07-19', '2019-07-18']),
-])
-def test_create_nightly_notification_status_triggers_tasks_for_days(notify_api, mocker, day_start, expected_kwargs):
+def test_create_nightly_notification_status_triggers_tasks_for_days(notify_api, mocker):
     mock_celery = mocker.patch('app.celery.reporting_tasks.create_nightly_notification_status_for_day')
-    create_nightly_notification_status(day_start)
+    create_nightly_notification_status()
+
+    assert mock_celery.apply_async.call_count == (
+        (4 * 3)  # four days, three notification types
```
idavidmcdonald: Optional: Could change your logic to be `4 * 2 + 10` but not fussed if you don't

PR #2673, opened 2019-12-06
Bill `NOTIFICATION_PENDING` notifications

idavidmcdonald:
## Example of bug in the wild
Screenshots included as the bug may fix itself if the messages transition out of a pending state

54 messages sent, of which they are all 2 fragments, meaning 108 expected text messages to be billed for. The billing dashboard is short by 4 fragments, ie two messages.

https://www.notifications.service.gov.uk/services/04cca2c6-3352-4da9-8c64-8782d7e47bf6/notifications/sms?status=sending
image

https://www.notifications.service.gov.uk/services/04cca2c6-3352-4da9-8c64-8782d7e47bf6/usage
image

If we look in the DB, we find there are the two sending messages with the status `pending`. These are the ones that are not being billed.

## Code change

SMS and emails may be marked as `NOTIFICATION_PENDING`. These will be
billed as they will have been sent to the provider and will eventually
turn to a final state such as `NOTIFICATION_DELIVERED` or
`NOTIFICATION_PERMANENT_FAILURE`.

This change will fix a discrepency on the billing page were the number
of messages being billed was less than the number of messages reported
as sent on a services dashboard when some of those messages were in a
pending state.

In reality, I don't think this bug would have had any longer affects for
incorrect billing as messages would not stay in the pending state for
too long and billing calculations would happen after that point.

## Questions
- Can an email be in a `pending` state? I have made the assumption so but would like to double check
- Can a letter be in a `pending` state and be billable? I have made the assumption not but would like to double check


tests/app/dao/test_ft_billing_dao.py
```diff
@@ -59,19 +59,20 @@ def set_up_yearly_data():
...
+
     for status in ['created', 'technical-failure']:
         create_notification(template=template, status=status)
-
-    today = convert_utc_to_bst(datetime.utcnow())
     results = fetch_billing_data_for_day(today.date())
     assert results == []
-    for status in ['delivered', 'sending', 'temporary-failure']:
+
+    for status in ['delivered', 'sending', 'temporary-failure', 'pending']:
```
idavidmcdonald: This check seems to be testing the same behaviour (or at least less of it) than `test_fetch_billing_data_for_today_includes_data_with_the_right_key_type`? If so, we could probably remove this section from this test altogether?

idavidmcdonald: In fact, is this whole test adding any more value over test_fetch_billing_data_for_today_includes_data_with_the_right_key_type? The negative cases for statuses that are not billable are tested implicitly there too

leohemsted:
`NOTIFICATION_PENDING` specifically only applies to firetext, as it's a status they use to distinguish between permanent and temporary failure. If we send a message and then receive a callback of "declined" we assume that's `permanent-failure`.

If they however send a callback of "pending", that means they've handed off to the networks, in practice this means they've confirmed the number is real, ie not a permanent-failure. If they then return a "declined" status update, that means it must be a `temporary-failure`.

If they're still in pending after 72 hours we time them out (set them to `temporary-failure`) so historical data won't have been affected by this.


TL;DR: Neither email or letter can ever be in `pending`. It is true that we have a line to check for pending in [ses callbacks](https://github.com/alphagov/notifications-api/blob/master/app/celery/process_ses_receipts_tasks.py#L53) and in a letters + emails function in [notifications_dao](https://github.com/alphagov/notifications-api/blob/master/app/dao/notifications_dao.py#L150), however, the only place where we actually set something to pending is through the [firetext response map](https://github.com/alphagov/notifications-api/blob/master/app/clients/sms/firetext.py#L11-L20)

tests/app/dao/test_ft_billing_dao.py
leohemsted: i think you mean `test_fetch_billing_data_for_day_bills_correctly_for_status` here? in which case, i agree and i think this test can be deleted.

though that said, i have some misgivings around both of these tests in general as if it fails it's not apparent _which_ notification status types is behaving unexpectedly

app/models.py
```diff
@@ -1297,6 +1297,7 @@ def check_code(self, cde):
     NOTIFICATION_SENDING,
     NOTIFICATION_SENT,
     NOTIFICATION_DELIVERED,
+    NOTIFICATION_PENDING,
```
idavidmcdonald: @leohemsted Based on your context, you are still fine with this being added to `NOTIFICATION_STATUS_TYPES_BILLABLE`, rather than us creating a new `NOTIFICATION_STATUS_TYPES_BILLABLE_FOR_SMS`?

tests/app/dao/test_ft_billing_dao.py
idavidmcdonald: I did, thanks.

And yes. Not only that, if you made a mistake to include one status you didn't mean to, whilst also excluding a different status you didn't mean to, the test would pass as it just tests the number. I'll have a think if there is an easy way to test this more thoroughly

app/models.py
servingUpAces: I'm ok with that. It's not stated anywhere in the code, but we do only use it in the Firetext callback scenario.

tests/app/dao/test_ft_billing_dao.py
idavidmcdonald: @leohemsted so I've removed this test. I couldn't think of a way to easily improve the other one though. Open to suggestions but if not then I think this may be ready to go.

PR #2675, opened 2019-12-10
Add new tasks for processing letters which have passed virus scan

klssmith:
The current `process_virus_scan_passed` task will be split into three separate tasks so that we can validate letters asynchronously. These new tasks are, in order:

#### 1. `sanitise_letter`
Added in this PR. This calls the template preview task with details of the letter to be sanitised.
#### 2. `sanitise_and_upload_letter`
Added to template preview [here](https://github.com/alphagov/notifications-template-preview/pull/399). This sanitises the letter and calls the third task with the results.
#### 3. `process_sanitised_letter`
Added in this PR. This moves the files into the relevant S3 buckets and update the notification

Once deployed, the new tasks won't be used until notifications-antivirus is updated to call `sanitise_letter` instead of `process_virus_scan_passed`.

[Pivotal story](https://www.pivotaltracker.com/story/show/169148522)

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -27,6 +27,7 @@ from app.dao.notifications_dao import (
     dao_update_notifications_by_reference,
 )
 from app.errors import VirusScanError
+from app.exceptions import NotificationTechnicalFailureException
 from app.letters.utils import (
     copy_redaction_failed_pdf,
     get_billable_units_for_letter_page_count,
@@ -35,6 +36,7 @@ from app.letters.utils import (
     upload_letter_pdf,
     ScanErrorType,
     move_failed_pdf,
+    move_sanitised_letter_to_test_or_live_pdf_bucket,
     move_scan_to_invalid_pdf_bucket,
     move_error_pdf_to_scan_bucket,
     get_file_names_from_error_bucket,
@@ -43,6 +45,7 @@ from app.models import (
     KEY_TYPE_TEST,
     NOTIFICATION_CREATED,
     NOTIFICATION_DELIVERED,
+    NOTIFICATION_PENDING_VIRUS_CHECK,
     NOTIFICATION_TECHNICAL_FAILURE,
     NOTIFICATION_VALIDATION_FAILED,
     NOTIFICATION_VIRUS_SCAN_FAILED,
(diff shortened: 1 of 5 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -262,6 +265,100 @@ def process_virus_scan_passed(self, filename):
...
+@notify_celery.task(bind=True, name='sanitise-letter', max_retries=15, default_retry_delay=300)
+def sanitise_letter(self, filename):
+    try:
+        reference = get_reference_from_filename(filename)
+        notification = dao_get_notification_by_reference(reference)
+
+        current_app.logger.info('Notification ID {} Virus scan passed: {}'.format(notification.id, filename))
+
+        if notification.status != NOTIFICATION_PENDING_VIRUS_CHECK:
+            current_app.logger.exception('Sanitise letter called for notification {} which has is in {} state'.format(
```
servingUpAces: Why are we logging an exception here? Isn't it ok if the notification has already been marked as `created`?

app/celery/letters_pdf_tasks.py
```diff
@@ -262,6 +265,100 @@ def process_virus_scan_passed(self, filename):
...
+        try:
+            current_app.logger.exception(
+                "RETRY: calling sanitise_letter task for notification {} failed".format(notification.id)
+            )
+            self.retry(queue=QueueNames.RETRY)
+        except self.MaxRetriesExceededError:
+            message = "RETRY FAILED: Max retries reached. " \
+                      "The task sanitise_letter failed for notification {}. " \
+                      "Notification has been updated to technical-failure".format(notification.id)
+            update_notification_status_by_id(notification.id, NOTIFICATION_TECHNICAL_FAILURE)
```
servingUpAces: This is fine - except when the template preview app throws an exception but it's actually a file that we can not validate. Until we fix template-preview so that it doesn't throw exceptions we will set letters to tech-failure when it's actually invalid.

app/celery/letters_pdf_tasks.py
```diff
@@ -262,6 +265,100 @@ def process_virus_scan_passed(self, filename):
...
+    invalid_pages,
+    validation_status,
+    filename,
+    notification_id,
+):
+    current_app.logger.info('Processing sanitised letter with id {}'.format(notification_id))
+    notification = get_notification_by_id(notification_id, _raise=True)
+
+    if notification.status != NOTIFICATION_PENDING_VIRUS_CHECK:
+        current_app.logger.exception(
```
servingUpAces: Is this an exception if the letter has already been marked as `created`. If for some reason this task is called twice.

app/config.py
```diff
@@ -29,6 +29,7 @@ class QueueNames(object):
     CALLBACKS = 'service-callbacks'
     LETTERS = 'letter-tasks'
     ANTIVIRUS = 'antivirus-tasks'
+    TEMPLATE_PREVIEW = 'template-preview-tasks'
```
servingUpAces: Not a big deal - but should this be called `validate-letter-tasks`

app/celery/letters_pdf_tasks.py
klssmith: If the notification is already marked as `created` then the task has already run - I had assumed that that was an issue and that we'd want to know about it, but if that's not the case then I can change it to a different level of log message

app/config.py
klssmith: I went for `template-preview-tasks` to match `antivirus-tasks` above it, but I'll change it

app/celery/letters_pdf_tasks.py
klssmith: We catch all exceptions from template preview and mark the notification as validation-failed, so this shouldn't be an issue. We were already doing that in an error handler, and in my template preview PR I've kept that behaviour but moved it to the function that is sanitising the letter - https://github.com/alphagov/notifications-template-preview/pull/399/files#diff-cc3699ca803f977365cdb0525d419092R123.

app/celery/letters_pdf_tasks.py
klssmith: I've downgraded this log message and other similar one from `exception` to `info`

PR #2676, opened 2019-12-10
Add returned letters table

servingUpAces:
The services do not have a way to get returned letters from the app. 

Create a table to store the notification_id, reported_at date and service_id for each returned letters. Adding the notification_id to this table allows us to join to notification_history by the primary key which will mean the query should be fast. 

When the returned letters are uploaded via the UI, the task will also update the returned_letters table. 
There is an endpoint to return the summary data, a count of letter per report date, typically there is one per month. 
Pending is another endpoint to return the data required by the service as described in https://www.pivotaltracker.com/story/show/167920694

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2018,8 +2018,7 @@ class Complaint(db.Model):
     __tablename__ = 'complaints'
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
-    notification_id = db.Column(UUID(as_uuid=True), db.ForeignKey('notification_history.id'),
-                                index=True, nullable=False)
+    notification_id = db.Column(UUID(as_uuid=True), index=True, nullable=False)
     service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), unique=False, index=True, nullable=False)
     service = db.relationship(Service, backref=db.backref('complaints'))
     ses_feedback_id = db.Column(db.Text, nullable=True)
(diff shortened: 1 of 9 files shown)


tests/app/dao/test_returned_letters_dao.py
```diff
@@ -0,0 +1,114 @@
...
+        insert_or_update_returned_letters(['ref1', 'ref2'])
+
+        returned_letters = ReturnedLetter.query.all()
+        assert len(returned_letters) == 2
+        for x in returned_letters:
+            assert x.reported_at == datetime(2019, 12, 10, 14, 20)
+            assert x.notification_id in [notification.id, history.id]
+
+
+def test_insert_or_update_returned_letters_when_no_notification(sample_letter_template):
```
klssmith: This test could use the `notify_db_session` fixture instead of `sample_letter_template`, which isn't used

tests/app/service/test_rest.py
```diff
@@ -3372,3 +3372,12 @@ def test_get_monthly_notification_data_by_service(mocker, admin_request):
 
     dao_mock.assert_called_once_with(start_date, end_date)
     assert response == []
+
+
+def test_get_returned_letter_summary(admin_request, sample_service):
```
klssmith: Might be worth testing the JSON that is actually returned from the endpoint as well as how many items are in it.

tests/app/dao/test_returned_letters_dao.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def test_get_returned_letter_summary_orders_by_reported_at(sample_service):
+    now = datetime.utcnow()
+    last_month = datetime.utcnow() - timedelta(days=30)
+    create_returned_letter(sample_service, reported_at=now)
+    create_returned_letter(sample_service, reported_at=now)
+    create_returned_letter(sample_service, reported_at=now)
+    create_returned_letter(sample_service, reported_at=last_month)
+    create_returned_letter(sample_service, reported_at=last_month)
```
klssmith: You could create a returned letter from a different service here to make sure that only letters from the sample_service get returned

app/models.py
```diff
@@ -2071,3 +2070,13 @@ def serialize(self):
             "created_at": self.created_at.strftime(DATETIME_FORMAT),
             "updated_at": self.updated_at.strftime(DATETIME_FORMAT) if self.updated_at else None,
         }
+
+
+class ReturnedLetter(db.Model):
+    __tablename__ = 'returned_letters'
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    reported_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
```
klssmith: Is there any chance that we will populate the table twice on the same day? If there is, then having this column as a date, not a datetime, would make things easier because we are only interested in the daily results, so we wouldn't have to then aggregate the results by day.

PR #2678, opened 2019-12-11
Use .format instead of concatenation to avoid type issues

CrystalPea:
Trying to concatenate uuid onto a string was throwing an error.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -271,14 +271,20 @@ def check_for_services_with_high_failure_rates_or_sending_to_tv_numbers():
             len(services_with_failures)
         )
         for service in services_with_failures:
-            service_dashboard = current_app.config['ADMIN_BASE_URL'] + "/services/" + service.service_id
+            service_dashboard = "{}/services/{}".format(
+                current_app.config['ADMIN_BASE_URL'],
+                str(service.service_id),
+            )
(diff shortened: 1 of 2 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -271,14 +271,14 @@ def check_for_services_with_high_failure_rates_or_sending_to_tv_numbers():
             len(services_with_failures)
         )
         for service in services_with_failures:
-            service_dashboard = current_app.config['ADMIN_BASE_URL'] + "/services/" + service.service_id
+            service_dashboard = current_app.config['ADMIN_BASE_URL'] + "/services/" + str(service.service_id)
```
quis: I think this is one reason why we prefer `str.format` over string concatenation, ie this works without doing any type casting:
```suggestion
            service_dashboard = "{}/services/{}".format(
                current_app.config['ADMIN_BASE_URL'],
                str(service.service_id),
            )
```

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -519,16 +519,16 @@ def test_check_for_missing_rows_in_completed_jobs_uses_sender_id(mocker, sample_
         [MockServicesWithHighFailureRate("123", 0.3)],
```
idavidmcdonald: I think the reason this error wasn't caught in your unit tests is because your tests don't match production data close enough. Here I think we should be setting the first argument of `MockServicesWithHighFailureRate` with a UUID as opposed to a string

PR #2680, opened 2019-12-11
Catch `TokenAlgorithmError`

idavidmcdonald:
Instead of letting it go uncaught and causing an error, we now show the
user an appropriate error message.

Requires https://github.com/alphagov/notifications-python-client/pull/155


CrystalPea:
I think you need to run `make freeze-requirements` :)

PR #2681, opened 2019-12-12
tend providers back 

leohemsted:
every five minutes, check the providers. If they haven't been updated in the last hour, then push them by 10 priority back towards a 60/40 mmg/firetext split.


tests/app/dao/test_provider_details_dao.py
```diff
@@ -122,6 +125,71 @@ def test_get_alternative_sms_provider_fails_if_unrecognised():
...
+    ).all()
+
+    assert len(updated_provider_history_rows) - len(old_provider_history_rows) == 1
+    assert updated_provider_history_rows[0].version - old_provider_history_rows[0].version == 1
+    assert updated_provider_history_rows[0].priority == 50
+
+
+@freeze_time('2016-01-01 00:30')
+def test_get_sms_providers_for_update_returns_providers(restore_provider_details):
+    six_minutes_ago = datetime(2016, 1, 1, 0, 24)
```
idavidmcdonald: Optional: The choice of five minutes for the treshold is fine but wondering if instead using one hour would make this test more reflective of the rest of the code and reduces a bit of the time needed to figure out what this test is testing. Could also change test name to something like `test_get_sms_providers_for_update_returns_providers_when_not_edited_within_threshold` or similar

tests/app/dao/test_provider_details_dao.py
```diff
@@ -136,14 +204,14 @@ def test_get_alternative_sms_provider_fails_if_unrecognised():
     ({'mmg': 50, 'firetext': -100}, {'mmg': 40, 'firetext': -90}),
 ])
 def test_reduce_sms_provider_priority_switches_provider(
```
idavidmcdonald: Optional: It it worth changing this test name now? Seems to more reflect older behaviour

tests/app/dao/test_provider_details_dao.py
```diff
@@ -155,56 +223,81 @@ def test_reduce_sms_provider_priority_switches_provider(
...
-    ).all()
 
-    assert len(updated_provider_history_rows) - len(provider_history_rows) == 1
-    assert updated_provider_history_rows[0].version - provider_history_rows[0].version == 1
-    assert updated_provider_history_rows[0].priority == 90
+@pytest.mark.parametrize('existing_mmg, existing_firetext, new_mmg, new_firetext', [
+    (50, 50, 60, 40),  # not just 50/50 - 60/40 specifically
+    (65, 35, 60, 40),  # doesn't overshoot if there's less than 10 difference
+    (0, 100, 10, 90),  # only adjusts by 10
+    (100, 100, 90, 90),  # it tries to fix weird data
```
idavidmcdonald: Optional: Bit unclear what this test case is trying to test as the result new priorities haven't changed from the existing ones

app/dao/provider_details_dao.py
```diff
@@ -34,57 +33,92 @@ def dao_get_provider_versions(provider_id):
...
     q = ProviderDetails.query.filter(
         ProviderDetails.notification_type == 'sms',
         ProviderDetails.active
     ).with_for_update().all()
 
-    providers = {provider.identifier: provider for provider in q}
-    other_identifier = get_alternative_sms_provider(identifier)
-
     # if something updated recently, don't update again. If the updated_at is null, treat it as min time
     if any((provider.updated_at or datetime.min) > datetime.utcnow() - time_threshold for provider in q):
```
idavidmcdonald: Optional: I feel that `_get_sms_providers_for_update` is doing too much. I think it could just return the provider it has found and then a different method is responsible for returning a boolean of if providers have been updated too recently.

tests/app/dao/test_provider_details_dao.py
leohemsted: i've tried to clarify the comment a little bit - i wanted to prove that the values are independent and it doesn't just bump one and lower the other all the time

app/dao/provider_details_dao.py
```diff
@@ -34,57 +33,92 @@ def dao_get_provider_versions(provider_id):
     ).all()
 
 
-@transactional
-def dao_reduce_sms_provider_priority(identifier, *, time_threshold):
-    """
-    Will reduce a chosen sms provider's priority, and increase the other provider's priority by 10 points each.
-    If either provider has been updated in the last `time_threshold`, then it won't take any action.
+def _adjust_provider_priority(provider, new_priority):
+    print('Adjusting provider priority - {} going from {} to {}'.format(
```
servingUpAces: Do you want to make this a log statement or get rid?

leohemsted: whoops good spot, that should be a logger statement - i'd just changed it to help test

PR #2682, opened 2019-12-12
Allow searching notifications by reference as well as recipient

quis:
We have a team who want to find emails that might have been sent to an incorrect address. Therefore they can’t search by the correct address, because it won’t match.

What they do have is the reference number of the user’s application, which is also stored in the `client_reference` field on the notification.

So when a user is searching we should also look at the client reference, as well as the recipient, allowing the user to enter either in the search box.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -17,7 +17,7 @@ from notifications_utils.recipients import (
 )
 from notifications_utils.statsd_decorators import statsd
 from notifications_utils.timezones import convert_bst_to_utc, convert_utc_to_bst
-from sqlalchemy import (desc, func, asc, and_)
+from sqlalchemy import (desc, func, asc, and_, or_)
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.sql import functions
@@ -569,7 +569,7 @@ def dao_update_notifications_by_reference(references, update_dict):
 
 
 @statsd(namespace="dao")
-def dao_get_notifications_by_to_field(service_id, search_term, notification_type=None, statuses=None):
+def dao_get_notifications_by_recipient_or_reference(service_id, search_term, notification_type=None, statuses=None):
     if notification_type is None:
         notification_type = guess_notification_type(search_term)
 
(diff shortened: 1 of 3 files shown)


app/dao/notifications_dao.py
```diff
@@ -588,13 +584,17 @@ def dao_get_notifications_by_to_field(service_id, search_term, notification_type
             normalised = search_term.lower()
 
     else:
-        raise InvalidRequest("Only email and SMS can use search by recipient", 400)
+        raise InvalidRequest("Must specify email or sms to search by recipient", 400)
```
CrystalPea: The error message sounds vague. I think:
"Searching by recipient is only available for sms and email messages"

Would work better.

Also earlier notification_type defaulted to None, now it does not have a default value. Is it possible this change of behaviour would break something?

app/dao/notifications_dao.py
```diff
@@ -569,10 +568,7 @@ def dao_update_notifications_by_reference(references, update_dict):
 
 
 @statsd(namespace="dao")
-def dao_get_notifications_by_to_field(service_id, search_term, notification_type=None, statuses=None):
-    if notification_type is None:
-        notification_type = guess_notification_type(search_term)
-
+def dao_get_notifications_by_to_field(service_id, search_term, notification_type, statuses=None):
```
CrystalPea: should we change function name to: `dao_get_notifications_by_recipient_or_reference` or something like that to reflect that the function does more now?

tests/app/dao/notification_dao/test_notification_dao.py
```diff
@@ -1131,6 +1131,49 @@ def test_dao_get_notifications_by_to_field_escapes(
...
+    ('%', 2),
+    ('_', 1),
+    ('/', 1),
+    ('\\', 1),
+    ('baz\\baz', 1),
+    ('%foo', 0),
+    ('%_%', 0),
+    ('test@example.com', 5),
+])
+def test_dao_get_notifications_by_to_field_escapes_reference(
```
CrystalPea: what does it mean that it *escapes* reference?

quis: `%` and `_` are wildcard characters in SQL, but we want them to be interpreted literally. There’s a bit more detail in the original search by recipient implementation: bdd77f9150fde59d71fdc837ff4cd2cbf0d62896

CrystalPea: ahh, I see, thanks 💖 
Maybe we could change test name to `test_dao_get_notifications_by_to_field_escapes_special_chars_for_reference` ?
What do you think :)

CrystalPea:
Good to go 👍 
Only I think your code style check is failing

app/dao/notifications_dao.py
quis: Changed the behaviour to accept `None` for notification type because users without the _See dashboard_ permission can still search across emails and text messages (they see both on one screen).

PR #2687, opened 2019-12-19
Add and write into Inbound sms history

CrystalPea:
We do not have historical data about inbound sms that our services receive. 
This can be a problem, because sometimes we need statistics about inbound sms to learn if there were some issues with delivery etc.

In this PR:
- create `InboundSmsHistory` model
- create `inbound_sms_history` table in the database through migration
- insert into the `InboundSmsHistory` table when deleting inbound sms at the end of retention period.

Story ticket: https://www.pivotaltracker.com/story/show/169809179


app/dao/inbound_sms_dao.py
```diff
@@ -62,6 +63,35 @@ def dao_count_inbound_sms_for_service(service_id, limit_days):
     ).count()
 
 
+def _insert_update_notification_history(subquery, query_limit=10000):
```
idavidmcdonald: Will look at this function after lunch

CrystalPea: Cool, if it makes it easier, I based that function on an existing function doing similar thing for Notifications: https://github.com/alphagov/notifications-api/blob/master/app/dao/notifications_dao.py#L373

app/models.py
```diff
@@ -1898,6 +1898,18 @@ def serialize(self):
         }
 
 
+class InboundSmsHistory(db.Model, HistoryModel):
+    __tablename__ = 'inbound_sms_history'
+    id = db.Column(UUID(as_uuid=True), primary_key=True)
+    created_at = db.Column(db.DateTime, nullable=False)
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), index=True, unique=False)
+    service = db.relationship('Service', backref='inbound_sms_history')  # what does this one do?
```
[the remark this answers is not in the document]
CrystalPea: yeah, backref and relationship... what does this line do in general? Do we need it? I copied it over from `InboundSms` table and put in the comment to remember to ask someone, so it worked well hahah :D

tests/app/dao/test_inbound_sms_dao.py
```diff
@@ -90,7 +93,7 @@ def test_count_inbound_sms_for_service_filters_messages_older_than_n_days(sample
 
 
 @freeze_time("2017-06-08 12:00:00")
-def test_should_delete_inbound_sms_according_to_data_retention(notify_db_session):
+def test_should_delete_inbound_sms_according_to_data_retention(notify_db_session, mocker):
```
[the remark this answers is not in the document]
CrystalPea: No it does not, the mocker is vestigial, thanks for spotting 👍

tests/app/dao/test_inbound_sms_dao.py
```diff
@@ -115,6 +117,9 @@ def test_should_delete_inbound_sms_according_to_data_retention(notify_db_session
 
     deleted_count = delete_inbound_sms_older_than_retention()
 
+    history = InboundSmsHistory.query.all()
+    assert len(history) == 7
```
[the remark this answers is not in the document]
CrystalPea: I added checking the keys for now. I can also do another test where I would be checking if data is right.

app/dao/inbound_sms_dao.py
```diff
@@ -62,6 +63,35 @@ def dao_count_inbound_sms_for_service(service_id, limit_days):
...
+    while offset < inbound_sms_count:
+        statement = insert(InboundSmsHistory).from_select(
+            InboundSmsHistory.__table__.c,
+            inbound_sms_query.limit(query_limit).offset(offset)
+        )
+
+        statement = statement.on_conflict_do_update(
+            constraint="inbound_sms_history_pkey",
+            set_={
+                "created_at": statement.excluded.created_at,
```
idavidmcdonald: hmm wondering if we should be setting the created_at as when the original inbound sms was created or if instead we should be saving that information on a different field (does `provider_date` cover it?) and letting `created_at` refer to the creation of the `inbound_sms_history` record.

tests/app/dao/test_inbound_sms_dao.py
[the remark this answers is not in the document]
idavidmcdonald: I would indeed be keen for a test of the data being right please

app/models.py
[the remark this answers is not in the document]
idavidmcdonald: https://docs.sqlalchemy.org/en/13/orm/backref.html

Have a good slow read of that (took me a while to get my head around it). 

Essentially, it means that not only will `inboundSmsHistory.service` return the Service object (which would happen if we didn't include the backref) but now also if you have a service then `service.inbound_sms_history` would return a list of all the inbound sms history objects for the service.

I guess the first question is do we want every service to have `inbound_sms_history` available as a property on it. If we never use that relationship in our code then we might not need it. If we did think that we need it, then one thing we would want to look into is the `lazy` argument as we probably don't want every time we get a service from the DB to also have to go and get all it's historic inbound sms's as this is quite DB intensive and we may never use this data.

My initial thoughts are that we therefore don't want to keep the backref as we aren't going to use it anywhere. This also matches the relationship we have for notification history and a service as per https://github.com/alphagov/notifications-api/blob/master/app/models.py#L1646

app/dao/inbound_sms_dao.py
idavidmcdonald: Thanks, that was useful to see

app/dao/inbound_sms_dao.py
CrystalPea: Provider date is when we received the inbound sms from provider I think, and created at is when we created an object for it. 

So probably good call that in history table, created_at would be when we create a history object.

app/models.py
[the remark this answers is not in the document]
CrystalPea: Ahh, thanks a lot, that's much clearer now :)

I updated the Model and migration file ☺️

app/dao/inbound_sms_dao.py
leohemsted: Hmm, I think your original method was better - my reckon is that we're more likely to want to check an inbound sms by when it was originally created at. The nightly task is an implementation detail and having the created_at change meaning between the two tables feels like a big gotcha. It might be useful to know when the inbound sms was received if we need to align it with log messages, service api calls, etc.

in outbound-land, notification_history.created_at is a copy of notification.created_at. I think we should maintain that behaviour. You can verify that behaviour by running `select max(created_at) from notification_history;`. For example, despite the task having most recently run last night, the newest created_at time on the preview notification_history table is from the 14th.

CrystalPea: Ok, going with @leohemsted 's suggestion, it is also much easier to leave it that way 😁

app/dao/inbound_sms_dao.py
```diff
@@ -62,6 +63,34 @@ def dao_count_inbound_sms_for_service(service_id, limit_days):
...
+        statement = statement.on_conflict_do_update(
+            constraint="inbound_sms_history_pkey",
+            set_={
+                "service_id": statement.excluded.service_id,
+                "notify_number": statement.excluded.notify_number,
+                "provider_date": statement.excluded.provider_date,
+                "provider_reference": statement.excluded.provider_reference,
+                "provider": statement.excluded.provider
+            }
+        )
```
idavidmcdonald: @leohemsted 

I'm wondering whether these lines are adding value. They aren't being tested (I've removed them from our code and no tests fail).

Further, I'm trying to work out in which scenario we expect them to run. A conflict seems unlikely but I guess if the delete statement that happened after this failed, and then this function gets run a second time it will raise conflicts. If that happens, do we want to treat this like an UPSERT, or should we just do nothing on the conflict. My assumption is that the inbound SMS row wouldn't get updated after creation (there is no updated_at field) so there would never be any information to update, therefore making most of this redundant and we should instead do an `on_conflict_do_nothing`?

idavidmcdonald: Oh this corresponding commit may be of slight interest to see where this code has been taken from https://github.com/alphagov/notifications-api/commit/0def0b7fd089aaa62bcc3892ddab25771e49182b

leohemsted: hmmm i think i'm fine with on_conflict_do_nothing. @servingUpAces?

servingUpAces: It's a similar pattern to the one we use for NotificationHistory. You are correct that unlike Notifications InboundSms don't have a lifecycle to the update will not change anything (except the audit columns).  Your suggestion of using `on_conflict_do_nothing` is appropriate. We need to be able to handle the insert happening twice, in case something goes wrong during the task and it's re-run. But there should be a unit test for this.

PR #2692, opened 2020-01-06
Put address in to field for precompiled

servingUpAces:
# Context
When a precompiled letter is sent via the admin app, we now pass in the address which can be set in the Notifications.to field.

Once a precompiled letters sent by the API has passed validation we can set the address in Notifications.to field. The template-preview app returns the address is the http request to sanitise the letter. 

# What
Save recipient address in the "to" field of a notification

# Coming soon
The celery tasks to validate precompiled letters sent by the API will be done in another PR.

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -309,7 +309,6 @@ def process_sanitised_letter(
     validation_status,
     filename,
     notification_id,
-    recipient_address=None
 ):
     current_app.logger.info('Processing sanitised letter with id {}'.format(notification_id))
     notification = get_notification_by_id(notification_id, _raise=True)
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -255,7 +255,8 @@ def process_virus_scan_passed(self, filename):
         update_letter_pdf_status(
             reference=reference,
             status=NOTIFICATION_DELIVERED if is_test_key else NOTIFICATION_CREATED,
-            billable_units=billable_units
+            billable_units=billable_units,
+            recipient_address=sanitise_response.get("recipient_address")
         )
         scan_pdf_object.delete()
     except BotoClientError:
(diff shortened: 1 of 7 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -474,14 +475,14 @@ def process_virus_scan_error(filename):
     raise error
 
 
-def update_letter_pdf_status(reference, status, billable_units):
+def update_letter_pdf_status(reference, status, billable_units, recipient_address=None):
+
+    update_dict = {'status': status, 'billable_units': billable_units, 'updated_at': datetime.utcnow()}
+    if recipient_address:
+        update_dict['to'] = recipient_address
```
klssmith: For templated letters we only store the first line in `to` - is there a reason why we're now storing the whole address instead?

app/service/send_pdf_letter_schema.py
```diff
@@ -8,6 +8,7 @@
         "filename": {"type": "string"},
         "created_by": {"type": "string"},
         "file_id": {"type": "string"},
+        "recipient_address": {"type": "string"}
     },
-    "required": ["postage", "filename", "created_by", "file_id"]
+    "required": ["postage", "filename", "created_by", "file_id", "recipient_address"]
```
klssmith: If you try and upload a letter with no address you now get a `500` error page instead of being able to send the letter. Maybe we need to add a story to make template preview check that there is actually an address at some point so that the error message is clearer

quis:
If we store the whole address then:
- we can change our mind about how to display it on the admin app and have them be effective immediately 
- we can let users search by any part of the address, when we we build search for letters

app/celery/letters_pdf_tasks.py
servingUpAces: We might want check if the address passed in are valid or search on that field. I think we would start to persist the whole address block on templated letters rather than restrict precompiled letters to the first line.

app/service/send_pdf_letter_schema.py
servingUpAces: Good thinking - I'll add a story.

servingUpAces: This has been done. The template-preview app will return a validation error if the address is missing.

PR #2693, opened 2020-01-06
Test isolation fix

leohemsted:
test_config manipulates os.environ. os.environ is an `environ` object, which acts like a dict but isn't in some subtle unknowable ways. The `reload_config` fixture would create a dict copy of the env, and then just call `os.environ = old_env` afterwards.

Boto3 would then complain that it couldn't load credentials (despite us using the mock_s3 fixture and also not having creds in the environment in the first place). Not entirely sure why this happens, but it does. For some reason, it being a `dict` instead of an `environ` object causes the mocking of boto3 to fail.

The solution is to not overwrite os.environ entirely, rather, use the standard dictionary setitem syntax to update the values to their previous values. Use `clear` to empty the environment too.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -40,12 +40,12 @@ alembic==1.3.2
 amqp==1.4.9
 anyjson==0.3.3
 attrs==19.3.0
-awscli==1.16.305
+awscli==1.16.310
 bcrypt==3.1.7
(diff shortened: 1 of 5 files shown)


idavidmcdonald:
Nice work Leo

PR #2696, opened 2020-01-10
handle document download errors properly

leohemsted:
if doc download returns a 403, that's a screw-up on our side. it's not helpful to a notify user for that to be passed on. the only thing they should care about is if it's a 400, because they uploaded a filetype we don't allow.

Everything else should return 500 internal server error.

### DIFF
--- a/app/clients/document_download.py
+++ b/app/clients/document_download.py
@@ -12,13 +12,8 @@ class DocumentDownloadError(Exception):
 
     @classmethod
     def from_exception(cls, e):
-        try:
-            message = e.response.json()['error']
-            status_code = e.response.status_code
-        except (TypeError, ValueError, AttributeError, KeyError):
-            message = 'connection error'
-            status_code = 503
-
+        message = e.response.json()['error']
+        status_code = e.response.status_code
         return cls(message, status_code)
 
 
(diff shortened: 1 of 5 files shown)


app/clients/document_download.py
```diff
@@ -45,11 +40,13 @@ def upload_document(self, service_id, file_contents):
 
             response.raise_for_status()
         except requests.RequestException as e:
-            error = DocumentDownloadError.from_exception(e)
-            current_app.logger.warning(
-                'Document download request failed with error: {}'.format(error.message)
-            )
-
-            raise error
+            if e.response.status_code == 400:
```
idavidmcdonald: Optional: I feel that this would be a good place for an inline code comments. It's obviously very easy to spot that we have different behaviour for a 400, but it's very unclear WHY the behaviour is different for 400. Obviously people could go search back in commit history but not ideal

tests/app/clients/test_document_download.py
```diff
@@ -33,26 +32,27 @@ def test_upload_document(document_download):
...
 
-    assert excinfo.value.message == 'Invalid encoding'
-    assert excinfo.value.status_code == 403
+    assert excinfo.value.message == 'Invalid mime type'
+    assert excinfo.value.status_code == 400
 
 
-def test_should_raise_for_connection_errors(document_download):
-    with pytest.raises(DocumentDownloadError) as excinfo, requests_mock.Mocker() as request_mock:
+def test_should_raise_other_statuses_as_exceptions(document_download):
```
idavidmcdonald: Optional:

Could rename test

`test_should_raise_non_400_statuses_as_exceptions` which is a bit more explicit to what is being tested.

I don't know if we expect any other response codes from doc download, if so you could parametrize them

leohemsted: i had a look, we don't return any other user errors, just two 400s (file has a virus or file isn't of a recognised type)

PR #2700, opened 2020-01-24
Add email_access_valdiated_at field to user table, populate it

CrystalPea:
## In this commit:
Add `email_access_valdiated_at` field to user table, populate it, and update it when users have to use their email to interact with Notify service.

### Initial population:
- If user has email_auth, set `last_validated_at` to `logged_in_at`.
- If user has sms_auth,  set`last_validated_at` to `created_at`.

### Then:
Update `email_access_valdiated_at` date when:
- user with `email_auth` logs in
- new user is created
- user resets password when logged out, meaning we send them an
email with a link they have to click to reset their password.


app/models.py
```diff
@@ -114,6 +114,7 @@ class User(db.Model):
     platform_admin = db.Column(db.Boolean, nullable=False, default=False)
     current_session_id = db.Column(UUID(as_uuid=True), nullable=True)
     auth_type = db.Column(db.String, db.ForeignKey('auth_type.name'), index=True, nullable=False, default=SMS_AUTH_TYPE)
+    last_validated_at = db.Column(db.DateTime, nullable=True)
```
idavidmcdonald: Do we want to allow it to be nullable?

idavidmcdonald: Also I think it would be a good idea to give it an even more verbose name like `email_access_last_validated_at`? Just if I was new the codebase, I probably wouldn't know this was to do with their email account. You could also a code comment above here describing what this variable is trying to record/solve (government users losing access to their email but retaining their SMS device and therefore still being able to access notify).

tests/app/dao/test_users_dao.py
```diff
@@ -148,11 +148,16 @@ def test_update_user_attribute(client, sample_user, user_attribute, user_value):
     assert getattr(sample_user, user_attribute) == user_value
 
 
+@freeze_time('2020-01-24T12:00:00')
 def test_update_user_password(notify_api, notify_db, notify_db_session, sample_user):
```
idavidmcdonald: This tests a user updating their password. Do we also need to test the case where a user is created for the first time that they have their last_validated_at set correctly?

app/models.py
CrystalPea: > Do we want to allow it to be nullable?

Hmmm you are right, we don't need it to be nullable

> Also I think it would be a good idea to give it an even more verbose name like email_access_last_validated_at

Good idea 👍

tests/app/dao/test_users_dao.py
CrystalPea: Cool, I will add a test like this 👍

app/dao/users_dao.py
```diff
@@ -28,15 +28,16 @@ def save_user_attribute(usr, update_dict={}):
...
 
-def save_model_user(usr, update_dict={}, pwd=None):
-    if pwd:
-        usr.password = pwd
-        usr.password_changed_at = datetime.utcnow()
+def save_model_user(user, update_dict={}, password=None):
+    if password:
+        user.password = password
+        user.password_changed_at = datetime.utcnow()
+        user.email_access_last_validated_at = datetime.utcnow()
```
idavidmcdonald: I'm not sure if we should have this line of code here for the following reason.

This function is `save_model_user` which may be used in varying places around our codebase without it being aware that if you change the password then you also set the `email_access_last_validated_at` unless you look at the code. This means there could be a risk of us bumping the last validated at date without being aware we are doing it and may not want to do so.

At the moment it isn't causing an issue with our existing codebase, the functionality seems to work as expected, however I feel it adds some opaque behaviour to the `save_model_user` function. 

I think I like how you've done it for `update_user_password(user, password, from_email=False)` with the `from_email` param, although I'm not sure I'd suggest doing it that way for this function itself. 

I'm wondering if we should make it super explicit what we are doing by adding a function to be called in the view code like `validate_email_access` which then each view would be responsible for deciding whether email access has been used.

I'm not 100% on this approach so would be keen to hear if you have any thoughts on potential approaches (including this one) and pros and cons.

app/user/rest.py
```diff
@@ -197,6 +197,8 @@ def verify_user_code(user_id):
 
     user_to_verify.current_session_id = str(uuid.uuid4())
     user_to_verify.logged_in_at = datetime.utcnow()
+    if user_to_verify.auth_type == 'email_auth':
```
idavidmcdonald: By the way, these two lines of code here may be similar to the way I was suggesting in my above larger comment. Each individual view is responsible for deciding whether the action has been done by email and therefore bumping the last validated at field.

migrations/versions/0313_add_last_validated_at.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+revision = '0313_add_last_validated_at'
+down_revision = '0312_populate_returned_letters'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('users', sa.Column('email_access_last_validated_at', sa.DateTime(), nullable=True))
+    # if user has email_auth, set email_access_last_validated_at on last login, else set it at user created_at date.
```
idavidmcdonald: Do we record anything in the DB that would indicate a user using sms auth reset their password at a particular time (as the resetting of a password does require email access)? 

If not, then this migration looks good

CrystalPea: We do not record what way password has been changed, so we have no way of knowing retroactively if a user had to use their email to change the password or not.

idavidmcdonald:
I've been doing some reading on database migrations (GOV.UK Pay had an incident a few days ago because of one).

I shared in slack but do take a read of 
https://gocardless.com/blog/zero-downtime-postgres-migrations-the-hard-parts/
https://www.braintreepayments.com/blog/safe-operations-for-high-volume-postgresql/#footnote-1

What I think is going to happen for your DB migration is it will open a transaction, the first statement will run `ALTER TABLE`. This will take an `ACCESS EXCLUSIVE` lock as per 
 https://www.postgresql.org/docs/9.5/sql-altertable.html.

Then we will update every column and then alter the table again.

Then the transaction will finish and the lock will be dropped.

This will mean we have an `ACCESS EXCLUSIVE` lock on the table for the whole of this migration. This type of lock, as per https://www.postgresql.org/docs/9.5/explicit-locking.html, means that we can't do READS of the table during the whole of this transaction. 

In reality, this table only has about 18000 rows so I assume everything will happen very quickly and although some queries could be waiting on it, this would be a very short time where we wouldn't see any user impact.

If this was happening on the notifications (8 million rows) or notification_history (many many millions) then I would be more concerned but I feel like it's probably OK for the user table. Just thought I'd share all my thinking and findings for learnings.

app/dao/users_dao.py
```diff
@@ -28,15 +28,17 @@ def save_user_attribute(usr, update_dict={}):
     db.session.commit()
 
 
-def save_model_user(usr, update_dict={}, pwd=None):
-    if pwd:
-        usr.password = pwd
-        usr.password_changed_at = datetime.utcnow()
+def save_model_user(user, update_dict={}, password=None, create_user=False):
```
idavidmcdonald: Optional: both this method and `update_user_password` could change the arguments you added to both be named the same thing for consistency e.g. `validate_email_access`

tests/app/dao/test_users_dao.py
```diff
@@ -42,11 +43,13 @@ def test_create_user(notify_db_session, phone_number):
...
     user = User(**data)
-    save_model_user(user)
+    save_model_user(user, password='password', create_user=True)
     assert User.query.count() == 1
-    assert User.query.first().email_address == email
-    assert User.query.first().id == user.id
-    assert User.query.first().mobile_number == phone_number
+    user = User.query.first()
+    assert user.email_address == email
+    assert user.id == user.id
```
idavidmcdonald: This line isn't testing anything (it's checking a variable is equal to itself)

CrystalPea: Good catch, I fixed it now.

app/dao/users_dao.py
CrystalPea: I think I will leave that one.

CrystalPea: hmm, I thought about it some more, actually I like the idea, I'll change it.

PR #2701, opened 2020-01-24
Catch uncaught JWT exceptions

idavidmcdonald:

For further context see https://github.com/alphagov/notifications-python-client/pull/156

## What does commit does do

- improves some of our error messages
- improves code comments of this pretty confusing section of our code
- adds some todos for bugs which I would like to fix in the future

## What I originally thought this PR would do

I previously added functionality to the python client to give error messages which link to our REST api guidance

https://github.com/alphagov/notifications-python-client/pull/156/files#diff-e4bd66cc2ec07db666fac1080662df81R12

My mistake was that I assumed once the client had been bumped in the API, these new pretty messages would be shown to users. That is not the case as we don't actually pass through the exceptions message from the python client but instead write our own in here. Now I've had time to think about it, I think this is the appropriate place to add the link to the guidance, rather than having it in the python client. 

## What still needs to be done
PRs will need to be opened for our API docs to update error messages that are found in there and have changed in this PR.

## How to test
```curl http://localhost:6011/v2/notifications -H "Authorization: Bearer eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJkNmFhMmM2OC1hMmQ5LTQ0MzctYWIxOS0zYWU4ZWIyMDI1NTMiLCJpYXQiOjE1Njg4MTg1Nzh9.ui3nIAi6-hP5pFqbSLlCJP78Mv_CPuKiB3i4HBf-eg9zrxH5FCk9HTCIfg2c92bBhPnEo0m36BTIZK2i_N5oSg"```

Replace the long token with one generated from https://jwt.io/ using

```
{
  "alg": "HS256",
  "typ": "JWT"
}
```

```
{
    'iss': YOUR SERVICE ID HERE,
    'iat': CURRENT UNIX TIMESTAMP,
    'aud': 'notifications.service.gov.uk' # this is the bad bit that will cause the error
}
```

And encrypt it with your secret from your API key as noted in https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header.

Note, token only valid for 30 seconds from timestamp so be quick!

You should get the new general token error message: Invalid token: make sure your API token matches the example at https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -10,6 +10,9 @@ from sqlalchemy.orm.exc import NoResultFound
 from app.dao.services_dao import dao_fetch_service_by_id_with_api_keys
 
 
+GENERAL_TOKEN_ERROR_MESSAGE = 'Invalid token: make sure your API token matches the example at https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header'  # noqa
+
+
 class AuthError(Exception):
     def __init__(self, message, code, service_id=None, api_key_id=None):
         self.message = {"token": [message]}
@@ -36,12 +39,12 @@ class AuthError(Exception):
 def get_auth_token(req):
     auth_header = req.headers.get('Authorization', None)
     if not auth_header:
-        raise AuthError('Unauthorized, authentication token must be provided', 401)
+        raise AuthError('Unauthorized: authentication token must be provided', 401)
 
     auth_scheme = auth_header[:7].title()
 
     if auth_scheme != 'Bearer ':
-        raise AuthError('Unauthorized, authentication bearer scheme must be used', 401)
+        raise AuthError('Unauthorized: authentication bearer scheme must be used', 401)
 
     return auth_header[7:]
 
@@ -60,17 +63,17 @@ def requires_admin_auth():
         g.service_id = current_app.config.get('ADMIN_CLIENT_USER_NAME')
         return handle_admin_key(auth_token, current_app.config.get('ADMIN_CLIENT_SECRET'))
     else:
-        raise AuthError('Unauthorized, admin authentication token required', 401)
+        raise AuthError('Unauthorized: admin authentication token required', 401)
 
 
 def requires_auth():
     request_helper.check_proxy_header_before_request()
 
     auth_token = get_auth_token(request)
-    client = __get_token_issuer(auth_token)
+    issuer = __get_token_issuer(auth_token)  # ie the `iss` claim which should be a service ID
 
     try:
-        service = dao_fetch_service_by_id_with_api_keys(client)
+        service = dao_fetch_service_by_id_with_api_keys(issuer)
     except DataError:
         raise AuthError("Invalid token: service id is not the right data type", 403)
     except NoResultFound:
@@ -92,10 +95,15 @@ def requires_auth():
             err_msg = "Invalid token: algorithm used is not HS256"
             raise AuthError(err_msg, 403, service_id=service.id, api_key_id=api_key.id)
         except TokenDecodeError:
+            # we attempted to validate the token but it failed meaning it was not signed using this api key.
+            # Let's try the next one
+            # TODO: Change this so it doesn't also catch `TokenIssuerError` or `TokenIssuedAtError` exceptions (which
+            # are children of `TokenDecodeError`) as these should cause an auth error immediately rather than
+            # continue on to check the next API key
             continue
         except TokenError:
-            err_msg = 'Invalid token: signature, api token is not valid'
-            raise AuthError(err_msg, 403, service_id=service.id, api_key_id=api_key.id)
+            # General error when trying to decode and validate the token
+            raise AuthError(GENERAL_TOKEN_ERROR_MESSAGE, 403, service_id=service.id, api_key_id=api_key.id)
 
         if api_key.expiry_date:
             raise AuthError("Invalid token: API key revoked", 403, service_id=service.id, api_key_id=api_key.id)
@@ -103,7 +111,7 @@ def requires_auth():
         g.service_id = api_key.service_id
         _request_ctx_stack.top.authenticated_service = service
         _request_ctx_stack.top.api_user = api_key
-        current_app.logger.info('API authorised for service {} with api key {}, using client {}'.format(
+        current_app.logger.info('API authorised for service {} with api key {}, using issuer {}'.format(
             service.id,
             api_key.id,
             request.headers.get('User-Agent')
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: app/authentication/auth.py [hunk 1]
```diff
@@ -92,33 +92,42 @@ def requires_auth():
...
         except TokenDecodeError:
+            # we attempted to validate the token but it failed meaning it was not signed using this api key.
+            # Let's try the next one
+            # TODO: Change this so it doesn't also catch `TokenIssuerError` or `TokenIssuedAtError` exceptions (which
+            # are children of `TokenDecodeError`) as these should cause an auth error immediately rather than
+            # continue on to check the next API key
             continue
+        except TokenError:
+            # General error when trying to decode and validate the token
+            err_msg = 'Invalid token: API token is not valid'
```
idavidmcdonald: @karlchillmaid in this case, the error is supposed to act as the catch all where we can't give more details. Therefore the best we can do now is to hand them off to the guidance e.g.

`Invalid token. Make sure your API token matches our guidance at https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header`

app/authentication/auth.py
```diff
@@ -92,33 +92,42 @@ def requires_auth():
...
 def __get_token_issuer(auth_token):
     try:
-        client = get_token_issuer(auth_token)
+        issuer = get_token_issuer(auth_token)
     except TokenIssuerError:
         raise AuthError("Invalid token: iss field not provided", 403)
     except TokenDecodeError:
-        raise AuthError("Invalid token: signature, api token is not valid", 403)
-    return client
+        raise AuthError("Invalid token: API token is not valid", 403)
```
idavidmcdonald: @karlchillmaid This case it would be hard to pinpoint exactly what they have done wrong or how to fix the problem. I think general guidance suggesting they look at our docs might be the most useful.

app/authentication/auth.py
```diff
@@ -127,4 +136,4 @@ def handle_admin_key(auth_token, secret):
     except TokenExpiredError:
         raise AuthError("Invalid token: expired, check that your system clock is accurate", 403)
     except TokenDecodeError:
-        raise AuthError("Invalid token: signature, api token is not valid", 403)
+        raise AuthError("Invalid token: API token is not valid", 403)
```
idavidmcdonald: Pointing them to the docs don't quite make sense here as this is an admin endpoint (ie us or one of the apps). We could just keep this short and sweet - `Invalid token` or `Invalid token: Could not decode your token`

karlchillmaid: 👍 
What about `Invalid token: Could not decode your API token`?

app/authentication/auth.py
karlchillmaid: How about `Invalid token: Make sure your API token matches the example at https://...`?

app/authentication/auth.py
karlchillmaid: Does `Make sure your API token matches the example at https://...` work here, or does the message need to be a more generic call to action?

karlchillmaid:
@idavidmcdonald that sounds sensible.
Once the guidance is published I'd planned to carry out a more general review of the documentation. In December I noticed some inconsistencies – I'm not sure whether these exist because of actual differences between clients/methods or whether they've just crept in over time.
Perhaps this is something we could pair on?

idavidmcdonald:
@karlchillmaid I think I've made our adjustments as discussed. Would be good to get your eyes on please.

One thing I've noticed a mismatch of is whether we capitalise after a colon e.g. 
`Invalid token: Make sure...` vs `Unauthorized: authentication bearer...`

app/authentication/auth.py
```diff
@@ -10,6 +10,9 @@
 from app.dao.services_dao import dao_fetch_service_by_id_with_api_keys
 
 
+GENERAL_TOKEN_ERROR_MESSAGE = 'Invalid token: Make sure your API token matches the example at https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header'  # noqa
```
karlchillmaid:
```suggestion
GENERAL_TOKEN_ERROR_MESSAGE = 'Invalid token: make sure your API token matches the example at https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header'  # noqa
```

app/authentication/auth.py
```diff
@@ -36,12 +39,12 @@ def to_dict_v2(self):
 def get_auth_token(req):
     auth_header = req.headers.get('Authorization', None)
     if not auth_header:
-        raise AuthError('Unauthorized, authentication token must be provided', 401)
+        raise AuthError('Unauthorized: authentication token must be provided', 401)
```
karlchillmaid: @idavidmcdonald Is in line with our style guide?

https://github.com/alphagov/notifications-manuals/wiki/A-to-Z-of-style#authorisation

app/authentication/auth.py
```diff
@@ -127,4 +138,4 @@ def handle_admin_key(auth_token, secret):
     except TokenExpiredError:
         raise AuthError("Invalid token: expired, check that your system clock is accurate", 403)
     except TokenDecodeError:
-        raise AuthError("Invalid token: signature, api token is not valid", 403)
+        raise AuthError("Invalid token: Could not decode your API token", 403)
```
karlchillmaid:
```suggestion
        raise AuthError("Invalid token: could not decode your API token", 403)
```

tests/app/authentication/test_authentication.py
```diff
@@ -123,7 +123,30 @@ def test_admin_auth_should_not_allow_request_with_no_iat(client, sample_api_key)
     request.headers = {'Authorization': 'Bearer {}'.format(token)}
     with pytest.raises(AuthError) as exc:
         requires_admin_auth()
-    assert exc.value.short_message == 'Invalid token: signature, api token is not valid'
+    assert exc.value.short_message == "Invalid token: Could not decode your API token"
```
karlchillmaid:
```suggestion
    assert exc.value.short_message == "Invalid token: could not decode your API token"
```

tests/app/authentication/test_authentication.py
```diff
@@ -253,7 +276,7 @@ def test_authentication_returns_error_when_admin_client_has_no_secrets(client):
             headers={'Authorization': 'Bearer {}'.format(token)})
     assert response.status_code == 403
     error_message = json.loads(response.get_data())
-    assert error_message['message'] == {"token": ["Invalid token: signature, api token is not valid"]}
+    assert error_message['message'] == {"token": ["Invalid token: Could not decode your API token"]}
```
karlchillmaid:
```suggestion
    assert error_message['message'] == {"token": ["Invalid token: could not decode your API token"]}
```

tests/app/authentication/test_authentication.py
```diff
@@ -268,7 +291,7 @@ def test_authentication_returns_error_when_admin_client_secret_is_invalid(client
         headers={'Authorization': 'Bearer {}'.format(token)})
     assert response.status_code == 403
     error_message = json.loads(response.get_data())
-    assert error_message['message'] == {"token": ["Invalid token: signature, api token is not valid"]}
+    assert error_message['message'] == {"token": ["Invalid token: Could not decode your API token"]}
```
karlchillmaid:
```suggestion
    assert error_message['message'] == {"token": ["Invalid token: could not decode your API token"]}
```

app/authentication/auth.py
idavidmcdonald: Yep, happy with this as it matches the http 401 error name of https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

idavidmcdonald:
@karlchillmaid Think that's all your content comments addressed. If you could take a final look please.

PR #2705, opened 2020-01-31
Populate email_access_validated_at_column and make it non-nullable

CrystalPea:
This is 2nd part of the original PR, which we had to split to avoid null values appearing in a non-nullable column in between deploying the migration and the code.


migrations/versions/0314_populate_email_access.py
```diff
@@ -0,0 +1,43 @@
...
+    """)
+    op.execute("""
+        UPDATE
+            users
+        SET
+            email_access_validated_at = logged_in_at
+        WHERE
+            auth_type = 'email_auth'
+        AND
+            logged_in_at IS NOT NULL
```
leohemsted: do we need `AND email_access_validated_at IS NULL` in this query too?

CrystalPea: The function it serves is: if someone has sms_auth and they have just reset their password using email link, they now have value in the `email_access_validated_at` field.
That check ensures this is not overwritten by an older date.

It's not essential, but it does something :)

leohemsted: ah fair enough - is it right then that we don't expect `email_access_validated_at` to be useful for users with email_auth because we just refer to `logged_in_at` instead?

or should `email_access_validated_at` be equal to `logged_in_at` for email auth?

CrystalPea: > or should email_access_validated_at be equal to logged_in_at for email auth?
Exactly this, `email_access_validated_at` is updated each time they log in for people with `email_auth`

PR #2706, opened 2020-02-03
Validate email access based on code_type and not auth_type

CrystalPea:
when verifying the code is correct. This way if user has sms_auth
and we send them verification code to validate their email access,
and they click the link in the email, their access will be validated
correctly.

### More changes added in that PR:
- Format email_access_validated_at when serializing
This is to bring it in line with other serialized dates in User
model, like logged_in_at and password_changed_at.

- Also get rid of check if password_changed_at has value, as
it is a non-nullable column, so it needs to always have value.

- Also set a default value for email_access_validated_at, to bring
it in line with other non-nullable columns.


CrystalPea:
@leohemsted I added more stuff to this PR following Becca's review of my related PR on admin.

PR #2708, opened 2020-02-05
Added a new endpoint to return the last used date for a template.

servingUpAces:
The existing endpoint returned a whole notification for the last time the template was used. But this only takes into account data in the last week. This new methods allows us to be specific about when the template was last used if ever but looking into the ft_notification_status table as well.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -72,7 +72,18 @@ def dao_get_last_template_usage(template_id, template_type, service_id):
 
 
 @statsd(namespace="dao")
-def dao_get_last_date_template_was_used(template_id, template_type, service_id):
+def dao_get_last_date_template_was_used(template_id, service_id):
+    last_date_from_notifications = db.session.query(
+        functions.max(Notification.created_at)
+    ).filter(
+        Notification.service_id == service_id,
+        Notification.template_id == template_id,
+        Notification.key_type != KEY_TYPE_TEST
+    ).scalar()
+
+    if last_date_from_notifications:
+        return last_date_from_notifications
+
     last_date = db.session.query(
         functions.max(FactNotificationStatus.bst_date)
     ).filter(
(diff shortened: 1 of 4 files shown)


app/dao/notifications_dao.py
```diff
@@ -70,6 +71,37 @@ def dao_get_last_template_usage(template_id, template_type, service_id):
...
+    ).filter(
+        Notification.service_id == service_id,
+        Notification.notification_type == template_type,
+        Notification.template_id == template_id,
+        Notification.key_type != KEY_TYPE_TEST
+    ).scalar()
+
+    if last_date and last_date_from_notifications:
+        if datetime.combine(last_date, datetime.utcnow().min.time()) >= last_date_from_notifications:
+            return last_date
```
leohemsted: is this ever going to be the case?

> "we have more recent ft_notification_status data than what is in the `Notification` table"

i don't think we need this check at all, and should always just return last_date_from_notifications in this case.

then all of this can just be `return last_date_from_notifications or last_date`

app/dao/notifications_dao.py
```diff
@@ -70,6 +71,37 @@ def dao_get_last_template_usage(template_id, template_type, service_id):
     ).first()
 
 
+@statsd(namespace="dao")
+def dao_get_last_date_template_was_used(template_id, template_type, service_id):
```
leohemsted: do we need this template_type filter here? it seems redundant, and we've got indexes on key_type, service_id and template_id already so i'm not sure it adds any speed-up.

tests/app/dao/notification_dao/test_notification_dao_template_usage.py
```diff
@@ -54,3 +54,56 @@ def test_last_template_usage_should_be_able_to_get_no_template_usage_history_if_
...
+    assert last_used_date == last_notification_date
+
+
+def test_dao_get_last_date_template_was_used_returns_none_if_never_used(sample_template):
+    assert not dao_get_last_date_template_was_used(template_id=sample_template.id,
+                                                   template_type='sms',
+                                                   service_id=sample_template.service_id)
+
+
+def test_dao_get_last_date_template_was_used_correct_date(sample_template):
```
leohemsted: i don't see how this test is differnet to the two at the top of the page, specifically the second test in `test_dao_get_last_date_template_was_used_returns_created_at_from_notifications`

tests/app/template_statistics/test_rest.py
```diff
@@ -189,3 +191,42 @@ def test_get_template_statistics_for_template_returns_empty_for_old_notification
...
+
+    json_resp = admin_request.get(
+        'template_statistics.get_last_used_datetime_for_template',
+        service_id=str(sample_template.service_id),
+        template_id=sample_template.id
+    )
+    assert json_resp['last_date_used'] == date_from_notification.strftime(DATETIME_FORMAT)
+
+
+def test_get_last_used_datetime_for_template_returns_400_if_service_does_not_exist(
```
leohemsted: minor, but this and the function below should be `...returns_404_if....`, not returns 400

PR #2709, opened 2020-02-06
Remove task that no longer runs

idavidmcdonald:
We no longer puts files in these s3 buckets (and have in fact deleted
the buckets) therefore this task is redundant and can be removed.


servingUpAces:
And fix the unit test :)

PR #2710, opened 2020-02-06
Fix None type error in last-used endpoint.

servingUpAces:


tests/app/template_statistics/test_rest.py
```diff
@@ -210,6 +210,17 @@ def test_get_last_used_datetime_for_template(
...
 
+def test_get_last_used_datetime_for_template_returns_none_if_no_usage_of_template(
+    admin_request, sample_template
+):
+    json_resp = admin_request.get(
+        'template_statistics.get_last_used_datetime_for_template',
+        service_id=str(sample_template.service_id),
+        template_id=sample_template.id
+    )
+    assert not json_resp['last_date_used']
```
idavidmcdonald: Optional: Change this to assert `is None` rather than `not` which is a bit more explicit (means we aren't returning 0 or False) and matches your test name

servingUpAces: I prefer to use is None, for some reason I thought that was not ok.

PR #2714, opened 2020-02-12
Docs in the db

leohemsted:
create a new `document_count` field in the Notification and NotificationHistory table. When someone posts with a document (or multiple documents), set it to a non-zero value. If someone posts without any documents, keep it set to null

There's an assumption that this will never be set to zero.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1407,6 +1407,8 @@ class Notification(db.Model):
 
     reply_to_text = db.Column(db.String, nullable=True)
 
+    document_download_count = db.Column(db.Integer, nullable=True)
+
     postage = db.Column(db.String, nullable=True)
     CheckConstraint("""
         CASE WHEN notification_type = 'letter' THEN
@@ -1682,6 +1684,8 @@ class NotificationHistory(db.Model, HistoryModel):
             postage is null
         END
     """)
+
+    document_download_count = db.Column(db.Integer, nullable=True)
 
     __table_args__ = (
         db.ForeignKeyConstraint(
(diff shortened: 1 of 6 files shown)


migrations/versions/0315_document_count.py
```diff
@@ -0,0 +1,26 @@
...
+import sqlalchemy as sa
+
+
+revision = '0315_document_count'
+down_revision = '0314_populate_email_access'
+
+
+def upgrade():
+    op.add_column('notifications', sa.Column('document_count', sa.Integer(), nullable=True))
+    op.create_index(op.f('ix_notifications_document_count'), 'notifications', ['document_count'], unique=False)
```
idavidmcdonald: From what I've been reading recently (see https://medium.com/braintree-product-technology/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#04b4 and section on `Create index`), this will hold an `ACCESS EXCLUSIVE` lock and mean no other other queries, including select statements can access the table which this migration happens. https://www.postgresql.org/docs/9.1/explicit-locking.html

Given the size of the notification and notification history tables, this has the potential to block all requests to the database for a long duration. There is a chance that because we are creating an index on a column that will just be nulls if the speed of creating the index will be super speedy and so it wouldn't matter anyway or if it may still cause a long running transaction.

You could look into creating the index concurrently as suggested in the first link.

leohemsted: update: removed index as decided it's not necessary yet

PR #2717, opened 2020-02-18
Look at all previous days when sending letters

idavidmcdonald:
## What
https://www.pivotaltracker.com/n/projects/1443052

Previously, when running the `collate_letter_pdfs_for_day` task, we
would only send letters that were created between 5:30pm yesterday and
5:30 today.

Now we send letters that were created before 5:30pm today and that are
still waiting to be sent. This will help us automatically attempt to
send letters that may have fallen through the gaps and not been sent the
previous day when they should have been.

Previously we solved the problem of letters that had fallen the gap by
having to run the task with a date parameter for example
`collate_letter_pdfs_for_day('2020-02-18'). We no longer need this date
parameter as we will always look back across previous days too for
letters that still need sending.

Note, we have to change from using the pagination `list_objects_v2` to
instead getting each individual notification from s3. We reduce load by
using `HEAD` rather than `GET` but this will still greatly increase the
number of API calls. We acknowledge there will be a small cost to this,
say 50p for 5000 letters and think this is tolerable. Boto3 also handles
retries itself so if when making one of the many HEAD requests, there is
a networking blip then it should be retried automatically for us.

## How to test locally
Make sure you have sent a letter before 5:30pm that exists in the `development-letters-pdf` bucket for yesterdays print run.
Make sure you have sent a letter before 5:30pm that exists in  `development-letters-pdf` bucket for the print run 2 days ago.

Note, you can likely set this up by uploading two letters today. And then:
- change their created at date in your database
- move the file in the s3 bucket to the relevant days sending folder
- change the filename of the file you are moving in s3 to match your new created_at timestamp

```
source environment.sh
flask shell
from app.celery.letters_pdf_tasks import collate_letter_pdfs_for_day
collate_letter_pdfs_for_day()
```

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -25,7 +25,6 @@ from app.dao.notifications_dao import (
     update_notification_status_by_id,
     dao_update_notification,
     dao_get_notification_by_reference,
-    dao_get_notifications_by_references,
     dao_update_notifications_by_reference,
     dao_get_letters_to_be_printed,
 )
@@ -180,10 +179,6 @@ def get_key_and_size_of_letters_to_be_sent_to_print(print_run_date):
             postage=letter.postage
         )
 
-        current_app.logger.info(
-            f'Found notification {letter.id} to send to DVLA to print: {letter_file_name}'
-        )
-
         letter_head = s3.head_s3_object(current_app.config['LETTERS_PDF_BUCKET_NAME'], letter_file_name)
         letter_pdfs.append({"Key": letter_file_name, "Size": letter_head['ContentLength']})
 
@@ -199,7 +194,7 @@ def group_letters(letter_pdfs):
     running_filesize = 0
     list_of_files = []
     for letter in letter_pdfs:
-        if letter['Key'].lower().endswith('.pdf') and letter_in_created_state(letter['Key']):
+        if letter['Key'].lower().endswith('.pdf'):
             if (
                 running_filesize + letter['Size'] > current_app.config['MAX_LETTER_PDF_ZIP_FILESIZE'] or
                 len(list_of_files) >= current_app.config['MAX_LETTER_PDF_COUNT_PER_ZIP']
@@ -213,22 +208,6 @@ def group_letters(letter_pdfs):
 
     if list_of_files:
         yield list_of_files
-
-
-def letter_in_created_state(filename):
-    # filename looks like '2018-01-13/NOTIFY.ABCDEF1234567890.D.2.C.C.20180113120000.PDF'
-    subfolder = filename.split('/')[0]
-    ref = get_reference_from_filename(filename)
-    notifications = dao_get_notifications_by_references([ref])
-    if notifications:
-        if notifications[0].status == NOTIFICATION_CREATED:
-            return True
-        current_app.logger.info('Collating letters for {} but notification with reference {} already in {}'.format(
-            subfolder,
-            ref,
-            notifications[0].status
-        ))
-    return False
 
 
 @notify_celery.task(bind=True, name='process-virus-scan-passed', max_retries=15, default_retry_delay=300)
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -132,8 +132,6 @@ def collate_letter_pdfs_to_be_sent():
     if print_run_date.time() < LETTER_PROCESSING_DEADLINE:
         print_run_date = print_run_date - timedelta(days=1)
 
-    # We can truncate the datetime to a date and add our own time (5:30pm) because UTC to BST does not
-    # make a difference to the date since it is triggered mid afternoon.
     print_run_deadline = print_run_date.replace(
         hour=17, minute=30, second=0, microsecond=0
     )
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -128,15 +128,17 @@ def collate_letter_pdfs_for_day():
     that have not yet been sent.
     If run after midnight, it will collect up letters created before 5:30pm the day before.
     """
-    date = convert_utc_to_bst(datetime.utcnow())
-    if date.time() < LETTER_PROCESSING_DEADLINE:
-        date = date - timedelta(days=1)
+    print_run_date = convert_utc_to_bst(datetime.utcnow())
+    if print_run_date.time() < LETTER_PROCESSING_DEADLINE:
+        print_run_date = print_run_date - timedelta(days=1)
 
-    # Using the truncated date is ok because UTC to BST does not make a difference to the date,
-    # since it is triggered mid afternoon.
-    print_run_date = date.strftime("%Y-%m-%d")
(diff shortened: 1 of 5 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -117,26 +121,31 @@ def get_letters_pdf(template, contact_block, filename, values):
...
+    """
+    date = convert_utc_to_bst(datetime.utcnow())
+    if date.time() < LETTER_PROCESSING_DEADLINE:
+        date = date - timedelta(days=1)
+
+    # Using the truncated date is ok because UTC to BST does not make a difference to the date,
+    # since it is triggered mid afternoon.
+    print_run_date = date.strftime("%Y-%m-%d")
+
+    letters_to_print = get_key_and_size_of_letters_to_be_sent_to_print(print_run_date)
```
idavidmcdonald: Wondering if it's best to past a "%Y-%m-%d" string into this function or instead maybe a datetime

app/celery/letters_pdf_tasks.py
```diff
@@ -159,6 +168,28 @@ def collate_letter_pdfs_for_day(date=None):
...
+        )
+
+        current_app.logger.info(
+            f'Found notification {letter.id} to send to DVLA to print: {letter_file_name}'
+        )
+
+        letter_head = s3.head_s3_object(current_app.config['LETTERS_PDF_BUCKET_NAME'], letter_file_name)
+        letter_pdfs.append({"Key": letter_file_name, "Size": letter_head['ContentLength']})
+
+    return letter_pdfs
```
idavidmcdonald: They are no longer sorted by `Key` like they used to be but now `created_at`. We feel that is OK?

app/celery/letters_pdf_tasks.py
```diff
@@ -159,6 +168,28 @@ def collate_letter_pdfs_for_day(date=None):
         )
 
 
+def get_key_and_size_of_letters_to_be_sent_to_print(print_run_date):
+    letters_awaiting_sending = dao_get_letters_to_be_printed(print_run_date)
+
+    letter_pdfs = []
+    for letter in letters_awaiting_sending:
+        letter_file_name = get_letter_pdf_filename(
```
idavidmcdonald: I won't lie, I copied and adapted this code from elsewhere and then matched my tests results to what I was getting which feels naughty. I'm wondering if I should introduce something like `get_letter_pdf_filename_from_notification` which can then be unit tested and make sure that given a letter, the correct arguments are passed into the `get_letter_pdf_filename` function?

app/celery/letters_pdf_tasks.py
servingUpAces: Makes sense, since the first thing the dao method does is convert it to datetime

app/celery/letters_pdf_tasks.py
servingUpAces: I think that's ok

app/dao/notifications_dao.py
```diff
@@ -731,6 +731,24 @@ def notifications_not_yet_sent(should_be_sending_after_seconds, notification_typ
...
+    Given a date for a print run, Return all letters created before 5:30pm that day that have not yet been sent
+    """
+    last_processing_deadline = datetime.strptime(print_run_date, "%Y-%m-%d").replace(
+        hour=17, minute=30, second=0, microsecond=0
+    )
+
+    notifications = Notification.query.filter(
+        Notification.created_at < convert_bst_to_utc(last_processing_deadline),
+        Notification.notification_type == LETTER_TYPE,
+        Notification.status == NOTIFICATION_CREATED
```
servingUpAces: Add a filter Notification.key_type == NORMAL

servingUpAces: or key_type != TEST

app/celery/letters_pdf_tasks.py
idavidmcdonald: Decided not to make any changes as not to go down a rabbit hole, plus I understood the code much better after you explained what `dont_use_sending_date` was for so I'm more confident in my code changes.

tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -242,52 +241,196 @@ def test_create_letters_gets_the_right_logo_when_service_has_letter_branding_log
...
-        {'Key': 'B.pDf', 'Size': 2},
-        {'Key': 'A.PDF', 'Size': 1},
-        {'Key': 'C.pdf', 'Size': 3}
+@freeze_time('2020-02-17 18:00:00')
+def test_get_key_and_size_of_letters_to_be_sent_to_print(notify_api, mocker, sample_letter_template):
+    create_notification(
+        template=sample_letter_template,
+        status='created',
+        reference='ref0',
+        created_at=(datetime.now() - timedelta(hours=2)).isoformat()
```
servingUpAces: Do we need isoformat?

tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -242,52 +241,196 @@ def test_create_letters_gets_the_right_logo_when_service_has_letter_branding_log
...
+
+    mocker.patch('app.celery.tasks.s3.head_s3_object', side_effect=[
+        {'ContentLength': 2},
+        {'ContentLength': 1},
+        {'ContentLength': 3},
     ])
-    mock_group_letters = mocker.patch('app.celery.letters_pdf_tasks.group_letters', return_value=[
-        [{'Key': 'A.PDF', 'Size': 1}, {'Key': 'B.pDf', 'Size': 2}],
-        [{'Key': 'C.pdf', 'Size': 3}]
+    mocker.patch('app.celery.letters_pdf_tasks.group_letters', return_value=[
```
servingUpAces: Do we need to mock this method?

tests/app/celery/test_letters_pdf_tasks.py
idavidmcdonald: It appears we do not. I will remove

tests/app/celery/test_letters_pdf_tasks.py
idavidmcdonald: Technically not, I'll remove it

app/celery/letters_pdf_tasks.py
```diff
@@ -115,28 +118,35 @@ def get_letters_pdf(template, contact_block, filename, values):
...
-        key=lambda letter: letter['Key']
+    This would usually be run at 5.50pm and collect up letters created between before 5:30pm today
+    that have not yet been sent.
+    If run after midnight, it will collect up letters created before 5:30pm the day before.
+    """
+    print_run_date = convert_utc_to_bst(datetime.utcnow())
+    if print_run_date.time() < LETTER_PROCESSING_DEADLINE:
+        print_run_date = print_run_date - timedelta(days=1)
+
+    # We can truncate the datetime to a date and add our own time (5:30pm) because UTC to BST does not
```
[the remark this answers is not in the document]
idavidmcdonald: Removed

servingUpAces:
Tested locally, looks good

PR #2721, opened 2020-02-20
Allow multiple keys for the `ADMIN_CLIENT_SECRET`

idavidmcdonald:
## What
https://www.pivotaltracker.com/story/show/171352263 - but just does the admin client secret to start small, we can do the other as separate PRs once we are happy with this

## Requires 
https://github.com/alphagov/notifications-credentials/pull/143 to be merged first

## How to test

Pull down this branch
In `app/config.py` change your development config to
`ADMIN_CLIENT_SECRETS = json.loads(os.environ.get('ADMIN_CLIENT_SECRETS', '[]'))`

```
export ADMIN_CLIENT_SECRETS='["key1", "key2"]'
./scripts/run_app.sh
```

Then open the admin app code and change the following
https://github.com/alphagov/notifications-admin/blob/master/app/config.py#L106

to be

`ADMIN_CLIENT_SECRET="key1"`.

Spin up your apps and log into the admin app, this will validate that it can talk using `key1`.

Then quit the admin app, change `ADMIN_CLIENT_SECRET="key2"` and retest again.

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -61,7 +61,12 @@ def requires_admin_auth():
 
     if client == current_app.config.get('ADMIN_CLIENT_USER_NAME'):
         g.service_id = current_app.config.get('ADMIN_CLIENT_USER_NAME')
-        return handle_admin_key(auth_token, current_app.config.get('ADMIN_CLIENT_SECRET'))
+
+        secret = ""
+        if len(current_app.config.get('ADMIN_CLIENT_SECRETS')):
+            secret = current_app.config.get('ADMIN_CLIENT_SECRETS')[0]
+
+        return handle_admin_key(auth_token, secret)
     else:
         raise AuthError('Unauthorized: admin authentication token required', 401)
 
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -62,18 +62,17 @@ def requires_admin_auth():
     if client == current_app.config.get('ADMIN_CLIENT_USER_NAME'):
         g.service_id = current_app.config.get('ADMIN_CLIENT_USER_NAME')
 
-        if len(current_app.config.get('ADMIN_CLIENT_SECRETS')):
-            for secret in current_app.config.get('ADMIN_CLIENT_SECRETS'):
-                try:
-                    decode_jwt_token(auth_token, secret)
-                    return
-                except TokenExpiredError:
-                    raise AuthError("Invalid token: expired, check that your system clock is accurate", 403)
-                except TokenDecodeError:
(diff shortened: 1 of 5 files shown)


app/authentication/auth.py
```diff
@@ -61,7 +61,22 @@ def requires_admin_auth():
 
     if client == current_app.config.get('ADMIN_CLIENT_USER_NAME'):
         g.service_id = current_app.config.get('ADMIN_CLIENT_USER_NAME')
-        return handle_admin_key(auth_token, current_app.config.get('ADMIN_CLIENT_SECRET'))
+
+        if len(current_app.config.get('ADMIN_CLIENT_SECRETS')):
```
leohemsted: don't think you need this if check. the for loop will just do nothing if the current app has an empty array

app/authentication/auth.py
```diff
@@ -61,7 +61,22 @@ def requires_admin_auth():
...
+            for secret in current_app.config.get('ADMIN_CLIENT_SECRETS'):
+                try:
+                    decode_jwt_token(auth_token, secret)
+                    return
+                except TokenExpiredError:
+                    raise AuthError("Invalid token: expired, check that your system clock is accurate", 403)
+                except TokenDecodeError:
+                    # TODO: Change this so it doesn't also catch `TokenIssuerError` or `TokenIssuedAtError` exceptions
+                    # (which are children of `TokenDecodeError`) as these should cause an auth error immediately rather
+                    # than continue on to check the next API key
```
leohemsted: minor, comment suggestion
```suggestion
                    # than continue on to check the next admin client secret
```

app/authentication/auth.py
idavidmcdonald: Fixed

app/authentication/auth.py
idavidmcdonald: Fixed

app/config.py
```diff
@@ -369,7 +369,7 @@ class Development(Config):
     TRANSIENT_UPLOADED_LETTERS = 'development-transient-uploaded-letters'
     LETTER_SANITISE_BUCKET_NAME = 'development-letters-sanitise'
 
-    ADMIN_CLIENT_SECRET = 'dev-notify-secret-key'
+    API_INTERNAL_SECRETS = json.loads(os.environ.get('API_INTERNAL_SECRETS', '[]'))
```
servingUpAces: If we aren't going to have a default here can we add a comment in the README (in the section we talk about environment variables), please?

idavidmcdonald: Sorry, that was an accidental commit. I've put it back to the default that it should have been. This should fix the concourse tests too.

PR #2723, opened 2020-02-24
Organisation usage

servingUpAces:
Add queries to handle returning usage for all services associated to a given organisation.
Only data for live services are returned. 
Added an endpoint to retrieve the organisation usage data. The endpoint may need to change, but I'd like to test it on a real data set. Once we are happy with the performance, I'll come back for validate the dataset has everything it needs and write all the tests. 

We were populating ft_billing for "yesterday" and "today", but I have decided to remove "yesterday". The only reason to include "yesterday" is so that the reports are accurate after midnight and before 02:00, but I think the performance gain is more important than the off chance someone is going to use the report that early in the morning

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -244,11 +244,9 @@ def fetch_monthly_billing_for_year(service_id, year):
     today = convert_utc_to_bst(datetime.utcnow()).date()
     # if year end date is less than today, we are calculating for data in the past and have no need for deltas.
     if year_end_date >= today:
-        yesterday = today - timedelta(days=1)
-        for day in [yesterday, today]:
-            data = fetch_billing_data_for_day(process_day=day, service_id=service_id, check_permissions=True)
-            for d in data:
-                update_fact_billing(data=d, process_day=day)
+        data = fetch_billing_data_for_day(process_day=today, service_id=service_id, check_permissions=True)
+        for d in data:
+            update_fact_billing(data=d, process_day=today)
 
     email_and_letters = db.session.query(
         func.date_trunc('month', FactBilling.bst_date).cast(Date).label("month"),
@@ -648,12 +646,10 @@ def fetch_usage_year_for_organisation(organisation_id, year):
     services = dao_get_organisation_live_services(organisation_id)
     # if year end date is less than today, we are calculating for data in the past and have no need for deltas.
     if year_end_date >= today:
-        yesterday = today - timedelta(days=1)
         for service in services:
-            for day in [yesterday, today]:
-                data = fetch_billing_data_for_day(process_day=day, service_id=service.id)
-                for d in data:
-                    update_fact_billing(data=d, process_day=day)
+            data = fetch_billing_data_for_day(process_day=today, service_id=service.id)
+            for d in data:
+                update_fact_billing(data=d, process_day=today)
     service_with_usage = {}
     # initialise results
     for service in services:
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -12,7 +12,7 @@ from app.dao.date_util import (
     get_financial_year,
     get_financial_year_for_datetime
 )
-from app.dao.organisation_dao import dao_get_organisation_services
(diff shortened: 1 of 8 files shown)


app/dao/fact_billing_dao.py
```diff
@@ -539,3 +540,153 @@ def create_billing_record(data, rate, process_day):
...
+    query = db.session.query(
+        Service.name.label("service_name"),
+        Service.id.label("service_id"),
+        func.sum(FactBilling.notifications_sent * FactBilling.rate).label("letter_cost")
+    ).select_from(
+        Service
+    ).join(
+        FactBilling, FactBilling.service_id == Service.id,
+    ).filter(
+        FactBilling.service_id == Service.id,
```
idavidmcdonald: check if can get rid of

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -625,3 +629,67 @@ def test_fetch_letter_line_items_for_all_service(notify_db_session):
...
+def test_fetch_usage_year_for_organisation(notify_db_session):
+    org, org_2, service, service_2, service_3, service_sms_only, \
+        org_with_emails, service_with_emails = set_up_usage_data(datetime(2019, 5, 1))
+    service_with_emails_for_org = create_service(service_name='Service with emails for org')
+    dao_add_service_to_organisation(service=service_with_emails_for_org, organisation_id=org.id)
+    template = create_template(service=service_with_emails_for_org, template_type='email')
+    create_ft_billing(bst_date=datetime(2019, 5, 1),
+                      template=template,
+                      notifications_sent=1100)
+    results = fetch_usage_year_for_organisation(org.id, 2019)
```
idavidmcdonald: This appears to be testing this financial year right? As in 2019-2020. If so, I'd both a test that tests the current year (could be done by freezetiming this test to our current date) but also a test that tests getting usage for a whole previous financial year that we are not in please

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -625,3 +629,67 @@ def test_fetch_letter_line_items_for_all_service(notify_db_session):
...
+    assert second_row['service_name'] == service_with_emails_for_org.name
+    assert second_row['free_sms_limit'] == 0
+    assert second_row['sms_remainder'] == 0
+    assert second_row['chargeable_billable_sms'] == 0
+    assert second_row['sms_cost'] == 0
+    assert second_row['letter_cost'] == 0
+    assert second_row['emails_sent'] == 1100
+
+
+def test_fetch_usage_year_for_organisation_populates_ft_billing_for_today(notify_db_session):
```
idavidmcdonald: Should this also test that it creates the ft_billing for yesterday too?

app/dao/fact_billing_dao.py
```diff
@@ -539,3 +540,152 @@ def create_billing_record(data, rate, process_day):
...
+    year_end_date = convert_utc_to_bst(year_end_datetime).date()
+
+    today = convert_utc_to_bst(datetime.utcnow()).date()
+    services = dao_get_organisation_live_services(organisation_id)
+    # if year end date is less than today, we are calculating for data in the past and have no need for deltas.
+    if year_end_date >= today:
+        yesterday = today - timedelta(days=1)
+        for service in services:
+            for day in [yesterday, today]:
+                data = fetch_billing_data_for_day(process_day=day, service_id=service.id)
```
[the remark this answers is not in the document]
servingUpAces: Ah wait, I think it only needs to update today, since yesterday should have been updated in the nightly task. I'll take another look.

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -582,8 +585,6 @@ def test_fetch_sms_billing_for_all_services_with_remainder(notify_db_session):
 
     results = fetch_sms_billing_for_all_services(datetime(2019, 5, 1), datetime(2019, 5, 31))
     assert len(results) == 3
-    # (organisation_name, organisation_id, service_name, free_sms_fragment_limit, sms_rate,
```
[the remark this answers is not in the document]
servingUpAces: I've removed it

PR #2727, opened 2020-02-25
Give existing services upload letters permission

quis:
This gives the feature to all existing users.

Depends on: 
- [x] https://github.com/alphagov/notifications-api/pull/2726

After this we will:
- remove this permission from the codebase entirely so that everyone has this feature and can’t switch it off

### DIFF
--- a/migrations/versions/0317_uploads_for_all.py
+++ b/migrations/versions/0317_uploads_for_all.py
@@ -1,0 +1,41 @@
+"""
+
+Revision ID: 0317_uploads_for_all
+Revises: 0316_int_letters_permission
+Create Date: 2019-05-13 10:44:51.867661
+
+"""
+from alembic import op
+from app.models import UPLOAD_LETTERS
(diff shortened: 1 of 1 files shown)


migrations/versions/0316_uploads_for_all.py
```diff
@@ -0,0 +1,41 @@
...
+           )
+    """.format(
+        permission=UPLOAD_LETTERS
+    ))
+
+
+def downgrade():
+    op.execute("DELETE from service_permissions where permission = '{}'".format(
+        UPLOAD_LETTERS
+    )
```
leohemsted: this will delete everything, including stuff previously added manually. i think i'd rather no possible downgrade for this migration i think?

quis: Yeah I was in 2 minds about what the right way to go there was.

Updated to `pass` in 6677e32

PR #2732, opened 2020-02-27
Return recipient for letter uploads

quis:
If your caseworking system always spits out files with the same name it will be hard to differentiate them when looking at the uploads page.

Seeing who the letter was sent to will help you differentiate them.

We can’t do this until the API returns the recipient.


app/upload/rest.py
```diff
@@ -45,6 +45,7 @@ def get_paginated_uploads(service_id, limit_days, page):
                 "%Y-%m-%d %H:%M:%S") if upload.scheduled_for else upload.created_at.strftime("%Y-%m-%d %H:%M:%S"),
             'upload_type': upload.upload_type,
             'template_type': None,
```
servingUpAces: template_type can just be upload.template_type, then you can remove line 63 where it's set for jobs only.

quis: Changed in f57b544

PR #2734, opened 2020-02-27
Add a filter to exclude services that do not count_as_live

servingUpAces:
This will exclude any test services, this query is used for the organistion usage page.

### DIFF
--- a/app/dao/organisation_dao.py
+++ b/app/dao/organisation_dao.py
@@ -34,8 +34,9 @@ def dao_get_organisation_services(organisation_id):
 def dao_get_organisation_live_services(organisation_id):
     return Service.query.filter_by(
         organisation_id=organisation_id,
-        restricted=False
-    ).all()
+        restricted=False,
+        count_as_live=True
+    ).order_by(Service.name).all()
 
 
 def dao_get_organisation_by_id(organisation_id):
--- a/tests/app/dao/test_organisation_dao.py
+++ b/tests/app/dao/test_organisation_dao.py
@@ -15,7 +15,8 @@ from app.dao.organisation_dao import (
     dao_add_service_to_organisation,
     dao_get_invited_organisation_user,
     dao_get_users_for_organisation,
-    dao_add_user_to_organisation
+    dao_add_user_to_organisation,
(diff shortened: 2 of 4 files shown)


app/dao/organisation_dao.py
```diff
@@ -34,8 +34,9 @@ def dao_get_organisation_services(organisation_id):
 def dao_get_organisation_live_services(organisation_id):
     return Service.query.filter_by(
         organisation_id=organisation_id,
-        restricted=False
-    ).all()
+        restricted=False,
```
idavidmcdonald: Do we still need the check on restricted False? I don't think we can have something with a count_as_live being True AND having restricted False too so this line isn't really doing anything? Or are we just being extra cautious?

tests/app/dao/test_organisation_dao.py
```diff
@@ -356,3 +357,22 @@ def test_get_organisation_by_email_address_ignores_gsi_gov_uk(notify_db_session)
...
     assert org == found_org
+
+
+def test_dao_get_organisation_live_services(notify_db_session):
+    org = create_organisation()
+    live_service = create_service(service_name='live service', restricted=False, count_as_live=True)
+    do_not_count_as_live_service = create_service(
+        service_name='do_not_count_as_live_service', restricted=False, count_as_live=False
+    )
+    archived_service_but_count_as_live = create_service(
```
idavidmcdonald: I think we should move all testing of logic about which services are live into this test (test_fetch_usage_year_for_organisation_does_not_return_data_for_service_where_count_as_live_is_false) and then assert in `test_fetch_usage_year_for_organisation_only_returns_data_for_live_services` that we are calling this method.

Then this test can test the following cases
- live service
- archived service
- trial service
- live service not to be counted as live because it's one for our testing
- any other types?

and then it's very clear and in one place the behaviour we expect

quis:
We’ve changed our mind about this. We still want ones that we’re not showing on the live services list to show up on the organisation page if it’s possible for them to incur debt.

We’ve done some manual work to tidy up the archived ones from the Cabinet Office page by setting them back to trial mode.

PR #2737, opened 2020-03-03
Add an endpoint to return returned letter stats

quis:
Currently the dashboard in the admin app pull the entire returned letter summary for a service to calculate how many letters have been returned in the last seven days.

Adding a separate endpoint for this purpose is better because:
- it’s a more efficient query
- it’s less data to send down the pipe
- it gives us a place to return the complete datetime, so the dashboard can be more precise about when the most recent report was

### DIFF
--- a/app/dao/returned_letters_dao.py
+++ b/app/dao/returned_letters_dao.py
@@ -51,7 +51,7 @@ def insert_or_update_returned_letters(references):
         db.session.connection().execute(stmt)
 
 
-def fetch_returned_letter_count(service_id):
+def fetch_recent_returned_letter_count(service_id):
     return db.session.query(
         func.count(ReturnedLetter.notification_id).label('returned_letter_count'),
     ).filter(
--- a/app/dao/returned_letters_dao.py
+++ b/app/dao/returned_letters_dao.py
@@ -13,6 +13,7 @@ from app.models import (
     Template,
     User,
 )
+from app.utils import midnight_n_days_ago
 
 
 def _get_notification_ids_for_references(references):
@@ -48,6 +49,25 @@ def insert_or_update_returned_letters(references):
             }
         )
         db.session.connection().execute(stmt)
+
+
+def fetch_returned_letter_count(service_id):
+    return db.session.query(
+        func.count(ReturnedLetter.notification_id).label('returned_letter_count'),
+    ).filter(
+        ReturnedLetter.service_id == service_id,
+        ReturnedLetter.reported_at > midnight_n_days_ago(7),
+    ).one()
+
+
+def fetch_most_recent_returned_letter(service_id):
+    return db.session.query(
+        ReturnedLetter.reported_at,
+    ).filter(
+        ReturnedLetter.service_id == service_id,
+    ).order_by(
+        desc(ReturnedLetter.reported_at)
+    ).first()
 
 
 def fetch_returned_letter_summary(service_id):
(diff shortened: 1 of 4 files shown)


app/service/rest.py
```diff
@@ -946,6 +948,25 @@ def check_if_reply_to_address_already_in_use(service_id, email_address):
         )
 
 
+@service_blueprint.route('/<uuid:service_id>/returned-letter-statistics', methods=['GET'])
+def returned_letter_statistics(service_id):
+
+    most_recent = fetch_most_recent_returned_letter(service_id)
```
idavidmcdonald: This will return dates that are longer than 7 days ago. If it does this then we will still run the `fetch_returned_letter_count` function which will be of no use as there won't be anything in the last 7 days so a wasted DB query

app/dao/returned_letters_dao.py
```diff
@@ -50,6 +51,25 @@ def insert_or_update_returned_letters(references):
         db.session.connection().execute(stmt)
 
 
+def fetch_returned_letter_count(service_id):
```
idavidmcdonald: Optional: Was vaguely wondering about joining these two queries together so you only issue one query that returns both fields. Obviously this way is more general and could have more reuse but also YAGNI. I'm in two minds and don't feel strongly either way but just thought I'd raise it as something to consider if you hadn't already.

idavidmcdonald: Wonder if function name should be `fetch_recent_returned_letter_count` to indicate the behaviour with only finding things reported in the last 7 days

tests/app/dao/test_returned_letters_dao.py
```diff
@@ -90,6 +98,62 @@ def test_insert_or_update_returned_letters_with_duplicates_in_reference_list(sam
...
+        reported_at=datetime(2010, 11, 8, 10, 0),
+    )
+
+    with freeze_time('2010-11-08 10:10'):
+        result = fetch_returned_letter_count(sample_service.id)
+
+    assert result.returned_letter_count == 3
+
+
+def test_fetch_most_recent_returned_letter_for_service(sample_service):
```
idavidmcdonald: Good test, I like the comments you've used

tests/app/service/test_rest.py
```diff
@@ -3382,6 +3382,20 @@ def test_get_monthly_notification_data_by_service(mocker, admin_request):
     assert response == []
 
 
+@freeze_time('2019-12-11 13:30')
+def test_get_returned_letter_statistics(admin_request, sample_service):
```
idavidmcdonald: Should probably do a test for the case were there are no recent returned letters too please

app/service/rest.py
quis: I’ve removed the wasted query in 8c47d07. This means it makes sense to keep the two separate DAO methods.

app/dao/returned_letters_dao.py
quis: Changed in 1c058e07

tests/app/service/test_rest.py
quis: Done in 02c824a

app/service/rest.py
```diff
@@ -946,6 +948,35 @@ def check_if_reply_to_address_already_in_use(service_id, email_address):
...
+    if most_recent_reported_at < midnight_n_days_ago(7):
+        return jsonify({
+            'returned_letter_count': 0,
+            'most_recent_report': most_recent.reported_at.strftime(DATETIME_FORMAT_NO_TIMEZONE),
+        })
+
+    count = fetch_recent_returned_letter_count(service_id)
+
+    return jsonify({
+        'returned_letter_count': count.returned_letter_count,
```
idavidmcdonald: Optional: The key could be something more descriptive e.g. `returned_letters_in_last_7_days_count`

quis: Keeping it as-is for consistency with the other returned letter endpoint responses.

PR #2742, opened 2020-03-09
Only add overlay for pages which are invalid for precompiled letters uploaded via the API

idavidmcdonald:
https://www.pivotaltracker.com/story/show/171130095

## Design decision
As suggested by the story, 'The code path for the notifications page goes via the API (whereas the upload page goes straight to the template preview app). We should probably investigate if it still needs to do this. It would be nice if the viewnotification page called the TemplatePreview.from(in)valid_pdf functions. '

I did look into this here which is a very rough draft:
https://github.com/alphagov/notifications-admin/compare/template-preview-page?expand=1

I managed to get it working but it seemed like it would be more work and this is all fairly complex already so didn't want to change too much in one go. Would potentially involve moving a bunch of functionality into utils or having things not very DRY. This PR instead seems like a much smaller change that isn't massively making things worse so that's why I'm suggesting this approach.

## How to review
I suggest reviewing commit by commit (but may not be necessary)

image

image

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -248,12 +248,9 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
         page_number = page if page else "1"
         content = base64.b64encode(pdf_file).decode('utf-8')
         content_outside_printable_area = metadata.get("message") == "content-outside-printable-area"
+        page_is_in_invalid_pages = page_number in metadata.get('invalid_pages', '[]')
 
-        show_overlay_for_page = False
-        if content_outside_printable_area and page_number in metadata.get('invalid_pages', '[]'):
-            show_overlay_for_page = True
-
-        if show_overlay_for_page or (content_outside_printable_area and file_type == "pdf"):
+        if content_outside_printable_area and (file_type == "pdf" or page_is_in_invalid_pages):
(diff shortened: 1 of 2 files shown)


app/template/rest.py
```diff
@@ -245,11 +245,15 @@ def preview_letter_template_by_notification_id(service_id, notification_id, file
...
         page_number = page if page else "1"
+        content = base64.b64encode(pdf_file).decode('utf-8')
+        content_outside_printable_area = metadata.get("message") == "content-outside-printable-area"
+
+        show_overlay_for_page = False
+        if content_outside_printable_area and page_number in metadata.get('invalid_pages', '[]'):
+            show_overlay_for_page = True
 
-        if overlay:
+        if show_overlay_for_page or (content_outside_printable_area and file_type == "pdf"):
```
idavidmcdonald: Oh, I'm open to suggestions on how to improve all the logic in this method as it's all fairly complex so if you can see a nicer way to reorganise everything then I'd like to hear about it

idavidmcdonald: Nice suggestion, I've implemented.

Agreed the whole function is gnarly. I think I'd like to potentially see it refactored so that pdf and pngs are just different routes but something I won't look at doing now.

PR #2747, opened 2020-03-11
Update postage db constraints for international letters

klssmith:
The `notifications`, `notification_history`, `templates` and `templates_history`
tables all had a check constraint on the postage column which specified
that the postage had to be `first` or `second` if the notification or
template was a letter. We now have two more options for postage -
`Europe` and `rest of world`.

It's not possible to alter a check constraint, so the constraints have
to be dropped then recreated. We are not recreating the constraint on
the `notification_history` table since values here are always copied
from the `notifications` table.

The constraints get added as `NOT VALID` at first - this stage will lock
the tables, so updating the `notification` table and `templates` and
`templates_history` are done in separate migrations so that we don't lock
all tables at the same time. In a third migration we then run
`VALIDATE CONSTRAINT` for all tables - this will lock a row at a time,
not the whole table.

https://www.pivotaltracker.com/story/show/171622825

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -903,7 +903,7 @@ class TemplateBase(db.Model):
     postage = db.Column(db.String, nullable=True)
     CheckConstraint("""
         CASE WHEN template_type = 'letter' THEN
-            postage is not null and postage in ('first', 'second')
+            postage is not null and postage in ('first', 'second', 'europe', 'rest-of-world')
         ELSE
             postage is null
         END
@@ -1352,10 +1352,13 @@ DVLA_RESPONSE_STATUS_SENT = 'Sent'
 
 FIRST_CLASS = 'first'
 SECOND_CLASS = 'second'
-POSTAGE_TYPES = [FIRST_CLASS, SECOND_CLASS]
+EUROPE = 'europe'
+REST_OF_WORLD = 'rest-of-world'
 RESOLVE_POSTAGE_FOR_FILE_NAME = {
     FIRST_CLASS: 1,
-    SECOND_CLASS: 2
+    SECOND_CLASS: 2,
+    EUROPE: 'E',
+    REST_OF_WORLD: 'N',
 }
 
 
@@ -1432,7 +1435,7 @@ class Notification(db.Model):
     postage = db.Column(db.String, nullable=True)
     CheckConstraint("""
         CASE WHEN notification_type = 'letter' THEN
-            postage is not null and postage in ('first', 'second')
+            postage is not null and postage in ('first', 'second', 'europe', 'rest-of-world')
         ELSE
             postage is null
         END
@@ -1697,13 +1700,6 @@ class NotificationHistory(db.Model, HistoryModel):
     created_by_id = db.Column(UUID(as_uuid=True), nullable=True)
 
     postage = db.Column(db.String, nullable=True)
-    CheckConstraint("""
-        CASE WHEN notification_type = 'letter' THEN
-            postage is not null and postage in ('first', 'second')
-        ELSE
-            postage is null
-        END
-    """)
 
     document_download_count = db.Column(db.Integer, nullable=True)
 
--- a/migrations/versions/0321_update_postage_constraint_1.py
+++ b/migrations/versions/0321_update_postage_constraint_1.py
@@ -1,0 +1,80 @@
+"""
+
+Revision ID: 0321_update_postage_constraint_1
+Revises: 0320_optimise_notifications
+Create Date: 2020-03-11 12:01:41.533192
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = '0321_update_postage_constraint_1'
+down_revision = '0320_optimise_notifications'
+
+
+def upgrade():
+    op.drop_constraint('chk_notifications_postage_null', 'notifications')
+    op.execute("""
+        ALTER TABLE notifications ADD CONSTRAINT "chk_notifications_postage_null"
+        CHECK (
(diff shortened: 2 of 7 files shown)


app/models.py
```diff
@@ -1351,7 +1351,8 @@ def check_code(self, cde):
 
 FIRST_CLASS = 'first'
 SECOND_CLASS = 'second'
-POSTAGE_TYPES = [FIRST_CLASS, SECOND_CLASS]
+INTERNATIONAL = 'international'
+POSTAGE_TYPES = [FIRST_CLASS, SECOND_CLASS, INTERNATIONAL]
```
servingUpAces: We need Europe and Non-Europe, instead of international. 
And we'll need to update `RESOLVE_POSTAGE_FOR_FILE_NAME={
 FIRST_CLASS:1, SECOND_CLASS: 2, EUROPE: 'E', NON_EUROPE:'N'
}`

migrations/versions/0318_update_postage_constraint.py
```diff
@@ -0,0 +1,76 @@
...
+        CHECK (
+            CASE WHEN notification_type = 'letter' THEN
+                postage is not null and postage in ('first', 'second', 'international')
+            ELSE
+                postage is null
+            END
+        )
+        NOT VALID
+    """)
+    op.execute("""
```
servingUpAces: I don't think we need to check constraint on `NOTIFICATION_HISTORY` because the data is copied from `NOTIFICATIONS` as long as that does have the constraint we should be alright.

app/models.py
quis: I went with `Europe` and `rest of world` here: https://github.com/alphagov/notifications-utils/pull/703/files#diff-0cdb242f4e660d4fe27f8b6bba754b52R60-R63

Would be nice to be consistent.

migrations/versions/0318_update_postage_constraint.py
klssmith: `notification_history` has already got a check constraint on the postage - I could just remove it instead of updating it. If we're going to remove it, then maybe we should also remove the constraint from `templates_history` to be consistent.

app/models.py
klssmith: I didn't realise we already had names for these in utils - I've updated this to use the same names

migrations/versions/0318_update_postage_constraint.py
klssmith: I've dropped the constraint from `notification_history` since the table is so large, but kept the constraint on the other 3 tables like we discussed

klssmith:
I've given this a bit of a complete rewrite since it was first reviewd and split it up into 3 different migration files:
1. Adds a constraint to `notifications` without validating it
2. Adds a constraint to `templates` and `templates_history` without validating them
3. Validates all 3 constraints

migrations/versions/0318_update_postage_constraint_1.py
```diff
@@ -0,0 +1,34 @@
...
+            ELSE
+                postage is null
+            END
+        )
+        NOT VALID
+    """)
+
+
+def downgrade():
+    pass
```
servingUpAces: Why no downgrade?

klssmith: Because although the change is reversible, I thought that ideally it shouldn't be downgraded in only 3 migrations since that might lock access to the tables for too long. This is because if we downgrade we'd need to add the constraint back to `notifications_history` too, which means (I think) we'd want 4 migrations to downgrade:

1. drop constraints from `templates` and `templates_history` and add the original constraints with `NOT VALID`
2. drop constraint from `notifications` and add the original constraint with `NOT VALID`
3. drop constraint from `notification_history` and add the original constraint with `NOT VALID`
4. validate all 4 constraints

Not sure if there's a way round this, or if it would be ok to somehow fit all of that into 3 migrations after all?

app/models.py
```diff
@@ -1352,10 +1352,13 @@ def check_code(self, cde):
 
 FIRST_CLASS = 'first'
 SECOND_CLASS = 'second'
-POSTAGE_TYPES = [FIRST_CLASS, SECOND_CLASS]
+EUROPE = 'Europe'
+REST_OF_WORLD = 'rest of world'
```
quis, line 1356: This should change (again sorry) to:
```suggestion
EUROPE = 'europe'
REST_OF_WORLD = 'rest-of-world'
```
Based on @leohemsted’s comment here: https://github.com/alphagov/notifications-utils/pull/703#pullrequestreview-385420470

klssmith: Thanks, I've updated the PR to use those new values

migrations/versions/0321_update_postage_constraint_1.py
```diff
@@ -0,0 +1,81 @@
...
+import sqlalchemy as sa
+
+
+revision = '0321_update_postage_constraint_1'
+down_revision = '0320_optimise_notifications'
+
+
+def upgrade():
+    op.drop_constraint('chk_notifications_postage_null', 'notifications')
+    op.drop_constraint('chk_notification_history_postage_null', 'notification_history')
```
servingUpAces, line 18: Let's try running this script on the copy-production-notify-db on staging. Be good to know how long it will take. 
We might also want to drop and create the constraint on notifications at the same time rather than after both drop.

klssmith: I've reordered the lines in this migration so that it drops the notification constraint, then creates the new notification constraint, and lastly drops the notification_history constraint. I'll also try and find a way to test this with copy-production-notify-db on staging.

PR #2748, opened 2020-03-12
Format postcode for CSV letter rows

idavidmcdonald:
Story ticket: https://www.pivotaltracker.com/story/show/171622677

We are formatting the postcode here, because if we did it in template
preview, that could break flows like API and admin one-off, since
we are not validating postcode there yet, and format_postcode
needs a nice validated postcode.

We are not doing it in admin, as then we would have to either
rewrite the CSV file or pass data differently to API. First would
be nasty, second is a lot of overhead.

In the long run we might want to move postcode formatting to
template preview so that the postcode in letter preview looks the same
before and after user sends it, but now to get it out quickly it's better
to format the postcode here in the task (if we did it in template preview now,
it would also try to format postcode for API flow and one-off flow, where it is
not being validated yet, and so if the postcode data is shorter than 4 characters
it would throw an exception, as postcode formatter relies on data being a real UK postcode)

Requires 
- https://github.com/alphagov/notifications-utils/pull/704
- a PR on admin be released which validates postcodes for CSV uploads

### DIFF
--- a/tests/app/celery/test_tasks.py
+++ b/tests/app/celery/test_tasks.py
@@ -961,7 +961,7 @@ def test_save_letter_saves_letter_to_database(mocker, notify_db_session):
         'addressline4': 'Wibble',
         'addressline5': 'Wobble',
         'addressline6': 'Wubble',
-        'postcode': 'Flob',
+        'postcode': 'SE1 2SA',
     }
     notification_json = _notification_json(
         template=job.template,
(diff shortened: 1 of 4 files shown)


tests/app/celery/test_tasks.py
```diff
@@ -1021,6 +1021,31 @@ def test_save_letter_saves_letter_to_database_with_correct_postage(mocker, notif
...
+def test_save_letter_saves_letter_to_database_with_formatted_postcode(mocker, notify_db_session):
+    service = create_service(service_permissions=[LETTER_TYPE])
+    template = create_template(service=service, template_type=LETTER_TYPE)
+    letter_job = create_job(template=template)
+
+    mocker.patch('app.celery.tasks.letters_pdf_tasks.create_letters_pdf.apply_async')
+    notification_json = _notification_json(
+        template=letter_job.template,
+        to='Foo',
+        personalisation={'addressline1': 'Foo', 'addressline2': 'Bar', 'postcode': 'se1 64sa'},
```
servingUpAces: This PR is fine, but it might be night to have a test that illustrates the different formatting that might happen, for instance adding the space

CrystalPea: good point, we have lots of cases tested in utils, where that logic lives, I hope that is sufficient for now.

PR #2751, opened 2020-03-13
Validate and format postcode for the API letter sending flow.

CrystalPea:
Validate and format postcode of a letter that we got through API POST request.

Story ticket: https://www.pivotaltracker.com/story/show/171622677

Should go in after: https://github.com/alphagov/notifications-api/pull/2748 as we are pulling in new version of utils there. Tests will fail until that PR is merged, too.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -271,7 +271,7 @@ def process_letter_notification(*, letter_data, api_key, template, reply_to_text
 
     postcode = letter_data['personalisation']['postcode']
     if not is_a_real_uk_postcode(postcode):
-        raise ValidationError(message='A real UK postcode is required.')
+        raise ValidationError(message='Must be a real UK postcode')
 
     test_key = api_key.key_type == KEY_TYPE_TEST
 
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -2,7 +2,9 @@ import base64
 import functools
 
 from flask import request, jsonify, current_app, abort
-from notifications_utils.recipients import try_validate_and_format_phone_number
+from notifications_utils.recipients import (
+    format_postcode_for_printing, is_a_real_uk_postcode, try_validate_and_format_phone_number
+)
 
 from app import api_user, authenticated_service, notify_celery, document_download_client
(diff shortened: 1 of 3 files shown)


app/v2/notifications/post_notifications.py
```diff
@@ -267,12 +269,18 @@ def process_letter_notification(*, letter_data, api_key, template, reply_to_text
                                                         template=template,
                                                         reply_to_text=reply_to_text)
 
+    postcode = letter_data['personalisation']['postcode']
+    if not is_a_real_uk_postcode(postcode):
+        raise ValidationError(message='A real UK postcode is required.')
```
servingUpAces: Has this message been reviewed by @karlchillmaid or @quis ?

CrystalPea: not yet, I will ask Chris for a review.

CrystalPea:
@servingUpAces if we merge this PR after https://github.com/alphagov/notifications-api/pull/2748 we will get the utils library update from there.

app/v2/notifications/post_notifications.py
quis: I’d suggest
```suggestion
        raise ValidationError(message='postcode must be a real UK postcode`')
```

This uses the same terminology as we’re using on the frontend:
image

And the tone of the message matches how we communicate other validation-type errors:
image

`postcode` should be lowercase because it’s referring to the name of the field in the JSON.

app/v2/notifications/post_notifications.py
```diff
@@ -271,7 +271,7 @@ def process_letter_notification(*, letter_data, api_key, template, reply_to_text
 
     postcode = letter_data['personalisation']['postcode']
     if not is_a_real_uk_postcode(postcode):
-        raise ValidationError(message='A real UK postcode is required.')
+        raise ValidationError(message='postcode is not a real UK postcode`')
```
karlchillmaid: Think we had a version that was ```Must be a real UK postcode```

CrystalPea: Should I change it to that then, it does sound clearer to me.

karlchillmaid: If @quis and the team are happy with that.

PR #2752, opened 2020-03-16
Degrade inbound number not associated with service message from error…

CrystalPea:
… to warning


CrystalPea:
This error actually proved useful this weekend: we learnt that an inbound number we got from mmg is already in use and avoided giving it to a service. Hence, we are keeping it as an error for now.

PR #2753, opened 2020-03-16
Increase the queue pool size to 30.

servingUpAces:
This should add an extra 1000 connections maximum, which will not tip us over the allowable 5000 db connections. And it may help with the queue pool connection errors.

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -2,7 +2,7 @@
   'notify-api': {
     'NOTIFY_APP_NAME': 'api',
     'disk_quota': '2G',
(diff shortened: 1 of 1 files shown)


idavidmcdonald:
I'm going to suggest we don't merge this (or maybe even close it) until we get some more evidence that this is needed.

We thought it may have been useful regarding queue pool exceptions but we may have found the root cause and this change may no longer be necessary. If it ain't broke, let's not try and fix it just in case this has an adverse affect.

idavidmcdonald:
Opening

PR #2757, opened 2020-03-17
Add a task to handle SMS responses

klssmith:
This moves the code from the `process_sms_client_response` function to a new `process_sms_client_response` task. The task will use a new queue, `sms-callbacks`, which the `delivery-worker-receipts` will watch.

The task essentially does the same things, with one or two minor differences, e.g. since the work is being done in a task, we can no longer return a success message from the `/notifications/sms/firetext` and `/notifications/sms/mmg` endpoints.

https://www.pivotaltracker.com/story/show/171791142

### DIFF
--- a/app/celery/process_sms_client_response_tasks.py
+++ b/app/celery/process_sms_client_response_tasks.py
@@ -1,20 +1,17 @@
 import uuid
-
 from datetime import datetime
+
 from flask import current_app
+from notifications_utils.statsd_decorators import statsd
 from notifications_utils.template import SMSMessageTemplate
 
-from app import statsd_client
+from app import notify_celery, statsd_client
 from app.clients import ClientException
-from app.dao import notifications_dao
 from app.clients.sms.firetext import get_firetext_responses
 from app.clients.sms.mmg import get_mmg_responses
-from app.celery.service_callback_tasks import (
-    send_delivery_status_to_service,
-    create_delivery_status_callback_data,
-)
+from app.celery.service_callback_tasks import send_delivery_status_to_service, create_delivery_status_callback_data
 from app.config import QueueNames
-from app.dao.notifications_dao import dao_update_notification
+from app.dao import notifications_dao
 from app.dao.service_callback_api_dao import get_service_delivery_status_callback_api_for_service
 from app.dao.templates_dao import dao_get_template_by_id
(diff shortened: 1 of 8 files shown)


app/celery/process_sms_client_response_tasks.py
```diff
@@ -1,78 +1,59 @@
...
-
-
-def process_sms_client_response(status, provider_reference, client_name):
-    success = None
-    errors = None
+@notify_celery.task(bind=True, name="process-sms-client-response", max_retries=5, default_retry_delay=300)
+@statsd(namespace="tasks")
+def process_sms_client_response(self, status, provider_reference, client_name):
     # validate reference
-    if provider_reference == 'send-sms-code':
```
klssmith: `provider_reference` should normally be the notification_id. It couldn't find anywhere (apart from the tests) where this function would get called with a provider reference of `send-sms-code` so removed it.

app/celery/process_sms_client_response_tasks.py
```diff
@@ -1,78 +1,59 @@
...
-        return success, errors
+    except ValueError as e:
+        current_app.logger.exception(f'{client_name} callback with invalid reference {provider_reference}')
+        raise e
 
-    try:
-        response_parser = sms_response_mapper[client_name]
-    except KeyError:
-        return success, 'unknown sms client: {}'.format(client_name)
+    response_parser = sms_response_mapper[client_name]
```
klssmith: I took away the exception handling round this line - I couldn't find an example of when we've received a response from an unknown client, and if that did happen in future we would be alerted by the error.

scripts/paas_app_wrapper.sh
```diff
@@ -43,7 +43,7 @@ case $NOTIFY_APP_NAME in
     ;;
   delivery-worker-receipts)
     exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=11 \
-    -Q ses-callbacks 2> /dev/null
+    -Q ses-callbacks,sms-callbacks 2> /dev/null
```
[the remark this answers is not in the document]
leohemsted: i think its fine to have the same worker - it'll pull both queues at the same time so will process from both queues at once. so if emails has a large backlog it won't affect sms.

we may need to scale it up (a little bit) though

PR #2759, opened 2020-03-18
Add an endpoint to delete a contact list

quis:
This was one of things we de-scoped when we first shipped this feature.

In order to safely delete a list, we first need to make sure any jobs aren’t referencing it.

Then we can remove the CSV file itself from S3.

### DIFF
--- a/migrations/versions/0319_contact_list_archived.py
+++ b/migrations/versions/0319_contact_list_archived.py
@@ -1,0 +1,26 @@
+"""
+
+Revision ID: 0319_contact_list_archived
+Revises: 0318_service_contact_list
+Create Date: 2020-03-26 11:16:12.389524
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+revision = '0319_contact_list_archived'
+down_revision = '0318_service_contact_list'
+
+
+def upgrade():
+    op.add_column(
+        'service_contact_list',
+        sa.Column('archived', sa.Boolean(), nullable=False, server_default=sa.false()),
+    )
+
+
+def downgrade():
+    op.drop_column(
+        'service_contact_list',
+        'archived',
+    )
--- a/tests/app/job/test_rest.py
+++ b/tests/app/job/test_rest.py
@@ -238,12 +238,21 @@ def test_create_scheduled_job(client, sample_template, mocker, fake_uuid):
     assert resp_json['data']['notification_count'] == 1
 
 
-def test_create_job_with_contact_list_id(client, mocker, sample_template, fake_uuid):
+@pytest.mark.parametrize('contact_list_archived', (
+    True, False,
+))
+def test_create_job_with_contact_list_id(
+    client,
+    mocker,
+    sample_template,
+    fake_uuid,
+    contact_list_archived,
+):
     mocker.patch('app.celery.tasks.process_job.apply_async')
     mocker.patch('app.job.rest.get_job_metadata_from_s3', return_value={
         'template_id': str(sample_template.id)
     })
-    contact_list = create_service_contact_list()
+    contact_list = create_service_contact_list(archived=contact_list_archived)
     data = {
         'id': fake_uuid,
         'valid': 'True',
--- a/tests/app/service/test_service_contact_list_rest.py
+++ b/tests/app/service/test_service_contact_list_rest.py
@@ -120,7 +120,8 @@ def test_dao_get_contact_list_by_id_does_not_return_if_contact_list_id_for_anoth
     assert response['message'] == "No result found"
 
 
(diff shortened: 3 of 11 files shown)


quis:
Archiving is probably better because we’re keeping a better record of who did what then.

It also addresses your second point which is that if we try to create a job referencing a contact list that’s just ben deleted I think it will cause problems at the moment.

quis:
@idavidmcdonald I‘ve changed it to archive in 594eab9c

quis:
The flow is:

### Choose the contact list to start the job with

You can’t get to this page if the contact list has been deleted

### Contact list is copied to the jobs bucket

Tiny, tiny, tiny possibility of an exception if someone presses the delete button exactly in between the contact list being loaded from the DB and the file being loaded from S3. Like you would have to press it in between these lines of code running: https://github.com/alphagov/notifications-admin/blob/e66f189b884d03e8787482c61ebc644dcc2a397b/app/main/views/send.py#L523-L531

### Job is created with file from the jobs bucket 

The job does not need a contact list at this point. It will record which contact list it was created from, but this will work whether or not the contact list is list archived.

***

TL;DR the job copies but doesn’t rely on the contact list, so the only possible error is during the copy, which I think in real timescales would be impossible.

migrations/versions/0319_contact_list_archived.py
```diff
@@ -0,0 +1,26 @@
...
+import sqlalchemy as sa
+
+revision = '0319_contact_list_archived'
+down_revision = '0318_service_contact_list'
+
+
+def upgrade():
+    op.add_column(
+        'service_contact_list',
+        sa.Column('archived', sa.Boolean(), nullable=False),
```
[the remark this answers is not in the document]
quis: Doesn’t seem to need it. I think because the default value is defined on the model SQLAlchemy takes care of it.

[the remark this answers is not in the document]
idavidmcdonald: If I add an emergency contact list to my local, then run the migrations it fails for me due to `sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) column "archived" contains null values`. I think you may still need it

idavidmcdonald:
Good context on the job stuff, cheers :)

migrations/versions/0319_contact_list_archived.py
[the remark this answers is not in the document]
quis: Ah yeah, that makes sense. Have changed and squashed now.

tests/app/job/test_rest.py
```diff
@@ -238,12 +238,21 @@ def test_create_scheduled_job(client, sample_template, mocker, fake_uuid):
...
+    sample_template,
+    fake_uuid,
+    contact_list_archived,
+):
     mocker.patch('app.celery.tasks.process_job.apply_async')
     mocker.patch('app.job.rest.get_job_metadata_from_s3', return_value={
         'template_id': str(sample_template.id)
     })
-    contact_list = create_service_contact_list()
+    contact_list = create_service_contact_list(archived=contact_list_archived)
```
idavidmcdonald: So this is saying that we are able to create a job with an archived contact list. Does that seem right? I would have assumed it wouldn't be allowed?

tests/app/service/test_service_contact_list_rest.py
```diff
@@ -113,3 +129,77 @@ def test_dao_get_contact_list_by_id_does_not_return_if_contact_list_id_for_anoth
...
+
+
+def test_archive_contact_list_by_id(mocker, admin_request, sample_service):
+    mock_s3 = mocker.patch('app.service.rest.s3.remove_contact_list_from_s3')
+    service_1 = create_service(service_name='Service under test')
+    template_1 = create_template(service=service_1)
+    expected_list = create_service_contact_list(service=service_1)
+    other_list = create_service_contact_list(service=service_1)
+
+    job_1 = create_job(template=template_1, contact_list_id=expected_list.id)
```
idavidmcdonald: Optional: Are all the jobs so this test is not just testing that you archive a contact list by ID but also testing that you can archive a contact list which has already been linked to a job? If so, would be nice to make that explicit with either a comment or splitting those into two tests as they are testing two different things

tests/app/job/test_rest.py
quis: Correct. I think this is OK. The same as if you schedule a job but then delete the template it doesn’t stop the job from going out.

tests/app/service/test_service_contact_list_rest.py
quis: Comment added in 14605764

PR #2760, opened 2020-03-19
Add a try/except around the code to get the files.

servingUpAces:
The idea is to log the exception but keep going. That way the "good" files still get sent and we can investigate why a file failed.

### DIFF
--- a/tests/app/celery/test_letters_pdf_tasks.py
+++ b/tests/app/celery/test_letters_pdf_tasks.py
@@ -315,6 +315,48 @@ def test_get_key_and_size_of_letters_to_be_sent_to_print(notify_api, mocker, sam
     ]
 
 
+@freeze_time('2020-02-17 18:00:00')
+def test_get_key_and_size_of_letters_to_be_sent_to_print_catches_exception(
+    notify_api, mocker, sample_letter_template
+):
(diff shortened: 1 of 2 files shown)


tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -315,6 +315,48 @@ def test_get_key_and_size_of_letters_to_be_sent_to_print(notify_api, mocker, sam
...
+
+    results = get_key_and_size_of_letters_to_be_sent_to_print(datetime.now() - timedelta(minutes=30))
+
+    assert mock_head_s3_object.call_count == 2
+    mock_head_s3_object.assert_has_calls(
+        [
+            call(current_app.config['LETTERS_PDF_BUCKET_NAME'], '2020-02-17/NOTIFY.REF1.D.2.C.C.20200217150000.PDF'),
+            call(current_app.config['LETTERS_PDF_BUCKET_NAME'], '2020-02-17/NOTIFY.REF0.D.2.C.C.20200217160000.PDF'),
+        ]
+    )
```
idavidmcdonald: Would be nice to also assert our log message for the error case but we can let that slide for the moment given time

PR #2762, opened 2020-03-19
Remove the join on NotificationHistory.

servingUpAces:
This is saying that we are confident that the NotificationHistory has been records before we delete the Notification.


CrystalPea:
Hmm... if we drop it and some data does go missing, we will know from statistics, right?
And then we could turn it back on.
Or we could make this change only for high volume services?

PR #2763, opened 2020-03-20
This option removes the subquery all together.

servingUpAces:
It has been recommended that subqueries are really inefficient especially on a delete statement.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -342,40 +342,12 @@ def delete_notifications_older_than_retention_by_type(notification_type, qry_lim
 
 
 def _delete_notifications(notification_type, date_to_delete_from, service_id, query_limit):
-    subquery = db.session.query(
-        Notification.id
-    ).join(NotificationHistory, NotificationHistory.id == Notification.id).filter(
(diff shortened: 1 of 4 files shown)


app/dao/notifications_dao.py
```diff
@@ -342,40 +342,12 @@ def delete_notifications_older_than_retention_by_type(notification_type, qry_lim
...
 def _delete_notifications(notification_type, date_to_delete_from, service_id, query_limit):
-    subquery = db.session.query(
-        Notification.id
-    ).join(NotificationHistory, NotificationHistory.id == Notification.id).filter(
+    deleted = Notification.query.filter(
         Notification.notification_type == notification_type,
         Notification.service_id == service_id,
         Notification.created_at < date_to_delete_from,
-    ).limit(query_limit).subquery()
+    ).delete()
```
idavidmcdonald: I'm not sure if we are OK to drop the query limit from here as we are using it further up?

From what I can understand, `delete_notifications_older_than_retention_by_type` needs to use a query limit and subquery everywhere or it doesn't use a query_limit at all?

I haven't got enough brain power to figure it out but I feel like there is something around this that needs looking at

PR #2767, opened 2020-03-23
Transaction to insert history and delete notifications.

servingUpAces:
Need to test the performance of this function, then we can call it from the task.
- Create a temporary table to insert ids of the desired rows, limit by 10K (might be too low).
- Insert into NotifcationHistory select from notification where id in temp table
- Delete from Notifications where id in temp table
- drop temp table

We should be able to iterate of this. The query stats for the query to create the temp table are very good, 17ms.

Next PR will call this code in the loop to delete/insert notifications.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -332,6 +332,63 @@ def delete_notifications_older_than_retention_by_type(notification_type, qry_lim
     return deleted
 
 
+@statsd(namespace="dao")
+@transactional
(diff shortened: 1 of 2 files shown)


risicle:
(You might find it useful to have a way of measuring how long the generation of the temp table takes vs the actual operation)

servingUpAces:
yes - statsd can be added in the next PR... was hoping to do some testing with this

PR #2768, opened 2020-03-24
Refactor delete_notifications_older_than_retention_by_type 

servingUpAces:
Refactor delete_notifications_older_than_retention_by_type to use a new strategy.

The insert_notification_history_delete_notifications function uses a temp table to store the data to insert and delete. This will save extra queries while performing the insert and delete operations.
The function is written in such a way that if the task is stop while processing when it's started up again it will just pick up where it left off.

I've made a decision to delete all test data in one query, I don't anticipate a problem with that.

The performance of this might also be better than last nights test because we are inserting everything we need for the NotificationHistory insert, so we don't need the join to Notifications to perform the insert.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -311,9 +311,11 @@ def delete_notifications_older_than_retention_by_type(notification_type, qry_lim
 
         day_to_delete_backwards_from = get_london_midnight_in_utc(
             convert_utc_to_bst(datetime.utcnow()).date()) - timedelta(days=f.days_of_retention)
-
-        deleted += _move_notifications_to_notification_history(
-            notification_type, f.service_id, day_to_delete_backwards_from, qry_limit)
+        delete_count_per_call = 1
+        while delete_count_per_call > 0:
+            delete_count_per_call = _move_notifications_to_notification_history(
+                notification_type, f.service_id, day_to_delete_backwards_from, qry_limit)
+            deleted += delete_count_per_call
 
     current_app.logger.info(
         'Deleting {} notifications for services without flexible data retention'.format(notification_type))
@@ -350,7 +352,7 @@ def insert_notification_history_delete_notifications(
           FROM notifications
         WHERE service_id = :service_id
           AND notification_type = :notification_type
-          AND created_at <= :timestamp_to_delete_backwards_from
+          AND created_at < :timestamp_to_delete_backwards_from
           AND key_type = 'normal'
           AND notification_status in ('delivered', 'permanent-failure', 'temporary-failure')
         limit :qry_limit
@@ -415,6 +417,7 @@ def _move_notifications_to_notification_history(notification_type, service_id, d
             qry_limit=qry_limit
         )
 
+    # Deleting test Notifications, test notifications are not persisted to NotificationHistory
     Notification.query.filter(
         Notification.notification_type == notification_type,
         Notification.service_id == service_id,
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -21,7 +21,6 @@ from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.sql import functions
 from sqlalchemy.sql.expression import case
-from sqlalchemy.dialects.postgresql import insert
 from werkzeug.datastructures import MultiDict
 
 from app import db, create_uuid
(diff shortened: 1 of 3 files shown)


app/dao/notifications_dao.py
```diff
@@ -335,31 +334,33 @@ def delete_notifications_older_than_retention_by_type(notification_type, qry_lim
...
     drop_table_if_exists = """
         DROP TABLE if exists NOTIFICATION_ARCHIVE
         """
     select_into_temp_table = """
          CREATE TEMP TABLE NOTIFICATION_ARCHIVE AS
-         SELECT id
+         SELECT id, job_id, job_row_number, service_id, template_id, template_version, api_key_id,
+             key_type, notification_type, created_at, sent_at, sent_by, updated_at, reference, billable_units,
+             client_reference, international, phone_prefix, rate_multiplier, notification_status,
+              created_by_id, postage, document_download_count
```
idavidmcdonald: Optional: If we were to add a new column to notification history we would always need to remember it also add it here. Would be nice to have this happen automatically but think it's not essential for this PR

app/dao/notifications_dao.py
```diff
@@ -406,103 +408,22 @@ def _move_notifications_to_notification_history(notification_type, service_id, d
...
-        Notification.service_id == service_id,
-        Notification.created_at < date_to_delete_from,
-    ).limit(query_limit).subquery()
-
-    deleted = _delete_for_query(subquery)
-
-    subquery_for_test_keys = db.session.query(
-        Notification.id
-    ).filter(
+    Notification.query.filter(
```
idavidmcdonald: Just a little code comment above here saying 'deleting test notifications' or similar would be useful, took me a second to work out that we weren't trying to do two delete statements after the after.

app/dao/notifications_dao.py
```diff
@@ -406,103 +408,22 @@ def _move_notifications_to_notification_history(notification_type, service_id, d
...
         Notification.notification_type == notification_type,
         Notification.service_id == service_id,
-        Notification.created_at < date_to_delete_from,
+        Notification.created_at < day_to_delete_backwards_from,
```
idavidmcdonald: Very minor but the delete above for the normal notifications is `<=` and this one for the test ones is just `<`

tests/app/dao/notification_dao/test_notification_dao_delete_notifications.py
```diff
@@ -157,28 +156,24 @@ def test_delete_notifications_inserts_notification_history(sample_service):
     assert NotificationHistory.query.count() == 2
 
 
-def test_delete_notifications_updates_notification_history(sample_email_template, mocker):
+def test_delete_notifications_does_nothing_if_notification_history_if_row_already_exists(
```
idavidmcdonald: Optional: This test name doesn't quite read correctly, should maybe be `test_delete_notifications_does_nothing_if_notification_already_exists_in_notification_history`

tests/app/dao/notification_dao/test_notification_dao_delete_notifications.py
```diff
@@ -357,29 +254,78 @@ def test_insert_notification_history_delete_notifications(sample_email_template)
...
+                        created_at=datetime.utcnow() + timedelta(minutes=30), status='temporary-failure')
+
+    del_count = insert_notification_history_delete_notifications(
+        notification_type=sample_template.template_type,
+        service_id=sample_template.service_id,
+        timestamp_to_delete_backwards_from=datetime.utcnow() + timedelta(hours=1),
+        qry_limit=1
+    )
+
+    assert del_count == 1
```
idavidmcdonald: I'm confused with this test. Am I right in thinking that are 3 notifications that should be moved but they would only do so one by one as qry_limit is one. `insert_notification_history_delete_notifications` will only delete maximum `qry_limit` meaning that if we want to move all our history, we need to call `insert_notification_history_delete_notifications` multiple times until it's got through them all? I don't see this happening anywhere?

app/dao/notifications_dao.py
servingUpAces: Yes - there is a way to just use the list of columns from a table. This code can be refactored later for things like that.

tests/app/dao/notification_dao/test_notification_dao_delete_notifications.py
servingUpAces: This method should adhere to the query limit... but there should be another test for `delete_notifications_older_than_retention_by_type` to make sure if notification count is greater than query limit it will delete them all, this `test_delete_notifications_calls_subquery_multiple_times` should test that. This test is poorly named.

app/dao/notifications_dao.py
servingUpAces: Yes - there is a way to just use the list of columns from a table. This code can be refactored later for things like that.

tests/app/dao/notification_dao/test_notification_dao_delete_notifications.py
idavidmcdonald: Strange, I can see the test `test_delete_notifications_calls_subquery_multiple_times` which looks like it's testing the right thing, however I can't find the corresponding logic in the app which actually makes it happen. Could you point me at it please?

servingUpAces: Ah... there is a small problem here. This test works because it iterates over the day hour by hour. But we need another loop to complete the work before moving onto the next hour. 
I've now added it: `app/dao/notifications_dao.py:315`

tests/app/dao/notification_dao/test_notification_dao_delete_notifications.py
```diff
@@ -222,10 +222,11 @@ def test_delete_notifications_does_try_to_delete_from_s3_when_letter_has_not_bee
     mock_get_s3.assert_not_called()
 
 
+@freeze_time('2020-03-25 00:01')
 def test_delete_notifications_calls_subquery_multiple_times(sample_template):
-    create_notification(template=sample_template, created_at=datetime.now() - timedelta(days=8), status='delivered')
-    create_notification(template=sample_template, created_at=datetime.now() - timedelta(days=8), status='delivered')
-    create_notification(template=sample_template, created_at=datetime.now() - timedelta(days=8), status='delivered')
+    create_notification(template=sample_template, created_at=datetime.now() - timedelta(days=7, minutes=3), status='delivered')
```
idavidmcdonald: I think now our original test would have passed without these changes (as we made the changes because of the hourly chunking). Happy either way

PR #2770, opened 2020-03-24
Increase default query limit to 50,000 from 10,000

servingUpAces:
The original intention was to increase the limit to 50K, which takes about 48seconds. 
10K rows takes 11 seconds.


idavidmcdonald:
5*11 = 55 seconds which isn't too disimilar to 48 seconds. Is the small gain in speed the only reason we are upping to 50k? Or are there other advantages?

servingUpAces:
An improvement is an improvement

PR #2771, opened 2020-03-25
Reduce the pressure on the db for API post email requests.

servingUpAces:
Instead of saving the email notification to the db add it to a queue to save later.
This is an attempt to alleviate pressure on the db from the api requests.
This initial PR is to trial it see if we see improvement in the api performance an a reduction in queue pool errors. If we are happy with this we could remove the hard coding of the service id.

In a nutshell:
 - If POST /v2/notification/email is from our high volume service (hard coded for now) then create a notification to send to a queue to persist the notification to the db.
 - create a save_api_email task to persist the notification
 - return the notification
 - New worker app to process the save_api_email tasks.


app/celery/tasks.py
```diff
@@ -293,6 +293,52 @@ def save_email(self,
...
+            queue=q
+        )
+        current_app.logger.info(f"Email {notification['id']} has been persisted.")
+    except IntegrityError:
+        current_app.logger.info(f"Email {notification['id']} already exists.")
+
+    except SQLAlchemyError as e:
+
+        try:
+            self.retry(queue=QueueNames.RETRY, exc=e)
```
klssmith: From looking at [these docs](https://docs.celeryproject.org/en/3.1/reference/celery.app.task.html#celery.app.task.Task.retry) it looks like the `exc` argument is used to specify what error should be raised when the task has retried the max number of times. I don't think we want to change the default from `MaxRetriesExceededError`, so don't need that argument here.

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -951,3 +952,82 @@ def test_post_email_notification_when_data_is_empty_returns_400(
...
+        headers=[('Content-Type', 'application/json'), create_authorization_header(service_id=service.id)]
+    )
+
+    json_resp = response.get_json()
+
+    assert response.status_code == 201
+    assert json_resp['id']
+    assert json_resp['content']['body'] == "Dear citizen, have a nice day"
+    assert json_resp['template']['id'] == str(template.id)
+    save_email_task.assert_called_once_with([mock.ANY], queue='save-api-email-tasks')
```
idavidmcdonald: Optional: Should we also add an assertion to show it's not saving it to the DB as well as adding it to the queue?

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -951,3 +952,82 @@ def test_post_email_notification_when_data_is_empty_returns_400(
...
+    )
+
+    json_resp = response.get_json()
+
+    assert response.status_code == 201
+    assert json_resp['id']
+    assert json_resp['content']['body'] == "Dear citizen, have a nice day"
+    assert json_resp['template']['id'] == str(template.id)
+    assert mocked_send_task.called
+    assert not save_email_task.called
```
idavidmcdonald: Optional: Should we check it is saving to the DB

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -951,3 +952,82 @@ def test_post_email_notification_when_data_is_empty_returns_400(
...
+        data=json.dumps(data),
+        headers=[('Content-Type', 'application/json'), create_authorization_header(service_id=service.id)]
+    )
+
+    json_resp = response.get_json()
+
+    assert response.status_code == 201
+    assert json_resp['id']
+    assert mocked_send_task.called
+    assert not save_email_task.called
```
idavidmcdonald: Optional: Same as above

app/celery/tasks.py
```diff
@@ -293,6 +293,52 @@ def save_email(self,
...
+    except IntegrityError:
+        current_app.logger.info(f"Email {notification['id']} already exists.")
+
+    except SQLAlchemyError:
+
+        try:
+            self.retry(queue=QueueNames.RETRY)
+        except self.MaxRetriesExceededError:
+            current_app.logger.error('Max retry failed' + f"Failed to persist notification {notification['id']}")
+
```
idavidmcdonald: What would happen if an exception is thrown in this method that isn't' caught by one of these two `except` statements? Would that notification be lost forever? Or does celery/SQS do something smart where it will realise it didn't  succeed and replay it again itself?

app/celery/tasks.py
```diff
@@ -293,6 +293,52 @@ def save_email(self,
...
+            queue=q
+        )
+        current_app.logger.info(f"Email {notification['id']} has been persisted.")
+    except IntegrityError:
+        current_app.logger.info(f"Email {notification['id']} already exists.")
+
+    except SQLAlchemyError:
+
+        try:
+            self.retry(queue=QueueNames.RETRY)
```
idavidmcdonald: Optional: Could test this behaviour

app/celery/tasks.py
leohemsted: Yeah, it would be lost forever

tests/app/v2/notifications/test_post_notifications.py
servingUpAces: It hasn't been saved to the db yet, it's been put on a queue to be saved to the db

tests/app/v2/notifications/test_post_notifications.py
servingUpAces: It hasn't been saved to the db yet, it's been put on a queue to be saved to the db

tests/app/v2/notifications/test_post_notifications.py
servingUpAces: It hasn't been saved to the db yet, it's been put on a queue to be saved to the db

tests/app/v2/notifications/test_post_notifications.py
idavidmcdonald: Sorry, to be clear, assert that it HASN'T been saved to the DB. As in we aren't trying to add it to both the DB and also the queue at the same time?

app/celery/tasks.py
idavidmcdonald: This is the question @servingUpAces, given what Leo says, is this an issue, should we do anything about it etc?

PR #2773, opened 2020-03-27
Move high volume config

servingUpAces:
To make the deployment and testing a little easier move the high volume service ids to the credential repo.

This way we can only add the ids when we are ready and all the infrastructure for the new service has been applied.


app/config.py
```diff
@@ -334,10 +332,11 @@ class Config(object):
     FIRETEXT_INBOUND_SMS_AUTH = json.loads(os.environ.get('FIRETEXT_INBOUND_SMS_AUTH', '[]'))
     MMG_INBOUND_SMS_AUTH = json.loads(os.environ.get('MMG_INBOUND_SMS_AUTH', '[]'))
     MMG_INBOUND_SMS_USERNAME = json.loads(os.environ.get('MMG_INBOUND_SMS_USERNAME', '[]'))
-
     ROUTE_SECRET_KEY_1 = os.environ.get('ROUTE_SECRET_KEY_1', '')
     ROUTE_SECRET_KEY_2 = os.environ.get('ROUTE_SECRET_KEY_2', '')
 
+    HIGH_VOLUME_SERVICE = json.loads(os.environ.get('HIGH_VOLUME_SERVICE', '[]'))
```
idavidmcdonald: Optional, make this key plural `HIGH_VOLUME_SERVICES` given it's now a list

idavidmcdonald: This also needs a corresponding change in the manifest file

PR #2774, opened 2020-03-27
Fix report and delete tasks

servingUpAces:
Delete notifications for all notification_status types.

This will ensure the whole day has been deleted. The stats table could get the wrong updates if there is partial data for a day.
Why?
If there is partial data for a day the reporting tasks for a service can be off. This corrects that by deleting an entire day. 

Also in the PR is a refactor of process_ses_receipt.
Why?
It is possible a service has data retention that is smaller than the time it takes to get a delivery receipt.
This PR refactors process_ses_receipt to update NotificationHistory if the Notification has already been purged.

Also removed some logging, I don't recall ever looking for `SES delivery failed: notification id` log message. It doesn't tell me anything that I can't get from the db.
Also removed a duplicate log in `tasks.save_api_email`

### DIFF
--- a/app/celery/process_ses_receipts_tasks.py
+++ b/app/celery/process_ses_receipts_tasks.py
@@ -39,7 +39,7 @@ def process_ses_results(self, response):
         reference = ses_message['mail']['messageId']
 
         try:
-            notification = notifications_dao.dao_get_notification_history_by_reference(reference=reference)
+            notification = notifications_dao.dao_get_notification_or_history_by_reference(reference=reference)
         except NoResultFound:
             message_time = iso8601.parse_date(ses_message['mail']['timestamp']).replace(tzinfo=None)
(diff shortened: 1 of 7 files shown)


app/celery/process_ses_receipts_tasks.py
```diff
@@ -39,7 +39,7 @@ def process_ses_results(self, response):
         reference = ses_message['mail']['messageId']
 
         try:
-            notification = notifications_dao.dao_get_notification_by_reference(reference)
+            notification = notifications_dao.dao_get_notification_history_by_reference(reference=reference)
```
leohemsted: optional: is it possible for us to rename this function? took me quite a while to realise that it checks notifications first, the notification_history. i'd prefer something like `dao_get_notification_or_history_by_reference`

PR #2775, opened 2020-03-27
Add retries to process_sanitised_letter task

klssmith:
This task didn't have retries before, based on the assumption that if
the task failed it was likely to be due to a Boto error, so retrying
would not help because a file was probably not in the expected bucket.
During an incident with the database, we had some letters that were
stuck in the `pending-virus-check` state because this task failed.

This change adds retries to the task if there was an Exception that was
not a `BotoClientError`.

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -288,7 +288,7 @@ def sanitise_letter(self, filename):
         current_app.logger.info('Notification ID {} Virus scan passed: {}'.format(notification.id, filename))
 
         if notification.status != NOTIFICATION_PENDING_VIRUS_CHECK:
-            current_app.logger.info('Sanitise letter called for notification {} which has is in {} state'.format(
+            current_app.logger.info('Sanitise letter called for notification {} which is in {} state'.format(
                 notification.id, notification.status))
             return
 
@@ -314,8 +314,8 @@ def sanitise_letter(self, filename):
             raise NotificationTechnicalFailureException(message)
 
 
-@notify_celery.task(name='process-sanitised-letter')
-def process_sanitised_letter(sanitise_data):
+@notify_celery.task(bind=True, name='process-sanitised-letter', max_retries=15, default_retry_delay=300)
+def process_sanitised_letter(self, sanitise_data):
     letter_details = encryption.decrypt(sanitise_data)
 
     filename = letter_details['filename']
@@ -326,7 +326,7 @@ def process_sanitised_letter(sanitise_data):
 
     if notification.status != NOTIFICATION_PENDING_VIRUS_CHECK:
         current_app.logger.info(
-            'process-sanitised-letter task called for notification {} which has is in {} state'.format(
+            'process-sanitised-letter task called for notification {} which is in {} state'.format(
                 notification.id, notification.status)
         )
         return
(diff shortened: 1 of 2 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -354,22 +354,36 @@ def process_sanitised_letter(sanitise_data):
         billable_units = get_billable_units_for_letter_page_count(letter_details['page_count'])
         is_test_key = notification.key_type == KEY_TYPE_TEST
 
-        move_sanitised_letter_to_test_or_live_pdf_bucket(filename, is_test_key, notification.created_at)
-        # We've moved the sanitised PDF from the sanitise bucket, but still need to delete the original file:
-        original_pdf_object.delete()
+        # Updating the notification needs to happen before the file is moved. This is so that if updating the
+        # notification fails, the task can retry because the file is in the same place.
```
idavidmcdonald: Glad you've given this stuff about ordering some thought.

So if updating the notification status fails, the task will rerun and everything should be able to run successfully the second time.

If the `move_sanitised_letter_to_test_or_live_pdf_bucket` failed due to a boto error then the task would be marked as technical failure. If it failed due to a different type of exception, would this task fail to run for the second time as the notification status would now be in delivered as opposed to `NOTIFICATION_PENDING_VIRUS_CHECK`?

Similarly, if the `delete` failed due to boto then the task would be marked as technical failure. But if it failed due to a different type of exception, then this task would again fail to run a second time?

Simiarly, if the app got killed mid statement, between the updating of the status and the moving and deleting, would we have a notification in an incorrect state? However, I think this is handled by celery allowing 10 seconds for a task to finish processing before killing the app.

So the above 2 problem cases make me a tiny bit nervous but I think the risk of us encountering them is also pretty low so I might be willing to accept the risk. Some my above assumptions could also be wrong as I haven't spent that much time with Celery. What do you think?

I think maybe ideally what this task would do, but this may be something if we have more time, is something like:

check if the file has been moved to the correct s3 bucket. If it has, continue, if it hasn't, then move it
Check if the original file has been deleted. If it has continue, if it hasn't, then delete it
If both of those two checks have succeeded, then mark the notification to the relevant status.

app/celery/letters_pdf_tasks.py
```diff
@@ -354,22 +354,36 @@ def process_sanitised_letter(sanitise_data):
...
             status=NOTIFICATION_DELIVERED if is_test_key else NOTIFICATION_CREATED,
             billable_units=billable_units,
             recipient_address=letter_details['address']
         )
+        move_sanitised_letter_to_test_or_live_pdf_bucket(filename, is_test_key, notification.created_at)
+        # We've moved the sanitised PDF from the sanitise bucket, but still need to delete the original file:
+        original_pdf_object.delete()
 
     except BotoClientError:
         current_app.logger.exception(
```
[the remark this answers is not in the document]
klssmith: 👍comment added

app/celery/letters_pdf_tasks.py
```diff
@@ -326,7 +326,7 @@ def process_sanitised_letter(sanitise_data):
 
     if notification.status != NOTIFICATION_PENDING_VIRUS_CHECK:
```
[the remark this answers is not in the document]
klssmith: Although the status of letters at this point is `pending-virus-check`, they have actually been virus scanned by antivirus and sanitised by template-preview. This is the last task they go through - in this task, the file gets moved and the letter will be given the status of `created`.

app/celery/letters_pdf_tasks.py
klssmith: I'm not too concerned about the two potential problem cases, just because I think they are both very unlikely.

If either `move_sanitised_letter_to_test_or_live_pdf_bucket` or `original_pdf_object.delete()` fail due to a non-Boto error, then things will end up in a bad state - the notification will be `created` but the files won't have been moved / deleted. But I think this is low risk - apart from Boto stuff, there's only 1 very simple function that gets called, and `ClientError` is one of Boto's base exceptions, so the chance of seeing another error class is low. If that did occur, we would see the error in the logs and would then have to go and manually see what had happened.

Likewise, the chance of the app being killed when the code was between those two statements _and_ the task then taking over 10 seconds doesn't seem very high. I don't think we've really covered for that possibility in any of our tasks - most of them would do unexpected things if that happened.

Just to double check I understand correctly, what's the benefit of changing the task to do your suggestion of checking S3 before moving / deleting the file and then updating the notification, or what error is that trying to prevent?

app/celery/letters_pdf_tasks.py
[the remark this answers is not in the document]
idavidmcdonald: Thanks for that. Could be worth another code comment around the check of the status being pending-virus-scan.

app/celery/letters_pdf_tasks.py
idavidmcdonald: Cool, I'm happy then. Thanks for that.

It just means that the task is always able to replay regardless of if it got into a bad state. The task essentially is saying - 'what should I look like after this task has run, what have I still got to do to get there, OK, let me do that'. The task will always end up with the same result and can be repeated as much as we want.

klssmith: Great, I'll keep that change in mind for a future enhancement then.

PR #2778, opened 2020-03-30
Cache template when posting a notification

quis:
This commit proposes two types of caching to speed up the post notification processing time.

Since every notification needs a template, if we can avoid going to the database to get the template then we should save some time.

The first cache is an LRU cache, which means that, for a given `template_id`, `service_id` and `version` the app will go to memory rather than the database to get the template. This should be much faster, and we should be guaranteed that a combination of those three parameters should always return the same result.

But, we don’t know the template version at the time of making the call. So this commit also adds another layer of caching to find the most recent version number of a template. Since this can be changed by other instances we need to cache it externally, so this cache uses Redis. Redis will be a lot slower than going to memory, but hopefully quicker than going to the database. By only caching the minimal amount of data possible in Redis (just the version as a number) we’re hopefully minimising the performance hit from going to an external service.

***

Questions:
- will this work?
- will this make things faster?
- is this massive increase in traffic to Redis a problem?

### DIFF
--- a/app/dao/templates_dao.py
+++ b/app/dao/templates_dao.py
@@ -1,10 +1,12 @@
 from datetime import datetime
+from functools import lru_cache
 import uuid
 
 from flask import current_app
-from sqlalchemy import asc, desc
+from sqlalchemy import asc, desc, func
+from sqlalchemy.orm.exc import NoResultFound
 
-from app import db
+from app import db, redis_store
 from app.models import (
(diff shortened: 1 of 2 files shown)


quis:
Closing since we’re not going to do this. Discussion continued on this ticket: https://www.pivotaltracker.com/story/show/172050089

PR #2780, opened 2020-03-30
fix purge functional test data task

leohemsted:
* it doesn't delete service email reply to or letter contacts, or contact lists. I don't think the contact lists will ever be an issue but it doesn't hurt to add it to the list of things to remove.
* it doesn't remove users from organisations before deleting the users

there may be more tables that link to Service that should be deleted, but for now just add these ones that I could spot.

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -106,7 +106,7 @@ def purge_functional_test_data(user_email_prefix):
     """
     Remove non-seeded functional test data
 
-    users, services, etc. Give an email prefix. Probably "notify-test-preview".
+    users, services, etc. Give an email prefix. Probably "notify-tests-preview".
     """
     users = User.query.filter(User.email_address.like("{}%".format(user_email_prefix))).all()
     for usr in users:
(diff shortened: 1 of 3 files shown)


app/commands.py
```diff
@@ -106,7 +106,7 @@ def purge_functional_test_data(user_email_prefix):
     """
     Remove non-seeded functional test data
 
-    users, services, etc. Give an email prefix. Probably "notify-test-preview".
+    users, services, etc. Give an email prefix. Probably "notify-tests-preview".
```
CrystalPea: Do we need the word "Probably" here?

leohemsted: yup. you _might_ want to run it on something else, for example if you were running from your local but pointed at preview

leohemsted: but most of the time it's (infrequently) used is to clear out data cos theres too much and we weren't anticipating the weird data. in this case, preview has 2705 org users and that's quite high (prod has 17 😂 ).

CrystalPea: Ahh, cheers, I get it now ☺️

PR #2784, opened 2020-03-31
Add team key to query

servingUpAces:
Add team key to the notifications to delete/insert into notifications/notification_history

notifications created with team keys have not been moved and purged.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -351,7 +351,7 @@ def insert_notification_history_delete_notifications(
         WHERE service_id = :service_id
           AND notification_type = :notification_type
           AND created_at < :timestamp_to_delete_backwards_from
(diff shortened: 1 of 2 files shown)


app/dao/notifications_dao.py
```diff
@@ -351,7 +351,7 @@ def insert_notification_history_delete_notifications(
         WHERE service_id = :service_id
           AND notification_type = :notification_type
           AND created_at < :timestamp_to_delete_backwards_from
-          AND key_type = 'normal'
+          AND key_type in ('normal', 'team')
```
idavidmcdonald: Optional: This is fine logically. Did you do any investigation to performance for using an `in` vs using an `OR`. I wouldn't be surprised if the query plan for the `in` got changed to an `OR` anyway and made no difference but just thought I'd check.

PR #2785, opened 2020-03-31
Optimise indexes

servingUpAces:
Dropping unnecessary indexes on Notification and Notification_History
Create new composite indexes that should give queries better performance
Drop and create new indexes.

We are not running this migration on production because we want to run each one at a time.
Run this on staging then start run some load tests

### DIFF
--- a/migrations/versions/0320_optimise_notifications.py
+++ b/migrations/versions/0320_optimise_notifications.py
@@ -1,0 +1,79 @@
+"""
+
+Revision ID: 0320_optimise_notifications
+Revises: 0319_contact_list_archived
+Create Date: 2020-03-26 11:16:12.389524
+
+"""
+import os
+
+from alembic import op
+
+revision = '0320_optimise_notifications'
+down_revision = '0319_contact_list_archived'
+environment = os.environ['NOTIFY_ENVIRONMENT']
+
+
+def upgrade():
+    # We like to run this operation on live via the command prompt, to watch the progress and stop if necessary
+    if environment not in ["live", "production"]:
+        # Drop indexes notifications - no need to recreate these indexes
+        op.execute('DROP INDEX IF EXISTS ix_notifications_key_type')
+        op.execute('DROP INDEX IF EXISTS ix_notifications_api_key_id')
+        op.execute('DROP INDEX IF EXISTS ix_notifications_notification_status')
+        op.execute('DROP INDEX IF EXISTS ix_notifications_notification_type')
+        op.execute('DROP INDEX IF EXISTS ix_notifications_service_id')
+
+        # Create new composite indexes
+        # PLEASE NOTE: that if you create index on production you need to add concurrently to the create statement,
+        # however we are unable to do that inside a transaction like this upgrade method
+        ix_notifications_service_id_composite = """
+        CREATE INDEX ix_notifications_service_id_composite 
+            on notifications (service_id, notification_type, notification_status, created_at)
+        """
+        op.execute(ix_notifications_service_id_composite)
+
+        ix_notifications_notification_type_composite = """
+            CREATE INDEX ix_notifications_notification_type_composite 
+            on notifications (notification_type, notification_status, created_at)
+        """
+        op.execute(ix_notifications_notification_type_composite)
+        # DROP and CREATE all other indexes
+        op.execute('DROP INDEX IF EXISTS ix_notifications_service_id_created_at')
+        op.execute(
+            'CREATE INDEX ix_notifications_service_id_created_at ON NOTIFICATIONS (service_id, created_at)'
+        )
+
+        op.execute('DROP INDEX IF EXISTS ix_notifications_client_reference')
+        op.execute('CREATE INDEX ix_notifications_client_reference on notifications (client_reference)')
+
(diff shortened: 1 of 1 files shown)


migrations/versions/0320_optimise_notifications.py
```diff
@@ -0,0 +1,79 @@
...
+            on notifications (service_id, notification_type, notification_status, created_at)
+        """
+        op.execute(ix_notifications_service_id_composite)
+
+        ix_notifications_notification_type_composite = """
+            CREATE INDEX ix_notifications_notification_type_composite 
+            on notifications (notification_type, notification_status, created_at)
+        """
+        op.execute(ix_notifications_notification_type_composite)
+        # DROP and CREATE all other indexes
```
leohemsted: why are we doing this?

migrations/versions/0320_optimise_notifications.py
```diff
@@ -0,0 +1,79 @@
...
+        op.execute('DROP INDEX IF EXISTS ix_notifications_service_id_created_at')
+        op.execute(
+            'CREATE INDEX ix_notifications_service_id_created_at ON NOTIFICATIONS (service_id, created_at)'
+        )
+
+        op.execute('DROP INDEX IF EXISTS ix_notifications_client_reference')
+        op.execute('CREATE INDEX ix_notifications_client_reference on notifications (client_reference)')
+
+        op.execute('DROP INDEX IF EXISTS ix_notifications_client_reference')
```
idavidmcdonald: This appears to be a repeat of the above two lines?

migrations/versions/0320_optimise_notifications.py
```diff
@@ -0,0 +1,79 @@
...
+
+        op.execute('DROP INDEX IF EXISTS ix_notifications_job_id')
+        op.execute('CREATE INDEX ix_notifications_job_id ON notifications (job_id)')
+
+        op.execute('DROP INDEX IF EXISTS ix_notifications_reference')
+        op.execute('CREATE INDEX ix_notifications_reference ON notifications (reference)')
+
+        op.execute('DROP INDEX IF EXISTS ix_notifications_service_created_at')
+        op.execute(
+            'CREATE INDEX ix_notifications_service_created_at ON notifications (service_id, created_at)')
```
idavidmcdonald: `CREATE INDEX ix_notifications_service_created_at ON notifications (service_id, created_at)` this is the same as `ix_notifications_service_id_created_at` above apart from the name. Is it intentional that we have both?

migrations/versions/0320_optimise_notifications.py
```diff
@@ -0,0 +1,79 @@
...
+        op.execute('DROP INDEX IF EXISTS ix_notifications_template_id')
+        op.execute('CREATE INDEX ix_notifications_template_id ON notifications (template_id)')
+
+        # Drop indexes notification_history
+        op.execute('DROP INDEX IF EXISTS ix_notification_history_key_type')
+        op.execute('DROP INDEX IF EXISTS ix_notification_history_api_key_id')
+        op.execute('DROP INDEX IF EXISTS ix_notification_history_notification_status')
+        op.execute('DROP INDEX IF EXISTS ix_notification_history_notification_type')
+        op.execute('DROP INDEX IF EXISTS ix_notification_history_service_id')
+        op.execute('DROP INDEX IF EXISTS ix_notification_history_week_created')
```
leohemsted: have you tested the nightly tasks? they'll use notification history for anything with 3 days retention or less (ie both govuk and nhs coronavirus sms), as they both update for last four days.

are we worried that they might be super slow after this?

migrations/versions/0320_optimise_notifications.py
servingUpAces: It was from a query on the list of bad performing queries... I thought it was for the platform-admin/summary page. But it's actually used for `timeout_notifications`

migrations/versions/0320_optimise_notifications.py
servingUpAces: Typo - I can remove it.

migrations/versions/0320_optimise_notifications.py
servingUpAces: That task has taken up to 15 minutes before - we think it's worth it

leohemsted: no i mean, why are we dropping and recreating all the other existing indexes?

migrations/versions/0320_optimise_notifications.py
servingUpAces: These indexes are not used for that query. We found that just deleting the unused indexes helps performance on the notification_history table.

servingUpAces:
We have a running document with recorded query plans and reasons for making decisions. Which will be shared. 
The only reason to have an alembic script is to record the changes and run it on other environments.

migrations/versions/0320_optimise_notifications.py
leohemsted: answered via slack:

> So by dropping an index we saw a decrease in 10GB used for a single 
> we are doing the drop and create index concurrently to clean the shared buffers
> when we re-created it it only took 1GB

Also of note: REINDEX holds an exclusive table lock, while this can be recreated concurrently

PR #2788, opened 2020-04-01
Force sequence of our related nightly tasks

idavidmcdonald:
https://www.pivotaltracker.com/story/show/172062501

We previously sequenced the
`timeout_notifications`, `create_nightly_notification_status` and
`delete_<notification_type>_notifications` tasks by setting
them up with gaps in crontab. This means they would run in
order and likely not conflict however not guaranteed
given that one of them might run for much longer and conflict with the
next task.

If tasks were to conflict, this may cause problems. For example, if
notifications aren't all timed out before creating the
notification_stats data, then the data will be incorrect for the given
day. If that is the case for an SMS retention of 3 days, given we
calculate the past 4 days of stats, we would not naturally recalculate
the stats for the last day if they had incorrect data due to the above
clash. Simiarly, the `delete-email-notifications` and
`delete-sms-notifications` tasks could clash and slow each other down,
we believe because they are both doing a lot of writes to the same
index.

To solve this we also considered using celery chaining, see
http://docs.celeryproject.org/en/latest/userguide/canvas.html#chains.
However the main benefit of chaining is the ability to pass the return value
of the previous task into the next task. We don't need this behaviour.
Apart from that, if an app crashed during an execution of chained tasks,
it would not automatically restart from where it left off. It would
require a human to spot this and kick it off again manually. This is the
same for the approach that we've taken. Therefore we went for the simplest
option of just writing a simple function instead of trying chaining.

I checked to see whether `create-nightly-billing` and
`send-daily-performance-platform-stats` need to be run in sequence too.
It appears not so they can just be run at any time during the middle of
the night. Neither depend on notification_status being in temp failure
rather than sending (which would be done by the timeout task).

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -11,6 +11,7 @@ from sqlalchemy.exc import SQLAlchemyError
 
 from app import notify_celery, performance_platform_client, zendesk_client
 from app.aws import s3
+from app.celery.reporting_tasks import create_nightly_notification_status
 from app.celery.service_callback_tasks import (
     send_delivery_status_to_service,
     create_delivery_status_callback_data,
@@ -120,6 +121,24 @@ def delete_letter_notifications_older_than_retention():
     except SQLAlchemyError:
         current_app.logger.exception("Failed to delete letter notifications")
         raise
+
+
+@notify_celery.task(name="clean-notifications-table-and-create-stats")
+@statsd(namespace="tasks")
+def clean_notifications_table_and_create_stats():
+    """
+    A group of functions to run sequentially as their order matters
+
+    We timeout any notifications that are still in 'sending'. When that is done we can then create
+    the rows in the notification_status table using the data from the notifications table.
+    Then we have finished interpreting the notifications table and can move notifications into the
+    notification_history table and delete the originals from `notifications`.
+    """
+    timeout_notifications()
+    create_nightly_notification_status()
+    delete_email_notifications_older_than_retention()
+    delete_sms_notifications_older_than_retention()
+    delete_letter_notifications_older_than_retention()
 
 
 @notify_celery.task(name='timeout-sending-notifications')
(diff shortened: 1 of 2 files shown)


app/celery/nightly_tasks.py
```diff
@@ -122,6 +123,24 @@ def delete_letter_notifications_older_than_retention():
...
 
 
+@notify_celery.task(name="clean-notifications-table-and-create-stats")
+@statsd(namespace="tasks")
+def clean_notifications_table_and_create_stats():
+    """
+    A group of functions to run sequentially as their order matters
+
+    We timeout any notifications that are still in 'sending'. When that is done we can then create
+    the rows in the notification_status table using the data from the notifications table.
```
leohemsted, line 133: does the solution also need to apply to `create_nightly_billing`? it's coded the same way as `create_nightly_notification_status` (ie: spawning more async tasks).

It doesn't depend on notifications having timed out, but if the delete task runs part way through it, then there could be issues as it could miss notifications that get moved to notification history while it's running

app/celery/nightly_tasks.py
```diff
@@ -122,6 +123,24 @@ def delete_letter_notifications_older_than_retention():
...
+    Then we have finished interpreting the notifications table and can move notifications into the
+    notification_history table and delete the originals from `notifications`.
+    """
+    timeout_notifications()
+    create_nightly_notification_status()
```
leohemsted, line 138: `create_nightly_notification_status` spawns a load of `create_nightly_notification_status_for_day` tasks. This new task won't control when those tasks run unless the entire thing is triggered after the delete tasks finish.

Those tasks all call `get_notification_table_to_use` to figure out if notifications have been deleted yet. That function has a flag for if the delete tasks have already run today so we can use that

servingUpAces:
We could just start with sequentially running the delete-<notification-type>-notifications tasks. 
And schedule those to run earlier. Instead of starting the delete tasks at 04:15, we could start at 03:00

idavidmcdonald:
Going to close this as we've merged the sequential running of the deleting tasks. We can turn our attention back to this approach if we need to do more work on the running order.

PR #2793, opened 2020-04-03
Allow any three lines in the address for templated letters

quis:
This is part of moving away from `postcode` and towards `address line 7`

We think it will be easier for people to map their existing data to our API if we let them fill in any 3 lines, instead of requiring at least 1, 2, and postcode specifically.

***

This pull request also removes the reformatting of the postcode. We’re going to let template preview handle this when it’s rendering the PDF.

It’s better (and less code) to store what people give us, so we can give them back the same thing.

***

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/713
- [x] https://github.com/alphagov/notifications-template-preview/pull/438

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -3,10 +3,7 @@ from datetime import datetime
 from collections import namedtuple, defaultdict
 
 from flask import current_app
-from notifications_utils.recipients import (
-    format_postcode_for_printing,
-    RecipientCSV
-)
+from notifications_utils.recipients import RecipientCSV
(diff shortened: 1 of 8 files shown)


quis:
Not clear from this story about when exactly it should happen: https://www.pivotaltracker.com/story/show/171622940

I guess there are three changes we need to communicate:
- validation of postcode
- move from 1, 2 and postcode to any 7 lines
- allowing country

I think the initial reckon was that we’d bundle them all into one docs change, so that we don’t keep moving the goalposts.

PR #2795, opened 2020-04-06
Look in all parts of a letter template to find placeholders

quis:
Text messages have placeholders in their body.

Emails have them in their subject line too.

Letters have them in their body, subject line and contact block.

We were only looking in the the body and subject when processing a job, therefore the thing assembling the letter was not looking in all the CSV columns it needed to, because it hadn’t been told about any placeholders in the contact block.

Fixing this means always making sure we use the correct `Template` instance for the type of template we’re dealing with. Which we were already doing in a different part of the codebase. So it makes sense to reuse that.

Turns out we fixed the same bug for email subjects over 3 years ago: 3ed9715

---

This commit also updates the API to check the contact block when validating the placeholders for creating a notification or preview of a template. 

---

Fixes https://www.pivotaltracker.com/story/show/172171068

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -983,8 +983,13 @@ class TemplateBase(db.Model):
         if self.template_type == LETTER_TYPE:
             return LetterPrintTemplate(
                 self.__dict__,
-                contact_block=self.service.get_default_letter_contact(),
+                contact_block=self.get_reply_to_text(),
             )
+
+    def _as_utils_template_with_personalisation(self, values):
+        template = self._as_utils_template()
+        template.values = values
+        return template
 
     def serialize(self):
         serialized = {
--- a/app/models.py
+++ b/app/models.py
@@ -977,16 +977,12 @@ class TemplateBase(db.Model):
 
     def _as_utils_template(self):
         if self.template_type == EMAIL_TYPE:
-            return PlainTextEmailTemplate(
-                {'content': self.content, 'subject': self.subject}
-            )
+            return PlainTextEmailTemplate(self.__dict__)
         if self.template_type == SMS_TYPE:
-            return SMSMessageTemplate(
-                {'content': self.content}
-            )
+            return SMSMessageTemplate(self.__dict__)
         if self.template_type == LETTER_TYPE:
             return LetterPrintTemplate(
-                {'content': self.content, 'subject': self.subject},
+                self.__dict__,
                 contact_block=self.service.get_default_letter_contact(),
             )
 
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -32,11 +32,10 @@ from app.dao.notifications_dao import (
 )
 
 from app.v2.errors import BadRequestError
-from app.utils import get_template_instance
 
 
 def create_content_for_notification(template, personalisation):
-    template_object = get_template_instance(template.__dict__, personalisation)
+    template_object = template._as_utils_template_with_personalisation(personalisation)
     check_placeholders(template_object)
 
     return template_object
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -8,6 +8,7 @@ from flask import request, jsonify, current_app, abort
 from notifications_utils.recipients import (
     format_postcode_for_printing, is_a_real_uk_postcode, try_validate_and_format_phone_number
 )
+from notifications_utils.template import WithSubjectTemplate
 
 from app import (
     api_user,
@@ -166,23 +167,25 @@ def post_notification(notification_type):
     if notification_type == SMS_TYPE:
         create_resp_partial = functools.partial(
             create_post_sms_response_from_notification,
-            from_number=reply_to
+            from_number=reply_to,
+            content=str(template_with_content),
         )
     elif notification_type == EMAIL_TYPE:
         create_resp_partial = functools.partial(
             create_post_email_response_from_notification,
             subject=template_with_content.subject,
-            email_from='{}@{}'.format(authenticated_service.email_from, current_app.config['NOTIFY_EMAIL_DOMAIN'])
+            email_from='{}@{}'.format(authenticated_service.email_from, current_app.config['NOTIFY_EMAIL_DOMAIN']),
+            content=WithSubjectTemplate.__str__(template_with_content),
         )
     elif notification_type == LETTER_TYPE:
         create_resp_partial = functools.partial(
             create_post_letter_response_from_notification,
             subject=template_with_content.subject,
+            content=WithSubjectTemplate.__str__(template_with_content),
         )
 
     resp = create_resp_partial(
         notification=notification,
-        content=str(template_with_content),
         url_root=request.url_root,
         scheduled_for=scheduled_for
     )
--- a/tests/app/celery/test_tasks.py
+++ b/tests/app/celery/test_tasks.py
@@ -9,7 +9,11 @@ from freezegun import freeze_time
 from requests import RequestException
 from sqlalchemy.exc import SQLAlchemyError
 from celery.exceptions import Retry
-from notifications_utils.template import SMSMessageTemplate, WithSubjectTemplate
+from notifications_utils.template import (
+    LetterPrintTemplate,
+    PlainTextEmailTemplate,
+    SMSMessageTemplate,
+)
 from notifications_utils.columns import Row
 
 from app import (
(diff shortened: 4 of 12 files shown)


app/models.py
```diff
@@ -977,19 +977,20 @@ def is_precompiled_letter(self, value):
...
         if self.template_type == SMS_TYPE:
-            return SMSMessageTemplate(
-                {'content': self.content}
-            )
+            return SMSMessageTemplate(self.__dict__)
         if self.template_type == LETTER_TYPE:
             return LetterPrintTemplate(
-                {'content': self.content, 'subject': self.subject},
+                self.__dict__,
                 contact_block=self.service.get_default_letter_contact(),
```
quis, line 986: This should get the actual contact block for the template or notification

app/notifications/process_notifications.py
```diff
@@ -32,11 +32,10 @@
 )
 
 from app.v2.errors import BadRequestError
-from app.utils import get_template_instance
```
idavidmcdonald, line 35: Can we delete `get_template_instance` now?

app/v2/notifications/post_notifications.py
```diff
@@ -166,23 +167,25 @@ def post_notification(notification_type):
...
             subject=template_with_content.subject,
+            content=PlainTextEmailTemplate.__str__(template_with_content),
         )
 
     resp = create_resp_partial(
         notification=notification,
-        content=str(template_with_content),
         url_root=request.url_root,
         scheduled_for=scheduled_for
     )
```
idavidmcdonald, line 191: Might need to talk to this change in person as I'm not quite sure what the difference is or why the difference is needed please

tests/app/notifications/rest/test_send_notification.py
```diff
@@ -142,18 +142,18 @@ def test_send_notification_with_placeholders_replaced(notify_api, sample_email_t
...
         (
             'Hello \n\n'
-            '* Jo\n'
-            '* John\n'
-            '* Josephine\n'
-            'This is an email from GOV.UK'
+            '• Jo\n'
+            '• John\n'
+            '• Josephine\n'
+            'This is an email from GOV.\u200bUK\n'
```
idavidmcdonald, line 148: Could you explain why you are making all the unicode changes in these tests please as I'm not sure why?

tests/app/notifications/rest/test_send_notification.py
```diff
@@ -400,7 +400,7 @@ def test_should_allow_valid_email_notification(notify_api, sample_email_template
             assert response.status_code == 201
             assert notification_id
             assert response_data['subject'] == 'Email Subject'
-            assert response_data['body'] == sample_email_template.content
+            assert response_data['body'] == sample_email_template.content + '\n'
```
idavidmcdonald, line 403: Where does the `\n` come from please?

tests/app/v2/notifications/test_post_letter_notifications.py
```diff
@@ -450,15 +485,34 @@ def test_post_letter_notification_is_delivered_and_has_pdf_uploaded_to_test_lett
...
 
-def test_post_letter_notification_persists_notification_reply_to_text(
+def test_post_letter_notification_ignores_reply_to_text_for_service(
+    client, notify_db_session, mocker
+):
+    mocker.patch('app.celery.letters_pdf_tasks.create_letters_pdf.apply_async')
+
+    service = create_service(service_permissions=[LETTER_TYPE])
+    create_letter_contact(service=service, contact_block='ignored', is_default=True)
+    template = create_template(service=service, template_type='letter')
```
idavidmcdonald, line 495: To clarify, this test isn't testing when the contact block has no placeholders, it testing that we didn't provide  `reply_to=letter_contact_block.id` in the template?

tests/app/v2/notifications/test_post_letter_notifications.py
```diff
@@ -467,7 +521,7 @@ def test_post_letter_notification_persists_notification_reply_to_text(
 
     notifications = Notification.query.all()
     assert len(notifications) == 1
-    assert notifications[0].reply_to_text == service_address
```
idavidmcdonald, line 470: It might be the git diff but are we dropping a test here? The one that checks that the default service address is used if we don't override it at the create_template level?

tests/app/celery/test_tasks.py
```diff
@@ -1298,13 +1303,81 @@ def test_should_cancel_job_if_service_is_inactive(sample_service,
...
 
 
-@pytest.mark.parametrize('template_type, expected_class', [
-    (SMS_TYPE, SMSMessageTemplate),
-    (EMAIL_TYPE, WithSubjectTemplate),
-    (LETTER_TYPE, WithSubjectTemplate),
-])
-def test_get_template_class(template_type, expected_class):
-    assert get_template_class(template_type) == expected_class
+def test_get_email_template_instance(mocker, sample_email_template, sample_job):
```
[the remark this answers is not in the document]
quis: It’s also making sure that `RecipientCSV` is getting called with the right value for the  `placeholders` argument, which was the source of the bug.

app/models.py
```diff
@@ -983,9 +983,14 @@ def _as_utils_template(self):
         if self.template_type == LETTER_TYPE:
             return LetterPrintTemplate(
                 self.__dict__,
-                contact_block=self.service.get_default_letter_contact(),
+                contact_block=self.get_reply_to_text(),
```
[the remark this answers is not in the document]
quis: No – this is a change to the `_as_utils_template_method` to make it possible to use it for checking personalisation. Without this change it would always be looking for placeholders in the service’s default contact block, not the contact block for that specific template.

app/notifications/process_notifications.py
quis: It’s still used in a few places where we only have template JSON, not a database object.

app/v2/notifications/post_notifications.py
```diff
@@ -166,23 +167,25 @@ def post_notification(notification_type):
...
             subject=template_with_content.subject,
-            email_from='{}@{}'.format(authenticated_service.email_from, current_app.config['NOTIFY_EMAIL_DOMAIN'])
+            email_from='{}@{}'.format(authenticated_service.email_from, current_app.config['NOTIFY_EMAIL_DOMAIN']),
+            content=PlainTextEmailTemplate.__str__(template_with_content),
         )
     elif notification_type == LETTER_TYPE:
         create_resp_partial = functools.partial(
             create_post_letter_response_from_notification,
             subject=template_with_content.subject,
+            content=PlainTextEmailTemplate.__str__(template_with_content),
```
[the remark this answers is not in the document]
quis: I’ve changed this to `WithSubjectTemplate` now, which:
- works for both emails and letters 
- is what we being used before to render templates into the JSON response
- doesn’t add trailing `\n` characters to the output

app/v2/notifications/post_notifications.py
```diff
@@ -166,23 +167,25 @@ def post_notification(notification_type):
     if notification_type == SMS_TYPE:
         create_resp_partial = functools.partial(
             create_post_sms_response_from_notification,
-            from_number=reply_to
+            from_number=reply_to,
+            content=str(template_with_content),
```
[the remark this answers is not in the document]
quis: `template_with_content` is an instance of `SMSMessageTemplate`; calling `str` on it returns the string representation of its content.

app/models.py
```diff
@@ -977,16 +977,12 @@ def is_precompiled_letter(self, value):
...
         if self.template_type == SMS_TYPE:
-            return SMSMessageTemplate(
-                {'content': self.content}
-            )
+            return SMSMessageTemplate(self.__dict__)
         if self.template_type == LETTER_TYPE:
             return LetterPrintTemplate(
-                {'content': self.content, 'subject': self.subject},
+                self.__dict__,
                 contact_block=self.service.get_default_letter_contact(),
```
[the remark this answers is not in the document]
quis: Yeah, that would be nicer, but also a lot more work.

tests/app/notifications/rest/test_send_notification.py
quis: I’ve found a way of doing this without the unicode changes now (they are replacements we make specifically for emails to stop browsers making GOV.UK into a link).

tests/app/v2/notifications/test_post_letter_notifications.py
quis: The expected behaviour is that the default service address _shouldn’t_ be used if `template.service_letter_contact_id` is set to `None`. 

That should be treated as the letter contact block being blank instead:
image

PR #2798, opened 2020-04-07
Delete notifications older than retention

idavidmcdonald:
Please read https://github.com/alphagov/notifications-api/pull/2788 and then review this commit by commit

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -62,6 +62,14 @@ def _remove_csv_files(job_types):
         current_app.logger.info("Job ID {} has been removed from s3.".format(job.id))
 
 
+@notify_celery.task(name="delete-notifications-older-than-retention")
+@statsd(namespace="tasks")
+def delete_notifications_older_than_retention():
+    delete_email_notifications_older_than_retention()
+    delete_sms_notifications_older_than_retention()
+    delete_letter_notifications_older_than_retention()
+
+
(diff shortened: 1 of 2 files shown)


app/celery/nightly_tasks.py
```diff
@@ -62,6 +62,14 @@ def _remove_csv_files(job_types):
...
 
+@notify_celery.task(name="delete-notifications-older-than-retention")
+@statsd(namespace="tasks")
+def delete_notifications_older_than_retention():
+    delete_email_notifications_older_than_retention()
+    delete_sms_notifications_older_than_retention()
+    delete_letter_notifications_older_than_retention()
+
+
 @notify_celery.task(name="delete-sms-notifications")
```
leohemsted, line 73: i like that we're keeping the existing tasks so we can still run them manually if needed

idavidmcdonald:
Looking at the past 14 days of data (which includes some deleting of days where we sent 7 million+ emails), the longest it took to delete per notification was:

emails: 1h 56m
sms: 52m
letters: 1h 26m

So if we assume we had the worst case for each notification type on the same day then that would be 4h 24 minutes. If that kicked off at the new time of 3am then it would finish by 7.30AM.

I think this is acceptable, but shout if you think otherwise?

Given those times as well, I don't think we will run into issues with cronitor, at least not after the first run but will keep an eye on it.

PR #2802, opened 2020-04-14
fix timezone related cancel letter job bug

leohemsted:
bugs bug bugs!

also would have bugged out on single notification pages, except the button to cancel wasn't visible due to the bug on admin


CrystalPea:
> also would have bugged out on single notification pages, except the button to cancel wasn't visible due to the bug on admin

😬

PR #2803, opened 2020-04-14
Use firetext response code to see if temporary or permanent failure if available

CrystalPea:
Currently, when we get first pending status, then permanent failure status from Firetext, we class it as temporary failure. But this is not always right. 

Hence, when we receive that succession of statuses, we will check for a code that tells us exactly what happened, so we know the status for sure.

Story ticket: https://www.pivotaltracker.com/story/show/171742892

Questions:
- maybe we would like to defer to the code every time we receive a failure from firetext?
- Since each code has a detailed explanation for failure, how about exposing those to our users? They often ask why messages failed, so more detailed explanation might help them with their work.

### DIFF
--- a/app/clients/sms/firetext.py
+++ b/app/clients/sms/firetext.py
@@ -20,15 +20,15 @@ firetext_responses = {
 }
 
 firetext_codes = {
-    '101': 'permanent-failure',  # Unknown Subscriber
-    '102': 'temporary-failure',  # Absent Subscriber
-    '103': 'temporary-failure',  # Subscriber Busy
-    '104': 'temporary-failure',  # No Subscriber Memory
-    '201': 'permanent-failure',  # Invalid Number
-    '301': 'permanent-failure',  # SMS Not Supported
-    '302': 'temporary-failure',  # SMS Not Supported
-    '401': 'permanent-failure',  # Message Rejected
-    '900': 'temporary-failure',  # Routing Error
+    '101': {'status': 'permanent-failure', 'reason': 'Unknown Subscriber'},
+    '102': {'status': 'temporary-failure', 'reason': 'Absent Subscriber'},
+    '103': {'status': 'temporary-failure', 'reason': 'Subscriber Busy'},
+    '104': {'status': 'temporary-failure', 'reason': 'No Subscriber Memory'},
+    '201': {'status': 'permanent-failure', 'reason': 'Invalid Number'},
+    '301': {'status': 'permanent-failure', 'reason': 'SMS Not Supported'},
+    '302': {'status': 'temporary-failure', 'reason': 'SMS Not Supported'},
+    '401': {'status': 'permanent-failure', 'reason': 'Message Rejected'},
+    '900': {'status': 'temporary-failure', 'reason': 'Routing Error'},
 }
 
 
@@ -36,8 +36,8 @@ def get_firetext_responses(status):
     return firetext_responses[status]
 
 
-def get_message_status_from_firetext_code(code):
-    return firetext_codes[code]
+def get_message_status_and_reason_from_firetext_code(code):
+    return firetext_codes[code]['status'], firetext_codes[code]['reason']
 
 
 class FiretextClientResponseException(SmsClientResponseException):
--- a/app/clients/sms/firetext.py
+++ b/app/clients/sms/firetext.py
@@ -19,9 +19,25 @@ firetext_responses = {
     '2': 'pending'
 }
 
+firetext_codes = {
+    '101': 'permanent-failure',  # Unknown Subscriber
+    '102': 'temporary-failure',  # Absent Subscriber
+    '103': 'temporary-failure',  # Subscriber Busy
+    '104': 'temporary-failure',  # No Subscriber Memory
+    '201': 'permanent-failure',  # Invalid Number
+    '301': 'permanent-failure',  # SMS Not Supported
+    '302': 'temporary-failure',  # SMS Not Supported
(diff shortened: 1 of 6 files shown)


app/dao/notifications_dao.py
```diff
@@ -89,11 +90,15 @@ def dao_create_notification(notification):
     db.session.add(notification)
 
 
-def _decide_permanent_temporary_failure(current_status, status):
+def _decide_permanent_temporary_failure(current_status, status, code=None):
     # Firetext will send pending, then send either succes or fail.
-    # If we go from pending to delivered we need to set failure type as temporary-failure
+    # If we go from pending to failure we need to set failure type as temporary-failure
+    # if we get a detailed code from firetext, we should use that code to set status instead
```
idavidmcdonald, line 96: I suggest capatilising this first word and adding a full stop the line above so it's clear they are two separate sentences

tests/app/notifications/rest/test_callbacks.py
```diff
@@ -147,7 +147,7 @@ def test_firetext_callback_should_return_200_and_call_task_with_valid_data(clien
     assert json_resp['result'] == 'success'
 
     mock_celery.assert_called_once_with(
-        ['0', 'notification_id', 'Firetext'],
+        ['0', 'notification_id', 'Firetext', None],
         queue='sms-callbacks',
     )
```
idavidmcdonald, line 152: We should add a test for getting the callback with a code please

app/clients/sms/firetext.py
```diff
@@ -19,11 +19,27 @@
     '2': 'pending'
 }
 
+firetext_codes = {
+    '101': 'permanent-failure',  # Unknown Subscriber
```
idavidmcdonald, line 23: Good use of comments

app/dao/notifications_dao.py
```diff
@@ -89,11 +90,15 @@ def dao_create_notification(notification):
...
     if current_status == NOTIFICATION_PENDING and status == NOTIFICATION_PERMANENT_FAILURE:
-        status = NOTIFICATION_TEMPORARY_FAILURE
+        if code:
+            status = get_message_status_from_firetext_code(code)
```
idavidmcdonald, line 99: If we get a code back from firetext which isn't in our dict (maybe they change their codes or add a new code and we aren't aware), we will throw a `KeyError` here. It feels unlikely though, as that doc they sent has changed in 4 years. Does that feel OK? I'm not decided if I think it's fine like that or we should catch it and do something with it (but I don't know what I think we should do).

What do you think?

app/dao/notifications_dao.py
CrystalPea: They actually are not totally separate sentences. I made that clearer now.

app/clients/sms/firetext.py
CrystalPea: Thanks :)

tests/app/notifications/rest/test_callbacks.py
CrystalPea: done 👍

CrystalPea:
> I don't understand your question: `maybe we would like to defer to the code every time we receive a failure from firetext?`. Could you rephrase this please?

So as this PR stands now, and as the story was written, we would only look at firetext failure codes if we got 2 receipts from firetext, first: (status 2) pending, and then: (status 1) permanent-failure. 
But I think there might be value in looking for the code each time we get a failure, even if we did not get "pending" status earlier. I'd like to hear @servingUpAces opinion on this, as maybe she knows more than we do.

> I definitely agree with you that there is value we are missing by not capturing these. I think we have three options
> 1. Do nothing with the data

I think this would be a lost opportunity.

> 2. Log it so we can go analyse it ourselves through the logs, would help us investigate cases were the user wants to know why it failed. I think this might be worth doing. Assuming 10% of firetext messages give us a code for this, it wouldn't be masses of logs per day to store this information

I think this would be a good thing to do as part of this PR. Quick and potentially useful to investigate failures and provide more information to users upon request.

> 3. Add a new DB column to the notifications table to record this type of thing HOWEVER this is probably a bit challenging to do at the moment and would require work so that might be out of scope of this PR. It's probably worth checking with Becca or Pete whether we should write a story for it

I really like this option and think we should consider making a story about this so we remember to do it later. I don't think it's super urgent, but I think it it will provide value to our users.

app/dao/notifications_dao.py
CrystalPea: Ahh, good point, I think we should catch it and proceed with our old method as a fallback, and at the same time log an error with the new / unknown code in it, so we can investigate and add it to our list of codes. 

What do you think?

idavidmcdonald: Sounds like a good solution, good for it.

idavidmcdonald:
Cool. So let's go for option 2 and put some basic logging in which may help in the future?

Also, I didn't realise they sent us a code at any other time? Do they always send codes or is it just when they send us the permanent failure after a temp failure?

CrystalPea:
> Do they always send codes or is it just when they send us the permanent failure after a temp failure?
@idavidmcdonald it's always when they send us failure I think.

They have 3 basic statuses: success, pending, and failure

Each time they send us failure status, they also send us the exact code for this failure as far as I know. So we could use it each time we get a failure, and fail back to what we do now if we do not get the detailed code.

I have changed our code now to always look at the Firetext failure code first (check the latest commit). I also added logging and a catch for KeyError if the code is not recognised.

Let me know what you think :)

PR #2810, opened 2020-04-20
Fix bug where preview for templated letters would not show

CrystalPea:
We bumped version of utils here: https://github.com/alphagov/notifications-template-preview/pull/441

What we did not realise is that now template preview would require more information from API.

This PR fixes that omission.

Before:
Screenshot 2020-04-20 at 15 24 03

After:
Screenshot 2020-04-20 at 15 23 37


CrystalPea:
Sweet, thanks a lot @quis

PR #2811, opened 2020-04-20
Add GDSMetrics package

idavidmcdonald:
As per instructions https://github.com/alphagov/gds_metrics_python

The celery workers don't have an HTTP endpoint so no point in trying to
get prometheus to scrape them.

image

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -7,6 +7,7 @@ from flask import _request_ctx_stack, request, g, jsonify, make_response
 from flask_sqlalchemy import SQLAlchemy as _SQLAlchemy
 from flask_marshmallow import Marshmallow
 from flask_migrate import Migrate
+from gds_metrics import GDSMetrics
 from time import monotonic
 from notifications_utils.clients.zendesk.zendesk_client import ZendeskClient
 from notifications_utils.clients.statsd.statsd_client import StatsdClient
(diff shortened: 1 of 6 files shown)


klssmith:
What is the reason for adding this directly to `notify-api` instead of relying on `notify-metric-exporter` to get the stats for Prometheus?

idavidmcdonald:
Good question Katie. So there is a difference in the metrics they produce.

The `notify-metric-exporter` is responsible for paas level stats (measured at the go router) about CPU, memory, disk space, request counts and latency. We will still keep this for the API.

The GDS Metrics library instead offers us python level metrics measured at flask and will give us the request counts and latency, but the key difference is it will do so broken down by HTTP endpoint. So we can compare the speed of requests between say two different routes. This will help us find slow running requests or measure performance changes to an endpoint.

idavidmcdonald:
Oh also note, we already have this is in place for document download API so you can compare the code here to the code there.

It will enable us to make something like this but for the API https://grafana-paas.cloudapps.digital/d/zxzfy2qZk/document-download-api-http?orgId=2&from=now-6h&to=now&var-space=production&var-app=document-download-api&var-path=%2Fservices%2F%3Cuuid:service_id%3E%2Fdocuments%2F%3Cuuid:document_id%3E

klssmith:
I thought I vaguely remembered that there was a reason why document-download-api could use GDS metrics, but the other apps didn't / couldn't and had to use the `notify-metric-exporter` instead - do you know why that was?

idavidmcdonald:
Oh no, absolutely no context on that. I assumed that doc download api was more of a trial app to see whether it was useful and then we likely just decided to exclusively use the notify-metric-exporter as it's much quicker to set up per app. So yeah, no idea I'm afraid if that is the case... I asked Leo to take a look at this PR so he might have some context

PR #2814, opened 2020-04-21
Let users search for letters

quis:
Like we have search by email address or phone number, finding an individual letter is a common task. At the moment users are having to click through pages and pages of letters to find the one they’re looking for.

We have to search in the `to` and `normalised_to` fields for now until we’ve been storing the normalised version for a few weeks.

## How we store addresses

At the moment we’re not consistent:

**Precompiled (API and one-off)**
`to` has the whole address
`normalised_to` has nothing

**Templated (API, CSV and one off)**
`to` has the first line of the address
`normalised_to` has nothing

This pull request makes us consistently store the whole address in the `to` field. We think that people might want to search by postcode, not just first line of the address.

This commit also starts to populate the `normalised_to` field with the address lowercased and with all spaces removed, to make it easier to search on.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -342,7 +342,6 @@ def save_letter(
 ):
     notification = encryption.decrypt(encrypted_notification)
 
-    # we store the recipient as just the first item of the person's address
     recipient = PostalAddress.from_personalisation(
         Columns(notification['personalisation'])
     ).normalised
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -345,7 +345,7 @@ def save_letter(
     # we store the recipient as just the first item of the person's address
     recipient = PostalAddress.from_personalisation(
         Columns(notification['personalisation'])
-    ).normalised_lines[0]
+    ).normalised
 
     service = dao_fetch_service_by_id(service_id)
     template = dao_get_template_by_id(notification['template'], version=notification['template_version'])
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -26,7 +26,6 @@ from werkzeug.datastructures import MultiDict
 from app import db, create_uuid
 from app.aws.s3 import remove_s3_object, get_s3_bucket_objects
 from app.dao.dao_utils import transactional
-from app.errors import InvalidRequest
 from app.letters.utils import get_letter_pdf_filename
 from app.models import (
     FactNotificationStatus,
@@ -600,11 +599,8 @@ def dao_get_notifications_by_recipient_or_reference(service_id, search_term, not
         except InvalidEmailError:
             normalised = search_term.lower()
 
-    elif notification_type == LETTER_TYPE:
-        raise InvalidRequest("Only email and SMS can use search by recipient", 400)
-
     else:
-        normalised = search_term.lower()
+        normalised = ''.join(search_term.split()).lower()
 
     normalised = escape_special_characters(normalised)
     search_term = escape_special_characters(search_term)
@@ -612,6 +608,7 @@ def dao_get_notifications_by_recipient_or_reference(service_id, search_term, not
     filters = [
         Notification.service_id == service_id,
         or_(
+            Notification.to.ilike("%{}%".format(search_term)),
             Notification.normalised_to.like("%{}%".format(normalised)),
             Notification.client_reference.ilike("%{}%".format(search_term)),
         ),
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -599,8 +599,19 @@ def dao_get_notifications_by_recipient_or_reference(service_id, search_term, not
         except InvalidEmailError:
             normalised = search_term.lower()
 
-    else:
+    elif notification_type in {LETTER_TYPE, None}:
+        # For letters, we store the address without spaces, so we need
+        # to removes spaces from the search term to match. We also do
+        # this when a notification type isn’t provided (this will
+        # happen if a user doesn’t have permission to see the dashboard)
+        # because email addresses and phone numbers will never be stored
+        # with spaces either.
         normalised = ''.join(search_term.split()).lower()
+
+    else:
+        raise TypeError(
+            f'Notification type must be {EMAIL_TYPE}, {SMS_TYPE}, {LETTER_TYPE} or None'
+        )
 
     normalised = escape_special_characters(normalised)
     search_term = escape_special_characters(search_term)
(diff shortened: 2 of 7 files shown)


app/dao/notifications_dao.py
```diff
@@ -600,18 +599,16 @@ def dao_get_notifications_by_recipient_or_reference(service_id, search_term, not
         except InvalidEmailError:
             normalised = search_term.lower()
 
-    elif notification_type == LETTER_TYPE:
-        raise InvalidRequest("Only email and SMS can use search by recipient", 400)
-
     else:
```
idavidmcdonald, line 602: Two questions around this.

Why did we have an `else` statement originally here? If it's not sms, email or letter, surely something is up? Either that or we are allowing you to search across all notification types?

Secondly, I think I'd like it to see laid out as

```
if sms:
    do thing

if email:
    do other thing

if letter:
    do different thing

else:
    do this, with a comment explaining why the generic behaviour is appropriate for all notification types
```

It would be much more explicit this way about the behaviour for all notifications and letters. Because here it appears not are only are we adding functionality for searching for letters, but we are also changing the value of `normalised` for when we aren't searching by notification type

app/celery/tasks.py
```diff
@@ -345,7 +345,7 @@ def save_letter(
     # we store the recipient as just the first item of the person's address
```
idavidmcdonald, line 345: Is this comment now incorrect?

tests/app/celery/test_tasks.py
```diff
@@ -950,26 +950,48 @@ def test_save_sms_does_not_send_duplicate_and_does_not_put_in_retry_queue(sample
...
+        'SE1 2SA'
+    ), (
+        'foobarbazwibblewobblewubblese12sa'
+    )),
+    ({
         # The address isn’t normalised when we store it in the
-        # `personalisation` column, but the first line is normalised for
-        # Storing in the `to` column
+        # `personalisation` column, but is normalised for storing in the
+        # `to` column
```
idavidmcdonald, line 976: This comment confuses me?
I thought we store it as is in the personalisation column, as a single string in the `to` column and as a normalised string in the `normalised_to` column? But this says we are storing it normalised in the `to` column?

tests/app/celery/test_tasks.py
```diff
@@ -950,26 +950,48 @@ def test_save_sms_does_not_send_duplicate_and_does_not_put_in_retry_queue(sample
     assert not retry.called
 
 
-@pytest.mark.parametrize('personalisation', (
-    {
+@pytest.mark.parametrize('personalisation, expected_to, expected_normalised', (
```
idavidmcdonald, line 953: Good test here and nice and clear variable names, nice :)

tests/app/celery/test_tasks.py
idavidmcdonald: So I think what I've figured out is that there is actually two times we normalise the data
Once from 7 different strings to a single string that has spaces
A secondly from a single string that has spaces to a single string without spaces and lowercase
This then means that when we refer to `normalise` around the codebase to do it with, it's quite confusing and maybe a little misleading. Is there a quick fix to this? If it's a long one we may have to live with it though

app/celery/tasks.py
quis: Removed in 11008af9

app/dao/notifications_dao.py
quis: For users who don’t have permission to see the dashboard all the notifications for their service appear on one page. Which means the search works across multiple notification types. There is an existing test for this: https://github.com/alphagov/notifications-api/blob/d88b20beeca799f920eb84fb1695f336e4fb5a29/tests/app/dao/notification_dao/test_notification_dao.py#L1314-L1318

When the only two searchable types of notification were `email` and `sms` then the lowest common denominator normalisation was to just lowercase the search term. But now that letters are searchable we need a normalisation that will also find letters. And that means smooshing spaces together, because our normalised addresses will have all their spaces smooshed together (in other words a search for ‘David M’ needs to match `davidm` in the database).

Is this worth adding as a comment in the code?

tests/app/celery/test_tasks.py
quis: The two meanings are normalised for search and normalised for the postal system. Might have to live with it since it goes as deep as the name of a database column though 😬

app/dao/notifications_dao.py
quis: I’ve added a comment, and removed the catch-all `else` in 64674be

app/dao/notifications_dao.py
```diff
@@ -600,18 +599,27 @@ def dao_get_notifications_by_recipient_or_reference(service_id, search_term, not
...
+            f'Notification type must be {EMAIL_TYPE}, {SMS_TYPE}, {LETTER_TYPE} or None'
+        )
 
     normalised = escape_special_characters(normalised)
     search_term = escape_special_characters(search_term)
 
     filters = [
         Notification.service_id == service_id,
         or_(
+            Notification.to.ilike("%{}%".format(search_term)),
```
idavidmcdonald, line 622: The only thing I'm wondering is that I think it's only useful for us to search the `to` field for letters, is that correct? As for emails and SMS it's taken care of by the search of `normalised_to`? 

I'm just wondering if this is going to have a performance impact for searching through SMS and emails as now it's also going to be doing an `ilike` on the `to` field which seems like it could be a fairly intensive query (and I don't know what indexing we have on that column to support that also)?

Do you think it might be an issue too? Ideally we could know by doing a before and after test what the impact would be and look at some stats (like https://grafana-paas.cloudapps.digital/d/u5LtjAqZk/notify-api-http?orgId=2) but it may just be worth trying to do some logic to only include the or statement for letters or `none`?

Sorry I didn't think about this in the previous PR review

quis: Some quick testing on the production database, running these queries 10 times:

```sql
select
  "to" from notifications
where
  service_id = '<GOV.UK Email>' and "normalised_to" like '%chris%'
limit 50;
```
Slowest query: 573ms
Fastest query: 32ms


```sql
select
  "to" from notifications
where
  service_id = '<GOV.UK Email>' and "to" ilike '%chris%'
limit 50;
```
Slowest query: 850ms
Fastest query: 28ms

***

I expect the fast queries are Postgres doing some kind of caching. The slowest queries for the `ilike` version don’t seem too much worse, so reckon we can live with them bring in there for a week until the `normalised_to` column has populated?

idavidmcdonald: You are probably right that we can deal with it for a week.

Also, I'm not sure if those queries are for the difference I was getting at. It's not whether the `like` on `normalised_to` is different in speed compared to the `ilike` on `to`, it's that how much slower is it to search with BOTH the `like` on `normalised_to` and `ilike` on `to` versus just the `like` on `normalised_to`, when in the case of emails and SMS, the `ilike` on `to` is of no benefit and is just adding query time.

Saying that, given you've tested this on GOV.UK email, I think most smaller services that use search would see little difference in response times.

PR #2817, opened 2020-04-22
Update _status page - remove references to Travis

CrystalPea:
Story ticket: https://www.pivotaltracker.com/story/show/170997266

Other PR in series: https://github.com/alphagov/notifications-admin/pull/3423

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ help:
 
 .PHONY: generate-version-file
 generate-version-file: ## Generates the app version file
-	@echo -e "__travis_commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"\n__travis_job_number__ = \"${BUILD_NUMBER}\"\n__travis_job_url__ = \"${BUILD_URL}\"" > ${APP_VERSION_FILE}
+	@echo -e "__git_commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"" > ${APP_VERSION_FILE}
 
 .PHONY: test
 test: generate-version-file ## Run tests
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -35,7 +35,7 @@ help:
 
 .PHONY: generate-version-file
 generate-version-file: ## Generates the app version file
-	@echo -e "__travis_commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"\n__travis_job_number__ = \"${BUILD_NUMBER}\"\n__travis_job_url__ = \"${BUILD_URL}\"" > ${APP_VERSION_FILE}
+	@echo -e "__git_commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"\n__job_url__ = \"${BUILD_URL}\"" > ${APP_VERSION_FILE}
```
klssmith, line 38: I think that you can just remove `__job_url__` completely from this command and all references to `BUILD_URL` from the Makefile too. It looks like we stopped using that in notifications-aws a while ago.

app/status/healthcheck.py
```diff
@@ -19,8 +19,7 @@ def show_status():
     else:
         return jsonify(
             status="ok",  # This should be considered part of the public API
-            travis_commit=version.__travis_commit__,
-            travis_build_number=version.__travis_job_number__,
+            github_commit=version.__git_commit__,
```
klssmith, line 22: `git` is used in some places and `github` in others, so it would be nice if that was consistent across the files. I prefer `git`, since GitHub is just the tool to view the git commits.

Makefile
CrystalPea: Ahh, good point, I did this now 👍

PR #2819, opened 2020-04-24
Additional prometheus metrics

leohemsted:
add the following prometheus metrics to keep track of general app
instance health.

you can see how these look locally by checking out the code and then going to http://localhost:6011/metrics (after doing some stuff locally to gather some data)

 # sqs_apply_async_duration

how long does the actual SQS call (a standard web request to AWS) take.
a histogram with default bucket sizes, split up per task that was
created.

 # concurrent_web_request_count

how many web requests is this app currently serving. this is split up
per process, so we'd expect multiple responses per app instance

 # db_connection_total_connected

how many connections does this app (process) have open to the database.
They might be idle.

 # db_connection_total_checked_out

how many connections does this app (process) have open that are
currently in use by a web worker

 # db_connection_open_duration_seconds

a histogram per endpoint of how long the db connection was taken from
the pool for. won't have any data if a connection was never opened.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,3 +1,4 @@
+import time
 import os
 import random
 import string
@@ -8,12 +9,14 @@ from flask_sqlalchemy import SQLAlchemy as _SQLAlchemy
 from flask_marshmallow import Marshmallow
 from flask_migrate import Migrate
 from gds_metrics import GDSMetrics
+from gds_metrics.metrics import Gauge, Histogram
 from time import monotonic
 from notifications_utils.clients.zendesk.zendesk_client import ZendeskClient
 from notifications_utils.clients.statsd.statsd_client import StatsdClient
 from notifications_utils.clients.redis.redis_client import RedisClient
 from notifications_utils.clients.encryption.encryption_client import Encryption
 from notifications_utils import logging, request_helper
+from sqlalchemy import event
 from werkzeug.exceptions import HTTPException as WerkzeugHTTPException
 from werkzeug.local import LocalProxy
 
@@ -107,6 +110,9 @@ def create_app(application):
     # avoid circular imports by importing this file later
     from app.commands import setup_commands
     setup_commands(application)
+
+    # set up sqlalchemy events
+    setup_sqlalchemy_events(application)
 
     return application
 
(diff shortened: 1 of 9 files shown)


app/celery/celery.py
```diff
@@ -12,6 +13,12 @@ def worker_process_shutdown(sender, signal, pid, exitcode, **kwargs):
 
 
 def make_task(app):
+    SQS_APPLY_ASYNC_DURATION = Histogram(
+        'sqs_apply_async_duration',
```
idavidmcdonald, line 17: do we want `duration_seconds` for consistency and I think best practice?

leohemsted:
1. pid is added to gauges automatically - see https://github.com/prometheus/client_python/blob/master/prometheus_client/values.py#L31.
2. deployment strategy. i thought it'd be fine pushing through to production as the performance issues are on read, not on write, and we've disabled production. this also includes the prometheus client bump so i hoped that would solve the performance issues.
3. didn't realise celery won't be gathered. the `prometheus_client` lib comes with its own built in http server, so we could easily add a sidecar in the future to expose these, though i don't think it's a priority now

idavidmcdonald:
Thanks Leo. In regards to turned off on production, I don't see that this is the case? It's still running in production, see https://grafana-paas.cloudapps.digital/d/u5LtjAqZk/notify-api-http?orgId=2

I manually unbound it for a while when the metrics page kept 502ing but whenever we deploy it will get rebound (my plan was to upgrade the version library which you have already done to fix this and then rebind it if someone else hadn't deployed already)

app/__init__.py
```diff
@@ -300,3 +317,71 @@ def process_user_agent(user_agent_string):
...
+        'How long db connections are held open for in seconds',
+        ['method', 'host', 'path']
+    )
+
+    # need this or db.engine isn't accessible
+    with app.app_context():
+        @event.listens_for(db.engine, 'connect')
+        def connect(dbapi_connection, connection_record):
+            # connection first opened with db
+            TOTAL_DB_CONNECTIONS.inc()
```
idavidmcdonald, line 345: Optional: Add an if statement to only increment this and decrease on close if `has_request_context` is true as the metrics themselves are not still not exposed for celery apps etc

leohemsted: hmmm, i don't think i'll do this right now as im not concerned about the actual act of writing metrics on boxes that don't expose things, i'm only concerned about errors introduced by metrics requiring access to variables they don't have (eg the `request.host` flask variable and stuff like that). i think the penalty of incrementing/decrementing a single gauge should be low enough that i'm not worried about it

app/__init__.py
```diff
@@ -379,17 +378,9 @@ def checkout(dbapi_connection, connection_record, connection_proxy):
...
                     'url_rule': request.url_rule.rule if request.url_rule else 'No endpoint'
                 }
-            # celery apps
-            elif current_task:
-                print(current_task)
-                connection_record.info['request_data'] = {
-                    'method': 'celery',
-                    'host': current_app.config['NOTIFY_APP_NAME'],  # worker name
-                    'url_rule': current_task.name,  # task name
-                }
```
leohemsted: this needs to stay, as the `checkin` will still happen in celery tasks, and will crash if connection_record.info['request_data'] isn't set. the print can go though

PR #2821, opened 2020-04-27
Reduce concurrency and optimise prefetching behaviour for the celery reporting app

idavidmcdonald:
Please review commit by commit.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -176,6 +176,9 @@ class Config(object):
     CELERY_TASK_SERIALIZER = 'json'
     # on reporting worker, restart workers after each task is executed to help prevent memory leaks
     CELERYD_MAX_TASKS_PER_CHILD = os.getenv('CELERYD_MAX_TASKS_PER_CHILD')
+    # we can set celeryd_prefetch_multiplier to be 1 for celery apps which handle only long running tasks
+    if os.getenv('CELERYD_PREFETCH_MULTIPLIER'):
+        CELERYD_PREFETCH_MULTIPLIER = os.getenv('CELERYD_PREFETCH_MULTIPLIER')
     CELERY_IMPORTS = (
         'app.celery.tasks',
         'app.celery.scheduled_tasks',
--- a/scripts/paas_app_wrapper.sh
+++ b/scripts/paas_app_wrapper.sh
@@ -29,7 +29,7 @@ case $NOTIFY_APP_NAME in
     -Q periodic-tasks 2> /dev/null
     ;;
   delivery-worker-reporting)
-    exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=11 \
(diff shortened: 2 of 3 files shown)


scripts/paas_app_wrapper.sh
```diff
@@ -30,7 +30,7 @@ case $NOTIFY_APP_NAME in
     ;;
   delivery-worker-reporting)
     exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=11 \
-    -Q reporting-tasks 2> /dev/null
+    -Q reporting-tasks -Ofair 2> /dev/null
```
servingUpAces, line 33: If we set prefetch multiplier to 1 do we need the -Ofair option?

servingUpAces:
Yes I think we do want to reduce the concurrency for the reporting app. There is another PR. Might be worth having a huddle about this app, perhaps put the changes with explanation into a single PR.

scripts/paas_app_wrapper.sh
idavidmcdonald: Yep, The prefetch multiplier is how many items to prefetch from SQS.

The `-Ofair` option is about once those items have been prefetched from SQS and are being held by the main master celery master process, does it automatically assign them to a celery child process or does it hold on to them and wait for a child process to free itself and then give the task to the child process.

They are both about "prefetching" but the first is from the broker and the second is from the master celery process. They solve different problems but are both beneficial.

app/config.py
```diff
@@ -176,6 +176,9 @@ class Config(object):
     CELERY_TASK_SERIALIZER = 'json'
     # on reporting worker, restart workers after each task is executed to help prevent memory leaks
     CELERYD_MAX_TASKS_PER_CHILD = os.getenv('CELERYD_MAX_TASKS_PER_CHILD')
+    # we can set celeryd_prefetch_multiplier to be 1 for celery apps which handle only long running tasks
+    if os.getenv('CELERYD_PREFETCH_MULTIPLIER'):
```
servingUpAces, line 180: Do we need this if statement? The CELERYD_MAX_TASKS_PER_CHILD option is not using one.

idavidmcdonald: I'm not sure what is different about the max tasks per child setting but if we don't include this statement then we get the following error when I run `./scripts/run_celery.sh` locally

```
File "/Users/davidmcdonald/.virtualenvs/notifications-api/lib/python3.6/site-packages/celery/worker/__init__.py", line 381, in setup_defaults
    'prefetch_multiplier', prefetch_multiplier,
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

leohemsted: you could have it default to `4` rather than `None` i guess using the optional arg in os.getenv

idavidmcdonald: I considered this and then decided I didn't want to replicate a setting in celery just in case celery later on decides to change it and we aren't following and update ours too

PR #2822, opened 2020-04-27
Add in lots of logging to our reporting tasks

idavidmcdonald:
We've seen only some of these reporting tasks happen but with no log messages to indicate what happened and no app crashes. 

For example, here you can only see 10 log lines from our `create-nightly-notification-status-for-day` tasks but we are expecting 18 (10 days for letters, 4 days for text messages, 4 days for emails). 
https://kibana.logit.io/s/9423a789-282c-4113-908d-0be3b1bc9d1d/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:'2020-04-26T22:46:17.785Z',mode:absolute,to:'2020-04-27T08:15:23.893Z'))&_a=(columns:!(message,router.response_time,access.response_code,syslog_proc),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:cf.app,negate:!f,type:phrase,value:notify-delivery-worker-reporting),query:(match:(cf.app:(query:notify-delivery-worker-reporting,type:phrase))))),index:'logstash-*',interval:auto,query:(query_string:(query:'%22data%20fetched%22')),sort:!('@timestamp',asc)). 

This PR will hopefully will give us a better picture of the timeline to further debug this.

Note, I've tried to make our message format very consistent and good for
searching for in kibana so I've changed that across this whole file for
consistency.

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -23,6 +23,7 @@ from app.models import (
 @cronitor("create-nightly-billing")
 @statsd(namespace="tasks")
 def create_nightly_billing(day_start=None):
+    current_app.logger.info("create-nightly-billing task: started")
     # day_start is a datetime.date() object. e.g.
     # up to 4 days of data counting back from day_start is consolidated
     if day_start is None:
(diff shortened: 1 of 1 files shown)


app/celery/reporting_tasks.py
```diff
@@ -31,37 +32,47 @@ def create_nightly_billing(day_start=None):
...
+        f'create-nightly-billing-for-day task for {process_day}: data fetched in {(end - start).seconds} seconds'
+    )
 
     for data in transit_data:
         update_fact_billing(data, process_day)
 
     current_app.logger.info(
-        f"create-nightly-billing-for-day task complete. {len(transit_data)} rows updated for day: {process_day}"
+        f"create-nightly-billing-for-day task for {process_day}: "
+        f"task complete. {len(transit_data)} rows updated for {process_day}"
```
leohemsted, line 67: don't think we need to keep the second `{process_day}` here

idavidmcdonald: Good spot, removed and rebased

PR #2825, opened 2020-04-28
Limit search by recipient to 50 results

quis:
Things could get ugly if you use a short search string on a service with lots of notifications…

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -586,7 +586,14 @@ def dao_update_notifications_by_reference(references, update_dict):
 
 
 @statsd(namespace="dao")
-def dao_get_notifications_by_recipient_or_reference(service_id, search_term, notification_type=None, statuses=None):
+def dao_get_notifications_by_recipient_or_reference(
+    service_id,
+    search_term,
+    notification_type=None,
+    statuses=None,
+    page=1,
+    page_size=None,
+):
 
     if notification_type == SMS_TYPE:
         normalised = try_validate_and_format_phone_number(search_term)
@@ -636,8 +643,7 @@ def dao_get_notifications_by_recipient_or_reference(service_id, search_term, not
     results = db.session.query(Notification)\
         .filter(*filters)\
         .order_by(desc(Notification.created_at))\
-        .limit(current_app.config['PAGE_SIZE'])\
-        .all()
+        .paginate(page=page, per_page=page_size)
     return results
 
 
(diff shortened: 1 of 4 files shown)


app/dao/notifications_dao.py
```diff
@@ -634,7 +634,11 @@ def dao_get_notifications_by_recipient_or_reference(service_id, search_term, not
     if notification_type:
         filters.append(Notification.notification_type == notification_type)
 
-    results = db.session.query(Notification).filter(*filters).order_by(desc(Notification.created_at)).all()
+    results = db.session.query(Notification)\
+        .filter(*filters)\
+        .order_by(desc(Notification.created_at))\
+        .limit(current_app.config['PAGE_SIZE'])\
```
idavidmcdonald, line 640: Optional: To pass the limit as an argument to the `dao` function and so it's more the view function that decides how many results should be retrieved which is more business logic (unless this a pattern that is already well established or other reasons)?

quis:
I think having a message saying ‘Only showing the first 50 results‘ would be fine – chances are if you have more than 50 you need to make your search more specific. So maybe this endpoint should limit at 51 results so we can know there are more than 50? Or is that too dirty a way of doing it?

idavidmcdonald:
Yeah, and only showing the first 50 is probably a quick solution for this case.

I think limiting to 51 isn't a bad shout, the question is just do you
-  send the 51 to the admin app and show the warning
-  send 50 and an extra variable saying more than 50 results.
- just sent 50 regardless of if there are more and then the admin shows the warning anyway 

I think I lean towards one of the later two as it feels weird to send 51 results across but I'd be happy with any really.

app/service/rest.py
```diff
@@ -475,10 +475,18 @@ def search_for_notification_by_to_field(service_id, search_term, statuses, notif
...
         statuses=statuses,
-        notification_type=notification_type
+        notification_type=notification_type,
+        page=1,
+        page_size=current_app.config['PAGE_SIZE'],
     )
     return jsonify(
-        notifications=notification_with_template_schema.dump(results, many=True).data
+        notifications=notification_with_template_schema.dump(results.items, many=True).data,
+        links=pagination_links(
```
idavidmcdonald, line 484: Optional: We have 4 other places we are returning paginated results. Three of the four seem to also return keys for total and page_size. Would be good to make it consistent I think?

Do we also want a unit test to check one of the responses at least contains the links key? I'm not sure if I'm too interested in testing the values of it as don't want to retest pagination_links itself but just as an indicator that we are returning paginated results?

PR #2829, opened 2020-05-01
set preload_app=True in gunicorn config

leohemsted:
preload_app is a flag that makes the gunicorn master process load the app code, and then copy that code to each worker. This is opposed to the default, where each worker initialises the app itself. We hope that by enabling this flag, we'll solve issues we've seen where instances sometimes return 502s when they start up. Our gut feel is that this happens because the workers are listening on the socket before they've finished loading the config etc. If the master thread loads the config, there shouldn't be this issue.

Advantages of preload_app: "save some RAM resources as well as speed up server boot times" [1]

Disadvantages of preload_app. Can no longer live-reload gunicorn workers to get new code. We don't do this anyway.

[1] https://docs.gunicorn.org/en/stable/settings.html#preload-app

#### Testing

I've locally tested this. And you can too! Just run `PORT=6011 gunicorn -c gunicorn_config.py application --error-logfile - --access-logfile -` to run gunicorn. You'll be able to see it only prints `Logging configured` once, before it prints "Starting gunicorn 20.0.4". If you run that command on master it'll print `Logging configured` four times (from each worker thread), after "Booting worker with pid: #####".

Start up times appear to be about the same (~3 seconds) using my very unscientific testing.

### DIFF
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -12,6 +12,7 @@ errorlog = "/home/vcap/logs/gunicorn_error.log"
 bind = "0.0.0.0:{}".format(os.getenv("PORT"))
 statsd_host = "{}:8125".format(os.getenv("STATSD_HOST"))
 gunicorn.SERVER_SOFTWARE = 'None'
+preload_app = True
 
 
 def on_starting(server):


gunicorn_config.py
```diff
@@ -12,6 +12,7 @@
 bind = "0.0.0.0:{}".format(os.getenv("PORT"))
 statsd_host = "{}:8125".format(os.getenv("STATSD_HOST"))
 gunicorn.SERVER_SOFTWARE = 'None'
+preload_app = False
```
idavidmcdonald, line 15: Do we want this to be set to True like the PR suggests :trollface:

leohemsted: no idea what you're talking about 😒

leohemsted:
Not merging until I've investigated the implications around queue pooling. See:

https://davidcaron.dev/sqlalchemy-multiple-threads-and-processes/

leohemsted:
I did have a look at this, but I'm ultimately closing this as, while I was able to get a hack in that would re-initialise the engine after the workers were forked by gunicorn, I didn't have enough confidence that this would solve it, while remaining clear and easy to read. I also wasn't entirely sure about what would happen with code that runs at import time, and if that'd affect anything.

leohemsted:
I'm closing this as, while the code snippet I tried did appear to work, I wasn't convinced that it was readable and debuggable and it felt very hacky. The advantages of preload_app didn't feel strong enough to risk that apps might break in strange ways, especially to do with things like code running on import vs in the app, that I didn't know how it'd handle.

PR #2831, opened 2020-05-04
don't expire email sign in codes on use

leohemsted:
we're seeing issues with email clients sniffing links, and causing them to expire before the user gets a chance to click on them. Temporarily disable the expiry while we work on a more permanent solution.

NB: The link will still expire after twenty four hours.

### DIFF
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -190,7 +190,8 @@ def verify_user_code(user_id):
         # only relevant from sms
         increment_failed_login_count(user_to_verify)
         raise InvalidRequest("Code not found", status_code=404)
-    if datetime.utcnow() > code.expiry_datetime or code.code_used:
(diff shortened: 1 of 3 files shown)


app/user/rest.py
```diff
@@ -192,8 +192,10 @@ def verify_user_code(user_id):
         raise InvalidRequest("Code not found", status_code=404)
     if datetime.utcnow() > code.expiry_datetime or code.code_used:
         # sms and email
-        increment_failed_login_count(user_to_verify)
-        raise InvalidRequest("Code has expired", status_code=400)
+        # TODO: Fix email flow so that clicking link doesn't expire emails
+        if data['code_type'] != 'email':
```
servingUpAces, line 196: Can we just check of code is used? That way the link still expires in 30 minutes?

leohemsted: good point. moved that check into the if statement above

PR #2832, opened 2020-05-04
Move pdf creation onto a template preview task

CrystalPea:
### In this PR:
- divide create_letters_pdf task into two tasks: one that sends data to template preview for pdf creation, and another task that sets billable units for notification (that task is called by the template preview task)
- move away uploading file to S3 onto the template preview task

#### For consideration:
I think `create_letters_pdf` task should now be named something like: `send_data_to_create_letter_pdf`
But this task is called in many places and I wondered if changing its name might cause problems during deploy?

--------------------

This template-preview PR needs to go in first: https://github.com/alphagov/notifications-template-preview/pull/447

Story ticket: https://www.pivotaltracker.com/story/show/171514311


servingUpAces:
If you would like to rename a task you need to create a task with the new name and keep the old task (both task use the same code so new task can call old task as a function), update code to call new task. Deploy that. 2nd deploy remove old task - by this point there should be not tasks on the queue using the old task name.

app/celery/letters_pdf_tasks.py
```diff
@@ -56,62 +50,63 @@
...
 
-    pages_per_sheet = 2
-    billable_units = math.ceil(int(resp.headers.get("X-pdf-page-count", 0)) / pages_per_sheet)
+    if notification.key_type != KEY_TYPE_TEST:
+        notification.billable_units = billable_units
+        dao_update_notification(notification)
 
-    return resp.content, billable_units
+    current_app.logger.info(
+        'Letter notification reference {reference}: billable units set to {billable_units}'.format(
```
servingUpAces, line 108: Can you include the notification_id in the log message, please? It's nice to search by id in the Kibanna

tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -56,188 +54,114 @@ def test_should_have_decorated_tasks_functions():
...
-    mock_get_letters_pdf.assert_called_once_with(
-        sample_letter_notification.template,
-        contact_block=sample_letter_notification.reply_to_text,
-        filename=None,
-        values=sample_letter_notification.personalisation
-    )
-
-
 # We only need this while we are migrating to the new letter_branding model
 def test_create_letters_gets_the_right_logo_when_service_has_letter_branding_logo(
```
servingUpAces, line 117: Is this different to test_create_letters_pdf_happy_path?

CrystalPea: The only difference is that branding is set for a service and as a result there is a logo filename for the letter. I could do a parametrize for it if you think it's too repetitive ☺️

app/celery/letters_pdf_tasks.py
CrystalPea: good suggestion, I don't know how I missed that, I will do that now 👍

app/notifications/process_notifications.py
```diff
@@ -138,7 +138,7 @@ def send_notification_to_queue(notification, research_mode, queue=None):
     if notification.notification_type == LETTER_TYPE:
         if not queue:
             queue = QueueNames.CREATE_LETTERS_PDF
```
servingUpAces, line 140: Since we are changing the name of the task do we want to change the name of the queue?

CrystalPea: what would you change it to?

PR #2836, opened 2020-05-06
Add support for CSV files

alixedi:
This addresses the second & third piece in the change proposed to support sending `.csv` files in email notifications here: https://github.com/alphagov/notifications-python-client/issues/164

The first piece for this has already been merged: https://github.com/alphagov/document-download-api/pull/122

### DIFF
--- a/app/clients/document_download.py
+++ b/app/clients/document_download.py
@@ -1,5 +1,3 @@
-import base64
-
 import requests
 
 from flask import current_app
@@ -26,15 +24,16 @@ class DocumentDownloadClient:
     def get_upload_url(self, service_id):
         return "{}/services/{}/documents".format(self.api_host, service_id)
 
-    def upload_document(self, service_id, file_contents):
+    def upload_document(self, service_id, file_contents, is_csv=None):
         try:
             response = requests.post(
                 self.get_upload_url(service_id),
                 headers={
                     'Authorization': "Bearer {}".format(self.auth_token),
                 },
-                files={
-                    'document': base64.b64decode(file_contents)
+                json={
+                    'document': file_contents,
+                    'is_csv': is_csv or False,
                 }
             )
 
--- a/tests/app/v2/notifications/test_post_notifications.py
+++ b/tests/app/v2/notifications/test_post_notifications.py
@@ -775,7 +775,16 @@ def test_post_email_notification_with_archived_reply_to_id_returns_400(client, s
     assert 'BadRequestError' in resp_json['errors'][0]['error']
 
 
-def test_post_notification_with_document_upload(client, notify_db_session, mocker):
+@pytest.mark.parametrize(
+    'csv_param',
+    (
+        {'is_csv': None},
+        {'is_csv': False},
+        {'is_csv': True},
+        {},
+    )
+)
+def test_post_notification_with_document_upload(client, notify_db_session, mocker, csv_param):
     service = create_service(service_permissions=[EMAIL_TYPE])
     service.contact_link = 'contact.me@gov.uk'
     template = create_template(
@@ -786,14 +795,14 @@ def test_post_notification_with_document_upload(client, notify_db_session, mocke
 
     mocker.patch('app.celery.provider_tasks.deliver_email.apply_async')
     document_download_mock = mocker.patch('app.v2.notifications.post_notifications.document_download_client')
-    document_download_mock.upload_document.side_effect = lambda service_id, content: f'{content}-link'
+    document_download_mock.upload_document.side_effect = lambda service_id, content, is_csv: f'{content}-link'
 
     data = {
         "email_address": service.users[0].email_address,
         "template_id": template.id,
         "personalisation": {
-            "first_link": {"file": "abababab"},
-            "second_link": {"file": "cdcdcdcd"}
+            "first_link": {"file": "abababab", **csv_param},
+            "second_link": {"file": "cdcdcdcd", **csv_param}
         }
     }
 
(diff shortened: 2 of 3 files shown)


tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -775,7 +775,16 @@ def test_post_email_notification_with_archived_reply_to_id_returns_400(client, s
     assert 'BadRequestError' in resp_json['errors'][0]['error']
 
 
-def test_post_notification_with_document_upload(client, notify_db_session, mocker):
+@pytest.mark.parametrize(
+    'csv_param',
```
idavidmcdonald, line 779: Good test coverage

app/clients/document_download.py
```diff
@@ -26,15 +24,16 @@ def init_app(self, app):
     def get_upload_url(self, service_id):
         return "{}/services/{}/documents".format(self.api_host, service_id)
 
-    def upload_document(self, service_id, file_contents):
+    def upload_document(self, service_id, file_contents, is_csv=None):
```
idavidmcdonald, line 27: Optional: I think it might make more sense to set the default argument up here to `False` rather than essentially doing it further down in the json block? It makes it a bit clearer about what the default behaviour is. So it would look like

```
def upload_document(self, service_id, file_contents, is_csv=False):
```

If you do go down this route, then it might also make sense to change `personalisation_data[key].get('is_csv')` to `personalisation_data[key].get('is_csv', False)` to match? Or even `service.id, personalisation_data[key]**`

Let me know what you think?

alixedi: 1. I am with you on:

```
def upload_document(self, service_id, file_contents, is_csv=False):
```

2. Changing `personalisation_data[key].get('is_csv')` to `personalisation_data[key].get('is_csv', False)` in the view would amount to a leaky abstraction.

I have updated the code to include (1) but not (2). If you would really like to have (2), I am happy to put that in as well. If not, we can totally merge this 😄 

Thank you for your time and patience 🙏

idavidmcdonald: So I think my concern with not doing (2) is that `personalisation_data[key].get('is_csv')` will almost always come back as `None` until everyone is using new versions of the API clients that support this new functionality. That `None` value is then passed through as `is_csv: None` into document download API and then even makes its way to be `is_csv = None` in document download API itself - https://github.com/alphagov/document-download-api/pull/122/files#diff-71d80e0e29b7065df20d6e25d268108bR28.

I see what you mean about the leaky abstraction but this feels like something I think we'd like to avoid, we would ideally be sure that document download is dealing with a boolean rather than risk a bug in the `None` case. We could change our code in document download API to ensure it is transformed into a boolean but it feels like this is an adequate place to do that and stop the problem before it gets there.

So in short, if you'd be willing to compromise and change it `personalisation_data[key].get('is_csv', False)` or change it so that we don't pass that argument into `upload_document` if the value is `None` then that would be much appreciated :)

Thanks Ali!

alixedi: https://github.com/alphagov/notifications-api/pull/2836/commits/f151b5f5b1093729569216837243dca521480dd0

alixedi: I think I understand the problem now. Thank you for the excellent explanation 🙏 

My reading of the problem definition is that we (rightly) do not want to leak non-boolean values like `None` to the `document-download-api`. 

To this end, I can simply revoke my last commit and we should be good to go: https://github.com/alphagov/notifications-api/pull/2836/commits/f151b5f5b1093729569216837243dca521480dd0

Please let me know what you think and thank you again for your time and patience.

idavidmcdonald: So I think you are right about revoking your last commit would remove the leak of non boolean values like `None` to the `document-download-api`.

However, I think by setting the default parameter value of `is_csv` to `None`, like in your previous commit, it hides the default behaviour inside the method rather than makes the default behaviour explicit. It feels like we are giving the user of `upload_document` a boolean choice in terms of functionality and therefore that boolean choice should be apparent in the default arguments.

Given that, I think I'd still encourage us to change it to `personalisation_data[key].get('is_csv', False)` as that is then the view making the decision about what it's default behaviour should be when the user doesn't specify. 

This opinion isn't too strongly held though so I think let me know if you choose to
- revoke that commit or
- don't revoke that commit and change to using `personalisation_data[key].get('is_csv', False)`

Either way, it should be good to approve :)

alixedi: I am going to revoke the commit because of the following reason:

```
>>> def view(data):
...     foo(data.get('is_csv', False))
...
>>> def upload(is_csv=False):
...     print(is_csv)
...
>>> view({'is_csv': True})
True
>>> view({'is_csv': False})
False
>>> view({})
False
```

idavidmcdonald: Sounds good. Thanks for your time to discuss this design this properly.

idavidmcdonald:
This is looking good and the unit tests have passed on our CI.

I'm going to spin this up locally this afternoon and just give it a quick end to end test with the other apps and then I'll approve it and we can get it deployed

idavidmcdonald:
So I've just done so end to end testing of this locally. Good news and bad news

Good news is it looks like this code is doing what it should do.

Bad news is that when I test this locally, the file that ends up being saved in S3 by document-download-api is saved but is of 0 length. It appears that this is is being caused by an issue with document download API in the following line of code:

https://github.com/alphagov/document-download-api/blob/master/app/upload/views.py#L27

If I print out the value of `file_data` at that point I get `<FileStorage: None (None)>` which seems bad.

I've tested locally removing the use of `FileStorage` so instead it's just doing `file_data = BytesIO(raw_content)` and then everything fixes itself. I haven't taken the time to understand what the problem is but it looks like this is a solution.

When I've run both this PR and document download API with my quick fix, then everything works correctly, the file is saved in s3 correctly and I can then change my API call successfully to decide whether to save this as a CSV file or default to a plain text file.

In short, if you are able to put in a PR please to change the behaviour of document download API to not use the `FileStorage` (and ideally understand why) then I can give this PR a final approval. Let me know if you need any help with this!

Thanks again!

currycoder:
Thanks for the excellent writeup here @idavidmcdonald - I'll look in to that issue and get a PR in for it ASAP.

alixedi:
@idavidmcdonald Please see: https://github.com/alphagov/document-download-api/pull/127

alixedi:
Thank you @idavidmcdonald. Been a pleasure!

We would pop in a PR for the final piece (python client) shortly.

PR #2843, opened 2020-05-13
Update postage constraint (take 2)

klssmith:
_This is the same as https://github.com/alphagov/notifications-api/pull/2747, which was reverted except that it does not drop the constraint on `notification_history` in production since that has been done manually._

**Update postage db constraints for international letters**

The `notifications`, `notification_history`, `templates` and `templates_history`
tables all had a check constraint on the postage column which specified
that the postage had to be `first` or `second` if the notification or
template was a letter. We now have two more options for postage -
`europe` and `rest-of-world`.

It's not possible to alter a check constraint, so the constraints have
to be dropped then recreated. We are not recreating the constraint on
the `notification_history` table since values here are always copied
from the `notifications` table.

The constraints get added as `NOT VALID` at first - this stage will lock
the tables, so updating the `notification` table and `templates` and
`templates_history` are done in separate migrations so that we don't lock
all tables at the same time. In a third migration we then run
`VALIDATE CONSTRAINT` for all tables - this will lock a row at a time,
not the whole table.

**Update JSON schema postage validation for new values**

### DIFF
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -34,8 +34,8 @@ def validate_schema_email_address(instance):
 @format_checker.checks('postage', raises=ValidationError)
 def validate_schema_postage(instance):
     if isinstance(instance, str):
-        if instance not in ["first", "second"]:
-            raise ValidationError("invalid. It must be either first or second.")
+        if instance not in ["first", "second", "europe", "rest-of-world"]:
+            raise ValidationError("invalid. It must be first, second, europe or rest-of-world.")
     return True
 
 
--- a/migrations/versions/0321_update_postage_constraint_1.py
+++ b/migrations/versions/0321_update_postage_constraint_1.py
@@ -1,0 +1,83 @@
+"""
+
+Revision ID: 0321_update_postage_constraint_1
+Revises: 0320_optimise_notifications
+Create Date: 2020-05-12 16:17:21.874281
+
+"""
+import os
+
+from alembic import op
+
+
+revision = '0321_update_postage_constraint_1'
+down_revision = '0320_optimise_notifications'
+environment = os.environ['NOTIFY_ENVIRONMENT']
+
+
+def upgrade():
+    op.drop_constraint('chk_notifications_postage_null', 'notifications')
+    op.execute("""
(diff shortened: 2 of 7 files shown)


app/schema_validation/__init__.py
```diff
@@ -34,8 +34,8 @@ def validate_schema_email_address(instance):
 @format_checker.checks('postage', raises=ValidationError)
 def validate_schema_postage(instance):
     if isinstance(instance, str):
-        if instance not in ["first", "second"]:
-            raise ValidationError("invalid. It must be either first or second.")
+        if instance not in ["first", "second", "europe", "rest-of-world"]:
+            raise ValidationError("invalid. It must be first, second, europe or rest-of-world.")
```
CrystalPea, line 38:
```suggestion
            raise ValidationError("invalid postage. It must be first, second, europe or rest-of-world.")
```

migrations/versions/0321_update_postage_constraint_1.py
```diff
@@ -0,0 +1,85 @@
...
+            ELSE
+                postage is null
+            END
+        )
+        NOT VALID
+    """)
+
+    # The constraint on notification_history will be dropped manually
+    if environment not in ["live", "production"]:
+        op.drop_constraint('chk_notification_history_postage_null', 'notification_history')
```
CrystalPea, line 34: I had to comment out this drop constraint on my local, too. Probably because I had this upgrade installed before, and then had some trouble downgrading, and was deleting constraints manually.

So for preview and staging it should be fine, because downgrading went smoother there when we were reverting this PR or something?

klssmith: For preview and staging we need to run this step because the constraint still exists. In production the constraint has already been dropped manually, so this line would give an error if we tried to run it there.

servingUpAces: We can change the constraint to say `ALTER TABLE notification_history DROP CONSTRAINT IF EXISTSchk_notifications_postage_null` Then we don't need the `if environment` check.

klssmith: 👍 That's updated now @servingUpAces

PR #2850, opened 2020-05-21
Re run failed reporting tasks

leohemsted:
We've had issues with long running tasks dying for a variety of reasons, and us having to manually kick things off again.

when a task runs, we store a key in redis. The key looks like `task_last_run_timestamp_{task_name}_{kwargs}` where kwargs are `key_value` separeated by underscores. The value is the timestamp of when the task completed.

Then, every morning at 6am BST we can run a task that checks the timestamps of all the individual day tasks. If that timestamp is older than midnight, the task didn't run this morning, so kick it off again. To kick it off again it just runs the parent tasks with a new flag "only_run_missing_days", to keep code duplication down.

The reason that each task and param gets an individual key is so that each one can be given an expiry date. If it was a hash set inside redis, then we wouldn't have a nice automated way of expiring individual keys without doing the work ourselves.

I've set the expiry date to four days, so that we can check on things after the weekend if everything goes really wrong.

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -1,10 +1,12 @@
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, date
 
+import iso8601
 from flask import current_app
 from notifications_utils.statsd_decorators import statsd
 from notifications_utils.timezones import convert_utc_to_bst
 
-from app import notify_celery
+from app.utils import get_london_midnight_in_utc
+from app import notify_celery, redis_store
 from app.config import QueueNames
 from app.cronitor import cronitor
 from app.dao.fact_billing_dao import (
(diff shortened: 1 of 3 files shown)


app/celery/reporting_tasks.py
```diff
@@ -67,42 +110,52 @@ def create_nightly_billing_for_day(process_day):
...
     # letters get modified for a longer time period than sms and email, so we need to reprocess for more days
     for i in range(10):
         process_day = yesterday - timedelta(days=i)
-        create_nightly_notification_status_for_day.apply_async(
-            kwargs={'process_day': process_day.isoformat(), 'notification_type': LETTER_TYPE},
-            queue=QueueNames.REPORTING
-        )
+        tasks_to_run.append({'process_day': process_day.isoformat(), 'notification_type': LETTER_TYPE})
+
+    child_task_name = create_nightly_billing_for_day.name
```
servingUpAces, line 138: shouldn't this be create_nightly_notification_status_for_day.name?

tests/app/celery/test_reporting_tasks.py
```diff
@@ -506,3 +511,98 @@ def test_create_nightly_notification_status_for_day_respects_bst(sample_template
...
+    mock_billing = mocker.patch('app.celery.reporting_tasks.create_nightly_billing')
+
+    rerun_failed_nightly_tasks()
+
+    mock_status.assert_called_once_with(only_run_missing_days=True)
+    mock_billing.assert_called_once_with(only_run_missing_days=True)
+
+
+@freeze_time('2020-06-05 05:00')
+def test_create_nightly_billing_only_run_missing_days_flag(notify_api, mocker):
```
servingUpAces, line 589: Can we have a similar test for create_nightly_notification_status, please?

servingUpAces:
I like this approach

leohemsted:
this has problem - delete task will have run by the time this retries, so it should take care of that. we'll need to update the create-x-for-day tasks to have an optional "has_delete_task_run" flag or similar, which it can pass through `fetch_billing_data_for_day`/`fetch_notification_status_for_day`, so it can set it in `get_notification_table_to_use` properly. ughhhh

leohemsted:
This hasn't been an issue for a while, so we're closing this as it's relatively complex.

My recommendation, if this does come back, is we can add these checks but simply log an exception (which'll fire off an email) if the task isn't run.

No point spending time automating stuff if we dont know how common the problem it's solving is

PR #2855, opened 2020-05-28
Add property to contact lists to say if they’ve ever been used

quis:
At the moment we return a count of recent jobs for contact lists, where recent is defined as being within the service’s data retention period.

This lets us write nice bits of UI copy like ‘used 3 times in the last 7 days’. But it’s hard to write the copy for when the count is 0, because this could be for one of two reasons:
- the contact list has never been used
- the contact list has been used, but not within the data retention period for that channel

At the moment we can’t know which of those reasons is the case, so we can’t write nice clear content like ‘never been used’.

This commit adds a property to contact lists which says whether they’ve ever been used.

It also renames the existing, as-yet-unused property to make clear that it’s only counting within the data retention (so can still be 0 even if `has_jobs` is `True`).

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2136,13 +2136,19 @@ class ServiceContactList(db.Model):
             )
         ).count()
 
+    def get_has_jobs(self):
+        return bool(Job.query.filter(
+            Job.contact_list_id == self.id,
+        ).first())
+
     def serialize(self):
         created_at_in_bst = convert_utc_to_bst(self.created_at)
         contact_list = {
             "id": str(self.id),
             "original_file_name": self.original_file_name,
             "row_count": self.row_count,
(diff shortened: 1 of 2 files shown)


app/models.py
```diff
@@ -2152,13 +2152,19 @@ def get_job_count(self):
             )
         ).count()
 
+    def get_has_jobs(self):
```
tombye, line 2155: Could this be called `has_been_used`, to follow your description? If not, should it be `has_jobs` rather than `get_has_jobs`? Doesn't seem like something you'd think you could set.

quis: @tombye I had a think about this and decided the best thing to do was turn this method (and the one above it) into a property. See  313b69f9

PR #2856, opened 2020-05-28
Capture detailed delivery receipt status from MMG

CrystalPea:
This PR starts logging detailed delivery status for MMG. 

This is a good start, but I think we should do more (in a separate PR):

1. we should add a `detailed_status` for notifications in the database and save this additional info there.
2. we should then see what is the break down of delivery errors we get. There are a few errors in particular that we should look out for:
- "2053": "Source address blacklisted" <--- could mean us or MMG are on some blacklist
- "29": "Invalid delivery report" <--- we don't really know if message has been delivered or not
- "11": "Sender id not registered" <--- I don't think this one is very likely, the full description is: "Your account is set to accept only registered sender IDs, and the sender ID defined in the
request has not been registered on your account."
- "13": "Sender id blacklisted" <--- this I think is something that has happened in the past?
- "19": "Routing unavailable" <--- what does that even mean? And why with Firetext it's a temporary-failure, but for MMG it's permanent failure?
- "20": "Rejected by anti-flooding mechanism" <--- if we get many of those we might want to contact the sender?
- "21": "System error" <--- error description says to retry those messages or contact support. Should we retry and count them and subtract them from our bill? 😬 

And those are unlikely to happen hopefully, but if they do happen, we should investigate:
- "23": "Duplicate message id"
- "24": "Message formatted incorrectly"
- "25": "Message too long"
- "51": "Missing recipient value"
- "52": "Invalid destination"

Then if we see some worrying patterns we have more scope to discuss them with MMG.

Story ticket: https://www.pivotaltracker.com/story/show/170971077

### DIFF
--- a/app/celery/process_sms_client_response_tasks.py
+++ b/app/celery/process_sms_client_response_tasks.py
@@ -24,7 +24,7 @@ sms_response_mapper = {
 
 @notify_celery.task(bind=True, name="process-sms-client-response", max_retries=5, default_retry_delay=300)
 @statsd(namespace="tasks")
-def process_sms_client_response(self, status, provider_reference, client_name, code=None):
+def process_sms_client_response(self, status, provider_reference, client_name, detailed_status_code=None):
     # validate reference
     try:
         uuid.UUID(provider_reference, version=4)
@@ -36,7 +36,7 @@ def process_sms_client_response(self, status, provider_reference, client_name, c
 
     # validate status
     try:
-        notification_status, detailed_status = response_parser(status, code)
+        notification_status, detailed_status = response_parser(status, detailed_status_code)
         current_app.logger.info(
             f'{client_name} callback returned '
             f'status of {notification_status}: {detailed_status} for reference: {provider_reference}'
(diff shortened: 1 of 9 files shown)


tests/app/clients/test_firetext.py
```diff
@@ -8,19 +8,25 @@
...
 
 
-def test_should_return_correct_details_for_complaint():
-    get_firetext_responses('2') == 'pending'
+def test_get_firetext_responses_should_return_correct_details_for_complaint():
+    assert get_firetext_responses('2') == ('pending', None)
 
 
-def test_should_be_none_if_unrecognised_status_code():
+def test_get_firetext_responses_should_be_none_if_unrecognised_status_code():
```
idavidmcdonald, line 29: Option: Rename test to say a be a bit clearer about what it is returning ie `test_get_firetext_responses_should_raise_error_if_unrecognised_status_code`

app/clients/sms/mmg.py
```diff
@@ -4,15 +4,49 @@
...
+        "51": "Missing recipient value",
+        "52": "Invalid destination",
+    }},
 }
 
 
-def get_mmg_responses(status):
-    return mmg_response_map[status]
+def get_mmg_responses(status, substatus=None):
+    return (mmg_response_map[status]["status"], mmg_response_map[status]["substatus"].get(substatus, None))
```
idavidmcdonald, line 49: Optional, for the return of this function you are explicit about calling it a tuple by adding the brackets. In `get_firetext_responses` you don't add the brackets. Both end up resulting in a tuple being returned but would be nice to have the consistency I think in terms of formatting

https://note.nkmk.me/en/python-function-return-multiple-values/

app/celery/process_sms_client_response_tasks.py
```diff
@@ -36,9 +36,10 @@ def process_sms_client_response(self, status, provider_reference, client_name, c
 
     # validate status
     try:
-        notification_status = response_parser(status)
+        notification_status, detailed_status = response_parser(status, code)
```
idavidmcdonald, line 39: Optional: We have terminology `code`, `substatus`, `reason`, `detailed_status`. Worth thinking about if these could or should be more consistent. It may not be appropriate though so give it some thought

PR #2858, opened 2020-05-29
add raw request timings to provider send functions

leohemsted:
we're using statsd to monitor how long provider requests are taking. However, there's lots of busy work that happens inside our statsd metrics timing window. Things like json dumping and loading, building headers, exception handling, etc. I'm worried that our metrics might not be telling us the true story of where slowness is.

for firetext/mmg, the response object from requests has an elapsed property [1], which captures from sending raw data to parsing the response headers. for ses, it's a bit trickier, but boto3 exposes a few event hooks [2]. it's hard to find them without stepping through the code, but the interesting ones are before-call, after-call, after-call-error, request-created, and response-received. The before-call and after-call involve some marshalling, built-in retrying, etc, while request-created and response-received are much lower level. They might be called more than once per ses request, if boto3 itself retries the request on 5xx, 429 and low level socket errors [3].

Add these as new `raw-request-time` metrics rather than overwriting to avoid changing the meaning of an existing metric, and to let us compare the metrics to see if there's a noticeable difference at all

[1] https://requests.readthedocs.io/en/master/api/#requests.Response.elapsed
[2] https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html
[3] https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#legacy-retry-mode

### DIFF
--- a/app/clients/email/aws_ses.py
+++ b/app/clients/email/aws_ses.py
@@ -54,6 +54,22 @@ class AwsSesClient(EmailClient):
         self.name = 'ses'
         self.statsd_client = statsd_client
 
+        # events are generally undocumented, but some that might be of interest are:
+        # before-call, after-call, after-call-error, request-created, response-received
+        self._client.meta.events.register(f'request-created.ses.SendEmail', self.ses_request_created_hook)
+        self._client.meta.events.register(f'response-received.ses.SendEmail', self.ses_response_received_hook)
+
+    def ses_request_created_hook(self, **kwargs):
+        # request created may be called multiple times if the request auto-retries. We want to count all these as the
+        # same request for timing purposes, so only reset the start time if it was cleared completely
+        if self.ses_start_time == 0:
+            self.ses_start_time = monotonic()
+
+    def ses_response_received_hook(self, **kwargs):
+        # response received may be called multiple times if the request auto-retries, however, we want to count the last
+        # time it triggers for timing purposes, so always reset the elapsed time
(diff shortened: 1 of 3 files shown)


leohemsted:
Testing locally, not under load:

|          | existing metric     | new metric         | %age diff |
|----------|---------------------|--------------------|-----------|
| mmg      | 0.2673354550206568  | 0.262755           | 1.7%      |
| firetext | 0.19947180000599474 | 0.194072           | 2.7%      |
| ses      | 0.5020331389969215  | 0.4997435169934761 | 0.45%     |

app/clients/email/aws_ses.py
```diff
@@ -114,6 +132,8 @@ def send_email(self,
             elapsed_time = monotonic() - start_time
             current_app.logger.info("AWS SES request finished in {}".format(elapsed_time))
             self.statsd_client.timing("clients.ses.request-time", elapsed_time)
+            if self.ses_elapsed_time != 0:
```
idavidmcdonald, line 135: What could cause the case where the elapsed time is 0?

leohemsted: don't think we'll get anything since we're already catching all generic `Exception` objects above, but was just an abundance of caution

leohemsted: similarly i don't think we'll ever get a `response` object that exists but doesn't have `elapsed`, but better safe than sorry

PR #2862, opened 2020-06-03
Stub SES email client to avoid hitting SES during load testing

idavidmcdonald:
If we set a value for `SES_STUB_URL` environment variable, we can stub out calls to SES and send
them to our own stub app. If the environment variable is not set, things
work as normal.

To be used alongside
https://github.com/alphagov/notifications-email-provider-stub/pull/1

How to test:
1. Run the email stub locally - https://github.com/alphagov/notifications-email-provider-stub/tree/email-stub
2. Then run this app
```SES_STUB_URL=http://localhost:6301/ses ./scripts/run_celery.sh``` plus the HTTP API server and the admin app
3. Visit http://localhost:6012/ and trigger the admin app to send an email.
4. Check the notification just created in your DB now has a UUID set as it's reference (it's the one generated by the email provider stub).

When `SES_STUB_URL` is not set, the app should work normally, like the stub is not even there.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -88,10 +88,10 @@ def create_app(application):
     aws_ses_stub_client.init_app(
         application.config['AWS_REGION'],
         statsd_client=statsd_client,
-        stub_url=os.environ.get('SES_STUB_URL')
+        stub_url=application.config['SES_STUB_URL']
     )
     # If a stub url is provided for SES, then use the stub client rather than the real SES boto client
-    email_clients = [aws_ses_client] if not os.environ.get('SES_STUB_URL') else [aws_ses_stub_client]
+    email_clients = [aws_ses_stub_client] if application.config['SES_STUB_URL'] else [aws_ses_client]
     clients.init_app(sms_clients=[firetext_client, mmg_client], email_clients=email_clients)
 
     notify_celery.init_app(application)
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -21,6 +21,7 @@ from app.celery.celery import NotifyCelery
 from app.clients import Clients
 from app.clients.document_download import DocumentDownloadClient
 from app.clients.email.aws_ses import AwsSesClient
+from app.clients.email.aws_ses_stub import AwsSesStubClient
 from app.clients.sms.firetext import FiretextClient
 from app.clients.sms.mmg import MMGClient
 from app.clients.performance_platform.performance_platform_client import PerformancePlatformClient
@@ -49,6 +50,7 @@ notify_celery = NotifyCelery()
 firetext_client = FiretextClient()
 mmg_client = MMGClient()
 aws_ses_client = AwsSesClient()
+aws_ses_stub_client = AwsSesStubClient()
 encryption = Encryption()
 zendesk_client = ZendeskClient()
 statsd_client = StatsdClient()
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: app/clients/email/stub_email.py [hunk 1]
```diff
@@ -0,0 +1,49 @@
+import json
+
+from flask import current_app
+from requests import request
+from time import monotonic
+
+from app.clients.email import (EmailClientException, EmailClient)
+
+
+class AwsSesStubClientException(EmailClientException):
```
CrystalPea, line 10: do we want to keep this function empty?

app/clients/email/stub_email.py
```diff
@@ -0,0 +1,49 @@
...
+                   subject,
+                   body,
+                   html_body='',
+                   reply_to_address=None):
+        try:
+            start_time = monotonic()
+            response = request(
+                "POST",
+                self.url,
+                data={"id": "dummy-data"},
```
CrystalPea, line 35: we had a brief discussion if we want to send realistic data to the stub, but we have not reached a conclusion, right? What do you gather?

app/clients/email/stub_email.py
CrystalPea: ok, it's just a namespace

app/__init__.py
```diff
@@ -81,13 +83,22 @@ def create_app(application):
...
     mmg_client.init_app(application, statsd_client=statsd_client)
+
     aws_ses_client.init_app(application.config['AWS_REGION'], statsd_client=statsd_client)
+    aws_ses_stub_client.init_app(
+        application.config['AWS_REGION'],
+        statsd_client=statsd_client,
+        stub_url=os.environ.get('SES_STUB_URL')
+    )
+    # If a stub url is provided for SES, then use the stub client rather than the real SES boto client
+    email_clients = [aws_ses_client] if not os.environ.get('SES_STUB_URL') else [aws_ses_stub_client]
```
servingUpAces, line 94: Should we add something to ensure the stub is not used in production?

CrystalPea: Good call!

CrystalPea: I added a condition for that now 👍

app/__init__.py
```diff
@@ -81,13 +83,25 @@ def create_app(application):
...
     mmg_client.init_app(application, statsd_client=statsd_client)
+
     aws_ses_client.init_app(application.config['AWS_REGION'], statsd_client=statsd_client)
+    aws_ses_stub_client.init_app(
+        application.config['AWS_REGION'],
+        statsd_client=statsd_client,
+        stub_url=os.environ.get('SES_STUB_URL')
+    )
+    # If a stub url is provided for SES, then use the stub client rather than the real SES boto client
+    if os.environ.get('SES_STUB_URL') and notify_environment not in ['live', 'production']:
```
leohemsted, line 94: you should use `application.config['SES_STUB_URL']` rather than `os.environ` - just to keep with the pattern that we only fetch from `os.environ` in config.py.

also, we could hard code `SES_STUB_URL = None` in the `Live` config, that way we wouldn't need this gross `notify_environment` check as even if it was set, it would be overwritten by class inheritance in prod.

idavidmcdonald:
@leohemsted @servingUpAces Think this is everyones comments addressed, let me know what you think please!

PR #2864, opened 2020-06-08
Drop postage constraints

klssmith:
We've decided to drop the postage constraints, so this adds a migration to do that (the constraint on notifications and notification_history has already been dropped in production and preview).

Also removes a few tests that are no longer needed and updates the JSON schemas for the new postage values we want to allow.

### DIFF
--- a/migrations/versions/0321_drop_postage_constraints.py
+++ b/migrations/versions/0321_drop_postage_constraints.py
@@ -1,0 +1,69 @@
+"""
+
+Revision ID: 0321_drop_postage_constraints
+Revises: 0320_optimise_notifications
+Create Date: 2020-06-08 11:48:53.315768
+
+"""
+import os
+
+from alembic import op
+
+
+revision = '0321_drop_postage_constraints'
+down_revision = '0320_optimise_notifications'
+environment = os.environ['NOTIFY_ENVIRONMENT']
+
+
+def upgrade():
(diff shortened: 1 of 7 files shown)


migrations/versions/0321_drop_postage_constraints.py
```diff
@@ -0,0 +1,28 @@
...
+    if environment not in ["live", "production"]:
+        op.execute('ALTER TABLE notifications DROP CONSTRAINT IF EXISTS chk_notifications_postage_null')
+        op.execute('ALTER TABLE notification_history DROP CONSTRAINT IF EXISTS chk_notification_history_postage_null')
+
+    op.execute('ALTER TABLE templates DROP CONSTRAINT IF EXISTS chk_templates_postage')
+    op.execute('ALTER TABLE templates_history DROP CONSTRAINT IF EXISTS chk_templates_history_postage')
+
+
+def downgrade():
+    pass
```
leohemsted, line 28: how come we're not re-adding them here?

klssmith: Just because you can't add them back in a single migration that is safe to run in production. I suppose it is possible to do a `if environment not in ["live", "production"]:` in the downgrade though - what do you think?

klssmith: Now updated to add in a downgrade that does not run in production

PR #2870, opened 2020-06-10
[work in progress] Cache API keys and service is active checks

tlwr:
***This is a work in progress**

Before each request for v2 notifications we make a database round trip to check whether the API key is present/valid/correct

We only care if API keys are eventually consistent, ie if someone deletes an API key and it is used for 5 more seconds after is deleted, this is not a big deal.

We can use a new dependency (cachetools) to cache these calls to the database, effectively trading memory for database roundtrips. We will cache the "service JOIN api_keys" database call that is performed for every request (only for v2 notifications)

This commit introduces a TTL cache which will cache these calls for 2s. I expect this to marginally improve performance in cases when a service is posting many notifications for a sustained period of time

I'm not entirely sure how effective this will be in an using eventlets, but it is worth testing IMO

I chose this dao because it is the first dao usage in the notifications codepath

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -66,7 +66,17 @@ metrics = GDSMetrics()
 clients = Clients()
 
 api_user = LocalProxy(lambda: _request_ctx_stack.top.api_user)
-authenticated_service = LocalProxy(lambda: _request_ctx_stack.top.authenticated_service)
+
+
+def lazy_authenticated_service():
+    svc = _request_ctx_stack.top.authenticated_service
+    if callable(svc):
+        return svc()
+    return svc
+
+
+authenticated_service = LocalProxy(lazy_authenticated_service)
+authenticated_service_id = LocalProxy(lambda: _request_ctx_stack.top.authenticated_service_id)
 
 CONCURRENT_REQUESTS = Gauge(
     'concurrent_web_request_count',
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -8,6 +8,10 @@ from sqlalchemy.exc import DataError
 from sqlalchemy.orm.exc import NoResultFound
 from gds_metrics import Histogram
 
+from app.cache import (
+    api_keys_for_service_cache,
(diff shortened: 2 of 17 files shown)


tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1005,6 +1006,10 @@ def test_post_notifications_saves_email_normally_if_save_email_to_queue_fails(cl
         "template_id": template.id,
         "personalisation": {"message": "Dear citizen, have a nice day"}
     }
+
+    # Caching
+    time.sleep(5)
```
quis, line 1011: You can use [freezegun](https://pypi.org/project/freezegun/) as a context manager to control time during tests:
```python

with freeze_time('1999-12-31'):
    pass  # make first call

with freeze_time('2000-01-01'):
    pass  # make second call

Comment (tlwr) [reply tests/app/v2/notifications/test_post_notifications.py:1011]: Thanks! I intend on fixing these tests properly once I fix the lazy loading, and adding a test that auth is eventually consistent

**Comment by tlwr** [comment]
I've rebased this onto https://github.com/alphagov/notifications-api/pull/2873

GitHub has absolutely wrecked the actual history again

```
Newest

fb7487ad Combine requires_auth and requires_auth_with_cache
a6feaa67 Pass around service id in request context
68891acd Make caches thread safe
65197c1a Add cache for service permissions
52134a50 Some caching in requires_auth_with_cache action
51e321eb add service is active cache
6c8fc375 add api keys cache
f8a6099d duplicate inline requires_auth before action
c73d2d87 ------------------------------------------------------------------------
chill -> 31973acc Don’t store the underlying dict
chill -> 4b1aee50 Remove a bunch of cruft from JSONModel
chill -> 3e8ecac9 Don’t interface with serialised data directly
chill ->21bd0be4 Serialise template immediately after fetching

Oldest
```

tlwr:
This has been superseded by #2887 and #2891

PR #2872, opened 2020-06-11
Remove the use of schedule_for in post_notifications.

servingUpAces:
Years ago we started to implement a way to schedule a notification. We hit a problem but we never came up with a good solution and the feature never made it back to the top of the prioriry list.

This PR removes the code for scheduled_for. There will be another PR to drop the scheduled_notifications table.

Unforetunately, I don't think we can remove the `scheduled_for` attribute from the notification.serialized method because out clients might fail if something is missing. For now I have left it in but defaulted the value to None.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -677,12 +677,6 @@ def dao_get_notifications_by_references(references):
     ).all()
 
 
-@statsd(namespace="dao")
-def dao_created_scheduled_notification(scheduled_notification):
-    db.session.add(scheduled_notification)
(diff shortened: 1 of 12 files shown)


klssmith:
You can now drop the `scheduled_for` field from the `post_email_request` and `post_sms_request` schemas too.

There's also the `schedule_notifications` entry in the `service_permission_types` table which isn't needed, but that could be removed when you drop the `ScheduledNotification` table.

servingUpAces:
closing in favour of https://github.com/alphagov/notifications-api/pull/2897

PR #2873, opened 2020-06-11
Serialise template as soon as possible 

quis:
This commit changes the code in post notification endpoint to handle a serialised template (ie a `dict`) rather than a database object.

This is the first step towards being able to cache the template and not hit the database on every request.

There should be no functional changes here, it’s just refactoring.

There are some changes to the tests where the signature of functions has changed.

Importing of the template schema has to be done at a function level, otherwise Marshmallow gets weird.


app/notifications/validators.py
```diff
@@ -147,17 +148,28 @@ def check_notification_content_is_not_empty(template_with_content):
...
         )
     except NoResultFound:
         message = 'Template not found'
         raise BadRequestError(message=message,
                               fields=[{'template': message}])
 
+    return template_schema.dump(fetched_template).data
+
+
+def get_template_model(template_id, service_id):
```
quis, line 166: We can add a short-lived in memory cache here, for example:
```suggestion
@cachetools.cached(cache=cachetools.TTLCache(ttl=5))
def get_template_model(template_id, service_id):
```

app/notifications/validators.py
```diff
@@ -147,17 +148,28 @@ def check_notification_content_is_not_empty(template_with_content):
         raise BadRequestError(message=message)
 
 
-def validate_template(template_id, personalisation, service, notification_type):
+def get_template_dict(template_id, service_id):
```
quis, line 151: We can add a long-lived cache here, which can be shared with, and invalidated by the admin app
```suggestion
@request_cache.set('template-{template_id}-version-None')
def get_template_dict(template_id, service_id):
```

app/json_models.py
```diff
@@ -0,0 +1,24 @@
+class JSONModel():
```
idavidmcdonald, line 1: Technically this class has nothing to do with JSON yet, just dictionaries. Is it named with JSON for a particular reason?

app/notifications/process_notifications.py
```diff
@@ -44,7 +49,34 @@
 
 
 def create_content_for_notification(template, personalisation):
-    template_object = template._as_utils_template_with_personalisation(personalisation)
+    if template.template_type == EMAIL_TYPE:
+        template_object = PlainTextEmailTemplate(
+            {
```
idavidmcdonald, line 54: Wonder if instead of explicitely having the logic here for what fields we expect to pass to the different template types we should just pass in all the keys for the template and then the class can decide which it wants to use?

Something like

```
PlainTextEmailTemplate(
    template.serialize()
    personalisation
)
```

app/notifications/validators.py
```diff
@@ -147,17 +148,28 @@ def check_notification_content_is_not_empty(template_with_content):
...
             template_id=template_id,
-            service_id=service.id
+            service_id=service_id
         )
     except NoResultFound:
         message = 'Template not found'
         raise BadRequestError(message=message,
                               fields=[{'template': message}])
 
+    return template_schema.dump(fetched_template).data
```
idavidmcdonald, line 163: Cool, so the template_schema is responsible for making sure `template.reply_to_text` has been worked out by calling `Template.get_reply_to_text`. Took me a little while to track that down

app/v2/notifications/post_notifications.py
```diff
@@ -169,6 +169,7 @@ def post_notification(notification_type):
             reply_to_text=reply_to
         )
 
+        # Think this is redundant
         template_with_content.values = notification.personalisation
```
idavidmcdonald, line 173: Shall we try removing it?

app/notifications/validators.py
idavidmcdonald: I'd be tempted to roll the cache change as a separate deploy so we can compare performance, thoughts?

app/notifications/validators.py
idavidmcdonald: Just to clarify, the risk here is that someone was sending notifications out, made a change to their template and then sent another notification within 5 seconds then it would use the old template? I think if we swapped to something like 2 seconds the risk of a user being quick enough to hit that edge case would be lower but I'm not opposed to keeping it at 5.

This would be good for GOV.UK though, reducing say 1000 DB calls to get templates over 5 seconds down to 1 DB call.

idavidmcdonald:
Oh also, I think it might benefit from squashing some commits, the first commit goes down one approach and then the second kind of takes it back in a different direction for example.

app/notifications/validators.py
quis: Agree, this was just to show what that change would look like, and why it makes sense to split the code up into these specific functions.

app/v2/notifications/post_notifications.py
```diff
@@ -102,13 +102,12 @@ def post_precompiled_letter_notification():
...
 
-    reply_to = get_reply_to_text(LETTER_TYPE, form, template)
-
     notification = process_letter_notification(
         letter_data=form,
         api_key=api_user,
         template=template,
-        reply_to_text=reply_to,
+        service=authenticated_service,
+        reply_to_text='',
```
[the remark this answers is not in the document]
quis: This bit of the code only deals with precompiled letters. Precompiled letters have a hidden template:
https://github.com/alphagov/notifications-api/blob/ca5cfe8d018003deccfce60af56549bb24685b56/app/v2/notifications/post_notifications.py#L97

…but they don’t use any of the content, subject or reply-to text (the contact block in the top-right of the letter) to render the letter.

app/v2/notifications/post_notifications.py
quis: It’s not redundant, it’s updating the personalisation to include any document download links. Will remove my comment.

app/notifications/validators.py
quis: Yeah, exactly. Can talk about the exact timings when we add the caching, but basing it on how quickly a user could edit a template then send themselves a message would be good.

app/notifications/validators.py
```diff
@@ -71,7 +71,7 @@ def check_template_is_for_notification_type(notification_type, template_type):
 
 
 def check_template_is_active(template):
```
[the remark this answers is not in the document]
quis: Resolving as outdated

app/v2/notifications/post_notifications.py
quis: Done

app/notifications/process_notifications.py
quis: Reason for doing it this way is deliberate, see 4d041f7

app/json_models.py
quis: have renamed to `SerialisedModel` in 2969d8d4 and  db7c59fe

app/serialised_models.py
```diff
@@ -0,0 +1,57 @@
+from abc import ABC, abstractmethod
+
+from app.dao import templates_dao
+
+
+class SerialisedModel(ABC):
+
+    """
+    A SerialisedModel takes a dictionary, typically created by
```
idavidmcdonald, line 9: This is good

app/serialised_models.py
```diff
@@ -0,0 +1,57 @@
...
+        'postage',
+        'process_type',
+        'reply_to_text',
+        'subject',
+        'template_type',
+        'version',
+    }
+
+    @classmethod
+    def from_id_and_service_id(cls, template_id, service_id):
```
idavidmcdonald, line 43: Optional: Consider adding a unit test for this class to make sure `from_id_and_service_id` and `get_dict` do what we expect them to

PR #2875, opened 2020-06-15
[needs work] Use opentelemetry project to instrument v2 post notifications using Jaeger

tlwr:
What
----

[OpenTelemetry](https://opentelemetry.io/) is a cross-platform set of APIs and SDKs and extensions for adding telemetry to applications.

The OpenTelemetry python libraries seem to work okay with `eventlet`.

This PR isn't production ready yet but could be useful in its current state for looking for performance improvements.

It adds a new package `app.tracing` which has a decorator `trace_request` and a convenience function `span`. Using `span` when not in a traced request is a runtime error.

Example usage
----

If you wanted to instrument:

```
@app.route('/do-two-things')
def do_two_things():
  do_thing_1()
  do_thing_2()
  return "done"
```

You would do:

```
@app.route('/do-two-things')
@trace_request
def do_two_things():
  with span('do-thing-1'):
    do_thing_1()
  with span('do-thing-2'):
    do_thing_2()

  return "done"
```

Which would generate a trace which looks like:

```
|========== total-request ==========|
 |== do-thing-1 ==|
                  |== do-thing-2 ==|
```

Instrumentation
----

When running the Jaeger all-in-one image locally, and running the tests, the API now generates traces for v2/notifications POSTs like:

image

Next steps
----

I want to think about this a bit more, and get comments/thoughts/feedback - you have done some tracing work already, as I may have duplicated what you've already done.

This also needs:

- Tests
- Tuning - such that we don't trace 100% of requests
- Proper configuration
- Rethinking how the current tracer is grabbed, the request context isn't good long term because it will prevent us from instrumenting daos low level

### DIFF
--- a/app/status/healthcheck.py
+++ b/app/status/healthcheck.py
@@ -7,12 +7,14 @@ from flask import (
 from app import db, version
 from app.dao.services_dao import dao_count_live_services
 from app.dao.organisation_dao import dao_count_organisations_with_live_services
+from app.tracing import trace_request
 
 status = Blueprint('status', __name__)
 
 
 @status.route('/', methods=['GET'])
 @status.route('/_status', methods=['GET', 'POST'])
+@trace_request
 def show_status():
     if request.args.get('simple', None):
         return jsonify(status="ok"), 200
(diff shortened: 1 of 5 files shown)


tlwr:
Needs a rebase and there are other things to do

PR #2877, opened 2020-06-16
This PR optimises the queries for notification by reference.

servingUpAces:
By added the notification_type to the filter we get a better performance on the query. Especially when selecting for NotificationHistory.

### DIFF
--- a/tests/app/dao/notification_dao/test_notification_dao.py
+++ b/tests/app/dao/notification_dao/test_notification_dao.py
@@ -28,7 +28,6 @@ from app.dao.notifications_dao import (
     update_notification_status_by_id,
     update_notification_status_by_reference,
     dao_get_notification_by_reference,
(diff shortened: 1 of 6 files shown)


tests/app/dao/notification_dao/test_notification_dao.py
```diff
@@ -1649,7 +1655,13 @@ def test_dao_get_notification_or_history_by_reference_with_multiple_matches_rais
         dao_get_notification_or_history_by_reference('REF1', 'letter')
 
 
-def test_dao_get_notification_or_history_by_reference_with_no_matches_raises_error(notify_db):
+def test_dao_get_notification_or_history_by_reference_with_no_matches_raises_error(sample_letter_template):
```
idavidmcdonald, line 1658: Are these test names the wrong way round? This test is about no matches for type but the one below is named that

PR #2878, opened 2020-06-16
Make sure people without international sms permission can send to crown dependencies

CrystalPea:
In this PR we make sure that people without international sms permission can send to crown dependencies after we start treating those numbers as international here: https://github.com/alphagov/notifications-utils/pull/749

A Jersey (tv) number, and my (standard UK) number, both sent to successfully, from a service without international sms permission, and with correct flags resulting in correct providers used:

Screenshot 2020-06-16 at 18 00 48

TODO: Import this version utils once they are merged: https://github.com/alphagov/notifications-utils/pull/749
Tests will fail without it.

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -120,11 +120,7 @@ def validate_and_format_recipient(send_to, key_type, service, notification_type,
     service_can_send_to_recipient(send_to, key_type, service, allow_whitelisted_recipients)
 
     if notification_type == SMS_TYPE:
-        international_phone_info = get_international_phone_info(send_to)
-
-        if international_phone_info.international and \
-                INTERNATIONAL_SMS_TYPE not in [p.permission for p in service.permissions]:
-            raise BadRequestError(message="Cannot send to international mobile numbers")
+        international_phone_info = check_if_service_can_send_to_number(service, send_to)
 
         return validate_and_format_phone_number(
             number=send_to,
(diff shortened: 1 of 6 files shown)


app/notifications/validators.py
```diff
@@ -134,6 +130,18 @@ def validate_and_format_recipient(send_to, key_type, service, notification_type,
         return validate_and_format_email_address(email_address=send_to)
 
 
+def check_if_service_can_send_to_number(service, number):
+    international_phone_info = get_international_phone_info(number)
+
+    if (
+        # if number is international and not a crown dependency
+        international_phone_info.international and international_phone_info.country_prefix != '44'
```
quis, line 138: I’m not too keen on having to look at the country prefix here. Having to look at the underlying data suggests that the properties of `international_phone_info` aren’t named well any more.

If you re-wrote your comment as code it would be clearer:
```python
if international_phone_info.international and not international_phone_info.crown_depedency:
```

Even then I’m not sure that’s totally accurate because Jersey and Guernsey aren’t a different nation.

Alternatively this code could stay the same and the code which sends the message to the provider could have something like:
```python
if international_phone_info.send_using_international_provider:
```

What do you think?

CrystalPea: to this thread: we decided to ship the code as it is to fix the bug, and add crown_depedency field to international_phone_info object next week.

tests/app/notifications/test_process_notification.py
```diff
@@ -327,7 +327,7 @@ def test_simulated_recipient(notify_api, to_address, notification_type, expected
 @pytest.mark.parametrize('recipient, expected_international, expected_prefix, expected_units', [
     ('7900900123', False, '44', 1),  # UK
     ('+447900900123', False, '44', 1),  # UK
-    ('07700900222', False, '44', 1),  # UK
+    ('07700900222', True, '44', 1),  # UK
```
quis, line 330: Should add a comment here to make it clear why `expectd_international` is `True` for a UK number:
```suggestion
    ('07700900222', True, '44', 1),  # UK (Jersey)
```

PR #2882, opened 2020-06-17
Reduce the use of db connections

servingUpAces:
In the effort to reduce the number of queries to the database this PR removes the need to fetch the notification and service after the commit when persisting the notification, this releases the db connection back to the connection pool.

If you look at the queries made during the POST /notification/email or sms you should see no new queries made after the insert into notifications transaction.

The letter flow will need some more work in order to remove the need for the db after the notification insert. This PR prioritises the email and sms flow first.

### DIFF
--- a/tests/app/notifications/test_process_notification.py
+++ b/tests/app/notifications/test_process_notification.py
@@ -23,7 +23,7 @@ from app.notifications.process_notifications import (
 )
 from notifications_utils.recipients import validate_and_format_phone_number, validate_and_format_email_address
 from app.v2.errors import BadRequestError
-from tests.app.db import create_service, create_template
+from tests.app.db import create_service, create_template, create_api_key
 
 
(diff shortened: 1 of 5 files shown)


tests/app/notifications/test_process_notification.py
```diff
@@ -197,44 +193,77 @@ def test_persist_notification_with_optionals(sample_job, sample_api_key, mocker)
...
-        api_key_id=sample_api_key.id,
-        key_type=sample_api_key.key_type,
+        api_key_id=api_key.id,
+        key_type=api_key.key_type,
+        reference="ref2")
+
+    mock_incr.assert_called_once_with(str(service.id) + "-2016-01-01-count", )
+
+
+def test_persist_notification_does_not_increments_cache_live_service(
```
idavidmcdonald, line 244: Optional: Could parametrise this test with the previous one to keep things DRY but this is very optional

PR #2888, opened 2020-06-19
Add international letter rates

klssmith:
**Migration to add international letter rates**
New rows giving the prices of letters with a post_class of `europe` and `rest-of-world` have been added to the `letter_rates` table. All rates are currently the same for international letters.

**Update letter data for usage-for-all-services report**
Usage for all services is a platform admin report that groups letters by postage. We want it to show `europe` and `rest-of-world` letters under a single category of `international`, so this updates the query to do that and to order appropriately.

[Pivotal story](https://www.pivotaltracker.com/story/show/171622825)


app/dao/fact_billing_dao.py
```diff
@@ -457,6 +458,10 @@ def get_rate(
     if notification_type == LETTER_TYPE:
         if letter_page_count == 0:
             return 0
+
+        if post_class in INTERNATIONAL_POSTAGE_TYPES:
```
idavidmcdonald, line 462: Optional, if we are to keep this approach then I would suggest an inline comment here saying
'we only have one billing rate for international letters so all postage types in INTERNATIONAL_POSTAGE_TYPES get mapped to it`

app/platform_stats/rest.py
```diff
@@ -106,3 +108,10 @@ def get_usage_for_all_services():
         x['organisation_name'],
         x['service_name']
     )))
+
+
+def postage_description(postage):
+    if postage in UK_POSTAGE_TYPES:
+        return f'{postage} class'
+    else:
+        return 'international'
```
idavidmcdonald, line 117: Optional: Do we want to break it down further into europe and rest of world here? Or is that even possible? I haven't looked to see if the letter_breakdown returns it broken down into europe and rest of world or just international?

app/dao/fact_billing_dao.py
klssmith: This commit has been removed, since we're now adding a separate row for `europe` and `rest-of-world` letters to the `letter_rates` table.

app/platform_stats/rest.py
klssmith: It would be broken down into `europe` and `rest-of-world` by default. We want to combine them though so that international letters show up as a single row in [this report](https://www.notifications.service.gov.uk/platform-admin/reports/usage-for-all-services)

klssmith:
After discussing it, I've updated this to have a row in `letter_rates` for `europe` and a row for `rest-of-world` since it will be simpler to change if the two rates ever do become different.

klssmith:
I've fixed the platform admin report to group international letters together, so this can be re-reviewed now

PR #2881, opened 2020-06-17
Serialise service, API keys and permissions at start of post notifications endpoint

quis:
By serialising these straight away we can:
- not go back to the database later, potentially closing the connection sooner
- potentially cache the serialised data, meaning we don’t touch the database at all

Branched off, and using the same techniques as:
- [x] https://github.com/alphagov/notifications-api/pull/2873

PR #2889, opened 2020-06-19
Commit the db session before sending notification to the queue.

servingUpAces:
When a high volume service posts a notification we want to commit the db session before we send the message to SQS. This will release the db connection earlier.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -15,8 +15,8 @@ from app import (
     notify_celery,
     document_download_client,
     encryption,
-    DATETIME_FORMAT
-)
+    DATETIME_FORMAT,
+    db)
 from app.celery.letters_pdf_tasks import get_pdf_for_templated_letter, sanitise_letter
 from app.celery.research_mode_tasks import create_fake_letter_response_file
 from app.celery.tasks import save_api_email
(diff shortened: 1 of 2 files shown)


app/v2/notifications/post_notifications.py
```diff
@@ -266,23 +271,26 @@ def save_email_to_queue(
...
+    api_key_id,
+    key_type,
+    template_id,
+    template_version,
     service_id,
     personalisation,
     document_download_count,
     reply_to_text=None
 ):
+    db.session.commit()
```
leohemsted, line 283: is there any way we can check if the session has been used, and only trigger if that happens? I think if we want to make a pattern of this I'd like it to be as reusable and plug-and-play-able as possible. i've done a bit of digging but haven't got anywhere.

what i would like is us to be able to set a flag, such that if we then create a new connection, it raises an exception. Or something like that. So we're aware of what's happening.

I worry that we're just making our code a bit too brittle, but not in a way that will lead to errors. But someone will add `service.id` somewhere, and the code will work great but all the performance gains we get will disappear as it suddenly consumes a second connection.

Not sure if there's anything smart we can do with the sqlalchemy event hooks or anything 🤔

leohemsted: also, at the very least, can you leave a comment here explaining why you're commiting. And leave a comment in `post_notifications` where save_email_to_queue is called explaining that it commits, and if we need stuff from the db we'll need to get it beforehand

quis: This is what https://github.com/alphagov/notifications-api/pull/2881 gives us. It:
- commits the session as soon as we no longer need it
- guarantees no subsequent database access, because `service` is no longer a SQLAlchemy object

servingUpAces: I'm not completely happy about having a commit here. But since this got in before the top down work Chris was doing, it ended up here. 
We can use db.session.expunge() which will throw an exception if you access a db object after the commit. However, we commit a notification in many places so introducing something like that will be a bit tricky.

PR #2891, opened 2020-06-22
Cache serialised template in Redis and in memory

quis:
This pull request caches just the serialised template object or dictionary in memory and Redis respectively.

It should improve performance for high volume services by releasing the DB connection more quickly.

### DIFF
--- a/tests/app/v2/notifications/test_post_notifications.py
+++ b/tests/app/v2/notifications/test_post_notifications.py
@@ -280,6 +280,37 @@ def test_should_cache_template_lookups_in_redis(mocker, client, sample_template)
     assert mock_redis_set.call_args[1]['ex'] == 604_800
 
 
+def test_should_return_template_if_found_in_redis(mocker, client, sample_template):
+
+    from app.schemas import template_schema
+    template_dict = template_schema.dump(sample_template).data
+
+    mocker.patch(
+        'app.redis_store.get',
+        return_value=json.dumps(template_dict).encode('utf-8')
+    )
+    mock_get_template = mocker.patch(
+        'app.dao.templates_dao.dao_get_template_by_id_and_service_id'
+    )
+
+    mocker.patch('app.celery.provider_tasks.deliver_sms.apply_async')
+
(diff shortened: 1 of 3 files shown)


tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -211,6 +212,73 @@ def test_notification_reply_to_text_is_original_value_if_sender_is_changed_after
...
+
+    for i in range(5):
+        auth_header = create_authorization_header(service_id=sample_template.service_id)
+        client.post(
+            path='/v2/notifications/sms',
+            data=json.dumps(data),
+            headers=[('Content-Type', 'application/json'), auth_header]
+        )
+
+    assert mock_get_template.call_args_list == [
```
servingUpAces, line 236: Optional: add `assert mock_get_template.call_count == 1` to be crystal clear.

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -211,6 +212,73 @@ def test_notification_reply_to_text_is_original_value_if_sender_is_changed_after
...
+        call(service_id=sample_template.service_id, template_id=str(sample_template.id))
+    ]
+    assert Notification.query.count() == 5
+
+
+def test_should_cache_template_lookups_in_redis(mocker, client, sample_template):
+
+    from app.schemas import template_schema
+
```
servingUpAces, line 245: Optional: Add a mock_get_template and assert it's not called.

tests/app/v2/notifications/test_post_notifications.py
quis, line 236:
```suggestion
    assert mock_get_template.call_count == 1
    assert mock_get_template.call_args_list == [
```

tests/app/v2/notifications/test_post_notifications.py
quis: That’s actually a separate test case (for when the template is already cached), I’ve added it in af1b021

PR #2901, opened 2020-06-25
Turn off statsd wrapper for synchronous statsd calls during POSTs

idavidmcdonald:
This commit removes the statsd wrapper function for
`dao_create_notification`. This is the only function that uses the
statsd wrapper during the processing of the HTTP POST to create either
an email or a text message. I've determined that this is the only
function by running the API locally, putting a log statement in the
statsd client and then POSTing notifications. Note, other functions
use the statsd wrapper as part of the subsequent sending of that
notification but that is done by the celery workers asynchronously.

I've removed this, and only this statsd wrapper, for three reasons:
- if there are anymore unforseen issues, either with the time it takes
  to resolve the DNS or with just managing to successfully resolve the
  DNS of statsd, then this will never happen as part of the creating of
  notifications which makes up the majority of traffic to our API and it's
  the code path we want to have the most reliable performance on.
- I haven't removed it from any of the other code paths in the API
  because these won't make up the majority of calls to StatsD so won't
reduce the amount of DNS resoloution needed
- I haven't removed it from the celery tasks, if DNS resoloution fails
  or is slow for these we don't mind as much as they will replay or not
 have as much user impact.
- We don't actually use this metric, I don't think anyone looks at the
  graph so there is little harm of dropping it. If we want we can talk
  about which metrics to move into Prometheus in the future but can drop
  this one sooner rather than later at little cost and just further
  reducing our risk of getting caught out again


leohemsted:
This won't impact govuk email as that happens from celery.

we also call statsd in a before_request handler to log the user agent: https://github.com/alphagov/notifications-api/blob/master/app/__init__.py#L271-L273

we also call statsd twice from utils after the request (to record status codes and response times) https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/logging.py#L73-L78

PR #2911, opened 2020-06-30
Clean up

servingUpAces:
This PR tidies up some of the code after the SerialisedService was introduced.

Removing the need to put service attributes in a local variable.


app/v2/notifications/post_notifications.py
```diff
@@ -175,7 +173,7 @@ def process_sms_or_email_notification(
     form_send_to = form['email_address'] if notification_type == EMAIL_TYPE else form['phone_number']
 
     send_to = validate_and_format_recipient(send_to=form_send_to,
-                                            key_type=api_key.key_type,
+                                            key_type=api_user.key_type,
```
quis, line 176: Where is `api_user` coming from here? I can’t see it defined above, or in the arguments this function takes.

servingUpAces: It's in the request context, set when we authenticate request, https://github.com/alphagov/notifications-api/blob/3954cd0e4a7a14824781740f1343b114738665ca/app/authentication/auth.py#L134

quis: `:magic:` – yup, this is good to go in that case.

PR #2914, opened 2020-07-03
Broadcast DB models

leohemsted:
#### add `broadcast` to template_type

this is the trickiest bit of this PR, if only because lots of places refer to template_type all over the place. I think I've caught everywhere, but not sure. This also involved resurrecting enum migration code from https://github.com/alphagov/notifications-api/blob/master/migrations/versions/0060_add_letter_template_type.py

#### add `broadcast_data` to template

this'll contain everything that we need to save on templates for a broadcast message. We don't know what that is yet, so it's just a jsonb field

### New model

BroadcastMessage model. A broadcast has a bunch of states - see the commit message for the proposed state diagram. I've tried to add as many timestamps/user ids in to the model, I wanted to avoid using an `updated_at` to refer to different things based on the state.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2148,3 +2148,65 @@ class ServiceContactList(db.Model):
             "created_at": created_at_in_bst.strftime("%Y-%m-%d %H:%M:%S"),
         }
         return contact_list
+
+
+class BroadcastStatusType(db.Model):
+    __tablename__ = 'broadcast_status_type'
+    DRAFT = 'draft'
+    PENDING_APPROVAL = 'pending-approval'
+    REJECTED = 'rejected'
+    BROADCASTING = 'broadcasting'
+    COMPLETED = 'completed'
+    CANCELLED = 'cancelled'
+    TECHNICAL_FAILURE = 'technical-failure'
+
+    STATUSES = [DRAFT, PENDING_APPROVAL, REJECTED, BROADCASTING, COMPLETED, CANCELLED, TECHNICAL_FAILURE]
+
+    name = db.Column(db.String, primary_key=True)
+
+
+class BroadcastMessage(db.Model):
+    __tablename__ = 'broadcast_message'
+    __table_args__ = (
+        db.ForeignKeyConstraint(
+            ['template_id', 'template_version'],
+            ['templates_history.id', 'templates_history.version'],
+        ),
+        {}
+    )
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True)
+
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'))
+    service = db.relationship('Service', backref='broadcast_messages')
+
+    template_id = db.Column(UUID(as_uuid=True), nullable=False)
+    template_version = db.Column(db.Integer, nullable=False)
+    template = db.relationship('TemplateHistory', backref='broadcast_messages')
+
+    _personalisation = db.Column(db.String, nullable=True)
+
+    status = db.Column(
+        db.String,
+        db.ForeignKey('broadcast_status_type.name'),
+        nullable=False,
+        default=BroadcastStatusType.DRAFT
+    )
+
+    # these times are related to the actual broadcast, rather than auditing purposes
+    starts_at = db.Column(db.DateTime, nullable=True)
+    finishes_at = db.Column(db.DateTime, nullable=True)  # isn't updated if user cancels
+
+    # these times correspond to when
+    created_at = db.Column(db.DateTime, nullable=False)
+    approved_at = db.Column(db.DateTime, nullable=True)
+    cancelled_at = db.Column(db.DateTime, nullable=True)
+    updated_at = db.Column(db.DateTime, nullable=True, onupdate=datetime.datetime.utcnow)
+
+    created_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey('users.id'), nullable=False)
+    approved_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey('users.id'), nullable=True)
+    cancelled_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey('users.id'), nullable=True)
+
+    created_by = db.relationship('User', foreign_keys=[created_by_id])
+    approved_by = db.relationship('User', foreign_keys=[approved_by_id])
+    cancelled_by = db.relationship('User', foreign_keys=[cancelled_by_id])
--- a/app/models.py
+++ b/app/models.py
@@ -26,6 +26,7 @@ from notifications_utils.template import (
     PlainTextEmailTemplate,
     SMSMessageTemplate,
     LetterPrintTemplate,
+    BroadcastMessageTemplate,
 )
 from notifications_utils.timezones import convert_utc_to_bst
 
@@ -44,8 +45,10 @@ from app.history_meta import Versioned
 SMS_TYPE = 'sms'
 EMAIL_TYPE = 'email'
 LETTER_TYPE = 'letter'
+BROADCAST_TYPE = 'broadcast'
 
-TEMPLATE_TYPES = [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]
+TEMPLATE_TYPES = [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE, BROADCAST_TYPE]
+NOTIFICATION_TYPES = [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]  # not broadcast
 
 template_types = db.Enum(*TEMPLATE_TYPES, name='template_type')
 
@@ -299,12 +302,12 @@ UPLOAD_DOCUMENT = 'upload_document'
 EDIT_FOLDER_PERMISSIONS = 'edit_folder_permissions'
 UPLOAD_LETTERS = 'upload_letters'
 INTERNATIONAL_LETTERS = 'international_letters'
-BROADCAST_TYPE = 'broadcast'
 
 SERVICE_PERMISSION_TYPES = [
     EMAIL_TYPE,
     SMS_TYPE,
     LETTER_TYPE,
+    BROADCAST_TYPE,
     INTERNATIONAL_SMS_TYPE,
     INBOUND_SMS_TYPE,
     SCHEDULE_NOTIFICATIONS,
@@ -314,7 +317,6 @@ SERVICE_PERMISSION_TYPES = [
     EDIT_FOLDER_PERMISSIONS,
     UPLOAD_LETTERS,
     INTERNATIONAL_LETTERS,
-    BROADCAST_TYPE,
 ]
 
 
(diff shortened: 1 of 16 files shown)


app/models.py
```diff
@@ -995,7 +1000,7 @@ def serialize(self):
             "created_by": self.created_by.email_address,
             "version": self.version,
             "body": self.content,
-            "subject": self.subject if self.template_type != SMS_TYPE else None,
+            "subject": self.subject if self.template_type in {EMAIL_TYPE, LETTER_TYPE} else None,
```
CrystalPea, line 1003: why is it curly braces here, not the square braces?

leohemsted: curly is a `set` rather than a `list`. technically it's marginally quicker to lookup an item in a set rather than in a list (constant time, rather than dependent on the length of the list). In practice in this example it'll make little to no difference

CrystalPea: Ahh, ok, thanks! :)

app/models.py
```diff
@@ -2142,3 +2148,56 @@ def serialize(self):
...
+        {}
+    )
+
+    STATES = [
+        'draft',
+        'pending-approval',
+        'rejected',  # rejected before leaving notify
+        'broadcasting',  # approved, now live
+        'completed',  # reached end_date
+        'cancelled',  # the broadcast went out but then we sent a second cancel message to end it prematurely
```
servingUpAces, line 2169: A user needs to cancel a broadcast, does that need approval? Is there a state between requesting to cancel a broadcast and networks accepting that that request.

app/models.py
```diff
@@ -2142,3 +2148,56 @@ def serialize(self):
...
+            ['templates_history.id', 'templates_history.version'],
+        ),
+        {}
+    )
+
+    STATES = [
+        'draft',
+        'pending-approval',
+        'rejected',  # rejected before leaving notify
+        'broadcasting',  # approved, now live
```
servingUpAces, line 2167: We've decided to go from pending-approval to broadcasting... do we need a state to represent the time from clicking the approve button to the network accepting the broadcast and the broadcast has started?

app/models.py
```diff
@@ -2142,3 +2148,56 @@ def serialize(self):
...
+    CheckConstraint(f"state in ({','.join(STATES)})")
+
+    # these times are related to the actual broadcast, rather than auditing purposes
+    starts_at = db.Column(db.DateTime, nullable=True)
+    finishes_at = db.Column(db.DateTime, nullable=True)  # isn't updated if user cancels
+
+    # these times correspond to when
+    created_at = db.Column(db.DateTime, nullable=False)
+    approved_at = db.Column(db.DateTime, nullable=True)
+    cancelled_at = db.Column(db.DateTime, nullable=True)
```
servingUpAces, line 2194: If cancelled is a terminating state do we need a separate `cancelled_at`?
There is no real harm in having the extra column, but I don't think a terminating state needs a separate timestamp

app/models.py
```diff
@@ -44,8 +45,10 @@
 SMS_TYPE = 'sms'
 EMAIL_TYPE = 'email'
 LETTER_TYPE = 'letter'
+BROADCAST_TYPE = 'broadcast'
 
-TEMPLATE_TYPES = [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]
+TEMPLATE_TYPES = [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE, BROADCAST_TYPE]
```
servingUpAces, line 50: Have we made sure the new template_type doesn't affect the front end?

migrations/versions/0323_broadcast_message.py
```diff
@@ -0,0 +1,94 @@
...
+    'broadcasting',
+    'completed',
+    'cancelled',
+    'technical-failure',
+]
+states_with_quotes_and_commas = ', '.join(f"'{x}'" for x in STATES)
+
+
+def upgrade():
+    op.execute(f'ALTER TYPE {name} RENAME TO {tmp_name}')
```
servingUpAces, line 42: Historically we the db user for our applications didn't have permission to `alter type {name} add value {value}`
I wonder if that is still true.

migrations/versions/0323_broadcast_message.py
```diff
@@ -0,0 +1,94 @@
...
+        sa.Column('cancelled_by_id', postgresql.UUID(as_uuid=True)),
+
+        sa.ForeignKeyConstraint(['approved_by_id'], ['users.id'], ),
+        sa.ForeignKeyConstraint(['cancelled_by_id'], ['users.id'], ),
+        sa.ForeignKeyConstraint(['created_by_id'], ['users.id'], ),
+        sa.ForeignKeyConstraint(['service_id'], ['services.id'], ),
+        sa.ForeignKeyConstraint(['template_id', 'template_version'], ['templates_history.id', 'templates_history.version'], ),
+        sa.PrimaryKeyConstraint('id')
+    )
+    op.create_check_constraint(
```
servingUpAces, line 74: Why use a check constraint here? It's introducing a new way to handle column types.
We already have 2 different ways in our model which is unfortunate.

app/models.py
leohemsted: hmmm, i didn't wanted to have to update `finishes_at` (which the users sets in a form) if the user clicks the cancel button though. i think it might still be useful for us to know "they originally planned for this to be active for three days, but stopped after one day". Or would you just rely on `updated_at` and checking the state? i guess that's not too bad actually

app/models.py
quis: The frontend is ready to handle the new type, see: https://github.com/alphagov/notifications-admin/pull/3498

migrations/versions/0323_broadcast_message.py
leohemsted: just tried it out on preview (creating a new enum and table) and it works fine.

last time we tried one of these was april 2017 (https://github.com/alphagov/notifications-api/blob/master/migrations/versions/0078_add_sent_notification_status.py). I can't get the exact date but it looks like we'd been live on paas for a month or so at that point (https://github.com/alphagov/notifications-api/pull/870)

migrations/versions/0323_broadcast_message.py
leohemsted: that's a good point. i'll make it a `broadcast_status_type` table to be consistent with existing practices 👍

app/models.py
```diff
@@ -1004,6 +1009,7 @@ def serialize(self):
                 for key in self._as_utils_template().placeholders
             },
             "postage": self.postage,
+            "broadcast_data": self.broadcast_data
```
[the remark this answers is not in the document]
leohemsted: yeah this was one thing i wanted to chat about - template.serialize is used in a few places.

three options i can think of:

1. return `broadcast_data` in api calls. Won't need code change in the client libs, but may need docs changes.
1. create a new `template.serialize_for_v2_endpoint` or similar that doesn't include this field
1. only include the _key_ if the template_type is broadcast (so rather than being null, it just doesn't exist)

I think I prefer the second option out of these. Thoughts?

app/models.py
leohemsted: hmmm, i don't think there's a need for MVP. trying to think of something that's almost similar, we don't, for example, have a state for "antivirus has completed but validation hasn't happened yet" for precompiled letters - we get the `process-antivirus-passed` task and immediately begin validation.

app/models.py
leohemsted: don't think cancellation needs approval, but i'll double check with the big P.

no state between requesting cancellation and networks accepting. we set `cancelled` when the user clicks the button, then trigger the tasks. this state field isn't the place for figuring out whether the network received those messages (we'd need to have a new table with a row per provider at least)

app/models.py
servingUpAces: No we don't have pending-validation, but now that it happens as part of a new task we may want to introduce it. However, we haven't seen a need to do that yet. 

If we use `start_at` as a send broadcast message start at that time, then we may want a state to represent that. Not sure so probably not needed yet

app/models.py
servingUpAces: Either way works. 
If you don't think `cancelled` is a terminating state we should keep the `cancelled_at` column. 
If you think it makes things a lot clearer that's reason enough to keep it.
Also means that some broadcast_messages will never set that column.

app/models.py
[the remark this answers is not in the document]
servingUpAces: Option 2 is good because the public API endpoints should be different to our app to app endpoints.

app/models.py
leohemsted: i think that, based on our current understanding of the CAB model, `cancelled` should always be a terminating state. but i think i'll keep `cancelled_at`, as i think it'll help with clarity. i don't think we'll really know what the data should look like until we've sent a few messages (at least test messages) unfortunately

app/models.py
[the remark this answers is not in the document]
leohemsted: turns out we don't use `template.serialize` anywhere else anyway (we use marshmallow schemas), so i just renamed it to `serialize_for_v2` for clarity

app/models.py
```diff
@@ -986,7 +991,7 @@ def _as_utils_template_with_personalisation(self, values):
         template.values = values
         return template
 
-    def serialize(self):
+    def serialize_for_v2(self):
```
servingUpAces, line 994: Why change the name? For clarity?

app/models.py
```diff
@@ -2142,3 +2147,65 @@ def serialize(self):
...
+    id = db.Column(UUID(as_uuid=True), primary_key=True)
+
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'))
+    service = db.relationship('Service', backref='broadcast_messages')
+
+    template_id = db.Column(UUID(as_uuid=True), nullable=False)
+    template_version = db.Column(db.Integer, nullable=False)
+    template = db.relationship('TemplateHistory', backref='broadcast_messages')
+
+    _personalisation = db.Column(db.String, nullable=True)
```
servingUpAces, line 2186: Will _personalisation only be used for place holders in the template?

app/models.py
leohemsted: yup - i did it to validate to myself that no-where else uses `serialize` (that might need broadcast stuff for admin)

app/models.py
leohemsted: Yup, that's my intention. There _may_ be some mandatory place holders (eg, for things like title), but we'll cross that bridge when we understand more about what the networks require

migrations/versions/0323_broadcast_message.py
leohemsted: update: misread this, learnt about `add value`, updated the upgrade script

leohemsted: update update: You can't run `alter table {name} add value {value}` from within a transaction. (This is changed in postgres v12)
This means the migration script isn't repeatable if it fails. I don't really like that, I would rather keep the existing way.

app/models.py
servingUpAces: I hope we don't need mandatory place holders

PR #2917, opened 2020-07-07
Broadcast Message CRUD endpoints

leohemsted:
# CRU~D~

* get single message
* get all messages for service
* create broadcast message
* update broadcast message (finish dates, personalisation, etc)
* update broadcast message state (approve, cancel, etc)

schemas included. tests included


app/models.py
```diff
@@ -2209,3 +2209,40 @@ class BroadcastMessage(db.Model):
...
+        self._personalisation = encryption.encrypt(personalisation or {})
+
+    def serialize(self):
+        return {
+            'id': self.id,
+
+            'service_id': self.service_id,
+
+            'template_id': self.template_id,
+            'template_version': self.template_version,
```
quis, line 2230: Can we add template name here?

app/broadcast_message/rest.py
```diff
@@ -0,0 +1,123 @@
...
+        return iso8601.parse_date(dt).replace(tzinfo=None)
+    return dt
+
+
+@broadcast_message_blueprint.route('', methods=['GET'])
+def get_broadcast_messages_for_service(service_id):
+    service = dao_fetch_service_by_id(service_id)
+    # TODO: should this return template content/data in some way? or can we rely on them being cached admin side.
+    # we might need stuff like template name for showing on the dashboard.
+    # TODO: should this paginate or filter on dates or anything?
```
quis, line 42: I think the dashboard should work like this:
- waiting for approval (as many as exist – no pagination)
- live (as many as exist – no pagination)
- previous (up to 50, then pagination)

So if there are more than 50 previous broadcasts then clicking ‘next →‘ will take you to page 2, which only has previous broadcasts.

This could either work by:
- admin app makes 3 calls to the API
- or admin app makes 1 call and API responds with a `dict` that has 3 keys

I’m happy to do either, or make no changes here for now and let the admin app do grouping but not pagination.

leohemsted: i'm gonna do nothing for now because we don't have 1 broadcast message, yet alone 50. but at some point it should start ordering by .... `updated_at` i think. not sure whether admin making three calls or the response being richer is preferable, but i think the admin making three calls is simpler to understand maybe?

quis: 👍 to doing nothing for now

app/broadcast_message/rest.py
```diff
@@ -0,0 +1,123 @@
...
+    broadcast_messages = [o.serialize() for o in service.broadcast_messages]
+    return jsonify(broadcast_messages=broadcast_messages)
+
+
+@broadcast_message_blueprint.route('/<uuid:broadcast_message_id>', methods=['GET'])
+def get_broadcast_message(service_id, broadcast_message_id):
+    # TODO: should this return template content/data in some way?
```
quis, line 49: Fine for admin to make a separate request for the template.

app/broadcast_message/rest.py
```diff
@@ -0,0 +1,123 @@
...
+    data = request.get_json()
+
+    validate(data, update_broadcast_message_status_schema)
+    broadcast_message = dao_get_broadcast_message_by_id_and_service_id(broadcast_message_id, service_id)
+
+    new_status = data['status']
+
+    # TODO: Restrict status transitions
+    # TODO: Do we need to validate that the user belongs to the same service, isn't the creator, has permissions, etc?
+    # or is that admin's job
```
quis, line 110: I think we should have the safety of checking in both place. But fine to leave as a TODO for now.

app/broadcast_message/broadcast_message_schema.py
```diff
@@ -0,0 +1,46 @@
...
+
+update_broadcast_message_schema = {
+    '$schema': 'http://json-schema.org/draft-04/schema#',
+    'description': 'POST update broadcast_message schema',
+    'type': 'object',
+    'title': 'Update broadcast_message',
+    'properties': {
+        'personalisation': {'type': 'object'},
+        'starts_at': {'type': 'string', 'format': 'date-time'},
+        'finishes_at': {'type': 'string', 'format': 'date-time'},
```
quis, line 29: This needs to allow updating the areas of the broadcast (while it’s in draft). Complete replacement only, the admin app can handle merging and de-duping new/existing areas.

tests/app/broadcast_message/test_rest.py
```diff
@@ -0,0 +1,259 @@
...
+        _expected_status=400
+    )
+
+    assert response['errors'] == [{
+        'error': 'ValidationError',
+        'message': 'Additional properties are not allowed (status was unexpected)'
+    }]
+
+
+def test_update_broadcast_message_status(admin_request, sample_service):
```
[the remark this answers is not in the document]
leohemsted: yeah - actual restriction logic is coming later. there's definitely something around who can edit the template vs who can approve it - i don't think `first_x` and `second_x` is a scalable solution though. Maybe only the person who creates the broadcast can edit/mark as pending review, and person B can only click approve, and not edit.

tests/app/broadcast_message/test_rest.py
```diff
@@ -0,0 +1,259 @@
...
+    t = create_template(sample_service, BROADCAST_TYPE)
+    bm = create_broadcast_message(t, status=BroadcastStatusType.DRAFT)
+
+    response = admin_request.post(
+        'broadcast_message.update_broadcast_message_status',
+        _data={'status': BroadcastStatusType.PENDING_APPROVAL, 'created_by': str(t.created_by_id)},
```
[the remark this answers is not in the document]
leohemsted: yeah, i agree. its an unfortunate pattern that stems from the admin app using this function to add the user id to a json blob. definitely a candidate for a refactor though. https://github.com/alphagov/notifications-admin/blob/2c33325dc247b52c7b72b34ff1d67b30f8d24904/app/notify_client/__init__.py#L12-L16

app/broadcast_message/broadcast_message_schema.py
```diff
@@ -0,0 +1,48 @@
...
+    'type': 'object',
+    'title': 'Create broadcast_message',
+    'properties': {
+        'template_id': uuid,
+        'service_id': uuid,
+        'created_by': uuid,
+        'personalisation': {'type': 'object'},
+        'starts_at': {'type': 'string', 'format': 'datetime'},
+        'finishes_at': {'type': 'string', 'format': 'datetime'},
+        'areas': {"type": "array", "items": {"type": "string"}},
```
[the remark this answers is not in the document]
leohemsted: i thought that too - since we're moving away from building up a session and towards just saving in the DB and then updating, that makes it a lot simpler. i don't think there's really too much risk from this though

app/broadcast_message/rest.py
```diff
@@ -0,0 +1,123 @@
...
+from app.errors import register_errors
+from app.models import BroadcastMessage, BroadcastStatusType
+from app.broadcast_message.broadcast_message_schema import (
+    create_broadcast_message_schema,
+    update_broadcast_message_schema,
+    update_broadcast_message_status_schema,
+)
+from app.schema_validation import validate
+
+broadcast_message_blueprint = Blueprint(
```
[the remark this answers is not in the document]
leohemsted: yup. that's coming after the demo - for now, the restrictions are in place on the admin app and that's enough for now

PR #2925, opened 2020-07-09
Send broadcast

leohemsted:
A new `send_broacast_message` task, triggered when a user hits approve (or send as it currently reads). doesn't trigger on cancellation yet.

the task takes a brodcast_message_id, and makes a post to the cbc-proxy for now, hardcode the url to the notify stub. the stub requires template as the admin/api get it, so use the marshmallow schema to json dump it. Note - this also required us to tweak the BroadcastMessage.serialize function so that it converts uuids in to ids - flask's jsonify function does that for free but requests.post doesn't sadly.

if the request fails (either 4xx or 5xx) just raise an exception and let it bubble up for now - in the future we'll add retry logic

### DIFF
--- a/app/broadcast_message/rest.py
+++ b/app/broadcast_message/rest.py
@@ -1,8 +1,9 @@
 from datetime import datetime
 
 import iso8601
-from flask import Blueprint, jsonify, request
+from flask import Blueprint, jsonify, request, current_app
 
+from app.config import QueueNames
 from app.dao.templates_dao import dao_get_template_by_id_and_service_id
 from app.dao.users_dao import get_user_by_id
 from app.dao.broadcast_message_dao import (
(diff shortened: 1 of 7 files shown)


app/broadcast_message/rest.py
```diff
@@ -118,6 +119,15 @@ def update_broadcast_message_status(service_id, broadcast_message_id):
...
 
+    current_app.logger.info(
+        f'broadcast_message {broadcast_message_id} moving from {broadcast_message.status} to {new_status}'
+    )
     dao_update_broadcast_message(broadcast_message)
 
+    if new_status == BroadcastStatusType.BROADCASTING:
+        send_broadcast_message.apply_async(
+            kwargs={'broadcast_message_id': str(broadcast_message.id)},
+            queue=QueueNames.NOTIFY
```
idavidmcdonald, line 130: Oh was this queue just picked as a short term solution and in the medium term we will put all the CBC related tasks on to their own queue?

PR #2926, opened 2020-07-10
Add better logging to starting collate-letter-pdfs-to-be-sent

idavidmcdonald:
This will help us better understand how far through the task has got if
it gets interrupted halfway (as was the case this morning and we
struggled to understand).

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -129,6 +129,7 @@ def collate_letter_pdfs_to_be_sent():
     that have not yet been sent.
     If run after midnight, it will collect up letters created before 5:30pm the day before.
     """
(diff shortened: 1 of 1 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -167,6 +169,9 @@ def collate_letter_pdfs_to_be_sent():
                 queue=QueueNames.PROCESS_FTP,
                 compression='zlib'
             )
+        current_app.logger.info(f"finished collate-letter-pdfs-to-be-sent processing for postage class {postage}")
+
+    current_app.logger.info("finished collate-letter-pdfs-to-be-sent")
```
klssmith, line 174: I think we can already tell that a task has finished from this log line https://github.com/alphagov/notifications-api/blob/master/app/celery/celery.py#L36 but it also doesn't do any harm to have more logging

PR #2927, opened 2020-07-10
Add gds-pre-commit file to repo

idavidmcdonald:
https://github.com/alphagov/gds-pre-commit

This should help us avoid accidently checking in any real secrets. It
has detected secrets in other repos at GDS so worth us trialling.

As part of this, it is suggested that we check in the
`.secrets.baseline` file as this will mean that the state of which
strings are safe is stored in the repo and won't be flagged as non safe
by any other team mates checking.

Interested to hear if you think this is worth adding?

image

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,0 +1,246 @@
+{
+  "exclude": {
+    "files": null,
+    "lines": null
+  },
+  "generated_at": "2020-09-07T15:19:39Z",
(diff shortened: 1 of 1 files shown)


idavidmcdonald:
Closing as GDS is no longer using gds pre commit. They suggest installing the generic yelp package for detect secrets instead

PR #2928, opened 2020-07-14
Rename API URLs for guest list to guest list

quis:
This is a small part of removing the term ‘white list’ from the API.

Once the admin app is pointed at these new URLs, we can remove the old ones.


quis:
@servingUpAces I’ve added some more commits which rename everything except the database table.

PR #2932, opened 2020-07-16
Broadcast crud validation

leohemsted:
Review commit by commit.

* forbid editing after broadcast message goes live …
* refactor status updating into its own function …
* enforce kwargs when creating user in tests …
* allow platform admins to approve their own messages …
* reject approvals from people outside your service …

### DIFF
--- a/app/broadcast_message/rest.py
+++ b/app/broadcast_message/rest.py
@@ -1,8 +1,7 @@
 from datetime import datetime
 
 import iso8601
-from flask import Blueprint, jsonify, request, current_app, abort
-
+from flask import Blueprint, jsonify, request, current_app
 from app.config import QueueNames
 from app.dao.templates_dao import dao_get_template_by_id_and_service_id
 from app.dao.users_dao import get_user_by_id
@@ -13,7 +12,7 @@ from app.dao.broadcast_message_dao import (
     dao_update_broadcast_message,
 )
 from app.dao.services_dao import dao_fetch_service_by_id
-from app.errors import register_errors
+from app.errors import register_errors, InvalidRequest
 from app.models import BroadcastMessage, BroadcastStatusType
 from app.celery.broadcast_message_tasks import send_broadcast_message
 from app.broadcast_message.broadcast_message_schema import (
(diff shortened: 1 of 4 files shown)


app/broadcast_message/rest.py
```diff
@@ -37,6 +37,41 @@ def _parse_nullable_datetime(dt):
     return dt
 
 
+def _update_broadcast_message(broadcast_message, new_status, updating_user):
```
klssmith, line 40: Does this function also need to check if the updating user has the right permissions for the service?

app/broadcast_message/rest.py
```diff
@@ -37,6 +37,41 @@ def _parse_nullable_datetime(dt):
...
+            f'User {updating_user.id} cannot approve broadcast_message {broadcast_message.id} from other service'
+        )
+
+    if new_status not in BroadcastStatusType.ALLOWED_STATUS_TRANSITIONS[broadcast_message.status]:
+        abort(
+            400,
+            f'Cannot move broadcast_message {broadcast_message.id} from {broadcast_message.status} to {new_status}'
+        )
+
+    # TODO: Restrict status transitions
```
klssmith, line 53: This comment looks like it no longer applies since it's been implemented above

app/models.py
```diff
@@ -2163,6 +2163,23 @@ class BroadcastStatusType(db.Model):
 
     STATUSES = [DRAFT, PENDING_APPROVAL, REJECTED, BROADCASTING, COMPLETED, CANCELLED, TECHNICAL_FAILURE]
 
+    EDITABLE_STATUSES = [DRAFT, PENDING_APPROVAL, REJECTED]
```
klssmith, line 2166: I got a bit confused at first reading through this PR because `EDITABLE_STATUSES` sounds like the status of the broadcast message can be changed, when it's the message itself which can be changed. You could maybe update the name just to make it extra clear what this refers to - up to you though.

leohemsted: Yeah, agreed that the name is a bit awkward.  On reflection I think it's probably good to contrast it to `LIVE_STATUSES` (as that's the key distinction), but `NON_LIVE_STATUSES` sounds a bit clunky - any suggestions?

klssmith: Maybe something like `EDITABLE_MESSAGE_STATUSES` to try and show it is a list of states where the message can be edited, not a list of states where the status can be edited. But it's hard to think of a good name, and I'm not sure that `EDITABLE_MESSAGE_STATUSES` is an improvement, so feel free to ignore the earlier comment.

app/broadcast_message/rest.py
leohemsted: yeah, probably - but i'll leave that until we understand a bit more about what kind of roles we have in a broadcast team. Not quite sure which permissions get to do what bit (eg: which permission do you need to propose a message? which do you need to click approve? will these teams just give every permission to every one, or will they really carefully control who gets what?

(current permissions:)

* manage_users
* manage_templates
* manage_settings
* send_texts
* send_emails
* send_letters
* manage_api_keys
* platform_admin
* view_activity

app/models.py
leohemsted: for now i changed them to `PRE_BROADCAST_STATUSES` as i think it's more useful to describe what they actually represent (messages that havent started broadcast yet), rather than how we intend to use the flag, which is a secondary effect

app/broadcast_message/rest.py
```diff
@@ -37,6 +37,40 @@ def _parse_nullable_datetime(dt):
     return dt
 
 
+def _update_broadcast_message(broadcast_message, new_status, updating_user):
+    if updating_user not in broadcast_message.service.users:
+        abort(
```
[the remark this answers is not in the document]
leohemsted: good point. just didnt think of invalidrequest. i prefer that, i'll change it

PR #2935, opened 2020-07-22
Stop raising error in scheduled task

servingUpAces:
We have a scheduled task to check that all the jobs have completed, this will catch if an app is shut down and the job is complete yet, we only wait 10 seconds before forcing the app to shut down.

The task was raising a JobIncompleteError, yet it is not an error the task is performing correctly and calling the appropriate task to restart the job.
Also used apply_sync to create the task instead of send_task.


servingUpAces:
There isn't a different between the apply_async and send_task, they both put a task on the queue. But using apply_async is nicer to read. We only need to use the send_task when the task is in a different repo and we can import the method, like the tasks in template_preview or the ftp app. I only changed it for readability and consistency.

PR #2936, opened 2020-07-24
add broadcast_event table

leohemsted:
It's clear that we need a way to track updates to a broadcast message. It's also clear that we'll need some kind of audit log that captures exactly what was sent out in a message.

This commit adds a new database table, `broadcast_event`, which maps 1:1 with CAP XML sent to the CBCs. We'll create one of these just before sending out.

The main driver for this was that cancel and update messages need to contain a list of references of all previous messages that they're amending. This is of format `{sender},{identifier},{sent_timestamp}`, and the identifier itself needs to be unique for each message.

--------

I've left lots of questions/thoughts in the comments

```python
# user creates draft
broadcast_message = BroadcastMessage(
    service_id=template.service_id,
    template_id=template.id,
    template_version=template.version,
    personalisation={},
    status=BroadcastStatusType.DRAFT,
    starts_at=None,
    finishes_at=None,
    created_by_id=None,
    areas=None,
)

# broadcast_message gets approved, we create a BroadcastEvent
original_alert_msg = BroadcastEvent(
    broadcast_message=broadcast_message_id
    msgType=alert
    transmitted_sender='covid-19@notify.gov.uk'
    transmitted_content = {'description': 'this is data from a template'},
    transmitted_areas=['england']
    transmitted_finishes_at=datetime(2020, 07, 23, 0, 0)
    transmitted_anything_else_derived_from_template_or_service=...?
    created_by_id = the_person_that_issued_the_thing?

    # not mvp
    status = 'draft', 'active'  # does this confuse things? do we need to think about clearing this up if people abandon updates etc?
)

# user updates content.
update_msg = BroadcastEvent(
    broadcast_message=broadcast_message_id  # same id
    msgType=update
    transmitted_content = {'description': 'this is data i wrote by hand'},
    transmitted_areas = ['england']  # unchanged, but we still copy it down
    transmitted_finishes_at = datetime(2020, 07, 23, 12, 0)
)

cancel_msg = BroadcastEvent(
    broadcast_message=broadcast_message_id  # same id
    msgType=cancel
    transmitted_content = None,
    transmitted_areas = ['england'],  # hunch: i think we'll always need this so provider knows which masts to send the message to.
    transmitted_finishes_at = None,
)
```

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2305,7 +2305,6 @@ class BroadcastEvent(db.Model):
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
 
-    # TODO: do we need this? or should we just join via broadcast_message.
     service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'))
     service = db.relationship('Service')
 
@@ -2318,18 +2317,18 @@ class BroadcastEvent(db.Model):
     # msgType. alert, cancel, or update. (other options in the spec are "ack" and "error")
     message_type = db.Column(db.String, nullable=False)
 
-    # reckon: this will be json containing {'headline': '...', 'description': '...', 'title': '...'}. anything that isnt
-    # hardcoded in utils/cbc proxy
+    # this will be json containing anything that isnt hardcoded in utils/cbc proxy. for now just body but may grow to
+    # include, eg, title, headline, instructions.
     transmitted_content = db.Column(
         JSONB(none_as_null=True),
-        nullable=True,
-        default=lambda: {'headline': '', 'description': '', 'title': ''}
+        nullable=True
     )
     # unsubstantiated reckon: even if we're sending a cancel, we'll still need to provide areas
     transmitted_areas = db.Column(JSONB(none_as_null=True), nullable=False, default=list)
     transmitted_sender = db.Column(db.String(), nullable=False)
 
-    # TODO: do we need this?
+    # we may only need this starts_at if this is scheduled for the future. Interested to see how this affects
+    # updates/cancels (ie: can you schedule an update for the future?)
     transmitted_starts_at = db.Column(db.DateTime, nullable=True)
     transmitted_finishes_at = db.Column(db.DateTime, nullable=True)
 
--- a/app/models.py
+++ b/app/models.py
@@ -39,6 +39,7 @@ from app import (
     encryption,
     DATETIME_FORMAT,
     DATETIME_FORMAT_NO_TIMEZONE)
+from app.utils import get_dt_string_or_none
 
 from app.history_meta import Versioned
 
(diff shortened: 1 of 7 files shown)


app/models.py
```diff
@@ -2251,3 +2251,85 @@ def serialize(self):
...
+    The only exception to this is that we will have to join to itself to find other broadcast_events with the
+    same broadcast_message_id when building up the `<references>` xml field for updating/cancelling an existing message.
+
+    As such, this shouldn't have foreign keys to things that can change or be deleted.
+    """
+    __tablename__ = 'broadcast_event'
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+
+    # TODO: do we need this? or should we just join via broadcast_message.
```
servingUpAces, line 2280: I think it's ok to leave it here. Similar to other models we have left a foreign key to service even if we can get it from another foreign key, for example template_id in notifications.

app/models.py
```diff
@@ -2251,3 +2251,85 @@ def serialize(self):
...
+    transmitted_content = db.Column(
+        JSONB(none_as_null=True),
+        nullable=True,
+        default=lambda: {'headline': '', 'description': '', 'title': ''}
+    )
+    # unsubstantiated reckon: even if we're sending a cancel, we'll still need to provide areas
+    transmitted_areas = db.Column(JSONB(none_as_null=True), nullable=False, default=list)
+    transmitted_sender = db.Column(db.String(), nullable=False)
+
+    # TODO: do we need this?
```
servingUpAces, line 2304: What will this represent? When we start sending to the MNOs. 
It's easier to remove a column later, let's keep it for now.

migrations/versions/0326_broadcast_event.py
```diff
@@ -0,0 +1,43 @@
...
+        sa.Column('transmitted_starts_at', sa.DateTime(), nullable=True),
+        sa.Column('transmitted_finishes_at', sa.DateTime(), nullable=True),
+        sa.ForeignKeyConstraint(['broadcast_message_id'], ['broadcast_message.id'], ),
+        sa.ForeignKeyConstraint(['service_id'], ['services.id'], ),
+        sa.PrimaryKeyConstraint('id')
+    )
+    # this shouldn't be nullable. it defaults to `[]` in python.
+    op.alter_column('broadcast_message', 'areas', existing_type=postgresql.JSONB(astext_type=sa.Text()), nullable=False)
+    # this can't be nullable. it defaults to 'draft' in python.
+    op.alter_column('broadcast_message', 'status', existing_type=sa.VARCHAR(), nullable=False)
```
servingUpAces, line 35: Isn't sa.VARCHAR() and sa.String() the same?

app/models.py
leohemsted: starts_at might be a time in the future if it's scheduled, I believe. Agree to leaving it in for now 👍

migrations/versions/0326_broadcast_event.py
leohemsted: yup - confirmed here https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.String

thats just the way alembic generated the alter statement tho - the thing that is changing is the nullable flag which was incorrectly set

PR #2943, opened 2020-08-04
Set international postage and international flag for international letters

servingUpAces:
Make sure to set the postage and international flag correctly for all letters. 

There are many ways to send a letter with Notify:

1. send one-off templated letter
2. upload CSV for a templated letter
3. send template letter via the API
4. send PDF letter via the API
5. upload PDF via the UI

Possible postage types are first | second | europe | rest-of-world.

### DIFF
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -107,7 +107,6 @@ def persist_notification(
     reply_to_text=None,
     billable_units=None,
     postage=None,
-    template_postage=None,
     document_download_count=None,
     updated_at=None
 ):
@@ -147,8 +146,8 @@ def persist_notification(
     elif notification_type == EMAIL_TYPE:
         notification.normalised_to = format_email_address(notification.to)
     elif notification_type == LETTER_TYPE:
-        notification.postage = postage or template_postage
-        notification.international = True if postage in INTERNATIONAL_POSTAGE_TYPES else False
+        notification.postage = postage
+        notification.international = postage in INTERNATIONAL_POSTAGE_TYPES
         notification.normalised_to = ''.join(notification.to.split()).lower()
 
     # if simulated create a Notification model to return but do not persist the Notification to the dB
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -27,7 +27,7 @@ from app.models import (
     LETTER_TYPE,
     NOTIFICATION_CREATED,
     Notification,
-)
+    INTERNATIONAL_POSTAGE_TYPES)
 from app.dao.notifications_dao import (
     dao_create_notification,
     dao_delete_notifications_by_id,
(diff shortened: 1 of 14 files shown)


app/notifications/process_notifications.py
```diff
@@ -147,7 +146,8 @@ def persist_notification(
     elif notification_type == EMAIL_TYPE:
         notification.normalised_to = format_email_address(notification.to)
     elif notification_type == LETTER_TYPE:
-        notification.postage = postage or template_postage
+        notification.postage = postage
+        notification.international = True if postage in INTERNATIONAL_POSTAGE_TYPES else False
```
quis, line 150:
```suggestion
        notification.international = (postage in INTERNATIONAL_POSTAGE_TYPES)
```

app/notifications/validators.py
```diff
@@ -214,3 +215,34 @@ def check_service_letter_contact_id(service_id, letter_contact_id, notification_
             message = 'letter_contact_id {} does not exist in database for service id {}'\
                 .format(letter_contact_id, service_id)
             raise BadRequestError(message=message)
+
+
+def validate_address(service, letter_data):
+    address = PostalAddress.from_personalisation(
+        letter_data,
+        allow_international_letters=(INTERNATIONAL_LETTERS in str(service.permissions)),
```
quis, line 223:
```suggestion
        allow_international_letters=service.has_permission(INTERNATIONAL_LETTERS),
```

app/notifications/process_notifications.py
servingUpAces: Ah - thanks. I meant to change that

app/notifications/validators.py
servingUpAces: service can either be a serialised json object or a db.Model type. If it's serialised there isn't a has_permissions object, if its a db.Model permissions is a resultset. So `(INTERNATIONAL_LETTERS in str(service.permissions))` works for both cases.

quis: Think there’s a better way (add a `has_permission` method to `SerialisedService`) but that’s a bigger refactor so best to leave out of this PR.

tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -646,6 +646,54 @@ def test_process_sanitised_letter_with_valid_letter(
...
+    template_preview_bucket_name = current_app.config['LETTER_SANITISE_BUCKET_NAME']
+    destination_bucket_name = current_app.config['LETTERS_PDF_BUCKET_NAME']
+    conn = boto3.resource('s3', region_name='eu-west-1')
+    conn.create_bucket(Bucket=scan_bucket_name)
+    conn.create_bucket(Bucket=template_preview_bucket_name)
+    conn.create_bucket(Bucket=destination_bucket_name)
+
+    s3 = boto3.client('s3', region_name='eu-west-1')
+    s3.put_object(Bucket=scan_bucket_name, Key=filename, Body=b'original_pdf_content')
+    s3.put_object(Bucket=template_preview_bucket_name, Key=filename, Body=b'sanitised_pdf_content')
```
quis, line 673: Shouldn’t this be mocked? Feels a bit weird to be connecting to an external service that could be unavailable from an integration test.

servingUpAces: That's what @mock_s3 does

quis: Oh, that’s fancy

PR #2944, opened 2020-08-04
Send events

leohemsted:
Now sends cancel messages.

When you click broadcasting or cancel, it creates new BroadcastEvents, and then sends them using the new send_broadcast_event task.

image

### DIFF
--- a/app/broadcast_message/rest.py
+++ b/app/broadcast_message/rest.py
@@ -14,7 +14,7 @@ from app.dao.broadcast_message_dao import (
 from app.dao.services_dao import dao_fetch_service_by_id
 from app.errors import register_errors, InvalidRequest
 from app.models import BroadcastMessage, BroadcastStatusType, BroadcastEvent, BroadcastEventMessageType
-from app.celery.broadcast_message_tasks import send_broadcast_message
+from app.celery.broadcast_message_tasks import send_broadcast_event
 from app.broadcast_message.broadcast_message_schema import (
     create_broadcast_message_schema,
     update_broadcast_message_schema,
@@ -196,7 +196,7 @@ def _create_broadcast_event(broadcast_message):
     # save to the DB
     dao_create_broadcast_message(event)
 
-    send_broadcast_message.apply_async(
-        kwargs={'broadcast_message_id': str(broadcast_message.id)},
+    send_broadcast_event.apply_async(
+        kwargs={'broadcast_event_id': str(event.id)},
         queue=QueueNames.NOTIFY
     )
--- a/app/broadcast_message/rest.py
+++ b/app/broadcast_message/rest.py
@@ -3,13 +3,12 @@ from datetime import datetime
 import iso8601
 from flask import Blueprint, jsonify, request, current_app
 from app.config import QueueNames
+from app.dao.dao_utils import dao_save_object
 from app.dao.templates_dao import dao_get_template_by_id_and_service_id
 from app.dao.users_dao import get_user_by_id
 from app.dao.broadcast_message_dao import (
-    dao_create_broadcast_message,
     dao_get_broadcast_message_by_id_and_service_id,
     dao_get_broadcast_messages_for_service,
-    dao_update_broadcast_message,
 )
 from app.dao.services_dao import dao_fetch_service_by_id
 from app.errors import register_errors, InvalidRequest
@@ -109,7 +108,7 @@ def create_broadcast_message(service_id):
         created_by_id=user.id,
     )
 
-    dao_create_broadcast_message(broadcast_message)
+    dao_save_object(broadcast_message)
 
     return jsonify(broadcast_message.serialize()), 201
 
@@ -137,7 +136,7 @@ def update_broadcast_message(service_id, broadcast_message_id):
     if 'areas' in data:
         broadcast_message.areas = data['areas']
 
-    dao_update_broadcast_message(broadcast_message)
+    dao_save_object(broadcast_message)
 
     return jsonify(broadcast_message.serialize()), 200
 
@@ -153,7 +152,7 @@ def update_broadcast_message_status(service_id, broadcast_message_id):
     updating_user = get_user_by_id(data['created_by'])
 
     _update_broadcast_message(broadcast_message, new_status, updating_user)
-    dao_update_broadcast_message(broadcast_message)
+    dao_save_object(broadcast_message)
 
     if new_status in {BroadcastStatusType.BROADCASTING, BroadcastStatusType.CANCELLED}:
         _create_broadcast_event(broadcast_message)
@@ -193,8 +192,7 @@ def _create_broadcast_event(broadcast_message):
         transmitted_finishes_at=transmitted_finishes_at,
     )
 
-    # save to the DB
-    dao_create_broadcast_message(event)
+    dao_save_object(event)
 
     send_broadcast_event.apply_async(
         kwargs={'broadcast_event_id': str(event.id)},
--- a/app/broadcast_message/rest.py
+++ b/app/broadcast_message/rest.py
@@ -13,7 +13,7 @@ from app.dao.broadcast_message_dao import (
 )
 from app.dao.services_dao import dao_fetch_service_by_id
 from app.errors import register_errors, InvalidRequest
-from app.models import BroadcastMessage, BroadcastStatusType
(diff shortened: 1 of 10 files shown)


app/broadcast_message/rest.py
```diff
@@ -155,10 +155,48 @@ def update_broadcast_message_status(service_id, broadcast_message_id):
...
+    # TODO: This doesn't support placeholders yet. We shouldn't use BroadcastMessageTemplate when we add placeholders
+    # as that just outputs XML, we need the raw text.
+    event = BroadcastEvent(
+        service=broadcast_message.service,
+        broadcast_message=broadcast_message,
+        message_type=msg_types[broadcast_message.status],
+        transmitted_content={"body": broadcast_message.template.content},
+        transmitted_areas=broadcast_message.areas,
+        # TODO: Probably move this somewhere more standalone too and imply that it shouldn't change. Should it include
+        # a service based identifier too? eg "flood-warnings@notifications.service.gov.uk" or similar
```
CrystalPea, line 187: so if it shouldn't change, we should probs store it in config?

But I like the idea of using from address for service as an identifier.

app/broadcast_message/rest.py
```diff
@@ -155,10 +155,48 @@ def update_broadcast_message_status(service_id, broadcast_message_id):
...
+        transmitted_sender='notifications.service.gov.uk',
+
+        # TODO: Should this be set to now? Or the original starts_at?
```
CrystalPea, line 190: When we send the broadcast for transmission, does it start broadcasting immediately or at a time we give it?

app/broadcast_message/rest.py
```diff
@@ -155,10 +155,48 @@ def update_broadcast_message_status(service_id, broadcast_message_id):
...
+        transmitted_starts_at=broadcast_message.starts_at,
+        # TODO: When cancelling, do we need to set this to now? Or should we keep it as the original time.
```
CrystalPea, line 192: It feels right to set it to now when we are cancelling, since we want to end transmission now, or am I missing something?

app/broadcast_message/rest.py
```diff
@@ -155,10 +155,48 @@ def update_broadcast_message_status(service_id, broadcast_message_id):
...
+        transmitted_finishes_at=transmitted_finishes_at,
+    )
+
+    # save to the DB
+    dao_create_broadcast_message(event)
```
CrystalPea, line 197: do we want to change this method name to ` dao_create_broadcast_event` since we are moving to events mental model?

app/models.py
```diff
@@ -2332,27 +2332,33 @@ class BroadcastEvent(db.Model):
...
 
     def serialize(self):
         return {
-            'id': self.id,
+            'id': str(self.id),
 
-            'service_id': self.service_id,
+            'service_id': str(self.service_id),
 
-            # 'reference': self.reference,
```
CrystalPea, line 2346: I'm not sure what is going on here: I get that we add `previous_event_references` to be able to get all related events into one bundle, but why are we deleting the single `reference`?

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -59,3 +60,59 @@ def test_send_broadcast_message_errors(sample_service):
             send_broadcast_message(broadcast_message_id=str(bm.id))
 
     assert ex.value.response.status_code == 503
+
+
+@freeze_time('2020-08-01 12:00')
+def test_send_broadcast_event_sends_data_correctly(sample_service):
+    t = create_template(sample_service, BROADCAST_TYPE)
+    bm = create_broadcast_message(t, areas=['london'], status=BroadcastStatusType.BROADCASTING)
+    be = create_broadcast_event(bm)
```
CrystalPea, line 69: do you think it would be ok to have more readable variable names, like words, to reduce things to keep in memory when reading the tests?

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -59,3 +60,59 @@ def test_send_broadcast_message_errors(sample_service):
...
+
+    assert request_mock.call_count == 1
+    assert request_mock.request_history[0].method == 'POST'
+    assert request_mock.request_history[0].headers["Content-type"] == "application/json"
+
+    cbc_json = request_mock.request_history[0].json()
+    assert cbc_json['id'] == str(be.id)
+    assert cbc_json['broadcast_message_id'] == str(bm.id)
+    assert cbc_json['sent_at'] == '2020-08-01T12:00:00.000000Z'
+    assert cbc_json['transmitted_starts_at'] is None
```
CrystalPea, line 83: Should we also check the type of event? I gather from lack of transmission start date that it's a cancel event?

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -59,3 +60,59 @@ def test_send_broadcast_message_errors(sample_service):
...
+    assert cbc_json['transmitted_areas'] == ['london']
+
+
+def test_send_broadcast_event_sends_references(sample_service):
```
CrystalPea, line 87: we could also have an assertion that logging works correctly potentially ☺️

app/broadcast_message/rest.py
leohemsted: dunno yet - i fully expect what timestamps we send and which ones are configurable etc to change after we get access to actual systems to test with.

app/broadcast_message/rest.py
leohemsted: agreed - didnt see point of putting it in config now. i dont know if it's expected that it's a real email address. for example the US weather warnings have the sender `w-nws.webmaster@noaa.gov` (https://alerts-v2.weather.gov/#/). Again, just wanted a placeholder to get things mergeable ahead of finding out stuff, especially around is this value viewable to end users.

app/broadcast_message/rest.py
leohemsted: dont know. i don't want to end transmission now if that means people won't receive the cancel message at all because its never broadcast, for example. Again, waiting on more info about the general lifecycle of systems to tweak this

app/broadcast_message/rest.py
leohemsted: good point - i'll rename it something generic. tbh i hate this function, it feels very unnecessary (it doesn't do anything broadcast specfic, just adds and commits) and is duplicated all over the place in loads of different dao files. but that's a wider problem for another day i think

app/models.py
leohemsted: so didnt realise it when i first wrote the function but `self.reference` is a property, that gets the reference by calling `BroadcastMessageTemplate.from_event(self.serialize()).reference`. so serialize would call `self.reference` which would call `self.serialize` which would call `self.reference`  .... meant an infinite recursion.

tests/app/celery/test_broadcast_message_tasks.py
leohemsted: this is an Alert event - starts_at doesnt need to be set in the CAP spec (sent_at always does, and if not specified, starts_at is assumed to be the same as sent_at in the spec) - this line was really making sure that we serialize null values correctly.

i'll add a check for the event type tho

app/broadcast_message/rest.py
CrystalPea: makes sense 👍

app/broadcast_message/rest.py
CrystalPea: good point!

tests/app/celery/test_broadcast_message_tasks.py
CrystalPea: ahh, ok, makes sense, thanks!

tests/app/celery/test_broadcast_message_tasks.py
leohemsted: we don't normally test logging messages, unless it a log message that we care about that changes (eg: a log message for an exception that we rely on to raise a pagerduty notification, or one that we specifically dont want to raise a pagerduty like in david's PR this morning) - is there something you specifically wanted to catch with these two log lines?

CrystalPea: no, makes sense, I try to test any log statements I put in, so there is less chance we silently stop getting those logs, but you ar probably right that it's not essential ☺️

PR #2945, opened 2020-08-05
Fix exception thrown in service_delivery_callback_task

servingUpAces:
The odd time we get a SES callback for a service with a short data retention (1 day), since there isn't a to field in the notification sent to create the status callback an exception is thrown. This PR fixes that by checking if the data type of the object passed in.

Question: what should the "notification_to" property be. I don't like returning None here as that might cause an exception with the service. Perhaps "recipient not available" but what if an invalid email address causes an issue as well.
Or do we not send the update?

The exception this PR removes is: 
```
message: Error processing SES results: <class 'AttributeError'>
traceback: Traceback (most recent call last):
  File "/home/vcap/app/app/celery/process_ses_receipts_tasks.py", line 74, in process_ses_results
    _check_and_queue_callback_task(notification)
  File "/home/vcap/app/app/notifications/notifications_ses_callback.py", line 73, in _check_and_queue_callback_task
    notification_data = create_delivery_status_callback_data(notification, service_callback_api)
  File "/home/vcap/app/app/celery/service_callback_tasks.py", line 113, in create_delivery_status_callback_data
    "notification_to": notification.to,
AttributeError: 'NotificationHistory' object has no attribute 'to'
```

### DIFF
--- a/app/celery/service_callback_tasks.py
+++ b/app/celery/service_callback_tasks.py
@@ -13,6 +13,7 @@ from app import (
     encryption
 )
 from app.config import QueueNames
+from app.models import Notification
 
 
 @notify_celery.task(bind=True, name="send-delivery-status", max_retries=5, default_retry_delay=300)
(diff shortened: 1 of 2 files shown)


idavidmcdonald:
So checking the data retention table, the only service on 1 or 2 days of data retention for emails is GOV.UK. Therefore you could ask them what they could handle?

This could also affect any services with SMS retention under 3 days but there are currently none of them (there are a handful with exact data retention of 3 days but that means 72-96 hours before we purge the notification so we shouldn't run into this case with any of them).

So I'd suggest a quick message to GOV.UK. Would love us to be able to move data retention back to 3 days for them in the future to stop us introducing weird edge cases for them but that's a whole different problem.

servingUpAces:
I've increased data retention to 3 days.

PR #2948, opened 2020-08-10
Set postage and international for letters uploaded with a CSV

servingUpAces:
If the letter is outside of the United Kingdom, set the postage and international flag.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -340,9 +340,9 @@ def save_letter(
 ):
     notification = encryption.decrypt(encrypted_notification)
 
-    recipient = PostalAddress.from_personalisation(
+    postal_address = PostalAddress.from_personalisation(
         Columns(notification['personalisation'])
-    ).normalised
+    )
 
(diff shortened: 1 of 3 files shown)


app/celery/tasks.py
```diff
@@ -354,8 +354,8 @@ def save_letter(
         saved_notification = persist_notification(
             template_id=notification['template'],
             template_version=notification['template_version'],
-            postage=template.postage,
-            recipient=recipient,
+            postage=postal_address.postage if postal_address.postage != 'united-kingdom' else template.postage,
```
quis, line 357: You can use the `international` property to avoid hard-coding `'united-kingdom'`:
```suggestion
            postage=postal_address.postage if postal_address.international else template.postage,
```

Under the hood it’s doing the same comparison – see: https://github.com/alphagov/notifications-utils/blob/b5a61bfb7b6b4c55898407304f2758a8f4be8585/notifications_utils/postal_address.py#L113-L115

servingUpAces: Updated

PR #2951, opened 2020-08-11
Add international letter service permissions for all services

servingUpAces:
Add international_letters service permission for all services that have the letter permission.

This is part 2 of 2 PRs to allow all services to have the international letter service permission. 

- [ ] part 1 https://github.com/alphagov/notifications-api/pull/2950


quis:
I think this should apply to all services, so that when they switch letters back on they get international letters by default, same as for a new service.

servingUpAces:
Query has been updated to add the permission for all services.

quis (approved):

PR #2958, opened 2020-08-17
Add endpoints to get invited users

klssmith:
We want to display flash messages in admin when invites have been
cancelled. This message needs to display the user's email address, so
this commit adds endpoints to GET a single invited service user and org
user so that we can look up the email address of a cancelled user.

[Pivotal story](https://www.pivotaltracker.com/story/show/174294749)

### DIFF
--- a/app/invite/rest.py
+++ b/app/invite/rest.py
@@ -60,6 +60,12 @@ def get_invited_users_by_service(service_id):
     return jsonify(data=invited_user_schema.dump(invited_users, many=True).data), 200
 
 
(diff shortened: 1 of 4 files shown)


klssmith:
The `admin_request` checks that the status code is `200` by default, so that was already being tested https://github.com/alphagov/notifications-api/blob/4ff0f5c4de5b459b65b74eaffd56c70acd1ccedc/tests/app/conftest.py#L874-L882 I've added in a couple of tests for the `404` scenario though.

PR #2963, opened 2020-08-28
Catch and log any exception thrown in the checkin event method.

servingUpAces:
We don't want an exception while recording metrics to affect a user action. A KeyError exception was thrown today, that meant that a user say a 500, the action being performed was to download a document from the document-download-frontend app. By catching the error we prevent the user from seeing a 500 when a recording the connection metric fails.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -352,49 +352,55 @@ def setup_sqlalchemy_events(app):
 
         @event.listens_for(db.engine, 'checkout')
         def checkout(dbapi_connection, connection_record, connection_proxy):
-            # connection given to a web worker
-            TOTAL_CHECKED_OUT_DB_CONNECTIONS.inc()
+            try:
+                # connection given to a web worker
+                TOTAL_CHECKED_OUT_DB_CONNECTIONS.inc()
 
-            # this will overwrite any previous checkout_at timestamp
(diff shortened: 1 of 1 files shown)


app/__init__.py
```diff
@@ -387,14 +387,17 @@ def checkout(dbapi_connection, connection_record, connection_proxy):
...
+                # duration that connection was held by a single web request
+                duration = time.monotonic() - connection_record.info['checkout_at']
+
+                DB_CONNECTION_OPEN_DURATION_SECONDS.labels(
+                    connection_record.info['request_data']['method'],
+                    connection_record.info['request_data']['host'],
+                    connection_record.info['request_data']['url_rule']
+                ).observe(duration)
+            except Exception:
+                current_app.logger.exception("Exception caught for checkin event.")
```
idavidmcdonald, line 403: Can we log the original exception message so we have some context if we see this?

idavidmcdonald (approved):
Seems reasonable. Could argue that we should have a similar statement on the `checkout` but we can cross that bridge if we have to come to it

app/__init__.py
servingUpAces: logger.exception will do that

idavidmcdonald: Ah OK, thanks

PR #2965, opened 2020-09-03
Persist simple polygons in the db.

CrystalPea:
They are being sent over from admin, and persisted in the db so we can send them on to the broadcast provider later on.

Also purge broadcast data so the new data type in areas column does not clash with old data type (it was an array of area codes. Now it will be a dictionary with are codes and simple polygons coordinates)

Also ensure broadcast with no areas cannot be approved (this is already checked by admin app, but we want to have a
fallback safety measure as a precaution).

Ticket: https://www.pivotaltracker.com/story/show/174385015
Admin PR: https://github.com/alphagov/notifications-admin/pull/3609 (they should go out together, API-first)


app/models.py
```diff
@@ -2263,7 +2263,7 @@ def serialize(self):
             'template_name': self.template.name,
 
             'personalisation': self.personalisation,
-            'areas': self.areas,
+            'areas': self.areas.get("areas", []),
```
idavidmcdonald, line 2266: Should we also expose `simple_polygons` too in a similar way, for example

```
'simple_poloygons': self.areas.get("simple_polygons", []),
```

migrations/versions/0328_purge_broadcast_data.py
```diff
@@ -0,0 +1,24 @@
...
+from alembic import op
+
+
+revision = '0328_purge_broadcast_data'
+down_revision = '0327_idx_notification_history'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute("TRUNCATE broadcast_event, broadcast_message;")
```
idavidmcdonald, line 17: This will get run before the rest of the code is deployed, which is fine but very small risk that we truncate the table, code begins deploying out, a user creates a broadcast with the old schema, code finishes getting deployed out. This would leave us with the old schema of the data in the DB which would need a manual tidy up. Very happy to tolerate that risk however, just thought I'd point it out just in case you weren't aware of it.

tests/app/broadcast_message/test_rest.py
```diff
@@ -151,7 +158,7 @@ def test_update_broadcast_message_allows_edit_while_not_yet_live(admin_request,
     )
 
     assert response['starts_at'] == '2020-06-01T20:00:01.000000Z'
-    assert response['areas'] == ['london', 'glasgow']
+    assert response['areas'] == ['london']
```
idavidmcdonald, line 161: We appear to be updating the broadcast and setting the areas to london and glasgow but the response shows just london. How is this possible? Does this mean the areas aren't being updated for some reason? Should we be updating `_data={'starts_at': '2020-06-01 20:00:01', 'areas': ['london', 'glasgow']},` to also include `simple_polygons`?

tests/app/broadcast_message/test_rest.py
```diff
@@ -354,6 +368,31 @@ def test_update_broadcast_message_status_rejects_approval_from_creator(
...
+        _data={'status': BroadcastStatusType.BROADCASTING, 'created_by': str(approver.id)},
+        service_id=template.service_id,
+        broadcast_message_id=broadcast.id,
+        _expected_status=400
+    )
+
+    assert mock_task.called is False
+    assert response[
+        'message'
+    ] == f'broadcast_message {broadcast.id} has no selected areas and so cannot be broadcasted.'
```
idavidmcdonald, line 393: Good test

idavidmcdonald (commented):
Generally looks good. Few minor comments

app/models.py
CrystalPea: good point, @servingUpAces what do you think?

migrations/versions/0328_purge_broadcast_data.py
CrystalPea: Thanks, I do realise there are a few things in those deploys that would be risky if the feature was live. As it is, the feature is not live yet and I plan to ask our team to not use the broadcast feature while I deploy.

tests/app/broadcast_message/test_rest.py
CrystalPea: thanks 💖

tests/app/broadcast_message/test_rest.py
CrystalPea: good catch!

app/models.py
CrystalPea: I added it, sounds like a sensible suggestion.

tests/app/broadcast_message/test_rest.py
CrystalPea: I fixed that test now, and I asserted that simple_polygons have updated correctly :).

tests/app/broadcast_message/test_rest.py
```diff
@@ -145,20 +145,24 @@ def test_update_broadcast_message_allows_edit_while_not_yet_live(admin_request,
...
     )
 
     response = admin_request.post(
         'broadcast_message.update_broadcast_message',
-        _data={'starts_at': '2020-06-01 20:00:01', 'areas': ['london', 'glasgow']},
+        _data={
+            'starts_at': '2020-06-01 20:00:01',
+            'areas': ['london', 'glasgow'],
+            "simple_polygons": [[50.12, 1.2], [50.13, 1.2], [50.14, 1.21]]
+        },
```
idavidmcdonald: So good that you fixed this test but what I don't get was how was it passing before. Was it the lack of simple_polygons in the post request that was causing the update of the areas not to happen? If so, should that not have 400'd rather than 200'd?

CrystalPea: Hmm, I'm not sure, it's not a problem we would have in real life, more like bad test data. I could add a guard statement against that if you feel it would be better to, it would look something like:
```
if ('areas' in data and 'simple_polygons' not in data) or ('areas' not in data and 'simple_polygons' in data):
        raise InvalidRequest(
            f'Cannot update broadcast_message {broadcast_message.id}, areas or polygons are missing.',
            status_code=400
        )
```

Not sure what the benefit would be though.

idavidmcdonald: So I'd argue that it is worth testing for this and catching the error. The API is treated in theory as separate from the admin app so even though the admin app won't do this, we shouldn't trust that all API calls will follow this pattern. Maybe in the future the API endpoint will be accessible to the public. Maybe in the future a developer won't realise that you need to send both of them and just sends one of them. I think for the cost of the extra few lines of code if we always want both areas and polygons or neither, then it's worth us checking for this in the API.

I was trying to think of a smart way to do the if statement and the best I could come up with is
```if len(set(data).intersection(("areas", "simple_polygons"))) == 1```
but not sure how readable that is. Your suggestion is definitely more readable and so probably worth going for but it was a fun little brain teaser for me

Comment (CrystalPea) [reply tests/app/broadcast_message/test_rest.py]: ok, I added that check ☺️ . Cool brain teaser with making the guard statement terse, code wars style :). As you say, I will keep the long-ish statement for readability ☺️

File: app/broadcast_message/rest.py [hunk 6]
```diff
@@ -101,7 +106,7 @@ def create_broadcast_message(service_id):
         template_id=template.id,
         template_version=template.version,
         personalisation=data.get('personalisation', {}),
-        areas=data.get('areas', []),
+        areas=data.get('areas', {}),
```
idavidmcdonald, line 109: Does this mean the interface for creating a broadcast is different to updating one.

So to create one I would post

```
{"areas": {"areas": [things here], "simple_polygons": [things here]}}
```

but to update it I would post
```
{"areas": [things here], "simple_polygons": [things here]}
```

If that is so then I don't think I like having that difference. Would you agree?

Simarly, if we were to create a broadcast, we could create it with just `areas` at the moment but nothing would stop us not giving `simple_polygons`. 

Essentially, I want the type of data and the validation rules to be consistent for creating and updating please.

Feel free to let me know if I've got this wrong or you don't think this is worth it. Sorry I didn't spot this earlier and there has been some back and forth from me stopping this going out so far.

CrystalPea: Ahh, I know why the disparity: when creating broadcast message, we are not specifying areas at all, we are not sending them over from admin.

But you are right I should update this statement in rest to be consistent and future-proof 👍

idavidmcdonald (approved):
I'm going to approve this as is as I think it's good enough for what we need. The interface is now the same and we have a good check on the update route to make sure we always send both areas and simple_polygons.

With hindsight vision, I think what I would suggest as further potential improvements would be that
- we also don't allow you to create a broadcast via a POST with only areas or only simple_polygons set, you must have neither or both
- given that we want protection for both create and updates and generally any way we create or edit a broadcastMessage in the DB, I would suggest something on the model, a property setter for areas, that would raise an exception if we tried to only set one of them. This could act as the real gateway instead of us having to do it in multiple places around the codebase

But like I said, I think how the code is written at the moment is good and safe enough and we should save any other improvements for a later time if at all so we can concentrate on delivering this quickly.

Thanks for your patience with me only spotting things and coming up with ideas at every time I've taken a look, rather than all of this at the beginning :)

CrystalPea:
@idavidmcdonald 

Thanks! I think it is standard thinking process to spot things as we go, and see more stuff as we look the n-th time.

I agree those measures you listed sound sensible, and I also agree we can do without them for now. Let's see if the shape of data mutates on in the coming weeks :)

PR #2968, opened 2020-09-09
Allow international letters to be cancelled

quis:
Our code was assuming that any notifications with `international` set to `True` were text messages. It was then trying to look up delivery information for a notification which wasn’t sent to a phone number, causing an exception.

***

Need to cancel this letter after deploying: https://www.notifications.service.gov.uk/services/d6aa2c68-a2d9-4437-ab19-3ae8eb202553/notification/6420f6d8-0fc8-4574-8e30-7beda9abb86b

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -140,7 +140,11 @@ def update_notification_status_by_id(notification_id, status, sent_by=None, deta
         _duplicate_update_warning(notification, status)
         return None
 
-    if notification.international and not country_records_delivery(notification.phone_prefix):
+    if (
+        notification.notification_type == 'sms'
+        and notification.international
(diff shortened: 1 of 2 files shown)


app/dao/notifications_dao.py
```diff
@@ -140,7 +140,11 @@ def update_notification_status_by_id(notification_id, status, sent_by=None, deta
         _duplicate_update_warning(notification, status)
         return None
 
-    if notification.international and not country_records_delivery(notification.phone_prefix):
+    if (
+        notification.notification_type == 'sms'
```
klssmith, line 144:
```suggestion
        notification.notification_type == SMS_TYPE
```

klssmith (approved):
Tiny optional suggestion, but looks good

PR #2972, opened 2020-09-15
Broadcast invite template

idavidmcdonald:
https://www.pivotaltracker.com/story/show/174527066

image

image


karlchillmaid (approved):
@quis how's this for a first pass? I kept it pretty close to the regular Notify invite email.

Is it worth putting service names in 'quotes'? I know we sometimes do that on Notify for clarity.

quis (approved):
@karlchillmaid Looks good to me. I’d stay consistent with the regular invite email on use of quotes for now. Although maybe we should revisit that, if we’re advising teams use names that don’t flow naturally in a sentence any more, for example :
- …has invited you to collaborate on Fishing Licences on GOV.UK Notify (not bad)
- …has invited you to collaborate on Bulky Waste – Borchester Council on GOV.UK Notify (hard to parse)

PR #2974, opened 2020-09-21
Make letter folder name code readable

leohemsted:
Had a little bit of a headache trying to read the get_folder_name function the other day.

Key things here are:

* Moving decision around whether a folder exists outside of `get_folder_name`
* renaming `sending_date` to `created_at`, because that's what it's always set to, even if the letter was sent many days later due to technical issues with us or with dvla


klssmith (approved):
Nice, I think this makes this much easier to understand

PR #2977, opened 2020-09-23
add broadcast org

leohemsted:
keep the ID the same as the existing prod broadcast org to keep things simple.

Move all services with a broadcast permission across to this new service. Note that we'll need to clear the service and organisation cache after deploying this.

- [ ] https://github.com/alphagov/notifications-admin/pull/3643


quis:
Shouldn’t this go after the admin PR? Otherwise you’d have a gap where services not caught by this migration would still get assigned to Cabinet Office.

quis (approved):

PR #2982, opened 2020-09-26
Improve efficiency of process missing rows task

quis:
For every missing row this was:
- downloading the CSV file from S3
- looping through every row in it until it found the one matching the index of the missing row 

`RecipientCSV` [implements `__getitem__`](https://github.com/alphagov/notifications-utils/blob/5ae0572d4188483fa140bccc128ecf80f3473000/notifications_utils/recipients.py#L78-L79) (which maybe it didn’t before) so we can create it once, then index the relevant row directly.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -237,14 +237,13 @@ def check_for_missing_rows_in_completed_jobs():
     jobs_and_job_size = find_jobs_with_missing_rows()
     for x in jobs_and_job_size:
         job = x[1]
+        recipient_csv, template, sender_id = get_recipient_csv_and_template_and_sender_id(job)
         missing_rows = find_missing_row_for_job(job.id, job.notification_count)
         for row_to_process in missing_rows:
-            recipient_csv, template, sender_id = get_recipient_csv_and_template_and_sender_id(job)
-            for row in recipient_csv.get_rows():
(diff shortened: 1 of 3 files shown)


idavidmcdonald (approved):
So this looks good and is definitely more efficient.

Two small things to check
- how come I can't find any logs when searching for "Processing missing row" in Kibana in production for the past 7 days. I can see some logs in preview but not in prod. Did we not run into this error in prod at all? Or was this avoided because we upped the time limit to 20 minutes?

- So this will make things very efficient compared to what it was like before. Could there be any knock on effect if this task processes really quickly in terms of clashes with the original job processing (assuming it's still going after 20 minutes). I don't think there will be any impact because this process is now much faster but just thought I'd see if you had come across anything

quis:
I’m not too worried about it bottlenecking somewhere else because it calls into the same [`process_row` function](https://github.com/alphagov/notifications-api/blob/1d50bfaafc97682a11992dbcef8510914c231244/app/celery/tasks.py#L134) that is used to process the job the first time around. It shouldn’t overlap with the original job processing because we set that 20 minute grace period.

Not sure why we didn’t see those log messages on Sunday because we did see a lot of the unique constraint errors happening when one of the jobs got interrupted by a deploy:
image (16)

idavidmcdonald:
👍

PR #2983, opened 2020-09-27
Make run-scheduled jobs to run every minute

servingUpAces:


quis:
The syntax is correct:
image

quis (approved):
Discussed in person, happy that this won’t result in jobs being picked up twice because:
- it’s using it’s own queue
- `process_job` sets the status to `JOB_STATUS_IN_PROGRESS` before doing any of the work

PR #2984, opened 2020-09-28
Add an endpoint for stats about scheduled jobs

quis:
At the moment we display the count of scheduled jobs on the dashboard by sending all the scheduled jobs to the admin app and letting it work out the stats.

This is inefficient and, because the get jobs response has a page size of 50, becomes incorrect if a service schedules more than 50 jobs.

This commit adds a separate endpoint which gives the admin app the stats it needs directly and correctly.

### DIFF
--- a/tests/app/job/test_rest.py
+++ b/tests/app/job/test_rest.py
@@ -910,3 +910,36 @@ def test_get_jobs_should_retrieve_from_ft_notification_status_for_old_jobs(admin
     assert resp_json['data'][1]['statistics'] == [{'status': 'created', 'count': 1}]
     assert resp_json['data'][2]['id'] == str(job_1.id)
     assert resp_json['data'][2]['statistics'] == [{'status': 'delivered', 'count': 6}]
+
+
+@freeze_time('2017-07-17 07:17')
+def test_get_scheduled_job_stats_when_no_scheduled_jobs(admin_request, sample_template):
+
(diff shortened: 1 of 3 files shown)


tests/app/job/test_rest.py
```diff
@@ -910,3 +910,36 @@ def test_get_jobs_should_retrieve_from_ft_notification_status_for_old_jobs(admin
...
+    resp_json = admin_request.get('job.get_scheduled_job_stats', service_id=service_id)
+    assert resp_json == {
+        'count': 0,
+        'soonest_scheduled_for': None,
+    }
+
+
+@freeze_time('2017-07-17 07:17')
+def test_get_scheduled_job_stats(admin_request, sample_template):
+    create_job(sample_template, job_status='scheduled', scheduled_for='2017-07-17 09:00')
```
idavidmcdonald, line 932: I would suggest also adding a job not in the scheduled status which will add test coverage to make sure we don't include scheduled jobs which have say completed or been cancelled etc. Also a job not for this service to make sure we don't include those either.

idavidmcdonald (commented):
Looks good but could do with some tighter test coverage please.

tests/app/job/test_rest.py
quis: Good call, added in 4cee47da

idavidmcdonald (approved):

PR #2986, opened 2020-09-28
Use cache to improve performance of CSV processing

quis:
When sending jobs from concurrent CSV files we seem to be limited by how fast we can save items to the database. We’ve made saving to the database faster in other places by taking advantage of the service and template cache to open the connection later. 

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/3074
- [x] clear Redis template cache in each environment

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -42,7 +42,7 @@ from app.dao.returned_letters_dao import insert_or_update_returned_letters
 from app.dao.service_email_reply_to_dao import dao_get_reply_to_by_id
 from app.dao.service_inbound_api_dao import get_service_inbound_api_for_service
 from app.dao.service_sms_sender_dao import dao_get_service_sms_senders_by_id
-from app.dao.services_dao import dao_fetch_service_by_id, fetch_todays_total_message_count
+from app.dao.services_dao import fetch_todays_total_message_count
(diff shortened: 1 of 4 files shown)


quis:
I‘ve done some performance testing of this locally, processing a 25 row CSV file. Here is how long the `save_email` task took in seconds.

  | Without cache | With cache
-- | -- | --
Min | 0.10 | 0.10
Max | 2.50 | 2.56
Median | 0.41 | 0.40
Mean | 0.58 | 0.60

In other words, there is no meaningful speed increase. Worth bearing in mind that:
- my local database doesn’t have many templates or services
- my local database isn’t under any load

Question is whether this is still worth it for the reduction in load on the database in production?

quis:
There is a noticeable improvement when timing only the part of the code which gets the service and template:

image

Some interesting observations: 
- looks like Postgres might be doing some cacheing because the first couple of hits to the database are really slow
- you can see the choppiness as we alternate from the memory cache back to the Redis cache, which is about 20&times; slower (but still 10&times; faster than Postgres)

Going from 10ms to 1ms to 50μs isn’t much in the context of the whole task (which takes on average 590ms), but it is a significant improvement…

CrystalPea (approved):
Good work 👍🏼

PR #2990, opened 2020-10-02
Serialise content in broadcast message response

quis:
This will let us show the content of the broadcast message in places we can’t at the moment.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2243,6 +2243,12 @@ class BroadcastMessage(db.Model):
     cancelled_by = db.relationship('User', foreign_keys=[cancelled_by_id])
 
     @property
+    def content(self):
+        return self.template._as_utils_template_with_personalisation(
+            self.personalisation
+        ).content_with_placeholders_filled_in
+
+    @property
     def personalisation(self):
         if self._personalisation:
             return encryption.decrypt(self._personalisation)
@@ -2261,6 +2267,7 @@ class BroadcastMessage(db.Model):
             'template_id': str(self.template_id),
             'template_version': self.template_version,
             'template_name': self.template.name,
+            'template_content': self.content,
 
             'personalisation': self.personalisation,
             'areas': self.areas.get("areas", []),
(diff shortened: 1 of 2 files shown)


app/models.py
```diff
@@ -2261,6 +2267,7 @@ def serialize(self):
             'template_id': str(self.template_id),
             'template_version': self.template_version,
             'template_name': self.template.name,
+            'template_content': self.content,
```
idavidmcdonald, line 2270: Does this make sense to be `template_content` rather than `content`? `template_content` to me is a template that hasn't had its placeholders filled in whereas content is where it would have (which is what has happened above)

tests/app/broadcast_message/test_rest.py
```diff
@@ -24,6 +24,8 @@ def test_get_broadcast_message(admin_request, sample_broadcast_service):
     )
 
     assert response['id'] == str(bm.id)
+    assert response['template_id'] == str(t.id)
+    assert response['template_content'] == t.content
```
idavidmcdonald, line 28: Should we be checking the fact that the content is rendered with placeholders filled in? I'm guessing the template in this case doesn't have any placeholders so the test passes regardless though? And I see another PR which is going to stop us from allowing placeholders in it so maybe we should instead change `def content` such that it doesn't try and fill in placeholders?

Or should we be

idavidmcdonald (commented):
I feel like this doesn't quite jam with https://github.com/alphagov/notifications-admin/pull/3672/files in terms of do we or do we not support templates with placeholders for broadcast messages. Thoughts?

tests/app/broadcast_message/test_rest.py
quis: I’ve added a test in 80352baf

idavidmcdonald: Cool, this is good. Thanks

app/models.py
idavidmcdonald: I think this comment still stands?

quis: Changed in 57473ef6

idavidmcdonald (approved):

PR #2992, opened 2020-10-06
Tailor message-too-long error message depending on the notification type

CrystalPea:
Up until now, only sms messages could get message-too-long error,
but now we also need to validate the size of email messages, so
the message content needs to be tailored to the notification type.

This utils PR needs to go in first: https://github.com/alphagov/notifications-utils/pull/798/files
Pivotal ticket: https://www.pivotaltracker.com/story/show/175115179

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -154,7 +154,7 @@ def check_message_is_not_too_long(template_with_content):
             )
         elif template_with_content.template_type == EMAIL_TYPE:
             message = (
-                f"Emails cannot be longer than 7500000 bytes. "
+                f"Emails cannot be longer than 1000000 bytes. "
                 f"Your message is {template_with_content.content_size_in_bytes} bytes."
             )
         raise BadRequestError(message=message)
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -144,10 +144,19 @@ def check_if_service_can_send_to_number(service, number):
         return international_phone_info
 
 
-def check_content_char_count(template_with_content):
+def check_message_is_not_too_long(template_with_content):
     if template_with_content.is_message_too_long():
-        message = f"Text messages cannot be longer than {SMS_CHAR_COUNT_LIMIT} characters. " \
-                  f"Your message is {template_with_content.content_count_without_prefix} characters"
+        message = "Your message is too long."
+        if template_with_content.template_type == SMS_TYPE:
+            message = (
+                f"Text messages cannot be longer than {SMS_CHAR_COUNT_LIMIT} characters. "
+                f"Your message is {template_with_content.content_count_without_prefix} characters long."
+            )
+        elif template_with_content.template_type == EMAIL_TYPE:
+            message = (
+                f"Emails cannot be longer than 7500000 bytes. "
+                f"Your message is {template_with_content.content_size_in_bytes} bytes."
+            )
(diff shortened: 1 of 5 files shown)


app/notifications/validators.py
```diff
@@ -144,10 +144,19 @@ def check_if_service_can_send_to_number(service, number):
         return international_phone_info
 
 
-def check_content_char_count(template_with_content):
+def check_content_character_count(template_with_content):
```
idavidmcdonald, line 147: Optional: The function name is better however this function checks character counts for sms and checks bytes for email which means the function name is a little misleading.

I don't think we can make the link back from bytes to characters though unfortunately (which would give a nicer message for the user). Maybe `check_message_is_not_too_long` or `check_message_length` would be better?

It might be worth tagging Karl for a quick review here too please.

app/notifications/validators.py
```diff
@@ -144,10 +144,19 @@ def check_if_service_can_send_to_number(service, number):
...
     if template_with_content.is_message_too_long():
-        message = f"Text messages cannot be longer than {SMS_CHAR_COUNT_LIMIT} characters. " \
-                  f"Your message is {template_with_content.content_count_without_prefix} characters"
+        message = "Your message is too long."
+        if template_with_content.template_type == SMS_TYPE:
+            message = (
+                f"Text messages cannot be longer than {SMS_CHAR_COUNT_LIMIT} characters. "
+                f"Your message is {template_with_content.content_count_without_prefix} characters."
```
idavidmcdonald, line 153: You've added a full stop on the end here. Is that intentional? I don't know if that would mean it no longer matches our API documentation? Or if we should have one to start with?

idavidmcdonald (approved):
Looks good technically. May just be worth flagging the content with Karl though

app/notifications/validators.py
CrystalPea: I like that idea, I did send a Slack message to @karlchillmaid about this PR when I first put it up ☺️

app/notifications/validators.py
CrystalPea: One for @karlchillmaid I guess, I changed it because it felt weird that we had no gull stop and then we had a capital letter 🤷

app/notifications/validators.py
idavidmcdonald: Ah, great minds think alike then Pea 👍

app/notifications/validators.py
karlchillmaid: I think the content looks good. Skimming through the other error messages in this file, it looks like we might be a bit inconsistent about full stops in API messages. Something to fix in the future?

karlchillmaid, line 153:
```suggestion
                f"Your message is {template_with_content.content_count_without_prefix} characters long."
```

app/notifications/validators.py
```diff
@@ -144,10 +144,19 @@ def check_if_service_can_send_to_number(service, number):
...
+            )
+        elif template_with_content.template_type == EMAIL_TYPE:
+            message = (
+                f"Email messages cannot be longer than 7500000 bytes. "
```
karlchillmaid, line 157:
```suggestion
                f"Emails cannot be longer than 7500000 bytes. "
```

Is 'longer than' the right way to describe something measured in bytes?

PR #2994, opened 2020-10-09
Add topic about tests with docker

caduvieira:
Just adding comment about tests with docker.
If accepted can a maintainer add a hacktoberfest-accepted label for the hacktoberfest? Thanks


idavidmcdonald:
Hi @caduvieira,

I don't think this is a beneficial enough change for our README unfortunately. The notify developers, who are the main users of this README, tend to not use docker for testing these changes locally. If they really wanted to they can find all the optional things they can do by running `make`. In this case, I don't think the benefit of having this new section outweight the extra maintenance cost of having the extra section.

Apologies but I hope you feel that is reasonable.

PR #2998, opened 2020-10-15
Adding service_id to the sort order for the letters being sent to print.

servingUpAces:
We have had a few instances where letters have caused problems. Particularly for precompiled letters, often the issue comes from the same service.
The hope is that by adding a sort order this will help the print provider narrow down the problem.

There is a small degradation of the performance of the query, but it's not enough to concern me.


CrystalPea (approved):
I love it 👍 
:shipit:  :shipit:  :shipit:

PR #3000, opened 2020-10-19
Add a task to save-api-sms for high volume services.

servingUpAces:
When we initially added a new task to persist the notifications for a high volume service we wanted to implement it as quickly as possible, so ignored SMS.
This will allow a high volume service to send SMS, the SMS will be sent to a queue to then persist and send the SMS, similar to emails.

At this point I haven't added a new application to consume the new save-api-sms-tasks. But we can add a separate application or be happy with how the app scales for both email and sms.

How to test:
Add a local service to the HIGH_VOLUME_SERVICE in app.config.Development or your source environment.sh, it's an array. Then post a sms and email from the python client using that service, need to use a live or team key. You should see the save-api-email-task or save-api-sms-task being called from the celery app.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -69,7 +69,6 @@ from app.v2.notifications.notification_schemas import (
 )
 from app.v2.utils import get_valid_json
 
-
 POST_NOTIFICATION_JSON_PARSE_DURATION_SECONDS = Histogram(
     'post_notification_json_parse_duration_seconds',
     'Time taken to parse and validate post request json',
@@ -201,11 +200,14 @@ def process_sms_or_email_notification(
         template_with_content=template_with_content
     )
 
-    if service.id in current_app.config.get('HIGH_VOLUME_SERVICE') and api_user.key_type == KEY_TYPE_NORMAL:
-        # Put GOV.UK Email notifications onto a queue
+    if service.id in current_app.config.get('HIGH_VOLUME_SERVICE') \
+        and api_user.key_type == KEY_TYPE_NORMAL \
+            and notification_type in [EMAIL_TYPE, SMS_TYPE]:
+        # Put service with high volumes of notifications onto a queue
         # To take the pressure off the db for API requests put the notification for our high volume service onto a queue
         # the task will then save the notification, then call send_notification_to_queue.
-        # We know that this team does not use the GET request, but relies on callbacks to get the status updates.
+        # NOTE: The high volume service should be aware that the notification is not immediately
+        # available by a GET request, it is recommend they use callbacks to keep track of status updates.
         try:
             save_email_or_sms_to_queue(
                 form=form,
@@ -291,7 +293,7 @@ def save_email_or_sms_to_queue(
 
     if notification_type == EMAIL_TYPE:
         save_api_email.apply_async([encrypted], queue=QueueNames.SAVE_API_EMAIL)
-    else:
+    elif notification_type == SMS_TYPE:
         save_api_sms.apply_async([encrypted], queue=QueueNames.SAVE_API_SMS)
 
     return Notification(**data)
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -18,7 +18,7 @@ from app import (
 )
 from app.celery.letters_pdf_tasks import get_pdf_for_templated_letter, sanitise_letter
 from app.celery.research_mode_tasks import create_fake_letter_response_file
-from app.celery.tasks import save_api_email
+from app.celery.tasks import save_api_email, save_api_sms
 from app.clients.document_download import DocumentDownloadError
 from app.config import QueueNames, TaskNames
 from app.dao.templates_dao import get_precompiled_letter_template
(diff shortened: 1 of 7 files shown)


leohemsted:
Given the caching that we added after adding the email function, have we taken a look t osee what is the actual response time difference between a post sms and a post email currently? 

as in: does the high volume service queue this result in real world improvements in API performance?

idavidmcdonald:
In addition to Leos comment, it would be nice for us to be clear on what this PR is solving. It was a while since we did the high volume email service and I just want to check, mostly for my own knowledge.

I can see two potential advantages to having a high volume service

## 1. Quicker response times for a POST request that uses the high volume service than doesn't
If response times are quicker then it means we can process a higher rate of requests to the API (in terms of the HTTP processing side of things, not in relation to the DB side of things). This is because our gunicorn/eventlet workers will be under lesser load than before and we will hopefully not reach the point of having all of our workers busy processing requests when new ones come in. Quicker response times are also better for users.

## 2. Protects the DB from using up too many connections at very high load
Without a high volume queue, were we to see a very large load of traffic to send emails/sms, assuming that we could handle it at the HTTP level, if that spike used up all of our DB connections then this would be very bad.
With a high volume queue, were to see a very large spike in traffic to send emails/sms, assuming that we could handle it at the HTTP level, we would only use up as many connections as we let the save-api-sms-workers were limited to based on the number of instances. This means that we would build up a queue of things to save but would stop us using up all our connections which is much preferable to the above scenario. 


## Summary
Leo seemed to ask if statement 1 is true, would response times decrease (at least substantially). It would be nice to know the answer but I think if we have the save-api-email-worker and save-api-sms-worker scaled appropriately, given the second statement, then I think this is still a beneficial change regardless.

app/v2/notifications/post_notifications.py
```diff
@@ -201,14 +201,13 @@ def process_sms_or_email_notification(
         template_with_content=template_with_content
     )
 
-    if service.id in current_app.config.get('HIGH_VOLUME_SERVICE') and api_user.key_type == KEY_TYPE_NORMAL \
-       and notification_type == EMAIL_TYPE:
+    if service.id in current_app.config.get('HIGH_VOLUME_SERVICE') and api_user.key_type == KEY_TYPE_NORMAL:
         # Put GOV.UK Email notifications onto a queue
```
idavidmcdonald, line 205: Comment could be updated as this is no longer just GOV.UK Email and the final sentence about `we know that time team does not use GET requests` could be made more generic to just describe the behaviour change, for example

High volume services may not be able to immediately GET their notification, so should either be aware of this or not need to GET request their notifications, for example by relying on callbacks instead to get status updates.

app/v2/notifications/post_notifications.py
```diff
@@ -201,14 +201,13 @@ def process_sms_or_email_notification(
         template_with_content=template_with_content
     )
 
-    if service.id in current_app.config.get('HIGH_VOLUME_SERVICE') and api_user.key_type == KEY_TYPE_NORMAL \
-       and notification_type == EMAIL_TYPE:
```
idavidmcdonald, line 205: We should make this strict so check that the notification type is in email or sms, just in case we get a high volume service who then decide to send a few letters

app/v2/notifications/post_notifications.py
```diff
@@ -290,7 +289,11 @@ def save_email_to_queue(
         data
     )
 
-    save_api_email.apply_async([encrypted], queue=QueueNames.SAVE_API_EMAIL)
+    if notification_type == EMAIL_TYPE:
+        save_api_email.apply_async([encrypted], queue=QueueNames.SAVE_API_EMAIL)
+    else:
```
idavidmcdonald, line 294: Optional: Could check explicitly that notification type is SMS and then fall back to raising an exception if it's not for some reason. Just in case we were ever to call this with a letter accidently

scripts/paas_app_wrapper.sh
```diff
@@ -51,7 +51,7 @@ case $NOTIFY_APP_NAME in
     ;;
   delivery-worker-save-api-notifications)
     exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=11 \
-    -Q save-api-email-tasks 2> /dev/null
+    -Q save-api-email-tasks, save-api-sms-tasks 2> /dev/null
```
idavidmcdonald, line 54: I think I like the fact they go on the same queue. These services aren't ones that are allowed to be worried about slight delays to their messages going out, so it doesn't matter if one service sending out lots of SMS caused some delays to GOV.UK email.

Also it means we only have to scale one worker to control how much DB resource we use by high volume services

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1122,36 +1121,39 @@ def test_post_notifications_doesnt_save_email_to_queue_for_test_emails(client, n
...
-    assert mock_send_task.called
-    assert not save_email_task.called
-    assert len(Notification.query.all()) == 1
+    # save email
+    save_email_task.assert_called_once_with([mock.ANY], queue='save-api-email-tasks')
+    mock_send_task.assert_called_once_with([json_resp['id']], queue='send-email-tasks')
+    assert Notification.query.count() == 1
 
 
-def test_post_notifications_doesnt_save_email_to_queue_for_sms(client, notify_db_session, mocker):
```
idavidmcdonald, line 1130: Optional: If I understand what this original test is doing, it might be worth adding a test to make sure that a letter sent by a high volume service processes fine and doesn't use the save-api tasks

idavidmcdonald (commented):
Looking pretty good. A couple of comments I think might need addressing and then a few small optional ones.

app/v2/notifications/post_notifications.py
servingUpAces: This code path is only for SMS and EMAIL

idavidmcdonald: Ah yes, thanks. All good as is then.

idavidmcdonald (approved):
Nice work

PR #3004, opened 2020-10-20
Divide letters by service when putting in ZIP files

CrystalPea:
When letters are sent to DVLA, we will now put them in a separate ZIP file for each service, so that if there are printing issues due to bad files from one service, other services will hopefully not be affected by that.

EDIT: reintroducing letters from Insolvency Service back to DVLA moved to this PR: https://github.com/alphagov/notifications-api/pull/3006

### DIFF
--- a/tests/app/celery/test_letters_pdf_tasks.py
+++ b/tests/app/celery/test_letters_pdf_tasks.py
@@ -42,7 +42,7 @@ from app.models import (
     NOTIFICATION_VIRUS_SCAN_FAILED,
 )
 
-from tests.app.db import create_notification, create_letter_branding, create_service
+from tests.app.db import create_notification, create_letter_branding, create_service, create_template
 
 from tests.conftest import set_config_values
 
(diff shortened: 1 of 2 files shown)


tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -232,9 +238,83 @@ def test_get_key_and_size_of_letters_to_be_sent_to_print(notify_api, mocker, sam
...
+    ])
+
+    results = get_key_and_size_of_letters_to_be_sent_to_print(datetime.now() - timedelta(minutes=30), postage='second')
+
+    assert mock_s3.call_count == 2
+    mock_s3.assert_has_calls(
+        [
+            call(
+                current_app.config['LETTERS_PDF_BUCKET_NAME'],
+                '2020-02-17/NOTIFY.INSOLVENCY.D.2.C.C.20200217150000.PDF'
```
idavidmcdonald, line 299: Just a quick cursory glance before stand up, sorry I don't have the time to do a proper review however one thing I think we need to do is keep the filenames of the letters the same and not change those. I don't think we be sure DVLA can support a change in filename and we want to keep this change as small as possible to minimise any portential unexpected side affects. It's definitely OK to change the name on the zip file but I think we need to keep the files inside the same please

leohemsted: @idavidmcdonald this is just because pea set the dvla ref to `insolvency` in the test setup on line 283

CrystalPea: Yeah, the pdf filename did not change, it's just I set ref to `INSOLVENCY`, maybe I could change it to `1N50LV3NCY` so it's less confusing? :D

idavidmcdonald: I think don't worry about it for the moment, but thanks. Sorry!

leohemsted (approved):
this looks good, and tests look good too - i reckon we should add an insolvency org on preview with the correct ID and test some stuff there too though

idavidmcdonald:
Heads up, I've emailed Carl to check if we are good to start sending insolvency letters again. Awaiting confirmation on this

CrystalPea:
@leohemsted that sounds good, so I guess we would be triggering the task on preview manually to test it before 17:30? (of course if we get go-ahead from Carl)

@idavidmcdonald sounds good, I have added some details about what we are planning to that convo, in case it makes a difference.

leohemsted (approved):

PR #3006, opened 2020-10-21
Start sending letters from Insolvency Service to DVLA again

CrystalPea:
Reintroduce the sending of Insolvency letters to DVLA and mark ZIP files with insolvency letters so DVLA knows to process those separately.

Review commit by commit.

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -181,6 +181,10 @@ def get_key_and_size_of_letters_to_be_sent_to_print(print_run_deadline, postage)
     letter_pdfs = []
     for letter in letters_awaiting_sending:
         try:
+            if str(letter.service.organisation_id) == 'f33fdfdd-7533-40cb-b5e8-cd78a1f5d21e':
+                letter_service_marking = str(letter.service.id) + ".INSOLVENCY"
+            else:
+                letter_service_marking = str(letter.service.id)
             letter_file_name = get_letter_pdf_filename(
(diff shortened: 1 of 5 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -181,6 +181,10 @@ def get_key_and_size_of_letters_to_be_sent_to_print(print_run_deadline, postage)
     letter_pdfs = []
     for letter in letters_awaiting_sending:
         try:
+            if str(letter.service.organisation_id) == 'f33fdfdd-7533-40cb-b5e8-cd78a1f5d21e':
```
leohemsted, line 184: hmmm, will this cause a lot of queries to the DB to get the service for every letter. not sure if sqlalchemy will cache the results?

leohemsted (approved):
code looks fine but i've got a question about performance - something to keep an eye on.

app/celery/letters_pdf_tasks.py
CrystalPea: Hmmm, good question, I was not very worried about it, because I saw we already ask for `letter.service.crown` a few lines below, and I assumed if it can do that, it must already pull the service somehow, but really I'm not that sure how it works 🤔

leohemsted: ah, good spot. if we already do crown, then no worries - this won't cause any more load. it's the `letter.service` bit that will fetch from the DB

PR #3012, opened 2020-10-23
Remove references to stub CBC

tlwr:
⚠️ Needs a rebase once https://github.com/alphagov/notifications-api/pull/3011 is merged

What
----

Removes references to the CBC stub

Wires up the update and cancel methods of the `cbc_proxy` client

Why
----

We are testing with live CBCs, we don't need to use a stub and the two codepaths will diverge and make things unnecessarily complicated and non-obvious


tlwr:
Rebased

idavidmcdonald (approved):

PR #3014, opened 2020-10-26
Optimise dao_get_letters_to_be_printed query

leohemsted:
Two main concepts here - for info on each one, check the commit messages for more info

#### return just the columns we need for collating letters

Rather than using `Notification.query` to return an ORM Notification object, we can just return the specific columns we need. We can join on service to fetch the crown status, there's much less data sent over the wire, less memory used on the app, and it's also useful for...

#### use yield_per instead of limit

Sqlalchemy supports not pre-fetching all results, but instead yielding every X (I set this to 10k, ignore the branch name). This is risky to use with more ORM features like referring to relationships (which is part of the motive for the first commit), but means that we don't fetch all data and build up lots of ORM objects in memory - only 10k each time, then we'll go get the files from s3 and start to send them before going back to the database for more.

### DIFF
--- a/tests/app/dao/notification_dao/test_notification_dao.py
+++ b/tests/app/dao/notification_dao/test_notification_dao.py
@@ -1688,10 +1688,20 @@ def test_letters_to_be_printed_sort_by_service(notify_db_session):
     second_service = create_service(service_name='second service', service_id='642bf33b-54b5-45f2-8c13-942a46616704')
     first_template = create_template(service=first_service, template_type='letter', postage='second')
     second_template = create_template(service=second_service, template_type='letter', postage='second')
-    notification_1 = create_notification(template=first_template, created_at=datetime(2020, 12, 1, 9, 30))
-    notification_2 = create_notification(template=first_template, created_at=datetime(2020, 12, 1, 12, 30))
-    notification_3 = create_notification(template=second_template, created_at=datetime(2020, 12, 1, 8, 30))
+    letters_ordered_by_service_then_time = [
+        create_notification(template=first_template, created_at=datetime(2020, 12, 1, 9, 30)),
+        create_notification(template=first_template, created_at=datetime(2020, 12, 1, 12, 30)),
+        create_notification(template=first_template, created_at=datetime(2020, 12, 1, 13, 30)),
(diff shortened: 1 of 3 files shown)


tests/app/dao/notification_dao/test_notification_dao.py
```diff
@@ -1688,10 +1688,20 @@ def test_letters_to_be_printed_sort_by_service(notify_db_session):
     second_service = create_service(service_name='second service', service_id='642bf33b-54b5-45f2-8c13-942a46616704')
     first_template = create_template(service=first_service, template_type='letter', postage='second')
     second_template = create_template(service=second_service, template_type='letter', postage='second')
-    notification_1 = create_notification(template=first_template, created_at=datetime(2020, 12, 1, 9, 30))
-    notification_2 = create_notification(template=first_template, created_at=datetime(2020, 12, 1, 12, 30))
-    notification_3 = create_notification(template=second_template, created_at=datetime(2020, 12, 1, 8, 30))
+    notis_ordered_by_service_then_time = [
```
CrystalPea, line 1691:
```suggestion
    letters_ordered_by_service_then_time = [
```

tests/app/dao/notification_dao/test_notification_dao.py
```diff
@@ -1688,10 +1688,20 @@ def test_letters_to_be_printed_sort_by_service(notify_db_session):
...
+        create_notification(template=second_template, created_at=datetime(2020, 12, 1, 8, 34))
+    ]
 
-    results = list(dao_get_letters_to_be_printed(print_run_deadline=datetime(2020, 12, 1, 17, 30), postage='second'))
-    assert len(results) == 3
-    assert results == [notification_1, notification_2, notification_3]
+    results = list(
+        dao_get_letters_to_be_printed(print_run_deadline=datetime(2020, 12, 1, 17, 30), postage='second', query_limit=4)
+    )
+    assert [x.id for x in results] == [x.id for x in notis_ordered_by_service_then_time]
```
CrystalPea, line 1707:
```suggestion
    assert [x.id for x in results] == [x.id for x in letters_ordered_by_service_then_time]
```

CrystalPea (approved):
That is very cool 👍 

I only have a small naming suggestion in the test.

PR #3022, opened 2020-10-28
client: cbc_proxy passes through sent/expires

tlwr:
What
----

A `BroadcastEvent` knows when an event was sent and should expire using the fields

* `sent` == `sent_at`
* `expires` == `transmitted_finishes_at`

We pass through these values directly to the CBC Proxy, because `BroadcastEvent` knows how they should be formatted

### DIFF
--- a/app/celery/broadcast_message_tasks.py
+++ b/app/celery/broadcast_message_tasks.py
@@ -29,6 +29,8 @@ def send_broadcast_event(broadcast_event_id):
             headline="GOV.UK Notify Broadcast",
             description=broadcast_event.transmitted_content['body'],
             areas=areas,
+            sent=broadcast_event.sent_at_as_cap_datetime_string,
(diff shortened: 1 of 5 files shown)


app/celery/broadcast_message_tasks.py
```diff
@@ -45,4 +49,6 @@ def send_broadcast_event(broadcast_event_id):
             description=broadcast_event.transmitted_content['body'],
             areas=areas,
```
rjbaker, line 50: I'm not sure we need the areas for a cancellation message

rjbaker (approved):

app/celery/broadcast_message_tasks.py
tlwr: `git grep 'unsubstantiated reckon'` 😆

rjbaker: We can tidy these up when we implement cancels and updates. Don't let it block merging.

PR #3027, opened 2020-11-04
Change how we validate the length of templates.

servingUpAces:
We want to add validation for an email that's too long, that way the user knows why the message is failing. At the moment if an email is too long it will get a technical failure, after the retries fail. This way the email post will get a validation error.

This PR adds a boolean to check_char_count, when validate_template is called. This is set to false if the notification is an email/sms from the API. We want to check the length after the document download file has been uploaded and the variable replaced with the link. Otherwise the check can fail.

Once this: https://github.com/alphagov/notifications-utils/pull/804 is reverted, we can update the utils version.

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -143,10 +143,19 @@ def check_if_service_can_send_to_number(service, number):
         return international_phone_info
 
 
-def check_content_char_count(template_with_content):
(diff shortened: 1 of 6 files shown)


CrystalPea (approved):
Looks good to me 👍

PR #3032, opened 2020-11-13
Log when we don't retry a callback

idavidmcdonald:
We don't retry any callbacks when it receives a 4xx status. We should
probably be aware of this happening and at the moment there is nothing
in our logs to easily identify whether the request failed and is being
retried or if it failed and is not being retried. This will enable us to
search our logs easily and figure out how much it's happening.

It's quite likely that we should in the future allow callbacks to retry
if they get a 429 http response (rate limiting) but we should do this in
a smart way (exponential backoff) and so this is a first step to being
aware of how big a problem it is in case we want to do something about
it.

Note, we already have a test for the behaviour that we don't retry for 4xx so I didn't need to add one
https://github.com/alphagov/notifications-api/blob/master/tests/app/celery/test_service_callback_tasks.py#L116

### DIFF
--- a/app/celery/service_callback_tasks.py
+++ b/app/celery/service_callback_tasks.py
@@ -85,7 +85,7 @@ def _send_data_to_service_callback_api(self, data, service_callback_url, token, 
         response.raise_for_status()
     except RequestException as e:
         current_app.logger.warning(
-            "{} request failed for notification_id: {} and url: {}. exc: {}".format(
+            "{} request failed for notification_id: {} and url: {}. exception: {}".format(
                 function_name,
                 notification_id,
                 service_callback_url,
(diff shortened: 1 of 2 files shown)


app/celery/service_callback_tasks.py
```diff
@@ -103,6 +103,15 @@ def _send_data_to_service_callback_api(self, data, service_callback_url, token,
                         notification_id
                     )
                 )
+        else:
+            current_app.logger.warning(
+                "{} callback is not being retried for notification_id: {} and url: {}. exc: {}".format(
```
CrystalPea, line 108: What is `exc`? Is it readable enough?

CrystalPea (approved):
Looks good, one little readability-related comment ☺️

PR #3035, opened 2020-11-17
Send broadcast events per provider

leohemsted:
depends on https://github.com/alphagov/notifications-api/pull/3036

At the moment, only EE is active.

Apologies for the big PR - it's best reviewed commit by commit, and any questions just ask. I'm _especially_ interested in naming things - I've struggled a huge amount with what to name these database tables, models and functions, and I'm very open to ideas! Even just highlighting things you don't think you'll enjoy typing in future queries and code. Anything where you think "i don't know what this is by looking at the name". I know some of these names are awful. Especially around the naming of `cbc_proxy_client`, which while it follows the naming convention of the other clients, is more of a wrapper/factory/etc.

Essentially this PR solves the two problems:

* how do we know if we sent a message to a specific provider
* every single piece of XML we send needs to have a unique identifier (ie: separate per provider).

This required the creation of a new table. A quick recap:

* `broadcast_message` contains notify-specific metadata. Holds draft broadcasts, approval statuses, etc.
* `broadcast_event` refers to a specific action we want to tell providers to do for a broadcast-message - alert, update, cancel
* `broadcast_provider_message` is a new table, and refers to a single blob of xml sent to a single provider

Now that we've got this table, we need to refer to it!

So split out the send message task (and the link test task) into separate tasks that call per provider. In the future we'll add logic to choose which providers to send to based on per environment platform settings, and any killswitches or test flags we might have.

We're calling the actual proxy client in a similar way, except, the identifier is now `broadcast_provider_message.id` instead of `broadcast_event.id`, and the previous references now get those provider message IDs rather than event IDs.

This uncovered the next issue, which is that the `CBCProxyClient` class had hardcoded definitions for lambda. We know that the different providers need to be called different ways, and they might have different responses, error codes, etc. So make a wrapper class that you ask for a proxy for a given provider, and it returns you a class that knows how to transform data etc.


servingUpAces:
This is looking good. 
It might be worth saving the actual XML to the broadcast_provider_message table. That way we have a record in case the code to create the XML changes. I suppose the lambda could return that XML in the response. 
Other considerations:
What state transitions are allowed for the broadcast_provider_message? Can you go from tech-failure to acknowledge?
Does the response from the lambda return a status of acknowledge or error? Or do we get that response asynchronously?
How are the response times from the lambda?
If we get an exception we will need to retry the task, for instance if we get a 500 from the lambda, we would want to try again. 
Do we only set it as technical-failure when we hit max retries?

servingUpAces (approved):
I've made some comments, however, those can be considered/implemented in a different PR.

leohemsted:
> This is looking good.
> It might be worth saving the actual XML to the broadcast_provider_message table. That way we have a record in case the code to create the XML changes. I suppose the lambda could return that XML in the response.

Definitely agree, that's worth looking at.

> Other considerations:
> What state transitions are allowed for the broadcast_provider_message? Can you go from tech-failure to acknowledge?

I've only envisioned one so far:

sending -> received_ack, received_err, technical-failure

i imagine technical failure being a terminating state, same as current. If we need to retry, then the broadcast_provider_message will stay in sending. Maybe it's worth renaming `sending` to `created` as that's more accurate

> Does the response from the lambda return a status of acknowledge or error? Or do we get that response asynchronously?

the lambda currently doesn't return anything, but we can easily set that up. For one2many at least, the lambda gets an XML blob in response, i've seen examples with <msgType>Ack</msgType>, i could have a dig around to see if we've ever got any error response.

> How are the response times from the lambda?

the BT/EE lambda at least takes 700ms-1500ms typically.

> If we get an exception we will need to retry the task, for instance if we get a 500 from the lambda, we would want to try again.

Yup. Haven't done any error handling yet.

> Do we only set it as technical-failure when we hit max retries?

That was my intention.

PR #3044, opened 2020-11-25
add CBC_PROXY_ENABLED config flag to control if tasks are triggered

leohemsted:
previously we made some incorrect assumptions about set-up on staging and prod - they currently don't have any cbc_proxy aws creds at all.

We shoudn't be attempting canaries or link tests when there's no AWS infrastructure to connect to.

We also shouldn't bother writing a row into the database at all for the broadcast_provider_message since we're not even attempting to send, and we shouldn't get confused between messages that failed and messages we never wanted to send at all.


CrystalPea (approved):
LGTM 👍

PR #3047, opened 2020-11-30
Tweak sending time metrics to only include live notifications

idavidmcdonald:
Review commit by commit

Changes the high volume and not high volume metrics to both only include
non test notifications. This is because when looking at the grafana
metrics, it was impossible to tell what affect the high volume/non high
volume effect was having vs the test/live notification effect.

This leaves us with no break down of high volume/not high volume sending
times for test notifications but I don't think we really need that.

image


CrystalPea (approved):
looks good to me 👍

PR #3048, opened 2020-12-01
Fix comparison of uuid to string

idavidmcdonald:
`service.id` is a uuid so will not be matched to anything in
`current_app.config.get('HIGH_VOLUME_SERVICE')` because that is a list
of strings.

This is why we are never falling into the first if statement and having
any metrics for high volume services on our dashboards at the moment.

Note, I had taken the existing line from the `post_notification`
endpoint, but that is using a serialised service which already has the
UUID converted to a string.

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -80,7 +80,7 @@ def send_sms_to_provider(notification):
             statsd_client.timing("sms.test-key.total-time", delta_seconds)
         else:
             statsd_client.timing("sms.live-key.total-time", delta_seconds)
-            if service.id in current_app.config.get('HIGH_VOLUME_SERVICE'):
+            if str(service.id) in current_app.config.get('HIGH_VOLUME_SERVICE'):
                 statsd_client.timing("sms.live-key.high-volume.total-time", delta_seconds)
             else:
                 statsd_client.timing("sms.live-key.not-high-volume.total-time", delta_seconds)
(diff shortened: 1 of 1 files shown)


CrystalPea (approved):

leohemsted (changes requested):
I notice that we're not converting to a string when chosing how to process the POST notification: https://github.com/alphagov/notifications-api/blob/master/app/v2/notifications/post_notifications.py#L207

at first i thought we see it working because we cache, which will probably store the uuid as a `string` rather than a `UUID`, but then the tests appear to work fine? Need to do some quick investigation here I think

servingUpAces:
The `service` used here: https://github.com/alphagov/notifications-api/blob/master/app/v2/notifications/post_notifications.py#L207 is from `authenticated_service` which is the serialised object we store after the auth function.

leohemsted (approved):
> So the difference, at least according to our tests is that the post_notifications route is
> `<class 'app.serialised_models.SerialisedService'> for the service`
> `<class 'str'> for service.id`
> Whereas the delivery one I was editing is dealing with
> `<class 'app.models.Service'> for the service`
> `<class 'uuid.UUID'> for service.id`

aahh that's what i hadn't twigged - that in the POST, even when the cache is empty/disabled and it hits the DB, it still gets wrapped in a SerialisedService object.

okay with that discovered, i think this PR is fine.

for completeness and to avoid confusion in the future it might be prudent to change the post_notification check to be `str(service.id)` - even though we know that currently it'll be a string anyway, it'd be good to prevent this kind of confusing gotcha in future.

quis:
I think a better future change would be to abstract this into the model so you can write:
```python
if service.is_high_volume:
  # do stuff
```
Then you don’t have to remember to do the string conversion the next time around.

idavidmcdonald:
Good idea CHS, should have thought of that before! If I touch any high volume code again then I'll implement that.

servingUpAces:
In this case you would need to implement `service.is_high_volume` on both the Service model and SerialisedService.

PR #3051, opened 2020-12-02
Restrict providers per service

leohemsted:
We'll just insert manually in the DB for now since this is such a low volume thing.

returns via api so admin app can style appropriately.

### DIFF
--- a/app/celery/broadcast_message_tasks.py
+++ b/app/celery/broadcast_message_tasks.py
@@ -16,13 +16,15 @@ def send_broadcast_event(broadcast_event_id):
         current_app.logger.info(f'CBC Proxy disabled, not sending broadcast_event {broadcast_event_id}')
         return
 
+    broadcast_event = dao_get_broadcast_event_by_id(broadcast_event_id)
     for provider in current_app.config['ENABLED_CBCS']:
-        # TODO: Decide whether to send to each provider based on platform admin, service level settings, broadcast
-        # level settings, etc.
-        send_broadcast_provider_message.apply_async(
-            kwargs={'broadcast_event_id': broadcast_event_id, 'provider': provider},
-            queue=QueueNames.NOTIFY
-        )
+        if broadcast_event.service.allowed_broadcast_provider in {None, provider}:
+            # There may be future checks here to decide whether to send to each provider based on platform admin level
+            # settings
+            send_broadcast_provider_message.apply_async(
+                kwargs={'broadcast_event_id': broadcast_event_id, 'provider': provider},
+                queue=QueueNames.NOTIFY
+            )
 
 
 @notify_celery.task(name="send-broadcast-provider-message")
--- a/app/celery/broadcast_message_tasks.py
+++ b/app/celery/broadcast_message_tasks.py
@@ -17,14 +17,11 @@ def send_broadcast_event(broadcast_event_id):
         return
 
     broadcast_event = dao_get_broadcast_event_by_id(broadcast_event_id)
-    for provider in current_app.config['ENABLED_CBCS']:
-        if broadcast_event.service.allowed_broadcast_provider in {None, provider}:
-            # There may be future checks here to decide whether to send to each provider based on platform admin level
-            # settings
-            send_broadcast_provider_message.apply_async(
-                kwargs={'broadcast_event_id': broadcast_event_id, 'provider': provider},
-                queue=QueueNames.NOTIFY
-            )
+    for provider in broadcast_event.service.get_available_broadcast_providers():
+        send_broadcast_provider_message.apply_async(
+            kwargs={'broadcast_event_id': broadcast_event_id, 'provider': provider},
+            queue=QueueNames.NOTIFY
+        )
 
 
 @notify_celery.task(name="send-broadcast-provider-message")
@@ -80,19 +77,6 @@ def send_broadcast_provider_message(broadcast_event_id, provider):
 
 @notify_celery.task(name='trigger-link-test')
 def trigger_link_test(provider):
-    """
-    Currently we only have one hardcoded CBC Proxy, which corresponds to one
-    CBC, and so currently we do not specify the CBC Proxy name
-
-    In future we will have multiple CBC proxies, each proxy corresponding to
-    one MNO's CBC
-
-    This task should invoke other tasks which do the actual link tests, eg:
-    for cbc_name in app.config.ENABLED_CBCS:
-        send_link_test_for_cbc(cbc_name)
-
-    Alternatively this task could be configured to be a Celery group
-    """
     identifier = str(uuid.uuid4())
     message = f"Sending a link test to CBC proxy for provider {provider} with ID {identifier}"
     current_app.logger.info(message)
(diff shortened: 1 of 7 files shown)


tests/app/service/test_rest.py
```diff
@@ -268,6 +271,7 @@ def test_get_service_by_id(admin_request, sample_service):
         'volume_email',
         'volume_letter',
         'volume_sms',
+
```
idavidmcdonald: This line probably isn't meant to be added?

tests/conftest.py
```diff
@@ -103,7 +103,7 @@ def sms_providers(notify_db):
 
 @pytest.fixture(scope='function')
 def notify_db_session(notify_db, sms_providers):
-    yield notify_db
+    yield notify_db.session
```
idavidmcdonald, line 106: Could you explain this change for me please?

app/celery/broadcast_message_tasks.py
```diff
@@ -16,13 +16,15 @@ def send_broadcast_event(broadcast_event_id):
...
     for provider in current_app.config['ENABLED_CBCS']:
-        # TODO: Decide whether to send to each provider based on platform admin, service level settings, broadcast
-        # level settings, etc.
-        send_broadcast_provider_message.apply_async(
-            kwargs={'broadcast_event_id': broadcast_event_id, 'provider': provider},
-            queue=QueueNames.NOTIFY
-        )
+        if broadcast_event.service.allowed_broadcast_provider in {None, provider}:
+            # TODO: Decide whether to send to each provider based on platform admin, service level settings, broadcast
+            # level settings, etc.
```
idavidmcdonald, line 23: Is this comment now redundant because we have chosen whether to send to each provider based on a service level setting?

app/celery/broadcast_message_tasks.py
```diff
@@ -16,13 +16,15 @@ def send_broadcast_event(broadcast_event_id):
...
 
+    broadcast_event = dao_get_broadcast_event_by_id(broadcast_event_id)
     for provider in current_app.config['ENABLED_CBCS']:
-        # TODO: Decide whether to send to each provider based on platform admin, service level settings, broadcast
-        # level settings, etc.
-        send_broadcast_provider_message.apply_async(
-            kwargs={'broadcast_event_id': broadcast_event_id, 'provider': provider},
-            queue=QueueNames.NOTIFY
-        )
+        if broadcast_event.service.allowed_broadcast_provider in {None, provider}:
```
idavidmcdonald, line 21: For my own learning, what made you pick a set here over a list or a tuple?

idavidmcdonald, line 21: Optional, wondering if so we avoid this same logic potentially being seen across several places in our code in the future we, maybe in time, introduce a `service.enabled_broadcast_providers` or equivalent such that

```
def enabled_broadcast_providers(self):
    if self.allowed_broadcast_provider:
        return [self.allowed_broadcast_provider]
    else:
        return current_app.config['ENABLED_CBCS']
```

note, above could be written pythonically in a one liner

Reasons for why it might be a good thing to move to in the future
- might reduce some code duplication and moves the complexity into one method
- codes in a quite defensive way, I don't want us to risk accidentally having a bug in our code and sending to fewer providers than we should

app/dao/broadcast_message_dao.py
```diff
@@ -10,12 +10,8 @@ def dao_get_broadcast_message_by_id_and_service_id(broadcast_message_id, service
     ).one()
 
 
-def dao_get_broadcast_message_by_id(broadcast_message_id):
-    return BroadcastMessage.query.get(broadcast_message_id)
-
-
 def dao_get_broadcast_event_by_id(broadcast_event_id):
-    return BroadcastEvent.query.get(broadcast_event_id)
+    return BroadcastEvent.query.filter(BroadcastEvent.id == broadcast_event_id).one()
```
idavidmcdonald, line 14: Could you explain this change for me please?

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -16,18 +22,48 @@
...
+
+    template = create_template(sample_service, BROADCAST_TYPE)
+    broadcast_message = create_broadcast_message(template, status=BroadcastStatusType.BROADCASTING)
+    event = create_broadcast_event(broadcast_message)
+
+    mock_send_broadcast_provider_message = mocker.patch(
+        'app.celery.broadcast_message_tasks.send_broadcast_provider_message',
+    )
+
+    with set_config(notify_api, 'ENABLED_CBCS', ['ee', 'vodafone']):
```
idavidmcdonald, line 62: Might be an edge case but wonder if it's worth testing for the case where the single restricted broadcast provider is not in `ENABLED_CBCS`

idavidmcdonald (commented):
Looks pretty decent and I don't think it adds much complexity. Good work on the doc block description for the `ServiceBroadcastProviderRestriction` model too.

Few small questions/comments but should be good to approve shortly.

tests/conftest.py
leohemsted: this was to let us access the sqlalchemy session object itself from within test fixtures so we can call `session.add` and `session.commit` methods. 

however when doing a ctrl+f while writing this comment i notice we're already using `notify_db.session`, and `from app import db; db.session` in other places, so i'll change this PR to just use one of those (probs `notify_db.session.add`).

that said, i'll have to keep a mental post it note to look at this some time, as i don't think we should be ever using `notify_db` without `notify_db_session` since the notify_db_session fixture is the one that wipes down the tables and i cant think of a time we don't want that fixture to run after a test. so ideally i'd combine them into one fixture. but that's a task for another time

app/celery/broadcast_message_tasks.py
leohemsted: yeah partially, we may still have other checks here (if we have platform-admin level provider killswitches for example). but i'll change the wording to reflect that

app/dao/broadcast_message_dao.py
leohemsted: `.get` returns `None` if there's no value. `.one()` raises a sqlalchemy `NoResultFound`, so is a bit clearer as to why it failed rather than you being surprised by having a None value later.

(and given that we're getting by ID, we never expect the value not to be present yet anywhere that we call this)

app/celery/broadcast_message_tasks.py
leohemsted: it won't make a difference here, but in a larger list of things it's computationally quicker since set lookup is O(1) (memory location is defined by a hash of the contents, so looking up is a constant time complexity) while tuples behave like a list, so lookup is O(n) (as it has to scan the entire list).

obviously for two items (especially when it's almost always going to be None), the difference in this is minimal, but i think it's a good habit to use sets more for this kind of stuff

tests/conftest.py
leohemsted: i've squashed out this change for now to use the existing `notify_db.session.add` pattern.

idavidmcdonald (approved):

PR #3052, opened 2020-12-07
use cf-cli@v7 in makefile

leohemsted:
a bug in cf-cli v6 caused us to get rate limited, one solution to this is to bump the version of cf-cli we're using to version 7. This has a few syntax changes as the old v3 commands become mainline.

To upgrade locally, grab the latest version from brew.

```sh
brew install cloudfoundry/tap/cf-cli@7
```

I've tested:

- [x] deploying
- [x] cancelling an on-going deployment
- [x] running a task (db migration)
- [x] checking a task's status

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,18 +154,17 @@ cf-deploy: ## Deploys the app to Cloud Foundry
 	@cf app --guid ${CF_APP} || exit 1
 
 	# cancel any existing deploys to ensure we can apply manifest (if a deploy is in progress you'll see ScaleDisabledDuringDeployment)
-	cf v3-cancel-zdt-push ${CF_APP} || true
+	cf cancel-deployment ${CF_APP} || true
 
-	cf v3-apply-manifest ${CF_APP} -f <(make -s generate-manifest)
-	CF_STARTUP_TIMEOUT=15 cf v3-zdt-push ${CF_APP} --wait-for-deploy-complete  # fails after 15 mins if deploy doesn't work
-
+	# fails after 15 mins if deploy doesn't work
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -154,18 +154,17 @@ cf-deploy: ## Deploys the app to Cloud Foundry
 	@cf app --guid ${CF_APP} || exit 1
 
 	# cancel any existing deploys to ensure we can apply manifest (if a deploy is in progress you'll see ScaleDisabledDuringDeployment)
-	cf v3-cancel-zdt-push ${CF_APP} || true
-
-	cf v3-apply-manifest ${CF_APP} -f <(make -s generate-manifest)
```
idavidmcdonald, line 159: Is there a reason you didn't use `cf apply-manifest` here instead of moving the manifest into the `cf push` statement? It looks like what you are proposing does roughly the same thing but I just wanted to double check there were no differences? It seems like a sensible change, just wondering if there is a reason we didn't do it like this already before

leohemsted: yeah - i also read up on the docs. looks like it doesn't do anything separate and keeps it cleaner and simpler to reason about.

i know andy had a big masterplan around breaking up the deploy process into its separate steps, and then reusing the same droplet with different manifests per environment, so i wonder if we originally did it to try that.

i'd be keen to revisit that some time, especially as cf v7 makes those commands mainline rather than bodging it with cf curl commands (see https://docs.cloudfoundry.org/devguide/push-sub-commands.html). But that would be a significant task and would need us to re-architect the pipelines to keep track of droplet IDs etc. (benefit: guaranteed identical code on diff environments, quicker deploys, quicker rollbacks)

idavidmcdonald (approved):

PR #3053, opened 2020-12-07
Add sequential message number to broadcast provider messages

CrystalPea:
We need that to send broadcast messages using proprietary IBAG
format that Vodafone currently uses.

Best to:
1. first read all commit messages in a list of commits (to see the scope)
2. review commit by commit


rjbaker (commented):
I think we'll need to enable the Vodafone CBC in the app config: https://github.com/alphagov/notifications-api/blob/master/app/config.py#L381

app/celery/broadcast_message_tasks.py
```diff
@@ -78,6 +87,11 @@ def send_broadcast_provider_message(broadcast_event_id, provider):
 @notify_celery.task(name='trigger-link-test')
 def trigger_link_test(provider):
     identifier = str(uuid.uuid4())
+    formatted_seq_number = None
+    if provider == BroadcastProvider.VODAFONE:
+        sequence = Sequence('broadcast_provider_message_number_seq')
+        sequential_number = db.session.connection().execute(sequence)
+        formatted_seq_number = format_sequential_number(sequential_number)
```
leohemsted, line 94: i think i'd rather have this kind of logic inside the `CBCProxyVodafone` class. If every provider has a different if block in here and in the `broadcast_provider_dao.py::create_broadcast_provider_message` function as well,  it might get a bit messy, and that's exactly what the different cbc_proxy_clients are there to avoid.

However, that means putting DB stuff inside the cbc_proxy.py file, and that's not consistent with how we've done things so far, so that makes me a bit hesitant. Happy to hear other thoughts from the rest of the dev team 🤔

app/celery/broadcast_message_tasks.py
```diff
@@ -29,7 +32,10 @@ def send_broadcast_event(broadcast_event_id):
 def send_broadcast_provider_message(broadcast_event_id, provider):
     broadcast_event = dao_get_broadcast_event_by_id(broadcast_event_id)
 
-    broadcast_provider_message = create_broadcast_provider_message(broadcast_event, provider)
+    broadcast_provider_message, message_number = create_broadcast_provider_message(broadcast_event, provider)
```
leohemsted, line 35: i'm not a fan of returning a tuple normally - i think it's often confusing and unexpected, especially when the second value is `None` a lot of the time.

if we add a relationship and an association proxy object linking the ProviderMessage to MessageNumber and vice versa, then we can write `broadcast_provider_message.message_number` to access the value - I've used this pattern recently (look up allowed_broadcast_provider on the service model) and it seems to work well - i'm happy to talk through it with you

leohemsted (commented):
a few comments i think are worth thinking about. They're a bit big and quite opinionated, so don't be afraid to say no if you think your way is better (or if you think the difference between the two ways is so small its not worth the effort to rewrite). After all, you're the one who's been looking at the code these last few days 😄

app/celery/broadcast_message_tasks.py
leohemsted: agree this is a bit weird but not a big issue right now. keep an eye on it if it grows as we add more providers/fields/etc and we might want to move it later if necessary

app/clients/cbc_proxy.py
```diff
@@ -61,15 +62,30 @@ def send_canary(
...
         self,
         identifier,
+        sequential_number
     ):
-        pass
+        """
+        link test - open up a connection to a specific provider, and send them an xml payload with a <msgType> of
+        test.
+        """
+        payload = {'message_type': 'test', 'identifier': identifier, 'message_number': sequential_number}
```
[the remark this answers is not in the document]
CrystalPea: ok, I have split them out now ☺️

app/celery/broadcast_message_tasks.py
CrystalPea: As per our chat, let's leave it here now to try and avoid putting db stuff in the clients, but if it becomes unwieldy as more providers are added, we will look to refactor it then.

app/celery/broadcast_message_tasks.py
CrystalPea: We did this now ☺️

migrations/versions/0334_broadcast_message_number.py
```diff
@@ -0,0 +1,48 @@
...
+        sa.PrimaryKeyConstraint('broadcast_provider_message_number')
+    )
+    op.execute(
+        """
+            INSERT INTO
+                broadcast_provider_message_number (broadcast_provider_message_id)
+            SELECT
+                id
+            FROM
+                broadcast_provider_message
```
leohemsted, line 38: dont think we need this, since we don't insert new rows for EE messages. we only want to do this where `provider == 'vodafone'` right? (which should be zero broadcast_provider_messages right now)

app/celery/broadcast_message_tasks.py
```diff
@@ -30,6 +33,9 @@ def send_broadcast_provider_message(broadcast_event_id, provider):
     broadcast_event = dao_get_broadcast_event_by_id(broadcast_event_id)
 
     broadcast_provider_message = create_broadcast_provider_message(broadcast_event, provider)
+    formatted_message_number = None
+    if provider == BroadcastProvider.VODAFONE:
+        formatted_message_number = format_sequential_number(broadcast_provider_message.message_number)
```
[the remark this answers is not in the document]
CrystalPea: totally ^_^

migrations/versions/0334_broadcast_message_number.py
CrystalPea: true! :)

CrystalPea: Done ☺️

leohemsted (approved):

PR #3055, opened 2020-12-07
put manifest in tmp folder

leohemsted:
for reasons unknown, using a file descriptor no longer works on concourse (or in that docker container generally). It might be a change within cf-cli v7 vs v6.

Either way, it does't work, so use a temporary file. Clean up the temporary file afterwards.

If the command fails, the temporary file will still stick around, so I've added the file to the /tmp/ folder instead. it's full of secret
keys and things so if you do have a deployment error while running locally, you should make sure you clean up the file (make cf-rollback and make clean will both do this for you).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,8 @@ CF_SPACE ?= ${DEPLOY_ENV}
 CF_HOME ?= ${HOME}
 $(eval export CF_HOME)
 
-CF_MANIFEST_FILE = manifest-$(firstword $(subst -, ,$(subst notify-,,${CF_APP})))-${CF_SPACE}.yml
+CF_MANIFEST_PATH ?= /tmp/manifest.yml
+
 
 NOTIFY_CREDENTIALS ?= ~/.notify-credentials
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -156,14 +157,23 @@ cf-deploy: ## Deploys the app to Cloud Foundry
...
 	cf cancel-deployment ${CF_APP} || true
 
+	# generate manifest (including secrets) and write it to CF_MANIFEST_PATH (in /tmp/)
+	make -s CF_APP=${CF_APP} generate-manifest > ${CF_MANIFEST_PATH}
+
 	# fails after 15 mins if deploy doesn't work
-	CF_STARTUP_TIMEOUT=15 cf push ${CF_APP} --strategy=rolling -f <(make -s generate-manifest)
+	# reads manifest from CF_MANIFEST_PATH
+	CF_STARTUP_TIMEOUT=15 cf push ${CF_APP} --strategy=rolling -f ${CF_MANIFEST_PATH}
+	# delete old manifest file
```
idavidmcdonald, line 166: Missing the delete command

leohemsted: d'oh! squashed in

idavidmcdonald (approved):

PR #3059, opened 2020-12-14
Add cancel routes to cbc proxy clients

CrystalPea:
This is so we could send cancel messages to the lambda.

### DIFF
--- a/app/clients/cbc_proxy.py
+++ b/app/clients/cbc_proxy.py
@@ -217,6 +217,9 @@ class CBCProxyVodafone(CBCProxyClientBase):
     def cancel_broadcast(
         self, identifier, previous_provider_messages, sent, message_number
     ):
+        # avoid cyclical import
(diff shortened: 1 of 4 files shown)


app/clients/cbc_proxy.py
```diff
@@ -163,6 +160,22 @@ def create_and_send_broadcast(
...
+    def cancel_broadcast(
+        self,
+        identifier, previous_provider_messages,
+        sent, message_number=None
+    ):
+        payload = {
+            'message_type': 'cancel',
+            'identifier': identifier,
+            'message_format': 'cap',
+            "references": [
```
rjbaker, line 172: Nice 👍 As discussed the reason for the array is because the CAP standard requires all references to a previous alert to be concatenated. E.g.
```
Alert->Update (references Alert)->Cancel (references Update & Alert)

rjbaker (approved):

PR #3060, opened 2020-12-14
Add public API endpoint to create broadcast messages

quis:
We know there is at least one system which wants to integrate with Notify to send out emergency alerts, rather than creating them manually.

This commit adds an endpoint to the public API to let them do that.

To start with we’ll just let the system create them in a single call, meaning they still have to be approved manually. This reduces the risk of an attacker being able to broadcast an alert via the API, should the other system be compromised.

***

This pull request also allows the API to accept CAP XML instead of JSON. We know the other system won’t easily be adapted to emit JSON, so it’s less work for us to do and be in control of the conversion.

***

At the moment this allows creating broadcast messages with any type of API key. We can lock this down a bit once we know what the model for test/live/team combined with live/training mode should look like.


app/v2/broadcast/post_broadcast.py
```diff
@@ -0,0 +1,66 @@
...
+        api_key_id=api_user.id,
+        # The client may pass in broadcast_json['expires'] but it’s
+        # simpler for now to ignore it and have the rules around expiry
+        # for broadcasts created with the API match those created from
+        # the admin app
+    )
+
+    dao_save_object(broadcast_message)
+
+    return jsonify(broadcast_message.serialize()), 201
```
leohemsted, line 66: i wonder if we should return values in xml if we receive xml? 🤔 

also, do you know if they expect anything specific in the response? they might want a link to the approval page, for example

requirements-app.txt
```diff
@@ -23,13 +23,15 @@ SQLAlchemy==1.3.22
 strict-rfc3339==0.7
 rfc3987==1.3.8
 cachetools==4.2.0
+beautifulsoup4==4.9.3
+lxml==4.6.2
```
leohemsted, line 27: we'll need to be super careful to make sure this is kept up-to-date as it's parsing untrusted xml from the public internet

app/v2/broadcast/post_broadcast.py
```diff
@@ -0,0 +1,66 @@
...
+        },
+        status=BroadcastStatusType.PENDING_APPROVAL,
+        api_key_id=api_user.id,
+        # The client may pass in broadcast_json['expires'] but it’s
+        # simpler for now to ignore it and have the rules around expiry
+        # for broadcasts created with the API match those created from
+        # the admin app
+    )
+
+    dao_save_object(broadcast_message)
```
leohemsted, line 64: can we log `created a broadcast_message for with id {x} for reference {y} for service {z}` here? or similar.

app/v2/broadcast/post_broadcast.py
```diff
@@ -0,0 +1,66 @@
...
+    broadcast_message = BroadcastMessage(
+        service_id=authenticated_service.id,
+        content=broadcast_json['content'],
+        reference=broadcast_json['reference'],
+        areas={
+            'areas': [
+                area['name'] for area in broadcast_json['areas']
+            ],
+            'simple_polygons': polygons.smooth.simplify.as_coordinate_pairs_long_lat,
+        },
```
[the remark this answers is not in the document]
quis: Pull request to make the frontend not baulk when it gets unknown areas: https://github.com/alphagov/notifications-admin/pull/3780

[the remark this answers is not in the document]
quis: We can think later about whether we want to store the original polygons somewhere for auditing. They’re maybe too big to go in the database.

app/v2/broadcast/post_broadcast.py
```diff
@@ -0,0 +1,66 @@
...
+@v2_broadcast_blueprint.route("", methods=['POST'])
+def create_broadcast():
+
+    check_service_has_permission(
+        BROADCAST_TYPE,
+        authenticated_service.permissions,
+    )
+
+    if request.content_type == 'application/json':
+        broadcast_json = request.get_json()
```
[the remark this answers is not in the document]
quis: I’ve removed the option to pass in JSON in e3e68e4

app/v2/broadcast/post_broadcast.py
```diff
@@ -0,0 +1,66 @@
...
+        if not validate_xml(cap_xml, 'CAP-v1.2.xsd'):
+            raise BadRequestError(
+                message=f'Request data is not valid CAP XML',
+                status_code=400,
+            )
+        broadcast_json = cap_xml_to_dict(cap_xml)
+    else:
+        raise BadRequestError(
+            message=f'Content type {request.content_type} not supported',
+            status_code=400,
```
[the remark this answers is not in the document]
quis: I like being pedantic about using the correct status code. Changed in 62c8403 🤓

app/v2/broadcast/post_broadcast.py
quis: We can ask this on Friday, find out what they do with the response for text messages at the moment.

app/v2/broadcast/post_broadcast.py
quis: Added in 12d56555

requirements-app.txt
quis: Yeah. It’s a shame there isn’t a library in [the `defusedxml` module](https://pypi.org/project/defusedxml/) which can validate XML against a schema.

leohemsted (approved):
think this looks good enough to get in front of EA for now 🤞

PR #3067, opened 2020-12-18
Fix cancel broadcast by converting reference date to string

CrystalPea:
Datetime object is not json-serializable, we have to convert it to string for the created_at field of previous broadcast provider messages.

Also move DATETIME_FORMAT from app to app.utils to avoid cyclical import issues

Best to look commit by commit ☺️

### DIFF
--- a/tests/app/clients/test_cbc_proxy.py
+++ b/tests/app/clients/test_cbc_proxy.py
@@ -1,10 +1,12 @@
 import json
 import uuid
 from collections import namedtuple
+from datetime import datetime
 from unittest.mock import Mock
 
 import pytest
 
+from app import DATETIME_FORMAT
 from app.clients.cbc_proxy import CBCProxyClient, CBCProxyException, CBCProxyEE, CBCProxyCanary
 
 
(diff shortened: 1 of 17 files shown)


tests/app/clients/test_cbc_proxy.py
```diff
@@ -161,11 +165,11 @@ def test_cbc_proxy_ee_cancel_invokes_function(mocker, cbc_proxy_ee):
     assert payload['references'] == [
         {
             "message_id": str(provider_messages[0].id),
-            "sent": provider_messages[0].created_at
+            "sent": provider_messages[0].created_at.strftime(DATETIME_FORMAT)
```
idavidmcdonald, line 168: Optional: Could be worth changing this assertion to check the string value rather than calling `strftime(DATETIME_FORMAT)` on it. Your test currently means you are confident that you are calling `strftime(DATETIME_FORMAT)` on `created_at` but it doesn't check that it is giving you the actual string you wish. If for some reason `strftime(DATETIME_FORMAT)` wasn't the right thing to call, or did something different to what you expected, then your test wouldn't currently pick this up.

An example would be

```
def times_number_by_2(number):
    return number ^ 2
```

```
def test_number_is_times_by_2():
    assert times_number_by_2(4) == 4 ^2
```

However, the mistake would be that you haven't timed the number by 2, you've infact put it to the power of 2.

Note, this approach isn't a hard rule that you should always check it in the way I've suggested as I'm fairly confident that the change you'ved made is correct but it is one to consider in future.

idavidmcdonald (approved):
Looks good 🚢

tests/app/clients/test_cbc_proxy.py
CrystalPea: That's a valid point, and I will keep it in mind for the future. Not making changes on this occasion just because it would be great to still do some testing today ☺️

PR #3069, opened 2020-12-22
Pass request_id onto the task if called from a task

servingUpAces:
If the request_id exists in the Flask global context, add it to the kwargs for the task.

The request_id is set is the task is created from a http request, if that task then calls through to another task this will set the request_id from the global context. We should then be able to follow the creation of a notification all the way from the original http request to the sending task.

### DIFF
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -58,6 +58,8 @@ def make_task(app):
 
             if has_request_context() and hasattr(request, 'request_id'):
                 kwargs['request_id'] = request.request_id
+            elif g.request_id:
+                kwargs['request_id'] = g.request_id
 
             with SQS_APPLY_ASYNC_DURATION_SECONDS.labels(self.name).time():
                 return super().apply_async(args, kwargs, task_id, producer, link, link_error, **options)
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -1,10 +1,9 @@
 import time
-
 from gds_metrics.metrics import Histogram
 from celery import Celery, Task
 from celery.signals import worker_process_shutdown
 from flask import g, request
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):
Would be nice if there was an easy test to support it but it's not the end of the world

servingUpAces:
I agree

app/celery/celery.py
```diff
@@ -58,6 +58,8 @@ def apply_async(self, args=None, kwargs=None, task_id=None, producer=None,
 
             if has_request_context() and hasattr(request, 'request_id'):
                 kwargs['request_id'] = request.request_id
+            elif g.request_id:
```
idavidmcdonald, line 61: Oh, will this raise an attribute error if there is g does not have `request_id` as an attribute?

I just opened a `flask shell` and did `g.thing` and got an attribute error so wondering if we are at risk of the same here?

app/celery/celery.py
```diff
@@ -58,6 +57,11 @@ def apply_async(self, args=None, kwargs=None, task_id=None, producer=None,
 
             if has_request_context() and hasattr(request, 'request_id'):
                 kwargs['request_id'] = request.request_id
+            else:
```
idavidmcdonald, line 60: Optional: What you've written is all good I think it looks a bit cleaner as the following

```
elif g.hasattr("request_id") and g.request_id:
    kwargs['request_id'] = g.request_id
```

But as I said, all fine so you can pick what you like most

idavidmcdonald (approved):

app/celery/celery.py
servingUpAces: g doesn't have hasattr(), but I've taken the same approach as notifications-utils

PR #3070, opened 2020-12-22
Remove mutable defaults from test fixture

quis:
Using mutable default arguments is an anti pattern. They are created during function definition time. All calls to the function reuse this one instance of that data structure, persisting changes between them. This results in unexpected bugs.


quis:
Closing in favour of https://github.com/alphagov/notifications-api/pull/3071

PR #3071, opened 2020-12-22
Do extra code style checks with flake8-bugbear

quis:
Flake8 Bugbear checks for some extra things that aren’t code style errors, but are likely to introduce bugs or unexpected behaviour. A good example is having mutable default function arguments, which get shared between every call to the function and therefore mutating a value in one place can unexpectedly cause it to change in another.

This commit enables all the extra warnings provided by Flake8 Bugbear,
except for:
- the line length one (because we already lint for that separately)
- B903 Data class should either be immutable or use `__slots__` because this seems to false-positive on some of our custom exceptions
- B902 Invalid first argument 'cls' used for instance method because some SQLAlchemy decorators (eg `declared_attr`) make things that aren’t formally class methods take a class not an instance as their first argument

It disables:
- _B306: BaseException.message is removed in Python 3_ because I think our exceptions have a custom structure that means the `.message` attribute is still present

Matches the work done in other repos:
- https://github.com/alphagov/notifications-admin/pull/3172

### DIFF
--- a/requirements_for_test.txt
+++ b/requirements_for_test.txt
@@ -1,5 +1,6 @@
 -r requirements.txt
 flake8==3.8.4
+flake8-bugbear==20.11.1
 moto==1.3.16
 pytest==6.1.2
 pytest-env==0.6.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,4 +4,5 @@ xfail_strict=true
 
 [flake8]
 # W504 line break after binary operator
-extend_ignore=W504
+extend_ignore=B306, W504
+select=B901
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,6 +3,7 @@ xfail_strict=true
 
 
 [flake8]
+exclude = venv*,__pycache__,node_modules,cache,migrations,build
 # W504 line break after binary operator
 extend_ignore=B306, W504
 select=B901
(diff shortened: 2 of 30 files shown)


setup.cfg
```diff
@@ -4,4 +4,5 @@ xfail_strict=true
 
 [flake8]
```
idavidmcdonald, line 5: We also have configuration in `.flake8`. From my reading of the docs I suspect that this file is taking precendent but I'm not 100% sure on that. https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations

Does this mean that we can get rid of the `.flake8` file?

requirements_for_test.txt
```diff
@@ -1,5 +1,6 @@
 -r requirements.txt
 flake8==3.8.4
+flake8-bugbear==20.11.1
```
idavidmcdonald, line 3: Do we want to use flake8-print too?

idavidmcdonald (approved):

requirements_for_test.txt
quis: We use `print` in a few different files in this repo, for example [in `commands.py`](https://github.com/alphagov/notifications-api/blob/7152ac7cba4a8f1cf9c9b378d3fe6bae0cf75b4f/app/commands.py) so I decided against it.

setup.cfg
quis: Removed `.flake8` in 0ed1f329

PR #3075, opened 2020-12-23
Cache provider lookups for 10 seconds

quis:
For every email or text message we send we have to work out which provider to send it with. Every time we do this we go and load the list of providers from the database.

For emails, the result will always be the same.

For text messages the result is randomly chosen to balance the load between the providers.

For international text messages the result is always the same (we only have one international text message provider).

This pull request adds an in-memory cache with a 2 second time-to-live so that we’re not fetching the providers from the database every time, which should speed things up a bit.

This does mean that, for text messages, the random choice will ‘stick’ for two seconds on each instance, before being re-chosen. I think this is OK because it will even out to the same distribution over time.

I don’t like having to clear the cache in the tests, so would welcome suggestions on a better way of doing this…

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -148,7 +148,7 @@ def update_notification_to_sending(notification, provider):
     dao_update_notification(notification)
 
 
-provider_cache = TTLCache(maxsize=8, ttl=2)
+provider_cache = TTLCache(maxsize=8, ttl=10)
 
 
 @cached(cache=provider_cache)
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -1,7 +1,7 @@
 import random
 from urllib import parse
 from datetime import datetime, timedelta
-
+from cachetools import TTLCache, cached
 from flask import current_app
 from notifications_utils.recipients import (
     validate_and_format_phone_number,
@@ -148,6 +148,10 @@ def update_notification_to_sending(notification, provider):
     dao_update_notification(notification)
 
 
+provider_cache = TTLCache(maxsize=8, ttl=2)
+
+
+@cached(cache=provider_cache)
 def provider_to_use(notification_type, international=False):
     active_providers = [
         p for p in get_provider_details_by_notification_type(notification_type, international) if p.active
--- a/tests/app/delivery/test_send_to_providers.py
+++ b/tests/app/delivery/test_send_to_providers.py
@@ -47,6 +47,23 @@ def test_provider_to_use_should_return_random_provider(mocker, notify_db_session
     assert ret.get_name() == 'mmg'
 
 
+def test_provider_to_use_should_cache_repeated_calls(mocker, notify_db_session):
+    mock_choices = mocker.patch(
+        'app.delivery.send_to_providers.random.choices',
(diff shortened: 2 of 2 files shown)


quis:
Here’s a chart showing how long it takes to choose a provider with no caching, with a cache that lives for 2 seconds and a cache that lives for 5 seconds:

image

(run over a CSV of 25 rows, with 4 workers)

app/delivery/send_to_providers.py
```diff
@@ -148,6 +148,10 @@ def update_notification_to_sending(notification, provider):
     dao_update_notification(notification)
 
 
+provider_cache = TTLCache(maxsize=8, ttl=2)
```
idavidmcdonald, line 151: Am I correct in thinking that maxsize 8 is chosen from 4 choices of `notification_type` multiplied by 2 choices of `international`? And if we were to add another notification type we would need to change this value?

idavidmcdonald, line 151: TTL 2 seconds? Yet you say in your comment `I think this shows that you really start to see the benefit with the 10s cache, where >50% of queries are skipping the database.`. Does that not encourage a longer TTL?

tests/app/delivery/test_send_to_providers.py
```diff
@@ -624,6 +643,7 @@ def test_should_send_sms_to_international_providers(
         international=True,
         reply_to_text=sample_template.service.get_default_sms_sender()
     )
+    send_to_providers.provider_cache.clear()
```
idavidmcdonald, line 646: Hmm, it's not so much that I hate that we have to include these lines to clear the cache, it's more that if we were to write another test, we would likely forget about this and not include it. 

Option. Put all these tests in a class and then use a method setup (https://docs.pytest.org/en/stable/xunit_setup.html#method-and-function-level-setup-teardown) to ensure the cache is cleared. To be super safe you'd add it as a setup to ALL tests where we delete all caches everywhere before any test but I think that is probably overkill.

idavidmcdonald (commented):
Few comments.

If we wanted to go even further (but I don't suggest you do now, if at all), you could argue that we could rearrange our code such that email providers, letter providers, broadcast providers and international sms are cached for a much longer time as it's really only the SMS balancing we want to not be cached for too long. A toss up between hardcoding some of those things in our code though and it being quicker vs our code being more flexible for adding more providers.

You could also put the active providers in redis given that information doesn't change very often. But one for another day if we decide to do that.

app/delivery/send_to_providers.py
quis: Changed to 10 seconds in 55afc9a

tests/app/delivery/test_send_to_providers.py
quis: Added a setup function in 624bd1d

app/delivery/send_to_providers.py
quis: 2 choices of notification type, because letters don’t use a provider in the same way. So I did 3, rounded up to nearest power of 2, then doubled. The aim was to get a number big enough that we never have to worry about it being a constraint, but not so big we ever have to worry about it consuming a lot of memory.

idavidmcdonald (approved):

PR #3076, opened 2020-12-23
Fix app logging

sakisv:
The increased traffic highlighted two issues with our logs:

The first issue was that sender workers, because of the `celery multi`, were keeping duplicate logs: The ones generated from our application and the ones generated from celery itself.

The second issue was around handling the scenario were the instances would run out of disk space. Given the constraints of our runtime environment the simplest solution is to recycle the instances, when their disk usage is over a certain threshold.

For now, I set the threshold to 75%, same as the alert threshold.

### DIFF
--- a/scripts/paas_app_wrapper.sh
+++ b/scripts/paas_app_wrapper.sh
@@ -22,7 +22,7 @@ case $NOTIFY_APP_NAME in
     ;;
   delivery-worker-sender)
     exec scripts/run_multi_worker_app_paas.sh celery multi start 3 -c 10 -A run_celery.notify_celery --loglevel=INFO \
-    -Q send-sms-tasks,send-email-tasks
+    --logfile=/dev/null --pidfile=/tmp/celery%N.pid -Q send-sms-tasks,send-email-tasks
     ;;
   delivery-worker-periodic)
     exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=2 \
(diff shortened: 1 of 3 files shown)


scripts/paas_app_wrapper.sh
```diff
@@ -22,7 +22,7 @@ case $NOTIFY_APP_NAME in
     ;;
   delivery-worker-sender)
     exec scripts/run_multi_worker_app_paas.sh celery multi start 3 -c 10 -A run_celery.notify_celery --loglevel=INFO \
-    -Q send-sms-tasks,send-email-tasks
+    --logfile=/dev/null --pidfile=/tmp/celery%N.pid -Q send-sms-tasks,send-email-tasks
```
idavidmcdonald, line 25: It's not just the sender worker that uses up disk space quickly. I assume this double logging is happening on all our celery workers. If that is correct, then we should probably apply this approach to all workers or is there a reason not to do that?

sakisv: True, it's not just the sender worker that uses up space quickly, but it is the only one that has duplicate logs. For the rest of the workers, by default, if you don't define a logfile, celery is printing to stderr [[link]] and we're redirecting the stderr to `/dev/null/`

[link]: https://docs.celeryproject.org/en/3.1/_modules/celery/app/log.html#Logging.setup_task_loggers

idavidmcdonald (approved):

PR #3082, opened 2021-01-04
Add disk space check for sender worker

sakisv:
Reused the existing `ensure_celery_is_running` function to terminate the
script

Complementary of #3076

### DIFF
--- a/scripts/run_multi_worker_app_paas.sh
+++ b/scripts/run_multi_worker_app_paas.sh
@@ -3,6 +3,7 @@
 set -e -o pipefail
 
 TERMINATE_TIMEOUT=9
+MAX_DISK_SPACE_USAGE=75
 readonly LOGS_DIR="/home/vcap/logs"
 
 function check_params {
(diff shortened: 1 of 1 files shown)


CrystalPea (approved):
Good work, although we could think about the threshold. For alert, it has been 75%, because that gave us plenty of time to react. Like, if the alert produced a ticket in the evening and we only saw it in the morning, it would still not be too late.

With automated termination of the instance, we could go for bigger threshold. But then I guess we would have to get rid of the alert (we have another more urgent alert when apps reach 95% disk space), and we do not have this automated termination turned on for all apps in the scope of the alert, so maybe increasing the threshold is something we could do when we turn this on for all apps, and then get rid of that alert altogether? What do you think?

sakisv:
I agree with everything you said :)

#3076 was done somewhat urgently because we had the 4 day Christmas holiday coming up, that's why I just went with the alarm threshold, and I used the same value here for consistency with that PR.

I think it's safe to remove the 75% warning for api and workers, because that should be taken care by these changes. I would leave the 95% alert in place because if we reach that point it means that for whatever reason this has failed and someone needs to take a look at what's happening _anyway_.

Do we also have alerts for admin, template-preview, etc?

CrystalPea:
@sakisv 

> I think it's safe to remove the 75% warning for api and workers, because that should be taken care by these changes. I would leave the 95% alert in place because if we reach that point it means that for whatever reason this has failed and someone needs to take a look at what's happening anyway.

Agreed ☺️ 

> Do we also have alerts for admin, template-preview, etc?

I don't think we do.

sakisv:
> I don't think we do.

So the alert is across all the apps: https://github.com/alphagov/prometheus-aws-configuration-beta/blob/master/terraform/modules/prom-ec2/alerts-config/alerts/notify-alerts.yml (and on [prom1](https://prom-1.monitoring.gds-reliability.engineering/graph?g0.range_input=1h&g0.expr=%20max(disk_utilization%7Bspace%3D%22production%22%2C%20organisation%3D%22govuk-notify%22%7D)%20by%20(app%2C%20space)&g0.tab=1))

This makes me more inclined to keep the alert in place as it is.

We don't need to worry about alert noise because this alert will only fire if the threshold is breached for 5 minutes, and we're checking every second.

CrystalPea:
@sakisv ahh, you're right :o. Makes sense!

PR #3083, opened 2021-01-08
Cache service callback API in memory 

quis:
This would mean the callback task for emails wouldn’t need to open a database connection at all.


quis:
Actually I understood the code wrong. This might mean closing the connection slightly earlier, but it wouldn’t skip the connection entirely.  And it would only speed up how fast we can create the callback tasks, not how fast we send them out. The task which sends the callback to the service already avoids the database.

PR #3084, opened 2021-01-08
add content to broadcast_message and make template fields nullable

leohemsted:
we want to be able to create broadcast messages without templates. To start with, these will come from the API, but in future we may want to let people create via the admin interface without creating a template too.

populate a non-nullable content field with the values supplied via the template (or supplied directly if via api).


quis:
Is it worth migrating content into old broadcast messages from their respective templates? Otherwise old broadcast messages will start showing empty content in the admin app.

leohemsted:
@quis yes, definitely should, good point. however, for completeness we should do this _after_ this deploy goes out, so that we don't end up missing anything between the migration running and the API boxes deploying to start populating the field

quis (approved):
Nice and clear tests 👍🏻

PR #3086, opened 2021-01-11
Failover to second lambda on error

idavidmcdonald:
Introduce failover behaviour such that if any error is raised by the first lambda (that includes both invoke errors and function errors connected to trying to talk to the CBC), then our API will then retry by invoking the second lambda.

Logs error types returned by https://github.com/alphagov/notifications-broadcasts-infra/pull/79

Note, the second lambda doesn't exist yet. It also isn't defined yet whether the second lambda will be pointed at the second CBC or if it will still point at the first CBC (or if it needs to be able to decide on the fly).

Please review this commit by commit.

The annoyance we have is that AWS docs are a bit unclear with what values we get back from an exception in the lambda code. We've made assumptions about the interface in our tests but we will need to do some testing to be 100% sure of these.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.invoke
https://docs.aws.amazon.com/lambda/latest/dg/python-exceptions.html

### DIFF
--- a/app/clients/cbc_proxy.py
+++ b/app/clients/cbc_proxy.py
@@ -100,7 +100,7 @@ class CBCProxyClientBase(ABC):
     ):
         pass
 
-    def _invoke_lambda(self, payload):
+    def _invoke_lambda_with_failover(self, payload):
         if not self.lambda_name:
             current_app.logger.warning(
                 '{self.__class__.__name__} tried to send {payload} but cbc proxy aws env vars not set'
@@ -108,18 +108,22 @@ class CBCProxyClientBase(ABC):
             return
 
         payload_bytes = bytes(json.dumps(payload), encoding='utf8')
+        result = self._invoke_lambda(self.lambda_name, payload_bytes)
 
+        if 'FunctionError' in result:
+            raise CBCProxyException('Function exited with unhandled exception')
+
+        return result
+
+    def _invoke_lambda(self, lambda_name, payload_bytes):
         result = self._lambda_client.invoke(
-            FunctionName=self.lambda_name,
+            FunctionName=lambda_name,
             InvocationType='RequestResponse',
             Payload=payload_bytes,
         )
 
         if result['StatusCode'] > 299:
             raise CBCProxyException('Could not invoke lambda')
-
-        if 'FunctionError' in result:
-            raise CBCProxyException('Function exited with unhandled exception')
 
         return result
 
@@ -143,7 +147,7 @@ class CBCProxyCanary(CBCProxyClientBase):
         self,
         identifier,
     ):
-        self._invoke_lambda(payload={'identifier': identifier})
+        self._invoke_lambda_with_failover(payload={'identifier': identifier})
 
 
 class CBCProxyEE(CBCProxyClientBase):
@@ -167,7 +171,7 @@ class CBCProxyEE(CBCProxyClientBase):
             'message_format': 'cap'
         }
 
-        self._invoke_lambda(payload=payload)
+        self._invoke_lambda_with_failover(payload=payload)
 
     def create_and_send_broadcast(
         self, identifier, headline, description, areas, sent, expires, message_number=None
@@ -183,7 +187,7 @@ class CBCProxyEE(CBCProxyClientBase):
             'expires': expires,
             'language': self.infer_language_from(description),
         }
-        self._invoke_lambda(payload=payload)
+        self._invoke_lambda_with_failover(payload=payload)
 
     def cancel_broadcast(
         self,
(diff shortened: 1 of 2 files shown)


app/clients/cbc_proxy.py
```diff
@@ -100,27 +108,28 @@ def cancel_broadcast(
...
+
+    def _invoke_lambda(self, lambda_name, payload_bytes):
         result = self._lambda_client.invoke(
-            FunctionName=self.lambda_name,
+            FunctionName=lambda_name,
             InvocationType='RequestResponse',
             Payload=payload_bytes,
         )
 
         if result['StatusCode'] > 299:
```
idavidmcdonald, line 130: Should we take the same approach for a 299 status code and try with the second lambda and if that fails then we raise the exception?

leohemsted (commented):
I'm just wondering about how we handle retries more generally. Mainly thinking around should the failover be invoked in a separate celery job? (either using `self.retry` or creating a new task and calling `.apply_async`

Our retries shouldn't just be a "one and done" system, we should retry multiple times - and that should be handled by celery at a task level I reckon.

However, it might still be sensible to have the individual task try to send via both lambdas/AZs, each time (as you've done here). I think what you've written is pretty reasonable, but I'm just wondering if you thought about how this might tie in to a larger scale retry mechanic at the task level?

app/clients/cbc_proxy.py
leohemsted: yeah, assuming i'm reading the docs right and correlating this `StatusCode` with these potential errors: https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html#API_Invoke_Errors

if we accidentally misconfigure one of the lambdas, we should retry the other one.

app/clients/cbc_proxy.py
```diff
@@ -100,28 +108,42 @@ def cancel_broadcast(
...
-            raise CBCProxyException('Could not invoke lambda')
+            current_app.logger.info(
+                f"Error calling lambda {self.lambda_name} with status code { result['StatusCode']}, {result.get('Payload')}"
+            )
+            success = False
+
+        elif 'FunctionError' in result:
+            current_app.logger.info(
+                f"Error calling lambda {self.lambda_name} with function error { result['Payload']['errorType'] }"
```
leohemsted, line 139: do we want to log the full payload here? Or is the expectation that we should then find the full lambda logs to figure out what went wrong? (that's fine too).

app/clients/cbc_proxy.py
```diff
@@ -100,28 +108,42 @@ def cancel_broadcast(
...
 
+    def _invoke_lambda(self, lambda_name, payload_bytes):
+        current_app.logger.info(f"Calling lambda {lambda_name}")
         result = self._lambda_client.invoke(
-            FunctionName=self.lambda_name,
+            FunctionName=lambda_name,
             InvocationType='RequestResponse',
             Payload=payload_bytes,
         )
+        current_app.logger.info(f"Finished calling lambda {lambda_name}")
```
leohemsted, line 129: is it possible to log the AWS Lambda request id here? i'm not sure if it is, but i'm keen to know what gets returend from boto3 (i've been reading https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.invoke, interested to see what it looks like in practice)

app/clients/cbc_proxy.py
```diff
@@ -143,11 +168,12 @@ def send_canary(
         self,
         identifier,
     ):
-        self._invoke_lambda(payload={'identifier': identifier})
+        self._invoke_lambda_with_failover(payload={'identifier': identifier})
```
leohemsted, line 171: does this change even make sense? what does this benefit us over calling `self._invoke_lambda` like before?

app/clients/cbc_proxy.py
leohemsted: i dont mind if we put some verbose logging here, deploy that, and then pare down the logs once we know what we can actually expect to be returned)

app/clients/cbc_proxy.py
idavidmcdonald: Good point, we might as well log the full payload

app/clients/cbc_proxy.py
idavidmcdonald: Good point, will change this one to `_invoke_lambda`

app/clients/cbc_proxy.py
idavidmcdonald: I don't see anything about getting the lambda request ID, or at least not in the response from `invoke`. Let's leave the logging as this for the moment and then there is a chance we wish to remove the `finished calling` line in time.

leohemsted (approved):

PR #3087, opened 2021-01-11
add content to old broadcast messages with no content

leohemsted:
new broadcast messages will have content filled whether they have a template or not, but old ones won't so populate.

Stole the session constructor from 0044_jos_to_notification_hist.py

### DIFF
--- a/app/broadcast_message/rest.py
+++ b/app/broadcast_message/rest.py
@@ -189,19 +189,12 @@ def _create_broadcast_event(broadcast_message):
     else:
         transmitted_finishes_at = broadcast_message.finishes_at
 
-    # TODO: Remove this if statement after broadcast message content is guaranteed to always be populated.
-    if broadcast_message.content:
-        content = broadcast_message.content
-    else:
-        content = broadcast_message.template._as_utils_template_with_personalisation(
-            broadcast_message.personalisation
-        ).content_with_placeholders_filled_in
 
     event = BroadcastEvent(
         service=broadcast_message.service,
         broadcast_message=broadcast_message,
         message_type=msg_types[broadcast_message.status],
-        transmitted_content={"body": content},
+        transmitted_content={"body": broadcast_message.content},
         transmitted_areas=broadcast_message.areas,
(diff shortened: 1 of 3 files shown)


klssmith:
Can we now get rid of this logic? https://github.com/alphagov/notifications-api/blob/4980c3e0fa1a50ae917124981bf085341bee38a1/app/broadcast_message/rest.py#L192-L198

leohemsted:
ah yesss, you're right. i knew i had left a TODO somewhere

migrations/versions/0336_broadcast_msg_migrate_content.py
```diff
@@ -0,0 +1,33 @@
...
+def upgrade():
+    session = Session(bind=op.get_bind())
+
+    broadcast_messages = session.query(BroadcastMessage).filter(BroadcastMessage.content == None)
+
+    for broadcast_message in broadcast_messages:
+        broadcast_message.content = broadcast_message.template._as_utils_template_with_personalisation(
+            broadcast_message.personalisation
+        ).content_with_placeholders_filled_in
+
```
klssmith, line 28: I needed to add `session.commit()` here before the `alter_column` to get this migration to work - without that, the error was
```
sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) column "content" contains null values
```

migrations/versions/0336_broadcast_msg_migrate_content.py
```diff
@@ -0,0 +1,33 @@
+"""
+
+Revision ID: 0336_broadcast_msg_content_2
```
klssmith, line 3: The Revision ID doesn't match the filename, which I think will cause issues with the script to check if migrations need to run

leohemsted: ah good spot 👍

migrations/versions/0336_broadcast_msg_migrate_content.py
leohemsted: squashed out and force pushed

klssmith (approved):

PR #3088, opened 2021-01-12
Change the sort order for the organisation usage page

servingUpAces:
Ensure the archived services are at the bottom of the list. The organisation trial mode page already sorts the archived services to the bottom.

### DIFF
--- a/app/organisation/rest.py
+++ b/app/organisation/rest.py
@@ -135,7 +135,7 @@ def get_organisation_services_usage(organisation_id):
         return jsonify(result='error', message='No valid year provided'), 400
     services = fetch_usage_year_for_organisation(organisation_id, year)
     list_services = services.values()
-    sorted_services = sorted(list_services, key=lambda s: s['service_name'].lower())
+    sorted_services = sorted(list_services, key=lambda s: (-s['active'], s['service_name'].lower()))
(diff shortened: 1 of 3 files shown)


app/organisation/rest.py
```diff
@@ -135,7 +135,7 @@ def get_organisation_services_usage(organisation_id):
         return jsonify(result='error', message='No valid year provided'), 400
     services = fetch_usage_year_for_organisation(organisation_id, year)
     list_services = services.values()
-    sorted_services = sorted(list_services, key=lambda s: s['service_name'].lower())
+    sorted_services = sorted(list_services, key=lambda s: (-s['active'], s['service_name'].lower()))
```
CrystalPea, line 138: why is the dash there? Sorry if that's a silly question 😅

quis:
```python
-True
>>> -1

-False
>>> 0
```
😨 I have learned something new about Python today

servingUpAces: The dash reverses the order, so I get True before False

CrystalPea: sweet, thanks a lot, I had no idea this is a thing, today I've learnt :3

CrystalPea (approved):

PR #3090, opened 2021-01-13
Allow broadcast messages to be created with API key

quis:
When we have a public API there will be no human creating the broadcast message. Instead it will be created by an API integration, authenticated by a key (just like for emails or texts).

This updates the database to:
- add a new foreign key from BroadcastMessages to API keys
- make `created_by` nullable when `api_key` is non-null, and vice versa
- add a reference field which the frontend can use instead of template name, for broadcasts without templates

***

This commit also fixes a previous migration to remove use of a model class.

We shouldn’t import models into migrations because if the model changes later down the line then the migration can’t be re-run at a later date (for example to rebuild a database from scratch).

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2245,13 +2245,20 @@ class BroadcastMessage(db.Model):
     cancelled_at = db.Column(db.DateTime, nullable=True)
     updated_at = db.Column(db.DateTime, nullable=True, onupdate=datetime.datetime.utcnow)
 
-    created_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey('users.id'), nullable=False)
+    created_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey('users.id'), nullable=True)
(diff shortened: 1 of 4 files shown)


app/models.py
```diff
@@ -2245,14 +2245,21 @@ class BroadcastMessage(db.Model):
     cancelled_at = db.Column(db.DateTime, nullable=True)
     updated_at = db.Column(db.DateTime, nullable=True, onupdate=datetime.datetime.utcnow)
 
-    created_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey('users.id'), nullable=False)
+    created_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey('users.id'), nullable=True)
```
leohemsted, line 2248: i cant leave a comment on the line directly but on line 2294 we should now check if created_by_id is none before calling `str` on it

quis: I noticed this for the other optional joins to the `users` table too, have fixed in 78e8785

leohemsted (approved):

PR #3093, opened 2021-01-13
Broadcast tasks onto worker

idavidmcdonald:
https://www.pivotaltracker.com/story/show/176459474

Review commit by commit

Will require a manual `cf create-app notify-delivery-worker-broadcasts` in each CF space before this can be deployed. 

Will then need to manually scale the app up to 2 instances for high availability. Either that or we set up autoscaler config such that it's min and max number of apps is 2 instances (not particularly a pattern we do yet but maybe one to introduce). Introducing this pattern of adding config to the autoscaler will also give us the benefit of tracking the number of instances in the autoscaler grafana dashboard and how many tasks are being added to queues/processed by the broadcast worker etc. Would be interested to get your thoughts on this please.

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -64,6 +64,7 @@
   'notify-delivery-worker-letters': {'memory': '2G'},
   'notify-delivery-worker-retry-tasks': {},
   'notify-delivery-worker-internal': {},
+  'notify-delivery-worker-broadcasts': {},
   'notify-delivery-worker-receipts': {},
   'notify-delivery-worker-service-callbacks': {'disk_quota': '2G'},
   'notify-delivery-worker-save-api-notifications': {'disk_quota': '2G'},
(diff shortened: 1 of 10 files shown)


manifest.yml.j2
```diff
@@ -64,6 +64,7 @@
   'notify-delivery-worker-letters': {'memory': '2G'},
   'notify-delivery-worker-retry-tasks': {},
   'notify-delivery-worker-internal': {},
+  'notify-delivery-worker-broadcasts': {},
```
leohemsted, line 67: can put `'instances': {'preview': 2, 'staging': 2, 'production': 2}` in here if we want to define instances without adding to autoscaler.

but do we need to configure autoscaler anyway to get metrics though?

idavidmcdonald: Will configure it in the autoscaler to give us metrics so won't specify the instances in here. Given that, I think this is ready for a final review please Leo

leohemsted (approved):

PR #3094, opened 2021-01-14
Three proxy

idavidmcdonald:
Commit one creates the CBC proxy client for Three
Commit two turns it on.

@rjbaker could you confirm whether we are good to release this with it turned on or if we need to hold off on that until later. Note, I'm expecting it will be fine to release it because if Three is not ready then our calls to the lambda will fail and we will log a failure but nothing much else will happen

Decided not to abstract the EE and Three proxies into a `CBCProxyOne2Many` subclass just yet. Will probably be a sensible move in the future but going to hold off until we add O2 or we get Three working as that may provide different information about any differences between the proxies

Note, our tests are made a bit nasty by the fact that we have the `CBCProxyEE` class uses a proxy that is not called `ee-proxy-1` and instead uses `bt-ee-proxy-1`. I would suggest we change the proxy name to `ee-proxy-1` for consistency. Any objections to me doing that?

### DIFF
--- a/app/clients/cbc_proxy.py
+++ b/app/clients/cbc_proxy.py
@@ -45,6 +45,7 @@ class CBCProxyClient:
         proxy_classes = {
             'canary': CBCProxyCanary,
             BroadcastProvider.EE: CBCProxyEE,
+            BroadcastProvider.THREE: CBCProxyThree,
             BroadcastProvider.VODAFONE: CBCProxyVodafone,
         }
         return proxy_classes[provider](self._lambda_client)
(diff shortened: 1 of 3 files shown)


rjbaker:
> Commit one creates the CBC proxy client for Three
> Commit two turns it on.
> 
> @rjbaker could you confirm whether we are good to release this with it turned on or if we need to hold off on that until later. Note, I'm expecting it will be fine to release it because if Three is not ready then our calls to the lambda will fail and we will log a failure but nothing much else will happen
> 
> Decided not to abstract the EE and Three proxies into a `CBCProxyOne2Many` subclass just yet. Will probably be a sensible move in the future but going to hold off until we add O2 or we get Three working as that may provide different information about any differences between the proxies
> 
> Note, our tests are made a bit nasty by the fact that we have the `CBCProxyEE` class uses a proxy that is not called `ee-proxy-1` and instead uses `bt-ee-proxy-1`. I would suggest we change the proxy name to `ee-proxy-1` for consistency. Any objections to me doing that?

Thanks @idavidmcdonald. The lambda isn't deployed yet, so the lambda invocation will fail as you mention - which is fine. Also happy to rename the proxy function to `ee-proxy-1`.

rjbaker (approved):

PR #3095, opened 2021-01-15
Let the admin app create broadcasts without templates

quis:
This pull request updates the JSON schemas and REST endpoint to allow creation of a broadcast with either:
- a `template_id` or
- `content` and a `reference` 

`reference` is so we have something to show in the admin app in lieu of template name when a template isn’t used.

The JSONSchema errors are a bit gnarly, but I think this is OK because this is the internal API.

### DIFF
--- a/tests/app/broadcast_message/test_rest.py
+++ b/tests/app/broadcast_message/test_rest.py
@@ -143,9 +143,9 @@ def test_create_broadcast_message(admin_request, sample_broadcast_service):
     (
         {},
         [
-            {'error': 'ValidationError', 'message': 'template_id is a required property'},
             {'error': 'ValidationError', 'message': 'service_id is a required property'},
-            {'error': 'ValidationError', 'message': 'created_by is a required property'}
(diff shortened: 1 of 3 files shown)


tests/app/broadcast_message/test_rest.py
```diff
@@ -178,23 +178,106 @@ def test_create_broadcast_message_400s_if_json_schema_fails_validation(
...
+            'created_by': str(sample_broadcast_service.created_by_id),
+        },
+        service_id=sample_broadcast_service.id,
+        _expected_status=400
+    )
+    assert len(response['errors']) == 1
+    assert response['errors'][0]['error'] == 'ValidationError'
+    assert response['errors'][0]['message'].endswith(
+        'is not valid under any of the given schemas'
+    )
```
CrystalPea, line 281: Should we also have a test for when neither template_id nor content is provided?

CrystalPea (approved):
Good work, we could add one more test maybe, as per my comment.

quis:
@CrystalPea good catch, I’ve added that extra test case in 94aea8a

PR #3096, opened 2021-01-15
Add notes column to services table

CrystalPea:
For us admins to be able to add notes about any special requests or irregularities pertaining to services.


idavidmcdonald (approved):
I think this is good.

Note a few things that I looked into whilst reviewing this

- I gave some thought whether this should be excluded in either the `ServiceSchema` or `DetailedServiceSchema` in https://github.com/alphagov/notifications-api/blob/master/app/schemas.py. It felt reasonable to include it in both. My rough thought was that the notes field could be very large and therefore increase the amount of data but I think that is pretty unlikely and therefore I wasn't too concerned about adding it to both.

- I wondered if we should add some dummy data to our tests so perhaps the `sample_service` has some data in it's `notes` field but after looking at how we set up the `sample_service`, most of the fields are empty anyway and I think it's probably fine.

PR #3101, opened 2021-01-19
retry sending broadcasts

leohemsted:
Retry tasks if they fail to send a broadcast event. Note that each task tries the regular proxy and the failover proxy for that provider. This runs a bit differently than our other retries:

Retry with exponential backoff. Our other tasks retry with a fixed delay of 5 minutes between tries. If we can't send a broadcast, we want to try immediately. So instead, implement an exponential backoff (1, 2, 4, 8, ... seconds delay). We can't delay for longer than 310 seconds due to visibility timeout settings in SQS, so cap the delay at that amount.

Normally we give up retrying after a set amount of retries (often 4 hours). As broadcast content is much more important than normal notifications, we don't ever want to give up on sending them to phones...

...UNLESS WE DO!

Sometimes we do want to give up sending a broadcast though! Broadcasts have an expiry time, when they stop showing up on peoples devices, so if that has passed then we don't need to send the broadcast out.

Broadcast events can also be superceded by updates or cancels. Check that the event is the most recent event for that broadcast message, if not, give up, as we don't want to accidentally send out two conflicting events for the same message.

### DIFF
--- a/app/broadcast_message/rest.py
+++ b/app/broadcast_message/rest.py
@@ -56,7 +56,7 @@ def _update_broadcast_message(broadcast_message, new_status, updating_user):
                 f'User {updating_user.id} cannot approve their own broadcast_message {broadcast_message.id}',
                 status_code=400
             )
-        elif not broadcast_message.areas:
+        elif len(broadcast_message.areas['simple_polygons']) == 0:
             raise InvalidRequest(
                 f'broadcast_message {broadcast_message.id} has no selected areas and so cannot be broadcasted.',
                 status_code=400
--- a/app/celery/broadcast_message_tasks.py
+++ b/app/celery/broadcast_message_tasks.py
@@ -32,25 +32,61 @@ def get_retry_delay(retry_count):
     return min(delay, 300)
 
 
-def check_provider_message_should_retry(broadcast_provider_message):
-    this_event = broadcast_provider_message.broadcast_event
+def check_provider_message_should_send(broadcast_event, provider):
+    """
+    If any previous event hasn't sent yet for that provider, then we shouldn't send the current event. Instead, fail and
+    raise a P1 - so that a notify team member can assess the state of the previous messages, and if necessary, can
+    replay the `send_broadcast_provider_message` task if the previous message has now been sent.
 
-    if this_event.transmitted_finishes_at < datetime.utcnow():
-        print(this_event.transmitted_finishes_at, datetime.utcnow(),)
-        raise MaxRetriesExceededError(
-            f'Given up sending broadcast_event {this_event.id} ' +
-            f'to provider {broadcast_provider_message.provider}: ' +
-            f'The expiry time of {this_event.transmitted_finishes_at} has already passed'
+    Note: This is called before the new broadcast_provider_message is created.
+
+    # Help, I've come across this code following a pagerduty alert, what should I do?
+
+    1. Find the failing broadcast_provider_message associated with the previous event that caused this to trip.
+    2. If that provider message is still failing to send, fix the issue causing that. The task to send that previous
+       message might still be retrying in the background - look for logs related to that task.
+    3. If that provider message has sent succesfully, you might need to send this task off depending on context. This
+       might not always be true though, for example, it may not be necessary to send a cancel if the original alert has
+       already expired.
+    4. If you need to re-send this task off again, you'll need to run the following command on paas:
+       `send_broadcast_provider_message.apply_async(args=(broadcast_event_id, provider), queue=QueueNames.BROADCASTS)`
+    """
+    if broadcast_event.transmitted_finishes_at < datetime.utcnow():
+        # TODO: This should be a different kind of exception to distinguish "We should know something went wrong, but
+        # no immediate action" from "We need to fix this immediately"
+        raise CBCProxyFatalException(
+            f'Cannot send broadcast_event {broadcast_event.id} ' +
+            f'to provider {provider}: ' +
+            f'The expiry time of {broadcast_event.transmitted_finishes_at} has already passed'
         )
 
-    newest_event = max(this_event.broadcast_message.events, key=lambda x: x.sent_at)
+    # get events sorted from earliest to latest
+    events = sorted(broadcast_event.broadcast_message.events, key=lambda x: x.sent_at)
 
-    if this_event != newest_event:
-        raise MaxRetriesExceededError(
-            f'Given up sending broadcast_event {this_event.id} ' +
-            f'to provider {broadcast_provider_message.provider}: ' +
-            f'This event has been superceeded by {newest_event.message_type} broadcast_event {newest_event.id}'
-        )
+    for prev_event in events:
+        if prev_event.id != broadcast_event.id and prev_event.sent_at < broadcast_event.sent_at:
+            # get the record from when that event was sent to the same provider
+            prev_provider_message = prev_event.get_provider_message(provider)
+
+            # the previous message hasn't even got round to running `send_broadcast_provider_message` yet.
+            if not prev_provider_message:
+                raise CBCProxyFatalException(
+                    f'Cannot send {broadcast_event.id}. Previous event {prev_event.id} ' +
+                    f'(type {prev_event.message_type}) has no provider_message for provider {provider} yet.\n' +
+                    f'You must ensure that the other event sends succesfully, then manually kick off this event ' +
+                    f'again by re-running send_broadcast_provider_message for this event and provider.'
+                )
+
+            # if there's a previous message that has started but not finished sending (whether it fatally errored or is
+            # currently retrying)
+            if prev_provider_message.status != BroadcastProviderMessageStatus.ACK:
+                raise CBCProxyFatalException(
+                    f'Cannot send {broadcast_event.id}. Previous event {prev_event.id} ' +
+                    f'(type {prev_event.message_type}) has not finished sending to provider {provider} yet.\n' +
+                    f'It is currently in status "{prev_provider_message.status}".\n' +
+                    f'You must ensure that the other event sends succesfully, then manually kick off this event ' +
+                    f'again by re-running send_broadcast_provider_message for this event and provider.'
+                )
 
 
 @notify_celery.task(name="send-broadcast-event")
@@ -74,7 +110,9 @@ def send_broadcast_event(broadcast_event_id):
 def send_broadcast_provider_message(self, broadcast_event_id, provider):
     broadcast_event = dao_get_broadcast_event_by_id(broadcast_event_id)
 
-    # the broadcast_provider_message will already exist if we retried previously
+    check_provider_message_should_send(broadcast_event, provider)
+
+    # the broadcast_provider_message may already exist if we retried previously
     broadcast_provider_message = broadcast_event.get_provider_message(provider)
     if broadcast_provider_message is None:
         broadcast_provider_message = create_broadcast_provider_message(broadcast_event, provider)
@@ -138,19 +176,15 @@ def send_broadcast_provider_message(self, broadcast_event_id, provider):
                 sent=broadcast_event.sent_at_as_cap_datetime_string,
             )
     except CBCProxyRetryableException as exc:
-        # this will raise MaxRetriesExceededError if we no longer want to retry
-        # (because the message has expired)
-        check_provider_message_should_retry(broadcast_provider_message)
-
-        # TODO: Decide whether to set to TECHNICAL_FAILURE or ERROR based on response codes from cbc proxy
-        update_broadcast_provider_message_status(
-            broadcast_provider_message,
-            status=BroadcastProviderMessageStatus.TECHNICAL_FAILURE
+        delay = get_retry_delay(self.request.retries)
+        current_app.logger.exception(
+            f'Retrying send_broadcast_provider_message for broadcast_event {broadcast_event_id} and ' +
+            f'provider {provider} in {delay} seconds'
         )
 
         self.retry(
             exc=exc,
-            countdown=get_retry_delay(self.request.retries),
+            countdown=delay,
             queue=QueueNames.BROADCASTS,
         )
 
--- a/app/celery/broadcast_message_tasks.py
+++ b/app/celery/broadcast_message_tasks.py
@@ -9,8 +9,12 @@ from celery.exceptions import MaxRetriesExceededError
 from app import cbc_proxy_client, db, notify_celery
 from app.clients.cbc_proxy import CBCProxyFatalException, CBCProxyRetryableException
 from app.config import QueueNames
-from app.models import BroadcastEventMessageType, BroadcastProvider
-from app.dao.broadcast_message_dao import dao_get_broadcast_event_by_id, create_broadcast_provider_message
+from app.models import BroadcastEventMessageType, BroadcastProvider, BroadcastProviderMessageStatus
+from app.dao.broadcast_message_dao import (
+    dao_get_broadcast_event_by_id,
+    create_broadcast_provider_message,
+    update_broadcast_provider_message_status
+)
 
 from app.utils import format_sequential_number
 
@@ -138,11 +142,19 @@ def send_broadcast_provider_message(self, broadcast_event_id, provider):
         # (because the message has expired)
         check_provider_message_should_retry(broadcast_provider_message)
 
+        # TODO: Decide whether to set to TECHNICAL_FAILURE or ERROR based on response codes from cbc proxy
+        update_broadcast_provider_message_status(
+            broadcast_provider_message,
+            status=BroadcastProviderMessageStatus.TECHNICAL_FAILURE
+        )
+
         self.retry(
             exc=exc,
             countdown=get_retry_delay(self.request.retries),
             queue=QueueNames.BROADCASTS,
         )
+
+    update_broadcast_provider_message_status(broadcast_provider_message, status=BroadcastProviderMessageStatus.ACK)
 
 
 @notify_celery.task(name='trigger-link-test')
--- a/app/celery/broadcast_message_tasks.py
+++ b/app/celery/broadcast_message_tasks.py
@@ -70,7 +70,11 @@ def send_broadcast_event(broadcast_event_id):
 def send_broadcast_provider_message(self, broadcast_event_id, provider):
     broadcast_event = dao_get_broadcast_event_by_id(broadcast_event_id)
 
-    broadcast_provider_message = create_broadcast_provider_message(broadcast_event, provider)
+    # the broadcast_provider_message will already exist if we retried previously
+    broadcast_provider_message = broadcast_event.get_provider_message(provider)
+    if broadcast_provider_message is None:
+        broadcast_provider_message = create_broadcast_provider_message(broadcast_event, provider)
+
     formatted_message_number = None
     if provider == BroadcastProvider.VODAFONE:
         formatted_message_number = format_sequential_number(broadcast_provider_message.message_number)
--- a/app/celery/broadcast_message_tasks.py
+++ b/app/celery/broadcast_message_tasks.py
@@ -1,15 +1,52 @@
 import uuid
+from datetime import datetime
 
 from flask import current_app
 from notifications_utils.statsd_decorators import statsd
 from sqlalchemy.schema import Sequence
+from celery.exceptions import MaxRetriesExceededError
 
 from app import cbc_proxy_client, db, notify_celery
+from app.clients.cbc_proxy import CBCProxyFatalException, CBCProxyRetryableException
 from app.config import QueueNames
 from app.models import BroadcastEventMessageType, BroadcastProvider
 from app.dao.broadcast_message_dao import dao_get_broadcast_event_by_id, create_broadcast_provider_message
 
 from app.utils import format_sequential_number
+
+
+def get_retry_delay(retry_count):
+    """
+    Given a count of retries so far, return a delay for the next one.
+    `retry_count` should be 0 the first time a task fails.
+    """
+    # TODO: replace with celery's built in exponential backoff
+
+    # 2 to the power of x. 1, 2, 4, 8, 16, 32, ...
+    delay = 2**retry_count
+    # never wait longer than 5 minutes
+    return min(delay, 300)
+
+
+def check_provider_message_should_retry(broadcast_provider_message):
+    this_event = broadcast_provider_message.broadcast_event
+
+    if this_event.transmitted_finishes_at < datetime.utcnow():
+        print(this_event.transmitted_finishes_at, datetime.utcnow(),)
+        raise MaxRetriesExceededError(
+            f'Given up sending broadcast_event {this_event.id} ' +
+            f'to provider {broadcast_provider_message.provider}: ' +
+            f'The expiry time of {this_event.transmitted_finishes_at} has already passed'
+        )
+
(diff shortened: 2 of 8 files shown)


tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -331,6 +374,11 @@ def test_send_broadcast_provider_message_errors(mocker, sample_service):
         sent=event.sent_at_as_cap_datetime_string,
         expires=event.transmitted_finishes_at_as_cap_datetime_string,
     )
+    mock_retry.assert_called_once_with(
+        countdown=1,
```
idavidmcdonald, line 378: Optional: It took me a second to work out that `countdown=1` is because this is the first retry so it's 2^0=1 and so this is the assertion that is checking it's exponential (because otherwise it would be countdown 0 or something else). Your other option could be to mock `get_retry_duration` and assert the countdown is the return value of the mock

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -361,3 +409,90 @@ def test_trigger_link_tests_invokes_cbc_proxy_client(
...
+        transmitted_starts_at=datetime(2021, 1, 1, 0, 0),
+        transmitted_finishes_at=datetime(2021, 1, 2, 0, 0),
+    )
+
+    provider_message = create_broadcast_provider_message(current_event, 'ee')
+
+    # even though the task is going on until midnight tomorrow, we shouldn't send the update now, because the cancel
+    # message will be in the pipeline somewhere.
+    with pytest.raises(MaxRetriesExceededError) as exc:
+        check_provider_message_should_retry(provider_message)
```
idavidmcdonald, line 496: This test is only really testing that if you have a cancel event which has already started, then you should not retry the original `update`. My comments above about retry logic to do with cancelling/update might change the code but if it weren't to, then I would also like to see this test be duplicated so we see that the same behaviour happens for an update event which has already started. It's a bit confusing having both `future_event` and `futurest_event` in this test, I don't see too much benefit from having both of those fixtures here in this test

tests/app/db.py
```diff
@@ -1075,7 +1075,7 @@ def create_broadcast_event(
         transmitted_areas=transmitted_areas or broadcast_message.areas,
         transmitted_sender=transmitted_sender or 'www.notifications.service.gov.uk',
         transmitted_starts_at=transmitted_starts_at,
-        transmitted_finishes_at=transmitted_finishes_at or datetime.utcnow(),
+        transmitted_finishes_at=transmitted_finishes_at or datetime.utcnow() + timedelta(hours=1),
```
idavidmcdonald, line 1078: Optional: This is fine. Another option could be `transmitted_finishes_at=transmitted_finishes_at or transmitted_starts + timedelta(hours=24)` as that is likely what we will be having as real data

app/celery/broadcast_message_tasks.py
```diff
@@ -1,17 +1,54 @@
...
 from app import cbc_proxy_client, db, notify_celery
+from app.clients.cbc_proxy import CBCProxyException
 from app.config import QueueNames
 from app.models import BroadcastEventMessageType, BroadcastProvider
 from app.dao.broadcast_message_dao import dao_get_broadcast_event_by_id, create_broadcast_provider_message
 
 from app.utils import format_sequential_number
 
 
+def get_retry_duration(retry_count):
```
idavidmcdonald, line 18: Optional: Rename `get_retry_delay` as `delay` is the wording used by celery, rather than `duration` for this - https://docs.celeryproject.org/en/3.1/userguide/tasks.html#using-a-custom-retry-delay

idavidmcdonald (changes requested):

app/celery/broadcast_message_tasks.py
```diff
@@ -1,17 +1,54 @@
...
+        print(this_event.transmitted_finishes_at, datetime.utcnow(),)
+        raise MaxRetriesExceededError(
+            f'Given up sending broadcast_event {this_event.id} ' +
+            f'to provider {broadcast_provider_message.provider}: ' +
+            f'The expiry time of {this_event.transmitted_finishes_at} has already passed'
+        )
+
+    newest_event = max(this_event.broadcast_message.events, key=lambda x: x.sent_at)
+
+    if this_event != newest_event:
```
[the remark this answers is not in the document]
leohemsted: # First paragraph (what happens to future events if this one didnt go out)

Good point. Considering an Alert/Update/Cancel flow. As an extra sticking point, all these are _per provider_, so we may need to do one thing for one provider and a different thing for other providers.

### Alert fails

Update: Should now technically change to an alert, as it's not updating anything.
Cancel: If there's an update, should just use that in the reference that. 

### Update fails
Cancel: should go out, but only affect the original alert.

### Cancel fails
Nothing to do. Keep trying until the previous alerts/updates time out.


# Second paragraph

Your case isn't an issue, however it's made me realise there's a separate issue.

This described case wouldn't be an issue a cancel message is _not_ valid for an extra 24 hours from cancellation time. In fact, its time is set to the time at which the cancelled button is clicked, (https://github.com/alphagov/notifications-api/blob/master/app/broadcast_message/rest.py#L197-L198). So we won't keep cancelling for 24 hrs.

However, the issue is: 

If we send a cancel, and it succeeds, we never check what the transmitted_finishes_at time is, we just send it. But if sending it fails (and failover fails too) we'll check whether we should retry it. And at that point it'll say "I'm now a few milliseconds past the `transmitted_finishes_at` time so I shouldn't retry this" and we'll never issue the cancel. I'm not entirely sure on the impacts of setting the transmitted_finishes_at (`<expires>` in the xml) to be the current timestamp, but I suspect that was just me being overcautious. 

My gut feel is that it's easiest to reason about if the `transmitted_finishes_at` never changes between alert/update/cancel, and is always set to `broadcast_message.finishes_at` (which is currently set to 24 hrs from approval time https://github.com/alphagov/notifications-admin/blob/master/app/models/broadcast_message.py#L203-L205). That way, in your scenario, we'd keep trying to cancel from 11pm until 1am the next day.

I don't want to make that change without double checking with pete/richard though.

app/celery/broadcast_message_tasks.py
```diff
@@ -50,33 +88,44 @@ def send_broadcast_provider_message(broadcast_event_id, provider):
...
+                expires=broadcast_event.transmitted_finishes_at_as_cap_datetime_string,
+            )
+        elif broadcast_event.message_type == BroadcastEventMessageType.CANCEL:
+            cbc_proxy_provider_client.cancel_broadcast(
+                identifier=str(broadcast_provider_message.id),
+                message_number=formatted_message_number,
+                previous_provider_messages=broadcast_event.get_earlier_provider_messages(provider),
+                sent=broadcast_event.sent_at_as_cap_datetime_string,
+            )
+    except CBCProxyException as exc:
```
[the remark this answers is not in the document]
leohemsted: So really we need a bit wider exception base - I was thinking about if we'd end up with something like `CBCProxyFatalException` and `CBCProxyRetryableException` (or similar). And we'd only catch the retryable exceptions. Then we can P1 fi we have a fatal exception, and we'd know immediately.

I imagine these fatal exceptions are things where we don't expect a retry to solve. So these could be failures due to a misconfiguration in our lambda, an expired cert, invalid content within the broadcast, etc - something that would require some kind of deploy to fix. I think I rather that the task fails without retrying, raises a pagerduty immediately, and then we can get out of bed, fix the issue, and manually retry the task when we know it'll work.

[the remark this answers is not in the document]
leohemsted: however, note that `max_retries=None` won't affect anything around which exception types cause retries. the only thing that causes celery to retry is us calling `self.retry` manually. Any type of uncaught exception will cause the celery task to fail without retrying.

in celery 4+, we can let celery determine all this stuff by giving it a list of exception types to retry for (https://docs.celeryproject.org/en/stable/userguide/tasks.html#automatic-retry-for-known-exceptions).

app/celery/broadcast_message_tasks.py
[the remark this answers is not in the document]
leohemsted: Changing so we don't send anything out if a previous message hasn't gone out - instead we just P1.

So now the situation is:


## first paragraph

if a prev event is hanging, and a new event is issued: we just fall over for that provider and wake someone in notify up. That person can make a call about how to proceed.

## second paragraph

I've solved this by setting the cancelled at time to retain the original finishing time, and not the "cancelled_at" time. this feels nicer to me, the cancellation statement itself should handle invalidating the original message, and "what time does a cancel message expire" makes no sense.

tests/app/celery/test_broadcast_message_tasks.py
leohemsted: properly patched the retry count and parametrized

tests/app/db.py
leohemsted: its complicated, we dont currently set transmitted_starts as that's for if we want the CBC to delay the start of the msg which we currently don't do. but i bumped to `now + 24 hrs` to give it a more realistic buffer if you didn't specify it manually

app/broadcast_message/rest.py
```diff
@@ -50,13 +50,14 @@ def _update_broadcast_message(broadcast_message, new_status, updating_user):
         )
 
     if new_status == BroadcastStatusType.BROADCASTING:
+        print(broadcast_message.areas)
```
idavidmcdonald, line 53: Leftover print statement

app/celery/broadcast_message_tasks.py
```diff
@@ -1,17 +1,93 @@
...
+       message might still be retrying in the background - look for logs related to that task.
+    3. If that provider message has sent succesfully, you might need to send this task off depending on context. This
+       might not always be true though, for example, it may not be necessary to send a cancel if the original alert has
+       already expired.
+    4. If you need to re-send this task off again, you'll need to run the following command on paas:
+       `send_broadcast_provider_message.apply_async(args=(broadcast_event_id, provider), queue=QueueNames.BROADCASTS)`
+    """
+    if broadcast_event.transmitted_finishes_at < datetime.utcnow():
+        # TODO: This should be a different kind of exception to distinguish "We should know something went wrong, but
+        # no immediate action" from "We need to fix this immediately"
```
idavidmcdonald, line 56: Does this comment still stand? Is the CBCProxyFatalException not a way to show that this is something needs fixing immediately?

app/celery/broadcast_message_tasks.py
```diff
@@ -1,17 +1,93 @@
...
+        raise CBCProxyFatalException(
+            f'Cannot send broadcast_event {broadcast_event.id} ' +
+            f'to provider {provider}: ' +
+            f'The expiry time of {broadcast_event.transmitted_finishes_at} has already passed'
+        )
+
+    events = sorted(broadcast_event.broadcast_message.events, key=lambda x: x.sent_at)
```
idavidmcdonald, line 63: Optional: I had to google that this returns a list which is sorted in ascending order, so this is starting at the very first event and working forwards in time. If you were able to communicate that in a code comment or by a variable name might be useful

app/celery/broadcast_message_tasks.py
```diff
@@ -1,17 +1,93 @@
...
+    for prev_event in events:
+        if prev_event.id != broadcast_event.id and prev_event.sent_at < broadcast_event.sent_at:
+            # get the record from when that event was sent to the same provider
+            prev_provider_message = prev_event.get_provider_message(provider)
+
+            # the previous message hasn't even got round to running `send_broadcast_provider_message` yet.
+            if not prev_provider_message:
+                raise CBCProxyFatalException(
+                    f'Cannot send {broadcast_event.id}. Previous event {prev_event.id} ' +
+                    f'(type {prev_event.message_type}) has no provider_message for provider {   provider} yet.\n' +
```
idavidmcdonald, line 74: Some weird spacing around `{     provider }` here

app/celery/broadcast_message_tasks.py
```diff
@@ -27,12 +103,19 @@ def send_broadcast_event(broadcast_event_id):
...
+@notify_celery.task(bind=True, name="send-broadcast-provider-message", max_retries=None)
 @statsd(namespace="tasks")
-def send_broadcast_provider_message(broadcast_event_id, provider):
+def send_broadcast_provider_message(self, broadcast_event_id, provider):
     broadcast_event = dao_get_broadcast_event_by_id(broadcast_event_id)
 
-    broadcast_provider_message = create_broadcast_provider_message(broadcast_event, provider)
+    check_provider_message_should_send(broadcast_event, provider)
+
+    # the broadcast_provider_message will already exist if we retried previously
```
idavidmcdonald, line 114: Optional: Change word `will` to `may` as I'm not sure it's guaranteed it will exist?

app/celery/broadcast_message_tasks.py
```diff
@@ -50,35 +133,56 @@ def send_broadcast_provider_message(broadcast_event_id, provider):
...
-            description=broadcast_event.transmitted_content['body'],
-            areas=areas,
-            previous_provider_messages=broadcast_event.get_earlier_provider_messages(provider),
-            sent=broadcast_event.sent_at_as_cap_datetime_string,
-            expires=broadcast_event.transmitted_finishes_at_as_cap_datetime_string,
+
+        # TODO: Decide whether to set to TECHNICAL_FAILURE or ERROR based on response codes from cbc proxy
+        update_broadcast_provider_message_status(
+            broadcast_provider_message,
+            status=BroadcastProviderMessageStatus.TECHNICAL_FAILURE
```
idavidmcdonald, line 175: If for some reason it failed to call out to the CBC with the alert and raises a retryable exception, why would we update the broadcast provider message status? Surely it's still in a status of 'sending' until we decide to give up and no longer retry (at which point it gets marked as a failure state of some sort)?

app/celery/broadcast_message_tasks.py
```diff
@@ -50,35 +133,56 @@ def send_broadcast_provider_message(broadcast_event_id, provider):
...
+                expires=broadcast_event.transmitted_finishes_at_as_cap_datetime_string,
+            )
+        elif broadcast_event.message_type == BroadcastEventMessageType.CANCEL:
+            cbc_proxy_provider_client.cancel_broadcast(
+                identifier=str(broadcast_provider_message.id),
+                message_number=formatted_message_number,
+                previous_provider_messages=broadcast_event.get_earlier_provider_messages(provider),
+                sent=broadcast_event.sent_at_as_cap_datetime_string,
+            )
+    except CBCProxyRetryableException as exc:
```
idavidmcdonald, line 165: I'm still wondering if we catch any exception that isn't a `CBCProxyFatalException` if we should retry? Just wondering if there any other types of flakes that we might get that wouldn't trigger a `CBCProxyRetryableException` that I'd rather retry unless we are absolutely confident they shouldn't. Like maybe some other random network flake when connecting to AWS or something

app/celery/broadcast_message_tasks.py
```diff
@@ -50,35 +133,56 @@ def send_broadcast_provider_message(broadcast_event_id, provider):
...
-            previous_provider_messages=broadcast_event.get_earlier_provider_messages(provider),
-            sent=broadcast_event.sent_at_as_cap_datetime_string,
+
+        self.retry(
+            exc=exc,
+            countdown=delay,
+            queue=QueueNames.BROADCASTS,
         )
 
+    update_broadcast_provider_message_status(broadcast_provider_message, status=BroadcastProviderMessageStatus.ACK)
```
idavidmcdonald, line 184: Ah so we were missing this before in our code eh?

idavidmcdonald (commented):
Looking very good. Very readable code, some of your best.

A few minor optional comments and then one or two more interesting ones, in particular the one about the use of provider message statuses

app/celery/broadcast_message_tasks.py
leohemsted: yeah, this comment still stands, but i haven't perhaps explained myself quite right. This particular case doesn't require immediate action, as the alert has expired naturally so whatever message we were trying to send out no longer needs to be sent out. We should investigate why it didn't send the next day, but we don't necessarily need to get out of bed. 

Compare this case to, eg, an AttributeError - where regardless of what the retry loop does, we need to urgently look into it because a message might need to go out right now.

Perhaps you could make the case that we might still want to get out of bed to understand why a message didn't go out in time, so it doesn't matter what exception type we throw here as long as it's logged. 

But I think it's a bit premature to look into this too much now - I think once some of those code is a bit more stable and we understand a bit more especially around what kind of errors we might get back from proxy I'd like to look into how we log errors and what triggers different pagerduty settings.

app/celery/broadcast_message_tasks.py
leohemsted: yeah possibly. I kinda wanted to differentiate "Currently sending right now" from "Sitting in a queue waiting to go again".

idavidmcdonald: I'd be very keen to leave it in a sending type state rather then technical failure as this is a completely different mental model to how all our other notification types work. Stuff stays in sending until we know it's been received correctly or decided that it definitely can't be received.

I think we will be able to tell something is sat in a queue because it's still in sending after more than a minute or two and we've probably had a p1 go off about it or something?

leohemsted: yeah ive thought about it and i agree with you - if we do this we can also check for tech failure in the check at the top of the code, which means we've got a killswitch if we need to get something out of the retry loop for whatever reason - we can set its status to tech failure manually

app/celery/broadcast_message_tasks.py
idavidmcdonald: Gotcha!

That makes sense. Your code comment could be a little clearer to give that context you mentioned but its not the end of the world.

However, I also can't see us hopefully ever having to throw an exception here. On the assumption that at least for a while, alerts won't expire for 24 hours, it would mean that we would get an exception thrown here if we tried to send an alert, it was failing for a while and we couldn't fix it within 24 hours - I hope that never happens.

More likely would be a coding bug that for some reason we were trying to send an old event that had already expired when we shouldn't have.

leohemsted: yeah, either that or someone trying to send a cancel at exactly the time it expires anyway cos they expected to have a manual step

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -471,3 +581,137 @@ def test_trigger_link_tests_invokes_cbc_proxy_client(
...
+@freeze_time('2021-01-01 12:00')
+def test_check_provider_message_should_send_raises_if_older_event_hasnt_started_sending_yet(sample_template):
+    broadcast_message = create_broadcast_message(sample_template)
+    # event approved at midnight
+    past_succesful_event = create_broadcast_event(
+        broadcast_message,
+        message_type='alert',
+        sent_at=datetime(2021, 1, 1, 0, 0),
+    )
+    # event updated at 5am (this is the event we're currently trying to send)
```
idavidmcdonald, line 664: Suggest changing this to `event updated at 5am (this event is still trying to be sent)` to not confuse `currently` with `current_event`

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -471,3 +581,137 @@ def test_trigger_link_tests_invokes_cbc_proxy_client(
...
+    current_event = create_broadcast_event(
+        broadcast_message,
+        message_type='update',
+        sent_at=datetime(2021, 1, 1, 7, 0),
+    )
+
+    # no provider message for past_still_sending_event
+    create_broadcast_provider_message(past_succesful_event, provider='ee', status=BroadcastProviderMessageStatus.ACK)
+
+    # even though the task is going on until midnight tomorrow, we shouldn't send the update now, because the cancel
```
idavidmcdonald, line 680: `cancel message` comment seems irrelevant now? There are no cancels, just updates. Would suggest changing this comment to:

this task shouldn't send because the `past_still_sending_event` hasn't started sending yet for this provider

idavidmcdonald (changes requested):
So I think there are a few small changes needed to the tests to fix inaccurate comments and to reflect the change we made that when a broadcast provider event fails to invoke the CBC we will leave it's status in 'sending'.

Now on top of this, I've had fun thinking about lots of complex cases that I think we should have the following behaviour for with a test to back it up. I've made a table with each of them and the current state of play.

| Scenario      | Result | Currently implemented | Currently tested |
| ----------- | ----------- | ----------- | ----------- |
| There is not yet a broadcast provider event in the DB for this broadcast event and provider  | Send | Yes | Yes |
| The broadcast event transmitted finished at date has now passed  | Raise | Yes | Yes |
| There is already a broadcast provider event in the DB for this broadcast event and the status is in 'sending'   | Send | Yes | No (but I think it just needs a correction to a test to add this) |
| There is already a broadcast provider event in the DB for this broadcast event and the status is in 'ack'   | Raise (probably shouldn't send as it's already been sent so should raise but I can see a small argument for letting it send (a problem at the CBC means they've asked us to manually resend))  | No | N/A |
| There is already a broadcast provider event in the DB for this broadcast event and the status is in 'tech-failure' | Raise (You spoke about how we can mark a broadcast provider event as tech failure to stop it attempting to send which I think is a good suggestion)  | No | N/A |
| There is a previous broadcast provider event for this provider but it has not yet been saved in the DB | Raise  | Yes | Yes |
| There is a previous broadcast provider event for this provider in the DB and the status of the previous one is 'sending' | Raise  | Yes | Yes |
| There is a previous broadcast provider event for this provider in the DB and the status of the previous one is 'ack' | Send | Yes | No |
| There is a previous broadcast provider event for this provider in the DB and the status of that one is 'technical-failure'  | Probably raise (Not sure what we should do. Your code currently will raise but I'm not sure if this is intentional or if it's a leftover from when we previously had a technical-failure indicate that something was still needing to be sent) | Yes | No |

We are in pretty good shape but there is some functionality missing that I think we should add and some test coverage to add (there is also some test coverage that I think will be added by fixing some of the status change stuff from technical failure to sending).

HOWEVER! I think this PR has been open too long and that you should fix the immediate comments I've left in this review about incorrect tests but not try and add any new functionality or any new tests. If you could also add any changes in separate commits on top then it will be quick to review too. That should hopefully be quick and then we can merge this. And then I suggest you open another PR with the additional things in the table that aren't covered.

Let me know if you want to talk in person or pair on this stuff!

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -471,3 +581,137 @@ def test_trigger_link_tests_invokes_cbc_proxy_client(
...
+    # but this task is only interested in the current task (the update) so doesn't worry about that
+    check_provider_message_should_send(current_event, 'ee')
+
+
+def test_check_provider_message_should_send_doesnt_raise_if_current_event_already_has_provider_message(sample_template):
+    broadcast_message = create_broadcast_message(sample_template)
+    current_event = create_broadcast_event(broadcast_message, message_type='alert')
+
+    # this might be set to technical-failure if we're currently retrying this event
+    create_broadcast_provider_message(current_event, provider='ee', status='technical-failure')
```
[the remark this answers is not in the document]
leohemsted: parametrized test to test all four failures, with a note that the tech-failure one should fail

idavidmcdonald (approved):
Thanks Leo.

As discussed on Slack, I hadn't covered the status of `ERR` for broadcast provider messages and we should also think about these in a subsequent PR in terms of test cases for when the current provider message is in this status and also when previous event provider messages are in this status

PR #3103, opened 2021-01-21
Commit db session before sending http request

servingUpAces:
Add a db.session.commit before sending the HTTPS request to the provider to release the db connection for other worker threads.

This should have a positive affect when the workers are really busy and our providers are slow with the responses, which does happen.

NOTE: will want to do this for the send_email_to_provider as well, this is just a draft to get an opinion

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -9,7 +9,7 @@ from notifications_utils.recipients import (
 )
 from notifications_utils.template import HTMLEmailTemplate, PlainTextEmailTemplate, SMSMessageTemplate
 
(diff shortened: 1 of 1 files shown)


app/delivery/send_to_providers.py
```diff
@@ -58,11 +58,16 @@ def send_sms_to_provider(notification):
 
         else:
             try:
+                recipient = notification.to
+                is_international = notification.international
+                notification_id = notification.id
+                reply_to_text = notification.reply_to_text
+                db.session.commit()
```
quis, line 65: I’m sure these lines are unlikely to raise an exception, but if they ever do it will change the provider priority because they’re within the `try` block. That doesn’t feel right to me.

PR #3105, opened 2021-01-25
Rename bt-ee-proxy to ee-proxy

klssmith:
We want to rename the `bt-ee-1-proxy` lambda function to `ee-1-proxy`.
This change will need to be deployed at the same time that we change
the name of the lambda function in the Terraform.

https://www.pivotaltracker.com/story/show/176551400


idavidmcdonald (approved):
Nice one

PR #3106, opened 2021-01-25
Add billing details fields to Service model and db table

CrystalPea:
The fields are:
Purchase order number - string field
Billing contact name - text field to acommodate possible multiple
contacts
Billing contact email address - text field to acommodate possible
multiple contacts
Billing reference - string field

All these fields are nullable. Notify platform admins will be
able to check and edit those values in Service Settings
section in Notify interface.

This will help make billing tasks and reports simpler.

Similar fields will also be added to Organisation model and
db table.


idavidmcdonald (commented):
I think you will need to make corresponding changes to `tests/app/service/test_rest.py` as we have added new fields to the `Service` model.

I think my main question is around when we get a service by ID from the API, what fields we should return in the JSON by default. Should we return all the service notes and billing details? Or should we only return those fields if the front end requests a detailed service such as https://github.com/alphagov/notifications-api/blob/master/app/service/rest.py#L205. Note, I'm genuinely asking as I'm not too sure what our general approach is when deciding what should go in the normal GET request for a service vs the more detailed request for a service

app/models.py
```diff
@@ -488,6 +488,10 @@ class Service(db.Model, Versioned):
     organisation = db.relationship('Organisation', backref='services')
 
     notes = db.Column(db.Text, nullable=True)
+    purchase_order_number = db.Column(db.String(255), nullable=True)
+    billing_contact_name = db.Column(db.Text, nullable=True)
+    billing_contact_email_address = db.Column(db.Text, nullable=True)
```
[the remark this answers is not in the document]
CrystalPea: I'm happy to change field names to `billing_contact_names` and `billing_contact_email_addresses`

The new table sounds interesting, the data would be more structured, but I also agree that it could be a slight overkill, and making layouts for displaying and editing it would be more faff.

CrystalPea:
As to `get_service_by_id`, for sure now we use it to create Service model in admin app and do things like `current_service.attribute`, so not having billing details there we would have to figure out how to do it differently. Is performance of `get_service_by_id` worth the effort? I don't know really 🤷🏼‍♀️ . I'd probably try to solve this problem if we see any trouble in latency, but if not, then imo it's a lot of work for something that we are not sure would be a problem.

You are right I need to update `test_get_service_by_id`, so I will push that and it should hopefully help Concourse pick the PR up :).

migrations/versions/0339_service_billing_details.py
```diff
@@ -0,0 +1,39 @@
...
+down_revision = '0338_add_notes_to_service'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('services', sa.Column('billing_contact_email_address', sa.Text(), nullable=True))
+    op.add_column('services', sa.Column('billing_contact_name', sa.Text(), nullable=True))
+    op.add_column('services', sa.Column('billing_reference', sa.String(length=255), nullable=True))
+    op.add_column('services', sa.Column('purchase_order_number', sa.String(length=255), nullable=True))
+    op.add_column('services_history', sa.Column('billing_contact_email_address', sa.Text(), nullable=True))
```
klssmith, line 22: If we don't need to store the billing history, I'd be curious to know if we can just not add these columns to `services_history`. A quick glance at the code suggests that would work (though I haven't tried it out) - not sure it really matters either way though

servingUpAces: I would include the history, it would be confusing if the tables were different and it will be useful to have the history for those columns.

app/models.py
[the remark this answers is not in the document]
servingUpAces: The only reason to use db.String(255) is to limit the length, there is no real difference between using db.text or  db.String(255) when it comes to performance or storage. Unless you have a reason to restrict the string length there is no advantage to using db.String(255). 255 seems an arbitrary number in this case.

servingUpAces:
Sounds like we are planning on putting the billing data on the service-settings page. That view uses current_service to display all the attributes, which means we need to include it in the service_schema.

migrations/versions/0339_service_billing_details.py
CrystalPea: I can ask Caley tomorrow if having historical records of billing details could be useful. If not, I can try and test what happens when we do not have those columns and update billing details.

klssmith: Sounds like it's good to have the history columns even if we don't look at them from Becca's comment, so I don't think you need to change anything here.

CrystalPea:
@servingUpAces it works for me as is (although I did have to update attribute names in admin PR after changing them here following @idavidmcdonald 's comment) 🤔

idavidmcdonald (approved):
Nice one

migrations/versions/0339_service_billing_details.py
CrystalPea: Definitely, sorry, Becca's comment didn't load for me yet when I was making my comment.

app/models.py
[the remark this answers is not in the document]
CrystalPea: That makes sense @servingUpAces , good to know it does not make difference to performance.

PR #3107, opened 2021-01-26
Add proxy client for o2 cell broadcasting

rjbaker:
o2 use One-2-many CBC so we can use the O2M/CAP client.

Once differences between CBCs have been worked out we can consolidate O2M clients to reduce duplication.


idavidmcdonald (approved):
Nice one

PR #3108, opened 2021-01-26
Stub training broadcasts

CrystalPea:
This PR stubs out broadcasts created from training mode accounts. For these broadcast messages:
- `BroadcastMessage` is created, with `stubbed` (new column, wow wow, check it out! :D ) set to True
- `BroadcastEvent` is created for each status change, as usual, but does not trigger `send_broadcast_event` task. Hence, `BroadcastProviderMessage`s are not created and cbc_proxy_client does not send any lambda invocations.

Before we merge it in, we have to ensure all MNO accounts on preview are moved to LIVE, so they can continue their tests.


migrations/versions/0340_stub_training_broadcasts.py
```diff
@@ -0,0 +1,25 @@
...
+import sqlalchemy as sa
+
+
+revision = '0340_stub_training_broadcasts'
+down_revision = '0339_service_billing_details'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('broadcast_message', sa.Column('stubbed', sa.Boolean(), nullable=True))
```
leohemsted, line 18: we should add a second PR to make this column non-nullable later (but we'll need a second PR so we can ensure new rows always get a value, and then in the new script we can migrate old rows and set the column to nullable=False)

leohemsted (approved):

migrations/versions/0340_stub_training_broadcasts.py
CrystalPea: that sounds good 👍🏼

PR #3109, opened 2021-01-27
Add February 2021 letter rates

klssmith:
All rates are changing, so we add an end date for the current rates and insert new rates for every post_class, sheet count and crown status.

To review - run the migration and check the rates against the new rates in the story https://www.pivotaltracker.com/n/projects/1443052/stories/176222554


CrystalPea (approved):
All the new rates in my local table look correct 👍🏼

PR #3111, opened 2021-01-28
Check request size preview

sakisv:
This is to replicate the nginx behaviour now that we're going to retire them.

A couple of questions that I'd like to hear your thoughts about:

- How important it is to have this check?
- Do you see this as a way to protect our instances against big payloads or as a way to control the size of the files our users can upload?

Depending on your answers we can choose between a) keeping the check here and creating a similar one for admin b) replacing this check with a check on cloudfront using something like WAF, or c) a combination of both (i.e. a "hard" limit on the WAF and more adjustable limit in the app).

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -19,6 +19,7 @@ from notifications_utils.clients.encryption.encryption_client import Encryption
 from notifications_utils import logging, request_helper
 from sqlalchemy import event
 from werkzeug.exceptions import HTTPException as WerkzeugHTTPException
+from werkzeug.exceptions import RequestEntityTooLarge as WerkzeugRequestEntityTooLarge
 from werkzeug.local import LocalProxy
 
 from app.celery.celery import NotifyCelery
@@ -280,6 +281,15 @@ def init_app(app):
 
         g.start = monotonic()
         g.endpoint = request.endpoint
+
+    @app.before_request
+    def check_content_length():
+        if (
+            request.content_length is not None
+            and current_app.config['MAX_CONTENT_LENGTH'] is not None
+            and request.content_length > current_app.config['MAX_CONTENT_LENGTH']
+        ):
+            raise WerkzeugRequestEntityTooLarge()
 
     @app.after_request
     def after_request(response):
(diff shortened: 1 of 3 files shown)


app/__init__.py
```diff
@@ -281,6 +282,15 @@ def record_request_details():
...
 
+    @app.before_request
+    def check_content_length():
+        if (
+            request.content_length is not None
+            and current_app.config['MAX_CONTENT_LENGTH'] is not None
+            and request.content_length > current_app.config['MAX_CONTENT_LENGTH']
+        ):
+            raise WerkzeugRequestEntityTooLarge()
+
```
servingUpAces, line 293: calling `request.data` does the same thing, does it not? 
However, I don't object to this extra code because it is clear what it is doing.
As long as this doesn't affect the performance of our API I would leave this check here. I would be surprised if this does degrade performance.

sakisv: You're right, it does pretty much the same thing (if you follow the calls from here: https://github.com/pallets/werkzeug/blob/f3b081634c744ae1e0ae3f76034ee90bfdc2a808/src/werkzeug/wrappers/request.py#L447)

But I found it a bit too magic so, as you said, I tried to be more explicit about what's happening.

leohemsted: `len(request.data)` will read the entire request data into memory in python (it'll probs still be in the socket at this point). `request.content_length` reads the CONTENT_LENGTH header.

sakisv: When were testing it yesterday, just by accessing `request.data` it was triggering the check and throwing the RequestEntityTooLarge, that's the "magic" bit I was referring to.

benthorner:
This PR has been around for a while, do you still want to get this reviewed?

sakisv:
I'll close this for now because for the time being we implemented this functionality in WAF [1]. It's worth keeping the branch around in case we want to revisit when we eventually switch api traffic over to cloudfront.

1: https://github.com/alphagov/notifications-aws/blob/master/terraform/modules/paas-proxy/waf.tf

PR #3112, opened 2021-01-28
Set broadcast channel as a service setting

idavidmcdonald:
Part 2 of https://www.pivotaltracker.com/story/show/176459886

Allows a developer to go into the DB and set which channel a service should go on to. This is the quickest and least we can do to unblock the networks from testing on different channels as they need to next week potentially.

Review commit by commit

For more context, see https://github.com/alphagov/notifications-broadcasts-infra/pull/102

The following still needs to happen in future PRs

Part 3: Create an endpoint for the API that will convert a service from a normal service to a broadcast service (put a row in the ServiceBroadcastSettings table and give the service the `broadcast` permission.
Part 4: Get the admin app to call the new API endpoint to make a service into a broadcast service
Part 5: Backfill data for all broadcast services that do not yet have a row in the `ServiceBroadcastSettings` table


app/celery/broadcast_message_tasks.py
```diff
@@ -70,6 +75,7 @@ def send_broadcast_provider_message(broadcast_event_id, provider):
             previous_provider_messages=broadcast_event.get_earlier_provider_messages(provider),
             sent=broadcast_event.sent_at_as_cap_datetime_string,
             expires=broadcast_event.transmitted_finishes_at_as_cap_datetime_string,
+            channel=channel
```
leohemsted, line 78: i wonder what happens if this channel is different to the initial broadcast's channel? i dont think we necessarily need to change anything, but i wonder about the relative merits of re-reading it here rather than pulling the data out of the original `alert` (somehow - we don't currently store the channel it went out on 🤔)

leohemsted (approved):
Looks pretty solid 👍 

One question about sending updates, but I don't think we necessarily need to solve that right now, since I don't expect we'll ever really be changing these channel values much.

app/celery/broadcast_message_tasks.py
idavidmcdonald: Good spot. I might leave an inline comment here about this.

Note, we did have a brief discussion the other day about storing the channel on the broadcast so we would be able to reference it always from there but decided given the channel is unlikely to change, we should just go for the quicker option, at least for now and just take it from the service every time

PR #3114, opened 2021-01-29
Don't stub broadcasts on preview

CrystalPea:
So that MNOs can use training mode accounts to test end-to-end broadcast sending. This will enable them to approve their own broadcasts.

### DIFF
--- a/app/broadcast_message/rest.py
+++ b/app/broadcast_message/rest.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 
 import iso8601
+
 from flask import Blueprint, jsonify, request, current_app
 from app.config import QueueNames
 from app.dao.dao_utils import dao_save_object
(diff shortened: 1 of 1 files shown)


tests/app/broadcast_message/test_rest.py
```diff
@@ -583,7 +584,10 @@ def test_update_broadcast_message_status_updates_details_but_does_not_queue_task
     assert len(bm.events) == 1
     alert_event = bm.events[0]
 
-    assert len(mock_task.mock_calls) == 0
+    if current_app.config['NOTIFY_ENVIRONMENT'] == 'preview':
```
idavidmcdonald, line 587: For this test, the `current_app.config['NOTIFY_ENVIRONMENT']` is `test` and you are not changing it to `preview`, meaning that we aren't actually running the assertion on line 588.

You would need to parametrize this test to give it two environments (or maybe all the environments) and then set the config at the top for the appropriate assertion to be used

CrystalPea: ahh, good comment, now I think about it, changing the environment could break tests. Test environment is required to connect to the right db etc. So I have scrapped the assertion that didn't do anything.

idavidmcdonald (approved):

PR #3122, opened 2021-02-05
Add billing details for organisation

CrystalPea:
Add to the model, migrate, and then serialise to be utilised in admin

Pivotal ticket: https://www.pivotaltracker.com/story/show/176411641
related admin PR, that needs to go in after this PR: https://github.com/alphagov/notifications-admin/pull/3788


migrations/versions/0342_org_billing_details.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('organisation', sa.Column('billing_contact_email_addresses', sa.Text(), nullable=True))
+    op.add_column('organisation', sa.Column('billing_contact_names', sa.Text(), nullable=True))
+    op.add_column('organisation', sa.Column('billing_reference', sa.String(length=255), nullable=True))
+    op.add_column('organisation', sa.Column('notes', sa.Text(), nullable=True))
+    op.add_column('organisation', sa.Column('purchase_order_number', sa.String(length=255), nullable=True))
+    op.alter_column('rates', 'rate',
```
klssmith, line 23: I can see why the organisation columns are being added, but were the changes to the rates and services table intentional?

CrystalPea: To services yes, to rates no, thanks for spotting! :)

CrystalPea: It would be great to do something about this ever growing number of things auto-added to migration file. Just I'm not sure what that would entail. Maybe a discussion subject for one of our afternoon cuddles ☺️

klssmith: The commit messages and PR description don't mention `services` so I'm not sure what the changes to the `services` table are for. Could you edit the commit message to explain what's being changed in this table and why?

CrystalPea: ok, no worries 👍🏼

CrystalPea: Done now, I also changed migration number to accommodate a migration that just got merged into master.

klssmith: Thanks 🙂

migrations/versions/0343_org_billing_details.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('organisation', sa.Column('billing_contact_email_addresses', sa.Text(), nullable=True))
+    op.add_column('organisation', sa.Column('billing_contact_names', sa.Text(), nullable=True))
+    op.add_column('organisation', sa.Column('billing_reference', sa.String(length=255), nullable=True))
+    op.add_column('organisation', sa.Column('notes', sa.Text(), nullable=True))
+    op.add_column('organisation', sa.Column('purchase_order_number', sa.String(length=255), nullable=True))
+    op.alter_column('services', 'billing_contact_email_addresses',
```
klssmith, line 23: Maybe I'm looking at the wrong thing here, but it looks to me like the `billing_contact_email_addresses` and `billing_contact_names` columns on `services` and `services_history` already had the type of `Text`?

That's from looking at the [model code](https://github.com/alphagov/notifications-api/blob/541a765811dbb5592822f8ebcf87ddc87acebf16/app/models.py#L492-L493) and the type of the columns in the database before running this migration

CrystalPea: That's what I thought, that I already did it, too, but then those alter statements were generated automatically when I was creating the migration file, so I assumed it must have not happened. 

But since you checked and the type `Text` is there, I can drop those statements 👍🏼

klssmith: 👍 The columns seem to be text in preview, staging and production, so maybe you just need to run those lines locally so your db matches.

klssmith (approved):

migrations/versions/0343_org_billing_details.py
CrystalPea: ahh, that makes sense, that will probably also solve the other automatic statements in my migrations. Sometimes I'm so slow 🤦🏼‍♀️. Thanks a lot!

klssmith: To be fair, I think most (all?) of the other automatic statements do need fixing!

PR #3128, opened 2021-02-09
Move provider restriction into broadcast service settings table

idavidmcdonald:
Move provider restriction into broadcast settings

This means we will have a much easier way of knowing what the settings
are for a broadcast service.

Note, we can just move data directly into the newer table as there is
nothing on the API or admin app that is putting data in the
`service_broadcast_provider_restriction` table, this was being done
manually for the few services that needed it.

### DIFF
--- a/tests/app/celery/test_broadcast_message_tasks.py
+++ b/tests/app/celery/test_broadcast_message_tasks.py
@@ -9,7 +9,6 @@ from app.models import (
     BroadcastStatusType,
     BroadcastEventMessageType,
     BroadcastProviderMessageStatus,
-    ServiceBroadcastProviderRestriction,
     ServiceBroadcastSettings,
 )
 from app.celery.broadcast_message_tasks import send_broadcast_event, send_broadcast_provider_message, trigger_link_test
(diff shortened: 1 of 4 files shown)


tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -74,10 +72,9 @@ def test_send_broadcast_event_does_nothing_if_provider_set_on_service_isnt_enabl
     notify_api,
     sample_service
 ):
-    notify_db.session.add(ServiceBroadcastProviderRestriction(
-        service=sample_service,
-        provider='three'
-    ))
+    settings = ServiceBroadcastSettings(service=sample_service, channel="test", provider="vodafone")
+    settings.provider = "three"
```
leohemsted, line 76: unless im missing something this can just be

```suggestion
    settings = ServiceBroadcastSettings(service=sample_service, channel="test", provider="three")
```

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -47,10 +46,9 @@ def test_send_broadcast_event_only_sends_to_one_provider_if_set_on_service(
     notify_api,
     sample_service
 ):
-    notify_db.session.add(ServiceBroadcastProviderRestriction(
-        service=sample_service,
-        provider='vodafone'
-    ))
+    settings = ServiceBroadcastSettings(service=sample_service, channel="test", provider="vodafone")
+    settings.provider = "vodafone"
```
leohemsted, line 50: dont think we need second line here

```suggestion
    settings = ServiceBroadcastSettings(service=sample_service, channel="test", provider="vodafone")
```

idavidmcdonald:
Good spots there Leo, I have fixed both of those and force pushed

leohemsted (approved):

PR #3132, opened 2021-02-10
Improvements to our letter checking tasks

idavidmcdonald:
Review commit by commit


CrystalPea (approved):
Looks good.

One comment to 3rd commit message: it says `This doesn't just relate to precompiled letters` but actually as far as I am aware only precompiled letters get checked for viruses?

idavidmcdonald:
Ah yes, you are right there Pea! Thanks for spotting. I'll leave it as is as I don't think the commit message inaccuracy will trip anyone up in the future.

PR #3133, opened 2021-02-10
Bump the migration revision ID by 1

idavidmcdonald:
Concourse had passed the tests before someone elses migration had beaten
me to getting merged. So I merged and got an error.

This should fix it

### DIFF
--- a/migrations/versions/0345_move_broadcast_provider.py
+++ b/migrations/versions/0345_move_broadcast_provider.py
@@ -1,7 +1,7 @@
 """
 
-Revision ID: 0344_move_broadcast_provider
-Revises: 0343_org_billing_details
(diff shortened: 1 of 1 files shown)


migrations/versions/0345_move_broadcast_provider.py
```diff
@@ -1,16 +1,16 @@
 """
 
 Revision ID: 0344_move_broadcast_provider
```
klssmith, line 3: This Revision ID should be 0345_move_broadcast_provider

idavidmcdonald: Just fixed this and rebased

klssmith (approved):

PR #3134, opened 2021-02-11
Improve sender task

servingUpAces:
This PR is to optimise the methods used to send notifications to providers by remove calls to the database. 
It probably easiest to go through it commit by commit.

- Remove validate_and_format for email and phone numbers, using normalised_to instead because that function has already happened when persisting the notificaiton.

- Remove 2 extra select queries after the update and commit. Once a transaction is committed SQLAlchemy will query for the db model if referenced after a commit.
 
- Use the cached template object.
  By adding SerialisedTemplate we can avoid a database call for the template. This is useful when sending many many emails/sms for the same template/version.

- User cache for service in send_to_provider methods. 
  This will remove a call to the db if the service exists in the cache.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -430,6 +430,7 @@ class Development(Config):
 
 
 class Test(Development):
+    SQLALCHEMY_ECHO = False
     NOTIFY_EMAIL_DOMAIN = 'test.notify.com'
     FROM_NUMBER = 'testing'
     NOTIFY_ENVIRONMENT = 'test'
--- a/app/config.py
+++ b/app/config.py
@@ -430,7 +430,6 @@ class Development(Config):
 
 
 class Test(Development):
-    SQLALCHEMY_ECHO = False
     NOTIFY_EMAIL_DOMAIN = 'test.notify.com'
     FROM_NUMBER = 'testing'
     NOTIFY_ENVIRONMENT = 'test'
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -28,11 +28,11 @@ from app.models import (
     NOTIFICATION_SENDING,
     NOTIFICATION_STATUS_TYPES_COMPLETED
 )
-from app.serialised_models import SerialisedTemplate
+from app.serialised_models import SerialisedTemplate, SerialisedService
 
 
 def send_sms_to_provider(notification):
-    service = notification.service
+    service = SerialisedService.from_id(notification.service_id)
     service_id = service.id
     if not service.active:
         technical_failure(notification=notification)
@@ -42,7 +42,7 @@ def send_sms_to_provider(notification):
         provider = provider_to_use(SMS_TYPE, notification.international)
 
         template_dict = SerialisedTemplate.get_dict(template_id=notification.template_id,
-                                                    service_id=service_id,
+                                                    service_id=service.id,
                                                     version=notification.template_version)['data']
 
         template = SMSMessageTemplate(
@@ -87,7 +87,7 @@ def send_sms_to_provider(notification):
 
 
 def send_email_to_provider(notification):
-    service = notification.service
+    service = SerialisedService.from_id(notification.service_id)
     service_id = service.id
     if not service.active:
         technical_failure(notification=notification)
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -15,7 +15,6 @@ from app.dao.provider_details_dao import (
     dao_reduce_sms_provider_priority
 )
 from app.celery.research_mode_tasks import send_sms_response, send_email_response
-from app.dao.templates_dao import dao_get_template_by_id
 from app.exceptions import NotificationTechnicalFailureException
 from app.models import (
     SMS_TYPE,
@@ -41,9 +40,9 @@ def send_sms_to_provider(notification):
     if notification.status == 'created':
         provider = provider_to_use(SMS_TYPE, notification.international)
 
-        template_dict = SerialisedTemplate.get_dict(template_id=notification.template_id,
-                                                    service_id=service.id,
-                                                    version=notification.template_version)['data']
+        template_dict = SerialisedTemplate.from_id_and_service_id(template_id=notification.template_id,
+                                                                  service_id=service_id,
+                                                                  version=notification.template_version).__dict__
 
         template = SMSMessageTemplate(
             template_dict,
@@ -95,9 +94,9 @@ def send_email_to_provider(notification):
     if notification.status == 'created':
         provider = provider_to_use(EMAIL_TYPE)
 
-        template_dict = SerialisedTemplate.get_dict(template_id=notification.template_id,
-                                                    service_id=service_id,
-                                                    version=notification.template_version)['data']
+        template_dict = SerialisedTemplate.from_id_and_service_id(template_id=notification.template_id,
+                                                                  service_id=service_id,
+                                                                  version=notification.template_version).__dict__
 
         html_email = HTMLEmailTemplate(
             template_dict,
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -28,6 +28,7 @@ from app.models import (
     NOTIFICATION_SENDING,
     NOTIFICATION_STATUS_TYPES_COMPLETED
 )
+from app.serialised_models import SerialisedTemplate
 
 
 def send_sms_to_provider(notification):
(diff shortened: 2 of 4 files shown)


app/delivery/send_to_providers.py
```diff
@@ -107,34 +105,32 @@ def send_email_to_provider(notification):
...
-
             reference = provider.send_email(
                 from_address,
-                validate_and_format_email_address(notification.to),
+                notification.normalised_to,
                 plain_text_email.subject,
                 body=str(plain_text_email),
                 html_body=str(html_email),
-                reply_to_address=validate_and_format_email_address(email_reply_to) if email_reply_to else None,
+                reply_to_address=notification.reply_to_text,
```
benthorner, line 127: While this is probably OK, I'm having trouble finding proof that the text in here has been validated equivalently at a previous stage, or if that's even important. 

There are multiple potential origins (see [here](https://github.com/alphagov/notifications-api/blob/2270832873c9a44d9cc86fa40ee5d8e6ef618e4c/app/service/send_notification.py#L126)) and at least one of them does it's own validation ([here](https://github.com/alphagov/notifications-api/blob/2270832873c9a44d9cc86fa40ee5d8e6ef618e4c/app/service/service_senders_schema.py#L9)). It would be helpful to have a bit more explanation in the commit message about where this is set.

tests/app/delivery/test_send_to_providers.py
```diff
@@ -724,51 +730,3 @@ def test_send_email_to_provider_uses_reply_to_from_notification(
...
-def test_send_sms_to_provider_should_format_phone_number(sample_notification, mocker):
-    sample_notification.to = '+44 (7123) 123-123'
-    send_mock = mocker.patch('app.mmg_client.send_sms')
-
-    send_to_providers.send_sms_to_provider(sample_notification)
-
-    assert send_mock.call_args[1]['to'] == '447123123123'
-
-
-def test_send_email_to_provider_should_format_email_address(sample_email_notification, mocker):
```
idavidmcdonald, line 759: Optional: You could keep this test and the one above it but rejig them so it's

`test_send_email_to_provider_uses_normalised_to_address_rather_than_email_address`

which would persist a notification with one value for the `to` field and a different value for the `normalised_to` field and then checks that the we call `send_mock` with the normalised value and not the unnormalised value.

You have this behaviour potentially covered in some of your tests above but the testing of this functionality above is all implicit, a separate test would make clear the behaviour that we expect

app/delivery/send_to_providers.py
```diff
@@ -52,14 +49,14 @@ def send_sms_to_provider(notification):
             show_prefix=service.prefix_sms,
         )
 
+        created_at = notification.created_at
```
idavidmcdonald, line 55: Just to check, what benefit do you see us getting from not doing the extra select query after the insert?

- It will mean we do one fewer query which is positive, although I don't think of massive impact
- I don't think it will release a DB connection early? Leo mentioned something about lack of DB connection pooling and it being a one to one with celery workers anyway? Or did I misunderstand this?

tests/app/delivery/test_send_to_providers.py
```diff
@@ -730,3 +731,53 @@ def test_send_email_to_provider_uses_reply_to_from_notification(
...
+                                       to_field= '+447700900855',
+                                       normalised_to=validate_and_format_phone_number('+447700900855'))
+    mocker.patch(
+        'app.redis_store.get',
+        side_effect=[
+            json.dumps({'data': template_dict}).encode('utf-8'),
+        ],
+    )
+    mock_get_template = mocker.patch(
+        'app.dao.templates_dao.dao_get_template_by_id_and_service_id'
```
benthorner, line 751: Minor: initially I struggled to understand the relevance of this patch, until I looked at the definition of `SerialisedTemplate.get_dict`. This is coupled to it. We could simplify this test by just checking the call to the serialised model, and trust that it implements it efficiently (or have a separate test for the method itself).

app/delivery/send_to_providers.py
```diff
@@ -40,10 +41,12 @@ def send_sms_to_provider(notification):
     if notification.status == 'created':
         provider = provider_to_use(SMS_TYPE, notification.international)
 
-        template_model = dao_get_template_by_id(notification.template_id, notification.template_version)
+        template_dict = SerialisedTemplate.get_dict(template_id=notification.template_id,
```
benthorner, line 44: Could we use `from_id_and_service_id` instead? It looks like that may be even more efficient, due to `@memory_cache`.

tests/app/delivery/test_send_to_providers.py
```diff
@@ -736,48 +736,59 @@ def test_send_email_to_provider_uses_reply_to_from_notification(
 def test_send_sms_to_provider_should_return_template_if_found_in_redis(
         mocker, client, sample_template
 ):
-
-    from app.schemas import template_schema
+    from app.schemas import service_schema, template_schema
```
benthorner, line 739: This line is repeated in the the test for emails. Does it make sense to DRY it up as a global import for the module?

benthorner (commented):
Using the serialised models looks like a nice improvement, and it was helpful for me to learn the models exist. If we have an optimisation feature like that already then it just makes sense to consistently use it everywhere.

I've adjusted my comments so they don't repeat @idavidmcdonald's ones. A couple are just to check my understanding.

tests/app/delivery/test_send_to_providers.py
```diff
@@ -103,7 +103,9 @@ def test_should_send_personalised_template_to_correct_sms_provider_and_persist(
     mocker
 ):
     db_notification = create_notification(template=sample_sms_template_with_html,
-                                          to_field="+447234123123", personalisation={"name": "Jo"},
+                                          to_field="+447234123123",
```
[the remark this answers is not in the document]
servingUpAces: `to` is the required field, `normalised_to` is usually derived from the to field but the unit test isn't calling the `persist_notification` method.

app/delivery/send_to_providers.py
servingUpAces: If you look at the calls the database makes you will get the extra query, because if you access a data model object after a commit the object is considered "dirty", so it will re-query to get the "current" data model object. The extra query happens regardless of what's happening with the connection pooling, regardless of the connection being released to the pool or not.

app/delivery/send_to_providers.py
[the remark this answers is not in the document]
servingUpAces: The reply_to_text is validated when added to the service, I don't see the need to validate it each time the service sends an email.
I suspect the validation was there before we confirmed the validity of the email address from the admin app. I don't believe it is necessary now.

app/config.py
```diff
@@ -430,6 +430,7 @@ class Development(Config):
 
 
 class Test(Development):
+    SQLALCHEMY_ECHO = False
```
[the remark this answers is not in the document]
servingUpAces: I can remove it, I tend to add this line in a lot when I want to see what's going on with the db.

app/delivery/send_to_providers.py
servingUpAces: The reply_to_text is validated when added to the service, I don't see the need to validate it each time the service sends an email.
I suspect the validation was there before we confirmed the validity of the email address from the admin app. I don't believe it is necessary now.

tests/app/delivery/test_send_to_providers.py
servingUpAces: The service_schema and template_schema need to be imported for the test locally because of the dependency conflicts with Marshmallow and SQLAlchemy. 
We get ` AttributeError: 'DummySession' object has no attribute 'query'` error if you globally import the schemas.

tests/app/delivery/test_send_to_providers.py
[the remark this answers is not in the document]
benthorner: Could you explain what you mean by "required field". I know it's required at user input, but the code doesn't seem to rely on it at this point (with `normalised_to` populated), so I'm not sure why we need to specify it.

app/delivery/send_to_providers.py
benthorner: OK. We do validate them when they are added. If that's the only way for this field to be populated, then I agree there's no need to validate at this point.

tests/app/delivery/test_send_to_providers.py
benthorner: Ah, OK. Thanks for explaining. Sounds like an annoying quirk! It would be helpful to add that to the commit message, as someone in future may also ask why the import is at the function level.

benthorner (approved):
Looks good to me.

idavidmcdonald (approved):

PR #3135, opened 2021-02-11
Refactor the CBC proxy classes

klssmith:
By creating a new `CBCProxyOne2ManyClient` class for the three One2Many
clients to inherit from. These three clients are the same apart from the
`lambda_name` and the `failover_lambda_name`.


servingUpAces (approved):
looks good to me - always a fan of deleting code and code reuse.

PR #3144, opened 2021-02-16
Stop emailing Notify when an MOU is signed

klssmith:
We've decided we don't get any value from these emails any more, so this stops us (Notify support) receiving them. We still let teams know an MOU has been signed.

[Pivotal story](https://www.pivotaltracker.com/story/show/176907414)

**❓Question❓**
I haven't done anything in the code with the template that was being used - this could be either be deleted or moved to the "archive" folder. Any thoughts on where this should go, and whether this is best done in a migration or if just moving / deleting it through the admin interface is fine?


servingUpAces (approved):
It looks like we have a pattern of moving templates to the archived folder for the Notify service, however this template only sent messages to our team, so I would be alright with just deleting the template from the admin app after this deploy.

klssmith:
Thanks @servingUpAces, a migration would mean that the template also gets deleted for developers locally too. But having an extra template in the Notify service of your local environment doesn't cause any issues, so I like your suggestion of just deleting from admin

PR #3146, opened 2021-02-17
Send text messages from a number for international

servingUpAces:
Send text message that are to an international number from a number rather than "Notify"

This isn't production ready but just a sample of what we can do to handle the international text messages sent from Notify. We could use an inbound number for this.

We may want to implement this for other notifications from Notify but this is the one that prevents users from logging in.


migrations/versions/0346_notify_number_sms_sender.py
```diff
@@ -0,0 +1,39 @@
...
+    # by adding a row in inbound_number we ensure the number isn't added to the table and assigned to a service.
+    inbound_number_sql = f"""INSERT INTO INBOUND_NUMBERS (id, number, provider, active, created_at)
+            VALUES('{inbound_number_id}', '{INBOUND_NUMBER}', 'mmg', false, now())
+    """
+    op.execute(inbound_number_sql)
+
+
+def downgrade():
+    delete_sms_sender = f"delete from service_sms_senders where id = '{SMS_SENDER_ID}'"
+    delete_inbound_number = f"delete from inbound_numbers where number = '{INBOUND_NUMBER}'"
```
leohemsted, line 37: this won't work since INBOUND_NUMBER will always be `07xxx`. (assuming we'll run this script and then manually update the row directly in the db after migration has run)

we could fetch the inbound_number by grabbing it from the service_sms_senders just before we delete, or just leave a note here that there is a manual cleanup step that needs to happen

servingUpAces: This now uses the real inbound number - so the downgrade will work. 07xx was just to sample the code, so I can find out if I should proceed with the PR.

leohemsted: ah ok, wasnt sure if you wanted to hide the number form the upgrade. if we want the phone number to be publicly searchable (dont think that's the end of the world) then that's fine

app/config.py
```diff
@@ -179,6 +179,7 @@ class Config(object):
     MOU_SIGNED_ON_BEHALF_SIGNER_RECEIPT_TEMPLATE_ID = 'c20206d5-bf03-4002-9a90-37d5032d9e84'
     MOU_SIGNED_ON_BEHALF_ON_BEHALF_RECEIPT_TEMPLATE_ID = '522b6657-5ca5-4368-a294-6b527703bd0b'
     MOU_NOTIFY_TEAM_ALERT_TEMPLATE_ID = 'd0e66c4c-0c50-43f0-94f5-f85b613202d4'
+    NOTIFY_NUMBER_SMS_SENDER = '07984404008'
```
leohemsted, line 182: small detail, can we call this var something like `NOTIFY_INTERNATIONAL_SMS_SENDER` so it's obvious what it's used for?

leohemsted (approved):
looks good, just one small thing

migrations/versions/0346_notify_number_sms_sender.py
servingUpAces: If we haven't associated the number with a service I think it's okay.

app/config.py
servingUpAces: good idea - sometimes naming is hard

PR #3149, opened 2021-02-17
✨ Add new API endpoint for retrieving a "blocklist" of protected senders.

dvjones89:
To help avoid abuse of the CAST Notify service, as well as protect our
accounts with backend providers like `firetext`, we want to apply some
restrictions on the Sender IDs that users can apply to our outbound SMS
and emails.

This PR adds a new (admin-only) API endpoint to return a "blocklist" of
protected phrases (such as `GOVUK` or `NHS`) that shouldn't be allowed
to feature in user-submitted SMS senders. This new "blocklist" is
retrieved by our front-end admin app and enforced by a new custom
validator.

### DIFF
--- a/app/dao/service_sms_sender_dao.py
+++ b/app/dao/service_sms_sender_dao.py
@@ -5,6 +5,8 @@ from app.dao.dao_utils import transactional
 from app.exceptions import ArchiveValidationError
 from app.models import ServiceSmsSender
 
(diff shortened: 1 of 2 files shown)


dvjones89:
Oops, I intended to raise this against our own fork, rather than the official `alphagov` repository. Apologies for the confusion!

PR #3150, opened 2021-02-17
Simplify README and consolidate scripts

benthorner:
Partly addresses: https://github.com/alphagov/notifications-manuals/issues/9


README.md
```diff
@@ -43,11 +33,14 @@ export WERKZEUG_DEBUG_PIN=off
...
- * The SECRET_KEY and DANGEROUS_SALT should match those in the [notifications-admin](https://github.com/alphagov/notifications-admin) app.
- * The  unique prefix for the queue names prevents clashing with others' queues in shared amazon environment and enables filtering by queue name in the SQS interface.
+* Replace `YOUR_OWN_PREFIX` with `local_dev_<first name>`.
+* Run the following in the credentials repo to get the API keys.
+
+```
+notify-pass credentials/providers/api_keys
+```
 
 ### Postgres
```
servingUpAces, line 45: Can we update the Postgres part we don't need to down grade to version 9.6.

benthorner: I remember trying this app with v13, but some of the tests fail as we look for specific strings in the exceptions. We could say v9.X, or maybe point to the infra version setting. How about that?

scripts/run_app.sh
```diff
@@ -1,6 +0,0 @@
-#!/bin/bash
-
-set -e
```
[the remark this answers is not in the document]
sakisv: Makefile will by default exit on the first failed command (or more precisely, the first command with a non-zero exit code)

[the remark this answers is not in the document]
benthorner: Thanks for confirming @sakisv, and good question @idavidmcdonald. For this particular file, it's also irrelevant as the new command uses an `&&`. The exit behaviour is still important in general, especially if we want to move the test script in here.

README.md
servingUpAces: It should be at least version 11

benthorner: So it is. I've added another commit to update the instructions there. Good to know we're nearer the cutting edge 👍.

leohemsted: i'd remove the "up to", so rather than "works with versions up to v11" i'd say "works with version v11" or something. we dont want to encourage people to install pg versions that don't match prod.

leohemsted (approved):
just one minor comment on wording of the postgres version, but other than that looks really good, thanks for this!

README.md
benthorner: Sure, I've amended the commit with that 👍.

PR #3151, opened 2021-02-17
Send daily email with volumes of letters and sheets to DVLA

CrystalPea:
Send daily email with volumes of letters and sheets to DVLA around 6pm so they know how many letters to expect for printing.

To do this we need to:
- create an email template by migration - done here
- add DVLA email to our credentials - done: https://github.com/alphagov/notifications-credentials/pull/199
- gather volumes of letters and sheets - done here
- create a notification and put it on the queue - done here

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -14,12 +14,12 @@ from app import encryption, notify_celery
 from app.aws import s3
 from app.config import QueueNames, TaskNames
 from app.dao.notifications_dao import (
+    dao_get_letters_to_be_printed,
+    dao_get_notification_by_reference,
+    dao_update_notification,
+    dao_update_notifications_by_reference,
     get_notification_by_id,
     update_notification_status_by_id,
-    dao_update_notification,
-    dao_get_notification_by_reference,
-    dao_update_notifications_by_reference,
-    dao_get_letters_to_be_printed,
 )
 from app.letters.utils import get_letter_pdf_filename
 from app.errors import VirusScanError
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -14,6 +14,7 @@ from app import encryption, notify_celery
 from app.aws import s3
 from app.config import QueueNames, TaskNames
 from app.dao.notifications_dao import (
+    dao_get_letters_and_sheets_volume_by_postage,
     dao_get_letters_to_be_printed,
     dao_get_notification_by_reference,
     dao_update_notification,
@@ -21,6 +22,7 @@ from app.dao.notifications_dao import (
     get_notification_by_id,
     update_notification_status_by_id,
 )
+from app.dao.templates_dao import dao_get_template_by_id
 from app.letters.utils import get_letter_pdf_filename
 from app.errors import VirusScanError
 from app.exceptions import NotificationTechnicalFailureException
@@ -36,6 +38,8 @@ from app.letters.utils import (
 )
 from app.models import (
     INTERNATIONAL_LETTERS,
+    INTERNATIONAL_POSTAGE_TYPES,
+    KEY_TYPE_NORMAL,
     KEY_TYPE_TEST,
     NOTIFICATION_CREATED,
     NOTIFICATION_DELIVERED,
@@ -45,7 +49,9 @@ from app.models import (
     NOTIFICATION_VIRUS_SCAN_FAILED,
     POSTAGE_TYPES,
     RESOLVE_POSTAGE_FOR_FILE_NAME,
-    INTERNATIONAL_POSTAGE_TYPES)
+    Service,
+)
+
 from app.cronitor import cronitor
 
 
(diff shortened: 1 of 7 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -137,12 +142,23 @@ def collate_letter_pdfs_to_be_sent():
...
+        letters_volume[postage + '_sheets'] = 0
+
         for i, letters in enumerate(group_letters(letters_to_print)):
-            filenames = [letter['Key'] for letter in letters]
+            letters_volume[postage] += len(letters)
+
+            filenames = []
+            for letter in letters:
+                filenames.append(letter['Key'])
+                letters_volume[postage + '_sheets'] += letter['Size']
```
servingUpAces, line 160: `letter['size']` represents file size rather than the sheet count of the letter.

app/celery/letters_pdf_tasks.py
```diff
@@ -173,9 +189,53 @@ def collate_letter_pdfs_to_be_sent():
             )
         current_app.logger.info(f"finished collate-letter-pdfs-to-be-sent processing for postage class {postage}")
 
+    send_letters_volume_email_to_dvla(letters_volume)
     current_app.logger.info("finished collate-letter-pdfs-to-be-sent")
 
 
+def send_letters_volume_email_to_dvla(letters_volume):
```
servingUpAces, line 196: Have you considered using a query to get the volume data?
It would mean `send_letters_volume_email_to_dvla` could be run on it's own or as a task and we can send the email before we start sending the files. The query is efficient and would look something like this: 
```
select count(*), sum(billable_units), postage
from notifications 
where notification_type = 'letter'
and key_type = 'normal'
and notification_status = 'created'
and created_at <= print_run_deadline
group by postage
```

app/celery/letters_pdf_tasks.py
CrystalPea: ohh, ok, good to know, I will add billable units to the query instead. I thought I was being smart hahah :D

app/celery/letters_pdf_tasks.py
CrystalPea: Hmmm, that is a sound suggestion.

migrations/versions/0346_add_dvla_volumes_template.py
```diff
@@ -0,0 +1,81 @@
+"""
+
+Revision ID: 0346_add_dvla_volumes_template
```
servingUpAces, line 3: You will need to fix the merge conflict, there is a 0346 file already

migrations/versions/0347_add_dvla_volumes_template.py
```diff
@@ -0,0 +1,81 @@
...
+            email_template_content,
+            current_app.config['NOTIFY_SERVICE_ID'],
+            email_template_subject,
+            current_app.config['NOTIFY_USER_ID'],
+            'normal'
+        )
+    )
+
+
+def downgrade():
```
servingUpAces, line 76: We might consider putting a if environment NOT production then do delete.

app/celery/letters_pdf_tasks.py
```diff
@@ -170,6 +179,58 @@ def collate_letter_pdfs_to_be_sent():
...
+        "second_class_sheets": 0,
+        'international_sheets': 0
+    }
+    for item in letters_volumes:
+        personalisation['total_volume'] += item.letters_count
+        personalisation['total_sheets'] += item.sheets_count
+        if f"{item.postage}_class_volume" in personalisation:
+            personalisation[f"{item.postage}_class_volume"] = item.letters_count
+            personalisation[f"{item.postage}_class_sheets"] = item.sheets_count
+        else:
```
servingUpAces, line 204: Did DVLA ask for the rest-of-world and europe volumes to be aggregated as international?

servingUpAces (commented):
I have a couple of small questions otherwise it looks good.

app/celery/letters_pdf_tasks.py
CrystalPea: they have slight preference for aggregated, but also they said whatever is easier for us.

migrations/versions/0347_add_dvla_volumes_template.py
CrystalPea: I'm not sure I understand, what do you mean here?

CrystalPea: You mean we should not have a downgrade for this migration in PROD?

servingUpAces: Yes that's what I mean - I am nervous about deleting data in prod

app/celery/letters_pdf_tasks.py
servingUpAces: okay

migrations/versions/0347_add_dvla_volumes_template.py
CrystalPea: ok 👍🏼

servingUpAces (approved):

PR #3153, opened 2021-02-18
🔨 Add script to set sensible DB defaults in dev

dvjones89:
When the application is first setup using `scripts/bootstrap.sh`, the
database is seeded with some default records that aren't entirely suited
to local development. For example, the default SMS sender is set to
"GOVUK" (a sender ID that'll get you banned with SMS backends) as well
as seeding a default user without any permissions and a randomly
generated (and therefore unknown) login password.

This commit adds a new `apply_dev_defaults.py` script, adding/updating
database records so it's easier for a developer to login and get up and
running with minimal fuss.

### DIFF
--- a/scripts/apply_dev_defaults.py
+++ b/scripts/apply_dev_defaults.py
@@ -1,0 +1,51 @@
+##!/usr/bin/env python
+import os, sys
+sys.path.insert(1, os.path.join(sys.path[0], '..'))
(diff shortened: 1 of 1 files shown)


dvjones89:
Oops, I intended to raise this against our own fork, rather than the official alphagov repository. Apologies for the confusion!

PR #3154, opened 2021-02-18
Update SerialisedService

servingUpAces:
I need to update the SeriralisedService object before deploying https://github.com/alphagov/notifications-api/pull/3145

If the send_to_provider grabs a SerialisedService that doesn't have these properties the task will fail. If we deploy the change to the SerialisedService first, clear the cache, then deploy https://github.com/alphagov/notifications-api/pull/3145 we can avoid that issue.

### DIFF
--- a/app/serialised_models.py
+++ b/app/serialised_models.py
@@ -75,6 +75,7 @@ class SerialisedTemplate(SerialisedModel):
 class SerialisedService(SerialisedModel):
     ALLOWED_PROPERTIES = {
         'id',
+        'name',
         'active',
         'contact_link',
         'email_from',
@@ -83,6 +84,8 @@ class SerialisedService(SerialisedModel):
         'rate_limit',
         'research_mode',
         'restricted',
+        'prefix_sms',
+        'email_branding'
     }
 
     @classmethod


idavidmcdonald (approved):

idavidmcdonald:
So after some help from CHS, I think this PR does not need to be merged separately. Here is why.

So the keys that are found in redis is completely reliant on the key being set here - https://github.com/alphagov/notifications-admin/blob/master/app/notify_client/service_api_client.py#L42, which sets all the keys in redis for our service which come back in the API response. Note, the keys sent to us by the API are determined by `ServiceSchema.dump`.

In the API we have a list of allowed properties here - https://github.com/alphagov/notifications-api/blob/master/app/serialised_models.py#L76

In the admin we have a similar list of allowed properties here - https://github.com/alphagov/notifications-admin/blob/master/app/models/service.py#L31

In both of these two cases, the list of properties will take the keys that are stored in redis and put the allowed_properties as python properties on the python object. As long as the key exists in redis already (because the API is sending it out), then you can add a key to the list of `ALLOWED_PROPERTIES` and it will just work.

The purpose of the `ALLOWED_PROPERTIES` list is really about catching typos early, so it’s flagged if you try to do `email_branding_id` instead of `email_branding`.

Keeping the list smaller in the API does mean that the value we cache in memory is slightly smaller, but this is probably negligible and so there is no particular reason why the two lists of `ALLOWED_PROPERTIES` can't match.


So in conclusion, the only reason you need to do a two step deploy is when adding or removing the return value of ServiceSchema.dump, which is not the case here.

PR #3155, opened 2021-02-18
Backfill services_broadcast_settings table

idavidmcdonald:
https://www.pivotaltracker.com/story/show/176459886

Review commit by commit

This is the final piece of the puzzle and will complete the above pivotal ticket.

Relies on https://github.com/alphagov/notifications-admin/pull/3794 being merged first

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -25,7 +25,7 @@ from notifications_utils.recipients import (
 
 from app import ma
 from app import models
-from app.models import ServicePermission, BROADCAST_TYPE
+from app.models import ServicePermission
 from app.dao.permissions_dao import permission_dao
 from app.utils import DATETIME_FORMAT_NO_TIMEZONE, get_template_instance
 
@@ -242,14 +242,7 @@ class ServiceSchema(BaseSchema, UUIDsAsStringsMixin):
         return service.allowed_broadcast_provider
 
     def _get_broadcast_channel(self, service):
-        # TODO: Once we've migrated data so that all broadcast services have `service.broadcast_channel`
-        # set then we can remove this logic and related tests and instead just return
-        # `service.broadcast_channel`. For the moment though, as we have some services with the broadcast
-        # permission that do not have a row in the service_broadcast_settings table, we need to hardcode
-        # this in here to give them a default that the admin app can use
-        if BROADCAST_TYPE in self.service_permissions(service):
-            return service.broadcast_channel if service.broadcast_channel else "test"
-        return None
+        return service.broadcast_channel
 
     def get_letter_logo_filename(self, service):
         return service.letter_branding and service.letter_branding.filename
--- a/migrations/versions/0348_migrate_broadcast_settings_migrate_broadcast_settings.py
+++ b/migrations/versions/0348_migrate_broadcast_settings_migrate_broadcast_settings.py
@@ -1,0 +1,48 @@
+"""
+
+Revision ID: 0348_migrate_broadcast_settings
+Revises: 0347_add_dvla_volumes_template
+Create Date: 2021-02-18 15:25:30.667098
+
+"""
+from alembic import op
+import sqlalchemy as sa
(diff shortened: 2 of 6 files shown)


migrations/versions/0348_migrate_broadcast_settings_migrate_broadcast_settings.py
```diff
@@ -0,0 +1,48 @@
...
+revision = '0348_migrate_broadcast_settings'
+down_revision = '0347_add_dvla_volumes_template'
+
+
+def upgrade():
+  # For every service that has the broadcast permission we want it to have
+  # a row in the broadcast_service_settings table
+  # 
+  # If it doesnt have a row already, then:
+  # - if the service is in trial mode, add a row and set the channel as 'severe'
```
benthorner, line 21: Just to check my understanding: for the current test channel services in trial mode, does this mean we'll need to manually amend them in each environment? If so, do we have a good idea which services these are?

app/schemas.py
```diff
@@ -242,14 +242,7 @@ def _get_allowed_broadcast_provider(self, service):
         return service.allowed_broadcast_provider
 
     def _get_broadcast_channel(self, service):
```
benthorner, line 244: Do we need the method anymore? It looks like it's only used once, just above.

benthorner (commented):
Looks good - nice to see we can remove the temporary code 🧹. I've just got one more major question about the migration. I've tried it locally with some test data it it seems to work as expected, so it's just the overall behaviour I'm wondering about.

migrations/versions/0348_migrate_broadcast_settings_migrate_broadcast_settings.py
idavidmcdonald: I don't think we will need to do any manual intervention, this migration should take care of all services and put them in a legit state (one that meets one of the radio options on the platform admin page to choose a broadcast account type).

For current broadcast services in trial mode, they will be be changed by this migration to be in trial mode but instead send to the severe channel (rather than default to the test channel). This change shouldn't really have an impact on any services. On preview there will be a bunch of services that fall into this category but that is OK as it's our preview environment so we don't really mind what channel it gets sent out on. For production, as training mode services don't send alerts, again we don't really mind if we change their channel from test to severe.

Let me know if you want me to chat this through in person if that is easier.

app/schemas.py
idavidmcdonald: I think so, or at least I'm just copying what we are doing for `_get_allowed_broadcast_provider` so I assume that one was needed too (it's only used above as well).

benthorner: Oh yes, good point. I did a bit of quick reading marshmallow, and this does seem to be the technique to use if we want to serialize an association without defining a nested schema for it.

Note to self: these schemas are used for a combination of API responses ([here](https://github.com/alphagov/notifications-api/blob/1b22de3830181809fb790cb06b9c2b647ce01cb3/app/service/rest.py#L191)), for which this attribute is required; and reduced DB queries ([here](https://github.com/alphagov/notifications-api/blob/1b22de3830181809fb790cb06b9c2b647ce01cb3/app/celery/tasks.py#L194)), for which it's not necessary (`dump_only`).

benthorner (approved):
Thanks for looking into my comments. I'm still a bit unclear on the effect of the migration on users, but we can always manually correct that if it turns out we need to preserve the channel, so it's not a big deal.

benthorner:
> I'm still a bit unclear on the effect of the migration on users

After talking this through, @idavidmcdonald explained that the channel setting ("test" or "severe") has little impact on the behaviour in Preview:

- All broadcast messages only go to test labs for the MNOs.

- "severe" is generally opt-in, "test" opt-out, so forcing "severe" is unlikely to affect test outcomes, except that the behaviour of the phone (e.g. the sound) might differ slightly.

PR #3158, opened 2021-02-23
Turn on o2 and three mnos in prod

idavidmcdonald:
Supporting infrastructure is ready for these two mnos


rjbaker (approved):
Confirm production CB infra has been deployed

PR #3159, opened 2021-02-24
Add date to personalisation for DVLA email

CrystalPea:
Personalisation was missing date attribute. The email still got sent last night, just it didn't have a value for date placeholder.

Also add DVLA_EMAIL_ADDRESS to manifest file, so it gets picked up from Credentials.


servingUpAces (approved):
I think this is ok, it's worth running an end to end test on staging by triggering the task. Even if the query is returns nothing you should still send an email since you initialise the personalisation with zeros.

CrystalPea:
@servingUpAces  good suggestion, only in staging we do not pull the variable from CREDENTIALS at present. Would you like me to change that? And if so, what email should I use? My own, to see the email coming in?

PR #3161, opened 2021-02-24
Check international rules to decide on a sender

servingUpAces:
Check international number for alpha: NO if true then use number to send SMS.

This is not a catch all for international SMS, the rules are quite
complex and still not completely understood. We are talking with our
provider who maybe able to sort this out for us. But in the meantime,
this should solve for the case that we understand.

- [x]  requires https://github.com/alphagov/notifications-utils/pull/843

### DIFF
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -105,7 +105,7 @@ def update_user_attribute(user_id):
         elif 'mobile_number' in update_dct:
             template = dao_get_template_by_id(current_app.config['TEAM_MEMBER_EDIT_MOBILE_TEMPLATE_ID'])
             recipient = user_to_update.mobile_number
-            reply_to = set_reply_to(recipient, template)
+            reply_to = get_sms_reply_to_for_notify_service(recipient, template)
         else:
             return jsonify(data=user_to_update.serialize()), 200
         service = Service.query.get(current_app.config['NOTIFY_SERVICE_ID'])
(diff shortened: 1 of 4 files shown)


app/user/rest.py
```diff
@@ -133,6 +130,21 @@ def update_user_attribute(user_id):
     return jsonify(data=user_to_update.serialize()), 200
 
 
+def set_reply_to(recipient, template):
+    # if is_uk_phone_number(recipient):
```
idavidmcdonald, line 134: Is all this commented out code supposed to be here?

servingUpAces: remove it

app/user/rest.py
```diff
@@ -133,6 +130,14 @@ def update_user_attribute(user_id):
     return jsonify(data=user_to_update.serialize()), 200
 
 
+def set_reply_to(recipient, template):
```
idavidmcdonald, line 133: I would suggest adding some unit tests to this method please. The `test_post_user_attribute_with_updated_by_sends_notification_to_international_from_number` test is good but definitely doesn't cover all the logic of this method

idavidmcdonald, line 133: This method name is not as precise as I think it should be. I think there needs to be something in the name to show that it is to do with setting the reply to for messages sent from the Notify service, rather than any generic service.

Also this method isn't `setting` anything, it's `getting` something.

I'd suggest `get_reply_to_for_notify_service`

idavidmcdonald (commented):
Two comments which shouldn't take too long to resolve but apart from that, it looks good :)

idavidmcdonald (approved):

PR #3162, opened 2021-02-26
Change DVLA_EMAIL_ADDRESS to a list

servingUpAces:
DVLA would like to get the daily volume messages sent to a list of people. 

- [ ] https://github.com/alphagov/notifications-api/pull/3162

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -181,7 +181,7 @@ class Config(object):
     NOTIFY_INTERNATIONAL_SMS_SENDER = '07984404008'
     LETTERS_VOLUME_EMAIL_TEMPLATE_ID = '11fad854-fd38-4a7c-bd17-805fb13dfc12'
     # we only need real email in Live environment (production)
-    DVLA_EMAIL_ADDRESS = 'success@simulator.amazonses.com'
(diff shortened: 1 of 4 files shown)


app/config.py
```diff
@@ -539,7 +541,7 @@ class Live(Config):
     CRONITOR_ENABLED = True
 
     ENABLED_CBCS = {BroadcastProvider.THREE, BroadcastProvider.O2, BroadcastProvider.VODAFONE}
-    DVLA_EMAIL_ADDRESS = os.getenv('DVLA_EMAIL_ADDRESS')
+    DVLA_EMAIL_ADDRESSES = os.getenv('DVLA_EMAIL_ADDRESSES')
```
leohemsted, line 544: you'll need to call `json.loads` here (as we do elsewhere). and set a default val of `[]` probably.

also, can this be deduplicated by moving it from `Live` to the default config? that'll make it consistent with how we load all the other environment variables

leohemsted (changes requested):

app/config.py
servingUpAces: good catch - done.

leohemsted (approved):

PR #3163, opened 2021-03-01
Update usage endpoint with billing details for orgs and services

CrystalPea:
So that when we build the billing report, we get the new details:
- purchase order number
- billing contact names
- billing contact email addresses
- billing reference

Also only return services with bills to pay.

Pivotal ticket: https://www.pivotaltracker.com/story/show/176521236

From the ticket:
> In order to determine where to fetch these values from, first check if there is a ‘purchase order number’ present for the service. If there is then set these fields using the values stored against the service.

> If there is no purchase order number on the service, then check if there is a purchase order number on the parent organisation. If there is then set these fields using the values stored against the organisation.

> If neither is present leave those fields empty in the report.

How to review:
review the whole PR together.

Next steps:
Use this additional data to update the billing report we create in admin app.

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -683,3 +683,20 @@ def fetch_usage_year_for_organisation(organisation_id, year):
         service_with_usage[str(email_usage.service_id)]['emails_sent'] = email_usage.emails_sent
 
     return service_with_usage
+
+
+def fetch_billing_details_for_all_services():
+    billing_details = db.session.query(
+        Service.id.label('service_id'),
+        func.coalesce(Service.purchase_order_number, Organisation.purchase_order_number).label('purchase_order_number'),
+        func.coalesce(Service.billing_contact_names, Organisation.billing_contact_names).label('billing_contact_names'),
+        func.coalesce(
+            Service.billing_contact_email_addresses,
+            Organisation.billing_contact_email_addresses
+        ).label('billing_contact_email_addresses'),
+        func.coalesce(Service.billing_reference, Organisation.billing_reference).label('billing_reference'),
+    ).outerjoin(
+        Service.organisation
+    ).all()
+
+    return billing_details
--- a/app/platform_stats/rest.py
+++ b/app/platform_stats/rest.py
@@ -59,7 +59,8 @@ def validate_date_range_is_within_a_financial_year(start_date, end_date):
 
 
 @platform_stats_blueprint.route('usage-for-all-services')
-def get_usage_for_all_services():
+@platform_stats_blueprint.route('data-for-billing-report')
+def get_data_for_billing_report():
     start_date = request.args.get('start_date')
     end_date = request.args.get('end_date')
 
--- a/app/platform_stats/rest.py
+++ b/app/platform_stats/rest.py
@@ -77,17 +77,18 @@ def get_data_for_billing_report():
     ]
     combined = {}
     for s in sms_costs:
-        entry = {
-            "organisation_id": str(s.organisation_id) if s.organisation_id else "",
-            "organisation_name": s.organisation_name or "",
-            "service_id": str(s.service_id),
-            "service_name": s.service_name,
-            "sms_cost": float(s.sms_cost),
-            "sms_fragments": s.chargeable_billable_sms,
-            "letter_cost": 0,
-            "letter_breakdown": ""
-        }
-        combined[s.service_id] = entry
+        if float(s.sms_cost) > 0:
+            entry = {
+                "organisation_id": str(s.organisation_id) if s.organisation_id else "",
+                "organisation_name": s.organisation_name or "",
+                "service_id": str(s.service_id),
+                "service_name": s.service_name,
+                "sms_cost": float(s.sms_cost),
+                "sms_fragments": s.chargeable_billable_sms,
+                "letter_cost": 0,
+                "letter_breakdown": ""
+            }
+            combined[s.service_id] = entry
 
     for letter_cost in letter_costs:
         if letter_cost.service_id in combined:
--- a/app/platform_stats/rest.py
+++ b/app/platform_stats/rest.py
@@ -4,6 +4,7 @@ from flask import Blueprint, jsonify, request
 
 from app.dao.date_util import get_financial_year_for_datetime
 from app.dao.fact_billing_dao import (
+    fetch_billing_details_for_all_services,
     fetch_letter_costs_for_all_services,
     fetch_letter_line_items_for_all_services,
     fetch_sms_billing_for_all_services,
@@ -105,12 +106,24 @@ def get_usage_for_all_services():
     for service_id, breakdown in lb_by_service:
         combined[service_id]['letter_breakdown'] += (breakdown + '\n')
 
+    billing_details = fetch_billing_details_for_all_services()
+    for service in billing_details:
+        if service.service_id in combined:
+            combined[service.service_id].update({
+                    'purchase_order_number': service.purchase_order_number,
+                    'contact_names': service.billing_contact_names,
+                    'contact_email_addresses': service.billing_contact_email_addresses,
+                    'billing_reference': service.billing_reference
+                })
+
     # sorting first by name == '' means that blank orgs will be sorted last.
-    return jsonify(sorted(combined.values(), key=lambda x: (
+
+    result = sorted(combined.values(), key=lambda x: (
         x['organisation_name'] == '',
         x['organisation_name'],
         x['service_name']
-    )))
+    ))
+    return jsonify(result)
 
 
 def postage_description(postage):
--- a/tests/app/dao/test_ft_billing_dao.py
+++ b/tests/app/dao/test_ft_billing_dao.py
@@ -600,62 +600,111 @@ def test_fetch_sms_billing_for_all_services_with_remainder(notify_db_session):
 
 
 def test_fetch_sms_billing_for_all_services_without_an_organisation_appears(notify_db_session):
-    org, org_2, service, service_2, service_3, service_sms_only, \
-        org_with_emails, service_with_emails = set_up_usage_data(datetime(2019, 5, 1))
+    setup = set_up_usage_data(datetime(2019, 5, 1))
     results = fetch_sms_billing_for_all_services(datetime(2019, 5, 1), datetime(2019, 5, 31))
 
     assert len(results) == 2
     # organisation_name, organisation_id, service_name, service_id, free_sms_fragment_limit,
     # sms_rate, sms_remainder, sms_billable_units, chargeable_billable_units, sms_cost
-    assert results[0] == (org.name, org.id, service.name, service.id, 10, Decimal('0.11'), 8, 3, 0, Decimal('0'))
-    assert results[1] == (None, None, service_sms_only.name, service_sms_only.id, 10, Decimal('0.11'),
-                          0, 3, 3, Decimal('0.33'))
+    assert results[0] == (
+        setup["org_1"].name,
+        setup["org_1"].id,
+        setup["service_1_sms_and_letter"].name,
+        setup["service_1_sms_and_letter"].id,
+        10, Decimal('0.11'), 8, 3, 0, Decimal('0'))
+    assert results[1] == (
+        None,
+        None,
+        setup["service_with_sms_without_org"].name,
+        setup["service_with_sms_without_org"].id,
+        10, Decimal('0.11'), 0, 3, 3, Decimal('0.33')
+    )
 
 
 def test_fetch_letter_costs_for_all_services(notify_db_session):
-    org, org_2, service, service_2, service_3, service_sms_only, \
-        org_with_emails, service_with_emails = set_up_usage_data(datetime(2019, 6, 1))
+    setup = set_up_usage_data(datetime(2019, 6, 1))
 
     results = fetch_letter_costs_for_all_services(datetime(2019, 6, 1), datetime(2019, 9, 30))
 
     assert len(results) == 3
-    assert results[0] == (org.name, org.id, service.name, service.id, Decimal('3.40'))
-    assert results[1] == (org_2.name, org_2.id, service_2.name, service_2.id, Decimal('14.00'))
-    assert results[2] == (None, None, service_3.name, service_3.id, Decimal('24.45'))
+    assert results[0] == (
+        setup["org_1"].name, setup["org_1"].id,
+        setup["service_1_sms_and_letter"].name, setup["service_1_sms_and_letter"].id,
+        Decimal('3.40')
+    )
+    assert results[1] == (
+        setup["org_for_service_with_letters"].name, setup["org_for_service_with_letters"].id,
+        setup["service_with_letters"].name, setup["service_with_letters"].id,
+        Decimal('14.00')
+    )
+    assert results[2] == (
+        None, None,
+        setup["service_with_letters_without_org"].name, setup["service_with_letters_without_org"].id,
+        Decimal('24.45')
+    )
 
 
 def test_fetch_letter_line_items_for_all_service(notify_db_session):
-    org_1, org_2, service_1, service_2, service_3, service_sms_only, \
-        org_with_emails, service_with_emails = set_up_usage_data(datetime(2019, 6, 1))
+    setup = set_up_usage_data(datetime(2019, 6, 1))
 
     results = fetch_letter_line_items_for_all_services(datetime(2019, 6, 1), datetime(2019, 9, 30))
 
     assert len(results) == 7
-    assert results[0] == (org_1.name, org_1.id, service_1.name, service_1.id, Decimal('0.45'), 'second', 6)
-    assert results[1] == (org_1.name, org_1.id, service_1.name, service_1.id, Decimal("0.35"), 'first', 2)
-    assert results[2] == (org_2.name, org_2.id, service_2.name, service_2.id, Decimal("0.65"), 'second', 20)
-    assert results[3] == (org_2.name, org_2.id, service_2.name, service_2.id, Decimal("0.50"), 'first', 2)
-    assert results[4] == (None, None, service_3.name, service_3.id, Decimal("0.35"), 'second', 2)
-    assert results[5] == (None, None, service_3.name, service_3.id, Decimal("0.50"), 'first', 1)
-    assert results[6] == (None, None, service_3.name, service_3.id, Decimal("1.55"), 'international', 15)
+    assert results[0] == (
+        setup["org_1"].name, setup["org_1"].id,
+        setup["service_1_sms_and_letter"].name, setup["service_1_sms_and_letter"].id,
+        Decimal('0.45'), 'second', 6
+    )
+    assert results[1] == (
+        setup["org_1"].name, setup["org_1"].id,
+        setup["service_1_sms_and_letter"].name, setup["service_1_sms_and_letter"].id,
+        Decimal("0.35"), 'first', 2
+    )
+    assert results[2] == (
+        setup["org_for_service_with_letters"].name, setup["org_for_service_with_letters"].id,
+        setup["service_with_letters"].name, setup["service_with_letters"].id,
+        Decimal("0.65"), 'second', 20
+    )
+    assert results[3] == (
+        setup["org_for_service_with_letters"].name, setup["org_for_service_with_letters"].id,
+        setup["service_with_letters"].name, setup["service_with_letters"].id,
+        Decimal("0.50"), 'first', 2
+    )
+    assert results[4] == (
+        None, None,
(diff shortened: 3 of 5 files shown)


app/dao/fact_billing_dao.py
```diff
@@ -684,3 +684,21 @@ def fetch_usage_year_for_organisation(organisation_id, year):
         service_with_usage[str(email_usage.service_id)]['emails_sent'] = email_usage.emails_sent
 
     return service_with_usage
+
+
+def fetch_billing_details_for_all_services():
+    billing_details = db.session.query(
```
benthorner, line 690: I think we can simplify this query, since we're already doing a join to just get the whole service and organisation - it's not a big deal memory-wise, since we're only talking about a few thousand objects. Then we could just have:

```
db.session.query(Service).options(db.joinedload(Service.organisation))
```

so that in the code we can just do:

```
if service.purchase_order_number:
   ... (use service.purchase_order_number)
  else:
   ... (use service.organisation.purchase_order_number)
```

We could even DRY it up more with this technique, noting the column names are the same in both cases, by doing something like:

```
billing_details = service if service.purchase_order_number else organisation
```

tests/app/platform_stats/test_rest.py
```diff
@@ -122,7 +122,7 @@ def test_validate_date_is_within_a_financial_year_when_input_is_not_a_date(start
 
 
 def test_get_usage_for_all_services(notify_db_session, admin_request):
-    org, org_2, service, service_2, service_3, service_sms_only, \
+    org, org_3, service, service_3, service_without_org, service_sms_only, \
```
benthorner, line 125: It took me a while to understand the change here - are we trying to align with the returned names ([here](https://github.com/alphagov/notifications-api/blob/6d9a505ba65679d81cde5fa9607e908a7cdd2985/tests/app/db.py#L1019))? It's surprising to have `org` and `org_3`, but no `org_2`. Could we:

- Add some explanation for the change in a commit message.
- Change the names so they exactly match the returned names.

Using indices to represent different things is a bit of a smell, but that seems a bit out-of-scope to improve in this PR. If we did want to make a small improvement, we could:

- Rename `service_x` with something more distinctive, like `service_with_email`.
- Stop returning orgs, and just fetch them from a service when they're needed.

tests/app/platform_stats/test_rest.py
```diff
@@ -134,23 +134,35 @@ def test_get_usage_for_all_services(notify_db_session, admin_request):
...
+    assert response[2]["contact_names"] == "sms billing contact names"
+    assert response[2]["contact_email_addresses"] == "sms@billing.contact email@addresses.gov.uk"
+    assert response[2]["billing_reference"] == "sms billing reference"
 
     assert response[3]["organisation_id"] == ""
-    assert response[3]["service_id"] == str(service_3.id)
+    assert response[3]["service_id"] == str(service_without_org.id)
     assert response[3]["sms_cost"] == 0
     assert response[3]["sms_fragments"] == 0
     assert response[3]["letter_cost"] == 24.45
```
benthorner, line 168: Minor: we could also assert e.g. `purchase_order_number` is null.

benthorner (commented):
Nice one catching the edge cases in the test 🔢. I've left a couple of more major comments - one of the hints at a more general thing: could we squash the commits before merging?

tests/app/platform_stats/test_rest.py
CrystalPea: done ☺️

app/dao/fact_billing_dao.py
CrystalPea: I like how short and sweet the query would become. Do you think it would also be performant if we had 30 000 services?

tests/app/platform_stats/test_rest.py
CrystalPea: I did some refactoring to make this fixture and it's outputs more readable, and also to stop variable assignment being based on returns order.

app/dao/fact_billing_dao.py
benthorner: Yes, it should be. In my experience, there are two important concerns with DB queries:

- Number of records, as an order of magnitude. Eventually this will exceed available memory and we will need to batch the query. Asking for a subset of columns just buys us a bit more time.

- Query complexity (e.g. inefficient indices). We're not doing anything fancy here, though.

You make a good point about the future performance. Since this is only an internal feature, and the number of services grows slowly, I don't think it's worth the effort to optimise this just yet.

So I think it's safe to make the change I'm suggesting.

What do you think?

app/platform_stats/rest.py
```diff
@@ -102,12 +106,32 @@ def get_usage_for_all_services():
     for service_id, breakdown in lb_by_service:
         combined[service_id]['letter_breakdown'] += (breakdown + '\n')
 
+    billing_details = fetch_billing_details_for_all_services()
+    for service in billing_details:
+        if service.service_id in combined:
```
servingUpAces, line 111: Can this be simplified by removing the if statement and just use:
```
combined[service.service_id].update({
   'purchase_order_number': service.service_purchase_order_number or service.organisation_purchase_order_number,
...
})```

File: app/platform_stats/rest.py [hunk 5]
```diff
@@ -75,17 +75,18 @@ def get_data_for_billing_report():
...
-            "organisation_name": s.organisation_name or "",
-            "service_id": str(s.service_id),
-            "service_name": s.service_name,
-            "sms_cost": float(s.sms_cost),
-            "sms_fragments": s.chargeable_billable_sms,
-            "letter_cost": 0,
-            "letter_breakdown": ""
-        }
-        combined[s.service_id] = entry
+        if float(s.sms_cost) > 0:
```
benthorner, line 78: I think this is new since my original review. Could we DRY it up to avoid setting things like `organisation_id` in two places?

One way to do this would be to define a function to return the common `dict` between the two loops e.g. `initial_report_row_for_service`.

Another way would be to rewrite the three loops by doing the report in two steps:

1. "aggregation". First loop over the three arrays to generate a dict of the form `service_id => { "sms_cost" => ..., "letter_cost" => ..., "billing_details" => ...}`

2. "generation". Loop of the dict to generate the row for each service, depending on the keys in its dict (i.e. whether it has a `letter_cost` and/or `sms_cost`).

What do you think?

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -609,7 +609,7 @@ def test_fetch_sms_billing_for_all_services_without_an_organisation_appears(noti
     setup = set_up_usage_data(datetime(2019, 5, 1))
     results = fetch_sms_billing_for_all_services(datetime(2019, 5, 1), datetime(2019, 5, 31))
 
-    assert len(results) == 2
+    assert len(results) == 3
```
benthorner, line 612: How come this changed here (in the last commit)?

benthorner (commented):
> I did some refactoring to make this fixture and it's outputs more readable, and also to stop variable assignment being based on returns order.

Thanks, that's looking better 👍. I've gone through again and left some comments on the changes, and a new refactoring suggestion on one of the commits that I believe is new.

That `set_up_usage_data` fixture is a beast! Tricky one to refactor. I was wondering if we really need one method, or if it could be split into separate fixtures? I think this is a relevant question to ask in this PR, as we've added yet another service to the output. Do we need all the usage data in all the tests?

tests/app/dao/test_ft_billing_dao.py
CrystalPea: It's because I added extra item to the fixture (service with sms within allowance)

tests/app/platform_stats/test_rest.py
```diff
@@ -121,11 +121,15 @@ def test_validate_date_is_within_a_financial_year_when_input_is_not_a_date(start
...
-    response = admin_request.get("platform_stats.get_usage_for_all_services",
-                                 start_date='2019-05-01',
-                                 end_date='2019-06-30')
+    response = admin_request.get(
+        "platform_stats.get_data_for_billing_report",
+        start_date='2019-05-01',
+        end_date='2019-06-30'
+    )
+
+    # we set up 5 services, but only 4 returned. service_with_emails was skipped as it had no bills to pay
```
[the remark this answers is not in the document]
CrystalPea: Yay ☺️

app/platform_stats/rest.py
```diff
@@ -57,7 +57,8 @@ def validate_date_range_is_within_a_financial_year(start_date, end_date):
 
 
 @platform_stats_blueprint.route('usage-for-all-services')
-def get_usage_for_all_services():
+@platform_stats_blueprint.route('data-for-billing-report')
```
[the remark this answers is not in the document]
CrystalPea: I kinda feel billing report is this plus the formatting done in the admin app. And this would be data for it 🤷🏼 

As to first question, yes, if you look at admin app PR, you will see that I changed the url there ☺️

app/dao/fact_billing_dao.py
CrystalPea: > I don't think it's worth the effort to optimise this just yet

But the effort has been made already, no? It would actually be extra effort to make it shorter but slightly less performant 🤔 

I like the aesthetics of what you propose, but kinda struggle justifying potentially running into problems a bit quicker in the future so that we could have a bit simpler code now.

I think what I would like most is a third opinion from someone like @servingUpAces . That would put my mind at peace :)

benthorner: Sure, a third opinion might be easiest 👍.

I think we should be careful about performance: normally it's better to start with readable code, and only optimise it when we have data to show this is necessary. That's because optimised code is often less readable. Without the data, it's hard to know what to optimise, and by how much - so called "premature optimisation".

I might be missing something, though: do we know there's a performance issue already?

app/platform_stats/rest.py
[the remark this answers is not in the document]
benthorner: Fair enough. I don't feel strongly about it, so let's leave it as-is 👍.

app/platform_stats/rest.py
CrystalPea: Not sure, what would happen if service_id is not in combined? Since not all services have bills to pay? 
Wouldn't that break?

app/platform_stats/rest.py
CrystalPea: I just added a new conditional (`if float(s.sms_cost`). The rest is not added by me, it just has different indentation. If you feel like refactoring this bit of code, feel free to add a commit for it to this PR, and I could review that commit then? :)

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -606,62 +606,111 @@ def test_fetch_sms_billing_for_all_services_with_remainder(notify_db_session):
...
-    assert results[1] == (org_2.name, org_2.id, service_2.name, service_2.id, Decimal('14.00'))
-    assert results[2] == (None, None, service_3.name, service_3.id, Decimal('24.45'))
+    assert results[0] == (
+        setup["org_1"].name, setup["org_1"].id,
+        setup["service_1_sms_and_letter"].name, setup["service_1_sms_and_letter"].id,
+        Decimal('3.40')
+    )
+    assert results[1] == (
+        setup["org_for_service_with_letters"].name, setup["org_for_service_with_letters"].id,
+        setup["service_with_letters"].name, setup["service_with_letters"].id,
```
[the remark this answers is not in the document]
CrystalPea: I think it was named wrong then, service_with_emails should have only emails, this service definitely has letters.

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -606,62 +606,111 @@ def test_fetch_sms_billing_for_all_services_with_remainder(notify_db_session):
...
+        Decimal('14.00')
+    )
+    assert results[2] == (
+        None, None,
+        setup["service_with_letters_without_org"].name, setup["service_with_letters_without_org"].id,
```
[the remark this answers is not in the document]
CrystalPea: hmm no, we have two services with letters, one with and one without an org, this is the service with letters without an org, as can be seen from first two items in results (org name and org id) being empty.

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -606,62 +606,111 @@ def test_fetch_sms_billing_for_all_services_with_remainder(notify_db_session):
...
+        Decimal('0.45'), 'second', 6
+    )
+    assert results[1] == (
+        setup["org_1"].name, setup["org_1"].id,
+        setup["service_1_sms_and_letter"].name, setup["service_1_sms_and_letter"].id,
+        Decimal("0.35"), 'first', 2
+    )
+    assert results[2] == (
+        setup["org_for_service_with_letters"].name, setup["org_for_service_with_letters"].id,
+        setup["service_with_letters"].name, setup["service_with_letters"].id,
```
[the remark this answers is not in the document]
CrystalPea: answer as above.

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -606,62 +606,111 @@ def test_fetch_sms_billing_for_all_services_with_remainder(notify_db_session):
 
 
 def test_fetch_sms_billing_for_all_services_without_an_organisation_appears(notify_db_session):
-    org, org_2, service, service_2, service_3, service_sms_only, \
-        org_with_emails, service_with_emails = set_up_usage_data(datetime(2019, 5, 1))
+    setup = set_up_usage_data(datetime(2019, 5, 1))
```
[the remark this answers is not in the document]
CrystalPea: I was thinking and could not get the right word. `services` is more descriptive, but as you say, it is not correct, and changing the fixture just for the name to fit feels like doing things wrong way round. 

Also I did not want to make the name too long as that would also impact readability, with the long service and org names. In the end I thought since in the results we always call out the key names which are very descriptive, the name of the dict is less important as long as it's short and sweet.

app/dao/fact_billing_dao.py
CrystalPea: No, we do not know of any performance issues here at the moment, and I guess this is an operation we only perform sporadically.

tests/app/dao/test_ft_billing_dao.py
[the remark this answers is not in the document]
CrystalPea: Now I remember when changing this, I saw names for a few services were not correct in the fixture, so I changed them to what they really are, and grouped billing records by each service, so it is easier to see what that service does.

Feel free to go and check in the fixture ☺️

tests/app/dao/test_ft_billing_dao.py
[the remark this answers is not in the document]
benthorner: > changing the fixture just for the name to fit feels like doing things wrong way round

How would it be the wrong way round?

I realise you've done some work already to make this fixture a bit better, so I'm OK with keeping it as-is. I don't think `setup` is a good name, though, as it's not a term we use elsewhere and it also means other things in Python. If you'd like to keep the fixture as is, then I'd suggest a name like `fixtures` is more descriptive for what the variable contains.

> I was wondering if we really need one method, or if it could be split into separate fixtures?

☝️ not sure if you saw this in the main comment. I think this fixture shouldn't exist, as it does way too many things. But it would also be more work to unpick it for the different tests, so maybe something for the future.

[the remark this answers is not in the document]
CrystalPea: I can change name to fixtures, the rest I feel is out of scope for this PR ☺️

tests/app/dao/test_ft_billing_dao.py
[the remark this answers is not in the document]
benthorner: Ah yes, that's really confusing how the names are totally different in the two places. I can see what you mean now, so I'll go through and resolve the other comments. Thanks for explaining the difference.

I did find this commit quite hard to review, due to the number of variables being changed. As a future suggestion (although hopefully we don't have other code like this!), I think it would have been easier to review as two commits: one to make the variable names consistent everywhere; and then anther to change them to a hash.

app/platform_stats/rest.py
benthorner: Oh yes, I see now. That's fair to leave it as-is.

I wouldn't want to add a commit to this PR, as that would make the review situation ambiguous. I'd be up for it as a separate PR if I were making the original change. I don't feel strongly enough about it to pick it up myself.

tests/app/dao/test_ft_billing_dao.py
benthorner: 🙈 don't mind me - I think I must have misread something!

app/dao/fact_billing_dao.py
CrystalPea: Ahh, that is true, I guess we could get away with not checking if organisation exists if we filtered to only get live services, as all live services have orgs and only live services have bills to pay.

Still argument remains for restricting queries to return only what we need to save resources. That does sound more eco-friendly.

benthorner: Hmm...I can see the argument both ways. Thanks for your thoughts @servingUpAces. I'm not sure about _always_ restricting queries, as that would make all our code a lot more verbose.

On the other hand, I can see the advantage of being explicit in this case: it's not obvious that the two tables both have columns with exactly the same names, so can be used interchangeably for billing. So maybe it's better to pull out the columns explicitly, so that someone in the future doesn't need to figure that out.

I'm coming round to leaving it as-is 👍.

> Otherwise you'd have to check that organisation exists before calling service.organisation.purchase order.

That's a good spot!

So we'd need a new conditional for the case where a service didn't have an organisation or its own billing details.

servingUpAces: Actually I forgot to mention you could use the coalesce function in the select.
`func.coalesce(Service.purchase_order_number, Organisation.purchase_order_number).label('purchase_order_number'),`

app/platform_stats/rest.py
CrystalPea: Ahh yes, that would be nicer, but then I think it would not be exactly as requested in the ticket:

> In order to determine where to fetch these values from, first check if there is a ‘purchase order number’ present for the service. If there is then set these fields using the values stored against the service.

> If there is no purchase order number on the service, then check if there is a purchase order number on the parent organisation. If there is then set these fields using the values stored against the organisation.

> If neither is present leave those fields empty in the report

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -606,62 +606,111 @@ def test_fetch_sms_billing_for_all_services_with_remainder(notify_db_session):
...
     results = fetch_sms_billing_for_all_services(datetime(2019, 5, 1), datetime(2019, 5, 31))
 
     assert len(results) == 2
     # organisation_name, organisation_id, service_name, service_id, free_sms_fragment_limit,
```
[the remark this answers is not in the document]
CrystalPea: hmm yes, but if we do not list all fields in the comment, how do we know the order? I'd leave it.

app/dao/fact_billing_dao.py
benthorner: Oooh, that would be a great compromise!

If we did that, could we get rid of the service vs. org conditional?

CrystalPea: That is a cool trick, but not sure it would work for the requirements from the ticket:
```
In order to determine where to fetch these values from, first check if there is a 'purchase order number' present for the service. If there is then set these fields using the values stored against the service.
If there is no purchase order number on the service, then check if there is a purchase order number on the parent organisation. If there is then set these fields using the values stored against the organisation.
If neither is present leave those fields empty in the report.
```

CrystalPea: I asked @yahoopete about this, so let's wait and see.

benthorner: @CrystalPea how would it not match the behaviour described in the ticket?

I had a look and it looks like it should behave the same.

tests/app/dao/test_ft_billing_dao.py
benthorner: Ah wait, I did read it right: I think this should be in the _previous_ commit "Billing report only has services with bills to pay". It confused me when I first read it, but I'm used to it now, so I think it's OK to leave it as-is.

app/dao/fact_billing_dao.py
benthorner: Ah yes, I see what you mean. Not quite the utopia I imagine.

(Arguably this makes it the worst option, as the behaviour could be unpredictable 🙈 . Oh well, at least now we're all really familiar with this code!)

benthorner (approved):
Thanks for the effort on the comments - the billing tests are now much easier to understand with the new fixture 🤩. I agree with you that we've done enough extra work on them in this PR - we can always refactor them more if we need to work on them again in future.

The query discussion turned out to be much more interesting than I expected, and I liked the different perspectives we had on it. I can see the compromise more clearly now.

PR #3167, opened 2021-03-04
Add service versioning to broadcast account type change

idavidmcdonald:
We are using the `set_broadcast_service_type` route to make changes to
service objects. However, we had forgotten to add the `version_class`
decorator to it which will mean the changing of a service going from
training to live mode will also be recorded in the services_history
table for free. Whilst not essential, this easy change makes things more
consistent for how we update other services.

Have tested locally by making a service a broadcast service and it works
image

### DIFF
--- a/app/dao/broadcast_service_dao.py
+++ b/app/dao/broadcast_service_dao.py
@@ -3,11 +3,19 @@ from datetime import datetime
 from flask import current_app
 
 from app import db
-from app.models import ServiceBroadcastSettings, ServicePermission, Organisation, BROADCAST_TYPE, EMAIL_AUTH_TYPE
-from app.dao.dao_utils import transactional
+from app.dao.dao_utils import transactional, version_class
+from app.models import (
+    Service,
+    ServiceBroadcastSettings,
+    ServicePermission,
+    Organisation,
+    BROADCAST_TYPE,
+    EMAIL_AUTH_TYPE
+)
 
 
 @transactional
+@version_class(Service)
 def set_broadcast_service_type(service, service_mode, broadcast_channel, provider_restriction):
     insert_or_update_service_broadcast_settings(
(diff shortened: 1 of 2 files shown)


servingUpAces (commented):
This is good and I tested it. I'd like to see a unit test to make sure the new history row is created.
You can access services_history using something like `Services.get_history_model().query().()`
I also noticed that the unit test `test_set_as_broadcast_service_gives_broadcast_permission_and_removes_other_permissions` does not test that permissions is removed because the sample service starts without any permissions.
One question unrelated to this PR; Why do we need to set the organisation in `set_broadcast_service_type`? The organisation is not part of the change set.

idavidmcdonald:
Thanks Becca. I've pushed up a squashed change which adds the test.


> I also noticed that the unit test test_set_as_broadcast_service_gives_broadcast_permission_and_removes_other_permissions does not test that permissions is removed because the sample service starts without any permissions.

I might have misunderstood you but I think this is now defunct as `test_set_as_broadcast_service_gives_broadcast_permission_and_removes_other_channel_permissions` tests three cases, each with different starting permissions. Is that right?

> Why do we need to set the organisation in set_broadcast_service_type? The organisation is not part of the change set.

I don't quite get what you mean by 'change set' but the reason we change the organisation is because all broadcast services are currently stored under a single organisation so we can find them easily. Does that answer your question?

servingUpAces:
By change set I mean the post data passed in by Admin app. I forgot about the one org for all broadcast services, maybe a comment in the code would help for future references.

servingUpAces (approved):

idavidmcdonald:
Ah yes, they all go in the broadcast org for the moment. There is a comment on the API route about this but not on this dao method
https://github.com/alphagov/notifications-api/blob/master/app/service/rest.py#L1091

PR #3168, opened 2021-03-04
If client reference not given, try to get it from personalisation

CrystalPea:
This is mostly useful for letters.

For templated letters sent via interface, whether one-offs
or CSV uploads, we do not give our users a way to set client reference.

Still, they often have a placeholder with reference that we could use
to set client_reference field.

Why is this helpful?

When letter is returned, or when we experience some printing issues,
often it is difficult to identify letters after the retention period.

This change will make it easier for some users to identify letters.
It will have more impact if we inform our users of this in template
editing guidance.

Alternatively, we could give people explicit ways to set reference in the frontend, but that would be much more work.


servingUpAces (commented):
This is only for templated letters sent from the UI. 
This will trigger for any notification without a client reference, we need to consider if it will affect the performance of post notifications.
Setting the client_reference should perhaps happen in the admin app or LetterTemplate rather than the highly used persist_notification method.

CrystalPea:
@servingUpAces good call regarding performance. I didn't think that simple conditional and variable assignment would take much, but I guess there would be a cumulative effect as we hit that place in the code millions of times each day. I think I could do it a bit differently, let's see :).

servingUpAces (approved):

PR #3169, opened 2021-03-05
add POST get user by email endpoint

leohemsted:
the existing endpoint is a GET, and so leaves email addresses in log files.

we've got an existing POST find_users_by_partial_email, but not one that matches on a whole email address.

### DIFF
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -417,6 +417,19 @@ def set_permissions(user_id, service_id):
     return jsonify({}), 204
 
 
+@user_blueprint.route('/email', methods=['POST'])
+def fetch_user_by_email():
+
+    email, errors = email_data_request_schema.load(request.get_json())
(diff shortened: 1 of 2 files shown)


app/user/rest.py
```diff
@@ -417,6 +417,20 @@ def set_permissions(user_id, service_id):
     return jsonify({}), 204
 
 
+@user_blueprint.route('/email', methods=['POST'])
+def fetch_user_by_email():
+
+    email, errors = email_data_request_schema.load(request.get_json())
+    if errors:
+        raise InvalidRequest(message=errors, status_code=400)
+    print(email)
```
idavidmcdonald, line 426: leftover print

leohemsted: squashed out

idavidmcdonald (approved):

PR #3171, opened 2021-03-08
New invite endpoints

leohemsted:
Review commit by commit.

Note: 

I didn't realise at the time, but the check invite token endpoint and the fetch invite details based on ID token currently share an endpoint.

The two HTTP requests look like the following
```
GET /invite/organisation/ImUyY2Y4ZmIyLTgxYjEtNDJjNS1iNDAzLTNhZDY4ZDM4MGQxMyI.YEZ5WA.V_EwdFF-1hrGkDlynbWNmNUMMOM
GET /invite/organisation/e2cf8fb2-81b1-42c5-b403-3ad68d380d13
```

Now, this doesn't break anything - one route is matching on strings while the other is matching on UUIDs, so it'll choose the tighter route match and select the correct endpoint.

However, that's a bit confusing and I don't love it. Does anyone have better suggestions for the route for "given an `invited_user.id` UUID, return the metadata for that invite such as what service it's for, who created it, what status it is in, and what permissions the user has been given by the invite?


leohemsted:
I think conceptually I'd prefer to migrate the existing endpoint to a new route.

`GET /invite/check/organisation/khKAHdkhasfkjhas.hdhsajkfhyagdYU  # checks signature validity, expiry time, etc`
`GET /invite/service/12341234-1234-1234-1234-123412341234  # returns metadata`

app/global_invite/rest.py
```diff
@@ -50,3 +50,15 @@ def validate_invitation_token(invitation_type, token):
...
+@global_invite_blueprint.route('/service/<uuid:invited_user_id>', methods=['GET'])
+def get_invited_user(invited_user_id):
+    invited_user = get_invited_user_by_id(invited_user_id)
+    return jsonify(data=invited_user_schema.dump(invited_user).data), 200
+
+
+@global_invite_blueprint.route('/organisation/<uuid:invited_org_user_id>', methods=['GET'])
+def get_invited_org_user(invited_org_user_id):
+    invited_user = dao_get_invited_organisation_user(invited_org_user_id)
+    return jsonify(data=invited_user.serialize()), 200
```
benthorner, line 64: Note to self: the different use of `serialize` vs `dump` here appears to be historical.

- The service user invite schema adds an `auth_type` foreign key to the dumped data (see [here](https://github.com/alphagov/notifications-api/blob/4cbd4c092090739ebfefc9d2b906bf8b39b213e4/app/schemas.py#L624)) and otherwise dumps all columns.
- The org user invite defines its own `serialize` method to specify the columns to dump in a dictionary.

Both perform validation of the email address - the org user invite has a specific schema for the posted data to achieve this (see [here](https://github.com/alphagov/notifications-api/blob/4cbd4c092090739ebfefc9d2b906bf8b39b213e4/app/organisation/invite_rest.py#L34)).

It looks like we've taken two approaches:

- Some endpoints have dedicated schemas, and a `serialize` method on the model.
- Other endpoints share a schema, which is reused for serialization.

@leohemsted just to help me understand a bit better: do you have any thoughts on which one of these approaches we prefer to use?

benthorner (commented):
Code all looks good - I've just left a tangential comment to try and improve my understanding. With the naming, I agree the new state is more confusing, so we should try to avoid introducing that here.

> I think conceptually I'd prefer to migrate the existing endpoint to a new route.

Yes, this makes sense to me. Let's add a secondary route and migrate to use that 👍.

As well as the URL structure, the `accept_invite -> global_invite` rename also confused me, as "global" isn't well defined. I can see why it has to be separate, since the org/service-specific blueprints all require an ID. I think just "invite" is OK, since it represents general functionality for invites. Another idea, which is more effort but may be appealing, would be to:

- Push the URL prefixes of the org/service invite blueprints onto the individual routes, so that we're free to...
- Move/refactor the token validation endpoint into those specific blueprints (with a similar URL structure).

The current `accept_invite` module isn't really generic, as its behaviour differs for services vs. orgs. So it makes sense to split it up, and would only cost a small amount of duplication. Or we could just wrangle it a suitable name.

What do you think?

app/global_invite/rest.py
leohemsted: hey, good question, there's a bit of legacy here.

the stuff in `schemas.py` is marshmallow, which we don't love (we're on an old version cos new version breaks, it's full of mystery and magic and surprises) and are slowly (glacially) moving away from that (and the `dump`) method towards `serialize` methods on the models when returning, and using json schemas to validate input data (note that we still need to do the deserialization ourselves, eg to datetimes or from str to UUID etc).

benthorner: Thanks for explaining, that's good to know. Sounds like the new way has some advantages, especially as the load vs. dump are less coupled (and we're not stuck on some ancient version of something).

Do you think it's worth documenting this [here](https://github.com/alphagov/notifications-manuals/wiki/Decision-records)?

app/organisation/invite_rest.py
```diff
@@ -3,16 +3,18 @@
...
 from app.config import QueueNames
 from app.dao.invited_org_user_dao import (
     save_invited_org_user,
-    get_invited_org_user,
+    get_invited_org_user as dao_get_invited_org_user,
     get_invited_org_users_for_organisation
 )
 from app.dao.templates_dao import dao_get_template_by_id
-from app.errors import register_errors
+from app.dao.organisation_dao import dao_get_invited_organisation_user
```
benthorner, line 16: Could we replace this, with [this](https://github.com/alphagov/notifications-api/blob/a2cc0df5a70b918bad27beea48cff55362c838e5/app/dao/invited_org_user_dao.py#L16), to avoid the name clash?

app/organisation/invite_rest.py
```diff
@@ -106,3 +112,32 @@ def invited_org_user_url(invited_org_user_id, invite_link_host=None):
         invite_link_host = current_app.config['ADMIN_BASE_URL']
 
     return '{0}/organisation-invitation/{1}'.format(invite_link_host, token)
+
+
+@organisation_invite_blueprint.route('/invite/organisation/<uuid:invited_org_user_id>', methods=['GET'])
```
benthorner, line 117: What do you think about migrating these routes to be more consistent with each other? I'm not sure if that matters much, although it may be confusing to someone reading this in future as to why they have different structures.

app/service_invite/rest.py
```diff
@@ -3,25 +3,30 @@
     request,
     jsonify,
     current_app)
+from itsdangerous import SignatureExpired, BadData
 
+
+from notifications_utils.url_safe_token import check_token, generate_token
+
+from app.dao.invited_user_dao import get_invited_user_by_id
```
benthorner, line 11: We don't need this one - it's the same function as the one below.

app/global_invite/rest.py
leohemsted: yeah, almost certainly. i'll write something up later

tests/app/service_invite/test_service_invite_rest.py
```diff
@@ -274,10 +274,14 @@ def test_update_invited_user_for_invalid_data_returns_400(client, sample_invited
     assert response.status_code == 400
 
 
-def test_validate_invitation_token_returns_200_when_token_valid(client, sample_invited_user):
+@pytest.mark.parametrize('endpoint_format_str', [
+    '/invite/service/{}',
+    '/invite/check/service/{}',
```
benthorner, line 279: Minor: the "check" bit would be easier to interpret for me if it was on the end i.e. `/invite/service/{}/check`. I think that's because the first part of the path is scoping the resource "invite / service", while "check" is the action we're performing on it.

benthorner (approved):
Thanks for working through my suggestions, especially the bigger one 💯.

I've left a couple of questions, and two small things I reckon we should address before this goes in.

app/organisation/invite_rest.py
leohemsted: sorry, I don't follow. Which route should it be more consistent with? The ones above where it's `/service/<uuid>/invite/<uuid>`?

We don't have a service ID here so cannot do that.

app/service_invite/rest.py
leohemsted: one uses a service id and one doesn't. the one that requires a service id is technically safer as it makes it less likely that a malicious user is able to invite themselves to another service by modfiying URLs etc.

they're not the same, but could be named better. i'll updated the one that requires a service id so the name is explicit.

tests/app/service_invite/test_service_invite_rest.py
leohemsted: hmmm, i'm not sure it should be on the end, largely because the token is a long variable length string and hard to find the end of (compared to a uuid which is fixed length). i think it's easy to miss detail like `/check` if it's at the end (like the `/edit` at the end of a google doc link. but that's just me. thoughts?

app/service_invite/rest.py
benthorner: Ahh, sorry. I assumed it was similar to https://github.com/alphagov/notifications-api/pull/3171/files#r593097701 and missed the `_by_id` bit on the end 👍.

tests/app/service_invite/test_service_invite_rest.py
leohemsted: we could have something like `check_token` in the URL to make it super explicit that it's a token, not a uuid, if that helps 🤔

app/organisation/invite_rest.py
benthorner: You're right they can't have exactly the same structure. I was looking at the beginning:

- Some routes start with `/invite/organisation`
- Others start with `/organisation`

I'd (perhaps naively) expect all the routes in a file to at least start with the same thing, which in this file should probably be `/invite/organisation`. I think it's a minor one, though, as we don't seem to have any guidance on URL structure. Still, this might be a good opportunity to make them consistent, since we're having to do change some of the other routes anyway. Up to you, though - I don't feel strongly about it.

tests/app/service_invite/test_service_invite_rest.py
benthorner: > the token is a long variable length string and hard to find the end of 

That's fair. How about a compromise: `/invite/service/check/{}`?

app/organisation/invite_rest.py
leohemsted: ah, gotcha. i see what you mean, but if we do that we sacrifice clarity across different organisation endpoints. 

for example we currently have:

```
GET        /organisations/<uuid:organisation_id>  # get organisation info
GET        /organisations/<uuid:organisation_id>/services  # get info on services for organisation
GET        /organisation/<uuid:organisation_id>/invite  # get info on invites for organisation
```

(please ignore the pluralisation of organisation aaaaaagh 🐰🕳️😱)

think this is a six/half dozen problem, we're never going to be able to align invites (both with a ID and without ID) and also align organisation endpoints at the same time. i'll leave it as is to avoid extra work for myself

app/service_invite/rest.py
leohemsted: fixed in 68d24e5

app/organisation/invite_rest.py
leohemsted: fixed in 68d24e5

app/organisation/invite_rest.py
benthorner: Sure, no worries. I can only sympathise 😸.

PR #3172, opened 2021-03-09
Lookup PDFs by prefix in S3 (not whole filename)

benthorner:
https://www.pivotaltracker.com/story/show/176823265

A couple of extra notes:

- This duplicates an existing precedent for getting files in S3 by prefix [1].
Unfortunately it would be quite hard to DRY up the new code, so I'm going to
leave it as-is unless someone feels strongly.

- While this makes it a moot point whether we remove the crown / non-crown part
of the filename, I still intent to pursue this in follow-up work, just so that we
have one less thing to remember the meaning of.

- This makes additional use of S3, so warrants a bit of performance testing just
to make sure there's no unexpected impact from the change. I'll do that before
merging the PR.

[1]: https://github.com/alphagov/notifications-api/blob/6db9a7a2f2b7d9ebb24cae7f8a976f6eb42b3a88/app/letters/utils.py#L169

### DIFF
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -69,22 +69,6 @@ def remove_contact_list_from_s3(service_id, contact_list_id):
     return remove_s3_object(*get_contact_list_location(service_id, contact_list_id))
 
 
-def get_s3_bucket_objects(bucket_name, subfolder=''):
-    boto_client = client('s3', current_app.config['AWS_REGION'])
-    paginator = boto_client.get_paginator('list_objects_v2')
-    page_iterator = paginator.paginate(
-        Bucket=bucket_name,
-        Prefix=subfolder
-    )
-
-    all_objects_in_bucket = []
-    for page in page_iterator:
-        if page.get('Contents'):
-            all_objects_in_bucket.extend(page['Contents'])
-
-    return all_objects_in_bucket
-
-
 def remove_s3_object(bucket_name, object_key):
     obj = get_s3_object(bucket_name, object_key)
     return obj.delete()
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -27,9 +27,10 @@ from app.errors import VirusScanError
 from app.exceptions import NotificationTechnicalFailureException
 from app.letters.utils import (
     ScanErrorType,
+    find_letter_pdf_filename,
+    generate_letter_pdf_filename,
     get_billable_units_for_letter_page_count,
     get_file_names_from_error_bucket,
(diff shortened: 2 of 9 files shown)


leohemsted (approved):

benthorner:
@leohemsted I've done a bit more work on this before merging:

- I walked myself through the process of sending a letter, as a learning exercise. Doing this revealed one of the places ([here](https://github.com/alphagov/notifications-api/pull/3172/commits/b43a367d5f3dc56cea89353ee807f0e0436625f3#diff-926badba91648d56a973e16bd92da3345b23bc60dc89360119b1df08de52723fR62)) we used "find_letter..." was incorrect - I've amended it to "generate_letter_..." in the first commit.

- There are only two places we use "find_letter..." (now), and both are scheduled tasks: collate and archive/retention. As a simple performance test, I ran the collate task on 10K letters with/out the change. Stats are:

   - Without: 375 seconds
   - With: 1086 seconds

This seems OK to me, as I don't believe the collation task has any major time constraints on it.

What do you think?

leohemsted:
hmmm, so we used to do one s3 request, now we do two, and that new request takes twice as long as an old request?

i think that's probably fine but something to think about and keep in mind. i never love making things take three times as long, especially tasks that could take a long time to run, run after we've left for the day, and might be a pain to restart if a box is rotated or crashes etc.

if we changed this new function to return the whole object metadata rather than just the filename, we could probs remove the second call to get metadata by filename from collate (and possibly some other places) 🤔 . don't think we need to do that right now, but if it's not too hard to change i think it'd be nice to do so. thoughts?

servingUpAces:
I'm hesitant to deploy something that makes the code more complicated and degrades performance, especially to solve a rare bug. Are we sure about this one?

benthorner:
@servingUpAces @leohemsted I agree it's more slow than I expected. I think that's because the extra operation is a search (by prefix), rather than a straightforward query by ID.

Now that I'm more confident with the code, I followed @leohemsted's suggestion - see the two new/latest commits. Testing the collate task with 10K letters in preview now comes out at 666 seconds.

@servingUpAces I also wasn't keen on the subtle "find_..._filename" vs "generate_..._filename" distinction. The new commits change the name and behaviour, so now they're much more distinct.

On the whole I reckon this is (now) a small, net positive change:

\+ +174 −195 LOC
\+ Removes some complexity (extra method calls)
\+ Fixes a bug
\- Adds some complexity (choice of find/generate)
\- Adds some time overhead

What would you like to do?

servingUpAces (approved):

PR #3173, opened 2021-03-10
Add an endpoint to return all the data required for the performance platform page

servingUpAces:
The performance platform is being decommissioned March 15. 
We have all the data in our db, this PR returns all the data needed to recreate the performance platform page on Notify.


tests/app/performance_platform/test_rest.py
```diff
@@ -0,0 +1,48 @@
...
+    results = admin_request.get(endpoint="performance_platform.get_performance_platform",
+                                start_date='2021-03-01',
+                                end_date='2021-03-01')
+
+    assert results['total_notifications'] == 15+20+3
+    assert results['email_notifications'] == 15
+    assert results['sms_notifications'] == 20
+    assert results['letter_notifications'] == 3
+    assert results['notifications_by_type'] == [{"date": '2021-03-01', "emails": 15, "sms": 20, "letters": 3}]
+    assert results['processing_time'] == [({"date": "2021-03-01", "percentage_under_10_seconds": 97.1})]
```
idavidmcdonald, line 36: Optional: Would be good to see multiple dates worth of data here as I don't think we have a test anywhere for that and I think it's most likely that we will be querying say a months worth of data always rather than a single day

idavidmcdonald, line 36: How come processing time is a list of sets containing dicts rather than just a list of dicts like `notifications_by_type` is?

app/performance_platform/rest.py
```diff
@@ -0,0 +1,85 @@
...
+        # Is it ok to reuse this? - should probably create a new one
+        validate(request.args, performance_platform_request)
+
+    # If start and end date are not set, we are expecting today's stats.
+    today = str(datetime.utcnow().date())
+
+    start_date = datetime.strptime(request.args.get('start_date', today), '%Y-%m-%d').date()
+    end_date = datetime.strptime(request.args.get('end_date', today), '%Y-%m-%d').date()
+    results = get_total_notifications_for_date_range(start_date=start_date, end_date=end_date)
+    total_notifications, emails, sms, letters = transform_results_into_totals(results)
```
idavidmcdonald, line 31: Ah, there is one major difference here actually I think.

On the performance platform, the headline figures for how many emails, texts and letters we have sent are for ALL notifications over time I believe so from the start of Notify (or at least us recording the figure). Whereas, when we want to show say the last month of data for notifications sent per day, we only query for the last 30 days of data. At the moment, how this is written means our big totals at the top of the page would also only show the total number sent in the last 30 days which is probably not what we want.

app/dao/fact_processing_time_dao.py
```diff
@@ -30,3 +30,18 @@ def insert_update_processing_time(processing_time):
         }
     )
     db.session.connection().execute(stmt)
+
+
+def get_processing_time_percentage_for_date_range(start_date, end_date):
+    query = db.session.query(
+        FactProcessingTime.bst_date.cast(db.Text).label("date"),
```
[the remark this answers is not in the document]
servingUpAces: Just the same label used for the json and the old page.

app/performance_platform/rest.py
servingUpAces: That's just a mistake of extra brackets

tests/app/performance_platform/test_rest.py
servingUpAces: That's just a mistake of extra brackets

tests/app/dao/test_fact_processing_time_dao.py
```diff
@@ -40,3 +42,21 @@ def test_insert_update_processing_time(notify_db_session):
...
+def test_get_processing_time_percentage_for_date_range(notify_db_session):
+    data = FactProcessingTime(
+        bst_date=datetime(2021, 2, 22).date(),
+        messages_total=3,
+        messages_within_10_secs=2
+    )
+
+    fact_processing_time_dao.insert_update_processing_time(data)
+
+    results = get_processing_time_percentage_for_date_range('2021-02-22', '2021-02-22')
```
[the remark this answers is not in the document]
idavidmcdonald: This comment still stands and also you could now use `create_process_time` to enable you to set up your data in this test

tests/app/dao/test_fact_notification_status_dao.py
```diff
@@ -703,3 +703,39 @@ def test_fetch_monthly_notification_statuses_per_service_for_rows_that_should_be
...
+
+    create_ft_notification_status(bst_date=date(2021, 3, 1),
+                                  service=template_email.service,
+                                  template=template_email,
+                                  count=15)
+    create_ft_notification_status(bst_date=date(2021, 3, 1),
+                                  service=template_sms.service,
+                                  template=template_sms,
+                                  count=20)
+    create_ft_notification_status(bst_date=date(2021, 3, 1),
```
idavidmcdonald, line 733: Optional: This is better with the data for the 2021/2/28 but we still aren't testing the upper bound on time. If you added in some data for 2021/3/2 then it would test this. Not the end of the world and I can survive with the gap in test coverage but thought I'd point it out

idavidmcdonald (approved):
I think there could still be a bit of improvement to test coverage but apart from that I think this is good to go. Thanks

app/performance_platform/rest.py
servingUpAces: Good spot - fixed

tests/app/dao/test_fact_notification_status_dao.py
servingUpAces: I've added more data to the test

PR #3175, opened 2021-03-10
Add isort to build and fix issues

benthorner:
This is consistent with the build for the -admin and -utils
repos, and we can benefit from standardised imports here too,
as well as removing redundant ones.

Note that the one aspect of the config I'm a bit unsure about
is the line length, which has introduced some verbosity in a few
places (by adding "\" line continuations).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,3 +7,13 @@ exclude = venv*,__pycache__,node_modules,cache,migrations,build,sample_cap_xml_d
 max-line-length = 120
 # W504 line break after binary operator
 extend_ignore=B306, W504
+
+
+[isort]
+line_length=80
+indent='    '
+multi_line_output=3
+known_third_party=notifications_utils,notifications_python_client
+known_first_party=app,tests
+include_trailing_comma=True
+use_parentheses=True
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,22 +1,30 @@
-import time
 import os
 import random
 import string
+import time
 import uuid
+from time import monotonic
 
 from celery import current_task
-from flask import _request_ctx_stack, request, g, jsonify, make_response, current_app, has_request_context
-from flask_sqlalchemy import SQLAlchemy as _SQLAlchemy
+from flask import (
+    _request_ctx_stack,
+    current_app,
+    g,
+    has_request_context,
+    jsonify,
+    make_response,
+    request,
+)
 from flask_marshmallow import Marshmallow
 from flask_migrate import Migrate
+from flask_sqlalchemy import SQLAlchemy as _SQLAlchemy
 from gds_metrics import GDSMetrics
 from gds_metrics.metrics import Gauge, Histogram
-from time import monotonic
-from notifications_utils.clients.zendesk.zendesk_client import ZendeskClient
-from notifications_utils.clients.statsd.statsd_client import StatsdClient
-from notifications_utils.clients.redis.redis_client import RedisClient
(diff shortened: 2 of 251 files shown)


setup.cfg
```diff
@@ -7,3 +7,12 @@ exclude = venv*,__pycache__,node_modules,cache,migrations,build,sample_cap_xml_d
 max-line-length = 120
 # W504 line break after binary operator
 extend_ignore=B306, W504
+
+
+[isort]
+line_length=80
```
CrystalPea, line 13: do we want line length of 80 when our usual limit is 120?

setup.cfg
```diff
@@ -7,3 +7,12 @@ exclude = venv*,__pycache__,node_modules,cache,migrations,build,sample_cap_xml_d
...
+indent='    '
+multi_line_output=3
```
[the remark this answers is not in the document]
CrystalPea: Ahh, found it: https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes/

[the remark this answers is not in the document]
CrystalPea: They are different sorting modes

setup.cfg
CrystalPea: I guess if we use the same in admin it should be alright

CrystalPea (approved):
Note: I have not reviewed all 251 files with sorted imports, but I gather we would see in functional tests if anything is not right.

leohemsted:
according to https://pycqa.github.io/isort/docs/configuration/options/ there's `use_parentheses` which uses brackets over trailing slashes when the line needs wrapping.

i'd prefer us to have that

so 

```
# too long
from app.really.long.import.path import really_long_import as really_long_alias

# isort currently does this, i don't like this
from app.really.long.import.path import \
    really_long_import as really_long_alias

# with use_parentheses we'll do this, i believe
from app.really.long.import.path import (
    really_long_import as really_long_alias
)
```

it's technically one more line, but it looks the same as our multiple-import lines, so i think it'll be easier to scan through. thoughts?

setup.cfg
leohemsted: we should have it as 120 in both i think.

quis: I chose 80 originally. I try to write code to 80 characters, but sometimes there’s a good reason to make lines a bit longer, which is why our flake8 limit is set to 120. Somewhere in this [excellent Raymond Hettinger talk](https://www.youtube.com/watch?v=wf-BqAjZb8M) he talks about not being overly-dogmatic about line length, at the expense of readability.

An automated tool can’t make the same judgement calls about when it’s sensible to exceed 80 characters, so I don’t think we should let it.

CrystalPea: I like 120 limit more, as it allows for more descriptive method and variable names. I think 80 chars limit was set as a standard when screens had much lower resolution, right? 

Still, here it's not that important to me since it's just imports, so I'm happy either way ☺️

benthorner: I agree with @quis. If we're going to have a limit, then it doesn't make much difference what it is, since the output is basically automatic.

benthorner:
> there's use_parentheses which uses brackets over trailing slashes when the line needs wrapping

Nice find @leohemsted. I think that's a helpful addition, so I've amended the first commit to include it, with the aim of applying it in our other repos where we have isort already. For me, there are a couple of advantages:

- Line continuations are (I think) uncommon in Python, so the result is less surprising.
- Importing more attrributes from the same module in future will have a smaller diff.

PR #3176, opened 2021-03-10
Refactor method for deciding the failure type

servingUpAces:
We only need to check what the status is for firetext SMS responses.

This PR is to clean up that code a little bit.

I had a thought that this code doesn't belong in notification_dao but perhaps here app.notifications.notifications_sms_callback.process_firetext_response or app.celery.process_sms_client_response_tasks.process_sms_client_response. That's a larger refactor, but is it worth it?

### DIFF
--- a/app/clients/sms/firetext.py
+++ b/app/clients/sms/firetext.py
@@ -21,6 +21,7 @@ firetext_responses = {
 
 firetext_codes = {
     # code '000' means 'No errors reported'
+    '000': {'status': 'temporary-failure', 'reason': 'No error reported'},
     '101': {'status': 'permanent-failure', 'reason': 'Unknown Subscriber'},
     '102': {'status': 'temporary-failure', 'reason': 'Absent Subscriber'},
     '103': {'status': 'temporary-failure', 'reason': 'Subscriber Busy'},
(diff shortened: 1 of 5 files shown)


benthorner (approved):
Nice one! I'd say it's a clear improvement, as someone who's just trawled through the codebase to try and understand what's going on here.

Things I've learned:

- This function is called for all kinds of notification
- But it only has an effect for SMS notifications sent through Firetext - this includes the second conditional, as explained [here](https://github.com/alphagov/notifications-api/commit/97088a4af1027dd75f7dda92458198a782a45b41).
- It only affects Firetext notifications because the response hashes happen to have different keys and structures, so we always get a `KeyError` in `get_message_status_and_reason_from_firetext_code` for other notifications and providers. The new code makes this much clearer!
- `status` initially means "status code from the provider response", but later means "new status for the notification".

The tricky bit about refactoring this further is that second conditional, which depends on how the status has changed over time. The first conditional looks like it would be easier to refactor, and I like the suggestion of moving it to the `process_sms_client_response_tasks` module, as that's where the notification status gets decided first.

I think this is good to go on it's own, though 👍.

servingUpAces:
Thank you for your comments @benthorner I'm going to merge this but try to do the larger refactor next week.

PR #3177, opened 2021-03-12
Remove performance platform code

idavidmcdonald:
You will want to review commit by commit but in short, this does two things

commit 1 - moves us storing into our DB the nightly processing stat from happening as part of `send-daily-performance-platform-stats` into it's own new task
commit 2 - deletes everything to do with the performance platform

One important thing is that the performance platform will be switched off on Monday morning, around 10am. This means that if as the code stands, we will run into a problem with `send-daily-performance-platform-stats`.

https://github.com/alphagov/notifications-api/blob/master/app/performance_platform/processing_time.py#L28-L33
You can see here that it first will call out to performance platform, and then it will save the stat in the DB. As the code stands, on Tuesday at 2am, the call out to performance platform will fail and the function will error and it will therefore not insert the processing data in the DB. This gives us a few options

1. merge this PR sometime after Tuesday at 2am and backfill any data missed from Tuesday onwards into the processing db table
2. merge this PR between Monday 2am and Tuesday 2am. Should be a clean swap over
3. merge this PR before Monday 2am and the performance platform will not be sent the latest stats between when the PR gets merged and Monday (impact is very low)
4. merge the first commit anytime before Tuesday at 2am and then merge the 2nd commit anytime separately after that, acknowledging that the calls out to performance platform will fail until we merge the 2nd commit and error but it has no real impact

I suggest we go for option 2 and merge this on Monday, should be easy to do and a clean swap over.

Also note, this PR means we need to go get RE to change some things in our cronitor config.


benthorner (approved):
Nice one extracting out the remaining performance code - the comments in the tests make it much easier to understand what's going on 🏅. I've got one more minor comment on the tests.

I've also done a few quick checks and I think you've also caught all the straggly bits of code related to the platform, except perhaps this tangential command: [load_processing_time_data](https://github.com/alphagov/notifications-api/blob/6dc73dbc411a167002683321a9a9ebcd6520a168/app/commands.py#L958). Do we still need it?

PR #3178, opened 2021-03-12
Add missing statsd timers to celery tasks

benthorner:
https://www.pivotaltracker.com/story/show/176823265

All other tasks in app/celery/*_tasks.py have timers on them. Some
of these timers will be useful to check before/after performance as
a way to reassure ourselves about the impact of [1].

[1]: https://github.com/alphagov/notifications-api/pull/3172

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -119,6 +119,7 @@ def update_billable_units_for_letter(self, notification_id, page_count):
 
 @notify_celery.task(name='collate-letter-pdfs-to-be-sent')
 @cronitor("collate-letter-pdfs-to-be-sent")
+@statsd(namespace="tasks")
 def collate_letter_pdfs_to_be_sent():
     """
     Finds all letters which are still waiting to be sent to DVLA for printing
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):
so this is fine for us to do, but be aware that we don't currently have any grafana dashboards for this. To get any benefit from this change, you will likely need to also do https://www.pivotaltracker.com/story/show/176261229. Up for the challenge? Could be worth timeboxing if it looks like you would be going down a rabbit hole...

benthorner:
> we don't currently have any grafana dashboards for this

Yes, I'm aware of that. For the purpose of the load testing I'd like to do, I was going to just do some manual PromQL queries to get the before/after stats I'm interested in - might create a temporary dashboard if that easier.

> Up for the challenge?

Maybe 😉. I actually want to have a chat with you about something related to this, which I'll ping you about later. So let's see where we go from there...

idavidmcdonald:
Note, these metrics may or may not be coming through to prometheus. They go into prometheus via a statsd exporter which may or not be exposing these metrics to be collected by prometheus. You may need to make a change to the statsd configuration to be able to run some promql on these metrics

benthorner:
Yeah, I had fun figuring out how the conversion happens yesterday!

Fortunately, no action needed, as the new timers are covered by [this mapping](https://github.com/alphagov/notifications-aws/blob/master/paas/statsd/statsd-mapping.yml#L41).

idavidmcdonald:
Sweet, if that is the case then if you are going to go to the effort or writing some temporary queries, then going the full way and adding the dashboard would be great. But I'll leave it to you to decide what the best use of your time is :)

PR #3179, opened 2021-03-16
Add international API key for firetext

servingUpAces:
We want to start using Firetext for sending international SMS. They
require us to use a different API key for international SMS because it
requires a new code path to switch the sender ID to something that the
country will accept.
This PR does not include switching the sender of international SMS to
Firetext but sets us up to do so.


tests/app/clients/test_firetext.py
```diff
@@ -47,7 +47,7 @@ def test_send_sms_successful_returns_firetext_response(mocker, mock_firetext_cli
 
     with requests_mock.Mocker() as request_mock:
         request_mock.post('https://example.com/firetext', json=response_dict, status_code=200)
-        response = mock_firetext_client.send_sms(to, content, reference)
+        response = mock_firetext_client.send_sms(to, content, reference, False)
```
benthorner, line 50:
```suggestion
        response = mock_firetext_client.send_sms(to, content, reference, international=False)
```

Minor: this should make it easier for someone in future (avoids them having to find and read the function signature).

tests/app/clients/test_firetext.py
```diff
@@ -79,6 +79,30 @@ def test_send_sms_calls_firetext_correctly(mocker, mock_firetext_client):
...
+    assert request_mock.call_count == 1
+    assert request_mock.request_history[0].url == 'https://example.com/firetext'
+    assert request_mock.request_history[0].method == 'POST'
+
+    request_args = parse_qs(request_mock.request_history[0].text)
+    assert request_args['apiKey'][0] == 'international'
+    assert request_args['from'][0] == 'bar'
+    assert request_args['to'][0] == '607234567890'
+    assert request_args['message'][0] == content
+    assert request_args['reference'][0] == reference
```
benthorner, line 103: Minor: do we need all these assertions for this test?

benthorner (approved):
Nice one - that's great we're getting rid of MMG as the international snowflake 🎉. I assume there's a credentials PR coming up to add the international API key.

Code looks good - just a few comments around making the boolean flag clear everywhere, if possible. (I can see you've done that already in some of the tests.)

PR #3182, opened 2021-03-17
Check for incomplete pending jobs

klssmith:
We have a scheduled task that was checking for jobs still in progress.
We saw a case where a scheduled job was stuck in a `pending` status as a
result of an app shutting down. This changes the `check_job_status` task
so that it also checks for scheduled jobs which are still pending after
30 minutes.

[Pivotal story](https://www.pivotaltracker.com/story/show/177095265)

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -3,7 +3,7 @@ from datetime import datetime, timedelta
 
 from flask import current_app
 from notifications_utils.statsd_decorators import statsd
-from sqlalchemy import and_
+from sqlalchemy import between
 from sqlalchemy.exc import SQLAlchemyError
 
 from app import cbc_proxy_client, notify_celery, zendesk_client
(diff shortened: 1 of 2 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -132,19 +133,31 @@ def check_job_status():
...
+        Job.job_status == JOB_STATUS_PENDING,
+        Job.scheduled_for.isnot(None),
+        between(Job.scheduled_for, thirty_five_minutes_ago, thirty_minutes_ago)
+    )
+
+    jobs_not_complete_after_30_minutes = incomplete_in_progress_jobs.union(
+        incomplete_pending_jobs
+        ).order_by(
+            Job.processing_started, Job.scheduled_for
+        ).all()
```
leohemsted, line 160: looks like this should be unindented?
```suggestion
    ).order_by(
        Job.processing_started, Job.scheduled_for
    ).all()
```

leohemsted (approved):
looks good, just a couple of minor tweaks

app/celery/scheduled_tasks.py
klssmith: Thanks, I've squashed that change into the existing commit

PR #3185, opened 2021-03-23
Quick fix for dealing with missing crown on a service sending letters

servingUpAces:
Quick fix to ensure that billing doesn't fail if the crown is not set for the service.

The letters rates for cronw and non crown are the same. It would be nice
to remove the need for crown but for now this is a quick fix.


tests/app/celery/test_reporting_tasks.py
```diff
@@ -379,6 +379,17 @@ def test_get_rate_for_letter_latest(notify_db_session):
...
 
 
+def test_get_rate_for_letter_latest_if_crown_is_none(notify_db_session):
+    # letter rates should be passed into the get_rate function as a tuple of start_date, crown, sheet_count,
+    # rate and post_class
+    crown = create_letter_rate(datetime(2017, 12, 1), crown=True, sheet_count=1, rate=0.33, post_class='second')
+    non_crown = create_letter_rate(datetime(2017, 12, 1), crown=False, sheet_count=1, rate=0.35, post_class='second')
+    letter_rates = [crown, non_crown]
+
+    rate = get_rate([], letter_rates, LETTER_TYPE, date(2018, 1, 1), None, 1)
```
CrystalPea, line 389: maybe we could explicitly say `crown=None` when calling `get_rate()`?

CrystalPea (approved):
Small readability comment, but in general this is good to go 👍🏼

PR #3189, opened 2021-03-24
Reduce max concurrent 2 factor codes

quis:
I was doing some analysis and saw that in the last 24 hours the most codes that anyone had was in a 15 minute window was 3:

image

So I think we can safely reduce this to 5 to get a bit more security with enough headroom to not have any negative impact to the user.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -146,7 +146,7 @@ class Config(object):
     API_PAGE_SIZE = 250
     TEST_MESSAGE_FILENAME = 'Test message'
     ONE_OFF_MESSAGE_FILENAME = 'Report'
-    MAX_VERIFY_CODE_COUNT = 10
+    MAX_VERIFY_CODE_COUNT = 5
 
     # be careful increasing this size without being sure that we won't see slowness in pysftp
     MAX_LETTER_PDF_ZIP_FILESIZE = 40 * 1024 * 1024  # 40mb


CrystalPea (approved):

klssmith:
Reducing `MAX_VERIFY_CODE_COUNT` will stop more than 5 codes being sent, which is what we want to happen. But reducing `MAX_VERIFY_CODE_COUNT` to 5 seems to also mean that we won't verify a user's code if they have more than 5 failed log in attempts, so in effect it means you can't log after 5 failed log in attempts, not 10 (not necessarily an issue, but just wanted to check this is ok)
https://github.com/alphagov/notifications-api/blob/1903e8f26812f1ffd9257a98f24089ee826499d9/app/user/rest.py#L207

quis:
Good spot @klssmith – that’s not the behaviour we want because people do use more than 5 login attempts:
```
+----------------------+------------------+
| failed_login_count   | count_of_users   |
|----------------------+------------------|
| 0                    | 36581            |
| 1                    | 687              |
| 2                    | 216              |
| 3                    | 96               |
| 4                    | 41               |
| 5                    | 32               |
| 6                    | 15               |
| 7                    | 5                |
| 8                    | 3                |
| 9                    | 2                |
| 10                   | 16               |
+----------------------+------------------+
```

quis:
@klssmith I’ve split the config into two, and added an extra test to make sure that failed login count and max concurrent codes are both covered.

klssmith (approved):
Nice, splitting out the two variables makes this much clearer

PR #3191, opened 2021-03-25
send a p1 when a broadcast goes out

leohemsted:
it's important to keep tabs on when these things leave our system. Sending a zendesk ticket that triggers a P1 is probably our simplest way of notifying the team when this happens (it's what we do with out of hours emergencies on the admin app too). We don't have any direct pagerduty integrations from the api app, but we already have the zendesk client hooked up.

After broadcasts go live, we may want to change this to a P2 (but even then, there's arguments for keeping it P1 to start with I think).

### DIFF
--- a/app/celery/broadcast_message_tasks.py
+++ b/app/celery/broadcast_message_tasks.py
@@ -5,7 +5,7 @@ from flask import current_app
 from notifications_utils.statsd_decorators import statsd
 from sqlalchemy.schema import Sequence
 
-from app import cbc_proxy_client, db, notify_celery
+from app import cbc_proxy_client, db, notify_celery, zendesk_client
 from app.clients.cbc_proxy import (
     CBCProxyFatalException,
     CBCProxyRetryableException,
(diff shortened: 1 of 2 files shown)


app/celery/broadcast_message_tasks.py
```diff
@@ -111,6 +111,24 @@ def send_broadcast_event(broadcast_event_id):
...
+    # raise a P1 to alert team that broadcast is going out.
+    message = '\n'.join([
+        'Broadcast Sent',
+        '',
+        f'https://www.notifications.service.gov.uk/services/{broadcast_message.service_id}/current-alerts/{broadcast_message.id}',  # noqa
+        '',
+        f'This broacast has been sent on channel {broadcast_message.service.broadcast_channel}.',
+        f'This broadcast is targeted at areas {broadcast_message.areas.get("areas")}.',
+    ])
+    zendesk_client.create_ticket(
```
leohemsted, line 125: what happens on preview/staging?

app/celery/broadcast_message_tasks.py
```diff
@@ -111,6 +111,24 @@ def send_broadcast_event(broadcast_event_id):
...
+
+    broadcast_message = broadcast_event.broadcast_message
+    # raise a P1 to alert team that broadcast is going out.
+    message = '\n'.join([
+        'Broadcast Sent',
+        '',
+        f'https://www.notifications.service.gov.uk/services/{broadcast_message.service_id}/current-alerts/{broadcast_message.id}',  # noqa
+        '',
+        f'This broacast has been sent on channel {broadcast_message.service.broadcast_channel}.',
+        f'This broadcast is targeted at areas {broadcast_message.areas.get("areas")}.',
```
idavidmcdonald, line 123: Can we link to the runbook google doc here which includes instructions for what to do if an alert goes out accidentally.

Probably worth saying in the zendesk ticket, if the alert going out is unexpected then please follow runbook for accidental alert or similar...

Also given this alert could go off and need a response out of hours, we probably need to think about bringing Unboxed up to speed in the next week or two

app/celery/broadcast_message_tasks.py
CrystalPea: we could possibly check by filling a feedback form on these environments?

CrystalPea: I think there is a way to create an incident through there, right?

leohemsted: zendesk on preview/staging links to our normal zendesk.

i'll make this only run on prod.

idavidmcdonald (approved):
noice

PR #3193, opened 2021-03-29
Populate annual billing for 2021

servingUpAces:
The free SMS allowances are changing for the financial year starting April 1 2021.

In this PR there is a command to set annual_billing for all active
services with the the new defaults. The new command will work for previous year as well. 
The reason for the missing_services_only parameter for the command is so that we can set the defaults for any services that do not have annual billing for 2020. Which will be needed before April 1.

The new method `set_default_free_allowance_for_service` will also be
called in a PR to follow that will set a services free allowance to the
default if the organisation for the service is changed.

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -891,7 +891,7 @@ def populate_annual_billing_with_the_previous_years_allowance(year):
 def populate_annual_billing_with_defaults(year, missing_services_only):
     """
     Add or update annual billing with free allowance defaults for all active services.
-    DEFAULT_FREE_SMS_FRAGMENT_LIMITS is the new free allowances for the financial year starting 2021.
+    The default free allowance limits are in: app/dao/annual_billing_dao.py:57.
 
     If missing_services_only is true then only add rows for services that do not have annual billing for that year yet.
     This is useful to prevent overriding any services that have a free allowance that is not the default.
--- a/app/commands.py
+++ b/app/commands.py
@@ -885,7 +885,7 @@ def populate_annual_billing_with_the_previous_years_allowance(year):
 @notify_command(name='populate-annual-billing-with-defaults')
 @click.option('-y', '--year', required=True, type=int,
               help="""The year to populate the annual billing data for, i.e. 2021""")
-@click.option('-m', '--missing-services-only', default=False, type=bool,
+@click.option('-m', '--missing-services-only', default=True, type=bool,
               help="""If true then only populate services missing from annual billing for the year.
                       If false populate the default values for all active services.""")
 def populate_annual_billing_with_defaults(year, missing_services_only):
--- a/app/commands.py
+++ b/app/commands.py
@@ -12,6 +12,7 @@ from flask import current_app, json
 from notifications_utils.recipients import RecipientCSV
 from notifications_utils.statsd_decorators import statsd
 from notifications_utils.template import SMSMessageTemplate
+from sqlalchemy import and_
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
 
@@ -29,6 +30,7 @@ from app.celery.tasks import process_row, record_daily_sorted_counts
 from app.config import QueueNames
 from app.dao.annual_billing_dao import (
     dao_create_or_update_annual_billing_for_year,
+    set_default_free_allowance_for_service,
 )
 from app.dao.fact_billing_dao import (
     delete_billing_data_for_service_for_day,
@@ -67,6 +69,7 @@ from app.models import (
     NOTIFICATION_CREATED,
     PROVIDERS,
     SMS_TYPE,
+    AnnualBilling,
     Domain,
     EmailBranding,
     LetterBranding,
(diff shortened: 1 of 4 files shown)


app/dao/annual_billing_dao.py
```diff
@@ -49,3 +51,54 @@ def dao_get_all_free_sms_fragment_limit(service_id):
     return AnnualBilling.query.filter_by(
         service_id=service_id,
     ).order_by(AnnualBilling.financial_year_start).all()
+
+
+def set_default_free_allowance_for_service(service, year_start=None):
+    default_free_sms_fragment_limits = {
+        'central': {
```
idavidmcdonald, line 58: Worth looking at https://github.com/alphagov/notifications-admin/pull/3826#pullrequestreview-608658602 and https://github.com/alphagov/notifications-admin/pull/3830. The plan is that we don't have to maintain a list that we have to update every year.

Worth thinking what the implication of doing the same here would be for this PR

app/commands.py
```diff
@@ -877,3 +850,67 @@ def process_row_from_job(job_id, job_row_number):
...
+@notify_command(name='populate-annual-billing-with-defaults')
+@click.option('-y', '--year', required=True, type=int,
+              help="""The year to populate the annual billing data for, i.e. 2021""")
+@click.option('-m', '--missing-services-only', default=False, type=bool,
+              help="""If true then only populate services missing from annual billing for the year.
+                      If false populate the default values for all active services.""")
+def populate_annual_billing_with_defaults(year, missing_services_only):
+    """
+    Add or update annual billing with free allowance defaults for all active services.
+    DEFAULT_FREE_SMS_FRAGMENT_LIMITS is the new free allowances for the financial year starting 2021.
```
idavidmcdonald, line 894: Code comment referencing `DEFAULT_FREE_SMS_FRAGMENT_LIMITS` feels a little weird as this variable doesn't exist (at least not in all caps) and it exists inside a different function, not this one

app/commands.py
```diff
@@ -877,3 +850,67 @@ def process_row_from_job(job_id, job_row_number):
...
+
+    If missing_services_only is true then only add rows for services that do not have annual billing for that year yet.
+    This is useful to prevent overriding any services that have a free allowance that is not the default.
+
+    If missing_services_only is false then add or update annual billing for all active services.
+    This is useful to ensure all services start the new year with the correct annual billing.
```
idavidmcdonald, line 900: So if I understand this correctly, if `missing_services_only` is true then we only update ones that do not have billing for that year yet.

Whereas, `missing_services_only` if false then we will add or update billing for all services. Why would we want to do this? Do we have data already with incorrect annual allowances? Do we anticipate having them as a normal practice? I feel like this is risky to ever run this with `missing_services_only` as `False` as it would override any custom allowances we might have given a service. I think my suggestion is that we don't have this flag and the default is always to only add rows for services, not update them but I am not too familiar with billing so there could be a valid reason for this.

idavidmcdonald (commented):
Generally looks good but a few questions please

app/commands.py
servingUpAces: There are some rows in the annual_billing table already for 2021, We want to reset them to the default. 

I have found in the past that some services where missing annual billing and this will set those to the default without affecting any existing rows (that may have a custom free allowance). 
I will take a back up of the table before running any command (which should be normal practice).
The results will need to be reviewed by Pete.

app/dao/annual_billing_dao.py
servingUpAces: I have a fix for that.

app/commands.py
servingUpAces: Resolved - the defaults were in this function when I wrote the comment.

app/commands.py
```diff
@@ -877,3 +850,67 @@ def process_row_from_job(job_id, job_row_number):
...
+        print("create free limit of {} for service: {}".format(free_allowance[0], row.id))
+        dao_create_or_update_annual_billing_for_year(service_id=row.id,
+                                                     free_sms_fragment_limit=free_allowance[0],
+                                                     financial_year_start=int(year))
+
+
+@notify_command(name='populate-annual-billing-with-defaults')
+@click.option('-y', '--year', required=True, type=int,
+              help="""The year to populate the annual billing data for, i.e. 2021""")
+@click.option('-m', '--missing-services-only', default=False, type=bool,
```
idavidmcdonald, line 888: I think the default argument for this should be `True`. `False` is the much more destructive action as it will overwrite data. Either that or removing the default behaviour so the person running this has to think about what they are doing and make an explicit choice.

idavidmcdonald (approved):
Looks good.

One tiny comment that needs a change please but apart from that this is good I think.

Thanks!

app/commands.py
servingUpAces: ok

app/dao/annual_billing_dao.py
```diff
@@ -49,3 +51,54 @@ def dao_get_all_free_sms_fragment_limit(service_id):
     return AnnualBilling.query.filter_by(
         service_id=service_id,
     ).order_by(AnnualBilling.financial_year_start).all()
+
+
+def set_default_free_allowance_for_service(service, year_start=None):
```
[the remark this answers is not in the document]
servingUpAces: In my next PR I will use this method to update services when the org type changes, that will always pass None for year_start.

PR #3199, opened 2021-04-06
Allow platform admins to cancel broadcasts.

CrystalPea:
Also update error message for when someone does not have permissions.

The message referenced approving broadcasts specifically, whereas people would also see it if they try to cancel or reject broadcast without permission.

Why we allow platform admins to cancel broadcasts:
we do this so they can react quickly if a broadcast was approved by accident.

Pivotal ticket: https://www.pivotaltracker.com/story/show/177482418
Admin PR: https://github.com/alphagov/notifications-admin/pull/3854

### DIFF
--- a/tests/app/broadcast_message/test_rest.py
+++ b/tests/app/broadcast_message/test_rest.py
@@ -489,13 +489,20 @@ def test_update_broadcast_message_status_doesnt_let_you_update_other_things(admi
     }]
 
 
-def test_update_broadcast_message_status_stores_cancelled_by_and_cancelled_at(
-    admin_request, sample_broadcast_service, mocker
+@pytest.mark.parametrize('user_is_platform_admin', [True, False])
+def test_update_broadcast_message_allows_service_user_and_platform_admin_to_cancel(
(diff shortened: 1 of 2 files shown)


tests/app/broadcast_message/test_rest.py
```diff
@@ -489,13 +489,20 @@ def test_update_broadcast_message_status_doesnt_let_you_update_other_things(admi
     }]
 
 
+@pytest.mark.parametrize('user_is_platform_admin', [True, False])
 def test_update_broadcast_message_status_stores_cancelled_by_and_cancelled_at(
```
benthorner, line 493: Optional: since this test is now checking two things (the permission check is implicit in the test still passes), we could rename it to make this more explicit e.g. `test_update_broadcast_message_allows_service_user_and_platform_admin_to_cancel`.

benthorner (approved):
Looks good, nice reuse of the test 👍

tests/app/broadcast_message/test_rest.py
CrystalPea: Good idea 👍🏼

PR #3201, opened 2021-04-08
Migrate towards new metrics for Celery tasks

benthorner:

This adds code to send a new set of StatsD metrics about Celery tasks,
which will replace the existing stats when we have enough data.

The new metrics require less code to implement, are more consistent with
our logging approach, and contain more useful dimensions for filtering.

If the approach here is successful, we can duplicate it in the Antivirus
and Template Preview apps, which have a similar superclass for Celery tasks.

Note: this will temporarily result in us sending roughly double the number
of datapoints to StatsD. This shouldn't cause a problem, as the data is sent
over UDP. I will do a quick test on Preview to check for any warning signs.

### DIFF
--- a/tests/app/celery/test_celery.py
+++ b/tests/app/celery/test_celery.py
@@ -1,0 +1,76 @@
+import uuid
+
+import pytest
+from freezegun import freeze_time
+
+from app import notify_celery
+
+
+@pytest.fixture(scope='session')
+def celery_task():
+    @notify_celery.task(name=uuid.uuid4(), base=notify_celery.task_cls)
+    def test_task(delivery_info=None): pass
+    return test_task
+
+
+@pytest.fixture
+def async_task(celery_task):
+    celery_task.push_request(delivery_info={'routing_key': 'test-queue'})
+    yield celery_task
+    celery_task.pop_request()
+
+
+def test_success_should_log_and_call_statsd(mocker, notify_api, async_task):
+    statsd = mocker.patch.object(notify_api.statsd_client, 'timing')
+    logger = mocker.patch.object(notify_api.logger, 'info')
+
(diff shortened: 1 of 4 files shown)


tests/app/celery/test_celery.py
```diff
@@ -0,0 +1,118 @@
...
+
+    with freeze_time() as frozen:
+        async_task()
+        frozen.tick(5)
+
+        async_task.on_success(
+            retval=None, task_id=1234, args=[], kwargs={}
+        )
+
+    statsd.assert_called_with(f'celery.test-queue.{async_task.name}.success', 5.0)
```
leohemsted, line 37: optional (and may not be relevant to all these, for example the logger asserts), i prefer `assert_called_once_with` to `assert_called_with`. `assert_called_with` will look at all invocations, and assert that at least one of them has the correct matching call method. But it doesn't flag up if the mocked method was called multiple times or what the other times were caleld with.

might be worth changing some of these to `assert_called_once_with`. If they're called multiple times and we care about each call, you can check `my_mock.mock_calls == [call(x), call(y)]` with each different call's set of args and kwargs.

leohemsted (approved):
nice tests 👍 

just a quick optional comment about assert_called_with, i'll leave that up to you

tests/app/celery/test_celery.py
benthorner: Hmm...tricky. In general I think it's safe to assume our code is only executed once. Often a double invocation will lead to an error (so we'll know in the worst case), or be harmless (if the code is idempotent).

StatsD (and arguably the logger calls) is neither of these (it won't error, and the result is misleading data), so I agree it's appropriate to assert it's only called once. I've amended the first "test"  commit 👍.

PR #3202, opened 2021-04-08
fix environment var checking

leohemsted:
# look for 'live', not 'production

config['NOTIFY_ENVIRONMENT'] is hardcoded to `'live'` in the Live config class. The values as seen on the environment which we send real messages from:

```
>>> json.loads(os.environ['VCAP_APPLICATION'])['space_name']  # what cloudfoundry sets
'production'
>>> os.environ['NOTIFY_ENVIRONMENT']  # we set this from cloudfoundry
'production'
>>> current_app.config['NOTIFY_ENVIRONMENT']  # hardcoded in the Live config
'live'
>>> current_app.config['NOTIFICATION_QUEUE_PREFIX']  # pulled from env var of same name
'live'
>>> current_app.config['ENV']  # this is an unrelated flask variable
'production'
```

# only send zendesk P1 for alerts

we don't need to be re-notified when someone clicks cancel

### DIFF
--- a/tests/app/celery/test_broadcast_message_tasks.py
+++ b/tests/app/celery/test_broadcast_message_tasks.py
@@ -137,6 +137,25 @@ def test_send_broadcast_event_creates_zendesk_p1(mocker, notify_api, sample_broa
     assert "Dear Sir/Madam" in zendesk_args['message']
 
 
+def test_send_broadcast_event_doesnt_p1_when_cancelling(mocker, notify_api, sample_broadcast_service):
+    template = create_template(sample_broadcast_service, BROADCAST_TYPE)
+    broadcast_message = create_broadcast_message(
+        template,
+        status=BroadcastStatusType.BROADCASTING,
+        areas={'areas': ['wd20-S13002775', 'wd20-S13002773'], 'simple_polygons': []},
+    )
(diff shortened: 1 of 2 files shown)


idavidmcdonald (approved):

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -137,6 +137,25 @@ def test_send_broadcast_event_creates_zendesk_p1(mocker, notify_api, sample_broa
...
 
 
+def test_send_broadcast_event_doesnt_p1_when_cancelling(mocker, notify_api, sample_broadcast_service):
+    template = create_template(sample_broadcast_service, BROADCAST_TYPE)
+    broadcast_message = create_broadcast_message(
+        template,
+        status=BroadcastStatusType.BROADCASTING,
+        areas={'areas': ['wd20-S13002775', 'wd20-S13002773'], 'simple_polygons': []},
+    )
+    create_broadcast_event(broadcast_message, message_type=BroadcastEventMessageType.ALERT)
```
benthorner, line 147: Do we need this one if we're just testing the cancellation?

leohemsted: i wasn't sure, but it's more correct since you need an alert to have a cancel, and was worried that there might be data integrity concerns that might lead to unexpected behaviour.

PR #3204, opened 2021-04-12
Broadcast envars

idavidmcdonald:
Review commit by commit but the main problem this is solving is reducing the number of our API apps that have IAM creds for sending emergency alerts (only the broadcast worker needs them)


leohemsted (approved):

rjbaker (approved):
Looks good. Also stops the dev environment from triggering the staging/preview CBC proxy 👍

PR #3207, opened 2021-04-13
Add requested and allowed broadcast areas

CrystalPea:
`requested_allowed_broadcast_areas` and `allowed_broadcast_areas` are new columns on `service_broadcast_settings` table.

 `allowed_broadcast_areas` will store information about which areas a service can broadcast to.

`requested_allowed_broadcast_areas` store hold information about which areas a service requested to be able to broadcast to.

This PR also includes adding dao setter methods for each of new fields.

Story ticket: https://www.pivotaltracker.com/story/show/177696929

This is part of the wider work to restrict which areas a service can broadcast to: https://docs.google.com/document/d/1eoxpu50wo1GddVSWWTC6QZhggHN2U4UAGxwhX_2U5ig/edit?usp=sharing

## Note:
This is a draft PR, to discuss if this is the right route, what are the alternatives, and also what data structure we want to hold against those columns. I think it would be optimum to share the same format that we do for broadcast areas. Happy to be convinced otherwise.


idavidmcdonald:
Going to close this one unfortunately. Not because it isn't a good idea but because we aren't going to start this work ourselves. If anything, a new team delivering emergency alerts will take over on this but they should start from scratch, understand the user needs and decide how they want to implement. It also isn't a priority feature (not one that CCS think is important to happen in the first few months after launch).

PR #3208, opened 2021-04-13
Check if service is in live mode before sending a broadcast

klssmith:
**Don't create broadcast event unless necessary**
If we're not going to send a broadcast, we don't need to create the
BroadcastEvent in the database. The BroadcastMessage contains all the
data we need - the BroadcastEvent is not used.

Not creating the event when we won't send the broadcast (e.g. when the
broadcast message was created when the service was in trial mode) adds
an extra layer of security.

**Check if a service is live before sending a broadcast**
We only want to send a broadcast if the broadcast message is not stubbed
and the service is live at the point at which the broadcast event should
be created. This is to prevent the situation where a broadcast service is
switched to live / trial mode in between the message being created and
approved (we log an error if this happens).

A stubbed broadcast message with a trial mode service at the point of
approval is not an issue - trial mode services can approve their own
broadcasts. In this situation, we don't create the broadcast event but
also don't need to log an error.

[Pivotal story](https://www.pivotaltracker.com/story/show/177491060)

### DIFF
--- a/app/broadcast_message/rest.py
+++ b/app/broadcast_message/rest.py
@@ -192,29 +192,29 @@ def _create_broadcast_event(broadcast_message):
     """
     Creates a broadcast event, stores it in the database, and triggers the task to send the CAP XML off
     """
-    msg_types = {
-        BroadcastStatusType.BROADCASTING: BroadcastEventMessageType.ALERT,
-        BroadcastStatusType.CANCELLED: BroadcastEventMessageType.CANCEL,
-    }
-
-    event = BroadcastEvent(
-        service=broadcast_message.service,
-        broadcast_message=broadcast_message,
-        message_type=msg_types[broadcast_message.status],
-        transmitted_content={"body": broadcast_message.content},
-        transmitted_areas=broadcast_message.areas,
-        # TODO: Probably move this somewhere more standalone too and imply that it shouldn't change. Should it include
-        # a service based identifier too? eg "flood-warnings@notifications.service.gov.uk" or similar
-        transmitted_sender='notifications.service.gov.uk',
-
-        # TODO: Should this be set to now? Or the original starts_at?
-        transmitted_starts_at=broadcast_message.starts_at,
-        transmitted_finishes_at=broadcast_message.finishes_at,
-    )
-
-    dao_save_object(event)
-
     if not broadcast_message.stubbed:
+        msg_types = {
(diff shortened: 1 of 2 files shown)


app/broadcast_message/rest.py
```diff
@@ -190,32 +190,42 @@ def update_broadcast_message_status(service_id, broadcast_message_id):
...
-        # a service based identifier too? eg "flood-warnings@notifications.service.gov.uk" or similar
-        transmitted_sender='notifications.service.gov.uk',
-
-        # TODO: Should this be set to now? Or the original starts_at?
-        transmitted_starts_at=broadcast_message.starts_at,
-        transmitted_finishes_at=broadcast_message.finishes_at,
-    )
+    service = broadcast_message.service
+
+    if not broadcast_message.stubbed and not service.restricted:
```
benthorner, line 198: Minor: we could invert this and return early (after logging), to avoid some of the nesting. Then it would look like:

```
if stubbed or restricted:
  log_error
  return

rest_of_method
```

benthorner (approved):
🔒

Just a couple of minor suggestions that you may want to consider before merging.

app/broadcast_message/rest.py
```diff
@@ -219,3 +222,10 @@ def _create_broadcast_event(broadcast_message):
             kwargs={'broadcast_event_id': str(event.id)},
             queue=QueueNames.BROADCASTS
         )
+    elif broadcast_message.stubbed != service.restricted:
+        # We shouldn't have a situation where a service is live and the broadcast message is stubbed (or vice versa)
```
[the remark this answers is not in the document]
klssmith: 👍 comment updated

app/broadcast_message/rest.py
klssmith: We don't want to log an error if the broadcast message is stubbed _and_ the service is restricted (this is the scenario where a trial mode service can approve their own broadcasts) - only if there's a mismatch between the stubbed and restricted values. This means the logic would need to look slightly more complicated if changed:
```python
if stubbed or restricted:
    if stubbed != restricted:
        log_error
  return

rest_of_method
```

benthorner: Oh yes, thanks for explaining. In that case I don't think it makes much difference either way, so let's leave it as-is 👍.

PR #3215, opened 2021-04-20
Add unique key on annual_billing for service_id + financial_year_start.

servingUpAces:
This is an extra precaution for the table to ensure data integrity. Since we only update/insert the data using the annual_billing_dao methods the integrity is in tact. I've check the data on preview, staging and prod there are no violations of this unique key.

### DIFF
--- a/app/dao/annual_billing_dao.py
+++ b/app/dao/annual_billing_dao.py
@@ -102,7 +102,7 @@ def set_default_free_allowance_for_service(service, year_start=None):
                                 f"{default_free_sms_fragment_limits['other'][year_start]}")
         free_allowance = default_free_sms_fragment_limits['other'][year_start]
 
-    dao_create_or_update_annual_billing_for_year(
+    return dao_create_or_update_annual_billing_for_year(
         service.id,
         free_allowance,
         year_start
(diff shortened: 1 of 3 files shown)


app/dao/annual_billing_dao.py
```diff
@@ -102,7 +102,7 @@ def set_default_free_allowance_for_service(service, year_start=None):
                                 f"{default_free_sms_fragment_limits['other'][year_start]}")
         free_allowance = default_free_sms_fragment_limits['other'][year_start]
 
-    dao_create_or_update_annual_billing_for_year(
+    return dao_create_or_update_annual_billing_for_year(
```
CrystalPea, line 105: why do we start returning this? Do we use it somewhere? If we do, how did it work before?

servingUpAces: This was included in https://github.com/alphagov/notifications-api/commit/ae57521b3947a854ce8459b06c268e533676a5aa where I simplified the case when annual_billing is missing when trying to get annual_billing for a year. 
I'm not sure why I had to add it back.

CrystalPea: oh ok, weird :o

CrystalPea (approved):

leohemsted (approved):

PR #3217, opened 2021-04-20
Turn on StatsD for all processes

benthorner:

It's unclear why these should still be overridden.

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -3,9 +3,6 @@
     'NOTIFY_APP_NAME': 'api',
     'disk_quota': '2G',
     'sqlalchemy_pool_size': 30,
-    'additional_env_vars': {
-      'STATSD_HOST': None
-    },
     'routes': {
       'preview': ['api.notify.works'],
       'staging': ['api.staging-notify.works'],
@@ -23,9 +20,6 @@
   'notify-api-sms-receipts': {
     'NOTIFY_APP_NAME': 'api',
     'disk_quota': '2G',
-    'additional_env_vars': {
-      'STATSD_HOST': None
-    },
     'routes': {
       'preview': ['api.notify.works/notifications/sms/mmg', 'api.notify.works/notifications/sms/firetext'],
       'staging': ['api.staging-notify.works/notifications/sms/mmg', 'api.staging-notify.works/notifications/sms/firetext'],


servingUpAces (commented):
seems we only use statsd for delivery-workers and not the Flask apps. I'm not sure that we want to turn it on.

benthorner:
I've had another look at this, and can see we have some inconsistency across our repos:

- Admin and Download API don't have StatsD at all.
- Template Preview has it for Flask ([here](https://github.com/alphagov/notifications-template-preview/blob/f66bccd14d3293dea630b01a0e574c5be2c32413/manifest.yml.j2#L37)) and Celery ([here](https://github.com/alphagov/notifications-template-preview/blob/f66bccd14d3293dea630b01a0e574c5be2c32413/manifest-celery.yml.j2#L32)).
- Antivirus also has it for Flask ([here](https://github.com/alphagov/notifications-antivirus/blob/58d35467e16221b3c79aeccd3de8a213f4a56f2f/manifest-api.yml.j2#L21)) and Celery ([here](https://github.com/alphagov/notifications-antivirus/blob/58d35467e16221b3c79aeccd3de8a213f4a56f2f/manifest.yml.j2#L18)).
- Download Frontend has it configured ([here](https://github.com/alphagov/document-download-frontend/blob/392be78b07afea6ab5d469c8b1113afd27b07dae/app/__init__.py#L53)) but not enabled.

Our [docs](https://github.com/alphagov/notifications-manuals/wiki/Monitoring-and-alerting#statsd-exporter) are unclear about when StatsD should be used. My understanding is that we shouldn't have it in any of our Flask apps, as the Prometheus client can be used instead. How about the following:

- Keep the env override for Flask, but add a comment explaining why.
- Remove the env override for the worker, as it's unclear why it's disabled.

More generally, I could also open an issue with actions to:

- Clarify the docs for where StatsD should be used.
- Remove StatsD env vars from our Flask apps (disabling it).

What do you think?

@idavidmcdonald do you have any comments on the broader issue here?

idavidmcdonald:
So I think the general approach has been give statsd to apps that need it. We've never had a clear cut policy of which apps should have it and which haven't, let alone something documented.

> My understanding is that we shouldn't have it enabled in any of our Flask apps, as the Prometheus client can be used instead

This is my general understanding too. Although I'm a litttle bit less sure on this as I was 6 months ago. 6 months ago I was more along the lines of 'lets get rid of statsd entirely'. Now I see us using it more, which is fine, but then it means actually do we actually try and reduce the use of the python prometheus client and instead use the statsd client (rather than having two different tools). Or maybe we still have two different tools. I don't think we need to decide right now to deviate from anything we are currently doing.

Open to adding some more comments in our code about why things are the way they are (assuming it is not in the git commit too).

Main thing I'm cautious of is this ticket on adding a dashboard for how long celery tasks take to growing in scope to relooking at our statsd configuration in every app. I don't want you having to go too far down this rabbit hole essentially.

I personally wouldn't bother with the Github issue at the moment. This doesn't cause us regular pain and we have bigger areas of tech debt that are more important to address.

benthorner:
Thanks for your thoughts @idavidmcdonald. I agree it's a tangent, and I don't feel strongly about it, except that it caused me some confusion when I was looking through the manifest ("why this setting here and here, but no where else").

I've also realised I've gotten confused between the receipts app and the receipts worker; it's the former that has the override, which is consistent with the API app itself - both are Flask apps, which can use Prometheus directly.

With the StatsD vs. Prometheus balance, we're actually OK:

- We only have StatsD code in Celery workers, where it's needed.
- We have StatsD turned on for Flask, but it's not used for anything.

So my suggestion is turn it off consistently in other apps, to prevent it being used in Flask. I'm not fussed, though, as it's not causing any particular pain at the moment. I'll close this PR, as this discussion helped me discover my original assumptions were wrong - so at least we got something out of it 😄.

benthorner:
I've also updated [our docs](https://github.com/alphagov/notifications-manuals/wiki/Monitoring-and-alerting#application-level-metrics-provided-by-the-python-statsd-client-and-the-prometheus-statsd-exporter) to explain the subtleties of the two options a bit better. Previously we just talked about "Flask apps", but there's a clear distinction between Flask and Celery code when it comes to the choice of monitoring tool.

PR #3218, opened 2021-04-21
Introduce transaction for precompiled letters

servingUpAces:
Wrap the saving the notification and uploading the precompiled letter to s3 in a transaction.

If the upload to s3 fails the notification will not be saved to the
database.

### DIFF
--- a/tests/app/v2/notifications/test_post_letter_notifications.py
+++ b/tests/app/v2/notifications/test_post_letter_notifications.py
@@ -20,6 +20,9 @@ from app.models import (
     Job,
     Notification,
 )
+from app.notifications.process_letter_notifications import (
+    create_letter_notification,
+)
 from app.schema_validation import validate
 from app.v2.errors import RateLimitError
 from app.v2.notifications.notification_schemas import post_letter_response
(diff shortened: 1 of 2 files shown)


tests/app/v2/notifications/test_post_letter_notifications.py
```diff
@@ -673,6 +673,28 @@ def test_post_precompiled_letter_notification_returns_201(
...
+        "content": "bGV0dGVyLWNvbnRlbnQ="
+    }
+
+    auth_header = create_authorization_header(service_id=sample_service.id)
+    with pytest.raises(expected_exception=Exception):
+        client.post(
+            path="v2/notifications/letter",
+            data=json.dumps(data),
+            headers=[('Content-Type', 'application/json'), auth_header])
+
```
idavidmcdonald, line 693: Optional: I think what would make this test even more rigerous would be for us to assert that `create_letter_notification` was called. So then it's very clear that we called `create_letter_notification`, then an exception was thrown but at the end there was no notification saved to the DB

idavidmcdonald (approved):
Nice one.

As part of my review, I ran the unit test you wrote having removed the `with_transaction` statement you had added and it does indeed break so leaves me confident that this is doing what it says it does

tests/app/v2/notifications/test_post_letter_notifications.py
servingUpAces: I've never mocked a method, then used side_effect=method... neat

leohemsted:
hmm im surprised this works! given that within the transaction, `dao_create_notification` also has the `@transactional` wrapper - guess postgres transactions didn't work like i expected them to. 🤔

but if the tests pass... 🤷‍♂️ ✅

idavidmcdonald:
I was confused by this too but then found out it was because I needed to pull master and get new changes which change it to use `autocommit` rather than `transactional`.

See https://github.com/alphagov/notifications-api/blob/91542ad33eb995c9b3e617cde30a9714846dd37b/app/dao/notifications_dao.py#L82

PR #3221, opened 2021-04-26
Warn about SMS failures when it's cost effective

benthorner:
https://www.pivotaltracker.com/story/show/177908219

Previously we warned when as little as 100 SMS messages failed, which
equates to £1.60 in cost. This increases the absolute threshold so that
we avoid taking action when the cost in developer time is greater than
the money wasted. If it takes us an hour to solve the ticket, then that 
equates to roughly £100 in people costs.

In future, we should also consider simplifying this check to be about
permanent failures in general.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -274,7 +274,7 @@ def check_for_services_with_high_failure_rates_or_sending_to_tv_numbers():
             )
             message += "service: {} failure rate: {},\n".format(service_dashboard, service.permanent_failure_rate)
     elif services_sending_to_tv_numbers:
-        message += "{} service(s) have sent over 500 sms messages to tv numbers in last 24 hours:\n".format(
+        message += "{} service(s) have sent high volumes of sms messages to tv numbers in last 24 hours:\n".format(
             len(services_sending_to_tv_numbers)
         )
         for service in services_sending_to_tv_numbers:
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -542,7 +542,7 @@ def dao_fetch_active_users_for_service(service_id):
     return query.all()
 
 
-def dao_find_services_sending_to_tv_numbers(start_date, end_date, threshold=500):
+def dao_find_services_sending_to_tv_numbers(start_date, end_date, threshold=62_500):
     return db.session.query(
         Notification.service_id.label('service_id'),
         func.count(Notification.id).label('notification_count')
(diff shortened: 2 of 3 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -274,7 +274,7 @@ def check_for_services_with_high_failure_rates_or_sending_to_tv_numbers():
             )
             message += "service: {} failure rate: {},\n".format(service_dashboard, service.permanent_failure_rate)
     elif services_sending_to_tv_numbers:
-        message += "{} service(s) have sent over 500 sms messages to tv numbers in last 24 hours:\n".format(
+        message += "{} service(s) have sent high volumes of sms messages to tv numbers in last 24 hours:\n".format(
```
idavidmcdonald, line 277: Would encourage 'high volumes' to be a bit more explicit about what number we consider high volumes to be. For example, `have sent over 50,000 sms messages` or whatever limit we choose

idavidmcdonald (approved):
I think this is OK. Slightly cautious that it's not just wasting money that we want to be aware of these alerts for. Sometimes we want to just spot that a team have a high failure rate which is abnormal as it does likely mean they need to take some action and means they and their users get a better service.

Would be open to getting another opinion that this is the right thing to do (or at least that the threshold should be increased this much)

quis:
There’s something here about catching a team that are about to send 62,500 when they’ve only sent 500 being the best outcome, because you’ve only spent £1.60 plus £100, versus spending £100 plus £100. You could build a whole financial model around when the best time to intervene is, based on expected sending rates, one-off vs ongoing sending, etc, etc…

That said, in the last year, this alert has never triggered for `n > 1000`. So by setting the threshold to such a high number you’re effectively switching it off.

quis:
There’s also a question about why we even try to send these messages to a provider. I can’t remember why we don’t put them in the bin – that would save us 100% of the money for £0 ongoing expense.

benthorner:
Thanks @idavidmcdonald and @quis for your thoughts. Sounds like it's hard to make a decision here:

- Easy part: we don't want to waste lots of money (easy to define based on salaries).
- Hard part: we want to Notify works "well" for users (hard to define, more than just this alert).

As the current email discussion with the "Income Collection" service illustrates, pursuing this alert can easily be a waste of time. Personally, I think Zendesk should be reserved for things that are "definitely broken":

- Usage of TV numbers is "definitely broken".
- General SMS failures can be benign, especially at low levels.

Since the original card was only concerned with TV numbers, perhaps a more acceptable compromise is:

- Remove the extra check about general SMS failures, since it's unclear why we need it at all.
- Leave the threshold for TV numbers as-is (500 is a good number to filter out one-time accidental uses).
- Revert the runbook for this alert back to one about TV numbers.

Long term, it would make more sense for this check to trigger an automated email to the service, since the action should always be the same (for TV numbers). But for now, fewer arbitrary Zendesk tickets would be nice.

How does that sound?

benthorner:
@servingUpAces could you take a look at this one? The original work was from a card written by yourself, so perhaps you can comment on what we're aiming for as a team here (beyond this specific alert).

benthorner:
Thanks everyone for chatting the other day. I've made a new PR (a one word change!) based on that discussion, to avoid confusing the existing discussion on this one. Here it is: https://github.com/alphagov/notifications-api/pull/3234.

PR #3222, opened 2021-04-27
Move request_id injection into send_task override

benthorner:

This applies the same change we made in other apps [1][2]. Adding
the override here is special, though, because it means the others
will now get triggered, since this app is the start of the chain
of tasks for a request. We will also retain existing request_id
tracing for tasks within this app, since "apply_async" calls the
"send_task" method internally, which is the one we're overriding.

## Screenshots

These show the logs we get for sending a precompiled letter, filtered by request ID. The difference is that more of the logs now have the request ID field populated, and so show up with the filter applied.

| Before | After |
| ------ | ------ |
| screencapture-kibana-logit-io-s-4c3b5032-5afa-42e5-8211-bcba010be352-app-kibana-2021-04-27-11_29_28 | screencapture-kibana-logit-io-s-4c3b5032-5afa-42e5-8211-bcba010be352-app-kibana-2021-04-27-11_27_24 |

[1]: https://github.com/alphagov/notifications-template-preview/pull/531/commits/6f3c118a1e071f8169698806450b45aa685859aa
[2]: https://github.com/alphagov/notifications-antivirus/pull/69/commits/2e08b7aa954c120eb0265fda41ad2025616426ba

### DIFF
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -73,16 +73,6 @@ def make_task(app):
 
                 return super().__call__(*args, **kwargs)
 
-        def apply_async(self, args=None, kwargs=None, **other_kwargs):
-            kwargs = kwargs or {}
-
-            if has_request_context() and hasattr(request, 'request_id'):
-                kwargs['request_id'] = request.request_id
-            elif has_app_context() and 'request_id' in g:
-                kwargs['request_id'] = g.request_id
-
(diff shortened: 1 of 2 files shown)


leohemsted (approved):
So good to get to the end of this and for it all to click into place. Great work!

PR #3224, opened 2021-04-27
Add make help commands for commonly used tasks

idavidmcdonald:
if you run `make help` or just `make` then you get a nice list of the
tasks that you can run (or at least, the ones with help text added to
them.

We were missing these for some of the key commands that a developer
would want to know about.

By adding help text to them, they will now show up in `make` or `make
help` and saves a developer needing to go either look in the README or
go look in the Makefile to figure out what commands are available.

Note, there is no particular convention around which commands have help
comments. I don't think we need to figure out this but at the least, the
ones which developers may want to run locally I think should show up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,17 +22,17 @@ NOTIFY_CREDENTIALS ?= ~/.notify-credentials
 ## DEVELOPMENT
 
 .PHONY: bootstrap
-bootstrap: generate-version-file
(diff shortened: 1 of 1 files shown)


benthorner (approved):
I remember spotting this when I was changing the README stuff originally. At the time I wasn't sure who benefits from these comments.

Thanks for fixing!

PR #3225, opened 2021-04-27
Turn on StatsD for receipts worker

benthorner:

It's unclear why this was disabled when the original work (renamed
from callbacks [1]) was added [2]. Enabling StatsD means we'll get
metrics for tasks executed by this worker, which we're missing.

(I suspect the config was copied from the app manifest [3], where it
makes sense, since the app can expose a /metrics endpoint instead.)

[1]: https://github.com/alphagov/notifications-api/pull/3206
[2]: https://github.com/alphagov/notifications-api/commit/9d5b629dadb2ec057569f77414a7d131bbb5035e
[3]: https://github.com/alphagov/notifications-api/blob/39d7c347b047966bb56b1805ced0463e36e6654c/manifest.yml.j2#L7


benthorner:
Never mind - I got confused between `notify-api-sms-receipts` and `notify-delivery-worker-receipts`!

PR #3233, opened 2021-05-05
Add a type table for broadcast providers

klssmith:
This adds a type table for broadcast providers, which is the pattern we
follow with our models (e.g. we have a `broadcast_channel_types` table).

As well as the four providers, the migration populates it with `all`
which is the value that will replace `null` in a later change.

It should be safe to add the foreign key constraint to the
`service_broadcast_settings` in the same migration since the column is
still nullable and we don't have data in that column that is not in the
types table.

Step one of [Pivotal story](https://www.pivotaltracker.com/story/show/177759863)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -120,7 +120,8 @@ def notify_db_session(notify_db, sms_providers):
                             "broadcast_status_type",
                             "invite_status_type",
                             "service_callback_type",
-                            "broadcast_channel_types"]:
+                            "broadcast_channel_types",
+                            "broadcast_provider_types"]:
             notify_db.engine.execute(tbl.delete())
     notify_db.session.commit()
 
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -120,7 +120,8 @@ def notify_db_session(notify_db, sms_providers):
                             "broadcast_status_type",
                             "invite_status_type",
                             "service_callback_type",
-                            "broadcast_channel_types"]:
+                            "broadcast_channel_types",
```
CrystalPea, line 123: Ahh, is it where we clean test db when running tests?

CrystalPea (approved):
Looks good 👍🏼 

> the code will also have to do a couple of step changes, one to expect None or 'all'

is this something we could do in this PR?

tests/conftest.py
klssmith: Yes, and the tables in this list are the ones that we don't want to clean since they contain fixed data that should always exist.

CrystalPea: makes sense, thanks! ☺️

klssmith:
@CrystalPea You're right, that change could also have been added here. I had that next change on another branch, but I've moved the commit here instead - see what you think

CrystalPea (approved):
looks good 👍🏼

PR #3239, opened 2021-05-10
Add webauthn

leohemsted:
See commits for details. Took me a while to get my head around the data returned from the browser - keeping it simple by not splitting much of it, and just storing it in the database as binary blobs. The API doesn't need to know what the things mean, it just needs to return them back to the admin when needed.

noteably we're not doing any validation on whether the credentials are duplicated, nor is there a uniqueness constraint in the database - I'm not sure if the credential blob will always reliably be equivalent, and I believe the browser should hopefully take care of dupes.


app/models.py
```diff
@@ -2583,3 +2583,37 @@ class ServiceBroadcastProviderRestriction(db.Model):
...
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    updated_at = db.Column(db.DateTime, nullable=True, onupdate=datetime.datetime.utcnow)
+
+    def serialize(self):
+        return {
+            'id': str(self.id),
+            'user_id': str(self.user_id),
+            'name': self.name,
+            'credential_data': self.credential_data,
+            'registration_response': self.registration_response,
```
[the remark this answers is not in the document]
CrystalPea: the `registration_response`? I'm happy to remove it if we don't need it.

app/models.py
```diff
@@ -2593,3 +2593,37 @@ class ServiceBroadcastProviderRestriction(db.Model):
...
+    A table that stores data for registered webauthn credentials.
+    """
+    __tablename__ = "webauthn_credential"
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True, nullable=False, default=uuid.uuid4)
+
+    user_id = db.Column(UUID(as_uuid=True), db.ForeignKey('users.id'), nullable=False)
+    user = db.relationship(User, backref=db.backref("webauthn_credentials"))
+
+    name = db.Column(db.String, nullable=False)
```
klssmith, line 2609: Maybe not worth it in this PR (and it's not MVP) but I wondered whether it would be worth having some kind of constraint to not allow a user to create more than one key with the same name. Just to avoid confusion - similar to how we don't let a service name give multiple API keys the same name.

app/webauthn/rest.py
```diff
@@ -0,0 +1,66 @@
...
+        name=data['name'],
+        credential_data=data['credential_data'],
+        registration_response=data['registration_response']
+    )
+
+    return jsonify(data=webauthn_credential.serialize()), 201
+
+
+@webauthn_blueprint.route('/<uuid:webauthn_credential_id>', methods=['POST'])
+def update_webauthn_credential(user_id, webauthn_credential_id):
```
klssmith, line 42: The delete endpoint has a check that the user and credential match - that would be good here too.

klssmith (commented):
I didn't have the frontend to test this with, so haven't tried out the code. In future (and when it's not just platform admins using this), we could think about whether the key gets deleted if a user gets archived - we currently remove phone numbers when we archive a user.

migrations/versions/0353_add_webauthn_table.py
```diff
@@ -0,0 +1,35 @@
+"""
+
+Revision ID: 0353_add_webauthn_table
```
klssmith, line 3: I've just merged a migration 0353, so you'll need to update the number

tests/app/webauthn/test_rest.py
```diff
@@ -0,0 +1,173 @@
...
+
+    resp = admin_request.get(
+        'webauthn.get_webauthn_credentials',
+        user_id=me.id,
+    )
+
+    creds = sorted(resp['data'], key=lambda x: x['name'])
+    assert len(creds) == 2
+
+    assert creds[0] == {
```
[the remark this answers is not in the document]
leohemsted: not sure - we cant return whole model since the serialize function is how that gets into it. do you mind digging into this?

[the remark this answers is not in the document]
klssmith: @CrystalPea these tests weren't being run because there was an error importing this test file. If you add an `__init__.py` file in this same directory it fixes the set-up error and this test then fails.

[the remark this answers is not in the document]
CrystalPea: ahh, sweet, thanks for solving this @klssmith 💖

app/models.py
benthorner: We're not too concerned about this: names are optional, as they're purely for the convenience of the user. If two keys do end up with the same name, they could use the registration date to disambiguate.

app/webauthn/rest.py
```diff
@@ -0,0 +1,66 @@
...
+
+@webauthn_blueprint.route('/<uuid:webauthn_credential_id>', methods=['DELETE'])
+def delete_webauthn_credential(user_id, webauthn_credential_id):
+    webauthn_credential = dao_get_webauthn_credential_by_id(webauthn_credential_id)
+    user = get_user_by_id(user_id)
+
+    if webauthn_credential.user_id != user.id:
+        raise InvalidRequest('Webauthn credential does not belong to this user', status_code=400)
+
+    if len(user.webauthn_credentials) == 1:
```
benthorner, line 64: Minor: I don't think we need this check in the API. Reasons are:

- We're going to check it in Admin anyway.
- It's not a technical issue in the API if there are no credentials for a user.
- We're hoping to remove this constraint in future.

app/webauthn/webauthn_schema.py
```diff
@@ -0,0 +1,23 @@
+post_create_webauthn_credential_schema = {
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "POST webauthn_credential schema",
+    "type": "object",
+    "properties": {
+        "name": {"type": "string"},
+        "credential_data": {"type": "string"},
+        "registration_response": {"type": "string"},
+    },
+    "required": ["name", "credential_data", "registration_response"],
```
benthorner, line 10: Does this allow for blank names (empty string)? We decided we would allow this, since they're for the user's convenience (so it's not worth the effort doing validation).

tests/app/webauthn/test_rest.py
```diff
@@ -0,0 +1,173 @@
...
+    assert response['message'] == 'Cannot delete last remaining webauthn credential for user'
+
+
+def test_delete_webauthn_credential_errors_if_user_id_doesnt_match(admin_request, notify_db_session):
+    user_1 = create_user(email='1')
+    user_2 = create_user(email='2')
+    cred_1a = create_webauthn_credential(user_1)  # noqa
+    cred_1b = create_webauthn_credential(user_1)  # noqa
+    cred_2a = create_webauthn_credential(user_2)
+    cred_2b = create_webauthn_credential(user_2)  # noqa
```
benthorner, line 184: Minor: how come we need all these credentials, are we testing something implicitly with them? If we need them, we could remove the assignments to avoid `# noqa`.

app/webauthn/rest.py
```diff
@@ -0,0 +1,74 @@
...
+@webauthn_blueprint.route('/<uuid:webauthn_credential_id>', methods=['POST'])
+def update_webauthn_credential(user_id, webauthn_credential_id):
+    data = request.get_json()
+    validate(data, post_update_webauthn_credential_schema)
+
+    webauthn_credential = dao_get_webauthn_credential_by_id(webauthn_credential_id)
+
+    user = get_user_by_id(user_id)
+
+    check_credential_belongs_to_user(webauthn_credential.user_id, user.id)
```
benthorner, line 50: Minor: we could rework this to be part of `dao_get_webauthn_credential_by_id` instead, by making it scoped to the user. Then we'd correctly get a `404` if we tried to update or delete a credential for a different user, rather than a `400`.

app/webauthn/webauthn_schema.py
benthorner: Update: I checked myself by modifying the tests, and it does allow a blank name. A very minor suggestion would be to have a test for this edge case.

benthorner (approved):
Looks good to me 👍. I ended up leaving a few minor comments along the way. I appreciate they're coming quite late on, so we could just get this in and iterate them later.

app/webauthn/webauthn_schema.py
leohemsted: i'd rather not allow blank names. it will lead to more confusion than not. i'd rather update this to specifically check the length

benthorner: Hmm...I'm less certain, especially as it'll mean more work in the Admin app to handle / prevent the errors.

I think we can assume most people won't deliberately erase the name, and either enter their own, or stick with the default of "Unnamed key".

What's the downside of allowing a blank name?

leohemsted: it's a massive usability/debugability issue as the string won't be truthy, and for example has been a big pain point/gross hack in our handling of placeholders where via the api you can send in empty string to have a blank placeholder but in the admin app you can't do that if sending a one-off message. i'm happy to put my foot down on this. 

if we really don't want to force people to write a name, even a placeholder name like "default key" or whatever, then we should make the string nullable but enforce that if it is not null it has length of at least 1.

benthorner: That's interesting to hear it's been an issue elsewhere. Empty strings aren't truthy in Ruby, so that one didn't occur to me. I don't _think_ it would be an issue, since the name is just something we show to the user (we don't do any processing on it). However, I also agree that there's an element of best practice here, so let's just require a min length.

I've made a note in the planning doc, so whoever does the update/remove card in Admin knows to support it.

app/webauthn/rest.py
leohemsted: that's reasonable. we have precedence for doing this too: 

https://github.com/alphagov/notifications-api/blob/master/app/dao/jobs_dao.py#L52-L53

app/webauthn/rest.py
CrystalPea: I think it's usually good to have two lines of defence, just in case, at least that's what I've learnt both on my course and so far at GDS.

Why are we hoping to remove this constraint in the future?

tests/app/webauthn/test_rest.py
leohemsted: yeah it felt good at the time but on reflection i think you're right, removed the extra ones

app/webauthn/rest.py
benthorner: I'd usually agree, if the code is to protect us from a technical (or security) issue.

We're not sure if we're going to keep this restriction, as there are other ways to prevent a user being locked out if they remove their last key. In particular, we want to allow a user to switch of WebAuthn somehow.

So I don't think we should make our code to "hard", as this just means it's more work to change later on. On the other hand, it's not _that_ much work to remove the restriction here, which is why I put it down as a minor issue.

In future, we could scope this check to only error if the user has the `webauth_auth` type. Otherwise, it's absolutely fine for them to remove their last credential, since they're not using them for anything.

leohemsted: lets leave it as is for now, i've put a TODO comment in to add webauthn_auth type checking 👍

PR #3241, opened 2021-05-11
Add "government" as a broadcast channel type

klssmith:
This adds "government" to the broadcast_channel_types table. It also adds it to the service_broadcast_settings_schema so that notifications-admin can pass through a value of "government". We don't have any logic around the value of service.broadcast_channel, so no updates are needed to the tasks etc.

[Pivotal story](https://www.pivotaltracker.com/story/show/178110633)


leohemsted (approved):

quis:
This is working locally alongside my admin PR. Nice job.

PR #3243, opened 2021-05-12
Add webauthn as an auth type 

CrystalPea:
Both in our models and as a migration to add it to auth_types table.

Make sure that if we downgrade, we first clean up the data.

Also add webauthn to tests that include other auth type


leohemsted (approved):

app/models.py
```diff
@@ -124,7 +125,7 @@ class User(db.Model):
     )
 
     # either email auth or a mobile number must be provided
-    CheckConstraint("auth_type = 'email_auth' or mobile_number is not null")
+    CheckConstraint("auth_type in ('email_auth', 'webauthn_auth') or mobile_number is not null")
```
klssmith, line 128: I think this check constraint also needs updating in the database, so will need an update in the migration.

CrystalPea: Ahh, good point 👍🏼

CrystalPea: Added now :3

migrations/versions/0356_add_webautn_auth_type.py
```diff
@@ -0,0 +1,40 @@
...
+revision = '0356_add_webautn_auth_type'
+down_revision = '0355_add_webauthn_table'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute("INSERT INTO auth_type VALUES ('webauthn_auth')")
+
+    op.drop_constraint('ck_users_mobile_or_email_auth', 'users', type_=None, schema=None)
+    op.create_check_constraint(
```
servingUpAces, line 19: When an alter table command is issues any writes to the table will be blocked during the execution of the command. To mitigate this we can add a `NOT VALID` command. Then in another PR we can validate the check constraint.
This will reduce any risk of the data migration blocking any commits to the users table.

PR #3244, opened 2021-05-12
Allowlist

leohemsted:
problem is this brings in utils which bumps everything that im scared about


klssmith (approved):
Looks like the utils bump that this was originally doing has already been merged by someone else, so these tests look fine

PR #3250, opened 2021-05-25
Bump utils version for new invalid address character

servingUpAces:
Bump utils version which contains `~` as an invalid first character for a postal address.

### DIFF
--- a/app/v2/notifications/get_notifications.py
+++ b/app/v2/notifications/get_notifications.py
@@ -55,7 +55,7 @@ def get_pdf_for_notification(notification_id):
     except Exception:
         raise PDFNotReadyError()
 
-    return send_file(filename_or_fp=BytesIO(pdf_data), mimetype='application/pdf')
+    return send_file(path_or_file=BytesIO(pdf_data), mimetype='application/pdf')
 
 
 @v2_notification_blueprint.route("", methods=['GET'])
--- a/tests/app/user/test_rest.py
+++ b/tests/app/user/test_rest.py
@@ -1,8 +1,8 @@
 import json
+import unittest.mock as mock
 import uuid
 from datetime import datetime
 
-import mock
 import pytest
 from flask import current_app, url_for
 from freezegun import freeze_time
--- a/tests/app/v2/test_errors.py
+++ b/tests/app/v2/test_errors.py
@@ -128,7 +128,5 @@ def test_bad_method(app_for_test):
 
             assert response.status_code == 405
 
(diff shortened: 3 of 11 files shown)


tests/app/user/test_rest.py
```diff
@@ -1,8 +1,8 @@
 import json
+import unittest.mock as mock
```
CrystalPea, line 2: what happened with those imports? Did dependency bump mean they stopped working or something?

app/v2/notifications/get_notifications.py
```diff
@@ -55,7 +55,7 @@ def get_pdf_for_notification(notification_id):
     except Exception:
         raise PDFNotReadyError()
 
-    return send_file(filename_or_fp=BytesIO(pdf_data), mimetype='application/pdf')
+    return send_file(path_or_file=BytesIO(pdf_data), mimetype='application/pdf')
```
CrystalPea, line 58: why has this argument name changed? Also something to do with dependency updates?

tests/app/v2/test_errors.py
```diff
@@ -128,7 +128,5 @@ def test_bad_method(app_for_test):
 
             assert response.status_code == 405
 
-            assert response.json == {
-                "result": "error",
-                "message": "The method is not allowed for the requested URL."
-            }
+            assert response.get_data(as_text=True) == \
```
CrystalPea, line 131: what happened with this assertion? Sorry if I'm asking silly questions, it's just confusing 😅

CrystalPea (commented):
Dependency updates can be finicky, eh? :)

I have a few questions:

tests/app/user/test_rest.py
servingUpAces: Bumping the pytest dependencies meant the old mocks package was not being brought in anymore. 
We were already using unittest.mock elsewhere. From the mocks pypi: `mock is now part of the Python standard library, available as unittest.mock in Python 3.3 onwards.`

app/v2/notifications/get_notifications.py
servingUpAces: The send_file method has changed in the latest version of Flask

tests/app/v2/test_errors.py
servingUpAces: The test request response no longer has json as an attribute.
All the changes in this PR are because the dependencies were incremented.

CrystalPea (approved):

PR #3251, opened 2021-05-25
Force Platform Admins to use WebAuthn

benthorner:
https://www.pivotaltracker.com/story/show/178213469

This introduces a DB 'check' constraint to prevent a user being a
platform admin unless they are using WebAuthn as their 2FA method.
On the other hand, a normal user can still use WebAuthn, so:

- A new user can setup their credentials before becoming an admin.

- An existing Platform Admin who is unable to login with WebAuthn
must be demoted before they can use a less secure auth type.

Apart from enforcing these processes, having this check also means
we don't have to worry too much about fixing places in the code
where the auth type may be changed, such as a service invite [1].

[1]: https://github.com/alphagov/notifications-admin/blob/master/app/main/views/manage_users.py#L146

### DIFF
--- a/migrations/versions/0358_require_platform_admin_webauthn.py
+++ b/migrations/versions/0358_require_platform_admin_webauthn.py
@@ -1,0 +1,28 @@
+"""
+
+Revision ID: 4561e334fa59
+Revises: 0357_validate_constraint
+Create Date: 2021-05-25 10:43:33.880337
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = '4561e334fa59'
+down_revision = '0357_validate_constraint'
+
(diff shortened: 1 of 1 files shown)


benthorner:
(This isn't ready to merge until all Platform Admins are using WebAuthn, otherwise the constraint will fail.)

benthorner:
Closing as this won't work: we only expose the security keys feature to Platform Admins, so it wouldn't be possible for them to  register their security key initially. We could do something like:

- Change the Admin app to expose the feature if the auth type is set to WebAuthn;
- Ask the user to login, then set the auth type to WebAuthn so they can register;

but that would be quite a complex and unintuitive experience.

PR #3252, opened 2021-05-26
Add a error log for alert tasks.

servingUpAces:
Many of the team members do not look at emails from zendesk, adding a current_app.logger.error message for things we care about to give developers a better chance of seeing them.
I have purposely not added an erro log for `check_for_services_with_high_failure_rates_or_sending_to_tv_numbers` because it's not something we need to look at immediately.

### DIFF
--- a/app/celery/broadcast_message_tasks.py
+++ b/app/celery/broadcast_message_tasks.py
@@ -150,6 +150,7 @@ def send_broadcast_event(broadcast_event_id):
             ticket_type=zendesk_client.TYPE_INCIDENT,
             p1=True,
         )
+        current_app.logger.error(message)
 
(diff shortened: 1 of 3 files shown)


benthorner:
Could you explain a little more about the expectation here: what kind of improvement do we expect to see after we merge this PR? On the one hand it's useful to have more stuff in logs, but I'm a bit confused how this would have helped this morning.

servingUpAces:
We would have gotten an alert email.

benthorner:
So is the expectation that we'll fix the issue more quickly if more people spot it in their email? If we need to fix something quickly, then would it be more effective to turn these Zendesk tickets into Pagerduty alerts?

servingUpAces:
I don't believe pager duty alerts are necessary

PR #3254, opened 2021-05-27
Fix 'flask db upgrade' not working on CI

benthorner:
Related to: https://github.com/alphagov/notifications-aws/pull/905

Previously this would fail because the Docker image we use for CI
builds doesn't have an 'environment.sh' file; it uses preset env
vars instead. This makes the command to load the file optional,
for which the error should be self evident if it's missing locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ NOTIFY_CREDENTIALS ?= ~/.notify-credentials
 bootstrap: generate-version-file ## Set up everything to run the app
 	pip3 install -r requirements_for_test.txt
 	createdb notification_api || true
-	. environment.sh && flask db upgrade
+	(. environment.sh && flask db upgrade) || flask db upgrade
 
 .PHONY: run-flask
 run-flask: ## Run flask


Makefile
```diff
@@ -25,7 +25,7 @@ NOTIFY_CREDENTIALS ?= ~/.notify-credentials
 bootstrap: generate-version-file ## Set up everything to run the app
 	pip3 install -r requirements_for_test.txt
 	createdb notification_api || true
-	. environment.sh && flask db upgrade
+	. environment.sh || flask db upgrade
```
idavidmcdonald, line 28: Would be interested to know your choice of bash syntax here (partly for my own learning).

So this changes it from an `and` to an `or`. Do we want an `or` here? It means that if there was an environment.sh file in someones dev environment but their db upgrade failed then `make bootstrap` would still appear successful?

Could another option be 

```
. environment.sh || true	
flask db upgrade
```

I write bash fairly rarely so feel free to correct me with any incorrect statements/assumptions I may have made

In principle, I think the approach of trying to source the environment.sh file and it not succeeding but the process continuing on is acceptable

benthorner: Sure, I can explain more:

- First thing to note is that each line of the make rule is run in isolation, so doing `. environment.sh` on one line and `flask db upgrade` on another line won't work.

- You're right that this change means `. environment.sh` will fail silently if someone doesn't have the file. However, `flask db upgrade` _will_ then fail loudly.

I think this is OK because:

- Frankly, people should read the README, which tells them to create an `environment.sh` file 😉.
- The error from `flask db upgrade` is pretty good indicator that the `environment.sh` file is missing.

Does that help?

idavidmcdonald: Thanks, just to further clarify

> First thing to note is that each line of the make rule is run in isolation, so doing . environment.sh on one line and flask db upgrade on another line won't work.

Is the fact each rule is run in isolation because it is bash or is this because it is a make file?
And have you tested this assumption or got a link to back it up out of interest? If not I can run a quick test to confirm it for myself

Another option to ponder would be

```
(environment.sh || true) &&  flask db upgrade
```

Not sure if that is valid bash with the brackets but then it is essentially optional that environment.sh can be sourced but required that the db can be upgraded

idavidmcdonald (approved):
Regardless of my questions about the intricacies of bash, this is good to go. Feel free to make changes or not as you see fit based on my comments but I'm happy enough with the overall approach here

Makefile
benthorner: > Is the fact each rule is run in isolation because it is bash or is this because it is a make file?

It's because it's a Makefile. I think I learnt that through trial and error, so it's just background knowledge for me now. You can find accounts of it online too (e.g. [here](https://stackoverflow.com/questions/54481072/source-env-variables-files-inside-makefile)).

The default shell for `make` is (apparently) `/bin/sh` by the way.

> `(environment.sh || true) &&  flask db upgrade`

Oops. This led me to remember `||` is lazy, so `.environment.sh || flask db upgrade` won't get to `flask db upgrade` 😬. I've amended the commit with something that actually works.

Thanks for persisting with the discussion ⭐. Do you want to take another look?

Makefile
```diff
@@ -25,7 +25,7 @@ NOTIFY_CREDENTIALS ?= ~/.notify-credentials
 bootstrap: generate-version-file ## Set up everything to run the app
 	pip3 install -r requirements_for_test.txt
 	createdb notification_api || true
-	. environment.sh && flask db upgrade
+	(. environment.sh && flask db upgrade) || true
```
idavidmcdonald, line 28: I don't think this works.

I think it tries to run `. environment.sh` and it returns false so therefore does not run the `flask db upgrade` statement. Therefore meaning we don't upgrade the DB which we do want to do on concourse for our test runs

If I delete my environment.sh file and run this locally I get

```
...
createdb: database creation failed: ERROR:  database "notification_api" already exists
/bin/bash: environment.sh: No such file or directory
```

If I put my environment.sh file back and run locally I get

```
...
createdb: database creation failed: ERROR:  database "notification_api" already exists
2021-06-14T19:06:15 app app INFO no-request-id "Logging configured" [in /Users/davidmcdonald/.virtualenvs/notifications-api/lib/python3.6/site-packages/notifications_utils/logging.py:37]
```

benthorner: Oh noes! This like whack-a-mole.

> (. environment.sh || true ) && flask db upgrade

This won't work unfortunately, as the scope of commands run in `(...)` is limited to the parentheses. (I only realised this from trial and error.)

OK. I've now got something that should work in all cases:

```
(. environment.sh && flask db upgrade) || flask db upgrade
```

To spell it out and confirm I'm not missing anything:

- When running locally (assuming `environment.sh` exists), the first side of the `||` will run and thus `flask db upgrade`.
- When running locally (without `environment.sh`), the second side will run, and error with `KeyError: 'NOTIFY_ENVIRONMENT'`.
- When running on CI (without `environment.sh`), the second side will run, using the keys already in the CI environment.

So...I think this is the it. This is the one 😄.

idavidmcdonald (approved):

Makefile
benthorner: It wasn't the one 🤣.

In all the excitement about the logic, I forgot that CI doesn't run migrations, so that third statement is false:

> When running on CI (without environment.sh), the second side will run, using the keys already in the CI environment.

I've raised a PR to fix: https://github.com/alphagov/notifications-api/pull/3271

PR #3257, opened 2021-06-02
Refactor high volume into serialised service model

quis:
Just looks a bit tidier and less repetitive.

I’ve only done this for the serialised service because:
- we’re only checking this in places where we’re already using the serialised service
- if we want to check this elsewhere there’s a good chance that new code should be using the serialised service, since it’ll itself be doing some kind of performance optimisation


benthorner (approved):
Makes sense to me, nice tidy up 👍.

(We'll get an error if we try to use the same method on a real Service object, so that'll be our prompt to add it there, or switch to the serialised version like you suggested.)

PR #3261, opened 2021-06-08
Remove scheduled_notifications

servingUpAces:
All code has been removed for ScheduledNotifications. This PR just removes the table, it has never been used.

### DIFF
--- a/migrations/versions/0359_remove_sched_notifications_.py
+++ b/migrations/versions/0359_remove_sched_notifications_.py
@@ -1,7 +1,7 @@
 """
 
-Revision ID: 0358_remove_sched_notifications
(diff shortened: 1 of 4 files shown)


benthorner (approved):
Upgrade and downgrade work for me 👍.

migrations/versions/0359_remove_sched_notifications_.py
```diff
@@ -0,0 +1,35 @@
+"""
+
+Revision ID: 0359_remove_sched_notifications
```
klssmith, line 3: I'm afraid I've just merged a migration number 0359, so this will need to be updated as a result

PR #3265, opened 2021-06-14
Bind splunk service to broadcasts worker too

sakisv:


idavidmcdonald (commented):
This looks like it does what you want it to.

However, line 9 in https://docs.google.com/document/d/17eYGKLgkJ8z1i4GgaF9p9953BMAtPwJDHWJqYIWy0_Q/edit#heading=h.2xrdrir7yqt2 seems to say we want all of the celery workers bound to the notify splunk service? Although it also says that is done?

Bit confused by that...

sakisv:
Ah, perfect, thanks for pointing it out. The doc linked in the one you linked (https://docs.google.com/spreadsheets/d/1IM2efTLBS_V90qGUHkg0XpWZwIf3BG1oPOqu8n51eJE/edit#gid=1102171704) at lines 17 and 18 says that realtime log shipping to splunk is not required for any worker, including the broadcast one, as long as we keep them around for a sufficient period.

My talks with Cyber was based on that doc, I don't know who was involved in creating the doc you linked to and if/when/why it replaced the first doc.

idavidmcdonald:
This is all unclear, so for that reason let us close this and if it is important, it will come up again

PR #3266, opened 2021-06-14
Tidy up models

servingUpAces:
# What
- Update the Notification and NotificationHistory model to reflect the database.
- Updates to datatypes, removal of indexes and addition of indexes.

# Why
After running the `flask db migrate` command there are many deltas because we did some work to update the notification and notification_history tables, however, the SQLAlchemy models were not updated to reflect those changes. This PR cleans up all those deltas.
However, there are still some differences that can be done but we can look at that in another PR.

# Reviewing this PR

- Run `flask db upgrade --rev-id before` 
- move before.py out of migrations 
- Checkout this branch
- Run `flask db upgrade --rev-id after`

You should see a significant difference in the two scripts

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1472,7 +1472,20 @@ class Notification(db.Model):
             ['templates_history.id', 'templates_history.version'],
         ),
         UniqueConstraint('job_id', 'job_row_number', name='uq_notifications_job_row_number'),
-        {}
+        Index(
+            'ix_notifications_notification_type_composite',
+            'notification_type',
+            'status',
+            'created_at'
+        ),
+        Index('ix_notifications_service_created_at', 'service_id', 'created_at'),
+        Index(
+            "ix_notifications_service_id_composite",
+            'service_id',
+            'notification_type',
+            'status',
+            'created_at'
+        )
     )
 
     @property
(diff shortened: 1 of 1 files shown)


CrystalPea (commented):
When I run `FLASK_APP=application.py NOTIFY_ENVIRONMENT=development flask db migrate --rev-id before` on master, I do get a file with a big diff.

Then I delete that file, change branch to this branch, and run `FLASK_APP=application.py NOTIFY_ENVIRONMENT=development flask db migrate --rev-id after`, and no file is generated. I also tried doing a small change to `models.py` to see if that would be picked up, and it isn't. I change back to master and it's fie again 🤔 

I'm not sure why, but it doesn't want to work on my local.

app/models.py
```diff
@@ -1742,6 +1741,30 @@ def update_from_original(self, original):
         self.status = original.status
 
 
+# Indexes for notification_history and notifications to improve performance of fetch queries.
+Index(
```
benthorner, line 1745: Minor: could we inline these with the model? I notice we have [a precedent for the inline approach already](https://github.com/alphagov/notifications-api/blob/916a1b55c5b0fb222d0accc9f5cbe98c786a0505/app/models.py#L831), which I prefer as it's easier to see which indices go with which tables.

I did [a quick test](https://github.com/alphagov/notifications-api/pull/new/sync-migrations-ben-test) to see if it would work for these ones as well, and it does (`flask db migrate` does nothing).

benthorner (approved):
Works for me: `flask db migrate` doesn't detect any differences.

I found I _did_ need to run `flask db upgrade` on master, as otherwise I `flask db migrate` wouldn't generate a file on that branch either. I'm not sure why that is. It might explain why @CrystalPea was getting a different outcome to both of us.

PR #3268, opened 2021-06-15
Update the job_status to in-progress sooner.

servingUpAces:
We had a situation where the delivery-worker app instance was terminated before the job was marked as `in-progress`, presumably because the query to check the daily limits was taking too long to complete.
If the job was in progress the `check_job_status` task would have restarted the job.
Updating the status to in-progress sooner will help.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -71,6 +71,10 @@ def process_job(job_id, sender_id=None):
 
     service = job.service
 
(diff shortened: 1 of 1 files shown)


benthorner (approved):
Nice explanation 👍. While ideally our queueing system should recover at any stage, this PR does what it says and reduces the likelihood of a batch job task getting lost.

PR #3271, opened 2021-06-15
Don't try to upgrade DB on CI

benthorner:
This was added by mistake - the Concourse pipeline never did this
previously, and errors if we try (the necessary environment vars
aren't present, even if we wanted to).


idavidmcdonald (approved):
I'll approve this as it's likely correct but could you just remind me why concourse doesn't need to run `flask db upgrade` please? At what point does its DB get brought up to scratch, I'm guessing as part of `make test`?

benthorner:
> At what point does its DB get brought up to scratch, I'm guessing as part of make test?

This puzzled me as well. You're correct, it is part of the tests: https://github.com/alphagov/notifications-api/blob/master/tests/conftest.py#L85

PR #3274, opened 2021-06-17
Add new broadcast related permissions

klssmith:
We want to have new permissions which will be used specifically for broadcasts:
- `create_broadcasts`
- `approve_broadcasts`
- `reject_broadcasts`
- `cancel_broadcasts`

Cancel and reject will always go together, but having separate database permissions makes things easier to change in the future.

The permission column of the permissions table is an enum. We can add values in the alembic upgrade script, but removing individual values from an enum is not supported by Postgres. To remove values, we have to recreate the enum with the old values.

[Pivotal story](https://www.pivotaltracker.com/story/show/178482657)

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1870,6 +1870,10 @@ SEND_LETTERS = 'send_letters'
 MANAGE_API_KEYS = 'manage_api_keys'
 PLATFORM_ADMIN = 'platform_admin'
 VIEW_ACTIVITY = 'view_activity'
+CREATE_BROADCASTS = 'create_broadcasts'
+APPROVE_BROADCASTS = 'approve_broadcasts'
+CANCEL_BROADCASTS = 'cancel_broadcasts'
+REJECT_BROADCASTS = 'reject_broadcasts'
 
 # List of permissions
 PERMISSION_LIST = [
@@ -1881,7 +1885,12 @@ PERMISSION_LIST = [
     SEND_LETTERS,
     MANAGE_API_KEYS,
     PLATFORM_ADMIN,
-    VIEW_ACTIVITY]
+    VIEW_ACTIVITY,
+    CREATE_BROADCASTS,
+    APPROVE_BROADCASTS,
+    CANCEL_BROADCASTS,
+    REJECT_BROADCASTS,
+]
 
 
 class Permission(db.Model):
--- a/migrations/versions/0359_more_permissions.py
+++ b/migrations/versions/0359_more_permissions.py
@@ -1,0 +1,50 @@
+"""
+
+Revision ID: 0359_more_permissions
+Revises: 0358_operator_channel
+Create Date: 2021-06-15 17:47:16.871071
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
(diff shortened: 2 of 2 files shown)


app/models.py
```diff
@@ -1837,6 +1837,8 @@ def serialize(self):
 MANAGE_API_KEYS = 'manage_api_keys'
 PLATFORM_ADMIN = 'platform_admin'
 VIEW_ACTIVITY = 'view_activity'
+CREATE_BROADCASTS = 'create_broadcasts'
+APPROVE_BROADCASTS = 'approve_broadcasts'
```
idavidmcdonald, line 1841: I'm not looking properly at this PR but just worth pointing out that I assume this means that the approve broadcasts permission also is permission to reject broadcasts too?

Wonder if it is worth making that explicit in the name of it?

Note, I don't feel too strongly about this. @quis might have opinions

klssmith: Yes, it would let you approve, cancel or reject. These names will be hard to change once added, so it would be good to get them right now - I can't think of a good name which covers all three actions.

idavidmcdonald: Interesting. I hadn't assumed it would also give them permission to cancel an alert...

quis: I think creators or approvers should be able to cancel. Is it worth making it a separate permission in the DB if it’s going to be ambiguous rolling it into another permission?

klssmith: Now updated as discussed on Slack to add separate `cancel_broadcasts` and `reject_broadcasts` permissions.

idavidmcdonald (approved):
I've run this locally on both the upgrade and downgrade with no problems.

Worth pointing out in my research to make sure this was good a few things

https://stackoverflow.com/questions/56942421/how-can-i-add-a-new-value-to-an-enum-in-postgres-without-locking-the-table where it points out that this is not a table locking operation for the upgrade (although the original poster appears to say that our downgrade might be)

https://blog.yo1.dog/updating-enum-values-in-postgresql-the-safe-and-easy-way/

klssmith:
We've used the way I've done things in the downgrade before a few times for both an [upgrade and and a downgrade](https://github.com/alphagov/notifications-api/blob/master/migrations/versions/0323_broadcast_message.py) so hopefully if we did need to downgrade that wouldn't cause issues.

It's `ALTER TYPE ... ADD VALUE` that I'm trying in a migration for the first time - worked on my machine though 🤞

app/models.py
```diff
@@ -1848,7 +1852,12 @@ def serialize(self):
     SEND_LETTERS,
     MANAGE_API_KEYS,
     PLATFORM_ADMIN,
-    VIEW_ACTIVITY]
+    VIEW_ACTIVITY,
+    CREATE_BROADCASTS,
```
servingUpAces, line 1856: The button to create a broadcast is labelled `Prepare broadcast` should we use `prepare_broadcast`. I don't have a strong feeling but am wondering if reflecting the wording on the UI is better.

quis: We’re probably going to review the language in the UI so I think `create` is OK.

PR #3276, opened 2021-06-22
Correct the daily limits cache.

servingUpAces:
Last year we had an issue with the daily limit cache and the query that was populating it. As a result we have not been checking the daily limit properly. This PR should correct all that.

The daily limit cache is not being incremented in app.notifications.process_notifications.persist_notification, this method is and should always be the only method used to create a notification.
We increment the daily limit cache is redis is enabled (and it is always enabled for production) and the key type for the notification is team or normal.

We check if the daily limit is exceed in many places:
 - app.celery.tasks.process_job
 -  app.v2.notifications.post_notifications.post_notification
 - app.v2.notifications.post_notifications.post_precompiled_letter_notification
 - app.service.send_notification.send_one_off_notification
 - app.service.send_notification.send_pdf_letter_notification

If the daily limits cache is not found, set the cache to 0 with an expiry of 24 hours. The daily limit cache key is service_id-yyy-mm-dd-count, so each day a new cache is created.

The best thing about this PR is that the app.service_dao.fetch_todays_total_message_count query has been removed. This query did not perform well and had been wrong for ages.

Before we merge this PR we will need to increase the daily limit for some of the services that we expect will exceed their current daily limit.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -161,8 +161,11 @@ def process_row(row, template, job, service, sender_id=None):
 
 def __sending_limits_for_job_exceeded(service, job, job_id):
     try:
-        check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
-        return False
+        total_sent = check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
+        if total_sent + job.notification_count > service.message_limit:
+            raise TooManyRequestsError(service.message_limit)
+        else:
+            return False
     except TooManyRequestsError:
         job.job_status = 'sending limits exceeded'
         job.processing_finished = datetime.utcnow()
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -33,7 +33,6 @@ from app.dao.returned_letters_dao import insert_or_update_returned_letters
 from app.dao.service_email_reply_to_dao import dao_get_reply_to_by_id
 from app.dao.service_inbound_api_dao import get_service_inbound_api_for_service
 from app.dao.service_sms_sender_dao import dao_get_service_sms_senders_by_id
-from app.dao.services_dao import fetch_todays_total_message_count
 from app.dao.templates_dao import dao_get_template_by_id
 from app.exceptions import DVLAException, NotificationTechnicalFailureException
 from app.models import (
@@ -55,6 +54,7 @@ from app.models import (
     DailySortedLetter,
 )
 from app.notifications.process_notifications import persist_notification
+from app.notifications.validators import get_service_daily_limit_cache_value
 from app.serialised_models import SerialisedService, SerialisedTemplate
 from app.service.utils import service_allowed_to_send_to
 from app.utils import DATETIME_FORMAT, get_reference_from_personalisation
@@ -159,8 +159,8 @@ def process_row(row, template, job, service, sender_id=None):
 
 
 def __sending_limits_for_job_exceeded(service, job, job_id):
-    total_sent = fetch_todays_total_message_count(service.id)
-
+    total_sent = get_service_daily_limit_cache_value(KEY_TYPE_NORMAL, service)
+    print(total_sent)
     if total_sent + job.notification_count > service.message_limit:
         job.job_status = 'sending limits exceeded'
         job.processing_finished = datetime.utcnow()
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -54,10 +54,11 @@ from app.models import (
     DailySortedLetter,
 )
 from app.notifications.process_notifications import persist_notification
-from app.notifications.validators import get_service_daily_limit_cache_value
+from app.notifications.validators import check_service_over_daily_message_limit
 from app.serialised_models import SerialisedService, SerialisedTemplate
 from app.service.utils import service_allowed_to_send_to
 from app.utils import DATETIME_FORMAT, get_reference_from_personalisation
+from app.v2.errors import TooManyRequestsError
 
 
 @notify_celery.task(name="process-job")
@@ -159,9 +160,10 @@ def process_row(row, template, job, service, sender_id=None):
 
 
 def __sending_limits_for_job_exceeded(service, job, job_id):
-    total_sent = get_service_daily_limit_cache_value(KEY_TYPE_NORMAL, service)
-    print(total_sent)
-    if total_sent + job.notification_count > service.message_limit:
+    try:
+        check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
+        return False
+    except TooManyRequestsError:
         job.job_status = 'sending limits exceeded'
         job.processing_finished = datetime.utcnow()
         dao_update_job(job)
@@ -170,7 +172,6 @@ def __sending_limits_for_job_exceeded(service, job, job_id):
                 job_id, job.notification_count, service.message_limit)
         )
         return True
-    return False
 
 
 @notify_celery.task(bind=True, name="save-sms", max_retries=5, default_retry_delay=300)
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -148,10 +148,8 @@ def persist_notification(
     # if simulated create a Notification model to return but do not persist the Notification to the dB
     if not simulated:
         dao_create_notification(notification)
-        # Only keep track of the daily limit for trial mode services.
-        if service.restricted and key_type != KEY_TYPE_TEST:
-            if redis_store.get(redis.daily_limit_cache_key(service.id)):
-                redis_store.incr(redis.daily_limit_cache_key(service.id))
+        if key_type != KEY_TYPE_TEST and current_app.config['REDIS_ENABLED']:
+            redis_store.incr(redis.daily_limit_cache_key(service.id))
 
         current_app.logger.info(
             "{} {} created at {}".format(notification_type, notification_id, notification_created_at)
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -149,8 +149,15 @@ def persist_notification(
     if not simulated:
         dao_create_notification(notification)
         if key_type != KEY_TYPE_TEST and current_app.config['REDIS_ENABLED']:
-            redis_store.incr(redis.daily_limit_cache_key(service.id))
-
+            cache_key = redis.daily_limit_cache_key(service.id)
+            if redis_store.get(cache_key) is None:
+                # if cache does not exist set the cache to 1 with an expiry of 24 hours,
+                # The cache should be set by the time we create the notification
+                # but in case it is this will make sure the expiry is set to 24 hours,
+                # where if we let the incr method create the cache it will be set a ttl.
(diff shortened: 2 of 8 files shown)


tests/app/notifications/test_validators.py
```diff
@@ -55,135 +56,66 @@ def enable_redis(notify_api):
...
-    assert not app.notifications.validators.services_dao.mock_calls
+    mock_get.assert_called_once_with(f'{serialised_service.id}-{datetime.utcnow().strftime("%Y-%m-%d")}-count')
+    assert mock_get.called
+    assert not mock_set.called
 
 
 def test_should_not_interact_with_cache_for_test_key(sample_service, mocker):
     mocker.patch('app.notifications.validators.redis_store')
-    mocker.patch('app.notifications.validators.redis_store.get', side_effect=[None])
+    mock_get = mocker.patch('app.notifications.validators.redis_store.get', side_effect=[None])
```
benthorner, line 75: Minor: do we need this? We're not calling the `redis_store`, and we've got an assertion for that later on. I'm not sure why we need two assertions (and the extra mock).

tests/app/notifications/test_validators.py
```diff
@@ -55,135 +56,66 @@ def enable_redis(notify_api):
...
-        f'service-{serialised_service.id}',
-        ANY,
-        ex=ANY,
-    )
-    assert not app.notifications.validators.services_dao.mock_calls
-
-
-@pytest.mark.parametrize('key_type', ['test', 'team', 'normal'])
+@pytest.mark.parametrize('key_type', ['team', 'normal'])
 def test_check_service_message_limit_in_cache_under_message_limit_passes(
```
benthorner, line 60: Slightly out-of-scope: can we fix the test function names? Normally the first part is the name of the function under test, which in this case is `check_service_over_daily_message_limit`.

Later on, some of the test functions are missing the name altogether e.g. `test_should_not_interact_with_cache_for_test_key`. Can we fix those too?

tests/app/notifications/test_validators.py
```diff
@@ -511,6 +440,96 @@ def test_should_not_rate_limit_if_limiting_is_disabled(
         assert not app.redis_store.exceeded_rate_limit.called
 
 
+def test_check_rate_limiting_checks_daily_limit_if_trial_mode_service(
```
benthorner, line 443: Minor: could we reduce the number of tests by just doing both assertions for the method calls in a single test, instead of spreading them across two? Also, we could combine them further by parametrizing on `restricted=True`. Then we'd just have one block of code, which would be much easier to read / review.

app/notifications/process_notifications.py
```diff
@@ -149,8 +149,15 @@ def persist_notification(
...
         if key_type != KEY_TYPE_TEST and current_app.config['REDIS_ENABLED']:
-            redis_store.incr(redis.daily_limit_cache_key(service.id))
-
+            cache_key = redis.daily_limit_cache_key(service.id)
+            if redis_store.get(cache_key) is None:
+                # if cache does not exist set the cache to 1 with an expiry of 24 hours,
+                # The cache should be set by the time we create the notification
+                # but in case it is this will make sure the expiry is set to 24 hours,
+                # where if we let the incr method create the cache it will be set a ttl.
+                redis_store.set(cache_key, 1, ex=86400)
```
benthorner, line 158: Can we have a test for this bit?

benthorner (commented):
Nearly good to go 🚀. 

I've got a few more major comments, and some minor ones I'd like your thoughts on too. By the way, I would have found this one a bit easier to review if the changes were spread over a few commits.

From a brief scan of the Admin app, it looks like all the code is there to show error messages for jobs and one-off notifications  👍. (I wasn't sure as you mentioned previously the feature has been broken for a while.)

tests/app/celery/test_tasks.py
```diff
@@ -181,6 +182,7 @@ def test_should_not_process_sms_job_if_would_exceed_send_limits_inc_today(
     mocker.patch('app.celery.tasks.s3.get_job_and_metadata_from_s3',
                  return_value=(load_example_csv('sms'), {'sender_id': None}))
     mocker.patch('app.celery.tasks.process_row')
+    mocker.patch('app.celery.tasks.get_service_daily_limit_cache_value', return_value=2)
```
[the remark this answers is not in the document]
servingUpAces: I'm not either. I've changed the tests to have one test for when the limit is exceeded and one for when the limit is not exceeded.

tests/app/celery/test_tasks.py
```diff
@@ -232,6 +235,7 @@ def test_should_process_email_job_if_exactly_on_send_limits(notify_db_session,
     mocker.patch('app.celery.tasks.save_email.apply_async')
     mocker.patch('app.encryption.encrypt', return_value="something_encrypted")
     mocker.patch('app.celery.tasks.create_uuid', return_value="uuid")
+    mocker.patch('app.celery.tasks.get_service_daily_limit_cache_value', return_value=0)
```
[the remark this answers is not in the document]
servingUpAces: no, I've removed it

tests/app/notifications/test_process_notification.py
```diff
@@ -198,77 +196,48 @@ def test_persist_notification_with_optionals(sample_job, sample_api_key):
...
+            template_version=template.version,
+            recipient='+447111111122',
+            service=template.service,
+            personalisation={},
+            notification_type='sms',
+            api_key_id=api_key.id,
+            key_type=api_key.key_type,
+            reference="ref2")
+
+        assert mock_incr.called
```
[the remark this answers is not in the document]
servingUpAces: ok

tests/app/notifications/test_validators.py
servingUpAces: I've fixed all the function names

tests/app/notifications/test_validators.py
servingUpAces: refactor of the tests has been done

tests/app/notifications/test_validators.py
```diff
@@ -55,135 +55,64 @@ def enable_redis(notify_api):
...
+    mock_set = mocker.patch('app.notifications.validators.redis_store.set')
     check_service_over_daily_message_limit(key_type, serialised_service)
-    app.notifications.validators.redis_store.set.assert_called_once_with(
-        f'service-{serialised_service.id}',
-        ANY,
-        ex=ANY,
-    )
-    assert not app.notifications.validators.services_dao.mock_calls
+    mock_get.assert_called_once_with(f'{serialised_service.id}-{datetime.utcnow().strftime("%Y-%m-%d")}-count')
+    assert mock_get.called
```
benthorner, line 68: We don't need this: it's superseded by the line above.

tests/app/notifications/test_validators.py
benthorner: I think this one got missed, by the way: we're still mocking the Redis get operation twice (and asserting on it twice).

tests/app/notifications/test_validators.py
```diff
@@ -55,135 +55,64 @@ def enable_redis(notify_api):
...
-        assert e.value.message == 'Exceeded send limits (4) for today'
-        assert e.value.fields == []
-        app.notifications.validators.redis_store.set.assert_called_once_with(
-            f'service-{serialised_service.id}',
-            ANY,
-            ex=ANY,
-        )
-        assert not app.notifications.validators.services_dao.mock_calls
+def test_check_service_message_limit_over_message_limit_fails(key_type, mocker, notify_db_session):
+    service = create_service(restricted=True, message_limit=4)
```
benthorner, line 108: Minor: the `restricted` setting is irrelevant for this test, so we don't need it (or we could parametrize on it if we're worried about it having an effect).

tests/app/notifications/test_validators.py
benthorner: Nice, thanks 🥇

app/celery/tasks.py
```diff
@@ -159,9 +160,10 @@ def process_row(row, template, job, service, sender_id=None):
 
 
 def __sending_limits_for_job_exceeded(service, job, job_id):
-    total_sent = fetch_todays_total_message_count(service.id)
-
-    if total_sent + job.notification_count > service.message_limit:
+    try:
+        check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
```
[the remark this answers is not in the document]
servingUpAces: Good point I've fixed the function so it does the right thing

app/notifications/process_notifications.py
```diff
@@ -148,11 +148,16 @@ def persist_notification(
     # if simulated create a Notification model to return but do not persist the Notification to the dB
     if not simulated:
         dao_create_notification(notification)
-        # Only keep track of the daily limit for trial mode services.
-        if service.restricted and key_type != KEY_TYPE_TEST:
-            if redis_store.get(redis.daily_limit_cache_key(service.id)):
-                redis_store.incr(redis.daily_limit_cache_key(service.id))
-
+        if key_type != KEY_TYPE_TEST and current_app.config['REDIS_ENABLED']:
```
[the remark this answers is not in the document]
servingUpAces: I'm going to leave this for now

app/notifications/validators.py
```diff
@@ -59,24 +58,27 @@ def check_service_over_api_rate_limit(service, api_key):
...
+    if key_type == KEY_TYPE_TEST or not current_app.config['REDIS_ENABLED']:
+        return 0
+
+    cache_key = daily_limit_cache_key(service.id)
+    service_stats = redis_store.get(cache_key)
+    if service_stats is None:
+        # first message of the day, set the cache to 0 and the expiry to 24 hours
+        service_stats = 0
+        redis_store.set(cache_key, service_stats, ex=86400)
+        return service_stats
```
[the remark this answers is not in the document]
servingUpAces: Now using the service_stats value in the process_job task.

tests/app/celery/test_tasks.py
```diff
@@ -182,6 +182,25 @@ def test_should_not_process_if_send_limit_is_exceeded(
...
+    mocker.patch('app.celery.tasks.s3.get_job_and_metadata_from_s3',
+                 return_value=(load_example_csv('multiple_sms'), {'sender_id': None}))
+    mocker.patch('app.celery.tasks.process_row')
+    mocker.patch('app.celery.tasks.check_service_over_daily_message_limit',
+                 return_value=0)
+    process_job(job.id)
+
+    job = jobs_dao.dao_get_job_by_id(job.id)
+    assert job.job_status == 'sending limits exceeded'
+    assert s3.get_job_and_metadata_from_s3.called is False
```
benthorner: Minor: we could assign the mock to a variable and use that directly here instead. Here, it's unclear where `s3` is coming from.

(I can see it's meant to be similar to [this other test](https://github.com/alphagov/notifications-api/blob/527c54fb9bb3dd46c646c9f7baca585550f194a1/tests/app/celery/test_tasks.py#L116).)

Similarly for the one below.

benthorner (approved):
This looks good to go 👍. I've left three comments, but they're all minor ones.

PR #3277, opened 2021-06-22
Remove user permissions if service becomes a broadcast service

klssmith:
The "normal" service permissions and broadcast service permissions are
going to be different with no overlap. This means that if you were
viewing the team members page, there might be permissions in the
database that are not visible on the frontend if a service has changed
type. For example, someone could have the 'manage_api_keys' permission,
which would not show up on the team members page of a broadcast service.
To avoid people having permissions which aren't visible in admin, we now
remove all permissions from users when their service is converted to a
broadcast service.

Permisions for invited users are also removed.

It's not possible to convert a broadcast service to a normal service, so
we don't need to cover for this scenario.

### DIFF
--- a/app/dao/broadcast_service_dao.py
+++ b/app/dao/broadcast_service_dao.py
@@ -7,7 +7,10 @@ from app.dao.dao_utils import autocommit, version_class
 from app.models import (
     BROADCAST_TYPE,
     EMAIL_AUTH_TYPE,
+    INVITE_PENDING,
+    InvitedUser,
     Organisation,
+    Permission,
     Service,
     ServiceBroadcastSettings,
     ServicePermission,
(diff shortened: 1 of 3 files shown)


benthorner (commented):
This also removes all permissions if the broadcast settings change later on. In theory that's OK, because our plan is to avoid changing the settings once a service is setup.

Do we want to reset the permissions every time? Either way, I think it would be worth adding a quick warning to the final screen in the admin app, so it's not a surprise.

(I was wondering about the impact of this if a service had lots of members. This shouldn't be a problem, as long as we only add members after the service is setup.)

quis:
Agree. I think it’s reasonable that if a service is moved from the test channel to the severe channel, or from one MNO to all, that we are forced to confirm that those are still the right people to be creating and approving alerts.

klssmith:
I had forgotten that this code is also used when changing broadcast settings and not just when making a service a broadcast service for the first time. If we think it is ok that all team member permissions need to be reset even when switching between operator, live and government channels then I can add something to admin to try and make it clear what will happen, perhaps a new paragraph on the confirmation page

Screenshot 2021-06-23 at 13 42 47

Otherwise, if that would cause issues I can have a bit of a rewrite of this code.

benthorner (approved):
No worries @klssmith 🙂 . I think the code is good to go, and we've got a second opinion from @quis, so I think it's OK for this to go in if we make the change to Admin soonish.

PR #3279, opened 2021-06-23
Fix the endpoint for the monthly status report

klssmith:
This wasn't working - the error given when trying to access it was `TypeError: Object of type 'Row' is not JSON serializable` when we tried to serialize a SQLAlchemy Row.

I haven't looked too far into what has changed to stop this from working, but have just changed the endpoint to return a nested list instead.

### DIFF
--- a/tests/app/service/test_rest.py
+++ b/tests/app/service/test_rest.py
@@ -3512,22 +3512,21 @@ def test_cancel_notification_for_service_updates_letter_if_still_time_to_cancel(
     assert response['status'] == 'cancelled'
 
 
-def test_get_monthly_notification_data_by_service(mocker, admin_request):
-    dao_mock = mocker.patch(
-        'app.service.rest.fact_notification_status_dao.fetch_monthly_notification_statuses_per_service',
-        return_value=[])
-
-    start_date = '2019-01-01'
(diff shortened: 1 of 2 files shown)


tests/app/service/test_rest.py
```diff
@@ -3512,22 +3512,21 @@ def test_cancel_notification_for_service_updates_letter_if_still_time_to_cancel(
...
     response = admin_request.get(
         'service.get_monthly_notification_data_by_service',
-        start_date=start_date,
-        end_date=end_date
+        start_date='2019-01-01',
+        end_date='2019-06-17'
     )
 
-    dao_mock.assert_called_once_with(start_date, end_date)
-    assert response == []
```
benthorner, line 3530: I'm guessing this function wasn't written using TDD 😄.

benthorner (approved):
Nice 🥇.

(I wonder if this got missed in https://github.com/alphagov/notifications-api/commit/0161adca063ee58d2ee38b394a646e6afb8ebade.)

tests/app/service/test_rest.py
klssmith: When checking the Git history earlier, I quickly found that I am the guilty party here 😄

PR #3280, opened 2021-06-25
Add flag to say if user is eligible for WebAuthn

benthorner:
https://www.pivotaltracker.com/story/show/178609960

Currently we have some data-driven roles to say who can use this feature.
Adding a flag in the API means we can avoid API calls in the Admin app to
determine the same.

Allowing members of the GOV.UK Notify service to use the feature is a
workaround, so we can avoid making someone a Platform Admin before they've
protected their account with it.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -145,6 +145,9 @@ class User(db.Model):
         if self.platform_admin:
             return True
 
+        if self.auth_type == 'webauthn_auth':
+            return True
+
         return any(
             str(service.organisation_id) == current_app.config['BROADCAST_ORGANISATION_ID'] or
             str(service.id) == current_app.config['NOTIFY_SERVICE_ID']
--- a/app/models.py
+++ b/app/models.py
@@ -140,6 +140,17 @@ class User(db.Model):
     def password(self):
         raise AttributeError("Password not readable")
 
+    @property
+    def can_use_webauthn(self):
+        if self.platform_admin:
+            return True
+
+        return any(
+            str(service.organisation_id) == current_app.config['BROADCAST_ORGANISATION_ID'] or
+            str(service.id) == current_app.config['NOTIFY_SERVICE_ID']
+            for service in self.services
+        )
+
     @password.setter
     def password(self, password):
         self._password = hashpw(password)
@@ -179,6 +190,7 @@ class User(db.Model):
             'permissions': self.get_permissions(),
             'platform_admin': self.platform_admin,
             'services': [x.id for x in self.services if x.active],
+            'can_use_webauthn': self.can_use_webauthn,
             'state': self.state,
         }
 
--- a/tests/app/user/test_rest.py
+++ b/tests/app/user/test_rest.py
@@ -72,6 +72,7 @@ def test_get_user(admin_request, sample_service, sample_organisation):
     assert fetched['permissions'].keys() == {str(sample_service.id)}
     assert fetched['services'] == [str(sample_service.id)]
     assert fetched['organisations'] == [str(sample_organisation.id)]
+    assert fetched['can_use_webauthn'] is False
     assert sorted(fetched['permissions'][str(sample_service.id)]) == sorted(expected_permissions)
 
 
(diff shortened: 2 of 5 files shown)


app/models.py
```diff
@@ -140,6 +140,17 @@ class User(db.Model):
...
+    @property
+    def can_use_webauthn(self):
+        if self.platform_admin:
+            return True
+
+        return any([
+            str(service.organisation_id) == current_app.config['BROADCAST_ORGANISATION_ID'] or
+            str(service.id) == current_app.config['NOTIFY_SERVICE_ID']
+            for service in self.services
+        ])
```
leohemsted, line 152: You can remove the square brackets form this, that way it'll be passed in as a generator rather than evaluated in a list comprehension, which will speed it up as `any` can shortcut return as soon as the first matching service is seen.

tests/app/user/test_rest.py
```diff
@@ -72,6 +72,7 @@ def test_get_user(admin_request, sample_service, sample_organisation):
     assert fetched['permissions'].keys() == {str(sample_service.id)}
     assert fetched['services'] == [str(sample_service.id)]
     assert fetched['organisations'] == [str(sample_organisation.id)]
+    assert not fetched['can_use_webauthn']
```
leohemsted, line 75: Minor: You could consider changing this assert to `is False`, as that way it excludes a variety of things (empty lists/dicts/strings and None) that we wouldn't expect on the front end.

```suggestion
    assert fetched['can_use_webauthn'] is False
```

app/models.py
```diff
@@ -140,6 +140,17 @@ class User(db.Model):
...
         raise AttributeError("Password not readable")
 
+    @property
+    def can_use_webauthn(self):
+        if self.platform_admin:
+            return True
+
+        return any([
+            str(service.organisation_id) == current_app.config['BROADCAST_ORGANISATION_ID'] or
+            str(service.id) == current_app.config['NOTIFY_SERVICE_ID']
```
leohemsted, line 150: Who is this check for? What's the flow? have i understood this right?

* new user joins notify team
* A new manual step: an existing platform admin invites them to the notify service
* they can now use webauthn, so set up their security key
* we enable platform admin in the db after verifying they're set up to use webauthn_auth
* They can leave the notify service if they want

*repeat for all three environments*

if we're starting to automate this flow, is this the process we want to use? We still need to go into the database manually to set someone to platform_admin, so it's not a fully automated process (although this does remove one manual prod database write access, which is nice). I guess we won't use this process too much anyway so I'm not too worried.

tests/app/user/test_rest.py
benthorner: Ahh, I previously tried `== False`, but Flake didn't like that (and suggested "if not cond").

Agree it's better to be explicit 👍.

app/models.py
benthorner: Nice, that's a bit shorter! 

I'd forgotten about generators, and also that you don't need extra parentheses for them inside a function call.

app/models.py
benthorner: I mentioned this in the 2nd commit:

> Allowing members of the GOV.UK Notify service to use the feature
> is a workaround, so we can avoid making someone a Platform Admin
> before they've protected their account with [WebAuthn].

> if we're starting to automate this flow, is this the process we want to use?

No, definitely not 😄. However, I don't think we're going to get to a full roll out of WebAuthn for a long time. This isn't so much about making the process easier, as making it more secure (see above).

I've added a task to the card to remind me to update the guidance for new Platform Admins, if this goes in.

tests/app/user/test_rest.py
leohemsted: yeah, frustratingly pep8 and various guidance on blogs/stackoverflow is always tilted towards `if` statements in code, where good variable naming and sensible return values can help deal with that, rather than asserts, where we should be as explicit as possible

leohemsted (approved):

benthorner:
@leohemsted I've added a further commit to simplify the PR I'm doing in the Admin app, by doing more of the checking here. 

Could you take another look?

leohemsted (approved):

PR #3281, opened 2021-06-29
Fix bug in check_service_over_daily_limit

servingUpAces:
I forgot to return service_stats if the cache exists. And fixed the tests to check service_stats value.


app/notifications/validators.py
```diff
@@ -74,6 +74,7 @@ def check_service_over_daily_message_limit(key_type, service):
                 service.id, int(service_stats), service.message_limit)
         )
         raise TooManyRequestsError(service.message_limit)
+    return service_stats
```
benthorner, line 77: Minor: if you have this, you don't need the earlier `return`.

benthorner (approved):
👍

PR #3284, opened 2021-06-30
Remove original permissions from broadcast users

klssmith:
The broadcast user permissions are changing, so to avoid confusion and permissions which exist in the database but don't display on the frontend we are going to remove all existing permissions for users of broadcast services. This also updates the permissions of invited users who are still pending.

The exception to this is the `view_activity` permission, which we [always add for broadcast users](https://github.com/alphagov/notifications-admin/blob/aad017a184d314d9d192f4b9d6485915f83217c1/app/main/forms.py#L1043) even if they have no other permissions.

[Pivotal story](https://www.pivotaltracker.com/story/show/178482657)

### DIFF
--- a/migrations/versions/0361_new_user_bcast_permissions.py
+++ b/migrations/versions/0361_new_user_bcast_permissions.py
@@ -1,0 +1,32 @@
+"""
+
+Revision ID: 0361_new_user_bcast_permissions
+Revises: 0360_remove_sched_notifications
+Create Date: 2021-06-30 11:42:32.780734
+
+"""
+from alembic import op
+
+revision = '0361_new_user_bcast_permissions'
+down_revision = '0360_remove_sched_notifications'
+
+
+def upgrade():
+    """
+    Delete all permissions for broadcast service users and invited pending users, apart from 'view_activity'
+    which they always have.
+    """
+    op.execute(
+        "DELETE FROM permissions WHERE permission != 'view_activity' "
+        "and service_id in (select id from services where organisation_id = '38e4bf69-93b0-445d-acee-53ea53fe02df')"
+    )
+    op.execute(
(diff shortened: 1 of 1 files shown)


migrations/versions/0360_new_user_bcast_permissions.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Delete all permissions for broadcast service users and invited pending users, apart from 'view_activity'
+    which they always have.
+    """
+    op.execute(
+        "DELETE FROM permissions WHERE permission != 'view_activity' "
+        "and service_id in (select id from services where organisation_id = '38e4bf69-93b0-445d-acee-53ea53fe02df')"
+    )
+    op.execute(
+        "UPDATE invited_users SET permissions = 'view_activity' WHERE status = 'pending' "
```
benthorner, line 24: This will give invited users the view permission even if they didn't have it before. I don't think that's a big deal, though: I'm not aware of any reason why we need to hide anything from particular broadcast service users.

benthorner (approved):
Looks good to me 👍.

(I double-checked the UUID matches BROADCAST_ORGANISATION_ID.)

migrations/versions/0360_new_user_bcast_permissions.py
klssmith: The `view_activity` permission is [always added](https://github.com/alphagov/notifications-admin/blob/256c840b46d4095edbd0af105e5c1a340242575d/app/main/forms.py#L1042) to the list of permissions when either the `BroadcastPermissionsForm` or the `BroadcastInviteUserForm` is submitted, so this keeping the same behaviour.

benthorner: Ahh, that's good to know, thanks.

I never noticed they were absent from the form before.

PR #3286, opened 2021-07-07
Bump utils to bring in fix for optional placeholder bug

CrystalPea:
See https://github.com/alphagov/notifications-utils/pull/878 for details.

Changes we had to make for our app and tests to work correctly after the dependency updates:

1. Update emergency alerts polygons test because we changed how exact we are with locations of the points on the map.

2. Use Flask's g object to set additional request attributes

So far we have been storing them in _request_ctx_stack which is an innard for Flask's request context.

Because of major update to Werkzeug dependency, which Flask relies on, the way we were using it stopped working, so we had a new way to set those values.
The way we set those values now, by using g object, seems to also be favoured in Flask documentation:
https://flask.palletsprojects.com/en/1.1.x/reqcontext/#how-the-context-works

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -7,7 +7,6 @@ from time import monotonic
 
 from celery import current_task
 from flask import (
-    _request_ctx_stack,
     current_app,
     g,
     has_request_context,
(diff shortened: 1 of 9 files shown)


leohemsted (approved):
we also use _request_ctx_stack in admin app so would be good to update that code too some time soon :)

PR #3289, opened 2021-07-13
retry service callbacks on 429

leohemsted:
if we're served a 429, put the item on the retry queue and retry the same as if the service returned a 5xx. 429 is commonly returned for rate limit exceeding, and retrying on a delay is a typical response to that.


benthorner (approved):
This makes sense, since otherwise the service will have to dig out the status updates manually somehow. It's unlikely to fix the problem entirely, since we only retry a few times at a regular interval (which could equally be a busy time). But really the service should make sure they have enough capacity to receive the callbacks.

So just the right amount of helpful 👍.

PR #3290, opened 2021-07-14
Always keep view activity permissions for broadcast users

klssmith:
We made a change to remove all permissions from users and invited users when the broadcast service settings form is submitted (#3284). However, when the form is submitted, notifications-admin always [adds the `view_activity`
permission](https://github.com/alphagov/notifications-admin/blob/256c840b46d4095edbd0af105e5c1a340242575d/app/main/forms.py#L1042) even if no permission boxes are ticked, so we don't want to remove that one permission.

### DIFF
--- a/app/dao/broadcast_service_dao.py
+++ b/app/dao/broadcast_service_dao.py
@@ -8,6 +8,7 @@ from app.models import (
     BROADCAST_TYPE,
     EMAIL_AUTH_TYPE,
     INVITE_PENDING,
+    VIEW_ACTIVITY,
     InvitedUser,
     Organisation,
     Permission,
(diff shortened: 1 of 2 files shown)


benthorner (approved):
This makes sense 👍.

It's got me wondering (in case you're interested as well): why do we have the `view_activity` permission? I can see we check for it specifically in a few places, but since it's always added, that means `@user_has_permission('view_activity')` is the same as `@user_has_permission()`, which [checks the user belongs to the service](https://github.com/alphagov/notifications-admin/blob/9170c0b17553a841dedcfb80a4ce943eef0b3478/app/models/user.py#L213).

Do you think we benefit from the permission?

klssmith:
It's something we use for non-broadcast services, though it's not really adding anything for broadcast services. I think it is helpful keep user permissions consistent for now and to go ahead with this change, but then to revisit whether we need that permission for broadcast services at all.

PR #3292, opened 2021-07-19
Revert werkzeug to the last non-2.0.0 version

sakisv:
We observe high memory usage since we bumped it (along with other
things) and because it only appears on the API and not on the workers
the hypothesis is that Werkzeug is responsible for it.


leohemsted (commented):
minor thing, can you stick a comment in the file explaining why it's pinned, so if we're bumping reqs we can easily see and go check changelogs to see if it's still an issue?

leohemsted (approved):

PR #3293, opened 2021-07-19
Trigger separate link test for failover lambda

benthorner:
https://www.pivotaltracker.com/story/show/178738544

Previously there was a risk that we'd only find out that the failover lambda
was broken at the point when we needed it. This rewrites the link test task
and its supporting code to make it easy to test each lambda independently.

Paired with @rjbaker

### DIFF
--- a/app/clients/cbc_proxy.py
+++ b/app/clients/cbc_proxy.py
@@ -1,4 +1,5 @@
 import json
+import uuid
 from abc import ABC, abstractmethod
 
 import boto3
@@ -77,11 +78,11 @@ class CBCProxyClientBase(ABC):
     def __init__(self, lambda_client):
         self._lambda_client = lambda_client
 
-    def send_link_test(
-        self,
-        identifier,
-    ):
-        pass
+    def send_link_tests(self):
+        self._send_link_test(self.lambda_name)
+        self._send_link_test(self.failover_lambda_name)
+
+    def _send_link_test(self): pass
 
     def create_and_send_broadcast(
         self, identifier, headline, description, areas, sent, expires, channel, message_number=None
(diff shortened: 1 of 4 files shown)


app/clients/cbc_proxy.py
```diff
@@ -118,6 +122,10 @@ def _invoke_lambda_with_failover(self, payload):
     def _invoke_lambda(self, lambda_name, payload):
         payload_bytes = bytes(json.dumps(payload), encoding='utf8')
         try:
+            current_app.logger.info(
+                f"Calling lambda {lambda_name} for link test with payload {payload}"
```
klssmith, line 126: It looks to me like this log line wouldn't just get used for link tests, but for all broadcasts (since `_invoke_lambda` is used for real broadcasts too) - is that correct?

If so having "link test" in the message is a bit confusing, plus the payload could be very large (though we do log the whole payload if there's an error, so that's not necessarily an issue)

rjbaker: Good spot thank you. This function is also called when sending regular messages. I've amended the log line 👍

klssmith (approved):
Change to the log line looks better 👍. It might not be practical to include the whole payload every time the lambda is invoked, because the payload includes `areas`, which could make it too big. (I realised that the payload we add to the logs when there's an error isn't actually the payload we send the lambda, it's part of the return value from the lambda).

An alternative would be to move that log line to `_send_link_test` instead (where payload will always be small) or just to log bits of the payload. Could always see what this looks like first and make changes if necessary later.

benthorner:
Thanks for spotting @klssmith. I think that's a reasonable concern to address now: we wouldn't want the code to fail when it was actually used (if the actual use case led to a large payload).

I couldn't find an obvious limit to the log length, but practically we wouldn't want to read anything much longer than 1000 chars, so I've capped it at that in a further commit. What do you think?

I think it's worth keeping the log where it is since it provides additional insight for all uses of the CBC proxy - some of these may be covered by higher level logs, but it's hard to audit that.

klssmith:
Thanks @benthorner , limiting the log line length seems like a good solution - it means we'll get the full payload logged for link tests, but more details than before for other message types

rjbaker:
Thanks @benthorner. Agree that's a useful place to keep the log line.

PR #3294, opened 2021-07-19
Auto expire old broadcast messages

benthorner:
https://www.pivotaltracker.com/story/show/178926353

Since the expiry is sent as part of the message payload, we don't
need to invoke the CBC proxies (and indeed there's no way to do so
for an expired alert). In future we plan to extend this task so it
triggers the regeneration of content on gov.uk/alerts.

It's worth noting that 'finishes_at' can theoretically be None, in
which case it's unclear when the alert should expire. While alerts
from the Admin app should always have an expiry [1], we have many
in the DB that don't, so it's worth checking for this scenario.

[1]: https://github.com/alphagov/notifications-admin/blob/078ac10c8df855a84c157190d43daa9a8897832b/app/models/broadcast_message.py#L255

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -4,7 +4,7 @@ from flask import current_app
 from sqlalchemy import between
 from sqlalchemy.exc import SQLAlchemyError
 
-from app import notify_celery, zendesk_client
+from app import db, notify_celery, zendesk_client
 from app.celery.broadcast_message_tasks import trigger_link_test
 from app.celery.letters_pdf_tasks import get_pdf_for_templated_letter
 from app.celery.tasks import (
@@ -48,6 +48,8 @@ from app.models import (
     JOB_STATUS_IN_PROGRESS,
     JOB_STATUS_PENDING,
     SMS_TYPE,
+    BroadcastMessage,
+    BroadcastStatusType,
     Job,
 )
 from app.notifications.process_notifications import send_notification_to_queue
@@ -304,3 +306,16 @@ def trigger_link_tests():
     if current_app.config['CBC_PROXY_ENABLED']:
         for cbc_name in current_app.config['ENABLED_CBCS']:
             trigger_link_test.apply_async(kwargs={'provider': cbc_name}, queue=QueueNames.BROADCASTS)
+
+
+@notify_celery.task(name='auto-expire-broadcast-messages')
+def auto_expire_broadcast_messages():
+    expired_broadcasts = BroadcastMessage.query.filter(
+        BroadcastMessage.finishes_at <= datetime.now(),
+        BroadcastMessage.status == BroadcastStatusType.BROADCASTING,
+    )
+
+    for broadcast in expired_broadcasts:
+        broadcast.status = BroadcastStatusType.COMPLETED
+
+    db.session.commit()
--- a/tests/app/celery/test_scheduled_tasks.py
+++ b/tests/app/celery/test_scheduled_tasks.py
@@ -8,6 +8,7 @@ from freezegun import freeze_time
 
 from app.celery import scheduled_tasks
 from app.celery.scheduled_tasks import (
+    auto_expire_broadcast_messages,
     check_for_missing_rows_in_completed_jobs,
     check_for_services_with_high_failure_rates_or_sending_to_tv_numbers,
     check_if_letters_still_in_created,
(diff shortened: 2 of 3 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -304,3 +306,17 @@ def trigger_link_tests():
...
+@notify_celery.task(name='auto-expire-broadcast-messages')
+def auto_expire_broadcast_messages():
+    expired_broadcasts = BroadcastMessage.query.filter(
+        BroadcastMessage.finishes_at <= datetime.now(),
+        BroadcastMessage.status == BroadcastStatusType.BROADCASTING,
+    )
+
+    for broadcast in expired_broadcasts:
+        broadcast.status = BroadcastStatusType.COMPLETED
+        broadcast.updated_at = datetime.now()
```
klssmith, line 320: `updated_at` will get set automatically [when the model is changed](https://github.com/alphagov/notifications-api/blob/03d438b0cc93dec4fe4a9b65b334deee8f48b5d8/app/models.py#L2309), so you could get away without manually setting it here

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -624,3 +631,27 @@ def test_trigger_link_does_nothing_if_cbc_proxy_disabled(
...
+    sample_template
+):
+    message = create_broadcast_message(
+        status=status,
+        finishes_at=finishes_at,
+        template=sample_template,
+    )
+
+    auto_expire_broadcast_messages()
+    assert message.status == final_status
```
klssmith, line 657: Could consider adding an assertion for the `updated_at` time here since that can also change

klssmith (commented):
This looked good to me, but it may be good to get another person to double check this just to get another check for any edge cases that could cause issues

app/celery/scheduled_tasks.py
benthorner: That's good to know. I'm used to setting it manually but I realise that's mostly from doing manual SQL for support purposes. To be fair there is [some inconsistency in where we've applied this](https://github.com/alphagov/notifications-api/blob/03d438b0cc93dec4fe4a9b65b334deee8f48b5d8/app/models.py#L784).

Anyway, I've amended the commit to remove this.

tests/app/celery/test_scheduled_tasks.py
benthorner: I did think about it, but it would make the parametrization more complex.

Fortunately it's no longer in scope as we discussed in https://github.com/alphagov/notifications-api/pull/3294#discussion_r673913382.

klssmith (approved):

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -624,3 +631,27 @@ def test_trigger_link_does_nothing_if_cbc_proxy_disabled(
         trigger_link_tests()
 
     assert mock_trigger_link_test.called is False
+
+
+@freeze_time('2021-07-19 15:50')
+@pytest.mark.parametrize('status, finishes_at, final_status', [
+    (BroadcastStatusType.BROADCASTING, '2021-07-19 16:00', BroadcastStatusType.BROADCASTING),
+    (BroadcastStatusType.BROADCASTING, '2021-07-19 15:40', BroadcastStatusType.COMPLETED),
+    (BroadcastStatusType.BROADCASTING, None, BroadcastStatusType.BROADCASTING),
```
sakisv, line 640: Not relevant to this PR, just out of curiosity, but is this a valid state where we're broadcasting but we haven't set a `finishes_at`?

sakisv (approved):
Looks good to me, though I'm in the same boat with Katie given my (lack of) knowledge of the codebase.

Just had a minor question.

tests/app/celery/test_scheduled_tasks.py
benthorner: Good question. While [we do set this time in the Admin app](https://github.com/alphagov/notifications-admin/blob/da976696747aa982d56e54eac3e14de7c64f6583/app/models/broadcast_message.py#L255), there's [nothing actually stopping the field from being `None`](https://github.com/alphagov/notifications-api/blob/master/app/broadcast_message/rest.py#L144). Although the Admin app is currently the only way to approve a broadcast, I wanted to make sure this code didn't misbehave if in future we have some other way of approving that doesn't set an expiry.

PR #3296, opened 2021-07-23
Show separate error for when user tries to cancel letter

CrystalPea:
Show separate error for when user tries to cancel letter that is already cancelled. 

This is to show a better error messages to our users, and also so we have more detailed log messages for related errors.

Relies on: https://github.com/alphagov/notifications-utils/pull/886

Needed for: https://github.com/alphagov/notifications-admin/pull/3984

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -2,7 +2,10 @@ import itertools
 from datetime import datetime
 
 from flask import Blueprint, current_app, jsonify, request
-from notifications_utils.letter_timings import letter_can_be_cancelled
+from notifications_utils.letter_timings import (
+    letter_can_be_cancelled,
+    too_late_to_cancel_letter,
+)
 from notifications_utils.timezones import convert_utc_to_bst
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
@@ -473,9 +476,17 @@ def cancel_notification_for_service(service_id, notification_id):
         raise InvalidRequest('Notification cannot be cancelled - only letters can be cancelled', status_code=400)
     elif not letter_can_be_cancelled(notification.status, notification.created_at):
         print_day = letter_print_day(notification.created_at)
-
+        if too_late_to_cancel_letter(notification.created_at):
+            message = "It’s too late to cancel this letter. Printing started {} at 5.30pm".format(print_day)
+        elif notification.status == 'cancelled':
+            message = "This letter has already been cancelled."
+        else:
+            message = (
+                f"We could not cancel this letter. "
+                f"Letter status: {notification.status}, created_at: {notification.created_at}"
+            )
         raise InvalidRequest(
-            "It’s too late to cancel this letter. Printing started {} at 5.30pm".format(print_day),
+            message,
             status_code=400)
 
     updated_notification = notifications_dao.update_notification_status_by_id(
--- a/tests/app/service/test_rest.py
+++ b/tests/app/service/test_rest.py
@@ -3445,6 +3445,7 @@ def test_cancel_notification_for_service_raises_invalid_request_when_letter_is_i
     notification_status,
 ):
     sample_letter_notification.status = notification_status
+    sample_letter_notification.created_at = datetime.now()
 
     response = admin_request.post(
         'service.cancel_notification_for_service',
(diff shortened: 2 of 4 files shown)


app/service/rest.py
```diff
@@ -473,9 +473,12 @@ def cancel_notification_for_service(service_id, notification_id):
         raise InvalidRequest('Notification cannot be cancelled - only letters can be cancelled', status_code=400)
     elif not letter_can_be_cancelled(notification.status, notification.created_at):
         print_day = letter_print_day(notification.created_at)
-
+        if too_late_to_cancel_letter(notification.created_at):
+            message = "It’s too late to cancel this letter. Printing started {} at 5.30pm".format(print_day)
+        else:
+            message = "We could not cancel this letter. Probably it was already cancelled."
```
benthorner, line 479: We should avoid ambiguity in our log messages, as this creates more work for someone reading them: was it cancelled, or wasn't it?

How about we just print the status of the notification - "We could not cancel this letter. It has status X and was created at Y. [Optional: The letter processing deadline is Z.]" - so the reader can figure it out for themselves?

One advantage of this is that we don't need the utils changes, which helps keep the complexity down over here.

benthorner (commented):
Nice work fixing the broken log 🥇.

I thought of a simpler way we could fix it that provides even more information, which I left a comment about. Can you see any advantages of leaving it as-is, though?

app/service/rest.py
quis: Does the user see these error messages or are they only for our logs?

CrystalPea: Hey @quis that's actually a great question! The reason this error was ending up in our logs was because we were not handling it right in the admin. I made a PR now to handle it: https://github.com/alphagov/notifications-admin/pull/3984 ☺️ 

Hey @benthorner I like the idea of using notification status and date, I added it for the case when we do not know why the error happened. Since the errors will be user facing now (check above message for Chris), we will need to tell them exactly what happened, so I think making the change in utils makes sense. Especially since no new code was added in utils, I just moved the code for checking if it's too late to a separate method. So I'd say it's more of a little refactor rather than adding complexity ☺️

app/service/rest.py
```diff
@@ -473,9 +476,17 @@ def cancel_notification_for_service(service_id, notification_id):
         raise InvalidRequest('Notification cannot be cancelled - only letters can be cancelled', status_code=400)
     elif not letter_can_be_cancelled(notification.status, notification.created_at):
         print_day = letter_print_day(notification.created_at)
-
+        if too_late_to_cancel_letter(notification.created_at):
+            message = "It’s too late to cancel this letter. Printing started {} at 5.30pm".format(print_day)
+        elif notification.status == 'cancelled':
+            message = "This latter has already been cancelled."
```
benthorner, line 482:
```suggestion
            message = "This letter has already been cancelled."
```

tests/app/service/test_rest.py
```diff
@@ -3445,14 +3445,21 @@ def test_cancel_notification_for_service_raises_invalid_request_when_letter_is_i
...
 
     response = admin_request.post(
         'service.cancel_notification_for_service',
         service_id=sample_letter_notification.service_id,
         notification_id=sample_letter_notification.id,
         _expected_status=400
     )
-    assert response['message'] == 'It’s too late to cancel this letter. Printing started today at 5.30pm'
+    if notification_status == 'cancelled':
+        assert response['message'] == 'This latter has already been cancelled.'
```
benthorner, line 3457: Optional: we could simplify this test to match the simplicity of the code e.g.

```
@pytest.mark.parametrize('status,expected_error', [
  ('cancelled', 'This letter...'),
  ('sent', 'We could not...')
])
```

I'm not sure what we gain from exhaustively testing every single possible status (apart from a lot more typing). But I appreciate this was there before and the PR is meant to be a quick fix, so happy to leave it as-is.

benthorner (approved):
There's a typo that needs fixing before this goes in ⌨️.

I'll approve now though.

tests/app/service/test_rest.py
CrystalPea: cool, I will leave it as is :3. I guess what we would gain is failing tests if we changed which statuses are cancellable.

PR #3297, opened 2021-07-26
Support granular API auth for internal apps

benthorner:
https://www.pivotaltracker.com/story/show/178774818

Previously we just had a single array of API keys / secrets, any of
which could be used to get passed the "requires_admin_auth" check.
While multiple keys are necessary to allow for rotation, we should
avoid giving other apps access this way (too much privilege).

This converts the existing config vars into a new dictionary, keyed
by client_id. We can then use the dictionary to scope auth for new
API consumers like gov.uk/alerts to just the endpoints they need to
access, while maintaining existing access for the Admin app.

Once the new dictionary is available as a JSON environment variable,
we'll be able to remove the old credentials / config. In the next
commits, we'll look at more tests for the new functionality.

TODO:

 - Move tests into own files
 - Remove ADMIN_CLIENT config var
 - Refactor to test internal auth method

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -67,30 +67,37 @@ def requires_no_auth():
 
 
 def requires_admin_auth():
+    requires_internal_auth(current_app.config.get('ADMIN_CLIENT_USER_NAME'))
+
+
+def requires_internal_auth(expected_client_id):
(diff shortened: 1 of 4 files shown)


benthorner:
Closing as we've discussed the approach and agreed to proceed.

I've written a new card to reflect the work (now on a new branch).

PR #3299, opened 2021-07-29
Revoke API keys when changing broadcast settings

quis:
On a regular Notify service anyone with permission can create an API key. If this service then is given permission to send emergency alerts it will have an API key which can create emergency alerts. This feels dangerous.

Secondly, if a service which legitimately has an API key for sending alerts in training mode is changed to live mode you now have an API key which people will think isn’t going to create a real alert but actually will. This feels really dangerous.

Neither of these scenarios are things we should be doing, but having them possible still makes me feel uncomfortable.

This commit revokes all API keys for a service when its broadcast settings change, same way we remove all permissions for its users.

### DIFF
--- a/app/dao/broadcast_service_dao.py
+++ b/app/dao/broadcast_service_dao.py
@@ -70,7 +70,8 @@ def set_broadcast_service_type(service, service_mode, broadcast_channel, provide
 
     # Revoke any API keys to avoid a regular API key being used to send alerts
     ApiKey.query.filter_by(
-        service_id=service.id
+        service_id=service.id,
+        expiry_date=None,
     ).update({
         ApiKey.expiry_date: datetime.utcnow()
     })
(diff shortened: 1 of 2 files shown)


app/dao/broadcast_service_dao.py
```diff
@@ -67,6 +68,13 @@ def set_broadcast_service_type(service, service_mode, broadcast_channel, provide
         status=INVITE_PENDING
     ).update({'permissions': VIEW_ACTIVITY})
 
+    # Revoke any API keys to avoid a regular API key being used to send alerts
+    ApiKey.query.filter_by(
+        service_id=service.id
```
benthorner, line 73: This will re-revoke any existing API keys as well, which is unnecessary / inaccurate.

Should we just revoke the active ones?

quis: Good point, fixed in 132411b

benthorner (approved):
Looks good 🇳🇿.

The comments in the tests make it easy to see what we're testing 💯.

PR #3301, opened 2021-07-30
Set jobs worker memory to 2GB

sakisv:
This is to see if the worker requires slightly more memory than it has access to or to determine if there is a memory  leak somewhere in the code that needs to be further investigated.

This comes at the heels of yesterday's issue that we could not process the CSVs the users uploaded, where the memory graph for this worker showed that it was using almost all of its available memory, so a redeploy fixed the problem.

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -50,7 +50,7 @@
   },
 
   'notify-delivery-celery-beat': {'memory': '128M'},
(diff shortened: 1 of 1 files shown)


benthorner (approved):
Nice explanation 👌.

(Interesting to see the existing memory variability among our workers!)

PR #3304, opened 2021-08-05
Switch to per-app secrets from internal APIs

benthorner:
https://www.pivotaltracker.com/story/show/179039225

Relates to: [1]

[1]: https://github.com/alphagov/notifications-credentials/pull/231


benthorner:
Checked on Staging to confirm that I can query the new `/v2/govuk-alerts` endpoint with the govuk-alerts API key, but not with the notify-admin key ✅. The Admin app also still works ✅.

leohemsted (approved):

PR #3305, opened 2021-08-10
 get_broadcasts returns a list of alerts for gov.uk/alerts

CrystalPea:
### In this PR we:

- add a dao query to get a list of alerts (only some properties, as required by gov.uk/alerts)
- return those alerts through an API endpoint get_broadcasts

### Considerations:
Is the date format resulting from conversion by jsonify ok? (check tests to see examples)

Flask dumps the date to a string in RFC 1123 format, which is consistent with dates in other parts of HTTP requests and responses, which sounds good, but is different than what we have in data.yaml file so far.

I would be up for leaving it as is and modifying gov.uk/alerts to accept dates in that format - I think it would be neater that way. Alternatively we could modify JSONEncoder to automatically convert the datetimes into the format we normally use (check second answer here): https://stackoverflow.com/questions/43663552/keep-a-datetime-date-in-yyyy-mm-dd-format-when-using-flasks-jsonify


app/v2/govuk_alerts/get_broadcasts.py
```diff
@@ -1,8 +1,11 @@
...
 
+from app.dao.broadcast_message_dao import dao_get_all_broadcast_messages
 from app.v2.govuk_alerts import v2_govuk_alerts_blueprint
 
 
 @v2_govuk_alerts_blueprint.route('')
 def get_broadcasts():
-    return jsonify({})
+    broadcasts = dao_get_all_broadcast_messages()
+    broadcasts_dict = {"alerts": [broadcast._asdict() for broadcast in broadcasts]}
```
leohemsted, line 10: rather than calling the sqlalchemy builtin _asdict, we should manually define the json that we will return. That way we can add, remove, and rename fields more easily. Also this will solve the datetime problem. 

In existing serialize functions in models.py we use `some_time.strftime(DATETIME_FORMAT)` and `get_dt_string_or_none(some_nullable_time)`, with `from app.utils import DATETIME_FORMAT, get_dt_string_or_none`. We should follow the same patterns here.

leohemsted (changes requested):
i also appreciate you didn't choose where the file is, but i'm also not entirely sure why this is in `/v2/`, which is currently for our public API endpoints, rather than placed with the rest of our internal endpoints (which are used by admin, doc-dl-frontend, and maybe some other places). I'd like us to change that please, although that doesn't necessarily have to be in this PR it should happen before we build too much more. @benthorner was there any particular reason why you made the endpoint in /v2/?

This does have an actual impact as well as just consistency, the one I can think of is that the error handlers are completely different between app/v2/errors.py and app/errors.py, both in terms of what exceptions we automatically handle and how the messages are returned to the caller, and I'd like us to be confident that we're calling the right thing 🤔

benthorner:
> @benthorner was there any particular reason why you made the endpoint in /v2/?

Yup: it's the latest (only) version of an API 😄.

I wasn't aware the "v2" namespace was reserved for public API endpoints. Looks like no one else is either, which suggests we're missing some documentation e.g. a README.md in the `app/v2/` directory, or (my preference) a more general file like `docs/apis.md` that we could extend to describe other areas. I don't really understand the objective reasons for not putting stuff in v2 for a new app like govuk-alerts, so it would be helpful capture the "why" as well as the "what".

How about we leave it as-is for this PR and move it in a follow-up where we also add (or link to) some documentation about it? Leaving it as-is will make this PR easier to review.

CrystalPea:
ok, so to sum up changes needed:

1. manually define the json and do datetime conversion in there

2. move the file to `app/govuk_alerts` so that it is like the rest of internal API endpoints (possibly in a separate PR, but I would be happy to do it in this one)

leohemsted:
definitely prefer to keep any moving the api endpoint around to its own separate PR 👍

app/v2/govuk_alerts/get_broadcasts.py
CrystalPea: @leohemsted yeah, I looked at get_dt_string_or_none before, but that didn't work mid-query. I thought about doing `cast` in query to change data types, but couldn't find good examples for Datetime to String.

 I also thought about creating a `serialize_for_gov_uk_alerts()` method in BroadcastMessage in `models.py`, but we needed some data outside of this model, namely `channel` from `ServiceBroadcastSettings`.
 
 I now did the json definition in `get_broadcasts` like you suggested and transformed the datetimes there, check if it meets the bill :). My only reservation is that doing it this way is probably slower than doing it from inside the query, but probably it won't matter that much, as hopefully we won't have enough alerts for it to matter?

tests/app/v2/govuk_alerts/test_get_broadcasts.py
```diff
@@ -0,0 +1,39 @@
...
+        template_1,
+        starts_at=datetime(2021, 6, 15, 12, 0, 0),
+        status='cancelled')
+
+    broadcast_message_2 = create_broadcast_message(
+        template_1,
+        starts_at=datetime(2021, 6, 22, 12, 0, 0),
+        status='broadcasting')
+
+    govuk_alerts_jwt_client_id = current_app.config['GOVUK_ALERTS_CLIENT_ID']
```
benthorner, line 24: Now https://github.com/alphagov/notifications-api/pull/3303 is in you can just do [this](https://github.com/alphagov/notifications-api/pull/3303/commits/09e3ba6836df0a930392a81dd3569f10ab4d04b6#diff-a705d52d5175ca6300ff24be4c10207e0c197282c3f89d39435b8fb1e46506a3R83).

CrystalPea: cool stuff, I pulled in those changes now :3

leohemsted (approved):
nice one

PR #3307, opened 2021-08-17
Don’t update `email_access_validated_at` on password reset

quis:
As of https://github.com/alphagov/notifications-admin/pull/4000 the admin app is doing this, so we don’t need to do it here as well.

### DIFF
--- a/tests/app/dao/test_users_dao.py
+++ b/tests/app/dao/test_users_dao.py
@@ -159,19 +159,13 @@ def test_update_user_attribute(client, sample_user, user_attribute, user_value):
 
 
 @freeze_time('2020-01-24T12:00:00')
-@pytest.mark.parametrize('from_email', [True, False])
-def test_update_user_password(notify_api, notify_db, notify_db_session, sample_user, from_email):
+def test_update_user_password(notify_api, notify_db, notify_db_session, sample_user):
     sample_user.password_changed_at = datetime.utcnow() - timedelta(days=1)
-    sample_user.email_access_validated_at = datetime.utcnow() - timedelta(days=1)
     password = 'newpassword'
     assert not sample_user.check_password(password)
-    update_user_password(sample_user, password, validated_email_access=from_email)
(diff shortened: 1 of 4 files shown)


tests/app/user/test_rest.py
```diff
@@ -791,30 +791,20 @@ def test_send_user_confirm_new_email_returns_400_when_email_missing(client, samp
...
-    ({
-        '_password': '1234567890',
-        'validated_email_access': False,
-    }, datetime(2020, 2, 13, 12, 0))
-])
 @freeze_time('2020-02-14T12:00:00')
-def test_update_user_password_saves_correctly(client, sample_service, data, email_access_validated_at):
+def test_update_user_password_saves_correctly(client, sample_service):
     sample_user = sample_service.users[0]
     sample_user.email_access_validated_at = datetime(2020, 2, 13, 12, 0)
```
benthorner, line 797: Do we need to bother about this attribute at all in these tests?

quis: Nope – squashed out

tests/app/dao/test_users_dao.py
```diff
@@ -159,19 +159,15 @@ def test_update_user_attribute(client, sample_user, user_attribute, user_value):
 
 
 @freeze_time('2020-01-24T12:00:00')
-@pytest.mark.parametrize('from_email', [True, False])
-def test_update_user_password(notify_api, notify_db, notify_db_session, sample_user, from_email):
+def test_update_user_password(notify_api, notify_db, notify_db_session, sample_user):
     sample_user.password_changed_at = datetime.utcnow() - timedelta(days=1)
     sample_user.email_access_validated_at = datetime.utcnow() - timedelta(days=1)
```
benthorner, line 164: We can also remove this line and the assertion - I don't think there's any value in testing it's not changed, since that's not a behaviour I would expect of such a feature in the first place.

benthorner (approved):

PR #3310, opened 2021-08-20
Revert "Update utils to bring in coordinate transformation"

quis:
Reverts alphagov/notifications-api#3309

Looks like PaaS doesn’t like trying to install PyProj so the build is failing.


benthorner (approved):
Should we revert in utils for the time being as well?

PR #3321, opened 2021-09-06
Remove remaining areas migration scaffolding

benthorner:

Depends on: https://github.com/alphagov/notifications-admin/pull/4014

### DIFF
--- a/tests/app/v2/broadcast/test_post_broadcast.py
+++ b/tests/app/v2/broadcast/test_post_broadcast.py
@@ -103,11 +103,9 @@ def test_valid_post_cap_xml_broadcast_returns_201(
     assert response_json['personalisation'] is None
     assert response_json['service_id'] == str(sample_broadcast_service.id)
 
-    # TEMPORARY: while we repurpose "areas"
     assert len(response_json['areas']['simple_polygons']) == 1
     assert len(response_json['areas']['simple_polygons'][0]) == 23
     assert response_json['areas']['simple_polygons'][0][0] == [53.10562, 0.244127]
-    assert response_json['areas']['simple_polygons'][0][-1] == [53.10562, 0.244127]
     assert response_json['areas']['simple_polygons'][0][-1] == [53.10562, 0.244127]
 
     assert response_json['starts_at'] is None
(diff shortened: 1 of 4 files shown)


benthorner:
Adding "DNM" as one of the commits depends on the data migration, which we've not run yet.

tests/app/v2/broadcast/test_post_broadcast.py
```diff
@@ -103,12 +103,14 @@ def test_valid_post_cap_xml_broadcast_returns_201(
     assert response_json['personalisation'] is None
     assert response_json['service_id'] == str(sample_broadcast_service.id)
 
-    # TEMPORARY: while we repurpose "areas"
+    assert len(response_json['areas']['names']) == 1
```
klssmith, line 106: Really minor, but this assertion could be removed since the assertion below (`assert response_json['areas']['names'] == ['River Steeping in Wainfleet All Saints']`) shows there is only one.

tests/app/v2/broadcast/test_post_broadcast.py
```diff
@@ -103,12 +103,14 @@ def test_valid_post_cap_xml_broadcast_returns_201(
     assert response_json['personalisation'] is None
     assert response_json['service_id'] == str(sample_broadcast_service.id)
 
-    # TEMPORARY: while we repurpose "areas"
```
klssmith, line 106: The comment here has been removed, but no related code which I guess is just because that has already been deleted? This test does have a line repeated twice though (`assert response_json['areas']['simple_polygons'][0][-1] == [53.10562, 0.244127]`)

tests/app/v2/broadcast/test_post_broadcast.py
benthorner: No problem: I've removed it from the third commit.

tests/app/v2/broadcast/test_post_broadcast.py
benthorner: Good spot. I've amended the second commit to remove the extra line and explain both removals in the commit message.

klssmith (approved):

PR #3322, opened 2021-09-06
Add command to migrate custom broadcast areas

benthorner:

Unlike broadcasts created in the Admin app, these are only expected
to have "names" and "simple_polygons" in their areas column [1].

The migration command in the Admin app [2] isn't suitable for these
broadcasts as it would try to aggregate their areas, etc.

I've put a conditional on "areas" being present (in the areas column)
so this command doesn't pick up any new custom broadcasts.

[1]: https://github.com/alphagov/notifications-api/pull/3312/commits/023a06d5fb4c20e6bf5a1c9e65b473cb376b2b6bo
[2]: https://github.com/alphagov/notifications-admin/pull/4011

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -914,3 +914,26 @@ def populate_annual_billing_with_defaults(year, missing_services_only):
 
     for service in active_services:
         set_default_free_allowance_for_service(service, year)
+
+
+@notify_command(name='tmp-backfill-custom-broadcast-areas')
+def tmp_backfill_custom_broadcast_areas():
+    from app import db
+    from app.models import BroadcastMessage
+
+    custom_broadcasts = BroadcastMessage.query \
+        .filter(BroadcastMessage.api_key_id != None) \
+        .filter(BroadcastMessage.areas.has_key('areas'))  # noqa
+
+    for broadcast in custom_broadcasts:
+        old_areas = broadcast.areas
+
+        new_areas = {
+            'names': old_areas['areas'],
+            'simple_polygons': old_areas['simple_polygons']
+        }
+
+        broadcast.areas = new_areas
+
+        print(f'Migrating {broadcast.id}')
+        db.session.commit()


benthorner:
<details>
 <summary>Script I used for testing</summary>
<pre>
insert into broadcast_message (
  id,
  status,
  api_key_id,
  service_id,
  areas,
  created_at,
  stubbed
) values (
  gen_random_uuid(),
  'draft',
  (select id from api_keys limit 1),
  (select id from services limit 1),
  '{"areas": ["a"], "simple_polygons": [[[1,2]]]}',
  current_timestamp,
  false
);
</pre>

<pre>
insert into broadcast_message (
  id,
  status,
  api_key_id,
  service_id,
  areas,
  created_at,
  stubbed
) values (
  gen_random_uuid(),
  'draft',
  (select id from api_keys limit 1),
  (select id from services limit 1),
  '{"areas": ["a"], "simple_polygons": [[[]]]}',
  current_timestamp,
  false
);
</pre>
</details>

app/commands.py
```diff
@@ -914,3 +914,26 @@ def populate_annual_billing_with_defaults(year, missing_services_only):
 
     for service in active_services:
         set_default_free_allowance_for_service(service, year)
+
+
+@notify_command(name='tmp-backfill-custom-broadcast-areas')
+def tmp_backfill_custom_broadcast_areas():
+    from app import db
```
klssmith, line 921: `db` is already being imported at the top of the file, and it would be nicer to have `BroadcastMessage` imported at the top if possible too (where we import other models). However, if the plan is to remove this code after the command is run, then I don't think it matters too much if it's not changed (I can see how importing here could make the change easier to undo)

app/commands.py
```diff
@@ -914,3 +914,26 @@ def populate_annual_billing_with_defaults(year, missing_services_only):
...
+    from app.models import BroadcastMessage
+
+    custom_broadcasts = BroadcastMessage.query \
+        .filter(BroadcastMessage.api_key_id != None) \
```
klssmith, line 925: You can put both conditions in the `.filter` (e.g. we do that [here](https://github.com/alphagov/notifications-api/blob/master/app/dao/annual_billing_dao.py#L30)), but as with the other comment it's probably not worth making a change if this command will be removed soon.

klssmith (approved):

app/commands.py
benthorner: Oh yes, I could remove the `db` one, but yes the idea is to make it standalone and easy to remove.

app/commands.py
benthorner: I personally prefer is way as I find it more readable vs. having both conditions on the same line.

PR #3323, opened 2021-09-07
Make it easy to develop with broadcast services

benthorner:
Previously I had to handcraft some SQL to give myself access to a
broadcast service I created locally. I've done this enough times
that I think it's worth automating.

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -94,17 +94,25 @@ class notify_command:
         self.name = name
 
     def __call__(self, func):
-        # we need to call the flask with_appcontext decorator to ensure the config is loaded, db connected etc etc.
-        # we also need to use functools.wraps to carry through the names and docstrings etc of the functions.
-        # Then we need to turn it into a click.Command - that's what command_group.add_command expects.
-        @click.command(name=self.name)
-        @functools.wraps(func)
-        @flask.cli.with_appcontext
+        decorators = [
+            click.command(name=self.name),  # turn it into a click.Command
+            functools.wraps(func)  # carry through function name, docstrings, etc.
+        ]
+
+        # in the test environment the app context is already provided and having
+        # another will lead to the test db connection being closed prematurely
+        if os.getenv('NOTIFY_ENVIRONMENT', '') != 'test':
+            # with_appcontext ensures the config is loaded, db connected, etc.
+            decorators.insert(0, flask.cli.with_appcontext)
+
         def wrapper(*args, **kwargs):
             return func(*args, **kwargs)
 
-        command_group.add_command(wrapper)
+        for decorator in decorators:
+            # this syntax is equivalent to e.g. "@flask.cli.with_appcontext"
+            wrapper = decorator(wrapper)
 
+        command_group.add_command(wrapper)
         return wrapper
 
 
--- a/app/commands.py
+++ b/app/commands.py
@@ -1,6 +1,7 @@
 import csv
 import functools
 import itertools
+import os
 import uuid
 from datetime import datetime, timedelta
 from decimal import Decimal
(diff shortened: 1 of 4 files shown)


klssmith:
It feels like this could be a bit dangerous for a couple of reasons. It will put your local development environment in a state which isn't possible "normally" and there's a chance that could lead to us making errors when writing or testing new code locally - users wouldn't ever be able to have all the permissions in [this list](https://github.com/alphagov/notifications-api/blob/b2b921ef21e9b88fe7018af60f95aca81d9abfd2/app/models.py#L1869).

My other concern is that there is nothing to stop this command being run in a non-dev environment. Even though there is alerting in place, we maybe need to check if it's considered too risky to add this since it would provide an quick and easy way to give someone a lot of permissions while bypassing some of the audit trails.

benthorner:
Thanks for taking a look @klssmith. I put this up as a quick attempt but as you point out it has several potential flaws. I've now spent some more time on it and have replaced the original commit and added some tests in another. I've also added an optional, final refactor commit to fix a surprise I found when writing the tests.

> It will put your local development environment in a state which isn't possible "normally"

Agreed. That could lead to a confusing discrepancy if the permission code had a bug in it. I've changed it to just populate the permissions for a broadcast creator + approver, as well as restricting the change to just broadcast services.

> My other concern is that there is nothing to stop this command being run in a non-dev environment.

I'm not concerned about this as if you're able to run this command in Production then you can also run any code you like. However, it's good practice to be defensive and document more clearly where code is meant to be used, so I've added an environment restriction to it so it can only be run locally (test + development).

Would you like to take another look?

app/commands.py
```diff
@@ -914,3 +925,32 @@ def populate_annual_billing_with_defaults(year, missing_services_only):
...
+def local_dev_broadcast_permissions(user_id):
+    if os.getenv('NOTIFY_ENVIRONMENT', '') not in ['development', 'test']:
+        current_app.logger.error('Can only be run in development')
+        return
+
+    user = User.query.filter_by(id=user_id).one()
+
+    user_broadcast_services = Service.query.filter(
+        Service.permissions.any(permission='broadcast'),
+        Service.users.any(id=user_id)
```
klssmith, line 941: You could consider another filter to ensure that only active services are changed

tests/app/test_commands.py
```diff
@@ -0,0 +1,24 @@
+from app.commands import local_dev_broadcast_permissions
+from app.dao.services_dao import dao_add_user_to_service
+from tests.app.db import create_user
+
+
+def test_local_dev_broadcast_permissions(
+    sample_service,
+    sample_broadcast_service,
+    notify_api,
+    mocker,
```
klssmith, line 10: `mocker` is not used in this test

tests/app/conftest.py
```diff
@@ -828,22 +825,21 @@ def letter_volumes_email_template(notify_service):
 
 
 @pytest.fixture
-def notify_service(notify_db, notify_db_session):
-    user = create_user()
+def notify_service(notify_db, notify_db_session, sample_user):
```
klssmith, line 828: Only the `sample_user` fixture is needed for this test.

`sample_user` includes `notify_db_session` which in turn includes `notify_db`. Basically, we can remove some fixtures because none of the tests which have been updated to now use `sample_user` need `notify_db_session` or `notify_db`.

klssmith (commented):
I left a couple of very minor comments, but this looks good - nice job getting testing the command file to work!

tests/app/test_commands.py
benthorner: Good spot - I've amended the second commit to remove it.

tests/app/conftest.py
benthorner: Nice (grumbles about dependency injection). I've added an extra commit to remove all the redundant DB fixtures.

klssmith (approved):

PR #3325, opened 2021-09-08
Add validation to prevent blank area names

benthorner:

Now that these are used for display on gov.uk/alerts we need to
make sure the data is being set properly. We've already found an
example where it wasn't [1]. We validate external broadcasts in
two stages: with the official CAP XML schema [2] and then again
with our own, more specific schema for the converted JSON. Since
this validation is a custom requirement I've made it part of the
JSON schema. Note that jsonschema recommends avoiding metachars
like "\w" since they're not supported by all implementations [3].

I've tested the new validation manually and it works as expected
by disallowing e.g. "  " but still alowing "foo" and "foo bar".
I don't think we need a test for it, as it's declarative code
for an edge case concern.

[1]: https://www.notifications.service.gov.uk/services/120107d0-d99a-4c42-8b70-f37d2f28879b/rejected-alerts/d6e0c70e-60f6-4422-8589-2a2d159c63f2
[2]: https://github.com/alphagov/notifications-api/blob/81a25ff1effa821f0096fcc5edea79b0a5b17f5e/app/xml_schemas/CAP-v1.2.xsd
[3]: http://json-schema.org/understanding-json-schema/reference/regular_expressions.html

### DIFF
--- a/app/v2/broadcast/broadcast_schemas.py
+++ b/app/v2/broadcast/broadcast_schemas.py
@@ -76,6 +76,7 @@ post_broadcast_schema = {
             "properties": {
                 "name": {
                     "type": "string",
+                    "pattern": "([a-zA-Z1-9]+ )*[a-zA-Z1-9]+",
                 },
                 "polygons": {
                     "type": "array",
--- a/app/v2/broadcast/post_broadcast.py
+++ b/app/v2/broadcast/post_broadcast.py
@@ -41,26 +41,11 @@ def create_broadcast():
     broadcast_json = cap_xml_to_dict(cap_xml)
 
     validate(broadcast_json, post_broadcast_schema)
+    _validate_template(broadcast_json)
 
     polygons = Polygons(list(chain.from_iterable((
         area['polygons'] for area in broadcast_json['areas']
     ))))
-
-    template = BroadcastMessageTemplate.from_content(
-        broadcast_json['content']
-    )
-
-    if template.content_too_long:
-        raise ValidationError(
-            message=(
-                f'description must be {template.max_content_count:,.0f} '
-                f'characters or fewer'
-            ) + (
-                ' (because it could not be GSM7 encoded)'
-                if template.non_gsm_characters else ''
-            ),
-            status_code=400,
-        )
 
     broadcast_message = BroadcastMessage(
         service_id=authenticated_service.id,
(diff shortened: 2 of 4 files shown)


leohemsted:
Great fix 👍 

Worth pointing out that we may well expect more in the future, I'm not sure we've got a concrete plan on what to do ~~if~~ when electoral areas change. I think it's reasonable to expect that electoral wards may well merge or be renamed in due course.

benthorner:
Update: after discussing with the team I've modified this PR and the new validation is now just one line in the JSON schema for external broadcasts. I've kept the first couple of commits as I think they still add a little value.

CrystalPea (commented):
The regex looks good to me 👍🏼 . Shouldn't we have some test for this though?

benthorner:
@CrystalPea we could do and I did have one originally. The third commit message has more:

> I've tested the new validation manually and it works as expected
> by disallowing e.g. "  " but still alowing "foo" and "foo bar".
> I don't think we need a test for it, as it's declarative code
> for an edge case concern.

I don't feel too strongly about this, although I think it's worth asking: there are lots of other definitions and settings in that schema and the XML one that we validate with first, so where do we stop with the testing?

CrystalPea:
@benthorner I feel it still would be good to have a test about it, since it's a requirement that could affect the workings of gov.uk/alerts. So if someone were to alter the schema or something, they would at least see a failing test and reflect.

benthorner:
@CrystalPea sure, I've amended the final commit so it includes a test / test data.

Could you take another look?

CrystalPea (approved):
Looks good to me. Only, is `'areas does not match ([a-zA-Z1-9]+ )*[a-zA-Z1-9]+'` the error that user will get? If so, we could do a better one, like 'We cannot accept empty area names' or something.

benthorner:
> Only, is 'areas does not match ([a-zA-Z1-9]+ )*[a-zA-Z1-9]+' the error that user will get?

Yup. We don't control the errors though: this is the standard behaviour for JSON schema and [we let the user see these errors in other cases](https://github.com/alphagov/notifications-api/blob/4c7de35f68cedaf94d3326364e12872eaccd3931/tests/app/v2/broadcast/test_post_broadcast.py#L174). I'm going to merge this now as it's just a small bug I spotted and the main thing is that it's now protected by a validation. We can always try and rescue and customise the error message if someone reports it as confusing. On the positive side, the current error message makes it clear exactly what kinds of area names we _will_ accept, so let's see how it goes.

PR #3328, opened 2021-09-09
Don't raise P1 for broadcasts

sakisv:
This is happening on the AWS side now as part of
alphagov/notifications-broadcasts-infra#267 - but we still want to keep
the zendesk ticket as it contains useful context _and_ provides
visibility to the team.

### DIFF
--- a/tests/app/celery/test_broadcast_message_tasks.py
+++ b/tests/app/celery/test_broadcast_message_tasks.py
@@ -99,7 +99,7 @@ def test_send_broadcast_event_does_nothing_if_provider_set_on_service_isnt_enabl
     ({'names': ['England', 'Scotland']}, ['England', 'Scotland']),
     ({}, [])
 ])
-def test_send_broadcast_event_creates_zendesk_p1(
+def test_send_broadcast_event_creates_zendesk(
     area_data,
     expected_message,
     mocker,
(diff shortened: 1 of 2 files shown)


leohemsted (approved):

tests/app/celery/test_broadcast_message_tasks.py
```diff
@@ -95,43 +95,6 @@ def test_send_broadcast_event_does_nothing_if_provider_set_on_service_isnt_enabl
     assert mock_send_broadcast_provider_message.apply_async.called is False
 
 
-@pytest.mark.parametrize('area_data,expected_message', [
-    ({'names': ['England', 'Scotland']}, ['England', 'Scotland']),
-    ({}, [])
-])
-def test_send_broadcast_event_creates_zendesk_p1(
```
idavidmcdonald, line 102: Do we want to keep the test but just change it so it doesn't check it is a P1 but a non P1?

sakisv: I thought about it, but I don't think it would be giving us much.

First, the default value [is already `False`](https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/clients/zendesk/zendesk_client.py#L51) and the behaviour is [already tested](https://github.com/alphagov/notifications-utils/blob/master/tests/test_zendesk_client.py#L37) in the utils repo.

Second, even if the behaviour in the utils repo changes, this test will not catch that change because we're only testing that the arguments we passed to `create_ticket` includes a `p1=True`.

sakisv: Taking a closer look, this test was the only one that was testing that we were actually creating a zendesk ticket. I restored the test and clarified/renamed the next one to not mention "p1".

PR #3329, opened 2021-09-14
return contact list created_at in UTC, not BST

leohemsted:
make sure timestamps returned from the api are always consistent.

The only place in models where we're serializing a BST timestamp is on the Notification.serialize_for_csv method now, which at least is a bit different as this is user-facing (it also returns a formatted human-readable notification_status for example).

see also

- [ ] https://github.com/alphagov/notifications-admin/pull/4022

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2218,7 +2218,6 @@ class ServiceContactList(db.Model):
         ).first())
 
     def serialize(self):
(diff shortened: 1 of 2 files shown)


benthorner (approved):
Nice 💯

PR #3332, opened 2021-09-15
Add new guidance on writing public APIs

benthorner:
https://www.pivotaltracker.com/story/show/179208442

In response to: https://github.com/alphagov/notifications-api/pull/3305#issuecomment-896631475

👉 [Rendered version](https://github.com/alphagov/notifications-api/blob/api-docs-179208442/docs/writing-public-apis.md)

I've included everything I can think of and tried to structure it in a
way that's easy to digest. Happy to write more if there's more to add.


leohemsted (approved):
i think this looks pretty good, but would be keen to get some other peoples input in this. the moving of the create a new worker app section makes loads of sense too 👍

docs/creating-a-new-worker-app.md
```diff
@@ -0,0 +1,24 @@
+# To create a new worker app
+
+You need to:
+1. Create new entries for your app in `manifest.yml.j2` and `scripts/paas_app_wrapper.sh` ([example](https://github.com/alphagov/notifications-api/pull/2486/commits/6163ca8b45813ff59b3a879f9cfcb28e55863e16))
+1. Update the jenkins deployment job in the notifications-aws repo ([example](https://github.com/alphagov/notifications-aws/commit/69cf9912bd638bce088d4845e4b0a3b11a2cb74c#diff-17e034fe6186f2717b77ba277e0a5828))
+1. Add the new worker's log group to the list of logs groups we get alerts about and we ship them to kibana ([example](https://github.com/alphagov/notifications-aws/commit/69cf9912bd638bce088d4845e4b0a3b11a2cb74c#diff-501ffa3502adce988e810875af546b97))
```
leohemsted, line 6: these two are out of date (jenkins!) and instead we should probably refer to a more recent PR, such as this:

https://github.com/alphagov/notifications-aws/pull/865

docs/creating-a-new-worker-app.md
```diff
@@ -0,0 +1,24 @@
...
+
+## First time deployment of your new worker
+
+Our deployment flow requires that the app is present in order to proceed with the deployment.
+
+This means that the first deployment of your app must happen manually.
+
+To do this:
+
+1. Ensure your code is backwards compatible
```
leohemsted, line 21: what does this mean? if it's a new worker what does it have to be compatible with? What kind of things could it overwrite?

docs/creating-a-new-worker-app.md
```diff
@@ -0,0 +1,24 @@
...
+1. From the root of this repo run `CF_APP=<APP_NAME> make <cf-space> cf-push`
```
leohemsted, line 22: `cf-push` doesn't exist as a makefile command any more. not quite sure what we need to do here, but I think it's along the lines of:

create an empty paas app (how?)

then update concourse
then run pipeline as normal, and it'll overwrite that empty paas with the desired app with config, env vars, etc as per usual.

paas route service somewhere in here?

leohemsted (commented):
actually the worker app section is really badly out of date and missing lots of context. I don't think that needs to change in this PR but if not, then I'd like to remove the details and make it a skeleton for us to fill out later.

i think it's still a useful doc to have but if it's not up to date it's worthless

benthorner:
> actually the worker app section is really badly out of date and missing lots of context.

@leohemsted sure, I agree something that's so out-of-date is worse than useless. I don't think there's much value in having a skeleton doc: if it's truly valuable to have then we should write it up now, otherwise we'll be lacking...something. I don't think it's of significant value, so I suggest we delete it and write it again if / when it becomes apparent that we need it and when we'll likely have more context from (presumably) actually trying to add a new worker app.

Do you feel strongly about keeping it in some form?

leohemsted:
@benthorner fair point. I guess I was thinking about the benefits of a skeleton in this case because it sets the scope for the docs folder of "what kind of things might we want in this folder" for the future, given that having a docs folder with technical notes isn't really a pattern we've used much yet.

That said I won't lose sleep if we delete it, happy for you to just bin it

benthorner:
> That said I won't lose sleep if we delete it, happy for you to just bin it

@leohemsted OK, I've added a commit to do just that. I'm confident we'll grow into the docs pattern - the problem at the moment isn't so much having a pattern, just getting people to write stuff down 😉.

PR #3334, opened 2021-09-23
Use the new Zendesk form for all tickets

klssmith:
We introduced a new way of creating Zendesk tickets here https://github.com/alphagov/notifications-utils/pull/899 and tried it out first just on the request to go live tickets (https://github.com/alphagov/notifications-admin/pull/4024).

It worked! So now we want to use the new way of creating tickets everywhere.

[Pivotal story](https://www.pivotaltracker.com/story/show/179005041)

### DIFF
--- a/tests/app/celery/test_nightly_tasks.py
+++ b/tests/app/celery/test_nightly_tasks.py
@@ -374,7 +374,11 @@ def test_letter_raise_alert_if_no_ack_file_for_zip_does_not_raise_when_files_mat
 @freeze_time('2018-01-11T23:00:00')
 def test_letter_raise_alert_if_ack_files_not_match_zip_list(mocker, notify_db):
     mock_file_list = mocker.patch("app.aws.s3.get_list_of_files_by_suffix", side_effect=mock_s3_get_list_diff)
-    mock_zendesk = mocker.patch("app.celery.nightly_tasks.zendesk_client.create_ticket")
+    mock_create_ticket = mocker.spy(NotifySupportTicket, '__init__')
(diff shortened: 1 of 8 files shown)


tests/app/celery/test_nightly_tasks.py
```diff
@@ -232,34 +234,47 @@ def test_should_call_delete_inbound_sms(notify_api, mocker):
     assert nightly_tasks.delete_inbound_sms_older_than_retention.call_count == 1
 
 
-def test_create_ticket_if_letter_notifications_still_sending(mocker):
+def test_create_ticket_if_letter_notifications_still_sending(notify_api, mocker):
```
idavidmcdonald, line 237: How come we had to add the `notify_api` here?

idavidmcdonald (approved):
Lovely! Nice clean changes for swapping out the old way of doing things with the new way of doing things.

Well broken down commit by commit too

tests/app/celery/test_nightly_tasks.py
klssmith: This test (and the other test that I added the fixture to) need an application context in order to run otherwise they give an error. Since the `notify_api` fixture is scoped to the session, these tests were passing when run as part of a group which uses the fixture but giving an error when run individually.

idavidmcdonald: Thank you, good spot.

If you were going to make this PR even better (don't worry, it's already brilliant), then that would be good to put in a separate commit explaining the change.

But definitely not worth doing now

PR #3336, opened 2021-09-28
Don’t repeat digits in security codes

quis:
People with dyslexia and dyscalculia find it difficult to transpose codes which have consecutive, repeated digits<sup>1</sup>, for example `11122` or `98776`.

This commits enhances the algorithm for generating codes to not repeat the previous digit in a code.

This reduces the key space for our codes from 100,000 (10<sup>5</sup>) possibilities to 65,610 (10 &times; 9<sup>4</sup>) possibilities.

***

1. https://twitter.com/annaecook/status/1442567679710150662
2. Background reading: https://github.com/alphagov/govuk-design-system/pull/1265

### DIFF
--- a/app/dao/users_dao.py
+++ b/app/dao/users_dao.py
@@ -20,7 +20,15 @@ def _remove_values_for_keys_if_present(dict, keys):
 
 
 def create_secret_code():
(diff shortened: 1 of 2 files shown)


idavidmcdonald (approved):
Lovely

PR #3341, opened 2021-10-11
Add task to recreate the PDF file for a non-templated letter

klssmith:
**Add task to re-sanitise and replace a PDF for precompiled letter**
This adds a task which is designed to be used if we want to recreate the PDF for a precompiled letter (either one that has been created using the API or one that has been uploaded through the website).

The task takes the `notification_id` of the letter and passes template preview the details it needs in order to sanitise the original file and then replace the version in the letters-pdf bucket with the freshly sanitised version.

**Add command to run the new task**
We already had the `replay-create-pdf-for-templated-letter` command. This adds a new command, `recreate-pdf-for-precompiled-or-uploaded-letter` which does the same thing but for non-templated letters.

Part of [Pivotal story](https://www.pivotaltracker.com/story/show/178746435)

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -31,6 +31,7 @@ from app.letters.utils import (
     generate_letter_pdf_filename,
     get_billable_units_for_letter_page_count,
     get_file_names_from_error_bucket,
+    get_folder_name,
     get_reference_from_filename,
     move_error_pdf_to_scan_bucket,
     move_failed_pdf,
@@ -524,3 +525,37 @@ def replay_letters_in_error(filename=None):
                     [filename],
                     queue=QueueNames.LETTERS
                 )
+
+
+@notify_celery.task(name='resanitise-pdf')
+def resanitise_pdf(notification_id):
+    """
+    `notification_id` is the notification id for a PDF letter which was either uploaded or sent using the API.
+
+    This task calls the `recreate_pdf_for_precompiled_letter` template preview task which recreates the
+    PDF for a letter which is already sanitised and in the letters-pdf bucket. The new file that is generated
+    will then overwrite the existing letter in the letters-pdf bucket.
+    """
+    notification = get_notification_by_id(notification_id)
+
+    # folder_name is the folder that the letter is in the letters-pdf bucket e.g. '2021-10-10/'
+    folder_name = get_folder_name(notification.created_at)
+
+    filename = generate_letter_pdf_filename(
+            reference=notification.reference,
+            created_at=notification.created_at,
+            ignore_folder=True,
+            postage=notification.postage
+        )
+
+    notify_celery.send_task(
+        name=TaskNames.RECREATE_PDF_FOR_PRECOMPILED_LETTER,
+        kwargs={
+            'notification_id': str(notification.id),
+            'file_location': f'{folder_name}{filename}',
+            'allow_international_letters': notification.service.has_permission(
+                INTERNATIONAL_LETTERS
+            ),
+        },
+        queue=QueueNames.SANITISE_LETTERS,
+    )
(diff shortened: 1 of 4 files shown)


klssmith:
[Here's the recovery plan](https://docs.google.com/document/d/1sP2BMd59E0LoPHFpsJNoioBK9w0a9_fgYY1HznMUfVQ/edit#), which should hopefully address which problems we are attempting to fix here, and which are not covered.

app/celery/letters_pdf_tasks.py
```diff
@@ -524,3 +525,37 @@ def replay_letters_in_error(filename=None):
...
+@notify_celery.task(name='resanitise-pdf')
+def resanitise_pdf(notification_id):
+    """
+    `notification_id` is the notification id for a PDF letter which was either uploaded or sent using the API.
+
+    This task calls the `recreate_pdf_for_precompiled_letter` template preview task which recreates the
+    PDF for a letter which is already sanitised and in the letters-pdf bucket. The new file that is generated
+    will then overwrite the existing letter in the letters-pdf bucket.
+    """
+    notification = get_notification_by_id(notification_id)
```
CrystalPea, line 539: should we have a check here to make sure it's a precompiled letter, not a templated one?

Optional/maybe for separate PR/story: it would be cool to have a task / function that would recognise which letters are precompiled and which are templated and deal with them appropriately, what do yo think about this?

CrystalPea (commented):
Good work!

I left a comment, let me know what you think.

app/celery/letters_pdf_tasks.py
```diff
@@ -524,3 +525,37 @@ def replay_letters_in_error(filename=None):
...
+        name=TaskNames.RECREATE_PDF_FOR_PRECOMPILED_LETTER,
+        kwargs={
+            'notification_id': str(notification.id),
+            'file_location': f'{folder_name}{filename}',
+            'allow_international_letters': notification.service.has_permission(
+                INTERNATIONAL_LETTERS
+            ),
+        },
+        queue=QueueNames.SANITISE_LETTERS,
+    )
```
CrystalPea, line 561: if notification is in `sending` status, should we put it back to `created` so it gets sent to DVLA again? Or do we just do it all manually, like we chatted yesterday with ZIP substitution etc?

app/celery/letters_pdf_tasks.py
klssmith: I don't think we need to check that the letter is precompiled and not templated. The other letter tasks, for example `sanitise_letter` and `get_pdf_for_templated_letter`,  don't check that the letter type is right - we trust that the code is doing the right thing or that the person on support has used the right letter type. If someone did accidentally use this task with a templated letter, they would see an error in the logs when there's no file in the backup S3 folder but nothing would actually happen, so it wouldn't cause an issue.

I did think about having a task that recognises which letters are precompiled. One reason I didn't do it (apart from the fact that it involved writing more code!) is that the queries I have to find the different types use different end times. So my thinking was that we couldn't run a task for all letters at once, since there's not a single database query to find all affected letters.

app/celery/letters_pdf_tasks.py
klssmith: The idea is that that would be done manually after discussing with DVLA which ones should be resent.

benthorner (approved):
Thanks for sharing the overall recovery doc. It addresses a lot of the questions I had in my original comment. One thing that's still unclear is the proposed process for notifications in "sending". In [this comment](https://github.com/alphagov/notifications-api/pull/3341#discussion_r742788169) it sounds like we'll reset them manually, but the queries for IDs to call this task with include notifications in "sending" status.

I don't think we need to resolve that question here, as the code should cope with notifications in "sending" status. I have added a comment about asserting in code the statuses the task is meant to support, but that's more toward the optional side of things.

CrystalPea (approved):

PR #3355, opened 2021-11-01
Upgrade to Celery 5

benthorner:
https://www.pivotaltracker.com/story/show/180017131

### DIFF
--- a/scripts/paas_app_wrapper.sh
+++ b/scripts/paas_app_wrapper.sh
@@ -21,7 +21,7 @@ case $NOTIFY_APP_NAME in
     -Q research-mode-tasks 2> /dev/null
     ;;
   delivery-worker-sender)
-    exec scripts/run_multi_worker_app_paas.sh celery multi start 3 -c 10 -A run_celery.notify_celery --loglevel=INFO \
+    exec scripts/run_multi_worker_app_paas.sh celery multi start 3 -c 4 -A run_celery.notify_celery --loglevel=INFO \
     --logfile=/dev/null --pidfile=/tmp/celery%N.pid -Q send-sms-tasks,send-email-tasks
     ;;
   delivery-worker-periodic)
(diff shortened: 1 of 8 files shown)


rjbaker (approved):
Good description. Busy tasks such as sender seem stable and performant with the new concurrency settings. We can tune these further and resize instances as required for better efficiency. 🚢

scripts/paas_app_wrapper.sh
```diff
@@ -33,7 +33,7 @@ case $NOTIFY_APP_NAME in
     -Q reporting-tasks 2> /dev/null
     ;;
   delivery-worker-priority)
-    exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=5 \
+    exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=4 \
```
idavidmcdonald, line 36: Quick drive by comment. There are several apps here running with concurrency not equal to 4. It would be good I think to either
- make it consistent
- detail in the commit message which ones are not being set to 4 and why please

rjbaker: The lower concurrency tasks don't need to change at this point as they appear to be fine with long polling enabled. So, in the interests of changing one thing at a time, I feel it's okay to leave them as they are. The high-concurrency tasks were definitely running too many processes with long-polling enabled. I expect we may need to tweak these more scientifically, both for performance and economy.

sakisv: Made it consistent in https://github.com/alphagov/notifications-api/pull/3355/commits/9e9091e980c6c2a10a47cc7c443ecadbcaf29512

scripts/paas_app_wrapper.sh
```diff
@@ -5,15 +5,15 @@ case $NOTIFY_APP_NAME in
     exec scripts/run_app_paas.sh gunicorn -c /home/vcap/app/gunicorn_config.py application
     ;;
   delivery-worker-retry-tasks)
-    exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=11 \
+    exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=4 \
```
benthorner, line 8: @sakisv this should actually be unnecessary, as per [the docs](https://docs.celeryproject.org/en/stable/reference/cli.html#cmdoption-celery-worker-c). In fact, it would be better _not_ to specify it in case our CPU provision changes in future.

benthorner: Agreed to leave this as-is for now: the important thing is that it's relatively low and it's unlikely the number of cores on any worker instance will _decrease_, which was my main concern here.

PR #3358, opened 2021-11-04
Republish gov.uk/alerts every night to clear down planned tests

quis:
We have made it so that gov.uk/alerts shows a ‘1 planned test’ banner for the whole of the day when there has been an operator test on that day.

We need to remove the banner when the day is over.

The most straightforward way to do this is to republish the site at the start of every day. The [gov.uk/alerts code will work out if there are or aren’t any planned tests to show that day](https://github.com/alphagov/notifications-govuk-alerts/blob/5a274af6d0c202b7bd5e664967b1e113015b9a31/app/models/alerts.py#L38-L44).

### DIFF
--- a/tests/app/celery/test_scheduled_tasks.py
+++ b/tests/app/celery/test_scheduled_tasks.py
@@ -19,6 +19,7 @@ from app.celery.scheduled_tasks import (
     check_job_status,
     delete_invitations,
     delete_verify_codes,
+    remove_yesterdays_planned_tests_on_govuk_alerts,
     replay_created_notifications,
     run_scheduled_jobs,
     switch_current_sms_provider_on_slow_delivery,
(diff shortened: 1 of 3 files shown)


tests/app/celery/test_scheduled_tasks.py
```diff
@@ -700,3 +701,16 @@ def test_auto_expire_broadcast_messages(
         )
     else:
         assert not mock_celery.called
+
+
+def test_remove_yesterdays_planned_tests_on_govuk_alerts(
```
CrystalPea, line 706: should we have an end-to-end test, with freezetime to show that alert is created on one day, then next day it is not shown in planned alerts anymore?

quis: I don’t know how you can freeze time like that on a functional, end-to-end test.

I did look to see if there was at least tests for some of the other scheduled task timings but couldn’t see anything to copy.

CrystalPea: Ahh, maybe it's too much work to establish a totally new pattern then, it's ok 👍🏼

CrystalPea (approved):

tests/app/celery/test_scheduled_tasks.py
CrystalPea: I thought of an end-to-end unit test like this:

```
freeze time on 11.11.2021
create the alert on that day
freeze time on 12.11.2021
run remove_yesterdays_planned_tests_on_govuk_alerts()
assert the alert is treated differently

PR #3360, opened 2021-11-08
Optimise query to get notifications to "time out"

benthorner:
From experimenting in production we found a "!=" caused the engine
to use a sequential scan, whereas explicitly listing all the types
ensured an index scan was used.

We also found that querying for many (over 100K) items leads to
the task stalling - no logs, but no evidence of it running either -
so we also add a limit to the query.

Since the query now only returns a subset of notifications, we need
to ensure the subsequent "update" query operates on the same batch.
Also, as a temporary measure, we have a loop in the task code to
ensure it operates on the total set of notifications to "time out",
which we assume is less than 500K for the time being.

- [ ] Test this works manually in Production before deploying.

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -130,25 +130,35 @@ def delete_letter_notifications_older_than_retention():
 @notify_celery.task(name='timeout-sending-notifications')
 @cronitor('timeout-sending-notifications')
 def timeout_notifications():
-    technical_failure_notifications, temporary_failure_notifications = \
-        dao_timeout_notifications(current_app.config.get('SENDING_NOTIFICATIONS_TIMEOUT_PERIOD'))
+    # TEMPORARY: re-run the following code over small batches of notifications
+    # so that we can cope with a high volume that need processing. We've changed
+    # dao_timeout_notifications to return up to 100K notifications, so this task
+    # will operate on up to 500K - normally we only get around 20K.
+    for _ in range(0, 5):
+        technical_failure_notifications, temporary_failure_notifications = \
+            dao_timeout_notifications(current_app.config.get('SENDING_NOTIFICATIONS_TIMEOUT_PERIOD'))
 
-    notifications = technical_failure_notifications + temporary_failure_notifications
-    for notification in notifications:
-        # queue callback task only if the service_callback_api exists
-        service_callback_api = get_service_delivery_status_callback_api_for_service(service_id=notification.service_id)
-        if service_callback_api:
-            encrypted_notification = create_delivery_status_callback_data(notification, service_callback_api)
-            send_delivery_status_to_service.apply_async([str(notification.id), encrypted_notification],
-                                                        queue=QueueNames.CALLBACKS)
+        notifications = technical_failure_notifications + temporary_failure_notifications
+        for notification in notifications:
+            # queue callback task only if the service_callback_api exists
+            service_callback_api = get_service_delivery_status_callback_api_for_service(service_id=notification.service_id)  # noqa: E501
+            if service_callback_api:
+                encrypted_notification = create_delivery_status_callback_data(notification, service_callback_api)
+                send_delivery_status_to_service.apply_async([str(notification.id), encrypted_notification],
+                                                            queue=QueueNames.CALLBACKS)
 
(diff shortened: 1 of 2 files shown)


app/celery/nightly_tasks.py
```diff
@@ -130,25 +130,29 @@ def delete_letter_notifications_older_than_retention():
...
+        current_app.logger.info(
+            "Timeout period reached for {} notifications, status has been updated.".format(len(notifications)))
+        if technical_failure_notifications:
+            message = "{} notifications have been updated to technical-failure because they " \
+                      "have timed out and are still in created.Notification ids: {}".format(
+                          len(technical_failure_notifications), [str(x.id) for x in technical_failure_notifications])
+            raise NotificationTechnicalFailureException(message)
+
+        if len(notifications) == 0:
+            return
```
leohemsted, line 156: do we want to log something here (ie: at function root level) if we received over 500k timeouts and were still updating things?

app/celery/nightly_tasks.py
```diff
@@ -130,25 +130,29 @@ def delete_letter_notifications_older_than_retention():
...
+                send_delivery_status_to_service.apply_async([str(notification.id), encrypted_notification],
+                                                            queue=QueueNames.CALLBACKS)
+
+        current_app.logger.info(
+            "Timeout period reached for {} notifications, status has been updated.".format(len(notifications)))
+        if technical_failure_notifications:
+            message = "{} notifications have been updated to technical-failure because they " \
+                      "have timed out and are still in created.Notification ids: {}".format(
+                          len(technical_failure_notifications), [str(x.id) for x in technical_failure_notifications])
+            raise NotificationTechnicalFailureException(message)
```
leohemsted, line 152: any tech failures will abort the whole process even if there are still things in sending. my gut feel says i think this is probably okay because we'd end up looking into this the next morning anyway but worth a quick think

app/celery/nightly_tasks.py
```diff
@@ -130,25 +130,29 @@ def delete_letter_notifications_older_than_retention():
...
-                                                        queue=QueueNames.CALLBACKS)
-
-    current_app.logger.info(
-        "Timeout period reached for {} notifications, status has been updated.".format(len(notifications)))
-    if technical_failure_notifications:
-        message = "{} notifications have been updated to technical-failure because they " \
-                  "have timed out and are still in created.Notification ids: {}".format(
-                      len(technical_failure_notifications), [str(x.id) for x in technical_failure_notifications])
-        raise NotificationTechnicalFailureException(message)
+    for _ in range(0, 5):
```
leohemsted, line 133: can we include a comment here explaining that there's a limit within dao_timeout_notifications and that's _why_ we're looping 5 times, and what that means (that we'll process a max of 500k timed out notifications per day)

app/celery/nightly_tasks.py
benthorner: Hmm...tricky. This is meant to be temporary and the original task wouldn't have coped with this much anyway. I'm not sure a log will help much unless we're actively looking for it, but agree it's not great to leave the code path hanging like it is. I've gone with a `RuntimeError` to reflect we don't expect to get here (for now).

app/celery/nightly_tasks.py
benthorner: Ugh yuk, I hadn't read the code until now - that bit should probably be a log as it's an expected code path 🤷. You're right this bit doesn't play nicely with the looping. For now I think it's OK as we don't seem to get these normally and we can always re-run the task manually if we do get one. For the longterm I've added a suggested action to the incident report to revisit this behaviour.

app/celery/nightly_tasks.py
benthorner: Good plan! I've amended the commit to include a comment here and where we modified the DAO function.

leohemsted (approved):
thanks for updates to comments! 👍

benthorner:
I've made [a tiny change](https://github.com/alphagov/notifications-api/pull/3360/commits/cdb43fbaf6f9c1164017cdfc7a221c81678382bc) to this after seeing how it behaves in Production:

Screenshot 2021-11-09 at 15 42 48

PR #3361, opened 2021-11-09
Bump Celery to latest version

quis:
This brings in the version 5.2.1 of Kombu, which fixes a security vulnerability:

> Celery 5.2.0 includes 'kombu' v5.2.1, which includes dependencies updates that resolve security issues.

— https://pyup.io/repos/github/alphagov/notifications-api/commits/?page=1#b654c27699a5164cbbe50e042d5d34141f560255

This is the commit from Kombu: https://github.com/celery/kombu/commit/f3b04558fa0df4ecc383c93e0b3b300d95e17c47

I believe the dependency of Kombu which has issues is urllib3, which has two open advisories for versions less than 1.26.5:
- https://github.com/urllib3/urllib3/security/advisories/GHSA-q2q7-5pp4-w6pg
- https://github.com/urllib3/urllib3/security/advisories/GHSA-5phf-pp7p-vc2r

*** 

_Note this will not work on Python 3.6 because [support was dropped in Celery 5.2](https://github.com/celery/celery/pull/6934)_


requirements.txt
```diff
@@ -78,7 +76,7 @@ docopt==0.6.2
     #   notifications-python-client
 docutils==0.15.2
     # via awscli
-eventlet==0.30.2
+eventlet==0.30.2 # pyup: ignore
```
benthorner, line 79: Should we keep this?

quis: 0aa7cf1aaff9dd8fd4ec1fb09fa6d64226e159b1

> **Tell Pyup to ignore outdated Eventlet version**
>
> We already do this in the admin app:
> https://github.com/alphagov/notifications-admin/pull/3876/files
>
> Upgrading Eventlet is blocked until this change in Gunicorn is released:
> https://github.com/benoitc/gunicorn/pull/2581/files

quis: I’m not sure why Pyup wasn’t picking this up before, but it is now

benthorner: (Sorry, I read the diff the wrong way round 🙈.)

benthorner (approved):
🎉

PR #3363, opened 2021-11-10
Python 3.9

leohemsted:
Didn't need any code changes to get all the tests to work locally! 🥳 

i had a click about but haven't exhaustively tested things.

The dependencies did break (because one backfill module wasn't available in python 3.9), so I took the opportunity to update from alexey's manual freeze script to the pip-tools process we already use in admin and ftp


quis:
Nice. Would suggest updating `.pyup.yml` – here it’s pointing to a file that won’t exist any more:
https://github.com/alphagov/notifications-api/blob/770d323274adc305c33b8e85bf0c45b6168145b6/.pyup.yml#L7

quis (approved):
Tests pass locally in a fresh 3.9.5 venv.

Ran `make freeze-requirements` and got the same result in `requirements.txt`.

Good to go I reckon.

PR #3367, opened 2021-11-12
Improve log message searchability for duplicate receipts

idavidmcdonald:
There were two problems with the existing message.

1. There was no space between the new status and the time taken
   which made reading and searching harder
2. They key bits of information (before and after status) were
   separated by the time taken (which will always be unique) meaning
   you couldn't do an easy search for a message that is say in delivered
   being attempted to be set to temporary-failure.

Example logs at the moment in Logit which are a bit hard to search through:
image


klssmith (approved):
Nice, this should make searching much easier!

PR #3369, opened 2021-11-15
Remove -Ofair option from celery worker

idavidmcdonald:
In version 4.0 of celery, -Ofair became the default
scheduling strategy:
https://docs.celeryproject.org/en/latest/history/whatsnew-4.0.html?highlight=fair#ofair-is-now-the-default-scheduling-strategy

This appears to still be the case:
https://github.com/celery/celery/blob/5d68d781de807b4576cf5f574e5ba0aaf0d17388/celery/concurrency/asynpool.py#L80

Note, it took me a while to be certain of this as the documentation
for the celery CLI suggests a choice of `default` or `fair` which
isn't so useful as both of these are `fair`:
https://docs.celeryproject.org/en/latest/reference/cli.html#cmdoption-celery-worker-O

### DIFF
--- a/scripts/paas_app_wrapper.sh
+++ b/scripts/paas_app_wrapper.sh
@@ -29,7 +29,7 @@ case $NOTIFY_APP_NAME in
     -Q periodic-tasks 2> /dev/null
     ;;
   delivery-worker-reporting)
-    exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=2 -Ofair \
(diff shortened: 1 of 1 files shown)


benthorner (approved):
This is indeed the new default. Originally I left this as-is to minimise the diff when we upgraded (since I didn't need to touch this file for any other reason).

Happy to remove the default. What led you to come here?

idavidmcdonald:
It was based on investigating whether it should be `-Ofair` or `-O fair` which we had commented on before and weren't clear on the answer. I had a bit of time and fancied looking in to that and then came across the fact that rather than have to figure that out, we could just remove it :)

PR #3371, opened 2021-11-15
Reduce concurrency to match number of CPUs

benthorner:
https://www.pivotaltracker.com/story/show/180116935

This got missed in [1].

[1]: https://github.com/alphagov/notifications-api/pull/3355/commits/9e9091e980c6c2a10a47cc7c443ecadbcaf29512

### DIFF
--- a/scripts/paas_app_wrapper.sh
+++ b/scripts/paas_app_wrapper.sh
@@ -17,7 +17,7 @@ case $NOTIFY_APP_NAME in
     -Q database-tasks,job-tasks 2> /dev/null
     ;;
   delivery-worker-research)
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):
What would be good would be an explanation of why concurrency is 2 on the few that aren't on concurrency 4 but I'm not going to stop that blocking this change going in.

benthorner:
> why concurrency is 2 on the few that aren't on concurrency 4

This is indeed a mystery. In theory we shouldn't specify the concurrency at all, as [Celery will automatically match it to the number of CPUs](https://docs.celeryproject.org/en/stable/userguide/configuration.html#std-setting-worker_concurrency). The risk of specifying it ourselves, as here, is that we over or under-provision our workers relative to the machine they run on. I'm tempted to just remove it everywhere, but this at least means we're not overworking.

PR #3374, opened 2021-11-16
Log activity on all periodic Celery tasks

benthorner:
https://www.pivotaltracker.com/n/projects/1443052

As stated in the comment, this would have been helpful during an
incident to give further reassurance that a task had at least
started running - at the time the only evidence for this was the
Cronitor dashboard itself, which we don't often look at.

I've removed other, equivalent "starting" logs, but kept those
that provide additional information in the log message.

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -126,7 +126,6 @@ def collate_letter_pdfs_to_be_sent():
     that have not yet been sent.
     If run after midnight, it will collect up letters created before 5:30pm the day before.
     """
-    current_app.logger.info("starting collate-letter-pdfs-to-be-sent")
     print_run_date = convert_utc_to_bst(datetime.utcnow())
     if print_run_date.time() < LETTER_PROCESSING_DEADLINE:
         print_run_date = print_run_date - timedelta(days=1)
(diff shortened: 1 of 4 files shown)


idavidmcdonald (approved):
I am a little cautious on the approach here but I think it's OK and I'm happy to approve.

I was initially cautious about moving the logging in to the cronitor wrapper as it felt like hiding away this functionality in somewhere that you'd never expect a log message to say I'm starting a celery task. I was wondering if going for a more duplicative, but clearer approach of just having an explicit single log line in each task might have been better.

However, given the wording for the log line is cronitor specific, it doesn't feel like you've taken something existing and shoe horned it in to the cronitor wrapper to come up with a DRY way of doing it.

Only thing I'd point out is that there will be no log line in staging which may not be a great thing if we are doing debugging or testing in staging but I could live with that. 

I've also looked through all our tasks and it looks like we have really good cronitor coverage too, pretty much all the tasks that happen once overnight (or similar in that sense) are hooked up to cronitor so will get this coverage which is cool.

benthorner:
Thanks for your deep thinking here @idavidmcdonald. It echoes some of my own thoughts:

- I struggled to add a log to every task due to the existing ones and trying to decide a pattern.
- I noticed a trend: if we care about the task starting then it ought to be monitored by Cronitor.
- We want _a_ log line, but it doesn't have to be Celery-specific, hence the compromise here.

This approach also has the advantage of covering future periodic tasks. I hadn't thought about it only working in Production, though, but since it's a non-critical feature anyway I agree it's not a big deal.

PR #3378, opened 2021-11-22
Change timeout-notifications to run every hour

benthorner:

This should avoid building up a large quantity of notifications to process
at the end of the day, meaning we can remove the temporary hack we put in
to cope with high volumes. Normally this task takes a couple of minutes to
complete, so running it every hour should be more than enough to allow for
a bit of growth, and not put any significant load on the DB.

I've made a separate note to update Cronitor if this goes in, since we will 
want to know earlier than 1 day if the job stops running.

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -130,35 +130,25 @@ def delete_letter_notifications_older_than_retention():
 @notify_celery.task(name='timeout-sending-notifications')
 @cronitor('timeout-sending-notifications')
 def timeout_notifications():
-    # TEMPORARY: re-run the following code over small batches of notifications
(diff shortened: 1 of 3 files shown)


benthorner:
Thanks for taking a look @servingUpAces.

I don't think performance would be an issue: normally this task only takes 1-2 mins to run (but only up to 100K notifications, where it runs out of memory). So I'd expect it to be fine running more frequently.

I would, however, monitor for any impact manually if this were to be deployed. Just in case.

I put this PR up as a draft but have since been discussing options with @idavidmcdonald. I haven't made a final decision yet, but in the meantime I'll close this PR for now as it wasn't David's suggested option.

PR #3382, opened 2021-11-25
Report processing stats to 2dp rather than 1dp

idavidmcdonald:
We are starting to see lots of 100.0%s in the current table
and we think this looks suspiciously too good so think it is
beneficial to change it to be 2dp such that we get a few more
non 100.0% values.

For the admin app to be able to show things to 2dp, we need to
give at least 2dp of accuracy otherwise we are losing 1dp of
granularity.

See https://github.com/alphagov/notifications-admin/pull/4079 for admin app PR.

### DIFF
--- a/app/performance_dashboard/rest.py
+++ b/app/performance_dashboard/rest.py
@@ -78,7 +78,7 @@ def transform_into_notification_by_type_json(total_notifications):
 def transform_processing_time_results_to_json(processing_time_results):
     j = []
     for x in processing_time_results:
-        j.append({"date": x.date, "percentage_under_10_seconds": round(x.percentage, 1)})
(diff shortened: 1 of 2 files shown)


quis (commented):
I’m not sure why we wouldn’t leave this un-rounded and let the admin app alone deal with presenting the data. If, for example, we decided to change to 3 decimal places we’d need to make the same change in both places again. 

If un-rounded floats are hard to test then you can use `math.isclose`. To test result is correct to 5 decimal places you can write something like:
```python
assert math.isclose(result, expected, rel_tol=1e-05)
```

idavidmcdonald:
Good point. Have changed it to just return the raw float from the DB

quis (approved):

PR #3385, opened 2021-11-30
Improve query performance

servingUpAces:
Adding a filter to `app.dao.notifications_dao.is_delivery_slow_for_providers` query to improve the performance. By added Notifications.notification_type = 'sms' to the query it will improve the performance some analyse shows 500ms improvement, which is a good thing especially when the query is run once a minute.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -558,6 +558,7 @@ def is_delivery_slow_for_providers(
         ProviderDetails
     ).outerjoin(
         Notification, and_(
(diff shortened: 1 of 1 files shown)


benthorner (approved):
This seems reasonable to me:

- The query is already SMS-specific.
- Makes sense to outer join to less stuff.

I suspect we no longer need `ProviderDetails.notification_type == 'sms'` as all the notifications will now be limited to SMS providers. However, if this is already a bit faster then it obviously doesn't hurt to leave it in.

PR #3390, opened 2021-12-03
put delete tasks on the reporting worker

leohemsted:
They share a lot with the reporting tasks (creating ft_billing and ft_notification_status), in that they're run nightly, take a long time, and we see error messages if they get run multiple times (due to visibility timeout).

The periodic app has two concurrent processes - previously there was just one delete task, which would use one of those processes, while the other process would pick up anything else on the queue (at that time of night, the regular provider switch checks and scheduled job checks). However, when we switched to running the three delete notification types separately, we saw visibility timeout issues - three tasks would be created, all three would be picked up by one celery instance, the two worker processes would start on two of them, and the third would sit on the box, wait longer than the visibility timeout to be picked up (and acknowledged), and so SQS would assume the task was lost and replay it.

it's queues all the way down!

By putting them on the reporting worker we can take advantage of tuning that app (for example setting the prefetch multiplier to one) which is designed to run large tasks. We've also got more concurrent workers on this box, so we can run all three tasks at once.

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -65,9 +65,9 @@ def _remove_csv_files(job_types):
 
 @notify_celery.task(name="delete-notifications-older-than-retention")
 def delete_notifications_older_than_retention():
-    delete_email_notifications_older_than_retention.apply_async(queue=QueueNames.PERIODIC)
-    delete_sms_notifications_older_than_retention.apply_async(queue=QueueNames.PERIODIC)
-    delete_letter_notifications_older_than_retention.apply_async(queue=QueueNames.PERIODIC)
+    delete_email_notifications_older_than_retention.apply_async(queue=QueueNames.REPORTING)
+    delete_sms_notifications_older_than_retention.apply_async(queue=QueueNames.REPORTING)
+    delete_letter_notifications_older_than_retention.apply_async(queue=QueueNames.REPORTING)
 
 
 @notify_celery.task(name="delete-sms-notifications")
--- a/app/config.py
+++ b/app/config.py
@@ -267,7 +267,7 @@ class Config(object):
             'delete-notifications-older-than-retention': {
                 'task': 'delete-notifications-older-than-retention',
                 'schedule': crontab(hour=3, minute=0),  # after 'create-nightly-notification-status'
-                'options': {'queue': QueueNames.PERIODIC}
+                'options': {'queue': QueueNames.REPORTING}
             },
             'delete-inbound-sms': {
                 'task': 'delete-inbound-sms',


idavidmcdonald (approved):
I'm OK with this going in as is. But I'm a little hesistant due to the fact that it isn't really a reporting task, it does feel more like a periodic task. Maybe instead we could be tweaking the periodic worker to have a lower prefetch limit and more worker processes. Or maybe we need a new celery app set up to deal with this type of thing.

I think for speed and ease though, this is probably OK. I guess I don't want it to have any knock on effect on the tasks already going to the reporting worker though....

leohemsted:
should be fine, all the nightly tasks finish at ~1am - ~1:25am

leohemsted:
W/r to "its not reporting it's periodic", i don't really see it like that - we have a bunch of nightly maintenance tasks - migrating data about, changing statuses, and i think i view it as very similar to those. If they don't run, we can run them again the next day without any huge consequences. They're generally idempotent (actually most of our scheduled tasks are).

I think I'd consider splitting the tasks into three groups. How we split those groups between workers, queues, etc I'm less fussed about, but here are the logical groups I see:

## Frequent scheduled tasks

These tasks run frequently (side note: the mix of crontab, a syntax we don't use much, and timedelta, is super confusing. `timedelta(minutes=66)` is a fave of mine 🤔). If they don't run, they'll run shortly afterwards and pick themselves back up. If they're a task that looks for inconsistent/incorrect states within notify, they'll fix it themselves (eg by restarting a job)

* 'switch-current-sms-provider-on-slow-delivery', crontab()  # every minute
* 'check-job-status', crontab()  # every minute
* 'auto-expire-broadcast-messages', timedelta(minutes=5)
* 'tend-providers-back-to-middle', crontab(minute='*/5')
* 'check-for-missing-rows-in-completed-jobs', crontab(minute='*/10')
* 'trigger-link-tests', timedelta(minutes=15)
* 'runobs', crontab(minute='0,15,30,45')
* 'replay-created-notifications', crontab(minute='0, 15, 30, 45')
* 'delete-verify-codes', timedelta(minutes=63)
* 'delete-invitations', timedelta(minutes=66)

## Nightly tasks

These tasks run every night. They move things around. If one doesn't run, the database will be in a weird and inconsistent state, at best we'll need to re-run the task, probably manually so that people during the day don't see strange things.

Some of these jobs have complex orders of precedence (eg timeout task needs to complete before we create the notification status tables, which should finish before the delete notification tasks happen, etc). Others are self contained (remove_letter_jobs is isolated and idempotent)

* 'remove-yesterdays-planned-tests-on-govuk-alerts', crontab(hour=00, minute=00)
* 'create-nightly-billing', crontab(hour=0, minute=15)
* 'create-nightly-notification-status', crontab(hour=0, minute=30)
* 'timeout-sending-notifications', crontab(hour=0, minute=5)
* 'delete-inbound-sms', crontab(hour=1, minute=40)
* 'save-daily-notification-processing-time', crontab(hour=2, minute=0)
* 'delete-notifications-older-than-retention', crontab(hour=3, minute=0)
* 'remove_sms_email_jobs', crontab(hour=4, minute=0)
* 'remove_letter_jobs', crontab(hour=4, minute=20)
* 'collate-letter-pdfs-to-be-sent', crontab(hour=17, minute=50)

## Monitoring jobs

These jobs run daily(ish) and alert the team to inconsistencies that we might need to manually intervene. They often result in a zendesk ticket being created. If they don't run one day, I'd wait for the next day rather than manually triggering them.

* 'check-if-letters-still-in-created', crontab(day_of_week='mon-fri', hour=7, minute=0)
* 'check-if-letters-still-pending-virus-check', crontab(day_of_week='mon-fri', hour='9,15', minute=0)
* 'check-for-services-with-high-failure-rates-or-sending-to-tv-numbers', crontab(day_of_week='mon-fri', hour=10, minute=30)
* 'raise-alert-if-letter-notifications-still-sending', crontab(hour=15, minute=30)
* 'raise-alert-if-no-letter-ack-file', crontab(hour=23, minute=00)

benthorner:
Nice summary @leohemsted! 

"periodic tasks" has always puzzled me, because "scheduled tasks" and "nightly tasks" are also periodic by definition! Maybe they are "less periodic" than the true "periodic" ones 😆.

Some of the names also add to the confusion e.g. `check-for-missing-rows-in-completed-jobs` sounds like a monitoring task, but can actually change state to fix the problem itself.

I can see how it would be even more effort to agree a more intuitive file structure for these tasks. I'm pleased I saw this, though, as now I know it's not just me who finds it mysterious!

PR #3391, opened 2021-12-03
Pagination approach change for `get_notifications_for_service`

idavidmcdonald:
https://www.pivotaltracker.com/story/show/179338377

Review commit by commit

There is some room for more potential refactoring after this, for example

- to fix the TODO comment
- to rename `count_pages` to `include_pagination`

But I want to first 
- get this reviewed to validate the approach I'm taking
- get this released to confirm it does indeed improve performance on this endpoint

### DIFF
--- a/tests/app/service/test_rest.py
+++ b/tests/app/service/test_rest.py
@@ -1883,6 +1883,51 @@ def test_get_notifications_for_service_without_page_count(
     assert len(resp['notifications']) == 1
     assert resp['total'] is None
     assert resp['notifications'][0]['id'] == str(without_job.id)
+    assert 'prev' not in resp['links']
+    assert 'next' not in resp['links']
+    assert 'last' not in resp['links']
+
+
+def test_get_notifications_for_service_pagination_links(
+    admin_request,
+    sample_job,
+    sample_template,
+    sample_user,
+):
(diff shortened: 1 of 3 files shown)


tests/app/service/test_rest.py
```diff
@@ -1881,8 +1881,45 @@ def test_get_notifications_for_service_without_page_count(
         count_pages=False
     )
     assert len(resp['notifications']) == 1
-    assert resp['total'] is None
     assert resp['notifications'][0]['id'] == str(without_job.id)
+    assert 'prev' not in resp['links']
+    assert 'next' not in resp['links']
+
+
+def test_get_notifications_for_service_pagination_links(
```
servingUpAces, line 1889: Do we want a test for `count_pages = False`?

servingUpAces (commented):
Looks good. I managed to test it locally. I just have a quick question.

tests/app/service/test_rest.py
idavidmcdonald: I'm leaning towards no that would be testing the internal implementation, rather than the interface. I generally try and avoid tests of the internal implementation but happy to be convinced or told that this is something you see as essential

servingUpAces (approved):

PR #3394, opened 2021-12-09
Improve response times for figuring out pagination links by doing it ourselves rather than use Flask-Sqlalchemy

idavidmcdonald:
Review commit by commit

Near identical to https://github.com/alphagov/notifications-api/pull/3391

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -150,7 +150,6 @@ from app.utils import (
     DATETIME_FORMAT_NO_TIMEZONE,
     get_prev_next_pagination_links,
     midnight_n_days_ago,
-    pagination_links,
 )
 
 service_blueprint = Blueprint('service', __name__)
@@ -533,11 +532,9 @@ def search_for_notification_by_to_field(service_id, search_term, statuses, notif
     )
     return jsonify(
         notifications=notification_with_template_schema.dump(results.items, many=True).data,
-        # TODO: this may be a bug to include the pagination links as currently `search_for_notification_by_to_field`
-        # hardcodes the pages of results to always be the first page so not sure what benefit is to show a link to
-        # page 2 which would have the page parameter ignored
-        links=pagination_links(
-            results,
+        links=get_prev_next_pagination_links(
+            1,
+            results.has_next,
             '.get_all_notifications_for_service',
             statuses=statuses,
             notification_type=notification_type,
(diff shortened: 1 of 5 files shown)


app/service/rest.py
```diff
@@ -530,11 +530,28 @@ def search_for_notification_by_to_field(service_id, search_term, statuses, notif
         page=1,
         page_size=current_app.config['PAGE_SIZE'],
     )
+
+    # We try and get the next page of results to work out if we need provide a pagination link to the next page
+    # in our response if it exists. Note, this could be done instead by changing
+    # notifications_dao.dao_get_notifications_by_recipient_or_reference to be True which will enable us to use
```
CrystalPea, line 536: > this could be done instead by changing `notifications_dao.dao_get_notifications_by_recipient_or_reference` to be True
    
What does changing `notifications_dao.dao_get_notifications_by_recipient_or_reference` to True mean?
Do you mean it could return true? That would be a totally different function then, wouldn't it?

app/service/rest.py
```diff
@@ -530,11 +530,28 @@ def search_for_notification_by_to_field(service_id, search_term, statuses, notif
...
+    # Flask-Sqlalchemy to tell if there is a next page of results but this way is much more performant for
+    # services with many results (unlike Flask SqlAlchemy, this approach doesn't do an additional query to count
+    # all the results of which there could be millions but instead only asks for a single extra page of results).
+    next_page_of_pagination = notifications_dao.dao_get_notifications_by_recipient_or_reference(
```
CrystalPea, line 540: Do I understand correctly that doing the paginated query two times without counting overall number of notifications is faster than doing the query once, but including the count? Wow :o

CrystalPea (approved):
Looks good to me 👍🏼

app/service/rest.py
idavidmcdonald: That is correct!

Before this PR, you'd do 2 queries.
Get me the first 50
Get all notifications and count how many there are

Now we are doing
Get me the first 50
Get me the next 50

I saw that the "Get all notifications and count how many there are" may take 100 seconds for some high volume services (think having to query for the count of 10 million notifications with a variety of filters) but the first 'Get me the first 50' would take under a second.

There may be a few edge cases where this new suggested approach is actually a little slower, but that is potentially for services that send very few notifications and therefore it would be from a quick query to a very very slightly slower quick query which won't have any impact for users.

app/service/rest.py
idavidmcdonald: Good point Pea, this was inaccurate. I've updated the docs there.

PR #3397, opened 2021-12-10
run the notification delete task per service

leohemsted:
Long running tasks are risky. They can easily be disrupted by cells rolling, us deploying, or a bug causing an unexpected exception, and we'll have to run them again to fix them, taking a long time and blocking the pipeline for that duration. If a task is grabbed by celery but there is no worker available in the next five minutes, then that task will be duplicated due to SQS celery timeout.

Short tasks are less risky. Especially shorter than 10 seconds, when celery _should_ give them enough grace period to exit gracefully on event of an app being told to shutdown. 

The three notification-type delete tasks takes up to three hours to run at the moment (the sms and email tasks finish within an hour and the letter task takes longer as it has to go to s3 and delete each letter individually)

In this PR, I propose that we break up the delete task into a massively parallel series of tasks, per service and notification type. Note that in the previous version of this, we did that anyway - but in a for loop rather than in separate tasks.

The parallel tasks should be able to take advantage of our large postgres instance - something which is very good at handling concurrent interactions. I'm not really worried about the impact of concurrent writes to notification_history, since we're already doing this 3x with the three notification types.

I suspect, though I haven't confirmed, that this will be quicker than what we currently do, which is run the three notification types in parallel. Currently the tasks can take up to three hours to run three concurrent tasks. We currently always have three reporting boxes, and they have four workers each, for twelve concurrent tasks. So I'd hope that we'd see around a four times improvement in the duration it takes (ie: <45 minutes).


app/dao/notifications_dao.py
```diff
@@ -433,18 +379,23 @@ def insert_notification_history_delete_notifications(
...
+    qry_limit=50000
+):
     deleted = 0
     if notification_type == LETTER_TYPE:
         _delete_letters_from_s3(
-            notification_type, service_id, day_to_delete_backwards_from, qry_limit
+            notification_type, service_id, timestamp_to_delete_backwards_from, qry_limit
         )
     delete_count_per_call = 1
     while delete_count_per_call > 0:
```
leohemsted, line 394: if we wanted to be really visibility timeout-concious, we could split into tasks at this level.

ie: "delete 50,000 notifications. If there's potentially more to delete, then enqueue a new task to delete the next batch of 50,000".

That would ensure all the tasks were a bit smaller, and much much more likely to stay under 5 minutes.

I don't think we should do that now, and the complexity could be a bit hard to approach, but it's a potential next step if we really wanted to keep task durations lower.

app/celery/nightly_tasks.py
```diff
@@ -73,45 +78,74 @@ def delete_notifications_older_than_retention():
...
+def delete_notifications_for_service_and_type(service_id, notification_type, datetime_to_delete_before):
+    start = datetime.utcnow()
+    num_deleted = move_notifications_to_notification_history(
+        notification_type,
+        service_id,
+        datetime_to_delete_before,
+    )
+    end = datetime.utcnow()
+    current_app.logger.info(
+        f'Deleted {num_deleted} {notification_type} notifications for '
```
servingUpAces, line 147: Can we change this log message so that it's easier to search? For instance if the message starts with something like `delete-notifications-for-service-and-type has deleted {notification_type} for service {id} in {end-start} seconds`

servingUpAces (commented):
I think this is worth a try. I wonder if there is a way to reconcile if all the tasks that have been created are executed. Or if all the tasks get created in  `_delete_notifications_older_than_retention_by_type` before an app restart happens. 
Also I'm guessing you'll change the config scheduling before deploying this?

app/celery/nightly_tasks.py
leohemsted: yes! i thought that the other day and then promptly forgot about it, that's a great idea.

servingUpAces (approved):

PR #3398, opened 2021-12-13
Scale timeout task to work on arbitrary volumes

benthorner:

This makes the temporary hack we put in more permanent:
we can now time out any number of notifications, not just
the arbitrary 500K limit we put in.

Please review commit by commit.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -505,9 +505,6 @@ def dao_timeout_notifications(cutoff_time, limit=100000):
     ).limit(limit).all()
 
     Notification.query.filter(
-        Notification.created_at < cutoff_time,
-        Notification.status.in_(current_statuses),
-        Notification.notification_type.in_([SMS_TYPE, EMAIL_TYPE]),
         Notification.id.in_([n.id for n in notifications]),
     ).update(
         {'status': new_status, 'updated_at': updated_at},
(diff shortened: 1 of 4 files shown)


app/dao/notifications_dao.py
```diff
@@ -489,35 +489,22 @@ def dao_delete_notifications_by_id(notification_id):
...
         Notification.notification_type.in_([SMS_TYPE, EMAIL_TYPE])
-    ).limit(100000).all()
+    ).limit(limit).all()
 
     Notification.query.filter(
-        Notification.created_at < timeout_start,
-        Notification.status.in_(current_statuses),
-        Notification.notification_type.in_([SMS_TYPE, EMAIL_TYPE]),
         Notification.id.in_([n.id for n in notifications]),
     ).update(
```
servingUpAces, line 509: A thought about optimizing this update query: put the list of `notifications` in a temp table and join on that temp table instead of using the `in` filter. 
Using the `in` operator is inefficient for large lists, which this is. 
This slightly complicates the code, but I suspect the performance gain would be significant.

servingUpAces (approved):

app/dao/notifications_dao.py
benthorner: Oh, interesting, I've not used that technique before and can see how it's a more general solution to this kind of problem, which would mean we don't have to operate in batches. 

If I was starting this work again then I'd want to explore using a join table. For now, this is plenty fast enough for normal volumes (1-10 mins, on average, per day), and can also scale to cope with the odd deluge.

PR #3400, opened 2021-12-14
bump lxml to fix security warning

leohemsted:
two vulnerabilities in <4.6.5 (GHSL-2021-1037 and GHSL-2021-1038)
https://github.com/lxml/lxml/blob/master/CHANGES.txt

also removes docopt as we don't use it except for a dev script (which we might not need anyway)


requirements.txt
```diff
@@ -73,12 +73,10 @@ cryptography==3.3.2
...
     # via eventlet
 docopt==0.6.2
-    # via
-    #   -r requirements.in
-    #   notifications-python-client
+    # via notifications-python-client
 docutils==0.15.2
     # via awscli
-eventlet==0.30.2 # pyup: ignore
+eventlet==0.30.2
```
leohemsted, line 79: dunno why this line has changed 🤔

benthorner: I think we should put it back to stop PyUp trying to upgrade it.

leohemsted: hmmm i'm not sure, i don't think that's needed - we still have the line in requirements.in (https://github.com/alphagov/notifications-api/blob/master/requirements.in#L13) . we don't have any of the other pyup lines reflected in this file (cryptography, marshmallow, etc are all pinned with `# pyup:< 1.2.3` comments in reqs.in)

benthorner (approved):
Thanks for linking to the changelog. The other changes do indeed appear minor.

Can we just remove `scripts/delete_sqs_queues`?

- It's not documented anywhere.
- The file name differs from its function (delete **+ list**).
- It's in VC, we can always bring it back.

requirements.txt
benthorner: Maybe it's this: https://github.com/alphagov/notifications-admin/commit/fab44e016a2d65ff04675a139c5a8386dfd8a1d2

benthorner: Aha, yeah, that would explain why you're now getting the safety warning!

leohemsted: ahhh good ifnd. i'll re-add that back in.

PR #3408, opened 2021-12-21
Close DB connection whilst making HTTP to SMS providers

idavidmcdonald:
At the moment, when we are processing and sending an SMS we open
a DB connection at the start of the celery task and then close it
at the end of the celery task. Nice and simple.

However, during that celery task we make an HTTP call out to our
SMS providers. If our SMS providers have problems or response times
start to slow then it means we have an open DB connection sat waiting
for our SMS providers to respond which could take seconds. If our
SMS providers grind to a halt, this would cause all of the
celery tasks to hold on to their connections and we would run out
of DB connections and Notify would fall over.

We think we can solve this by closing the DB session which releases
the DB connection back to the pool.

Note, we've seen this happen in staging during load testing if our
SMS provider stub has fallen over. We've never seen it in production
and it may be less unlikely to happen as we are balancing traffic
across two providers and they generally have very good uptime.

One downside to be aware of is there could be a slight increase in
time spent to send an SMS as we will now spend a bit of extra time
closing the DB session and then reopening it again after the HTTP
request is done.

Note, there is no reason this approach couldn't be copied for our
email provider too if it appears successful.

Questions
- is it worth it or is it mostly protecting staging against our load test sms stub which could maybe have a different solution? Is this an approach we want to go down? 
- should we put this behind a feature flag for staging only?
- we should probably run a load test against this to check it doesn't introduce a performance degradation?

Note, I haven't tested locally as my celery is borked...

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -10,7 +10,7 @@ from notifications_utils.template import (
     SMSMessageTemplate,
 )
 
-from app import create_uuid, notification_provider_clients, statsd_client
+from app import create_uuid, db, notification_provider_clients, statsd_client
 from app.celery.research_mode_tasks import (
     send_email_response,
     send_sms_response,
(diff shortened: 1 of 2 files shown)


servingUpAces:
This is a good idea and I do not expect a performance degradation. I think we do not need to feature flag this and only add it for staging. This can help if our providers do give us slower responses to the send_sms HTTP request (which has happened in the past). I believe this is a viable approach.
I have tested this locally - all good.

sakisv:
Happy to give it a load test on staging before it lands on prod.

sakisv (approved):

PR #3412, opened 2021-12-23
Reduce pool size from 30 to 15 connections

sakisv:
Having a pool size of 30 connections means that if we receive a big
number of requests, with the current configuration, the API would end up
holding onto 30 connections per worker * 4 workers per instance * 35
instances = 4200 connections. With a limit of 5000 connections, this
means that we would only have 800 connections to share between the
workers or for overflow usage (btw, even the overflow for the API would
take us above the 5000 limit - 10 overflow connections per worker * 4 *
35 = 1400 connections, total 5600 _only_ for the API).

During our load tests this led to a deadlock situation where nothing
could retrieve connections to deal with a queue build-up.

The reduced pool size allowed for a much more graceful degradation of
the service where, after significant load we would increase the response
times but still manage to serve all the requests.

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -2,7 +2,7 @@
   'notify-api': {
     'NOTIFY_APP_NAME': 'api',
     'disk_quota': '2G',
-    'sqlalchemy_pool_size': 30,
+    'sqlalchemy_pool_size': 15,
     'additional_env_vars': {
       'STATSD_HOST': None
     },


idavidmcdonald (approved):

idavidmcdonald:
Just to add if someone comes back to visit this. 15 was chosen based on 4 (processes) x (15 conn + 10 overflow) x 35 instances = 3500 DB connections used maximum.

We haven't tested at other numbers but this seems in the right ballpark to save 1500 for our other apps. It may be possible to increase from 15 slightly but this is in the right ballpark to start with now.

benthorner:
For anyone else coming to this, the concept of "overflow" in SqlAlchemy was new to me: https://docs.sqlalchemy.org/en/14/core/pooling.html#sqlalchemy.pool.QueuePool.__init__

I wonder if we should allow overflow at all? I guess they buy us a bit of time to start other instances.

idavidmcdonald:
It is a good question about the overflow. Always something I've just taken for granted that we have.

sakisv:
A few thoughts around the overflow:

In the current state of the world, if we have a big wave of traffic which will cause our instances to use all the connections in their pool, we will end up having 2100 connections reserved (4 processes * 15 connections * 35 instances). 

These connections are not being released when the traffic goes down, even though we configure the [`POOL_RECYCLE`](https://github.com/alphagov/notifications-api/blob/494a01ba57aeb46d3622b4895a1373a3cc47371c/app/config.py#L147).

The overflow connections do not suffer from this problem: They are not kept in the pool, so they are released immediately and they become available to be used by other apps.

I can see three ways of tackling this issue:

1. Start scaling the API instances again. When we scale down, we'll release the connections reserved by that instance
2. Further reduce the connection pool size but increase the overflow
3. Investigate why the connections are not being released, _if_ they should be released, and how we can fix it properly

Obviously this is not as big a problem as it was before this PR, but holding on to so many connections when we don't need them, is still something that we should look at.

benthorner:
@sakisv could you clarify what the problem is here? Do you think some other app will need the connections?

Reading this again got me thinking: 35 is [the max instances for the API app](https://github.com/alphagov/notifications-paas-autoscaler/blob/master/Makefile#L115), but there's also the various workers, which have their own limits and their own connection pools. Were we seeing all the connections coming from the API app? 

I think you've proven the new limit works well in practice, based on your subsequent testing. It's just a bit mysterious is all: is it better to have more instances, or more connections per instance; what's the sweet spot? 

(Given the results of your tests, perhaps 15 is the sweet spot 😄.)

sakisv:
> Do you think some other app will need the connections?

Indeed that is exactly the problem.

> Were we seeing all the connections coming from the API app?

Yes, we actually ran a load test where we exhausted the connections, we stopped the load test and waited for things to settle down. Then we saw that the DB connections were still above 4000. Once we restarted the API instances, all these connections were released.

> It's just a bit mysterious is all: is it better to have more instances, or more connections per instance; what's the sweet spot?

That's a great question. I think we're more or less there, the current config seems to take us ~a long~ some way. 

The question I'm posing is whether or not we should do something about all these connections that are not being released. We may decide that given the current priorities/constraints this is an acceptable risk, which is fine as long as we're being explicit about it.

benthorner:
OK, so we're saying the API is hogging DB connections that something else wants to use i.e. we potentially still risk running out of available connections, unless we recycle these ones after X period of time (maybe immediately).

What do we know about these other apps? I'm wondering if it's like:

1. API app does a load of work (so takes all the connections).
2. Then the sender worker needs to do all the work to send (so now it wants all the connections).
3. Then the receipts worker need to do all the work (ditto).

I think in reality they should all be working flat out at the same time. Looking at the behaviour for each app seems like they way to go if we want to quantify the risk. As long as we don't hit connection issues with this change, then it's probably not worth it - assuming our load tests are realistic, of course.

By the way, I had a look at [the original commit](https://github.com/alphagov/notifications-api/commit/20d5a946f68d78d1cafbc9a93d13955597316f6a), which doesn't say much about the recycle setting. [The docs](https://flask-sqlalchemy.palletsprojects.com/en/2.x/config/) suggest it might not be applicable to Postgres, so maybe it never worked.

rjbaker:
Looks API's SQLAlchemy config actually needs some work. API is [currently using Flask 1.2](https://github.com/alphagov/notifications-api/blob/master/requirements.txt#L81)
In v2.4 all the `SQLALCHEMY_POOL_*` config vars (and others) passed to Flask are deprecated.

Instead SQLAlchemy is configured by Flask using `SQLALCHEMY_ENGINE_OPTIONS` ([from Ben's link above](https://flask-sqlalchemy.palletsprojects.com/en/2.x/config/)) which accepts a dictionary of options fed to SQLAlchemy create_engine method. All options are here: https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine

benthorner:
@rjbaker just to check: are you saying it's broken, or just needs a tidy up due to deprecations?

rjbaker:
I think it will need work to upgrade Flask to 2.4, I was just looking into why the `SQLALCHEMY_POOL_RECYCLE` config doesn't appear to be taking effect.

In response to your earlier points, Celery workers [use a single DB connection](https://github.com/alphagov/notifications-api/blob/master/manifest.yml.j2#L112), Celery itself maintains the worker pool size via the concurrency setting. So they're not as connection-hungry as the web service and use comparatively few, but retain some headroom for scaling out.

This way we can (currently very roughly) ensure that under max load there are enough connections available for all moving parts. Agree that this should be measured with everything working flat out 👍

benthorner:
@rjbaker oh that's interesting, thanks for the pointer. I didn't realise there was such a big difference. Then the connection recycling doesn't seem like such a big deal, as the total connections - 2.1K + overflow + 1 x each worker - should always be well within the limit, assuming API instances don't grow much more.

Still, I agree it's worth investigating why it's not behaving as expected: we may learn something useful, and it's confusing to have config that doesn't actually do anything.

PR #3413, opened 2021-12-24
Bump sqlalchemy statement timeout even higher for reporting worker

idavidmcdonald:
We saw it fail again last night to calculate how many notifications
were sent for one of our services to put in the ft_notification_status
table. It ran in to the sqlalchemy statement timeout again.
To get us through the holiday
period lets make it 2 hours as surely that will be enough and then
we can fix this properly

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -58,7 +58,7 @@
     'additional_env_vars': {
       'CELERYD_MAX_TASKS_PER_CHILD': 1,
       'CELERYD_PREFETCH_MULTIPLIER': 1,
-      'SQLALCHEMY_STATEMENT_TIMEOUT': 2400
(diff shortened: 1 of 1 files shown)


rjbaker (approved):
The Postgres `statement_timeout` option is applied in milliseconds. It's passed directly to the the Postgres session client beneath SQLAlchemy here, rather than being a specific SQLAlchemy setting: https://github.com/alphagov/notifications-api/blob/bffca39223b419d7f2f8a637d7b8be44a567184e/app/__init__.py#L49

Postgres docs: https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-STATEMENT-TIMEOUT

Agree this is an excessively long running query and either needs refactoring or different solution/data model to create the same report. 👍

PR #3419, opened 2022-01-06
Improve logging for adjusting provider priority on API failure

idavidmcdonald:
We can deduce that our traffic is being reduced for one of our providers
but we aren't logging that we are doing it so it is much harder to
tell.

This commit
- makes it explicit what we are doing with a log line
- also gives context on what the error was with the provider API
  so we can tell if it is say a 4xx or a 5xx

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -82,7 +82,12 @@ def send_sms_to_provider(notification):
             except Exception as e:
                 notification.billable_units = template.fragment_count
                 dao_update_notification(notification)
-                dao_reduce_sms_provider_priority(provider.get_name(), time_threshold=timedelta(minutes=1))
+
+                provider_name = provider.get_name()
+                current_app.logger.info(
+                    "Reducing priority for {provider_name} due to exception raised when calling provider API"
+                )
+                dao_reduce_sms_provider_priority(provider_name, time_threshold=timedelta(minutes=1))
                 raise e
             else:
                 notification.billable_units = template.fragment_count
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -24,7 +24,7 @@ def deliver_sms(self, notification_id):
     except Exception as e:
         if isinstance(e, SmsClientResponseException):
             current_app.logger.warning(
-                "SMS notification delivery for id: {} failed".format(notification_id)
+                "SMS notification delivery for id: {} failed due to {}".format(notification_id, e)
             )
         else:
             current_app.logger.exception(


app/delivery/send_to_providers.py
```diff
@@ -82,7 +82,12 @@ def send_sms_to_provider(notification):
             except Exception as e:
                 notification.billable_units = template.fragment_count
                 dao_update_notification(notification)
-                dao_reduce_sms_provider_priority(provider.get_name(), time_threshold=timedelta(minutes=1))
+
+                provider_name = provider.get_name()
+                current_app.logger.info(
+                    "Reducing priority for {provider_name} due to exception raised when calling provider API"
```
leohemsted, line 88: needs an `f` format string

app/delivery/send_to_providers.py
```diff
@@ -82,7 +82,12 @@ def send_sms_to_provider(notification):
...
+                )
+                dao_reduce_sms_provider_priority(provider_name, time_threshold=timedelta(minutes=1))
```
leohemsted, line 90: TLDR: I don't think we need this log line to answer the email. We can see that we had a bunch of MMG 503 failures [by searching for "API POST request failed" in kibana]( https://kibana.logit.io/s/9423a789-282c-4113-908d-0be3b1bc9d1d/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(message),filters:!(),index:'logstash-*',interval:auto,query:(query_string:(query:'%22API%20POST%20request%20failed%20on%22')),sort:!('@timestamp',desc)) )

unfortunately it might not be so simple. 

this function might not change sms provider priority if:

* the priority is already 0/100
* the priority has changed (in either direction) within the last `time_threshold`

this means that if we have some exceptions close to each other, we'll get more of these logs than we'll actually adjust the priority. That might be okay? (I think that's probably okay) but it's worth thinking about.

If we're asking "Did we lower the priority at 10:21am today due to errors" then we'll partially be able to answer that - the answer will be "maybe, but it might have also been due to delivery receipt slowness at the same time".

If we're asking "How many times did we lower the priority due to a 5xx" then we still won't be able to answer that.

If we're asking "Did a particular 5xx from firetext cause a shift in priority, we can already answer that by looking at the other log lines from the request id ([example]( https://kibana.logit.io/s/9423a789-282c-4113-908d-0be3b1bc9d1d/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:requestId,negate:!f,type:phrase,value:'7a565e1e4e330f06'),query:(match:(requestId:(query:'7a565e1e4e330f06',type:phrase))))),index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'*')),sort:!('@timestamp',desc)) )).

Do we have a metric for provider response codes? If we trust our app to act on those (which we do) is it easier to have a grafana visualisation of that rather than looking through logs? (You can see those logs by searching "API POST request failed on").

app/delivery/send_to_providers.py
leohemsted: Not essential, but it would be nice if this log line and the log line for slow providers (https://github.com/alphagov/notifications-api/blob/master/app/celery/scheduled_tasks.py#L117) looked similar so we could easily search for both of them to understand the full picture. Or following ben's lead, and make them a metric we can surface in grafana rather than searching in logs 🤔

PR #3417, opened 2022-01-05
Optimise query to populate notification statuses

benthorner:
https://www.pivotaltracker.com/story/show/180693991

Investigation with EXPLAIN and EXPLAIN ANALYZE for the notification history
table shows this is another instance of [1] but for the key type column.
Swapping "!=" for "IN" solves the problem.

## Query plans (EXPLAIN ANALYZE)

Before:

```
 GroupAggregate  (cost=48285296.89..48304545.97 rows=699630 width=72) (actual time=382883.029..383541.435 rows=80 loops=1)
   Group Key: template_id, service_id, job_id, key_type, notification_status
   ->  Sort  (cost=48285296.89..48287047.29 rows=700159 width=64) (actual time=382734.683..383303.675 rows=1464049 loops=1)
         Sort Key: template_id, job_id, key_type, notification_status
         Sort Method: external merge  Disk: 84736kB
         ->  Index Scan using ix_notification_history_service_id_composite on notification_history  (cost=0.70..48190993.82 rows=700159 width=64) (actual time=166574.747..381321.318 rows=1464049 loops=1)
               Index Cond: ((service_id = 'c5956607-20b1-48b4-8983-85d11404e61f'::uuid) AND (notification_type = 'email'::notification_type) AND (created_at >= '2021-12-20 00:00:00'::timestamp without time zone) AND (created_at < '2021-12-21 00:00:00'::timestamp without time zone))
               Filter: ((key_type)::text <> 'test'::text)
 Planning Time: 0.115 ms
 Execution Time: 383553.795 ms
```

After:

```
 GroupAggregate  (cost=2513920.30..2533169.38 rows=699630 width=72) (actual time=14248.780..14864.704 rows=80 loops=1)
   Group Key: template_id, service_id, job_id, key_type, notification_status
   ->  Sort  (cost=2513920.30..2515670.69 rows=700159 width=64) (actual time=14110.324..14638.207 rows=1464049 loops=1)
         Sort Key: template_id, job_id, key_type, notification_status
         Sort Method: external merge  Disk: 84736kB
         ->  Index Scan using ix_notification_history_service_id_composite on notification_history  (cost=0.70..2419617.23 rows=700159 width=64) (actual time=1.883..12851.183 rows=1464049 loops=1)
[...]

PR #3421, opened 2022-01-10
Add comment to explain status aggregation approach

benthorner:

This relates to the performance optimisation work we're doing [1].
Before optimising the task, it's worth asking if we can do less -
the comment explains why it has to be this way.

Some references to back up the comment:

- We do status updates in either table [2].
- We don't allow duplicate receipts for emails [3].
- We don't allow duplicate receipts for SMS [4].
- We don't expect duplicate receipts for letters.

This is something we would need to revisit if we want to support
additional status updates - we could reject based on the age of the
notification, rather than the status.

[1]: https://github.com/alphagov/notifications-api/pull/3417
[2]: https://github.com/alphagov/notifications-api/blob/20ead82463a9df1ce3a6325d103a00306609f5fb/app/dao/notifications_dao.py#L538
[3]: https://github.com/alphagov/notifications-api/blob/20ead82463a9df1ce3a6325d103a00306609f5fb/app/celery/process_ses_receipts_tasks.py#L58
[4]: https://github.com/alphagov/notifications-api/blob/20ead82463a9df1ce3a6325d103a00306609f5fb/app/dao/notifications_dao.py#L129-L135

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -66,6 +66,29 @@ def create_nightly_billing_for_day(process_day):
 @notify_celery.task(name="create-nightly-notification-status")
 @cronitor("create-nightly-notification-status")
 def create_nightly_notification_status():
+    """
+    Aggregate notification statuses into rows in ft_notification_status.
+    In order to minimise effort, this task assumes that:
+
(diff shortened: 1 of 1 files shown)


leohemsted (approved):
Fab stuff

PR #3422, opened 2022-01-11
Fix bug in organisation report for its services and usages.

servingUpAces:
If a service has not sent any SMS for the financial year the free allowance was showing up as 0 rather than the number in annual billing. The query has been updated to use an outer join so that the free allow will be returned when there is no ft_billing.

There is a potential performance enhancement to only return the data for the services of the organisation in the `fetch_sms_free_allowance_remainder_until_date` subquery. I will investigate in a subsequent PR.

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -599,7 +599,7 @@ def fetch_sms_billing_for_organisation(organisation_id, start_date, end_date):
     # ASSUMPTION: AnnualBilling has been populated for year.
     allowance_left_at_start_date_query = fetch_sms_free_allowance_remainder_until_date(start_date).subquery()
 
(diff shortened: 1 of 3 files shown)


klssmith (approved):

CrystalPea (commented):
Good spot 👏🏼 . Would we have to do the same for `fetch_sms_billing_for_all_services` on line 65? Those 2 functions use the same pattern.

servingUpAces:
> Good spot 👏🏼 . Would we have to do the same for `fetch_sms_billing_for_all_services` on line 65? Those 2 functions use the same pattern.

@CrystalPea Good question, but no we don't want to change that query. That query is for the billing report used for at the end of a quarter. If we include services with no ft_billing it will bloat the report with extra data. Caley is only interested in services that we need to bill.

PR #3425, opened 2022-01-12
Parallelise status aggregation by service and day

benthorner:

This follows a similar approach as [1]. Recently we've seen lots of errors
from this task, which we think are a consequence of it doing too much work
and tripping Celery's visibility timeout.

While we can optimise the query [2], it's likely the errors will return as
the number of live services grows. Parallelising the aggregation now will
make it more futureproof.

Because the new logs will be broken down by service, I'll also be able to 
confirm if we can safely revert https://github.com/alphagov/notifications-api/pull/3411 - at the moment, it's unclear what 
the max per-service query time is in the DAO loop.

I've broken down the changes into a few commits.

Tested locally to check the tasks are queued up as expected ✅.

## Deployment

Should only be deployed when today's status aggregation task has
finished. This should be done by the start of the working day, but worth 
checking in Cronitor to make sure it's not got stuck.

The parallelisation will result in a much larger number of tasks being 
generated, which could potentially starve other scheduled tasks. I don't 
think this will be a problem, given the current runtimes:

|         bin(1d)         | avg(duration) | max(duration) |
|-------------------------|---------------|---------------|
| 2022-01-18 00:00:00.000 | 341.0841      | 848.7002      |
| 2022-01-17 00:00:00.000 | 338.8308      | 948.1304      |
| 2022-01-16 00:00:00.000 | 551.6605      | 1684.3972     |
| 2022-01-15 00:00:00.000 | 632.2731      | 2115.1576     |
| 2022-01-14 00:00:00.000 | 794.273       | 2561.2193     |
| 2022-01-13 00:00:00.000 | 593.8976      | 1975.6033     |

With all the work split up, I expect much lower average runtimes per 
task. There also aren't any critical tasks scheduled to execute near 
this one, and we have plenty of workers ([3](https://github.com/alphagov/notifications-paas-autoscaler/blob/c217b56a06d36cd520a6ad868eb140f9257f30bd/Makefile#L109) x [4](https://github.com/alphagov/notifications-api/blob/5cd6fcbb4feec983ae39a7ae8b8ad3132f534e5d/scripts/paas_app_wrapper.sh#L32)).

[1]: https://github.com/alphagov/notifications-api/pull/3397
[2]: https://github.com/alphagov/notifications-api/pull/3417

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -91,30 +91,20 @@ def create_nightly_notification_status():
 
     yesterday = convert_utc_to_bst(datetime.utcnow()).date() - timedelta(days=1)
 
-    # email and sms
-    for i in range(4):
-        process_day = yesterday - timedelta(days=i)
-        for notification_type in [SMS_TYPE, EMAIL_TYPE]:
+    for notification_type in [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]:
+        days = 10 if notification_type == LETTER_TYPE else 4
+
+        for i in range(days):
+            process_day = yesterday - timedelta(days=i)
+
             create_nightly_notification_status_for_day.apply_async(
                 kwargs={'process_day': process_day.isoformat(), 'notification_type': notification_type},
                 queue=QueueNames.REPORTING
             )
             current_app.logger.info(
-                f"create-nightly-notification-status task: create-nightly-notification-status-for-day task created "
+                f"create-nightly-notification-status-for-day task created "
                 f"for type {notification_type} for {process_day}"
             )
-
-    # letters
-    for i in range(10):
-        process_day = yesterday - timedelta(days=i)
-        create_nightly_notification_status_for_day.apply_async(
-            kwargs={'process_day': process_day.isoformat(), 'notification_type': LETTER_TYPE},
-            queue=QueueNames.REPORTING
-        )
-        current_app.logger.info(
-            f"create-nightly-notification-status task: create-nightly-notification-status-for-day task created "
-            f"for type letter for {process_day}"
-        )
 
 
 @notify_celery.task(name="create-nightly-notification-status-for-day")
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -91,54 +91,58 @@ def create_nightly_notification_status():
 
     yesterday = convert_utc_to_bst(datetime.utcnow()).date() - timedelta(days=1)
 
-    for notification_type in [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]:
-        days = 10 if notification_type == LETTER_TYPE else 4
+    for (service_id,) in db.session.query(Service.id):
+        for notification_type in [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]:
+            days = 10 if notification_type == LETTER_TYPE else 4
 
-        for i in range(days):
-            process_day = yesterday - timedelta(days=i)
+            for i in range(days):
+                process_day = yesterday - timedelta(days=i)
 
-            create_nightly_notification_status_for_day.apply_async(
-                kwargs={'process_day': process_day.isoformat(), 'notification_type': notification_type},
-                queue=QueueNames.REPORTING
-            )
-            current_app.logger.info(
-                f"create-nightly-notification-status-for-day task created "
-                f"for {notification_type} and {process_day}"
-            )
+                create_nightly_notification_status_for_service_and_day.apply_async(
+                    kwargs={
+                        'process_day': process_day.isoformat(),
+                        'notification_type': notification_type,
+                        'service_id': service_id,
+                    },
+                    queue=QueueNames.REPORTING
+                )
+                current_app.logger.info(
+                    f"create-nightly-notification-status-for-day task created "
+                    f"for {service_id}, {notification_type} and {process_day}"
+                )
 
 
-@notify_celery.task(name="create-nightly-notification-status-for-day")
-def create_nightly_notification_status_for_day(process_day, notification_type):
+@notify_celery.task(name="create-nightly-notification-status-for-service-and-day")
+def create_nightly_notification_status_for_service_and_day(process_day, service_id, notification_type):
     process_day = datetime.strptime(process_day, "%Y-%m-%d").date()
     current_app.logger.info(
         f'create-nightly-notification-status-for-day task started '
-        f'for {notification_type} and {process_day}'
+        f'for {service_id}, {notification_type} and {process_day}'
     )
 
-    for (service_id,) in db.session.query(Service.id):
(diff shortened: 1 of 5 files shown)


benthorner:
## Update

Short version: this PR is necessary to stop the current errors

On Friday 14th January we had a couple of errors due to duplicate tasks, after two days of no errors. The following timing data backs up how https://github.com/alphagov/notifications-api/pull/3417 has (probably) helped, but not fully solved the problems we're seeing with this task.

Screenshot 2022-01-14 at 09 18 22

leohemsted (commented):
Just a couple of minor changes but it looks good overall and i'm deffo in favour of breaking up these big tasks into smaller parallel chunks

app/celery/reporting_tasks.py
```diff
@@ -124,17 +124,28 @@ def create_nightly_notification_status_for_day(process_day, notification_type):
...
-    start = datetime.utcnow()
-    transit_data = fetch_notification_status_for_day(process_day=process_day, notification_type=notification_type)
-    end = datetime.utcnow()
-    current_app.logger.info(
-        f'create-nightly-notification-status-for-day task for {process_day} type {notification_type}: '
-        f'data fetched in {(end - start).seconds} seconds'
-    )
+    for (service_id,) in db.session.query(Service.id):
+        start = datetime.utcnow()
+        transit_data = fetch_status_data_for_service_and_day(
```
CrystalPea: Optional: I know you didn't name this, but `transit_data` is such a nebulous variable name, it doesn't tell much about what it holds. When reviewing `update_fact_notification_status` function, it took me a bit to understand what it holds. Could we rename it to `new_status_data` or something to that effect here and in `update_fact_notification_status`?

app/dao/fact_notification_status_dao.py
```diff
@@ -86,18 +77,19 @@ def query_for_fact_status_data(table, start_date, end_date, notification_type, s
...
     FactNotificationStatus.query.filter(
         FactNotificationStatus.bst_date == process_day,
-        FactNotificationStatus.notification_type == notification_type
+        FactNotificationStatus.notification_type == notification_type,
+        FactNotificationStatus.service_id == service_id,
     ).delete()
 
-    for row in data:
+    for row in transit_data:
         stmt = insert(table).values(
```
CrystalPea, line 78: Optional: again, I know you didn't name this, but would it be possible to rename `stmt` to `statement`, or even better `new_notification_status` or something like that? It would reduce mental load when reading this part of code.

app/celery/reporting_tasks.py
```diff
@@ -103,15 +103,16 @@ def create_nightly_notification_status():
             )
             current_app.logger.info(
                 f"create-nightly-notification-status-for-day task created "
-                f"for type {notification_type} for {process_day}"
+                f"for {notification_type} and {process_day}"
```
CrystalPea: "task created for sms *for* 12-01-2022" makes a more readable sentence than "task created for sms *and* 12-01-2022"

app/commands.py
```diff
@@ -281,87 +276,6 @@ def setup_commands(application):
     application.cli.add_command(command_group)
 
 
-@notify_command(name='migrate-data-to-ft-billing')
-@click.option('-s', '--start_date', required=True, help="start date inclusive", type=click_dt(format='%Y-%m-%d'))
-@click.option('-e', '--end_date', required=True, help="end date inclusive", type=click_dt(format='%Y-%m-%d'))
-@statsd(namespace="tasks")
-def migrate_data_to_ft_billing(start_date, end_date):
```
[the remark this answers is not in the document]
CrystalPea: I agree this is one for @servingUpAces to double-check.

CrystalPea (approved):
Nice job, I left a few comments, mostly about readability.

app/celery/reporting_tasks.py
```diff
@@ -91,50 +91,58 @@ def create_nightly_notification_status():
...
-            f"create-nightly-notification-status task: create-nightly-notification-status-for-day task created "
-            f"for type letter for {process_day}"
-        )
-
-
-@notify_celery.task(name="create-nightly-notification-status-for-day")
-def create_nightly_notification_status_for_day(process_day, notification_type):
+    for (service_id,) in db.session.query(Service.id):
+        for notification_type in [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]:
+            days = 10 if notification_type == LETTER_TYPE else 4
```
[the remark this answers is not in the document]
benthorner: Hmm...this one is grabbing me as immediately better, so I'll leave as-is. 

Points for an interesting technique, though 😄.

app/dao/fact_notification_status_dao.py
```diff
@@ -36,33 +35,15 @@
...
-        all_data_for_process_day += data_for_service_and_type
-
-    return all_data_for_process_day
-
-
-def query_for_fact_status_data(table, start_date, end_date, notification_type, service_id):
-    query = db.session.query(
+    return db.session.query(
         table.template_id,
         table.service_id,
```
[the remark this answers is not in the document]
benthorner: Great spot! I've removed in https://github.com/alphagov/notifications-api/pull/3425/commits/9686595fa85c1fc5458f26c3afcc893a8aa93351.

app/celery/reporting_tasks.py
benthorner: Sure, I remember thinking the same thing.

I've gone with `new_status_rows`, where "rows" is even more specific.

app/dao/fact_notification_status_dao.py
benthorner: Hmm...I'm not sure we benefit from a variable there at all.

I've made it an inline expression in https://github.com/alphagov/notifications-api/pull/3425/commits/9686595fa85c1fc5458f26c3afcc893a8aa93351 (so no more `stmt` 😄).

Does that work for you?

app/celery/reporting_tasks.py
benthorner: Looking through these again I realise some of them are in different orders as well. In https://github.com/alphagov/notifications-api/pull/3425/commits/9686595fa85c1fc5458f26c3afcc893a8aa93351 I've gone through and made the ordering consistent, so we can have "for" on the end 👍.

CrystalPea: Nice one 👍🏼

app/commands.py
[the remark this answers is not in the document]
benthorner: I'm going to merge this now. 

We can always add these commands back later on (+ some tests to make sure they work 😉).

benthorner:
Thanks for both your reviews @leohemsted and @CrystalPea !

benthorner:
## Update: quick test on Preview

Testing on Preview (1500 services) yielded about 30K tasks, which took about 30 mins to process. This runtime is quite similar to Production, except that a few services take longer to execute and thus finish later on.

PR #3432, opened 2022-01-19
unpin cryptography

leohemsted:
we previously pinned cryptography to versions less than 3.4 since after that point, cryptography started using rust as a dependency. This isn't an issue if you install from wheel, but we found that the version of pip bundled with the python buildpack was too old to support this. However, since upgrading from python 3.6 to python 3.9, the pip version has been bumped and we now no longer need to pin cryptography as it installs correctly.

(i tested this by pushing an admin prototype to preview)

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -26,10 +26,6 @@ beautifulsoup4==4.9.3
 lxml==4.7.1
 Werkzeug==2.0.2
 
-# higher version causes build to fail on PaaS due to lack of Rust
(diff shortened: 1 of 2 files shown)


benthorner (approved):
Nice cleanup 🏅.

Thanks for explaining how you tested it.

benthorner:
I noticed we did the same thing in FTP as well, so I did a PR for that repo: https://github.com/alphagov/notifications-ftp/pull/362

PR #3435, opened 2022-01-21
Add index to improve page load times for service dashboard

idavidmcdonald:
We are seeing slow page load times for the service dashboards, ie
https://www.notifications.service.gov.uk/services/<service_id>

This appears to be caused by slow API request times for the
`get_template_statistics_for_service_by_day` route.

The route appears to be slow due to the query in
`fetch_notification_status_for_service_for_today_and_7_previous_days`
being slow.

That query unions the results from the `ft_notification_status` table
and the `notifications` table. It appears that the query on the
`notifications` table is the slow part:

```
SELECT
	CAST(notifications.notification_type AS TEXT) AS notifications_notification_type,
	notifications.notification_status AS notifications_notification_status,
	notifications.template_id AS notifications_template_id,
	count(*) AS count
FROM notifications
	WHERE
	notifications.created_at >= %(created_at_1)s
	AND notifications.service_id = %(service_id_2)s
	AND notifications.key_type != %(key_type_2)s
GROUP BY notifications.notification_type, notifications.template_id, notifications.notification_status
```

Doing an EXPLAIN ANALYZE on this will vary per service (due to
differing query plans due to different number of notifications per
service) but generally it will do an index scan on
`ix_notifications_service_created_at`. This would do it in two
parts. First it will search the index for all related rows and then
it will go and retrieve those rows. This might be millions of rows
that it needs to retreieve and if these rows are not in memory then
it may need to get these from disk which is particularly slow. We
see in production that often the first page load from 'cold' will
be over 30 seconds but then subsequent page load times shortly
after will be maybe a second or two.

We think one way to solve this is to introduce a new index such
that the index includes all the data in the index itself that this
query needs. An index like this would mean that when searching the
index for relevant rows, there would be no need for the subsequent
step to go and read all the rows (potentially some of them from disk).
This is known as an 'index only scan'
https://www.postgresql.org/docs/12/indexes-index-only-scans.html

The order of the keys in the index is important. We have included
them in the rough order of most differing values to fewest, ie
there are many different service_ids but very few key_types. This
will increase the likelyhood of our queries deciding to use the
new index.

Introducing this index manually and temporarily on staging and doing
some EXPLAIN ANALYZES shows that the slow query
- does an index only scan on our new index
- is dramatically faster (in the region of 90% faster)

Whilst this can't be guaranteed in production, because the data
in staging does not match production, we expect to see somewhat
similar results.

The downside to be aware of is that by adding a new index, everytime
we insert or update a row in the `notifications` table, it will have
to update this index, which takes time. However running a 5 minute
load test on staging at 100 notifications per second, we saw no
difference in response times to create notifications with and
without the new index.

To make this change in production, we must manually run an
`CREATE INDEX CONCURRENTLY` as:
- `CREATE INDEX` would incur downtime
- it is not possible to run `CREATE INDEX CONCURRENTLY`
  using SQLAlchemy

Finally, there may be other options that could solve this problem
but this solution is the proposed because we have good confidence
that it will work. Alternatives that have been considered

Partioning the notifications table:
May work but we aren't sure on what to partition and the amount
of work to figure that out is high and the amount of work to
transition is high (we can't partition the notifications table
itself, you have to create a new partioned table alongside and
mirror traffic)

pg_prewarm:
https://www.postgresql.org/docs/11/pgprewarm.html
This hypothetically could be helpful but we didn't investigate
as it is likely more complex than adding a new index

For further discussion and thoughts, see:
https://docs.google.com/document/d/1QZlYKTzFzeUkdDNm8Q_yiHme1edCpPPcA9ak3hvfnUk/edit#


app/models.py
```diff
@@ -1498,6 +1498,15 @@ class Notification(db.Model):
             'notification_type',
             'status',
             'created_at'
+        ),
+        Index(
+            "ix_notifications_service_created_at_template_composite",
```
benthorner, line 1503: Very optional: how about naming this index after the query? It's designed around that query and we don't expect it to be used for anything else. Richard was telling me that GOV.UK Pay have "an index for every query". We're not there yet, but this might be a helpful convention since we're adding more indices.

Possible name: `ix_notifications_template_statistics_for_service_by_day`. I think that's just short enough to be accepted.

idavidmcdonald: Excellent idea, love it. I struggled to name this due to the length constraint as I wanted to get every field in but that wasn't an option. This is a great alternative.

idavidmcdonald:
## Summary

So this was deployed about 5pm on the 24th of Jan in to production. In short, there is weak evidence that this may have made a small improvement. It is clear however that this has not made a monumental difference such that this new index will completely alleviate the problems our users are seeing.

## Percentile graphs for 3h time buckets

```
filter request like 'template-statistics'
| fields @timestamp, request, requestTime
| sort @timestamp desc
| stats pct(requestTime, 50), pct(requestTime, 90), pct(requestTime, 95), pct(requestTime, 99) by bin(3h)
```
on the `paas-api-nginx-access` group gives the following graphs for the last 4 weeks

image
image
image
image

## Percentile graphs for 1h time buckets

```
filter request like 'template-statistics'
| fields @timestamp, request, requestTime
| sort @timestamp desc
| stats pct(requestTime, 99), pct(requestTime, 99.5) by bin(1h)
```

on the `paas-api-nginx-access` group gives the following graphs for the last 4 weeks

image
image


## Request counts taking over 5 seconds for 1h time buckets

```
filter request like 'template-statistics'
| filter requestTime > 5
| fields @timestamp, request, requestTime
| sort @timestamp desc
| stats count() by bin(1h)
```

on the `paas-api-nginx-access` group gives the following graphs for the last 12 weeks

image


## Anecdotal evidence

Anecdotal evidence so far gathered is that
- I still can get pages to time out when attempting to load dashboards for some of our bigger services
- I have seen our slow use query plans that use an index only scan on the new index
- I have seen query plans that use this new index that take over 30 seconds

## Conclusion

It's clear from the above that this new index hasn't 'fixed' everything.

From the graphs there is debatably some weak evidence that it might be making a small positive difference. However, it is very hard to tell given a lot of variation in the percentile graphs before the change was released. Given this, I suggest we keep the index for another week or two to gather further data to prove or disprove that.

In the meantime, I will begin looking at table partitioning to see if that might be more successful.

We will need to wait longer to gain stronger evidence if this is making a positive difference.

idavidmcdonald:
## Why is the index only scan still slow?

So I think I know why the index hasn't made an improvement (whereas we did see a good improvement in staging and locally).

From the query plans above you can see

```
 HashAggregate  (cost=27134.89..27558.17 rows=28219 width=70) (actual time=71315.506..71315.583 rows=6 loops=1)
   Group Key: notification_type, template_id, notification_status
   Buffers: shared hit=601326 read=303783 dirtied=14431
   I/O Timings: read=69497.528
   ->  Index Only Scan using ix_notifications_template_statistics_for_service_by_day on notifications  (cost=0.69..26477.34 rows=65755 width=30) (actual time=1.517..71046.240 rows=607552 loops=1)
         Index Cond: ((service_id = '0cc696c6-b792-409d-99e9-64232f461b0f'::uuid) AND (created_at >= '2022-01-31 00:00:00'::timestamp without time zone))
         Filter: ((key_type)::text <> 'test'::text)
         Heap Fetches: 1230214
         Buffers: shared hit=601323 read=303783 dirtied=14431
         I/O Timings: read=69497.528
 Planning Time: 1.288 ms
 Execution Time: 71316.250 ms
 ```
 
 In particular:
 
 ```
 Heap Fetches: 1230214
 ```
 
A heap fetch is where it goes and retrieves a page of rows to read, which is exactly what we wanted to avoid by the index only scan (which just reads from the index directly rather than needing to go and fetch any pages). Heap fetches are expensive and take a lot of time.

So why is it doing a bunch of heap fetches?

In short, because the visibility map is out of date because there have been lots of updates to the table since the last VACUUM. 

https://blog.makandra.com/2018/11/investigating-slow-postgres-index-only-scans/
https://stackoverflow.com/questions/33009865/why-index-only-scan-is-taking-so-long
https://www.postgresql.org/docs/11/routine-vacuuming.html - the bit where it says it does VACUUMs 'To update the visibility map, which speeds up index-only scans.'

The solution for this is to run a VACUUM.

## But why are some query plans still quick?

Because of caching. If the index only scan requires lots of heap fetches, the first time that data is pulled from the heap it will be very slow. However on subsequent pulls, that data will be in the buffer cache and will be quick. 

## But what about autovacuum?

Doesn't autovacuum do all the vacuuming we need anyway? 

Looking at the error logs for our production DB for the 1-7 of Jan, I can see that the notifications table gets autovacuumed about 1 to 4 times a day. The time it takes to run the autovacuum varied between about 4 hours to 13 hours. When an autovacuum finishes, it doesn't automatically start autovacuuming the notifications table.

It may just be that our autovacuum isn't aggressive enough or isn't being run enough

### Our autovacuum settings

 autovacuum                                  | on                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 | Starts the autovacuum subprocess.
 autovacuum_analyze_scale_factor             | 0.05                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | Number of tuple inserts, updates, or deletes prior to analyze as a fraction of reltuples.
 autovacuum_analyze_threshold                | 50                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 | Minimum number of tuple inserts, updates, or deletes prior to analyze.
 autovacuum_freeze_max_age                   | 200000000                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          | Age at which to autovacuum a table to prevent transaction ID wraparound.
 autovacuum_max_workers                      | 3                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  | Sets the maximum number of simultaneously running autovacuum worker processes.
 autovacuum_multixact_freeze_max_age         | 400000000                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          | Multixact age at which to autovacuum a table to prevent multixact wraparound.
 autovacuum_naptime                          | 15s                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                | Time to sleep between autovacuum runs.
 autovacuum_vacuum_cost_delay                | 20ms                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | Vacuum cost delay in milliseconds, for autovacuum.
 autovacuum_vacuum_cost_limit                | 1200                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | Vacuum cost amount available before napping, for autovacuum.
 autovacuum_vacuum_scale_factor              | 0.1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                | Number of tuple updates or deletes prior to vacuum as a fraction of reltuples.
 autovacuum_vacuum_threshold                 | 50                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 | Minimum number of tuple updates or deletes prior to vacuum.
 autovacuum_work_mem                         | -1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 | Sets the maximum memory to be used by each autovacuum worker process.

## What would I like to test

I'd like to test what happens to if we run a VACUUM on the production notifications table and then see how our query plans perform. I think we should try running it during normal daytime running hours rather than at X in the morning. If it shows positive results, then we can look in to how we tweak autovacuum to be more aggresive (to mimic what vacuum does).

**Comment by risicle** [comment]
I think no matter what tricks you use, if you try calculating this big calculation with perfect accuracy and on the fly, you're always going to have problems.

If you try a trick like prewarming, the question is then - when are you going to prewarm? If you knew when users are about to make requests, you could just pre-calculate the aggregation for them. Continually prewarming for this query will do it at the expense of all other database load.

**Comment by benthorner** [comment]
Valiant effort @idavidmcdonald 🏅 

I agree with your conclusion: the index only scan has little to no benefit, since the time range of the query is focussed in on dirty rows. When I was doing my own testing on the new index, I see how I lost sight of this context.

Small request: the graphs in [this comment](https://github.com/alphagov/notifications-api/pull/3435#issuecomment-1024438810) could do with some titles to distinguish them. For example, I'm not sure why there are 4 3h graphs. Whatever they are, it doesn't change the conclusion, though.

> I wonder if it is because the vacuum took 1.5 hours so there were rows being dirtied all this time.

That makes sense: the weekly number of notifications for that service is consistent with a few 100K rows being dirty in the span of 1.5 hours.

It sounds like you're pursuing the use of Redis as an alternative solution - presumably to precompute the `for_today` part and use the query as a fallback. That does seem like the best option now 👍.

While this must have been a bit of a pain in the 🚪 to discover, at least we learnt something new 📚.

**Comment by benthorner** [comment]
Little update: this got me wondering if we would still see a benefit with the [the GIN index I'm thinking of adding](https://docs.google.com/document/d/1QZlYKTzFzeUkdDNm8Q_yiHme1edCpPPcA9ak3hvfnUk/edit#heading=h.3d1e0lch74k5). 

I think we will. Any use of an index will narrow the set of rows to retrieve, but we still need to: (a) check they are visible to the current transaction; (b) retrieve any data from them not contained in the index. (a) was a problem for this PR due to the very large number of volatile rows returned from the index; conversely, we only expect tiny numbers of (volatile) rows to be returned by the GIN index; we're expecting to access the heap for these as the index is only meant for filtering.

What confused me was: if indices are updated when a row is updated, then why do we need to check anything? Answer: the index doesn't store any visibility info, we need to check each row is visible to the current transaction. But then don't we need to check the whole table; what is the index achieving? No: the index returns a superset of the relevant rows across all active transactions, so the result set is precise and we can still benefit if the index significantly filters the number of rows.

**Comment by idavidmcdonald** [comment]
filter request like 'template-statistics'
| fields @timestamp, request, requestTime
| sort @timestamp desc
| stats pct(requestTime, 50), pct(requestTime, 90), pct(requestTime, 95), pct(requestTime, 99) by bin(3h)

on the paas-api-nginx-access group gives the following graphs for the last 16 weeks

50th percentile:
image

90th percentile:
image

95th percentile:
image

99th percentile:
image

So it has been a month and.... I'm still not 100% on whether this is useful.

I think there is a slight improvement after 24 Jan. 

I think there is little evidence that the 50th percentile time has improved.

I think there is some reasonable evidence that the 90th percentile has improved.

I think there is some evidence that the 95th percentile has improved.

I think there is little evidence that the 99th percentile has improved.

However it is hard to be very certain as all of this will depend on user traffic, how many views to the dashboard pages and how many notifications and how they are distributed across services in our DB.

I think I suggest we make it official and merge this PR. I'm confident it is not making things worse and somewhat confident that it is making a small improvement.

I suggest that after we take a different approach of using Redis for these slow queries, that we consider dropping this index. However I don't think that will be done in the next few weeks at least and therefore it is worth merging this code.

Could I get a second opinion please!

**Comment by idavidmcdonald** [comment]
Thanks Ben, in that case may I suggest I run the following to drop the index

```
DROP INDEX CONCURRENTLY ix_notifications_template_statistics_for_service_by_day;
```

I'm hoping that by dropping the index it may also give some more data for us now traffic is a bit lower to compare before and after with and without the index.

PR #3436, opened 2022-01-24
Stop killing reporting processes after each task

benthorner:

Previously we think this setting was necessary to avoid a memory
leak [1], but it's unclear if this is still an issue:

- We've advanced two major versions of Celery.
- Some of the tasks are now quicker and leaner.

Restarting worker sub-processes after each task is a big problem
for performance, as we move towards parallelising our reporting.

This is something of a test to see if we can manage without this
setting. Note that we need to unset the variable manually:

   cf unset-env notify-delivery-worker-reporting CELERYD_MAX_TASKS_PER_CHILD

In the worst case we can always re-run any failed tasks. To check
the worker is still behaving as expected, we can:

- Monitor CPU / memory graphs for it.
- Check `cf events` for unexpected restarts / crashes.
- Compare numbers of task completion logs to previous days.
- Check the number of new billing / status rows looks right.

[1]: https://github.com/alphagov/notifications-api/commit/ad419f7592524292f8da8447814cc0bbc37f240e

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -344,10 +344,6 @@ class Config(object):
     if os.getenv('CELERYD_PREFETCH_MULTIPLIER'):
         CELERY['worker_prefetch_multiplier'] = os.getenv('CELERYD_PREFETCH_MULTIPLIER')
 
-    # on reporting worker, restart workers after each task is executed to help prevent memory leaks
-    if os.getenv('CELERYD_MAX_TASKS_PER_CHILD'):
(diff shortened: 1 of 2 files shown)


leohemsted:
While we should call `cf unset-env`, we don't _need_ to do so immediately as we'll have removed the line that reads it in. As in, we should unset to practice good environment variable hygiene but I don't think we need to restage the app immediately after, that can just happen when we next deploy api

leohemsted (approved):

benthorner:
> we don't need to do so immediately as we'll have removed the code that reads it in

Good point @leohemsted 👍. I'll make a note and postpone removing it until tomorrow.

PR #3437, opened 2022-01-25
Attempt 2 of parallelising status aggregation more

benthorner:

This about these tasks:

    create_nightly_notification_status

    create_nightly_notification_status_for_day

We're still seeing occassional errors when one of these tasks is
taking too long to execute, causing duplication. This is likely
not a problem but there is potential for a race condition that
could result in missing rows in the status table.

Two things we can do to improve the situation:

- Stop the duplication in the first place (this PR).
- Make the task cope with duplication (future PR).

The original attempt [1] to fix the duplication and make the task
more scalable did work, but had a few issues:

1. The worker ran out of disk space. Fixed in [2].
2. It took too long to get through all the subtasks.

This PR reinstates the parallelisation but cuts the number of
subtasks that get created, so we avoid the second issue, which we
have also partially addressed in [3].

[1]: https://github.com/alphagov/notifications-api/pull/3425
[2]: https://github.com/alphagov/notifications-api/pull/3434
[3]: https://github.com/alphagov/notifications-api/pull/3436


app/dao/notifications_dao.py
```diff
@@ -806,3 +806,17 @@ def get_service_ids_that_have_notifications_from_before_timestamp(notification_t
...
+def get_service_ids_with_notifications_on_date(notification_type, date):
+    return {
+        row.service_id
+        for row in db.session.query(
+            Notification.service_id
+        ).filter(
+            Notification.notification_type == notification_type,
+            # using >= + < is much more efficient than date(created_at)
+            Notification.created_at >= date,
+            Notification.created_at < date + timedelta(days=1)
```
leohemsted, line 820: hmmm, alarm bells are ringing here. could this lead to weird bugs on the BST/GMT changeover dates? 

in fact now i see that we're already doing similar calculations in the outer loop just subtracting `days=1`.

I suspect that for example if you run this just after the late march switch, it'll keep the same time (11:00pm UTC) for each day, even if those days are before the DST switch and should be from 00:00. That could lead to a bit of double counting. 😬

I think (and this is very much something i've not tested) that we should do the calculation with `date` objects and then call `get_london_midnight_in_utc` on that date object each time we want to compare to a `created_at` timestamp

tests/app/celery/test_reporting_tasks.py
```diff
@@ -61,39 +61,58 @@ def test_create_nightly_billing_triggers_tasks_for_days(notify_api, mocker, day_
...
+@freeze_time('2019-08-01T00:30')
+@pytest.mark.parametrize('notification_date, expected_types_aggregated', [
+    ('2019-08-01', set()),
+    ('2019-07-31', {EMAIL_TYPE, SMS_TYPE, LETTER_TYPE}),
+    ('2019-07-28', {EMAIL_TYPE, SMS_TYPE, LETTER_TYPE}),
+    ('2019-07-27', {LETTER_TYPE}),
+    ('2019-07-22', {LETTER_TYPE}),
+    ('2019-07-21', set()),
+])
+def test_create_nightly_notification_status_triggers_relevant_tasks(
```
leohemsted, line 97: can we add a test where a service _isn't_ picked up when it has notifications on a different day?

it's probably easier to write that test forthe new `get_service_ids_with_notifications_on_date` function in the equivalent dao test file.

but i'm also interested to see us add some BST/GMT tests on this `create_nightly_notification_status` task where we ensure it sends the correct timestamps through to the `get_service_ids_with_notifications_on_date`

app/dao/notifications_dao.py
benthorner: Great spot @leohemsted !

It took me a little while to get my head around this, so I'll write some notes here with my understanding and then look at addressing your comment.

## What do we want?

- The day I think I'm running the task on is the day it actually runs on. This is always the case as the BST changeover is either at 1AM (forwards) or 2AM (backwards), so we'll always lock on to the correct day.

- The set of notifications for a day should match the timezone we were in when they were created e.g. if we think it was created late on Monday for us, then it should be attributed to Monday. This is the case for [the actual query](https://github.com/alphagov/notifications-api/blob/de3e0eeb954f6eacc33fc5d96683f30f94c96bea/app/dao/fact_notification_status_dao.py#L39) but, as you point out here, it's not the case for the pre-check to get the relevant service IDs.

This wouldn't lead to double counting but could lead to under counting ("no counting") if a service only sent some notifications between midnight and 1AM on a single, isolated day (once we move to BST):

- We look for 00:00 - 00:00 in the DB (UTC)
- But notifications really span from (in UTC):

   - 26/03 00:00 - 27/03 00:00 (for 26/03, GMT)
   - 27/03 00:00 - 27/03 23:00 (for 27/03, GMT/BST)
   - 27/03 23:00 - 28/03 23:00 (for 28/03, BST)

- So a notification sent at 28/03 00:30 (BST) - 27/03 23:30 (UTC) - would incorrectly trigger an aggregation for the associated service on the 27th, instead of the 28th. A bit of an edge case, but worth fixing 👍.

## How do we get it?

Do what you said ;-).

At the same time, we can simplify [the actual query](https://github.com/alphagov/notifications-api/blob/de3e0eeb954f6eacc33fc5d96683f30f94c96bea/app/dao/fact_notification_status_dao.py#L39) to use the same approach and put some tests around both functions to make the "what we want" more explicit (and correct).

tests/app/celery/test_reporting_tasks.py
benthorner: This is implicit in the first and last scenarios: a notification exists for a service, but nothing happens. I suppose there's a weird potential bug where we call the subtask for the full range of days if it has a notification on just one of them. That's covered by the previous test, though.

One thing I haven't done is check this works for multiple services when they _both_ have relevant notifications. I tend to avoid testing "is it a loop" though, on the basis that it's hard to write a non-looping bit of code that only runs once vs. one that can work for any number of items => a low value test.

Ramble over. Do you think there's still a missing test here?

app/dao/notifications_dao.py
benthorner: > At the same time, we can simplify [the actual query](https://github.com/alphagov/notifications-api/blob/de3e0eeb954f6eacc33fc5d96683f30f94c96bea/app/dao/fact_notification_status_dao.py#L39)

Addressed in https://github.com/alphagov/notifications-api/pull/3437/commits/f937d833b0aa3e0f00ac77603424e5570cd1657b

> Do what you said ;-).

Addressed in https://github.com/alphagov/notifications-api/pull/3437/commits/54163611e7f1f798b4dae99cfab0644d63c3763c.

tests/app/celery/test_reporting_tasks.py
leohemsted: i think what i was concerned about was that i'd spotted a bug, and we didn't have any tests that covered that. because your test  was only concerned with the outer python loops to ensure that the right notification types are invoked for the right day

get_service_ids_with_notifications_on_date now has tests (added in a 966c4db8c) which I think is what I was hoping for 👍

leohemsted (approved):
Nice work to clean up all the gnarly datetime calcs 👍

PR #3440, opened 2022-01-26
Audit api key id when cancelling broadcast via api

CrystalPea:
We want to audit API key both when we create a broadcast and when we cancel it.

Two ways we came up with to do this were to either create a new column on broadcast_message, or to add a row to "Events" table each time someone cancels a broadcast via API call. We decide to do the first of the two, because it is more consonant with how we have been doing stuff so far.

Needs to go in before: https://github.com/alphagov/notifications-api/pull/3448


app/models.py
```diff
@@ -2313,15 +2313,18 @@ class BroadcastMessage(db.Model):
...
+    created_by_api_key = db.relationship('ApiKey', foreign_keys=[created_by_api_key_id])
+    cancelled_by_api_key = db.relationship('ApiKey', foreign_keys=[cancelled_by_api_key_id])
 
     reference = db.Column(db.String(255), nullable=True)
     cap_event = db.Column(db.String(255), nullable=True)
 
     stubbed = db.Column(db.Boolean, nullable=False)
 
-    CheckConstraint("created_by_id is not null or api_key_id is not null")
+    CheckConstraint("created_by_id is not null or created_by_api_key_id is not null")
```
servingUpAces, line 2326: These check constraints are not in the migration script.

CrystalPea: You are right, I think the check constraint that they supersede is not in any migration script either. I added it now ☺️

migrations/versions/0363_cancelled_by_api_key.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = '0363_cancelled_by_api_key'
+down_revision = '0362_broadcast_msg_event'
+
+
+def upgrade():
```
servingUpAces, line 16: What are we doing with the data in the production broadcast_message table?
There currently are broadcast messages, unless we are purging that data we will need to do this migration in a different way. Let me know if you want my assistance with that migration.

servingUpAces (approved):

PR #3442, opened 2022-01-31
add script to run celery from within docker

leohemsted:
as a team we primarily develop locally. However, we've been experiencing issues with pycurl, a subdependency of celery, that is notoriously difficult to install on mac. On top of the existing issues, we're also seeing it conflict with pyproj in bizarre ways (where the order of imports between pyproj and pycurl result in different configurations of dynamically linked C libraries being loaded.

You are encouraged to attempt to install pycurl locally, following these instructions: https://github.com/alphagov/notifications-manuals/wiki/Getting-Started#pycurl

However, if you aren't having any luck, you can instead now run celery in a docker container.

`make run-celery-with-docker`

This will build a container, install the dependencies, and run celery (with the default of four concurrent workers).

It will pull aws variables from your aws configuration as boto would normally, and it will attempt to connect to your local database with the user `postgres`. If your local database is configured differently (for example, with a different user, or on a different port), then you can set the SQLALCHEMY_DATABASE_URI locally to override that.

nb: most of this code mimics, at least in part, the work done in template preview: https://github.com/alphagov/notifications-template-preview/pull/517. I removed a few steps and boilerplate since we only ever anticipate using this to run celery locally (ie: no test vs production image, no install test reqs, no need to have commands to run bash, etc)

Known limitations:

This won't work with modifying utils locally and installing it in `-e` mode. Further work would be required to copy that folder into the docker file if you wish to do that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,3 +177,8 @@ disable-failwhale: ## Disable the failwhale app and enable api
 	cf unmap-route notify-api-failwhale ${DNS_NAME} --hostname api
 	cf stop notify-api-failwhale
 	@echo "Failwhale is disabled"
+
+.PHONY: run-celery-with-docker
+run-celery-with-docker: ## Run celery in Docker container (useful if you can't install pycurl locally)
+	docker build -f docker/Dockerfile -t notifications-api .
+	./scripts/run_with_docker.sh make run-celery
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -177,3 +177,11 @@ disable-failwhale: ## Disable the failwhale app and enable api
 	cf unmap-route notify-api-failwhale ${DNS_NAME} --hostname api
 	cf stop notify-api-failwhale
 	@echo "Failwhale is disabled"
+
+.PHONY: bootstrap-with-docker
```
servingUpAces, line 181: why do we have this make task if it doesn't do anything?

sakisv (approved):
Looks good 👍

PR #3448, opened 2022-02-08
Drop api_key_id column from broadcast_message table

CrystalPea:
This column has been superseded by a new column named created_by_api_key_id.

Needs to go in after https://github.com/alphagov/notifications-api/pull/3440

### DIFF
--- a/migrations/versions/0364_drop_old_column.py
+++ b/migrations/versions/0364_drop_old_column.py
@@ -1,0 +1,33 @@
+"""
+
+Revision ID: 0364_drop_old_column
+Revises: 0363_cancelled_by_api_key
+Create Date: 2022-01-25 18:05:27.750234
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = '0364_drop_old_column'
+down_revision = '0363_cancelled_by_api_key'
+
+
+def upgrade():
+    # move data over
+    op.execute("UPDATE broadcast_message SET created_by_api_key_id=api_key_id WHERE created_by_api_key_id IS NULL")
+    op.create_check_constraint(
+        "ck_broadcast_message_created_by_not_null",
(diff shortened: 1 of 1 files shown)


servingUpAces (approved):

migrations/versions/0364_drop_old_column.py
```diff
@@ -0,0 +1,35 @@
...
+        "broadcast_message",
+        "created_by_id is not null or created_by_api_key_id is not null"
+    )
+    op.drop_column('broadcast_message', 'api_key_id')
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('broadcast_message', sa.Column('api_key_id', postgresql.UUID(), autoincrement=False, nullable=True))
+    op.get_bind()
```
benthorner, line 30: Do we need this? I can see it in other migrations, but I'm not sure what it's for.

benthorner (commented):
Just a couple of things to check / fix before this goes in. Initially I was surprised to see a constraint being created, but I can see how it makes sense at this point: we can't start enforcing it until the app is using the new column 👍 .

migrations/versions/0364_drop_old_column.py
leohemsted: https://alembic.sqlalchemy.org/en/latest/ops.html?#:~:text=For%20full%20interaction%20with%20a%20connected%20database%20where%20parameters%20can%20also%20be%20used%20normally%2C%20use%20the%20%E2%80%9Cbind%E2%80%9D%20available%20from%20the%20context%3A

looks like it;s not needed in this case, as that fn just returns something.

some alembic scripts do use the return value tho for passing through to other functions or running `.execute` on though.

CrystalPea: Ahh, good spot, I just copied from earlier migrations, I will get rid of it ☺️

PR #3452, opened 2022-02-09
Audit api key id when cancelling broadcast via api - second try

CrystalPea:
Needs to go in after https://github.com/alphagov/notifications-api/pull/3453

And needs to go in before:  https://github.com/alphagov/notifications-api/pull/3448

### DIFF
--- a/tests/app/broadcast_message/test_utils.py
+++ b/tests/app/broadcast_message/test_utils.py
@@ -9,7 +9,12 @@ from app.models import (
     BroadcastEventMessageType,
     BroadcastStatusType,
 )
-from tests.app.db import create_broadcast_message, create_template, create_user
+from tests.app.db import (
+    create_api_key,
+    create_broadcast_message,
+    create_template,
+    create_user,
+)
 
 
 def test_validate_and_update_broadcast_message_status_stores_approved_by_and_approved_at_and_queues_task(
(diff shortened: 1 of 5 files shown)


tests/app/broadcast_message/test_utils.py
```diff
@@ -87,12 +89,73 @@ def test_validate_and_update_broadcast_message_status_for_cancelling_broadcast(
...
+    assert broadcast_message.status == BroadcastStatusType.REJECTED
+    assert broadcast_message.cancelled_at is None
+    assert broadcast_message.cancelled_by_id is None
+    assert broadcast_message.updated_at is not None
+
+    assert not mock_task.called
+    assert len(broadcast_message.events) == 0
+
+
+def test_validate_and_update_broadcast_message_status_for_rejecting_broadcast(
```
benthorner, line 154:
```suggestion
def test_validate_and_update_broadcast_message_status_for_rejecting_broadcast_from_API_call(
```

benthorner (approved):
Looks good. I agree that, since these values are only set once, they should be on the model and not treated as events. Maybe we'll have `rejected_by` in future as well 😉.

Just one test I think we're missing: we should check the new `created_by_api_key_id` is populated correctly somewhere e.g. [in this test](https://github.com/alphagov/notifications-api/blob/e5f085fe11dfcdcf19f3d8173477fd947ce9e5b1/tests/app/v2/broadcast/test_post_broadcast.py#L71).

tests/app/broadcast_message/test_utils.py
```diff
@@ -87,12 +89,73 @@ def test_validate_and_update_broadcast_message_status_for_cancelling_broadcast(
...
+    mock_task = mocker.patch('app.celery.broadcast_message_tasks.send_broadcast_event.apply_async')
+
+    validate_and_update_broadcast_message_status(
+        broadcast_message, BroadcastStatusType.CANCELLED, updating_user=None, api_key_id=api_key.id
+    )
+
+    assert broadcast_message.status == BroadcastStatusType.CANCELLED
+    assert broadcast_message.cancelled_at is not None
+    assert broadcast_message.cancelled_by_id is None
+    assert broadcast_message.cancelled_by_api_key_id == api_key.id
```
CrystalPea, line 115: @benthorner thanks for reviewing 🙏🏼 . I think we are testing that `cancelled_by_api_key_id` gets populated correctly here ☺️

benthorner: Indeed we are. I was talking about `**created**_by_api_key_id`. The old `api_key_id` column wasn't being checked before, though, so it's not a new issue. However, since we're making a change to the column, this PR is a good place to make sure there's a test to cover it somewhere.

Does that make sense?

PR #3453, opened 2022-02-09
Migration adding cancelled_by_api_key_id to broadcast_message

CrystalPea:
table. Also add created_by_api_key_id that will supersede api_key_id column in that same table.

We do migration and code in separate PRs to make it easier to downgrade if anything goes wrong.

Needs to go in before:
https://github.com/alphagov/notifications-api/pull/3452
AND
https://github.com/alphagov/notifications-api/pull/3448


migrations/versions/0363_cancelled_by_api_key.py
```diff
@@ -0,0 +1,56 @@
...
+        ['cancelled_by_api_key_id'],
+        ['id']
+    )
+    op.get_bind()
+    op.execute("UPDATE broadcast_message SET created_by_api_key_id=api_key_id")  # move data over
+    op.create_check_constraint(
+        "ck_broadcast_message_created_by_not_null",
+        "broadcast_message",
+        "created_by_id is not null or created_by_api_key_id is not null"
+    )
```
leohemsted, line 41: this constraint cant be created until after we've updated all the code to create with the new way.

or between this and #3452 we'll create messages that will fail this check

CrystalPea: ahh you are right, I will move it, and copying of the data between columns, to the later migration.

servingUpAces:
The easiest way to check the db migration is backward compatible is to apply the db migration before updating the code and run the unit tests (assuming unit tests cover insert/updates to broadcast messages)

migrations/versions/0363_cancelled_by_api_key.py
CrystalPea: ok done now, I moved both data copying and create constraint to the later migration ☺️

migrations/versions/0363_cancelled_by_api_key.py
```diff
@@ -0,0 +1,45 @@
...
+down_revision = '0362_broadcast_msg_event'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('broadcast_message', sa.Column('created_by_api_key_id', postgresql.UUID(as_uuid=True), nullable=True))
+    op.add_column(
+        'broadcast_message', sa.Column('cancelled_by_api_key_id', postgresql.UUID(as_uuid=True), nullable=True)
+    )
+    op.drop_constraint('broadcast_message_api_key_id_fkey', 'broadcast_message', type_='foreignkey')
```
leohemsted, line 20: i'm not too worried about this but this doesnt need to happen really, i think. null entries still comply with foreign key constraints, and when we drop the column later that should also remove the associated constraints with it automatically

leohemsted (approved):

PR #3454, opened 2022-02-10
Rewrite status aggregation to be a bulk upsert

benthorner:

This significantly cuts the number of (insert) queries per service, and
also removes a minor issue of rows in the status table being deleted but
not replaced if the task quits halfway through (for whatever reason).

I think the new version is easier to follow as all the behaviour is in
one place, instead of passing "new_status_rows" around. It was also a bit
of a learning opportunity for me to get used to SQLAlchemy.

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -10,10 +10,7 @@ from app.dao.fact_billing_dao import (
     fetch_billing_data_for_day,
     update_fact_billing,
 )
-from app.dao.fact_notification_status_dao import (
-    fetch_status_data_for_service_and_day,
-    update_fact_notification_status,
-)
+from app.dao.fact_notification_status_dao import update_fact_notification_status
 from app.dao.notifications_dao import get_service_ids_with_notifications_on_date
 from app.models import EMAIL_TYPE, LETTER_TYPE, SMS_TYPE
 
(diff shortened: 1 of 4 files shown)


leohemsted (changes requested):
We delete the old stats rows because some of those rows will have changed. An upsert will correct numbers being different, but not if the row should no longer be there at all! This is super common - as in notifications that are sending but get a delivery status the next day, at some point we expect to have no notifications left in sending for that template/day so that row should have been deleted entirely.

There's a test (that has always been) missing in test_reporting_tasks. The following passes on master, and fails on this branch as after the task runs a second time there are two rows, when there should only be one

```
def test_create_nightly_notification_status_for_day_removes_old_data(sample_template):
    process_day = date.today()
    notification = create_notification(template=sample_template, status='sending')

    create_nightly_notification_status_for_day(str(process_day), 'sms')

    new_fact_data = FactNotificationStatus.query.all()

    assert len(new_fact_data) == 1
    assert new_fact_data[0].notification_count == 1
    assert new_fact_data[0].notification_status == 'sending'

    # notification gets a delivery receipt
    notification.status = 'delivered'

    create_nightly_notification_status_for_day(str(process_day), 'sms')

    # we should no longer have any rows for sending
    updated_fact_data = FactNotificationStatus.query.all()

    assert len(updated_fact_data) == 1
    assert updated_fact_data[0].notification_count == 1
    assert updated_fact_data[0].notification_status == 'delivered'
```

benthorner:
Argh damn. Thanks for spotting though! Seems obvious in retrospect 🤷‍♂️.

I can definitely add that missing test 💯.

Do you think it's worth continuing with this PR? I can add the deletion back in place of the conflict strategy. Overall it would still cut down on the number of queries and arguably be a bit simpler.

leohemsted:
I think it's probs still worthwhile doing the PR - it looks sensible enough (i agree it's more readable, and i'd also expect some kind of performance boost too) and re-adding the delete at the beginning won't change it that much. We shouldn't need the on conflict any more tho.

benthorner:
Cool, I've made those changes in https://github.com/alphagov/notifications-api/pull/3454/commits/fd4737ebf61471cf6ab4f3379d879304666bd315 - ready for another review.

leohemsted (approved):
looks good. any specific benchmarking we want to do before merging?

benthorner:
> any specific benchmarking we want to do before merging?

I had a go at this but it's quite hard to do locally due to the amount of test data needed. I suspect the performance change is marginal and may only be visible over lots of services.

> This PR complements the work I'm doing to parallelise status aggregation more [1] - both should help to speed up this task and therefore avoid the errors we currently get due to tasks getting duplicated.

This is sort of true: if it turned out we couldn't parallelise more then we may have seen an improvement from this. Now that we are parallelising more I think it's misleading (removed).

So at this point it's just a refactoring, which we may as well merge.

(And also a learning opportunity for me, which is important too 😄.)

PR #3458, opened 2022-02-14
More tweaks of broadcast message validation

quis:
## Test that alert with no references is OK

References are optional, and we fixed errors when they are not provided in bbc444699a6032ef855959d81aee905775be9862

This commit amends the tests so that they cover the `Alert` message type as well as the `Cancel` (which is already covered).

## Allow cancel of alert via API with empty `<description>`

The XML for an alerts requires a `<description>` field. The XML for a `<cancel>` may have the `<description>` field populated (although we ignore the contents) but it may also be empty.

This commit updates the schema to leave the all the validation to the view layer, which can decide when or when not to validate the content of the `<description>` field.

### DIFF
--- a/app/v2/broadcast/broadcast_schemas.py
+++ b/app/v2/broadcast/broadcast_schemas.py
@@ -52,7 +52,7 @@ post_broadcast_schema = {
         },
         "content": {
             "type": "string",
-            "minLength": 1,
+            "minLength": 0,
         },
         "web": {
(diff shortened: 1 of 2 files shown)


rjbaker (approved):
Looks good. I can see we already have coverage to ensure a `Cancel` message contains a reference (otherwise we wouldn't know what to cancel).

PR #3461, opened 2022-02-17
Be more robust in handling ambiguous references to cancel an alert

quis:
We’ve found two problems that can come up when trying to cancel an alert:
- the `<references>` field contains reference to the alert which should be cancelled, plus some older, already-cancelled alerts
- the `<references>` field contains references to multiple current or pending alerts, and it’s unclear which one should be cancelled

This pull request adds code to handle the first case and error in the second.

### DIFF
--- a/app/dao/broadcast_message_dao.py
+++ b/app/dao/broadcast_message_dao.py
@@ -26,6 +26,10 @@ def dao_get_broadcast_message_by_id_and_service_id(broadcast_message_id, service
 
 def dao_get_broadcast_message_by_references_and_service_id(references_to_original_broadcast, service_id):
     return BroadcastMessage.query.filter(
+        BroadcastMessage.status.in_((
+            BroadcastStatusType.PENDING_APPROVAL,
+            BroadcastStatusType.BROADCASTING,
+        )),
         BroadcastMessage.reference.in_(references_to_original_broadcast),
         BroadcastMessage.service_id == service_id
     ).one()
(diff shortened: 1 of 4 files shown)


rjbaker:
From the CAP spec: 

> “Cancel” - Cancels the earlier message(s) identified in <references>

The reference fields behave like a linked list, usually there's original `Alert` followed by one or more `Update`s and a final `Cancel`. Usually an entirely new separate `Alert` will be created following a `Cancel` as it implies ALL referenced messages should be disregarded.

quis:
@rjbaker we are letting people ‘reject’ (using a cancel message) alerts which are in the ‘pending approval‘ state. This is a state which isn’t in CAP. Therefore there’s more potential for ambiguity about which alert should be cancelled. The scenario we thought of is:
- someone issues an alert in the Flood Warning System
- they spot a spelling mistake
- they issue another alert in the Flood Warning System, to fix the spelling mistake 
- they try to cancel the original alert in the Flood Warning System
- the cancel now references the original alert and the corrected alert (this is how references work in the Flood Warning System)
- if we treat this a valid request we would reject the alert with the spelling mistake (good) but also the one with the correction (bad)

Operational procedures should stop this situation from arising, but if it does our preference is to have an error in the case of ambiguity, rather than risk cancelling something which should be going out. That’s what I’ve discussed with the Environment Agency.

leohemsted:
if they end up in this state what would their course of action be - log into notify and manually cancel the one with the typo?

quis:
@leohemsted yeah, that’s right.

rjbaker (approved):

PR #3462, opened 2022-02-17
Migration to make NHS email branding ID consistent across environments

klssmith:
With the changes we will be making to email branding in admin it's useful to know the ID of the NHS email brand. This makes it easier to show a preview of the branding.

This means that the ID will need to be consistent across environments, so this changes the details of the NHS brand in the dev, preview and staging environments to match the production data.

The migration
- Removes NHS branding from existing services
- Removes NHS branding from any orgs which had it as their default
- Deletes the NHS branding row if it exists
- Inserts a new NHS branding row with details which match those in production

This does have the effect of removing NHS branding from people's local and preview services and orgs (NHS branding doesn't exist on staging), but this does not affect real teams. The NHS logo with a filename of `1ac6f483-3105-4c9e-9017-dd7fb2752c44-nhs-blue_x2.png` exists in the logo S3 bucket for all environments.

[Pivotal story](https://www.pivotaltracker.com/story/show/180807168)

### DIFF
--- a/migrations/versions/0365_add_nhs_branding.py
+++ b/migrations/versions/0365_add_nhs_branding.py
@@ -1,0 +1,58 @@
+"""
+
+Revision ID: 0365_add_nhs_branding
+Revises: 0364_drop_old_column
+Create Date: 2022-02-17 16:31:21.415065
+
+"""
+import os
+
+from alembic import op
+
+revision = '0365_add_nhs_branding'
+down_revision = '0364_drop_old_column'
+
+environment = os.environ['NOTIFY_ENVIRONMENT']
+
+
+def upgrade():
+    if environment not in ["live", "production"]:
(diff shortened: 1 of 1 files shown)


benthorner (approved):
Nice! I like how this provides a clear audit trail of where this top-level branding was created from 💯. We could just update the existing IDs, but this is clearer and I agree the impact is minimal - the default branding should still be NHS for most services anyway.

klssmith:
@benthorner I've added one more step to the migration that I'd missed (the `UPDATE organisation` bit). As well as removing any current NHS branding from services, we also need to remove it from orgs in order to avoid a foreign key constraint error when deleting the NHS branding row. Can you check you're still happy please?

benthorner (approved):
Good spot 👍.

PR #3463, opened 2022-02-22
Clear cache when cancelling broadcast via the API

quis:
Before we implemented ‘cancel’ any updates to a broadcast went through the admin app. This meant the admin app could deal with clearing the cache any time a broadcast was updated by a user performing an action.

Now that a broadcast can be updated without the admin app being involved we have another place we need to clear the cache from.

If we don’t do this then the broadcast can look like it’s still going even though it’s successfully been cancelled.


benthorner (approved):
Good spot 👍.

This matches the key format used by Admin ✅.

PR #3466, opened 2022-02-23
Bump gunicorn

leohemsted:
Gunicorn has a bug where it is incompatible with eventlet>0.30.2

The code to fix this itself was merged into master in gunicorn's repo (https://github.com/benoitc/gunicorn/pull/2581) - however, that master code hasn't been released since it was merged 9 months ago.

There's a [security advisory for eventlet<=0.31.0](https://github.com/eventlet/eventlet/security/advisories/GHSA-9p9m-jm8w-94p2) which doesn't affect us as it only involves websockets, however, it'd be nice to be able to not have that as a lingering issue on our repo.

To fix this, pin github to the merge commit for the PR above. This can revert to a regular pypi requirement for gunicorn once 0.21.1 or higher is released.

The pinned commit adds nothing to gunicorn 20.1.0 other than the eventlet fix and some docs changes.

also note, i've changed the requirement to `gunicorn[eventlet]` - this means we don't need to explicitly list eventlet in reqs.in.

also note locally that you may need to run `pip install -r requirements.txt --no-cache-dir` to ensure you get the version from github (since the version # hasn't changed, pip will try and use the old wheel in the cache dir). That said we don't normally run under gunicorn/eventlet locally

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -9,8 +9,8 @@ Flask-Migrate==2.7.0
 git+https://github.com/mitsuhiko/flask-sqlalchemy.git@500e732dd1b975a56ab06a46bd1a20a21e682262#egg=Flask-SQLAlchemy==2.3.2.dev20190108
 Flask==1.1.2
 click-datetime==0.2
-eventlet==0.30.2 # pyup: ignore # 0.31 breaks Gunicorn
-gunicorn==20.1.0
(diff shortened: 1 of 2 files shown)


CrystalPea (approved):
I'm ok with this 👍🏼 .

Although as some others pointed in Slack, we should have some process for reviewing pinned dependencies like that 🤔

PR #3467, opened 2022-02-24
Iterate local development with Docker

benthorner:
This makes a few changes to:

- Make local development consistent with our other apps. It's now
faster to start Celery locally since we don't try to build the
image each time - this is usually quick, but unnecessary.

- Add support for connecting to a local Redis instance. Note that
the previous suggestion of "REDIS = True" was incorrect as this
would be turned into the literal string "True".

I've also co-located and extended the recipes in the Makefile to
make them a bit more visible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,13 +61,11 @@ export PATH=${PATH}:/Applications/Postgres.app/Contents/Versions/11/bin/
 
 ### Redis
 
-To switch redis on you'll need to install it locally. On a OSX we've used brew for this. To use redis caching you need to switch it on by changing the config for development:
+To switch redis on you'll need to install it locally e.g. `brew install redis`. To use redis caching you need to switch it on with an environment variable:
 
 ```
-REDIS_ENABLED = 1
+export REDIS_ENABLED=1
 ```
-
-Note that if you're running Celery with Docker locally, then you'll also need to set `REDIS_URL=redis://host.docker.internal:6379`.
 
 ##  To run the application
 
--- a/README.md
+++ b/README.md
@@ -63,8 +63,11 @@ export PATH=${PATH}:/Applications/Postgres.app/Contents/Versions/11/bin/
 
 To switch redis on you'll need to install it locally. On a OSX we've used brew for this. To use redis caching you need to switch it on by changing the config for development:
 
-        REDIS_ENABLED = True
+```
+REDIS_ENABLED = 1
+```
 
+Note that if you're running Celery with Docker locally, then you'll also need to set `REDIS_URL=redis://host.docker.internal:6379`.
 
 ##  To run the application
 
@@ -81,6 +84,18 @@ make run-celery
 # run scheduled tasks (optional)
 make run-celery-beat
 ```
+
+We've had problems running Celery locally due to one of its dependencies: pycurl. Due to the complexity of the issue, we also support running Celery via Docker:
+
+```
+# install dependencies, etc.
+make bootstrap-with-docker
+
+# run the background tasks
+make run-celery-with-docker
+
+# run scheduled tasks
+make run-celery-beat-with-docker
 
 ##  To test the application
 
--- a/README.md
+++ b/README.md
@@ -61,7 +61,15 @@ export PATH=${PATH}:/Applications/Postgres.app/Contents/Versions/11/bin/
 
 ### Redis
 
-To switch redis on you'll need to install it locally e.g. `brew install redis`. To use redis caching you need to switch it on with an environment variable:
+To switch redis on you'll need to install it locally. On a Mac you can do:
+
+```
+# assuming you use Homebrew
+brew install redis
+brew services start redis
+```
+
+To use redis caching you need to switch it on with an environment variable:
 
 ```
 export REDIS_ENABLED=1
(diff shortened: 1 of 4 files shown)


idavidmcdonald:
I might look at this properly tomorrow but just before I finish, check out this commit where I did similar, just to compare/contrast in case you missed/liked anything

https://github.com/alphagov/notifications-api/pull/3450/commits/d567d1529db7778b73205f73f9b5ec1a94b66461

benthorner:
Dah, I knew this was familiar from somewhere! Sorry for duplicating.

I think we're pretty much aligned and this PR goes a bit further in terms of support provided.

Are you OK if we continue with this one?

README.md
```diff
@@ -63,8 +63,11 @@ export PATH=${PATH}:/Applications/Postgres.app/Contents/Versions/11/bin/
 
 To switch redis on you'll need to install it locally. On a OSX we've used brew for this. To use redis caching you need to switch it on by changing the config for development:
 
-        REDIS_ENABLED = True
+```
```
idavidmcdonald, line 66: Optional: For the paragraph above this, I'd suggest the following change for how to run redis locally:

You can run redis locally using

```
brew install redis
redis-server
```

I feel this is more explicit and easy to follow.

README.md
```diff
@@ -63,8 +63,11 @@ export PATH=${PATH}:/Applications/Postgres.app/Contents/Versions/11/bin/
...
+REDIS_ENABLED = 1
+```
 
+Note that if you're running Celery with Docker locally, then you'll also need to set `REDIS_URL=redis://host.docker.internal:6379`.
```
idavidmcdonald, line 70: Optional: Can we set this environment variable in the dockerfile as a default to avoid people having to do this manually themselves so that it 'just works'?

Like at the bottom of https://github.com/alphagov/notifications-api/pull/3450/commits/d567d1529db7778b73205f73f9b5ec1a94b66461

README.md
```diff
@@ -63,8 +63,11 @@ export PATH=${PATH}:/Applications/Postgres.app/Contents/Versions/11/bin/
 
 To switch redis on you'll need to install it locally. On a OSX we've used brew for this. To use redis caching you need to switch it on by changing the config for development:
 
-        REDIS_ENABLED = True
+```
+REDIS_ENABLED = 1
```
idavidmcdonald, line 67: I'm confused. Is the suggestion here to do

`REDIS_ENABLED=1 make run-celery-with-docker`

or is it to go and change the app/config.py file and set `REDIS_ENABLED = 1`? 

If it is the former, then we would need to remove the space characters here and change the description of the comment above which talks about changing the development config which I read as related to app/config.py.

Or if it is the later, then I'm not quite sure why we need to change it to `1` as I feel that `True` would also turn redis on when editing the app/config.py directly?

scripts/run_with_docker.sh
```diff
@@ -14,6 +14,8 @@ docker run -it --rm \
   -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
   -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
   -e SQLALCHEMY_DATABASE_URI=$SQLALCHEMY_DATABASE_URI \
+  -e REDIS_ENABLED=${REDIS_ENABLED:-0} \
```
idavidmcdonald, line 17: I don't think we would need to do this if we are editing the app/config.py file directly (see my comment above)? But this may not be the case if we are using an environment variable instead to control whether we use redis or not.

idavidmcdonald (commented):
Generally good but one small clarification on how to do run this locally please

README.md
benthorner: Interesting. Happy to add `brew install redis` but I'd prefer to leave the reader to adopt their own approach e.g. brew prints out the following commands, which is what I use myself:

```
brew services start redis
brew services stop redis
```

OK if I just mention how to install it?

benthorner: Updated in https://github.com/alphagov/notifications-api/pull/3467/commits/6092baea643e5058237fd0d48e06f9d8f5918a65.

README.md
benthorner: Good point - I missed that in your one.

Amended in https://github.com/alphagov/notifications-api/pull/3467/commits/6092baea643e5058237fd0d48e06f9d8f5918a65.

README.md
benthorner: Ahhh, so _that's_ what it was referring to.

You're right about the spaces. I've amended in https://github.com/alphagov/notifications-api/pull/3467/commits/6092baea643e5058237fd0d48e06f9d8f5918a65.

scripts/run_with_docker.sh
benthorner: Indeed, we still need this as I'm not proposing someone have to go and edit a file. However, unless the environment variable is set you'll get an error, hence the default of "0" here.

README.md
idavidmcdonald: Optional: I'd prefer we include explicit instructions in here for starting redis too. I would like our READMEs to be explicit where possible, and I think including a command for how to start redis would be useful. I don't mind whether that is your brew commands there or `redis-server` but it prevents either
- someone not realising they need to start a redis server
- having to immediately google, how to start redis

In particular for someone never having used redis before or someone not familiar with brew this may be useful.

So more of a paved path, very happy for people to deviate with their own preferences if they wish but ideally 90% of the time they should be able to follow the readme without needing to turn to google.

idavidmcdonald (approved):
Thanks Ben, looking good.

I left one final comment in a thread that I won't die on a hill for so feel free to decide as you see fit and then get this merged!

README.md
benthorner: Sure, I've added this in https://github.com/alphagov/notifications-api/pull/3467/commits/bd690ab71825b227b01d9fc3aa347dff146e1a5a.

An installation with Homebrew will recommend to use `services start`, so that's what I've gone with.

PR #3469, opened 2022-02-25
Allow SMS fragment limit to be 0

klssmith:
This updates the `create_or_update_free_sms_fragment_limit_schema` to allow the `free_sms_fragment_limit` to have a minimum of 0, not 1.

[Pivotal story](https://www.pivotaltracker.com/story/show/181282523)


benthorner (approved):
Works for me locally in combination with https://github.com/alphagov/notifications-admin/pull/4168.

I had a quick look through the code to see if a value of 0 would cause any issues. I didn't see anything that looked brittle though, so I think this is good to go.

PR #3470, opened 2022-02-28
Make letters still sending check later

klssmith:
This changes the time of the scheduled task to raise an alert if letters are still sending from 1530 to 1700. DVLA have reported that our "monitoring is executing just before we actually mark them as ‘despatched’ and send you the feedback files." and asked us to make the check a little later.

We don't actually contact DVLA until the morning after the alert anyway, so this won't affect the process of getting in touch with them.

This change will require Cronitor to be updated for the new time.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -307,7 +307,7 @@ class Config(object):
             },
             'raise-alert-if-letter-notifications-still-sending': {
                 'task': 'raise-alert-if-letter-notifications-still-sending',
-                'schedule': crontab(hour=15, minute=30),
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):

leohemsted:
cronitor doesnt know the time things run, it just expects them to run within 24 hours.

so it'll fail after we merge this, go green two hours later when the task runs for real, then the next day it'll only check 24 hrs after the last run, so at 5pm

klssmith:
Thanks @leohemsted good to know there's nothing to do change for Cronitor

PR #3471, opened 2022-02-28
Fix test which was flakey due to order of items returned from db

klssmith:
The test was querying `FactNotificationStatus` and ordering the results
by bst_date and notification_type then checking the rows. However, the
bst_date and notification_type for each row is the same, so this test
could fail based on the order that the results came back in. By ordering
on the notification_status instead, we can be sure of the order of the
results.

### DIFF
--- a/tests/app/celery/test_reporting_tasks.py
+++ b/tests/app/celery/test_reporting_tasks.py
@@ -605,10 +605,7 @@ def test_create_nightly_notification_status_for_service_and_day_overwrites_old_d
     notification = create_notification(template=first_template, status='sending')
     create_nightly_notification_status_for_service_and_day(str(process_day), first_service.id, 'sms')
 
-    new_fact_data = FactNotificationStatus.query.order_by(
-        FactNotificationStatus.bst_date,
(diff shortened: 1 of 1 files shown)


leohemsted (approved):

tests/app/celery/test_reporting_tasks.py
```diff
@@ -605,10 +605,7 @@ def test_create_nightly_notification_status_for_service_and_day_overwrites_old_d
     notification = create_notification(template=first_template, status='sending')
     create_nightly_notification_status_for_service_and_day(str(process_day), first_service.id, 'sms')
 
-    new_fact_data = FactNotificationStatus.query.order_by(
-        FactNotificationStatus.bst_date,
-        FactNotificationStatus.notification_type
-    ).all()
+    new_fact_data = FactNotificationStatus.query.all()
```
idavidmcdonald, line 608: Just to check, does this need an order by too?

idavidmcdonald: And I immediately see that it doesn't a few lines below! Ignore me!

PR #3479, opened 2022-03-07
automatically retry letters stuck in pending-virus-scan

leohemsted:
Since sept 2019 we've had to log on to production around once every twenty days to restart the virus scan task for a letter. Most of the time this is just a case of making sure the file is in the scan bucket, and then triggering the task. If the file isn't in the scan bucket we'd need to do some more manual investigation to find out exactly where the file got stuck, but I can only remember times when it's been in the scan bucket.

So if the file is in the scan bucket, we can just check that with code and kick the task off automatically.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -236,10 +236,9 @@ def check_if_letters_still_pending_virus_check():
         letter_ids = [(str(letter.id), letter.reference) for letter in letters]
 
         msg = f"""{len(letters)} precompiled letters have been pending-virus-check for over 90 minutes.
-            We couldn't find them in the scan bucket.
+            We couldn't find them in the scan bucket. We'll need to find out where the files are and kick them off
+            again or move them to technical failure.
 
-            Follow runbook to resolve:
-            https://github.com/alphagov/notifications-manuals/wiki/Support-Runbook#Deal-with-letter-pending-virus-scan-for-90-minutes.
             Notifications: {sorted(letter_ids)}"""
 
         if current_app.config['NOTIFY_ENVIRONMENT'] in ['live', 'production', 'test']:
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -8,6 +8,7 @@ from sqlalchemy import between
 from sqlalchemy.exc import SQLAlchemyError
 
 from app import db, notify_celery, zendesk_client
+from app.aws import s3
 from app.celery.broadcast_message_tasks import trigger_link_test
 from app.celery.letters_pdf_tasks import get_pdf_for_templated_letter
 from app.celery.tasks import (
@@ -45,6 +46,7 @@ from app.dao.services_dao import (
     dao_find_services_with_high_failure_rates,
 )
 from app.dao.users_dao import delete_codes_older_created_more_than_a_day_ago
+from app.letters.utils import generate_letter_pdf_filename
 from app.models import (
     EMAIL_TYPE,
     JOB_STATUS_ERROR,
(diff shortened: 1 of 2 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -207,14 +209,38 @@ def replay_created_notifications():
...
+            notify_celery.send_task(
+                name=TaskNames.SCAN_FILE,
+                kwargs={'filename': filename},
+                queue=QueueNames.ANTIVIRUS,
+            )
+        else:
+            letters.append(letter)
 
     if len(letters) > 0:
         letter_ids = [(str(letter.id), letter.reference) for letter in letters]
```
servingUpAces, line 236: If there is a letter here, we could kick off the `app.celery.letters_pdf_tasks.process_sanitised_letter`, right?

app/celery/scheduled_tasks.py
```diff
@@ -207,14 +209,38 @@ def replay_created_notifications():
...
+            )
+            notify_celery.send_task(
+                name=TaskNames.SCAN_FILE,
+                kwargs={'filename': filename},
+                queue=QueueNames.ANTIVIRUS,
+            )
+        else:
+            letters.append(letter)
 
     if len(letters) > 0:
```
servingUpAces, line 235: If there is a letter here, we could kick off the app.celery.letters_pdf_tasks.process_sanitised_letter, right?

servingUpAces (commented):
This is great, I think we can take it a small step further

app/celery/scheduled_tasks.py
```diff
@@ -207,14 +209,38 @@ def replay_created_notifications():
...
         letter_ids = [(str(letter.id), letter.reference) for letter in letters]
 
-        msg = """{} precompiled letters have been pending-virus-check for over 90 minutes. Follow runbook to resolve:
+        msg = f"""{len(letters)} precompiled letters have been pending-virus-check for over 90 minutes.
+            We couldn't find them in the scan bucket.
+
+            Follow runbook to resolve:
             https://github.com/alphagov/notifications-manuals/wiki/Support-Runbook#Deal-with-letter-pending-virus-scan-for-90-minutes.
```
servingUpAces, line 242: This run book entry will need to be updated.

app/celery/scheduled_tasks.py
leohemsted: hmmm, i'm keen to think about other failure cases, but `process_sanitised_letter` expects the letter to be in the scan bucket. so that task would just fail.

I think it's fine to keep as is. The letter might have passed the virus scan and fallen over in the `sanitise_letter` task, but doing another virus scan on it won't hurt.

app/celery/scheduled_tasks.py
leohemsted: Yeah, it was on my radar to do when merging this.

That said, your comment prompted me to do some thinking about it and drafting up new words - and now I'm in favour of deleting the runbook. I don't think I can offer any concrete steps that we know you'll have to do when the letter is in `pending-virus-scan` but ISN'T in the scan bucket. I don't think we need a runbook just to say "you will have to investigate and find the letter and may need to run any number of different tasks to unstick this".

app/celery/scheduled_tasks.py
servingUpAces: oh right, it looks in the scan bucket too. Fair enough.

app/celery/scheduled_tasks.py
servingUpAces: Deleting the entry is probably best.

servingUpAces: Shall we remove the reference to it here?

leohemsted: Yeah lets delete it.

servingUpAces (approved):

PR #3488, opened 2022-03-17
Fix SMS priority adjustment if only 1 provider

benthorner:
https://www.pivotaltracker.com/story/show/181574489

Fixes:

    >       reduced_provider = providers[identifier]
    E       KeyError: 'firetext'


servingUpAces (approved):
Leaving a thought: if we only have one provider do we want to pause the `switch-current-sms-provider-on-slow-delivery` scheduled task that runs once a minute?

PR #3492, opened 2022-03-24
Add boilerplate for Reach SMS callbacks

benthorner:
This is enough to update a notification in the DB. See
the last commit for the main change / testing locally.

The first few commits are some small refactorings based
on problems I encountered getting to the last commit.


app/clients/sms/reach.py
```diff
@@ -0,0 +1,25 @@
...
+        return ("permanent-failure", "TODO: Permanent failure")
+    else:
+        raise KeyError
+
+
+class ReachClientResponseException(SmsClientResponseException):
+    pass  # TODO (custom exception for errors)
+
+
+class ReachClient(SmsClient):
```
servingUpAces, line 19: Should we add init_app and record_outcome to this boiler plate code?

benthorner: Eventually yes. The class isn't actually needed for this PR (callbacks), but I wanted to put something here in case this PR went in first.

My other PR about the sending side [does a bunch of refactoring to avoid copy/pasting init_app and record_outcome](https://github.com/alphagov/notifications-api/pull/3493/files#diff-95316b574f974237b3b7ce453fec09628bd1062087154789ff4d0176ea23c460R23).

PR #3493, opened 2022-03-25
Add boilerplate for sending SMS via Reach

benthorner:
https://www.pivotaltracker.com/story/show/181574489

Separate PR for callback boilerplate: https://github.com/alphagov/notifications-api/pull/3492

Like the callback PR, I've done a few refactorings to avoid
a lot of copy/pasta when creating the new provider for Reach.

See the last commit for instructions about local testing.

Note: this will break the "Change priority" screen in the Admin
app [^1], hence marking the PR with "DNM" for the time being. I
hope to have a PR that fixes that soon.

[^1]: https://www.notifications.service.gov.uk/provider/edit-sms-provider-ratio

### DIFF
--- a/app/clients/sms/__init__.py
+++ b/app/clients/sms/__init__.py
@@ -12,7 +12,7 @@ class SmsClientResponseException(ClientException):
         self.message = message
 
     def __str__(self):
-        return "Message {}".format(self.message)
+        return f"SMS client error ({self.message})"
 
 
 class SmsClient(Client):
--- a/app/clients/sms/__init__.py
+++ b/app/clients/sms/__init__.py
@@ -38,7 +38,7 @@ class SmsClient(Client):
             self.statsd_client.incr(f"clients.{self.name}.error")
             self.current_app.logger.warning(log_message)
 
-    def send_sms(self, to, content, reference, international, sender=None):
+    def send_sms(self, to, content, reference, international, sender):
         start_time = monotonic()
         sender = self.from_number if sender is None else sender
 
--- a/app/clients/sms/__init__.py
+++ b/app/clients/sms/__init__.py
@@ -23,6 +23,7 @@ class SmsClient(Client):
     def init_app(self, current_app, statsd_client):
         self.current_app = current_app
         self.statsd_client = statsd_client
+        self.from_number = self.current_app.config.get('FROM_NUMBER')
 
     def record_outcome(self, success):
         log_message = "Provider request for {} {}".format(
@@ -39,6 +40,7 @@ class SmsClient(Client):
 
     def send_sms(self, to, content, reference, international, sender=None):
         start_time = monotonic()
+        sender = self.from_number if sender is None else sender
 
         try:
             response = self.try_send_sms(to, content, reference, international, sender)
--- a/app/clients/sms/__init__.py
+++ b/app/clients/sms/__init__.py
@@ -1,3 +1,5 @@
+from time import monotonic
+
 from app.clients import Client, ClientException
 
 
@@ -18,6 +20,10 @@ class SmsClient(Client):
     Base Sms client for sending smss.
     '''
 
+    def init_app(self, current_app, statsd_client):
+        self.current_app = current_app
+        self.statsd_client = statsd_client
+
     def record_outcome(self, success):
         log_message = "Provider request for {} {}".format(
             self.name,
@@ -31,7 +37,23 @@ class SmsClient(Client):
             self.statsd_client.incr(f"clients.{self.name}.error")
             self.current_app.logger.warning(log_message)
 
-    def send_sms(self, *args, **kwargs):
+    def send_sms(self, to, content, reference, international, sender=None):
+        start_time = monotonic()
+
+        try:
+            response = self.try_send_sms(to, content, reference, international, sender)
+            self.record_outcome(True)
+        except SmsClientResponseException as e:
+            self.record_outcome(False)
+            raise e
+        finally:
+            elapsed_time = monotonic() - start_time
+            self.statsd_client.timing(f"clients.{self.name}.request-time", elapsed_time)
+            self.current_app.logger.info("Reach request for {} finished in {}".format(reference, elapsed_time))
+
+        return response
+
+    def try_send_sms(self, *args, **kwargs):
         raise NotImplementedError('TODO Need to implement.')
 
     @property
--- a/app/clients/sms/__init__.py
+++ b/app/clients/sms/__init__.py
@@ -40,6 +40,14 @@ class SmsClient(Client):
 
     def send_sms(self, to, content, reference, international, sender):
         start_time = monotonic()
+
+        if sender is None:
+            # temporary log to see if the following ternary is necessary
+            # or if it's safe to remove it - keep for 1-2 weeks
+            self.current_app.logger.warning(
+                f"send_sms called with 'sender' of 'None' for {reference}"
+            )
+
         sender = self.from_number if sender is None else sender
 
         try:
--- a/app/clients/sms/__init__.py
+++ b/app/clients/sms/__init__.py
@@ -18,6 +18,19 @@ class SmsClient(Client):
     Base Sms client for sending smss.
     '''
 
+    def record_outcome(self, success):
+        log_message = "Provider request for {} {}".format(
+            self.name,
+            "succeeded" if success else "failed",
+        )
+
+        if success:
+            self.current_app.logger.info(log_message)
+            self.statsd_client.incr(f"clients.{self.name}.success")
+        else:
+            self.statsd_client.incr(f"clients.{self.name}.error")
+            self.current_app.logger.warning(log_message)
+
     def send_sms(self, *args, **kwargs):
         raise NotImplementedError('TODO Need to implement.')
 
--- a/app/clients/sms/__init__.py
+++ b/app/clients/sms/__init__.py
@@ -21,5 +21,6 @@ class SmsClient(Client):
     def send_sms(self, *args, **kwargs):
         raise NotImplementedError('TODO Need to implement.')
 
-    def get_name(self):
+    @property
+    def name(self):
         raise NotImplementedError('TODO Need to implement.')
--- a/app/clients/sms/mmg.py
+++ b/app/clients/sms/mmg.py
@@ -1,5 +1,4 @@
 import json
-from time import monotonic
 
 from requests import RequestException, request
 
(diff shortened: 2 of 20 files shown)


app/clients/sms/mmg.py
```diff
@@ -77,22 +77,6 @@ def init_app(self, current_app, statsd_client, *args, **kwargs):
         self.statsd_client = statsd_client
         self.mmg_url = current_app.config.get('MMG_URL')
 
-    def record_outcome(self, success, response):
-        status_code = response.status_code if response else 503
-        log_message = "API {} request {} on {} response status_code {}".format(
```
leohemsted, line 83: we currently rely on finding these logs lines in cloudwatch to gather historical stats. i'm kinda interested in moving away from this method anyway since the cloudwatch insight always times out so is an approximation anyway, but we'd need something in place before we can change this log line.

https://github.com/alphagov/notifications-manuals/wiki/Support-Runbook#provider-5xx-vs-2xx-stats

app/clients/sms/mmg.py
```diff
@@ -77,22 +77,6 @@ def init_app(self, current_app, statsd_client, *args, **kwargs):
         self.statsd_client = statsd_client
         self.mmg_url = current_app.config.get('MMG_URL')
 
-    def record_outcome(self, success, response):
-        status_code = response.status_code if response else 503
```
leohemsted, line 82: we're losing the logging of this information - what the status code is is quite useful. Other things such as the response body might also be useful but we don't currently capture that at all.

(You can see by looking at existing log lines that we don't currently log this exception further up. eg this kibana link[^1] for a firetext 503 the other week


[^1]: https://kibana.logit.io/s/9423a789-282c-4113-908d-0be3b1bc9d1d/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:requestId,negate:!f,type:phrase,value:'216c191853a5f7c77da9d9e2ee056025'),query:(match:(requestId:(query:'216c191853a5f7c77da9d9e2ee056025',type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:cf.app,negate:!f,type:phrase,value:notify-delivery-worker-sender),query:(match:(cf.app:(query:notify-delivery-worker-sender,type:phrase))))),index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'*')),sort:!('@timestamp',desc))

app/clients/sms/__init__.py
```diff
@@ -23,6 +23,7 @@ class SmsClient(Client):
     def init_app(self, current_app, statsd_client):
         self.current_app = current_app
         self.statsd_client = statsd_client
+        self.from_number = self.current_app.config.get('FROM_NUMBER')
```
leohemsted, line 26: is this ever used now? relatively surprised that this is decided at the sms client level rather than earlier on.

looking here[^1] i expect that `sender` is always set and we can make the arg non-optional and remove the member variable entirely.


[^1]: https://github.com/alphagov/notifications-api/blob/master/app/delivery/send_to_providers.py#L77

leohemsted (commented):
i'm a little uncertain about making those tests more lax but will think overnight. but have another couple of comments to address while we think about that.

app/clients/sms/mmg.py
benthorner: Yeah, I noticed. I was hoping [this commit](https://github.com/alphagov/notifications-api/pull/3493/commits/2746be97c4f5b8266c03f7aeb3ac1c1d76e0559d) would be a reasonable / better replacement over our custom interpretation of the status code. Having the response at this level is tricky because it couples the base class to the way we make the request down below. We can of course move / have some logs back down - that commit is my attempt to have the best of both. What do you think?

leohemsted: Ah, good point. I'd missed that the `ReachClientResponseException` overrides the base exc class's `__str__` to contain the code. (And so does MMGClientResponseException and FiretextClientResponseException - those three exception class implementations look like a good candidate for DRYing out 🤔)

I like the approach you've added in provider_tasks. i think logging exceptions at the point that they're caught/suppressed is generally best practice.

i just tested it and the log line looks like the following on a 504, which i think is fine.

```
[2022-03-29 11:22:45,703: WARNING/ForkPoolWorker-2] SMS notification delivery for id: a2ff3cc8-2bd8-4ffd-ab1f-4886843417ab failed
Traceback (most recent call last):
  File "/home/vcap/app/app/clients/sms/reach.py", line 40, in try_send_sms
    response = request(
  File "/usr/local/lib/python3.9/site-packages/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/requests/sessions.py", line 528, in request
    prep = self.prepare_request(req)
  File "/usr/local/lib/python3.9/site-packages/requests/sessions.py", line 456, in prepare_request
    p.prepare(
  File "/usr/local/lib/python3.9/site-packages/requests/models.py", line 316, in prepare
    self.prepare_url(url, params)
  File "/usr/local/lib/python3.9/site-packages/requests/models.py", line 390, in prepare_url
    raise MissingSchema(error)
requests.exceptions.MissingSchema: Invalid URL 'TODO': No schema supplied. Perhaps you meant http://TODO?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/vcap/app/app/celery/provider_tasks.py", line 23, in deliver_sms
  File "/home/vcap/app/app/delivery/send_to_providers.py", line 86, in send_sms_to_provider
    raise e
  File "/home/vcap/app/app/delivery/send_to_providers.py", line 81, in send_sms_to_provider
    provider.send_sms(**send_sms_kwargs)
  File "/home/vcap/app/app/clients/sms/__init__.py", line 50, in send_sms
    raise e
  File "/home/vcap/app/app/clients/sms/__init__.py", line 46, in send_sms
    response = self.try_send_sms(to, content, reference, international, sender)
  File "/home/vcap/app/app/clients/sms/reach.py", line 56, in try_send_sms
    raise ReachClientResponseException(response=e.response, exception=e)
app.clients.sms.reach.ReachClientResponseException: Code 504 text Gateway Time-out exception Invalid URL 'TODO': No schema supplied. Perhaps you meant http://TODO?
```

so in summary: i think your approach is fine and losing that line from record_outcome is fine now that we print out the stack trace too.

as an aside it might be nice to DRY up the three SMS exception classes since the functions look identical to me and all override the base class

app/clients/sms/__init__.py
benthorner: Good spot. We can certainly make it a required arg: https://github.com/alphagov/notifications-api/pull/3493/commits/6cbaa1bd4af2e75628aae45c17c688d4f982eed0.

The ternary is harder: although the argument is always provided, the `sender` may still be `None`. We persist notifications in quite a few places and ways so it's quite hard to tell. I know:

- All services must have a default SMS sender, so I'd never _expect_ `None`.
- None of the notifications in our Production DB have null `reply_to_text`.
- The original code was added [a very long time ago](https://github.com/alphagov/notifications-api/commit/c29dd237027dcbe6bd2073aaa774d0f90039d75b#diff-5517993ae3247630cf07bd936b84e1382ba7c764e9377815a91e29efeebdea6bR74) so might be redundant.

I'm cautious about making this change as part of the PR, partly because removing it will be more work and partly because I'm not totally confident in just dropping it e.g. if the assumption breaks in future.

How about:

- A separate PR to look at that (after this one).
- Some more thinking about how to enforce its presence.

app/clients/sms/mmg.py
benthorner: Thank you 🙂.

I had the same thought about the exception classes (and the JSON validation). In theory - and in practice for Firetext - they _could_ be different so the duplication seemed justified for the validation at least.

For the exceptions, how do you feel about this?

https://github.com/alphagov/notifications-api/pull/3493/commits/0fa34339353e2e20b8fa2adf2e252307cf48d763

app/clients/sms/mmg.py
benthorner: Great catch 💯.

As discussed in Slack, we can still make future reports work by parsing logs based on "success" and "failed". We're about to start a new quarter, so now is the ideal time to change the reporting.

I asked Caley about this and it's unlikely we'll need reports for previous quarters. I can save the old query and mention it in the runbook - I assume we can just ditch it after a year or so.

Does that sound OK?

app/clients/sms/__init__.py
leohemsted: Sounds good! Good sleuthing around the impact. I suspect that it's never None any more but could have been in the early days. Agree that it best belongs in another PR

As a compromise, how do we feel about adding a log line if sender==None that we can then check for in a couple of weeks?

benthorner: Sure, that's a good idea - still possible of course but should be very niche.

I've added a log in https://github.com/alphagov/notifications-api/pull/3493/commits/0980e093bafafd799fdd94c709a1a9ca981d1f09 and a task on the card to revisit.

app/clients/sms/mmg.py
leohemsted: ooh not bad, didn't think about removing the subclasses entirely. i was initially a little worried that it might not contain enough cues as to which provider had the issues, but the stack trace includes the provider file:

```
[...]
  File "/home/vcap/app/app/clients/sms/reach.py", line 50, in try_send_sms
    raise SmsClientResponseException("Request failed")
```

(plus also there's the separate log line `Provider request for reach failed` but that doesn't have any context)

leohemsted (approved):

benthorner:
Discussed with @leohemsted and agreed we will ship this on Thursday or Friday to align with the new quarter. This is so [we can smoothly switch over the SMS report to a new query for the next and future quarters](https://github.com/alphagov/notifications-api/pull/3493#discussion_r836602643).

Currently this PR is "DO NOT MERGE" as it depends on https://github.com/alphagov/notifications-admin/pull/4199. However, in team discussions we've agreed we rarely use this UI so on balance it's OK for it to be broken until someone reviews the PR to fix it.

app/clients/sms/mmg.py
benthorner: Update: we agreed it's OK to change the logging at the start of the quarter so long as we update the query and runbook to match, and have a way to query the old data if we need to.

PR #3502, opened 2022-04-08
add new daily sms provider volume report

leohemsted:
code generally lifted almost exactly from the daily_volumes_report, but per provider and only for SMS.

see https://github.com/alphagov/notifications-admin/pull/4206

### DIFF
--- a/tests/app/dao/test_ft_billing_dao.py
+++ b/tests/app/dao/test_ft_billing_dao.py
@@ -10,6 +10,7 @@ from app.dao.fact_billing_dao import (
     delete_billing_data_for_service_for_day,
     fetch_billing_data_for_day,
     fetch_billing_totals_for_year,
+    fetch_daily_sms_provider_volumes_for_platform,
     fetch_daily_volumes_for_platform,
     fetch_letter_costs_and_totals_for_all_services,
     fetch_letter_line_items_for_all_services,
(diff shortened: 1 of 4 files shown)


tests/app/dao/test_ft_billing_dao.py
```diff
@@ -856,6 +857,84 @@ def test_fetch_daily_volumes_for_platform(
     assert results[1].letter_sheet_totals == 40
 
 
+def test_fetch_daily_sms_provider_volumes_for_platform_for_platform_groups_values_by_provider(
```
benthorner, line 860:
```suggestion
def test_fetch_daily_sms_provider_volumes_for_platform_groups_values_by_provider(
```

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -856,6 +857,84 @@ def test_fetch_daily_volumes_for_platform(
...
+    assert results[0].provider == 'bar'
+    assert results[0].sms_totals == 80
+    assert results[0].sms_fragment_totals == 160
+
+    assert results[1].provider == 'foo'
+    assert results[1].sms_totals == 5
+    assert results[1].sms_fragment_totals == 10
+
+
+def test_fetch_daily_sms_provider_volumes_for_platform_for_platform_calculates_funky_fragments_and_costs(
```
benthorner, line 890:
```suggestion
def test_fetch_daily_sms_provider_volumes_for_platform_for_platform_calculates_chargeable_units_and_costs(
```

I'm tempted to leave it but I suspect people in the future won't know what we're on about so we should stick to the official term.

tests/app/platform_stats/test_rest.py
```diff
@@ -238,3 +239,23 @@ def test_volumes_by_service_report(
...
+    )
+
+    assert len(resp) == 1
+    assert resp[0] == {
+        'day': '2022-03-01',
+        'provider': 'foo',
+        'sms_totals': 1,
+        'sms_fragment_totals': 2,
+        'sms_chargeable_units': 2,
+        'sms_cost': 3,
```
benthorner, line 260: You could make this a decimal to test the casting implicitly.

tests/app/dao/test_ft_billing_dao.py
benthorner: Optional: you could reasonably combine this and the previous test as the generic `test_fetch_daily_sms_provider_volumes_for_platform` - together they convey what the function returns and it's useful to see it in one place.

tests/app/dao/test_ft_billing_dao.py
```diff
@@ -856,6 +857,84 @@ def test_fetch_daily_volumes_for_platform(
...
+    create_ft_billing('2022-02-03', sample_template)
+    create_ft_billing('2022-02-04', sample_template)
+    create_ft_billing('2022-02-05', sample_template)
+
+    # too late
+    create_ft_billing('2022-02-06', sample_template)
+
+    results = fetch_daily_sms_provider_volumes_for_platform(start_date='2022-02-03', end_date='2022-02-05')
+
+    assert len(results) == 3
```
benthorner, line 918: Optional: you could cover the date ordering here as well.

benthorner (approved):
Nice work - all very clear 💯.

I've left a handful of comments but nothing to block this going in.

One thought I had while reviewing: would it be useful to separate out international SMS? The reason I ask is because, at the moment, international SMS always goes to one provider and is load balanced separately; if we need to adjust the load balancing for domestic SMS then we may get it slightly wrong if we're including SMS volumes in our calculations.

☝️ worth checking with Katie - could also be a follow-up bit of work depending on how big an impact international SMS has on projections (I haven't checked).

leohemsted:
I did check with katie, she confirmed that no she doesn't need international split out. unfortunately we had the chat in person and i cant remember exactly what she said!

tests/app/dao/test_ft_billing_dao.py
leohemsted: i think this is covered anyway in this test by virtue of checking results[0] and results[-1] equaling the lowest and the highest dates in the search

tests/app/platform_stats/test_rest.py
leohemsted: hmmm i'm not sure i follow - do you mean `assert isinstance(resp[0]['sms_cost'], float)` or similar? because the admin_request fixture takes json response and parses it - so that it was able to complete succesfully implies that we are casting all decimals to alternative json serialisable numerics

benthorner: I was just thinking: `'sms_cost': 3.797` (or whatever).

Very optional though: you're right this confirms it's implicitly working. Using a proper floating point number would just rule out things like the cast only working for integers, or the cast truncating to an integer.

leohemsted: ah right. Good point, that's a low effort change to the input vals. i'll do that 👍

PR #3503, opened 2022-04-08
Don't error sending a letter that's sent already

benthorner:
Fixes this error (in Admin):

           File "/home/vcap/app/app/notify_client/notification_api_client.py",
      line 74, in send_precompiled_letter
            return
      self.post(url='/service/{}/send-pdf-letter'.format(service_id), data=data)
          File "/home/vcap/app/app/notify_client/__init__.py", line 59, in post
            return super().post(*args, **kwargs)
          File
      "/home/vcap/deps/0/python/lib/python3.9/site-packages/notifications_python_client/base.py",
      line 48, in post
            return self.request("POST", url, data=data)
          File
      "/home/vcap/deps/0/python/lib/python3.9/site-packages/notifications_python_client/base.py",
      line 64, in request
            response = self._perform_request(method, url, kwargs)
          File
      "/home/vcap/deps/0/python/lib/python3.9/site-packages/notifications_python_client/base.py",
      line 118, in _perform_request
            raise api_error
        notifications_python_client.errors.HTTPError: 500 - Internal server
      error

Due to this error (in API):

           File "/home/vcap/app/app/service/send_notification.py", line 178, in
      send_pdf_letter_notification
            raise e
          File "/home/vcap/app/app/service/send_notification.py", line 173, in
      send_pdf_letter_notification
            letter =
      utils_s3download(current_app.config['TRANSIENT_UPLOADED_LETTERS'],
      file_location)
          File
      "/home/vcap/deps/0/python/lib/python3.9/site-packages/notifications_utils/s3.py",
      line 53, in s3download
            raise S3ObjectNotFound(error.response, error.operation_name)
        notifications_utils.s3.S3ObjectNotFound: An error occurred (NoSuchKey)
      when calling the GetObject operation: The specified key does not exist.

I checked the DB to verify the letter does actually exist i.e. it is an
instance of the problem we're fixing here.

Since I was in the area I also did a bit of light refactoring to the tests,
so best to review commit by commit.


app/service/send_notification.py
```diff
@@ -172,9 +175,14 @@ def send_pdf_letter_notification(service_id, post_data):
...
         letter = utils_s3download(current_app.config['TRANSIENT_UPLOADED_LETTERS'], file_location)
     except S3ObjectNotFound as e:
-        current_app.logger.exception('Letter {}.pdf not in transient {} bucket'.format(
+        current_app.logger.warning('Letter {}.pdf not in transient {} bucket'.format(
             post_data['file_id'], current_app.config['TRANSIENT_UPLOADED_LETTERS'])
         )
+
+        # notification already exists e.g. if the user clicked send in different tabs
+        if get_notification_by_id(post_data['file_id']):
+            return {'id': str(post_data['file_id'])}
```
leohemsted, line 184: nice. could this be up the top of the function so we get out of this fn even earlier if the notification is created and dont even check for the file?

leohemsted (approved):
one optional comment. but a nice sensible tweak

app/service/send_notification.py
benthorner: I thought about that but it would mean an extra unnecessary DB operation for the majority of requests. I appreciate this is just requests from the Admin app so we're not overly concerned about performance.

Do you have a strong preference either way?

leohemsted: yeah, given its just admin i thought it'd be minimal. also in the happy path, it's making at least three DB queries, getting and reading a file from s3, and moving a file between buckets, so a simple primary key lookup should be realllllly quick in comparison, even if it's not necessary for 99% of the time.

it also fits in well with similar checks for, eg, if the user has permissions in admin app. You shouldn't be in this function at all since the letter is already sent - so just bail out at the top. also i just prefer exception handling clauses to be a bit more simple in general.

that said none of these views are particularly strongly held, i'll leave it up to you

benthorner: Sure, changed in https://github.com/alphagov/notifications-api/pull/3503/commits/413c6c4c2653deba6692bed74deeb0bb8c4c7f37.

PR #3505, opened 2022-04-11
Fix default letter address not being set on new letter templates

idavidmcdonald:
We spotted that when you create a new letter template, it doesn't have the default sender address set. Chris agrees that this is a bug. This PR fixes that so when you create a new letter template, it uses the default. If you don't have a default then no sender address is set.

image

It is worth being aware of the impact of this change on a related letter sender address bug. There is a bug where if you copy a letter template, then the new template does not have the sender address of the copied template and is left blank. This PR changes the behaviour of that bug such that if you copy a letter template, then regardless of the value of the sender address for that template, the new template ends up with the services default sender address. Maybe this is an improvement or it could equally be seen as worse. Hopefully Chris has an opinion on whether this is fine or if this PR should only be merged if we also fix that bug too. I have done a quick bit of looking into fixing that bug but it looks like a harder thing to fix and is something I'm not going to try and tackle now.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -563,7 +563,7 @@ class Service(db.Model, Versioned):
 
     def get_default_letter_contact(self):
         default_letter_contact = [x for x in self.letter_contacts if x.is_default]
-        return default_letter_contact[0].contact_block if default_letter_contact else None
+        return default_letter_contact[0] if default_letter_contact else None
 
     def has_permission(self, permission):
         return permission in [p.permission for p in self.permissions]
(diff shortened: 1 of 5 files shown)


benthorner:
Before looking at the code in detail I'd like to clarify a few things:

- Does the default sender feature currently do anything for letters? What did it used to do?
- Is this the right place to fix the bug, or will it make the "copy template" even more surprising?*

\*Your code change prompted me to go check the behaviour for postage, which is also an overriding one: if I copy a first class letter the postage gets reset to second class.

I agree these are questions for @quis...

samuelhwilliams:
@idavidmcdonald do we still need this or shall we close it off?

idavidmcdonald:
@quis, are you able to make a call on the above questions please?

Very happy for this to be closed if we decide it isn't good to go in as is due to the related complications.

samuelhwilliams:
Shall we just close this off? 🤔

idavidmcdonald:
Yep, too much inaction

PR #3511, opened 2022-04-12
Move existing NHS orgs without branding onto NHS branding

CrystalPea:
This is done to make self-service branding easier to implement, and also because NHS branding makes much more sense for services in those orgs than GOV.UK branding.

Part of this story: https://www.pivotaltracker.com/story/show/181415981

Should go in after: https://github.com/alphagov/notifications-api/pull/3510


benthorner (approved):
This looks OK to me.

Worth noting that we don't need to update any NHS services as they should all have NHS branding already by default [^1].

[^1]: https://github.com/alphagov/notifications-api/blob/master/app/dao/services_dao.py#L328

PR #3515, opened 2022-04-14
Bump straightforward dependencies

klssmith:
This bumps the dependencies which don't need code changes in order to fix. Two of these are major version bumps, but I can't see anything in them which should affect our code.


idavidmcdonald (approved):
👍

PR #3527, opened 2022-04-28
add revision to fix apr rate

leohemsted:
this will update about 80-90k rows in ft billing.

Can be merged whenever - before or after. as long as it's not deploying while we're creating ft billing rows, it should be fine

### DIFF
--- a/migrations/versions/0370_fix_apr_2022_sms_rate.py
+++ b/migrations/versions/0370_fix_apr_2022_sms_rate.py
@@ -1,0 +1,34 @@
+"""
+
+Revision ID: 0370_fix_apr_2022_sms_rate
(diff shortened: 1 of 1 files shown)


benthorner (approved):
This looks good to me 👍.

Not something to hold up this PR, but why do we have [a `ProviderRates` table](https://github.com/alphagov/notifications-api/blob/66bc11bb65b2c3de13f9051312d54c61418052e5/app/models.py#L1149) as well as a `Rates` table? I can see `Rates` is the one that's actually used.

leohemsted:
was a decision made a long time ago, it seems. https://github.com/alphagov/notifications-api/commit/43226dc93252f16d85a8381836e661497c57efcc

PR #3531, opened 2022-05-03
clean up user tests

leohemsted:
an API build failed because one of the tests expected the database to be empty, but it actually wasn't. This is probably because another test run earlier on that worker did not clear down properly. i'm not 100% sure which test this was but i've fixed a couple of tests in this file. I've also changed those particular tests to delete any users at the beginning of the test.

I took the opportunity to refresh all of these tests to ensure they all correctly tear down, and also use the more modern admin_request fixture ratehr than the old client one that required us to specify headers and do json parsing etc.

### DIFF
--- a/tests/app/user/test_rest.py
+++ b/tests/app/user/test_rest.py
@@ -1,10 +1,9 @@
-import json
 import uuid
 from datetime import datetime
 from unittest import mock
 
 import pytest
-from flask import current_app, url_for
+from flask import current_app
 from freezegun import freeze_time
 
 from app.dao.permissions_dao import default_service_permissions
@@ -21,7 +20,6 @@ from app.models import (
     Permission,
     User,
 )
-from tests import create_admin_authorization_header
 from tests.app.db import (
     create_organisation,
     create_service,
@@ -99,11 +97,11 @@ def test_get_user_doesnt_return_inactive_services_and_orgs(admin_request, sample
     assert fetched['permissions'] == {}
 
 
-def test_post_user(client, notify_db, notify_db_session):
+def test_post_user(admin_request, notify_db_session):
     """
     Tests POST endpoint '/' to create a user.
     """
-    assert User.query.count() == 0
+    User.query.delete()
     data = {
         "name": "Test User",
         "email_address": "user@digital.cabinet-office.gov.uk",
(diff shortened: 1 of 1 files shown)


tests/app/user/test_rest.py
```diff
@@ -252,22 +231,20 @@ def test_cannot_create_user_with_empty_strings(admin_request, notify_db_session)
...
+    json_resp = admin_request.post(
+        'user.update_user_attribute',
+        user_id=sample_user.id,
+        _data=update_dict
+    )
 
-    assert resp.status_code == 200
-    json_resp = json.loads(resp.get_data(as_text=True))
     assert json_resp['data'][user_attribute] == user_value
+    assert getattr(sample_user, user_attribute) == user_value
```
benthorner, line 247: I was confused to see this addition at first but I think I get it: we want to check the model was actually updated, not just that the POSTed data is echoed back in the response.

tests/app/user/test_rest.py
```diff
@@ -1088,72 +1039,58 @@ def test_get_orgs_and_services_only_shows_users_orgs_and_services(admin_request,
...
+        _data=data,
     )
-    users = json.loads(response.get_data(as_text=True))
 
-    assert response.status_code == 200
     assert users['data'] == []
 
 
-def test_search_for_users_by_email_handles_incorrect_data_format(notify_db, client):
+def test_search_for_users_by_email_handles_incorrect_data_format(notify_db_session, admin_request):
```
benthorner, line 1083: This is interesting: I didn't realise we did our own database cleanup. Looks consistent within this file, but I can see a fair few non-`_session` uses out there. I suspect many of them should be changed and we should mark this fixture as semi-private.

How about a quick comment on each fixture to clarify when it should be used?

benthorner (approved):
Another great cleanup 💯.

It was hard to spot the other changes but I get how one thing can lead to another. Deleting all the users still seems brittle to me - prevents us having any seed users in future - but is still less brittle then "hoping" (`assert`) there are no users 👍.

tests/app/user/test_rest.py
leohemsted: yeah, quick comment makes a lot of sense. i'd love some kind of static analysis tool that tells us how our fixtures are used amongst the various unit tests to identify usages of notify_db

leohemsted: i've made a couple of commits to address this (by making `notify_db` private) but they're pretty big so I think I'll put them in a separate PR.

PR #3532, opened 2022-05-03
bump utils

leohemsted:
mostly to get rid of the security warning on pypdf2


quis (approved):
Same change as https://github.com/alphagov/notifications-admin/pull/4224 👍🏻

PR #3533, opened 2022-05-03
remove provider_rates table

leohemsted:
this was added five years ago but never used. if we want to bring back variable rates per client we might as well get a fresh start since a lot has changed since then.


benthorner (approved):
Nice, thanks for cleaning up ⭐ 🚀

PR #3534, opened 2022-05-03
Remove redundant fields from service usage APIs

benthorner:
https://www.pivotaltracker.com/story/show/181935935

These are no longer used since [^1] and [^2].

[^1]: https://github.com/alphagov/notifications-admin/pull/4225
[^2]: https://github.com/alphagov/notifications-admin/pull/4229


idavidmcdonald (approved):
Nice

benthorner:
I've manually verified the usage dashboards are identical before / after the deployment for services identified in [the original API PR](https://github.com/alphagov/notifications-api/pull/3520), including the current and previous years. It's safe to remove the old fields.

PR #3535, opened 2022-05-03
Notify db fixture cleanup

leohemsted:
Spurred by ben's comments[^1], I thought I'd make the `notify_db` fixture private and make it clearer what `notify_db_session` does. 

* `notify_db` runs once per session (test run/worker thread), connects to the database, and performs migrations
* `notify_db_session` runs for every test and deletes all non static data from the db after your test finishes. (except static type tables etc)

We shouldn't really be encouraging use of `notify_db` on its own, as it doesn't do any tearing down, so if you inadvertently create any entries in the db they'll linger around and screw up the next test that runs - this is super hard to debug when it happens as it's not easy to find out what the previous test was.

So, to solve this, make the `notify_db` fixture private, and remove all usage of it from the test suite. Anywhere that used to use it to access the `db.session` object now gets the reference from the notify_db_session fixture instead.

[^1]: https://github.com/alphagov/notifications-api/pull/3531#discussion_r863867876

### DIFF
--- a/tests/app/broadcast_message/test_utils.py
+++ b/tests/app/broadcast_message/test_utils.py
@@ -270,7 +270,6 @@ def test_update_broadcast_message_status_rejects_approval_of_broadcast_with_no_a
 
 
 def test_update_broadcast_message_status_allows_trial_mode_services_to_approve_own_message(
-    notify_db,
     sample_broadcast_service,
     mocker
 ):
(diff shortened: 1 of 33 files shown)


benthorner (approved):
Another great cleanup 🧹.

I like the parity of `notify_db.session` vs `notify_db_session`. I also noticed you were able to remove the fixture entirely in some tests.

Out of scope (mostly thinking to myself): I wonder if we should cleanup globally / by default. This is something I've seen in other repos and has the benefit of being reliable, but the drawback of a likely performance hit. This change is a clear improvement irrespective and there doesn't seem to be much motivating us to do something more global.

PR #3537, opened 2022-05-06
Fix existing marshmallow schemas and remove unused code

klssmith:
There were parts of the marshmallow schemas which either weren't needed or weren't working as expected. While mostly not causing issues now, these do cause issues when we try to upgrade marshmallow. See individual commits for details.


idavidmcdonald (approved):
Nice

PR #3546, opened 2022-05-19
Use notification view for status / billing tasks

benthorner:
https://www.pivotaltracker.com/story/show/178125825

This fixes a bug where (letter) notifications left in sending would
temporarily get excluded from billing and status calculations once the
service retention period had elapsed, and then get included once again when
they finally get marked as delivered.*

Status and billing tasks shouldn't need to have knowledge about which table
their data is in and getting this wrong is the fundamental cause of the bug
here. Adding a view across both tables abstracts this away while keeping
the query complexity the same.

Using a view also has the added benefit that we no longer need to care when
the status / billing tasks run in comparison to the deletion task, since we
will retrieve the same data irrespective (see the commits for a much more 
detailed discussion on data integrity).

*Such a scenario is rare but has happened.

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -467,11 +467,9 @@ def fetch_billing_data_for_day(process_day, service_id=None, check_permissions=F
 
 
 def _query_for_billing_data(notification_type, start_date, end_date, service):
-    table = NotificationAllTimeView
-
     def _email_query():
         return db.session.query(
-            table.template_id,
+            NotificationAllTimeView.template_id,
             literal(service.crown).label('crown'),
             literal(service.id).label('service_id'),
             literal(notification_type).label('notification_type'),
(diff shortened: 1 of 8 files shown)


rjbaker:
Smart use of a union/view! Any idea on result size & memory needed when this is run?

benthorner:
@rjbaker could you clarify what you mean by result size? The end results should be unchanged, but the intermediary union can of course be massive without filtering)due to the combined size of both tables; but it's not radically different to previous queries on the history table alone, which dwarfs the notifications table in size.

As to memory, I'm not sure. It would take quite a bit more effort to reinstate my local setup to test a full run before / after; the individuals queries alone are so quick that I don't think their effect on memory would be discernible. It's also possible the general performance will be worse than predicted - my local data is just a sample.

I think the pragmatic thing to do is just try this out - it's safe to revert if we need to.

app/models.py
```diff
@@ -1411,6 +1411,39 @@ class NotificationStatusTypes(db.Model):
...
 
+class NotificationAllTimeView(db.Model):
+    """
+    WARNING: this view is a union of rows in "notifications" and
+    "notification_history". Any query on this view will query both
+    tables and therefore rely on *both* sets of indices.
+    """
+    __tablename__ = 'notifications_all_time_view'
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True)
```
leohemsted, line 1422: rather than redefining everything, can we just make the class definition extend notification history?

`class NotificationAllTimeView(NotificationHistory)`

keeps it more resilient and reduces duplication. we do similar already with TemplateBase vs Template/TemplateHistory

app/dao/fact_billing_dao.py
```diff
@@ -468,7 +466,9 @@ def fetch_billing_data_for_day(process_day, service_id=None, check_permissions=F
     return transit_data
 
 
-def _query_for_billing_data(table, notification_type, start_date, end_date, service):
+def _query_for_billing_data(notification_type, start_date, end_date, service):
+    table = NotificationAllTimeView
```
leohemsted, line 470: i appreciate this keeping the diff short and the review simple, however, if I were writing an SQL query from scratch I'd use the class name each time we reference a column.

```python
return db.session.query(
    NotificationAllTimeView.template_id,
    ...
```

how do you feel about making that change in a separate commit? i don't think this is a 100% necessary change that blocks this PR, but it might be nice to pre-empt us having to stop to think "now what `table` are we talking about here" if we miss a definition that might be many lines up outside of the big sql query definition we're looking at.

benthorner: Sure, changed in https://github.com/alphagov/notifications-api/pull/3546/commits/8e837cf681ea1e02bcc1f87364c16a6b534e22bb.

I had mixed opinions about this myself, so happy to change it if you prefer the direct class version.

app/models.py
benthorner: I tried this quickly but it's not straightforward:

```
# inheriting from NotificationHistory
NotificationAllTimeView.__tablename__
'notification_history'
```

I expect there's a way around this but I prefer the non-DRY approach:

1. It's easier to see exactly which columns are in the table.
2. It's conceptually clearer: a NotificationAllTimeView isn't a NotificationHistory; it's its own (combo) thing.
3. The other methods and `__table_args__` don't make sense to inherit.

We could potentially have a `NotificationBase` to parallel `TemplateBase`, although it might be a bit weird for the base class of a notification not to have any kind of `to` field.

(1) has been helpful to me on several occasions so a base class path isn't a clear win. `TemplateBase` does set a precedent for it though. Do you think it's worth moving to that pattern for Notifications?

PR #3549, opened 2022-05-26
Upgrade marshmallow-sqlalchemy plus the unpinned dependencies

klssmith:
This upgrades marshmallow-sqlalchemy, which can be unpinned now that https://github.com/alphagov/notifications-api/pull/3541 has been merged. It also upgrades all the unpinned dependencies in `requirements.txt`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -150,7 +150,7 @@ marshmallow==3.15.0
     #   -r requirements.in
     #   flask-marshmallow
     #   marshmallow-sqlalchemy
-marshmallow-sqlalchemy==0.23.1
+marshmallow-sqlalchemy==0.28.0
     # via -r requirements.in
 mistune==0.8.4
     # via notifications-utils
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,19 +4,21 @@
 #
 #    pip-compile requirements.in
 #
-alembic==1.7.4
+alembic==1.7.7
     # via flask-migrate
-amqp==5.0.9
+amqp==5.1.1
     # via kombu
 arrow==1.2.2
     # via isoduration
-attrs==21.2.0
+async-timeout==4.0.2
+    # via redis
+attrs==21.4.0
     # via jsonschema
-awscli==1.21.4
+awscli==1.24.8
     # via awscli-cwlogs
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -197,17 +201,17 @@ python-dateutil==2.8.2
...
+pytz==2022.1
     # via
     #   celery
     #   notifications-utils
 pyyaml==5.4.1
     # via
     #   awscli
     #   notifications-utils
-redis==3.5.3
+redis==4.3.1
```
leohemsted, line 212: there's a few breaking changes wrapped up in here. i had a quick skim through and it appears okay but it would be good to test this, especially with gunicorn since we had the monkeypatching stuff i just wanna be sure we're handling correctly

requirements.txt
```diff
@@ -4,31 +4,33 @@
...
     # via -r requirements.in
-bcrypt==3.2.0
+bcrypt==3.2.2
     # via flask-bcrypt
 beautifulsoup4==4.11.1
     # via -r requirements.in
 billiard==3.6.4.0
     # via celery
-bleach==4.1.0
+bleach==5.0.0
```
leohemsted, line 27: there's a few breaking changes here (https://bleach.readthedocs.io/en/latest/changes.html)

i dont think this'll impact us - but could there be some weird potential for issues if someone passes in a html element with style tags? that's weird and unlikely but i dont want us to be open to bugs because of it (cc @quis)

klssmith: I've rebased this out so that it can be looked at separately, so there's now no change to the bleach version

requirements.txt
klssmith: We're running this version of redis in admin, which seems to do the same monkeypatching. Let me know if you think there is more that we should be testing apart from checking everything runs on preview - if so, I'll do this in it's own PR to be safe.

klssmith:
I've added an extra commit on the end of this PR to pin Werkzeug

requirements.txt
leohemsted: ah had missed that it was already onadmin. yeah thats fine then.

leohemsted (approved):

PR #3551, opened 2022-06-01
Bump utils to version 56.0.0

benthorner:
The only impactful change is the major version itself, where I've
fixed the breaking changes due to the upgrade of PyPDF2 [^1] and
checked there are no deprecation warnings when I run the tests.

[^1]: https://github.com/alphagov/notifications-utils/pull/973


leohemsted (approved):
really dont think we should be doing _any_ pdf stuff outside of template preview but that's a can of worms for another time. bump seems good 👍

PR #3554, opened 2022-06-08
Add endpoint to archive an organisation

klssmith:
The new endpoint checks that an org has no users, pending invited users,
or services (in any state). If it does, admin will display an error message
so that the platform admin user archiving the org can manually reassign
/ remove the team members and services.

Assuming that the org can be archived, we remove its email branding,
letter branding and domains, and set its state to `active=False`.

[Pivotal story](https://www.pivotaltracker.com/story/show/178519743)

### DIFF
--- a/app/dao/organisation_dao.py
+++ b/app/dao/organisation_dao.py
@@ -3,6 +3,7 @@ from sqlalchemy.sql.expression import func
 from app import db
 from app.dao.dao_utils import VersionOptions, autocommit, version_class
 from app.models import Domain, Organisation, Service, User
+from app.utils import get_archived_db_column_value
 
 
 def dao_get_organisations():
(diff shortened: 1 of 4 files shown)


app/dao/organisation_dao.py
```diff
@@ -105,6 +105,20 @@ def _update_organisation_services(organisation, attribute, only_where_none=True)
         db.session.add(service)
 
 
+@autocommit
+def dao_archive_organisation(organisation_id):
+    organisation = dao_get_organisation_by_id(organisation_id)
+
```
idavidmcdonald, line 111: I think in here we may also want to call `organisation.name = get_archived_db_column_value(organisation.name)` which is the equivalent to what we do for the service name in `dao_archive_service`

It probably makes very little difference but I guess 
- it keeps things consistent
- if we ever did want to create a new organisation later with the same name as the old one, we won't be blocked by the unique constraint on the organisation name

idavidmcdonald (approved):
Great work. This all looks good, simple and meets the expectations of the story.

One small comment that I think we should probably do but it is pretty minor.

app/dao/organisation_dao.py
klssmith: My thinking behind not doing that originally was that you could have a organisation, for example "Camden Council", which gets archived. If someone then wants to create a new organisation called "Camden Council" in future, it would essentially be the same organisation, so we'd want to just unarchive the existing org instead of setting up a brand new one with the same name. That might work better for the usage data.

This is a different to services, where the names can be anything, so we want to release the name for another team to use.

I can make the change to do what you suggest though if you think that we wouldn't want to unarchive an organisation like that

idavidmcdonald: Your call on that one. There are pros and cons.

At the moment though we don't offer a way to unarchive without going in the DB so if that is the case, I'd encourage someone to make a new organisation from scratch as that can be done using the user interface

I'll leave you to decide what approach to take as either case doesn't matter too much

klssmith: I've updated this to change the org name

idavidmcdonald (approved):
Thanks for the change. Looks good!

PR #3556, opened 2022-06-13
Rename fact_billing DAO functions consistently

benthorner:

Previously we had a wild variety of overlapping, top-level
and nested function names that made it very hard to navigate
the code used for the ~3 different reports supported here.

This takes each report and renames the associated functions
so they are clearly grouped together, using "_" to indicate
the nested functions that are only used in the file.

Making this change will make it easier to review future PRs.

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -64,7 +64,7 @@ def fetch_sms_free_allowance_remainder_until_date(end_date):
     return query
 
 
-def fetch_sms_billing_for_all_services(start_date, end_date):
+def fetch_usage_for_all_services_sms(start_date, end_date):
 
     # ASSUMPTION: AnnualBilling has been populated for year.
     allowance_left_at_start_date_query = fetch_sms_free_allowance_remainder_until_date(start_date).subquery()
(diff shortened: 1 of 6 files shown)


app/dao/fact_billing_dao.py
```diff
@@ -198,7 +198,7 @@ def fetch_usage_for_all_services_letter_breakdown(start_date, end_date):
     return query.all()
 
 
-def fetch_billing_totals_for_year(service_id, year):
+def fetch_usage_for_service_annual(service_id, year):
```
CrystalPea, line 201: Good change, mentions that it's for a single service 👏🏼

app/dao/fact_billing_dao.py
```diff
@@ -242,7 +242,7 @@ def fetch_billing_totals_for_year(service_id, year):
     ).all()
 
 
-def fetch_monthly_billing_for_year(service_id, year):
+def fetch_usage_for_service_by_month(service_id, year):
```
CrystalPea, line 245: Here we are losing information that it's for a year, but I'm not sure what would be an ideal name, so it's ok if it stays as is.

CrystalPea (approved):

app/dao/fact_billing_dao.py
benthorner: Yes I struggled to get to a set of names that worked for everything - this seems like the least worst option.

PR #3561, opened 2022-06-17
Support variable rates for all service SMS report

benthorner:

This makes a few initial changes to the very similar
code we have to calculate SMS usage for organisations,
which we then duplicate for the all service report.

By the end of this PR we have three slightly different
versions of the complex windowing query to calculate SMS
usage. It should be possible to consolidate them into one
query, which we will look at in a follow-up PR.

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -67,26 +67,21 @@ def fetch_sms_free_allowance_remainder_until_date(end_date):
 def fetch_usage_for_all_services_sms(start_date, end_date):
 
     # ASSUMPTION: AnnualBilling has been populated for year.
-    allowance_left_at_start_date_query = fetch_sms_free_allowance_remainder_until_date(start_date).subquery()
+    year = get_financial_year_for_datetime(get_london_midnight_in_utc(start_date))
+    ft_billing_subquery = _fetch_usage_for_all_services_sms_query(year).subquery()
 
-    chargeable_units = func.sum(FactBilling.billable_units * FactBilling.rate_multiplier)
-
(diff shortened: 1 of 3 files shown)


leohemsted (approved):
I think this looks really good. I've gone over it several times now and can't find anything obvious to note, certainly nothing I think we need to approach before this can go in. the commits were well split up and really helped me go through this, although the "support variable rates" big commit was still tough to follow, but i'm pretty confident given our earlier improvements to the test coverage

benthorner:
Deployed to Production. I've manually verified the change works before / after:

- No diff for a report from 1st Jan to 31st March.
- Expected diff from 1st April to 15th June.

As an anonymous example, one of the line items in the diff was as follows:

```
< 72037b7c-f5be-444a-b3bb-d330f1f69f38,XYZ,c3eb8161-6df3-40d0-be07-09463b82d58d,ABC,30.96,1800,0,0,,,,,
---
> 72037b7c-f5be-444a-b3bb-d330f1f69f38,XYZ,c3eb8161-6df3-40d0-be07-09463b82d58d,ABC,112.1268,6519,0,0,,,,,
```

If we do a manual calculation on ft_billing for this service (`c3eb...`) we get this:

```
select rate, sum(billable_units * rate_multiplier) from ft_billing where service_id='c3eb8161-6df3-40d0-be07-09463b82d58d' and bst_date >= '2022-04-01' and bst_date <= '2022-06-15' and notification_type='sms' group by 1;
  rate  |  sum  
--------+-------
 0.0161 |  4719
 0.0172 | 11800
```

The service has a free allowance of 10K, hence the previous 1800 number - where we only counted billing rows for one of the rates. The sum of both numbers is consistent with the new number in the report ✅.

PR #3562, opened 2022-06-20
Remove old performance platform code

idavidmcdonald:
We no longer talk to the performance platform so there is no use
for this.

Our code for our performance dashboard (https://www.notifications.service.gov.uk/performance)
still stands.


benthorner (approved):
Nice one - I've been meaning to delete this for ages 🙌

PR #3563, opened 2022-06-20
Remove a few unused functions

idavidmcdonald:
I can't see these being used anywhere so think they are good to delete and their related tests

Review commit by commit for an easier review

### DIFF
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -15,12 +15,6 @@ def get_s3_object(bucket_name, file_location):
     return s3.Object(bucket_name, file_location)
 
 
-def head_s3_object(bucket_name, file_location):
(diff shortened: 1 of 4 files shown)


benthorner (approved):
Also can't see these being used anywhere 👍 .

Out of interest, how did you come across all of these?

idavidmcdonald:
After spotting some unused code elsewhere I wondered if there were other unused functions floating around. I started going through the files in the `app` directory and searching for whether each function in those files was referenced anywhere else in the `app` directory (aware this isn't a watertight test but then should be backed up by the unit tests failing if I missed something).

I only made it down into the `dao` directory whilst I had a spare 15 minutes so stopped at that point.

PR #3565, opened 2022-06-27
Get an email branding pool for an organisation

BlessedDev:
This PR is related to and should be deployed before: https://github.com/alphagov/notifications-admin/pull/4282

In order to support the changes made to notifications-admin, empowering users to self select email branding available in their organisation branding pool, we added a new table, `email_branding_to_organisation` (backed by an OrganisationEmailBranding model), that connects email branding to organisation with a many-to-many relationship, similarly to how `user_to_service' table works.

Relevant rest and dao methods were also added to retrieve the branding pool, and unit tests were written for them.

Review commit by commit

See https://www.pivotaltracker.com/story/show/181416165 for further details.


app/models.py
```diff
@@ -419,6 +419,12 @@ class Organisation(db.Model):
     billing_contact_email_addresses = db.Column(db.Text, nullable=True)
     billing_reference = db.Column(db.String(255), nullable=True)
 
+    email_branding_pool = db.relationship(
+        'EmailBranding',
+        secondary='email_branding_to_organisation',
+        backref='organisation'
```
idavidmcdonald, line 425: So an `EmailBranding` can belong to multiple organisations. Therefore I don't think it makes sense for the backref to be `organisation` (singular) as it could return multiple organisations.

I think this should likely mean the backref is renamed to `organisations`.

Note, I've been reading this example to see a generic example for sqlalchemy example of a many to many relationship like this https://docs.sqlalchemy.org/en/14/orm/basic_relationships.html#many-to-many

app/dao/organisation_dao.py
```diff
@@ -158,3 +158,22 @@ def dao_add_user_to_organisation(organisation_id, user_id):
 @autocommit
 def dao_remove_user_from_organisation(organisation, user):
     organisation.users.remove(user)
+
+
+@autocommit
+def dao_add_email_branding_to_organisation_pool(organisation_id, email_branding_id):
+    organisation = dao_get_organisation_by_id(organisation_id)
+    email_branding = EmailBranding.query.filter_by(id=email_branding_id).one()
+    email_branding.organisation.append(organisation)
```
idavidmcdonald, line 167: Optional: You could consider changing this line to:

`organisation.email_branding_pool.append(email_branding)`

That way you are doing what the function name says, you are adding email branding to the organisations branding pool (rather than adding the organisation to the list of orgs using that email branding)

Caveat, I haven't tested the above to check it works

tests/app/dao/test_organisation_dao.py
```diff
@@ -383,3 +385,23 @@ def test_get_organisation_by_email_address_ignores_gsi_gov_uk(notify_db_session)
 
     found_org = dao_get_organisation_by_email_address('test_gsi_address@example.gsi.gov.uk')
     assert org == found_org
+
+
+def test_add_to_and_get_email_branding_pool_for_organisation(sample_organisation):
```
idavidmcdonald, line 390: I think it would be useful to  expand this to check that if there a third email branding that exists and is in the email branding pool of organisation 2 then it does not get included in the results for organisation 1 please

tests/app/organisation/test_rest.py
```diff
@@ -995,3 +996,29 @@ def test_get_organisation_services_usage_returns_400_if_year_is_empty(admin_requ
...
+
+    response = admin_request.get(
+        'organisation.get_organisation_email_branding_pool',
+        organisation_id=sample_organisation.id,
+        _expected_status=200
+    )
+
+    assert len(response['data']) == 2
+    assert response['data'][0]['id'] == str(first_branding.id)
+    assert response['data'][0]['id'] == str(first_branding.id)
```
idavidmcdonald, line 1024: This second assertion I think should be 

```
assert response['data'][1]['id'] == str(second_branding.id)
```

idavidmcdonald (changes requested):
Nice work both. Nicely broken up into small commits too!

Well done on the many to many relationship, it took me a while of slowly reading the sqlalchemy docs to make sure I understood how to do it myself.

1. I left requests for a few minor changes but they aren't too controversial.

2. I've left one more complex question about the interplay between the default email branding and the email branding pool and making sure they are distinctly and clearly named

3. A very open question. What do you think about the phrase 'email branding pool`? I've personally not been too keen as a 'pool' as I think it is less easy to understand than 'options'. But this is quite a personal choice and I don't feel too strongly about it either. Just want to check we are happy to commit to 'email branding pool' as we probably won't ever change it.

app/models.py
```diff
@@ -468,6 +474,16 @@ def serialize_for_list(self):
         }
 
 
+class OrganisationEmailBranding(db.Model):
```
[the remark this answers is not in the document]
CrystalPea: Hi David, thanks for the suggestion! We discussed with Gozie, and decided we will rename the new class to `OrganisationEmailBrandingPool`, and leave column names and table name in Postgres as they are.

app/models.py
CrystalPea: We will give this a go, thanks, great suggestion and the docs were really helpful.

app/dao/organisation_dao.py
CrystalPea: Cool, that sounds sensible, we will see if it works alright 👍🏼

tests/app/dao/test_organisation_dao.py
CrystalPea: We will add that in, will make the test more robust for sure :)

tests/app/organisation/test_rest.py
CrystalPea: Oops, good spot :D

app/models.py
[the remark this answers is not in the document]
CrystalPea: > One note of caution with your suggestion. OrganisationEmailBrandingPool is a model representing a single row in the database table.

Hi David, I don't think this is quite right. One pool is spread across possibly many rows - one row is a connection between an organisation and one branding, but a pool is a relation between one organisation and many brandings, so it'a possibly many rows, selected on organisation id.

We originally named this model `OrganisationEmailBranding`, because for a similar table, `user_to_service`, model name is `UserService`, so wanted to be consistent with that. Maybe we could call it 'EmailBrandingToOrganisation' instead?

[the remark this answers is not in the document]
CrystalPea: After some thought I agree with you best option is option number 2:

> Rename the attribute email_branding to default_email_branding on the Organisastion model class. You'd probably want to do the same for letter_branding to default_letter_branding as well for consistency.

I will add first steps for that change to this PR.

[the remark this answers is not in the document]
BlessedDev: I also agree with going with option number 2. I have made all the other requested changes made so that the only thing left to do is to carry out the suggested changes in option 2.

app/models.py
CrystalPea: done now ☺️

app/dao/organisation_dao.py
CrystalPea: Gozie made it work, so yay done now ☺️

tests/app/dao/test_organisation_dao.py
CrystalPea: Done now (thanks @BlessedDev ! )

app/models.py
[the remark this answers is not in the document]
idavidmcdonald: Maybe we go pragmatic. Nothing will ever be perfect. Here is a suggestion for something that is fairly easy and quick to implement.

Add an inline comment to the `Organisation` class `email_branding` and `letter_branding` attributes saying these are the default brandings and not to be confused with the email branding pool. Then we can keep the class `OrganisationEmailBranding` name as it currently is. This keeps it consistent with `UserService` for example.

Thoughts?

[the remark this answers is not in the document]
CrystalPea: Sounds good, I added the comments. I didn't say anything about letter branding pool though as we don't have one yet, so that could be confusing.

PR #3566, opened 2022-06-28
Add Docker hostname to access host machine network

rjbaker:
Docker containers can access the host machine network via the `host.docker.internal` name which points to the host network gateway IP. 

This allows access to services running on the host machine from inside the container.

This appears to have stopped working on Docker for Mac v20.10.13 running on M1.

We can use the `--add-host` directive to manually add the host pointing to the host-gateway IP.

To test this I ran a local server on the host machine: `python -m http.server 8000`

Running the container without the `--add-host` directive returns:
```
root@10832ebe0e06:/home/vcap/app# curl http://host.docker.local:8000
curl: (6) Could not resolve host: host.docker.local
```

Running with the `--add-host` directive:
```
root@3fa56df19740:/home/vcap/app# curl http://host.docker.internal:8000
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
<title>Directory listing for /</title>
...
```


BlessedDev (approved):

samuelhwilliams:
@rjbaker shall we merge this or do we no longer need it

PR #3567, opened 2022-06-28
Further refactoring for billing tests

benthorner:

In response to [^1].

This pushes some of the test setup back into each of the
tests for the 4 billing report functions we're working on
at the moment. We can still use some fixtures to make the
setup more focussed e.g. sample_service vs. create_service.

In the remaining commits, I've also converted the "set_up"
functions into fixtures, to match how we write most tests.

[^1]: https://github.com/alphagov/notifications-api/pull/3559#discussion_r902390463

### DIFF
--- a/tests/app/dao/test_fact_billing_dao.py
+++ b/tests/app/dao/test_fact_billing_dao.py
@@ -70,10 +70,10 @@ def sample_service_billing_fy_2016(sample_service):
     create_annual_billing(service_id=sample_service.id, free_sms_fragment_limit=1, financial_year_start=2016)
 
 
-def set_up_yearly_data_variable_rates():
-    service = create_service()
-    sms_template = create_template(service=service, template_type="sms")
-    letter_template = create_template(service=service, template_type="letter")
+@pytest.fixture
+def sample_service_billing_fy_2018_variable_rates(sample_service):
+    sms_template = create_template(service=sample_service, template_type="sms")
+    letter_template = create_template(service=sample_service, template_type="letter")
 
     create_ft_billing(bst_date='2018-05-16', template=sms_template, rate=0.162)
     create_ft_billing(bst_date='2018-05-17', template=sms_template, rate_multiplier=2, rate=0.0150, billable_unit=2)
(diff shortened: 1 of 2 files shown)


tests/app/dao/test_fact_billing_dao.py
```diff
@@ -39,6 +39,11 @@
 )
 
 
+@pytest.fixture
+def sample_service_sms_template(sample_service):
+    return create_template(service=sample_service, template_type="sms")
```
leohemsted, line 44: So this is very convoluted and confusing to explain but i think we can just use `sample_template` here.

sample_template calls `service = create_service(service_permissions=[EMAIL_TYPE, SMS_TYPE], check_if_service_exists=True)`

`check_if_service_exists` checks if the service with the given name exists. if it does, then it returns that. (else it creates a service with that name).

we don't specify a name so it defaults to `Sample service`.

This means it will create a service called `Sample service` if the fixture hasnt been invoked (and we can access that via `sample_template.service`), and if we've already invoked `sample_service` in the test, indirectly or otherwise, it'll attach the template to that service.

TLDR:

we dont need this fixture, and can just use `sample_template`

leohemsted (commented):
Just one relatively small comment

tests/app/dao/test_fact_billing_dao.py
benthorner: Aha, nice digging! I thought about using this originally but didn't look deep enough to see the connection.

Such a subtle link isn't ideal but doesn't bother me overly and I like the idea of reuse / avoiding a new fixture. What concerns me more is the name `sample_template` doesn't clearly convey it's an SMS template, which is specifically what we need in many of the tests. So that's two confusing subtleties of using this fixture.

Defining a new fixture isn't ideal either, so I'm not sure what's the lesser evil here.

Where do you lean?

benthorner: A fair compromise - added in https://github.com/alphagov/notifications-api/pull/3567/commits/469a2e0e417918f1981b0e80471b36daa921cd7a.

leohemsted (approved):

PR #3568, opened 2022-06-29
Deploy droplets

sakisv:
Adds two scripts:

1. `scripts/deploy.sh` which uses the CF API to reuse droplets and trigger deployments
2. `scripts/create-droplet.sh` which is a helper script to run locally if you need to deploy something to preview

Best reviewed commit-by-commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,12 +87,14 @@ clean:
 preview: ## Set environment to preview
 	$(eval export DEPLOY_ENV=preview)
 	$(eval export DNS_NAME="notify.works")
+	$(eval export USE_DROPLETS="yes")
 	@true
 
 .PHONY: staging
 staging: ## Set environment to staging
 	$(eval export DEPLOY_ENV=staging)
 	$(eval export DNS_NAME="staging-notify.works")
+	$(eval export USE_DROPLETS="yes")
 	@true
 
 .PHONY: production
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -135,9 +135,7 @@ cf-deploy: ## Deploys the app to Cloud Foundry
 	# generate manifest (including secrets) and write it to CF_MANIFEST_PATH (in /tmp/)
 	make -s CF_APP=${CF_APP} generate-manifest > ${CF_MANIFEST_PATH}
 
-	# fails after 15 mins if deploy doesn't work
-	# reads manifest from CF_MANIFEST_PATH
-	CF_STARTUP_TIMEOUT=15 cf push ${CF_APP} --strategy=rolling -f ${CF_MANIFEST_PATH}
+	CF_APP=${CF_APP} CF_MANIFEST_PATH=${CF_MANIFEST_PATH} ./scripts/deploy.sh
```
rjbaker, line 138: Presumably this change will begin the switch to the new deployment method in all environments?

rjbaker (approved):
Looks good and I understand the droplet build/copy/deploy process following our conversations. One question above about rolling this out.

Makefile
sakisv: Correcty. Do you think it warrants a preview+staging rollout first?

sakisv: Addressed in 8176014

PR #3573, opened 2022-07-06
refactor billing, remove duplicate functions

leohemsted:
merging the org biling reporting and the platform admin billing report. review commit by commit


app/dao/fact_billing_dao.py
```diff
@@ -74,7 +74,7 @@ def fetch_usage_for_all_services_sms(start_date, end_date):
         Service.name
     )
 
-    return query.all()
```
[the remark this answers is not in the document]
idavidmcdonald: Optional: Also for comparison `fetch_usage_for_all_services_letter` still calls `.all` so there is a bit of inconsistency between the `fetch_usage_for_all_services_x` functions and what they return. I can tolerate this though or maybe a slight change to the function name to say it returns a query?

app/dao/organisation_dao.py
```diff
@@ -26,11 +34,22 @@ def dao_get_organisation_services(organisation_id):
     ).one().services
 
 
-def dao_get_organisation_live_services(organisation_id):
-    return Service.query.filter_by(
-        organisation_id=organisation_id,
-        restricted=False
-    ).all()
+def dao_get_organisation_live_services_with_free_allowance(organisation_id, financial_year):
```
idavidmcdonald, line 37: Optional: I initially read 'with_free_allowance' as a filter. Ie only get services which have a free allowance and exclude those that don't have any free allowance.

Not sure if it is worth changing to something like 'dao_get_organisation_live_services_and_their_free_allowance'

app/dao/fact_billing_dao.py
```diff
@@ -826,23 +829,17 @@ def fetch_usage_for_organisation(organisation_id, year):
     letter_usages = _fetch_usage_for_organisation_letter(organisation_id, year_start, year_end)
     email_usages = _fetch_usage_for_organisation_email(organisation_id, year_start, year_end)
     for usage in sms_usages:
-        service_with_usage[str(usage.service_id)] = {
-            'service_id': usage.service_id,
-            'service_name': usage.service_name,
-            'free_sms_limit': usage.free_allowance,
+        # update sms fields
+        service_with_usage[str(usage.service_id)] |= {
```
idavidmcdonald, line 799: Nice use of |= and comment to go with it

app/dao/fact_billing_dao.py
```diff
@@ -813,8 +815,9 @@ def fetch_usage_for_organisation(organisation_id, year):
         service_with_usage[str(service.id)] = {
             'service_id': service.id,
             'service_name': service.name,
-            'free_sms_limit': 0,
-            'sms_remainder': 0,
+            'free_sms_limit': service.free_sms_fragment_limit,
+            # if sms usage is 0, then remainder is equiv to the free sms fragment limit
```
idavidmcdonald, line 785: Nice comment, this is helpful

idavidmcdonald (approved):
A few optional comments but I think this is all generally OK.

Definitely a complex area so keeping this code DRY feels like a good move.

app/dao/fact_billing_dao.py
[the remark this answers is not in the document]
leohemsted: yeah, to confirm, calling `iter(query)` executes it in the same way as `.all`. It doesn't functionally make any difference, it may in fact have very marginal performance improvements or RAM usage improvements by not calling `.all` since it might copy data about a little bit less, but i wouldn't expect that to be noticeable.

i'll take a look at the usage of `_letter`. happy to keep them all consistent.

app/dao/fact_billing_dao.py
leohemsted: good to make use of some python 3.9 features!

PR #3580, opened 2022-07-25
add api_client_request fixture

leohemsted:
in admin tests we have `client_request` fixture for handling boilerplate around testing user-facing requests - logging in the user, json dumping/loading, creating URLs from endpoints and args, etc.

in api tests we have `admin_request` for doing the same for requests coming from the admin.

now in api /v2/ tests, we have `api_client_request`, for requests made by an api client to our public endpoints! The only difference between this and `admin_request` is that it _requires_ a service id (to create a jwt token for), and has an optional _api_key_type arg that can be 'team', 'test' or 'normal' (the default).

if your service has any number of api keys for the given key type, it'll use an arbitrary one of them. If there isn't one, it'll create one of that type.

if your test is simulating a well formed request to a v2 endpoint, it can now use this request to handle boilerplate. Never call `create_service_authorization_header` again!

Except, you might still need to call the old `client` fixture occasionally. Mostly when you're testing:

* What happens if you pass something in that isn't valid json
* What happens if you don't pass in proper authorization header

This PR is pretty big, and adds more lines than it takes away - but most of those new lines are just making the code layout more readable. In terms of lines of executed code this is definitely a reduction.

### DIFF
--- a/tests/app/conftest.py
+++ b/tests/app/conftest.py
@@ -49,7 +49,10 @@ from app.models import (
     Template,
     TemplateHistory,
 )
-from tests import create_admin_authorization_header
+from tests import (
+    create_admin_authorization_header,
+    create_service_authorization_header,
+)
 from tests.app.db import (
     create_api_key,
     create_email_branding,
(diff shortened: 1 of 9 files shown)


klssmith (approved):
Nice, this makes the tests much simpler! Just need to fix an import error so the tests pass

PR #3581, opened 2022-07-25
allow Document download by email only

leohemsted:
This should be merged on friday 29th, along with some documentation changes that will be coming to a client repo near you shortly

- [x] https://github.com/alphagov/notifications-tech-docs/pull/171

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -203,6 +203,8 @@ def validate_template(template_id, personalisation, service, notification_type, 
     if check_char_count:
         check_is_message_too_long(template_with_content)
 
+    check_template_can_contain_documents(notification_type, personalisation)
+
     return template, template_with_content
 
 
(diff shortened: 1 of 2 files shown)


klssmith:
This all worked as expected when I tested it out, so just needs the content in the failing tests to be updated.

You could add a conditional to only call this function for emails now, but I'm not sure if that is worth doing - it probably doesn't add much to also continue calling it for text messages  https://github.com/alphagov/notifications-api/blob/93e9a3384ec6db3671b83b0dce4fda8ee2b29983/app/v2/notifications/post_notifications.py#L191

leohemsted:
@klssmith oops good spot on the test. i briefly thought about where the check happens - im not too worried, the check includes the template_type check as the first thing, so if it is an email template it will short-cut out as that boolean fails before it needs to look at the keys

klssmith (approved):

PR #3583, opened 2022-08-01
Add endpoint to update email branding pool

klssmith:
This adds an endpoint to update the email branding pool for an
organisation. The form that the data comes from allows multiple email
brandings to be added at once, so the endpoint takes this into account.
I've added a new dao function which allows bulk adding to the email branding
pool - this saves having to iterate through each new branding and doing
a database update for each one.

As with other POST endpoints, we check the data using a JSON schema,
however the request from admin should always have the data in the
required format, so I would not expect this to raise a validation error
in production. Instead, it acts to document the endpoint and to ensure
any future changes don't break things unexpectedly.

[Pivotal story](https://www.pivotaltracker.com/story/show/182771318)

### DIFF
--- a/tests/app/dao/test_organisation_dao.py
+++ b/tests/app/dao/test_organisation_dao.py
@@ -7,6 +7,7 @@ from sqlalchemy.exc import IntegrityError, SQLAlchemyError
 
 from app import db
 from app.dao.organisation_dao import (
+    dao_add_email_branding_list_to_organisation_pool,
     dao_add_email_branding_to_organisation_pool,
     dao_add_service_to_organisation,
     dao_add_user_to_organisation,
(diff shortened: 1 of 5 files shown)


tests/app/dao/test_organisation_dao.py
```diff
@@ -420,6 +421,24 @@ def test_add_to_and_get_email_branding_pool_for_organisation(sample_organisation
...
+    branding_2 = create_email_branding(logo='test_x2.png', name='branding_2')
+    branding_3 = create_email_branding(logo='test_x3.png', name='branding_3')
+
+    dao_add_email_branding_list_to_organisation_pool(sample_organisation.id, [branding_1.id])
+
+    assert sample_organisation.email_branding_pool == [branding_1]
+
+    dao_add_email_branding_list_to_organisation_pool(sample_organisation.id, [branding_2.id, branding_3.id])
+
+    # existing brandings remain in the pool, while new ones get added
```
idavidmcdonald, line 435: Nice work, good use of code comment here

tests/app/dao/test_organisation_dao.py
```diff
@@ -420,6 +421,24 @@ def test_add_to_and_get_email_branding_pool_for_organisation(sample_organisation
     assert third_branding not in results
 
 
+def test_dao_add_email_branding_list_to_organisation_pool(sample_organisation):
```
idavidmcdonald, line 424: Optional: Should we add a test for what happens if you call this with a branding that is already in the pool? I know we generally don't expect it but could have the admin app with a page open for the first user user, a second user updates the pool, the first user submits their form

idavidmcdonald (approved):
Looks good and thank you for a nice clear commit message that explains your choice of why you are using a schema.

tests/app/dao/test_organisation_dao.py
klssmith: This would fail with an `IntegrityError` if you try to add a branding that is already in the pool. When writing this I thought that was ok, since you can't add branding that is already in the pool through admin. When you click "Submit" on the form, it does another [check of what's in the branding pool](https://github.com/alphagov/notifications-admin/pull/4306/files#diff-11eecc50c648e6b1c5c13b38769c3e8b3135e392651c134664c6de99ee0de65eR455-R459) (i.e. the allowed options) on the POST request and does not let you submit the form if with something that's already in the pool. This catches the scenario where two people are using the branding form at the same time and person 2 tries to add a brand that person 1 has just added.

(The error shown on the form is not very user friendly since it comes straight from WTForms, but it does stop you from ever adding branding that is already in a pool and would be tricky to make nicer.)

Let me know if you think that's ok or not - I had thought it was given that form is platform admin only and the chances of two people using it at the same time, let alone adding the same branding to the same pool are really, really low.

idavidmcdonald: A perfect explanation. I think that is definitely OK. Thank you for explaining for me. 

🚢

PR #3584, opened 2022-08-01
Require limited charset for `Service.email_from`

samuelhwilliams:
Some users have been trying to create services with unicode service
names. When it comes to sending emails, not all unicode characters can
be 'cleaned' to the simple ASCII charset which is currently required to
build a valid local email fragment.

Rich unicode local email names can theoretically be supported by
converting them to punycode, but we don't want to do this currently as
it will lead to very peculiar-looking emails and we haven't researched
the potential consequences of this.

For a service to currently send emails, they must put in a basic ascii
service name. Let's help the user avoid the error case by enforcing this
strictly in the first place.

We leave it to the future to research and design a long-term solution
that will perhaps allow rich Unicode service names.

Note: the notifications-admin app performs some initial 'cleaning' on
service names which will remove diacritics/accents/etc and downcast some
rich characters to plain ascii. What this means is that we aren't
totally banning _all_ unicode characters from service names, but just
those which don't downcast easily to the ascii set - eg cyrillic.

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -1,3 +1,4 @@
+import string
 from datetime import datetime, timedelta
 from uuid import UUID
 
@@ -303,6 +304,15 @@ class ServiceSchema(BaseSchema, UUIDsAsStringsMixin):
         if len(set(permissions)) != len(permissions):
             duplicates = list(set([x for x in permissions if permissions.count(x) > 1]))
             raise ValidationError('Duplicate Service Permission: {}'.format(duplicates))
+
+    @validates('email_from')
+    def validate_email_from(self, value):
+        if not all(
+            char in string.ascii_lowercase + string.digits + '.' for char in value
+        ):
+            raise ValidationError(
+                'Unacceptable characters: `email_from` may only contain letters, numbers and full stops.'
+            )
 
     @pre_load()
     def format_for_data_model(self, in_data, **kwargs):
--- a/tests/app/service/test_rest.py
+++ b/tests/app/service/test_rest.py
@@ -703,6 +703,41 @@ def test_create_service_should_throw_duplicate_key_constraint_for_existing_email
             assert "Duplicate service name '{}'".format(service_name) in json_resp['message']['name']
 
 
+@pytest.mark.parametrize(
+    'email_from, should_error',
+    (
+            ('sams.sarnies', False),  # We are happy with plain ascii alnum/full-stops.
+            ('SAMS.SARNIES', True),  # We will reject anything with uppercase characters
+            ('sam\'s.sarnies', True),  # We reject punctuation other than a full-stop `.`
+            ('sams.Бутерброды', True),  # We reject unicode outside of the ascii charset
+            ('sams.ü', True),  # Even if it could theoretically be downcast to ascii
+            ('sams.u', False),  # Like this, which would be fine
+    )
+)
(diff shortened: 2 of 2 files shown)


app/schemas.py
```diff
@@ -304,6 +305,15 @@ def validate_permissions(self, value):
             duplicates = list(set([x for x in permissions if permissions.count(x) > 1]))
             raise ValidationError('Duplicate Service Permission: {}'.format(duplicates))
 
+    @validates('email_from')
+    def validate_email_from(self, value):
+        if not all(
+            char in string.ascii_lowercase + string.digits + '.' for char in value
+        ):
```
quis, line 312: A totally optional suggestion if you think it’s more elegant:
```suggestion
        if set(value).difference(
            string.ascii_lowercase + string.digits + '.'
        ):
```

tests/app/service/test_rest.py
```diff
@@ -990,6 +1033,44 @@ def test_update_service_permissions_will_add_service_permissions(client, sample_
     assert resp.status_code == 200
     assert set(result['data']['permissions']) == set([SMS_TYPE, EMAIL_TYPE, LETTER_TYPE])
 
+@pytest.mark.parametrize(
+    'email_from, should_error',
+    (
+            ('sams.sarnies', False),
```
quis, line 1039: Worth adding a big 🥪 here to be explicit that rejecting uppercase is an expected behaviour?
```suggestion
            ('sams.sarnies', False),
            ('SAMS.SARNIES', True),            
```

app/schemas.py
samuelhwilliams: It's more elegant, though I _think_ what I wrote is maybe a touch more readable. I will leave it since you said it's optional :)

tests/app/service/test_rest.py
samuelhwilliams: Good idea :)

tests/app/service/test_rest.py
```diff
@@ -703,6 +703,50 @@ def test_create_service_should_throw_duplicate_key_constraint_for_existing_email
...
+        'user_id': str(first_service.users[0].id),
+        'message_limit': 1000,
+        'restricted': False,
+        'active': False,
+        'email_from': email_from,
+        'created_by': str(sample_user.id)
+    }
+    auth_header = create_admin_authorization_header()
+    headers = [('Content-Type', 'application/json'), auth_header]
+    resp = client.post(
```
klssmith, line 733: We generally use the `admin_request` fixture to test requests from the admin app - it does some of the work for you (e.g. adding the headers), so if you use that fixture instead of `client` in the tests you can simplify them a bit

samuelhwilliams: Thanks, will update :)

tests/app/service/test_rest.py
```diff
@@ -991,6 +1026,45 @@ def test_update_service_permissions_will_add_service_permissions(client, sample_
...
+        client, sample_service, email_from, should_error
+):
+    auth_header = create_admin_authorization_header()
+
+    data = {
+        'service_name': 'Sam\'s sarnies',
+        'email_from': email_from
+    }
+
+    resp = client.post(
```
klssmith, line 1050: This test could also use the `admin_request` fixture

samuelhwilliams: It does now :)

klssmith (approved):

PR #3587, opened 2022-08-11
Add email verification service permission

samuelhwilliams:
While we're building out the document download email verification flow
we want to make sure the features are unavailable to most services, so
we'll use a service permission to essentially feature flag it as
disabled for everyone.

This gives us the ability to gradually allow services to test the new
features if we want to do a limited roll-out initially.

Ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/182950344


CrystalPea (approved):
Looks good to me 👍🏼 .

Do we add that permission setting to admin interface as a part of the same pivotal ticket, or a separate one?

samuelhwilliams:
It will be a separate one, if we do indeed do it at all. At the moment I haven't planned to add it to the admin interface because my intention is for it to just be a developer-level feature flag, almost.

CrystalPea:
@samuelhwilliams Ahh makes sense, thanks a lot!

PR #3588, opened 2022-08-15
Rename migration file to match migration name

klssmith:
When we deploy api, we check if migrations need to run by using `make check-if-migrations-to-run`. This compares the name of the latest migration file with the database version returned by the '/_status' endpoint. If they are different, we run migrations and run the functional tests an extra time.

This changes the filename of the latest migration to match the db migration name so that we don't do unnecessary steps when deploying.


samuelhwilliams (approved):
🪄 ✨

PR #3590, opened 2022-08-15
Verify email before download parameter

samuelhwilliams:
This PR adds support for a new key to the `personalisation` field of email notifications: `verify_email_before_download`. If present this field must be True/False - and defaults to False if not present.

When added, it causes the call to document-download-api to include a `verification_email` parameter, which will later be used to store the hashed email against the document for verification as part of the user download flow.

TIcket: https://www.pivotaltracker.com/n/projects/1443052/stories/182950357

### DIFF
--- a/app/clients/document_download.py
+++ b/app/clients/document_download.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 import requests
 from flask import current_app
 
@@ -23,17 +25,22 @@ class DocumentDownloadClient:
     def get_upload_url(self, service_id):
         return "{}/services/{}/documents".format(self.api_host, service_id)
 
-    def upload_document(self, service_id, file_contents, is_csv=None):
+    def upload_document(self, service_id, file_contents, is_csv=None, verification_email: Optional[str] = None):
         try:
+            data = {
+                'document': file_contents,
+                'is_csv': is_csv or False,
+            }
+
+            if verification_email:
+                data['verification_email'] = verification_email
+
             response = requests.post(
                 self.get_upload_url(service_id),
                 headers={
                     'Authorization': "Bearer {}".format(self.auth_token),
                 },
-                json={
-                    'document': file_contents,
-                    'is_csv': is_csv or False,
-                }
+                json=data
             )
 
             response.raise_for_status()
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from flask import current_app
 from gds_metrics.metrics import Histogram
 from notifications_utils import SMS_CHAR_COUNT_LIMIT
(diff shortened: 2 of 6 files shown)


app/clients/document_download.py
```diff
@@ -23,17 +25,22 @@ def init_app(self, app):
     def get_upload_url(self, service_id):
         return "{}/services/{}/documents".format(self.api_host, service_id)
 
-    def upload_document(self, service_id, file_contents, is_csv=None):
+    def upload_document(self, service_id, file_contents, is_csv=None, verification_email: Optional[str] = None):
```
CrystalPea, line 28: what does `Optional` give us compared to if we did not use it? I haven't seen it before.
Is it to make Python more statically typed? If so, do you think it's a direction we should take overall, and if so, why? Thanks!

app/v2/notifications/post_notifications.py
```diff
@@ -334,13 +336,22 @@ def process_document_uploads(personalisation_data, service, simulated=False):
...
+    error_if_service_using_email_verification_flow_without_permission(
+        {k: personalisation_data[k] for k in file_keys},
+        service.permissions
+    )
+
     for key in file_keys:
         if simulated:
             personalisation_data[key] = document_download_client.get_upload_url(service.id) + '/test-document'
         else:
+            verify_email = personalisation_data[key].get('verify_email_before_download') or False
```
CrystalPea, line 348: could we set this variable before calling `error_if_service_using_email_verification_flow_without_permission` and then pass it, and then `error_if_service_using_email_verification_flow_without_permission` could be simpler?

app/notifications/validators.py
```diff
@@ -126,6 +127,17 @@ def check_if_service_can_send_files_by_email(service_contact_link, service_id):
         )
 
 
+def error_if_service_using_email_verification_flow_without_permission(file_upload_personalisation, service_permissions):
```
CrystalPea, line 130: Should we test this error? That a service without permission tries to send file with verification and they get the right error message in the response?

CrystalPea (commented):
Nice work, I left a few comments - one question and two suggestions. Let me know what you think ☺️

app/clients/document_download.py
samuelhwilliams: For now, it just tells us (devs) that the value can be a string or it can be None and to handle that appropriately. I'd like for us to move towards having type hints in the code base and having these statically checked with mypy, but something to discuss more broadly before committing to it.

app/notifications/validators.py
samuelhwilliams: Not sure how I missed this! Thanks

samuelhwilliams: Added 👍

app/v2/notifications/post_notifications.py
samuelhwilliams: Done 👍

app/clients/document_download.py
CrystalPea: Ok, interesting, thanks! Yeah, definitely one to discuss on a tech catch up when David and Leo are back - or maybe even we could have a tech away day to discuss tech strategy in an ambitious way?

app/v2/notifications/post_notifications.py
CrystalPea: ✨

CrystalPea (approved):
:shipit:

app/clients/document_download.py
samuelhwilliams: That's a great idea, we should totally do that 👍

PR #3593, opened 2022-08-19
Add new default email branding to pool

samuelhwilliams:
When we set the default email branding for an organisation, let's also
add it to the email branding pool for that organisation so that services
can switch out of it and back at a later date if they change their minds
without having to contact us.

Ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/182762788

### DIFF
--- a/app/dao/organisation_dao.py
+++ b/app/dao/organisation_dao.py
@@ -110,6 +110,10 @@ def dao_update_organisation(organisation_id, **kwargs):
     if 'email_branding_id' in kwargs:
         _update_organisation_services(organisation, 'email_branding')
 
+        # This could be `None` to return an organisation to the default GOV.UK branding (which isn't a real brand).
+        if kwargs['email_branding_id']:
+            dao_add_email_branding_to_organisation_pool(organisation_id, kwargs['email_branding_id'])
+
     if 'letter_branding_id' in kwargs:
         _update_organisation_services(organisation, 'letter_branding')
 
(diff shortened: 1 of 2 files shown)


app/dao/organisation_dao.py
```diff
@@ -109,6 +109,7 @@ def dao_update_organisation(organisation_id, **kwargs):
 
     if 'email_branding_id' in kwargs:
         _update_organisation_services(organisation, 'email_branding')
+        dao_add_email_branding_to_organisation_pool(organisation_id, kwargs['email_branding_id'])
```
klssmith, line 112: If the branding chosen to be the default is the GOV.UK branding, then `email_branding_id` is `None` and the dao function gives an error.

klssmith:
We need to clear `organisation-<ord_id>-email-branding-pool` from the cache when changing anything in an organisation's branding pool. When I run this locally, the org settings page does not show the right thing until the cache is cleared.

samuelhwilliams:
> We need to clear organisation-<ord_id>-email-branding-pool from the cache when changing anything in an organisation's branding pool. When I run this locally, the org settings page does not show the right thing until the cache is cleared.

Weird ... I didn't have to clear the cache when I ran it locally 🤔  But yes you're right - we should clear it anyway.

klssmith:
Caching is off by default in development for both admin and api - you need to turn it on with an environment variable in each repo https://github.com/alphagov/notifications-admin/blob/098d3e97d1d6fc00b6baa5f9df021a43bc00f496/app/config.py#L120

samuelhwilliams:
> Caching is off by default in development for both admin and api - you need to turn it on with an environment variable in each repo https://github.com/alphagov/notifications-admin/blob/098d3e97d1d6fc00b6baa5f9df021a43bc00f496/app/config.py#L120

💀  I see. It's definitely not off for _everything_ because I've had to clear the cache locally before... Bit of a 'gotcha' I guess.

klssmith (approved):

PR #3594, opened 2022-08-22
Update top level dependencies where possible

klssmith:
This updates all test dependencies and all of the top level dependencies that it was possible to update.

There is a later version of Flask available than the one used here, but it conflicted with other requirements.


CrystalPea (approved):
:shipit:

PR #3598, opened 2022-08-26
Add support for `retention_period` to API

samuelhwilliams:
Update the `POST /v2/notification` endpoint to support a
`retention_period` parameter in file upload fields of personalisation.
This just passes the value directly through to the
document-download-api. Validation on the exact value of the
retention_period will happen in document-download-api - although
notify-api does enforce it as a string.

-----

Just reusing the existing service permission toggle ("feature flag") as we expect to release these at the same time, and we will remove the permission from the code shortly afterwards so the potential confusion over the name/value won't persist.

Ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/183092708

### DIFF
--- a/tests/app/v2/notifications/test_post_notifications.py
+++ b/tests/app/v2/notifications/test_post_notifications.py
@@ -881,6 +881,9 @@ def test_post_email_notification_with_archived_reply_to_id_returns_400(
         {'is_csv': True},
         {'verify_email_before_download': False},
         {'verify_email_before_download': True},
+        {'retention_period': None},
+        {'retention_period': '1 week'},
+        {'retention_period': '4 weeks'},
     )
 )
 def test_post_notification_with_document_upload(api_client_request, notify_db_session, mocker, extra):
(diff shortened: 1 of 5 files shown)


tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1058,7 +1074,7 @@ def test_post_notification_without_document_email_verification_permission(
 
     assert (
         resp['errors'][0]['message'] ==
-        'Email verification flow for document download has not been enabled for this service.'
+        "Email verification and/or custom retention for 'send files by email' has not been enabled for this service."
```
CrystalPea, line 1077: Might be worth checking this content with Karl?

CrystalPea (approved):
Looks good, only I'd check error content with Karl before merging 👍🏼

tests/app/v2/notifications/test_post_notifications.py
samuelhwilliams: It's only technically an error that a user could ever see by trying to use the feature before we announce/release it.

It's not real content per se - will be removed when we release.

CrystalPea: Makes sense 👍🏼

PR #3604, opened 2022-09-05
Proper error for unique email branding violation

samuelhwilliams:
If we receive a POST request to create an email branding with a name
that already exists, let's return a 400 rather than 500ing on the unique
constraint violation.

### DIFF
--- a/app/email_branding/rest.py
+++ b/app/email_branding/rest.py
@@ -1,4 +1,5 @@
-from flask import Blueprint, jsonify, request
+from flask import Blueprint, current_app, jsonify, request
+from sqlalchemy.exc import IntegrityError
 
 from app.dao.email_branding_dao import (
     dao_create_email_branding,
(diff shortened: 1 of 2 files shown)


klssmith:
I've just noticed we [handle this quite differently for letter branding](https://github.com/alphagov/notifications-api/blob/81e1c3873e67f9da969ab692820e1d6e3cc6c394/app/letter_branding/letter_branding_rest.py#L22). Since the branding forms are fairly similar, it might be nice if both branding forms handled integrity errors the same way (it could be either way)? I'd find the code a little easier to work with if it was consistent, though not sure what others think - if no one else minds, then maybe it's fine

samuelhwilliams:
@klssmith good call - updated to use a similar pattern to the one you pointed out. Think the create call is the most likely one to error but yeah let's do the update as well since we're here 👍  Updated both PRs.

app/email_branding/rest.py
```diff
@@ -18,6 +19,21 @@
...
+    Handle integrity errors caused by the unique constraint
+    """
+    if 'uq_email_branding_name' in str(exc):
+        return jsonify(
+            result='error',
+            message={'name': ["An email branding with that name already exists."]}
+        ), 400
+
+    current_app.logger.exception(exc)
+    return jsonify(result='error', message="Internal server error"), 500
```
CrystalPea, line 34: Wouldn't it be better to pass the error message we get here? "Internal server error" doesn't tell us much.

CrystalPea (commented):
Nice work, one comment ☺️

app/email_branding/rest.py
samuelhwilliams: It might be, but I'm not here to fix every problem 😅  This is exactly what the letter branding integrity exception handler does - and it's not worse than existing behaviour which would just bubble up the 500.

Just looking for a small+quick incremental improvement here (since it's unplanned work) :)

klssmith (approved):

app/email_branding/rest.py
CrystalPea: ok, fair do

PR #3607, opened 2022-09-16
Populate org email branding pools

CrystalPea:
Populate org email branding pools with brands of services belonging to those orgs. Exclude services with count as live = False, as those have been added by platform admins.

I tested this by:
1. Creating a new org with empty branding pool on my local
2. creating 4 services belonging to that org:
- a live service with "branding 1"
- another live service with "branding 1"
- a test service with "branding 2"
- a live service with "branding 3" that has set `count_as_live` to `False`

The result was as expected: "branding 1" and "branding 2" were added to the org branding pool, and no errors occurred.


samuelhwilliams:
> Before merging: double check that there are no silly brands belonging to services in the scope on prod (very unlikely there would be, but still).

Is the plan to just do this with a manual SQL query selecting all the relevant services and scanning the results?

CrystalPea:
@samuelhwilliams correct, I actually just did it, I run:

```
SELECT name, text FROM email_branding WHERE id in
(SELECT email_branding_id FROM services
            JOIN service_email_branding
            ON services.id = service_id
            WHERE email_branding_id IS NOT NULL
            AND count_as_live = true);
```

There is very many of them, feel free to look through if you have some time ;). But it's extremely unlikely one of these would be silly, since the query excludes services created by us Platform Admins and only we can add new logos.

migrations/versions/0377_populate_org_brand_pools.py
```diff
@@ -0,0 +1,33 @@
...
+def upgrade():
+    op.execute(
+        textwrap.dedent(
+            """
+            INSERT INTO email_branding_to_organisation
+            (organisation_id, email_branding_id)
+            (SELECT organisation_id, email_branding_id FROM services
+            JOIN service_email_branding
+            ON services.id = service_id
+            WHERE email_branding_id IS NOT NULL
```
samuelhwilliams, line 24: I think we also need to check `organisation_id IS NOT NULL` - unless you're sure that there are no services without an organisation that have a branding in prod?

I was able to make one locally and it would break the migration with eg:

```
ERROR:  null value in column "organisation_id" violates not-null constraint
DETAIL:  Failing row contains (null, e3957ce2-ef36-4b21-b318-a817d439f852).
```

CrystalPea: Ohhh good catch, thank you!

CrystalPea: Added now 👍🏼

samuelhwilliams (approved):
Looks good 👍

PR #3608, opened 2022-09-20
When creating nhs type org, add nhs branding to its branding pool

CrystalPea:


app/organisation/rest.py
```diff
@@ -105,11 +106,24 @@ def create_organisation():
     validate(data, post_create_organisation_schema)
 
     if data["organisation_type"] in NHS_ORGANISATION_TYPES:
-        data["email_branding_id"] = current_app.config['NHS_EMAIL_BRANDING_ID']
+        organisation = _create_nhs_organisation(data)
+    else:
+        organisation = Organisation(**data)
+        dao_create_organisation(organisation)
```
quis, line 112: I appreciate you’re just adding to what’s already here, but if you’re going to refactor this  logic I think it would fit more naturally in the `dao_create_organisation` function:
https://github.com/alphagov/notifications-api/blob/44f89e9239c5d160ac8d17d4a0ba74f33b0cba04/app/dao/organisation_dao.py#L79-L81

It feels similar to the logic that’s already inside the `dao_update_organisation` function: 
https://github.com/alphagov/notifications-api/blob/44f89e9239c5d160ac8d17d4a0ba74f33b0cba04/app/dao/organisation_dao.py#L84-L120

Moving it into `dao_create_organisation` would mean that if we decided to create an organisation from somewhere else in the codebase then it would always have the correct default branding.

CrystalPea: I added a commit to change it so.

quis (approved):
Thanks for changing, looks good now 👍🏻

PR #3611, opened 2022-09-28
Add scheduled nightly task to purge old events

samuelhwilliams:
Events are used to track certain actions users might take (eg logging in), and contain PII. Because of this we shouldn't keep this data forever. We've agreed a 1-year retention period and should regularly purge any data older than this.

Also needs https://github.com/alphagov/notifications-credentials/pull/295

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -18,6 +18,7 @@ from app.celery.tasks import (
     process_row,
 )
 from app.config import QueueNames, TaskNames
+from app.cronitor import cronitor
 from app.dao.invited_org_user_dao import (
     delete_org_invitations_created_more_than_two_days_ago,
 )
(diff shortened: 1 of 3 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -367,3 +369,17 @@ def remove_yesterdays_planned_tests_on_govuk_alerts():
...
+
+@notify_celery.task(name='delete-old-records-from-events-table')
+@cronitor('delete-old-records-from-events-table')
+def delete_old_records_from_events_table():
+    delete_events_before = datetime.utcnow() - timedelta(weeks=52)
+    event_query = Event.query.filter(Event.created_at < delete_events_before)
+
+    current_app.logger.info(
+        f"Preparing to delete {event_query.count()} historical events from before {delete_events_before}."
+    )
```
leohemsted, line 382: `delete()` returns the row count so we can save the extra query cycle with something like

```suggestion
    deleted_count = event_query.delete()
    current_app.logger.info(
        f"Deleted {deleted_count} historical events from before {delete_events_before}."
    )
```

leohemsted (commented):
looks good, just one minor tweak

PR #3618, opened 2022-10-05
Add validation for send-a-file upload dicts

samuelhwilliams:
If a user sends the wrong kind of data/value for the new send-a-file/document download security features, let's provide a user-friendly error message.

I would've ideally liked to do this as part of the generic incoming notification JSON schema, but in order to do that we'd need to use a `oneOf` subschema to allow for multiple data types and different validation depending on the vlaue of personalisation fields. When we do this, we lose the ability to (cleanly) provide explicit error messages about what exactly is wrong with the input, because JSON Schema wraps the root error message in some contextual one about all of the sub-schemas failing to validate.

So we've decided to just do the validation on known-file upload personalisation, which allows us to more cleanly present the error to the user.

### DIFF
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -1,4 +1,5 @@
 import json
+import re
 from datetime import datetime, timedelta
 from uuid import UUID
 
@@ -56,6 +57,32 @@ def validate_schema_date_with_hour(instance):
             raise ValidationError("datetime format is invalid. It must be a valid ISO8601 date time format, "
                                   "https://en.wikipedia.org/wiki/ISO_8601")
     return True
+
+
+@format_checker.checks('send_a_file_retention_period', raises=ValidationError)
+def validate_schema_retention_period(instance):
+    if instance is None:
+        return True
+
+    if isinstance(instance, str):
+        period = instance.strip().lower()
+        match = re.match(r'^(\d+) weeks?$', period)
+        if match and 1 <= int(match.group(1)) <= 78:
+            return True
+
+    raise ValidationError(
+        f"Unsupported value for retention_period: {instance}. Supported periods are from 1 to 78 weeks."
+    )
+
+
+@format_checker.checks('send_a_file_confirm_email', raises=ValidationError)
+def validate_schema_send_a_file_confirm_email(instance):
+    if instance is None or isinstance(instance, bool):
+        return True
+
+    raise ValidationError(
+        f"Unsupported value for confirm_email_before_download: {instance}. Use a boolean true or false value."
+    )
 
 
 @format_checker.checks('datetime', raises=ValidationError)
--- a/app/schema_validation/__init__.py
+++ b/app/schema_validation/__init__.py
@@ -75,8 +75,18 @@ def validate_schema_retention_period(instance):
     )
 
 
-@format_checker.checks('send_a_file_confirm_email', raises=ValidationError)
-def validate_schema_send_a_file_confirm_email(instance):
+@format_checker.checks('send_a_file_is_csv', raises=ValidationError)
+def send_a_file_is_csv(instance):
+    if instance is None or isinstance(instance, bool):
+        return True
+
+    raise ValidationError(
+        f"Unsupported value for is_csv: {instance}. Use a boolean true or false value."
+    )
+
+
+@format_checker.checks('send_a_file_confirm_email_before_download', raises=ValidationError)
+def send_a_file_confirm_email_before_download(instance):
     if instance is None or isinstance(instance, bool):
         return True
 
--- a/app/v2/notifications/notification_schemas.py
+++ b/app/v2/notifications/notification_schemas.py
@@ -122,7 +122,29 @@ get_notifications_response = {
     "definitions": {
         "notification": get_notification_response
     },
+}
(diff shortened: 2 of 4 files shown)


app/schema_validation/__init__.py
```diff
@@ -58,6 +59,32 @@ def validate_schema_date_with_hour(instance):
...
+        return True
+
+    if isinstance(instance, str):
+        period = instance.strip().lower()
+        match = re.match(r'^(\d+) weeks?$', period)
+        if match and 1 <= int(match.group(1)) <= 78:
+            return True
+
+    raise ValidationError(
+        f"Unsupported value for retention_period: {instance}. Supported periods are from 1 to 78 weeks."
```
leohemsted, line 74: bit late to the party now that we've already merged a bunch of docs changes, soz. one for @karlchillmaid but i'm not 100% sold on this error message as a general case.

i think when a user inputs `"99 weeks"` this is a good error message to return. 

However, I don't think it (or the API docs) are explicit enough on how to fix the input if a user enters other incorrect values.

things a user might foreseeably and reasonably try the following and expect them to work:

* `4` (the number, not a string)
* `"4"` (as a string but with no date unit)
* `"28 days"` or `"1 month"` (a string but the wrong date unit)

i guess i'm worried if a user enters one of those, we're telling them what the supported _date range_ is but not what the supported _format_ is. Thoughts?

app/schema_validation/__init__.py
```diff
@@ -58,6 +59,32 @@ def validate_schema_date_with_hour(instance):
...
+    )
+
+
+@format_checker.checks('send_a_file_confirm_email', raises=ValidationError)
+def validate_schema_send_a_file_confirm_email(instance):
+    if instance is None or isinstance(instance, bool):
+        return True
+
+    raise ValidationError(
+        f"Unsupported value for confirm_email_before_download: {instance}. Use a boolean true or false value."
```
leohemsted, line 84: do we need a custom error message for this? we don't have one for `is_csv` for example.

The default jsonschema error message is `'confirm_email_before_download potato is not of type boolean, null'`, which I think is probably fine?

If it's not fine, we might want to extend this validation/error message to apply to `is_csv` too.

app/v2/notifications/notification_schemas.py
```diff
@@ -122,7 +122,29 @@
...
+        "file": {
+            "type": "string",
+            "required": True,
+        },
+        "is_csv": {
+            "type": ["boolean", "null"],
+            "required": False,
+        },
+        "confirm_email_before_download": {
+            "format": "send_a_file_confirm_email",
```
leohemsted, line 140:
```suggestion
            "format": "send_a_file_confirm_email",
            "type": ["boolean", "null"],
```

Do we want to use jsonschema's type checking here? We already do with `is_csv`. (Depending on what we decide to do with the error messages at least)

app/v2/notifications/notification_schemas.py
```diff
@@ -122,7 +122,29 @@
...
+            "required": False,
+        },
+        "retention_period": {
+            "format": "send_a_file_retention_period",
```
leohemsted, line 144:
```suggestion
            "format": "send_a_file_retention_period",
            "type": ["string", "null"],
```

this could be a useful step to nudge users towards a string if they enter a number

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -500,6 +500,90 @@ def test_post_email_notification_returns_201(
     assert mocked.called
 
 
+@pytest.mark.parametrize(
```
leohemsted, line 503: love the comprehensive tests ✅

app/v2/notifications/notification_schemas.py
samuelhwilliams: No because then you lose control over the error message. But this comment is probably overlapping upon the conversation above so let's keep it there 👍

app/v2/notifications/notification_schemas.py
samuelhwilliams: As above.

app/schema_validation/__init__.py
samuelhwilliams: Just trying to go on what @karlchillmaid's come up with. I think this is opening up a wider issue around error messages for invalid inputs and this _probably_ isn't the time to have that wider discussion, though if it kicks that off I wouldn't be mad. I'm pretty sure with our current use of json schema we have a mix of content-designed error messaged and default JSON Schema ones.

app/schema_validation/__init__.py
samuelhwilliams: I broadly agree - I suggested one that I thought indicated the input values more clearly but didn't push back when Karl suggested alternatives here: https://github.com/alphagov/notifications-tech-docs/pull/172#discussion_r958611629

samuelhwilliams: Maybe we can take these messages and iterate if we hear from users that they're not clear enough?

leohemsted: Cool, lets agree to look at logs in a week or two and see what people have been trying to send in 👍

leohemsted (approved):

PR #3626, opened 2022-10-19
Add created_at/by, updated_at/by columns to EmailBranding

samuelhwilliams:
Now that we're working to make it possible for users to create their own email branding, let's keep a record of who's created or updated the branding and when in case we ever need to look back and see what's happened.

We also want to emit audit events for changes, but for now we'll do that from the admin app - as that's consistent with our current process. We may look to centralise this back to the API in the future to provide consistency and centralise that responsibility, but not today.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -240,6 +240,10 @@ class EmailBranding(db.Model):
     brand_type = db.Column(
         db.String(255), db.ForeignKey("branding_type.name"), index=True, nullable=False, default=BRANDING_ORG
     )
+    created_at = db.Column(db.DateTime, nullable=True, default=lambda: datetime.datetime.utcnow())
+    created_by = db.Column(UUID(as_uuid=True), db.ForeignKey("users.id"), nullable=True)
+    updated_at = db.Column(db.DateTime, nullable=True, onupdate=lambda: datetime.datetime.utcnow())
+    updated_by = db.Column(UUID(as_uuid=True), db.ForeignKey("users.id"), nullable=True)
 
     def serialize(self):
         serialized = {
--- a/tests/app/email_branding/test_rest.py
+++ b/tests/app/email_branding/test_rest.py
@@ -178,16 +178,23 @@ def test_post_update_email_branding_updates_field(admin_request, notify_db_sessi
     assert email_branding[0].text == email_branding[0].name
 
 
-@freezegun.freeze_time()
-def test_post_update_email_branding_updated_fields(admin_request, notify_db_session, sample_user):
+@freezegun.freeze_time(as_kwarg="frozen_time")
+def test_post_update_email_branding_updated_fields(admin_request, notify_db_session, sample_user, **kwargs):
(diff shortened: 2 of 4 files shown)


tests/app/email_branding/test_rest.py
```diff
@@ -149,6 +178,34 @@ def test_post_update_email_branding_updates_field(admin_request, notify_db_sessi
...
+        _data=update_data,
+        email_branding_id=email_branding_id,
+    )
+
+    email_brandings = EmailBranding.query.all()
+    email_branding = email_brandings[0]
+    assert len(email_brandings) == 1
+    assert str(email_branding.id) == email_branding_id
+    assert str(email_branding.updated_by) == update_data["updated_by"]
+    assert email_branding.updated_at == datetime.datetime.utcnow()
```
CrystalPea, line 206: should we maybe set a different time for create and update operations, and then assert that created_at was not updated when doing the update operation?

CrystalPea (commented):
Good work, and so speedy 👏🏼 . 

One suggestion for extending a test a little ☺️

tests/app/email_branding/test_rest.py
samuelhwilliams: Good call - pushed a commit adding this check :)

tests/app/email_branding/test_rest.py
```diff
@@ -149,6 +178,40 @@ def test_post_update_email_branding_updates_field(admin_request, notify_db_sessi
...
+    response = admin_request.post("email_branding.create_email_branding", _data=data, _expected_status=201)
+
+    email_brandings = EmailBranding.query.all()
+    email_branding = email_brandings[0]
+    assert len(email_brandings) == 1
+    assert email_branding.created_at == start_time
+    assert email_branding.updated_by is None
+    assert email_branding.updated_at is None
+
+    frozen_time.tick()
```
CrystalPea, line 196: ohh, nice, didn't know that one 😻

CrystalPea (approved):
Neat :shipit:

app/models.py
```diff
@@ -240,6 +240,10 @@ class EmailBranding(db.Model):
     brand_type = db.Column(
         db.String(255), db.ForeignKey("branding_type.name"), index=True, nullable=False, default=BRANDING_ORG
     )
+    created_at = db.Column(db.DateTime, nullable=True, default=lambda: datetime.datetime.utcnow())
+    created_by = db.Column(UUID(as_uuid=True), db.ForeignKey("users.id"), nullable=True)
+    updated_at = db.Column(db.DateTime, nullable=True, onupdate=lambda: datetime.datetime.utcnow())
```
klssmith, line 245: Optional, but we use a slightly different format for `default` and `onupdate` model values elsewhere in this file - we just pass the function name as an argument without needing a lambda (`default=datetime.datetime.utcnow`  or `onupdate=datetime.datetime.utcnow`). We could stick to that for consistency or to avoid confusion, but both ways of doing things work.

PR #3627, opened 2022-10-20
Bump notifications-utils to 57.1.1

samuelhwilliams:
This pulls in the updated URL escaping that prevents pre-escaped %-encoded URL values ending up rendered as their unescaped characters (eg a URL sent to us as `/blah?redirect=%2Flogin` was ending up in user inboxes as `/blah?redirect=/login` - which can be semantically different to the target server).

Would like to wait 24-48 hours before merging this to give the initial reporter time to acknowledge that this change is going in.

- [x] https://github.com/alphagov/notifications-utils/pull/994

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -29,7 +29,7 @@ notifications-python-client==6.3.0
 # PaaS
 awscli-cwlogs==1.4.6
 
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
This is ready for review - just leaving in draft so that I don't merge it immediately :)

leohemsted (approved):

PR #3629, opened 2022-10-28
Add endpoint to update letter branding pool

klssmith:
This adds an endpoint and new dao function to update the letter branding
pool for an organisation. The form that the data comes from allows multiple
letter brandings to be added at once, so the endpoint and dao function take
this into account by allowing the new brandings to be added in bulk.

As with other POST endpoints, we check the data using a JSON schema,
however the request from admin should always have the data in the
required format, so this acts more as documentation than a check that
might fail.

### DIFF
--- a/app/dao/organisation_dao.py
+++ b/app/dao/organisation_dao.py
@@ -5,6 +5,7 @@ from sqlalchemy.sql.expression import func
 from app import db
 from app.dao.dao_utils import VersionOptions, autocommit, version_class
 from app.dao.email_branding_dao import dao_get_email_branding_by_id
+from app.dao.letter_branding_dao import dao_get_letter_branding_by_id
 from app.models import (
     NHS_ORGANISATION_TYPES,
     AnnualBilling,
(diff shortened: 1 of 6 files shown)


leohemsted (approved):
Fundamentally I'm fine with this as stands. Just one small q.

If we try and add a duplicate branding, I think this implementation will hit a primary key constraint and return a 500. 

Should we return a 200 (and silently discard the duplicate), 400, or 500? I can see arguments for all three, though I'd lean towards a 200.

klssmith:
If we're going to silently ignore brandings that are already there, then maybe it could still return a `204`? That way we can keep the same status code for the case where someone adds a few brands and only one is already in the pool as well as someone only adding a brand which is in the pool.

leohemsted:
my mistake - i tested this locally and it works, which i didn't expect. sqlalchemy just handles this magically when you commit, which i didn't expect.

204 status code is fine, i was more interested in 2xx vs 4xx vs 5xx but 204 no body is fine.

```python
# setup
dao_add_letter_branding_list_to_organisation_pool(org, [1, 2])
assert org.letter_branding_pool == [1, 2]

dao_add_letter_branding_list_to_organisation_pool(org, [2, 3])

# sqlalchemy just makes this work but i didnt expect it to!
assert org.letter_branding_pool == [1, 2, 3]
```

i think it'd be nice to have a test confirming this to avoid confusion but otherwise this pr is good :+1:

klssmith:
👍 I've added a test to match the example you gave

PR #3630, opened 2022-11-01
Add endpoint and dao function to remove letter branding from pool

klssmith:
This adds a new endpoint to remove a particular letter brand from the letter branding pool for an organisation. As with the email branding equivalent, there is a check that you're not removing the default and that the brand you're trying to remove is in the pool. However, we do also do the same checks in admin, so do not expect these errors to be raised.

### DIFF
--- a/app/dao/organisation_dao.py
+++ b/app/dao/organisation_dao.py
@@ -246,3 +246,13 @@ def dao_add_letter_branding_list_to_organisation_pool(organisation_id, letter_br
     organisation.letter_branding_pool.extend(letter_brandings)
 
     db.session.add(organisation)
+
+
+@autocommit
+def dao_remove_letter_branding_from_organisation_pool(organisation_id, letter_branding_id):
+    organisation = dao_get_organisation_by_id(organisation_id)
+    letter_branding = dao_get_letter_branding_by_id(letter_branding_id)
+
+    organisation.letter_branding_pool.remove(letter_branding)
+
+    return letter_branding
--- a/app/organisation/rest.py
+++ b/app/organisation/rest.py
@@ -21,6 +21,7 @@ from app.dao.organisation_dao import (
     dao_get_organisations,
     dao_get_users_for_organisation,
     dao_remove_email_branding_from_organisation_pool,
+    dao_remove_letter_branding_from_organisation_pool,
     dao_remove_user_from_organisation,
     dao_update_organisation,
 )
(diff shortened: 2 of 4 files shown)


app/organisation/rest.py
```diff
@@ -262,6 +263,24 @@ def update_organisation_letter_branding_pool(organisation_id):
...
+    organisation = dao_get_organisation_by_id(organisation_id)
+    letter_branding_ids = {branding.id for branding in organisation.letter_branding_pool}
+
+    if letter_branding_id not in letter_branding_ids:
+        raise InvalidRequest(f"Letter branding {letter_branding_id} not in {organisation.name}'s pool", status_code=404)
+
+    if organisation.letter_branding_id == letter_branding_id:
+        raise InvalidRequest("You cannot remove an organisation's default letter branding", status_code=400)
+
+    dao_remove_letter_branding_from_organisation_pool(organisation_id, letter_branding_id)
```
samuelhwilliams, line 279: Some thoughts only, nothing to do here:

I think I need to understand better what these dao functions are for and how we want to be using them. We call `dao_get_organisation_by_id` at the start of this view, but then pass the organisation_id into this function which immediately calls `dao_get_organisation_by_id` again. And `dao_get_organisation_by_id` returns a SQLAlchemy model rather than plain data, so we're not hiding much away and still providing something that has direct access to the DB. Which means that on line 271 when we access `organisation.letter_branding_pool`, we are making more DB calls (perhaps unexpectedly).

Maybe one to discuss in the next tech chat ... as I think our abstraction layer here might not be hiding as much as we would hope.

app/dao/organisation_dao.py
```diff
@@ -246,3 +246,13 @@ def dao_add_letter_branding_list_to_organisation_pool(organisation_id, letter_br
     organisation.letter_branding_pool.extend(letter_brandings)
 
     db.session.add(organisation)
+
+
+@autocommit
```
samuelhwilliams, line 251: Does this immediately cause a commit to happen? Or are all of our requests wrapped in outer transactions that mean errors later in the view would stop this being committed?

samuelhwilliams (approved):
The view works as expected 👍 

I have some comments/thoughts around the DAO stuff but nothing really related to this PR :)

Nice test coverage 👏

app/dao/organisation_dao.py
klssmith: It will cause a [commit to happen (or a rollback if there is an error)](https://github.com/alphagov/notifications-api/blob/b2aeed613d1185c146414b75e2d8fce97f527ddb/app/dao/dao_utils.py#L9) when the dao function has finished executing all its code.

samuelhwilliams: 🤔 🤔 🤔

tests/app/dao/test_organisation_dao.py
```diff
@@ -532,3 +533,26 @@ def test_dao_add_letter_branding_list_to_organisation_pool(sample_organisation):
...
+
+
+def test_dao_remove_letter_branding_from_organisation_pool(sample_organisation):
+    branding_1 = create_letter_branding("branding_1", "filename_1")
+    branding_2 = create_letter_branding("branding_2", "filename_2")
+    branding_3 = create_letter_branding("branding_3", "filename_3")
+
+    dao_add_letter_branding_list_to_organisation_pool(
+        sample_organisation.id, [branding_1.id, branding_2.id, branding_3.id]
+    )
```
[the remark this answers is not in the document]
klssmith: This is only really being used as a helper method in the test, which I reckoned was fine - it's something we do a lot in the dao tests. But I could perhaps add an assertion after it to check that the pool contents are as expected before we remove any branding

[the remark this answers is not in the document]
samuelhwilliams: It's probably not necessary, don't worry

app/organisation/rest.py
klssmith: I agree, that sounds like a good topic to discuss at the tech catch up

PR #3633, opened 2022-11-09
Log iat/epoch time and cf request id

samuelhwilliams:
This will hopefully let us get some more useful information to share with AWS about why some requests to the API seem to be reaching/transitting CloudFront slowly.

Logs something like:

Error: Your system clock must be accurate to within 30 seconds(you: 1667977240, us: 1667977272) [X-Amz-Cf-Id: oMuw1qhJrBzOpJcRhhavISwB950YwYTubdTAVGuX2iBPWUNmBMqN0w==]

We can then share the cf-id with AWS.

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -1,3 +1,4 @@
+import time
 import uuid
 
 from flask import current_app, g, request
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):
Give it a go.

A quick review so we can get this out but I couldn't find evidence that `request.headers.get('x-amz-cf-id')` will return us something (but at least the use of `get` will return `None` if needed). Lets see what happens in prod!

samuelhwilliams:
I tested this manually in preview earlier:

image

PR #3636, opened 2022-11-15
Give NHS letter branding the same ID in each env and give NHS orgs NHS branding when created / updated

klssmith:
**Give NHS letter branding the same ID in each environment**
We gave the NHS email branding the same ID in each environment, which simplifies the code to give it to organisations and to use it on forms.

This does the same for the NHS letter branding ID. The logo exists in all environments. We delete the row in the `letter_branding` table from all environments apart from production, then re-insert it to the other environments with the ID that exists in production.

**Add NHS letter branding to pools and org when creating an NHS org**
When an organisation is created with one of the NHS org types, we now set its letter branding to the NHS brand and add the NHS brand to its branding pool.

We are already setting up NHS branding for services which are NHS services, so this change won't really affect people's experience - it just makes sure that the letter branding pools are set up with the right data.

**When changing an org to NHS type give it NHS letter branding and add the branding to the pool**
When updating an organisation to be an NHS org type, we set its letter branding to the NHS brand if it has no current letter branding.

We also add the branding to the pool. This behaviour is not NHS specific - we add the letter branding for any organisation to the branding pool if it's being made the default for that organisation.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -181,6 +181,7 @@ class Config(object):
     NOTIFY_INTERNATIONAL_SMS_SENDER = "07984404008"
     LETTERS_VOLUME_EMAIL_TEMPLATE_ID = "11fad854-fd38-4a7c-bd17-805fb13dfc12"
     NHS_EMAIL_BRANDING_ID = "a7dc4e56-660b-4db7-8cff-12c37b12b5ea"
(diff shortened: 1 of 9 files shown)


samuelhwilliams (approved):
Looks good, tested locally and seems to work as expected too 👍

PR #3638, opened 2022-11-16
Add webauthn_credential.logged_in_at

leohemsted:
If the admin passes through a credential id when logging a user in (which it will once https://github.com/alphagov/notifications-admin/pull/4452 is merged), then set `logged_in_at` to be the current timestamp in the database.


samuelhwilliams (approved):
Nice 👍  Tested locally, all looks good

PR #3640, opened 2022-11-17
write the current alembic head to file when running bootstrap

leohemsted:
this new file isn't used in our code anywhere, however, if you have a migration in a PR and someone else merges a different migration, this'll mean they also update this same file, thus you'll get a merge conflict and will be forced to rebase and fix the migration

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,0 +1,1 @@
+0383_webauthn_cred_logged_in_at
--- a/migrations/env.py
+++ b/migrations/env.py
@@ -1,9 +1,9 @@
-from __future__ import with_statement
-
 from logging.config import fileConfig
 
 from alembic import context
 from sqlalchemy import engine_from_config, pool
+
+from pathlib import Path
 
 # this is the Alembic Config object, which provides
 # access to the values within the .ini file in use.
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -26,6 +26,8 @@ bootstrap: generate-version-file ## Set up everything to run the app
 	pip3 install -r requirements_for_test.txt
 	createdb notification_api || true
 	(. environment.sh && flask db upgrade) || true
+	@# write the current version to a file so that mismatched migrations cannot be merged without conflicts
+	(. environment.sh && flask db current | grep -iv "logging configured" ) > ./migrations/.current-alembic-head || true
```
samuelhwilliams, line 30: This file doesn't get written out in CI so do we rely on devs running `make bootstrap` and hitting this issue locally? I'm not sure that's super reliable. Would be good if we could get this to run in CI as well.

leohemsted: i figured that if you didn't run `make bootstrap` then your migration won't have even run

leohemsted: the problem is, that when master updates, no CI actually runs. So we need the code change to have happened on the dev laptop before pushing. One other option that might be less brittle would be to write a pre-commit hook that writes this file if there have been changes within `./migration`? but i've no idea how to write a pre-commit hook

samuelhwilliams: oh I just run `flask db upgrade` manually lol, guess I could switch to using `make bootstrap` but.

This could be fixed by setting `require branch up-to-date` in github settings but maybe that's disabled because people found it too annoying?

leohemsted: we tried that briefly and then disabled it cos it's annoying, especially when you're trying to squeeze in quick content changes alongside someone elses deploy etc etc

samuelhwilliams (approved):
nice

PR #3643, opened 2022-11-18
Add default letter branding for orgs to branding pools

klssmith:
This adds the default letter branding for each organisation to the letter branding pool. We have already added the NHS branding for NHS orgs in a previous migration. In future work, we will look at adding brandings for services to the pool too.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0383_webauthn_cred_logged_in_at
+0385_letter_branding_pools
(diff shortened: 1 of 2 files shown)


leohemsted:
if you run this migration locally you should get the `migrations/.current-alembic-head` file modified - you should add that to your git commit too (to prevent broken builds down the line with other people who also add migrations)

klssmith:
That file's now added (I have tested this out locally, but it was before I'd rebased and got that migration check)

leohemsted (approved):

PR #3644, opened 2022-11-18
Email branding add alt text

leohemsted:
we set the text to equal the name if it's not passed in from the admin app. at the moment we always pass in, but just an empty string. really, we'd much rather just pass in null and set it to null.

also, add alt-text. it's nullable for now but we'll start setting it on the admin app and passing it through (and eventually we'll migrate old values and make it non-nullable.

question for reviewer: should the logic for deciding the name based on alt text be written in the admin app or here?

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0385_letter_branding_pools
+0386_email_branding_alt_text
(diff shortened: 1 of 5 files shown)


migrations/.current-alembic-head
```diff
@@ -1 +1 @@
-0385_letter_branding_pools
\ No newline at end of file
+0386_email_branding_alt_text
```
samuelhwilliams, line 1: 🎉

samuelhwilliams (approved):

PR #3645, opened 2022-11-21
Set and return email branding alt text

leohemsted:


app/email_branding/email_branding_schema.py
```diff
@@ -7,6 +7,7 @@
     "properties": {
         "colour": {"type": ["string", "null"]},
         "name": {"type": "string"},
+        "alt_text": {"type": "string"},
```
klssmith, line 10: It feels like this could match the other nullable database fields and be `{"type": ["string", "null"]}`, but I think this is fine if we're never planning on passing None through from admin

klssmith (approved):

app/email_branding/email_branding_schema.py
leohemsted: helpfully, this now _does_ need to be `null`. i'll update this PR

leohemsted: commits squashed.

PR #3647, opened 2022-11-22
Generate daily report for new email brands

samuelhwilliams:
## Requires
 * [x] https://github.com/alphagov/notifications-utils/pull/1004

## Summary

Every (week)day, generate a report and log it into Zendesk listing all of the new email brands created since the last weekday so that they can be manually reviewed for quality. We may want to tweak them or remove them from an organisation's branding pool if it doesn't make sense for the brand to be more widely available.

## Ticket
image

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -1,5 +1,8 @@
+import textwrap
+from collections import defaultdict
 from datetime import datetime, timedelta
 
+import jinja2
 from flask import current_app
 from notifications_utils.clients.zendesk.zendesk_client import (
     NotifySupportTicket,
@@ -56,8 +59,10 @@ from app.models import (
     SMS_TYPE,
     BroadcastMessage,
     BroadcastStatusType,
+    EmailBranding,
     Event,
     Job,
+    Organisation,
 )
 from app.notifications.process_notifications import send_notification_to_queue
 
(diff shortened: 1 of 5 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -374,3 +379,70 @@ def delete_old_records_from_events_table():
...
+
+        else:
+            for organisation in new_brand.organisations:
+                brands_by_organisation[organisation].append(new_brand)
+
+    template = jinja2.Template(
+        textwrap.dedent(
+            """
+            <p>There are new email brands to review since {{ yesterday }}.</p><hr>
+            {%- if brands_by_organisation -%}
```
klssmith, line 415: Really minor, but this `if` statement could be removed since if `brands_by_organisation` is empty the `for` loop won't run

app/celery/scheduled_tasks.py
```diff
@@ -374,3 +379,70 @@ def delete_old_records_from_events_table():
...
+                {%- for brand in brands_with_no_organisation -%}
+                    <li><a href="https://www.{{ domain }}/email-branding/{{ brand.id }}/edit">{{ brand.name }}</a></li>
+                {%- endfor -%}
+                </ul>
+            {%- endif -%}
+            """
+        ).strip()
+    )
+    message = template.render(
+        domain=current_app.config["NOTIFY_EMAIL_DOMAIN"],
```
klssmith, line 435: `current_app.config["ADMIN_BASE_URL"]` could be used here

app/celery/scheduled_tasks.py
```diff
@@ -374,3 +379,70 @@ def delete_old_records_from_events_table():
...
+        yesterday=previous_weekday.strftime("%A %-d %B %Y"),
+        brands_by_organisation=brands_by_organisation,
+        brands_with_no_organisation=brands_with_no_organisation,
+    )
+
+    ticket = NotifySupportTicket(
+        subject="Review new email brandings",
+        message=message,
+        ticket_type=NotifySupportTicket.TYPE_TASK,
```
klssmith, line 444: We only use the task ticket type manually when we want to create a reminder for someone on support since they give you a drop down date box with an alert. That probably wouldn't be useful here, so we could stick to the ticket type of `NotifySupportTicket.TYPE_INCIDENT`, which is what we mostly use

app/celery/scheduled_tasks.py
samuelhwilliams: Oof much better call, thanks!

app/celery/scheduled_tasks.py
samuelhwilliams: That's a fair suggestion, will update 👍

PR #3648, opened 2022-11-23
Fully populate letter branding pools

klssmith:
Previous migrations partially populated the letter branding pools by
- adding NHS branding to the pool for NHS organisations
- adding the default brand for each service to the pool

This migration adds every letter branding being used by a service to the pool for that service's organisation. We'll need to manually check the letter branding pool contents after this to remove any brands we don't think should be there.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0387_migrate_alt_text
+0388_populate_letter_branding
--- a/migrations/versions/0388_populate_letter_branding.py
+++ b/migrations/versions/0388_populate_letter_branding.py
@@ -1,0 +1,31 @@
+"""
+
+Revision ID: 0388_populate_letter_branding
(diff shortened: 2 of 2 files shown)


migrations/versions/0387_populate_letter_branding.py
```diff
@@ -0,0 +1,31 @@
...
+revision = '0387_populate_letter_branding'
+down_revision = '0386_email_branding_alt_text'
+
+
+def upgrade():
+    op.execute(
+        """
+        INSERT INTO letter_branding_to_organisation
+        (organisation_id, letter_branding_id)
+        (SELECT organisation_id, letter_branding_id FROM services
```
samuelhwilliams, line 19: Inconsequential for the number of rows we're dealing with, but `SELECT DISTINCT` could've been nice here.

samuelhwilliams (approved):
Looks good/runs fine 👍

PR #3650, opened 2022-11-23
Generate daily report for new email brands

samuelhwilliams:
Re-creation of https://github.com/alphagov/notifications-api/pull/3647

Now with an added test to detect/check that HTMLEmailTemplate can be handled for services with an email brand.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -1,5 +1,8 @@
+import textwrap
+from collections import defaultdict
 from datetime import datetime, timedelta
 
+import jinja2
 from flask import current_app
 from notifications_utils.clients.zendesk.zendesk_client import (
     NotifySupportTicket,
@@ -56,8 +59,10 @@ from app.models import (
     SMS_TYPE,
     BroadcastMessage,
     BroadcastStatusType,
+    EmailBranding,
     Event,
     Job,
+    Organisation,
 )
 from app.notifications.process_notifications import send_notification_to_queue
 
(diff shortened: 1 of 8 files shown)


samuelhwilliams:
This has sort-of already been reviewed by @klssmith (thank you) but there's a new test for the weird edge case with the new HTMLEmailTemplate now

app/celery/scheduled_tasks.py
```diff
@@ -375,3 +380,68 @@ def delete_old_records_from_events_table():
...
+        domain=current_app.config["ADMIN_BASE_URL"],
+        yesterday=previous_weekday.strftime("%A %-d %B %Y"),
+        brands_by_organisation=brands_by_organisation,
+        brands_with_no_organisation=brands_with_no_organisation,
+    )
+
+    ticket = NotifySupportTicket(
+        subject="Review new email brandings",
+        message=message,
+        ticket_type=NotifySupportTicket.TYPE_INCIDENT,
```
quis, line 443: Is `TYPE_INCIDENT` appropriate here? Might [`TYPE_TASK`](https://github.com/alphagov/notifications-utils/blob/main/notifications_utils/clients/zendesk/zendesk_client.py#L50) be better?

app/celery/scheduled_tasks.py
```diff
@@ -375,3 +380,68 @@ def delete_old_records_from_events_table():
...
+    brands_with_no_organisation = []
+    for new_brand in new_email_brands:
+        if not new_brand.organisations:
+            brands_with_no_organisation.append(new_brand)
+
+        else:
+            for organisation in new_brand.organisations:
+                brands_by_organisation[organisation].append(new_brand)
+
+    template = jinja2.Template(
```
quis, line 412: Could you pull this out into its own file please? Will make it easier to edit in the future when we add some guidance from https://docs.google.com/document/d/17tZKy0ogUOmUkjydjMMRTzEVZzpBrJFQEf3k789o3gs/edit#heading=h.xficsb7mfn5u

app/celery/scheduled_tasks.py
samuelhwilliams: I had it as type_task originally but it was suggested we rarely use that and incident is more common for us.

tests/app/delivery/test_send_to_providers.py
```diff
@@ -393,6 +393,23 @@ def test_send_email_should_use_service_reply_to_email(sample_service, sample_ema
     )
 
 
+@pytest.mark.parametrize("service_fixture", ["sample_service", "sample_service_with_email_branding"])
+def test_send_email_works_with_and_without_email_branding(request, service_fixture, sample_email_template, mocker):
+    service = request.getfixturevalue(service_fixture)
+    mocker.patch("app.aws_ses_client.send_email", return_value="reference")
+
+    db_notification = create_notification(template=sample_email_template, reply_to_text="foo@bar.com")
+    create_reply_to_email(service=service, email_address="foo@bar.com")
```
klssmith, line 402: The tests pass without this line, so I don't think we need to create the reply_to email address for the test

app/celery/scheduled_tasks.py
samuelhwilliams: Have switched to TYPE_TASK on second thought, it really does just feel more appropriate...

tests/app/conftest.py
```diff
@@ -157,6 +157,14 @@ def sample_service(sample_user):
     return service
 
 
+@pytest.fixture(scope="function")
+def sample_service_with_email_branding(sample_service):
+    sample_service.email_branding = create_email_branding(
+        id=uuid.uuid4(), logo="1ac6f483-3105-4c9e-9017-dd7fb2752c44-nhs-blue_x2.png", name="NHS"
```
klssmith, line 163: You could optionally use a non-NHS example for this fixture just in case someone tries to use this along with the `nhs_email_branding` fixture in the same test and gets a surprising integrity error - but that's probably very unlikely to happen.

tests/app/delivery/test_send_to_providers.py
samuelhwilliams: Patched in fixup commit 5da28b6b9ed665cef0dbe809bb83c286ea0c9b43

tests/app/conftest.py
samuelhwilliams: Patched in fixup commit 5da28b6b9ed665cef0dbe809bb83c286ea0c9b43

klssmith (approved):

PR #3651, opened 2022-11-25
add an endpoint to generate a name for a provided alt text

leohemsted:
alt text is not unique, it doesn't need to be. however, names should be unique so people can distinguish email branding options in a list of them.

this endpoint takes in a prospective alt_text, and checks the DB. if there's nothing in the DB with that name, it returns the alt text as provided.

If there is, it tries `{name} (alternate 1)`, and goes up by 1 each time. It compares case-insensitive, though always returns a name with the same case as the input alt_text.

If a branding has 100 alternates already, it gives up. We should probably look into what's happening before that.

### DIFF
--- a/app/dao/email_branding_dao.py
+++ b/app/dao/email_branding_dao.py
@@ -3,6 +3,17 @@ from app.dao.dao_utils import autocommit
 from app.models import EmailBranding
 
 
+def dao_get_existing_alternate_email_branding_for_name(name):
+    """
+    Returns any email branding with name of format `{name} (alternate {x})`
+
+    where name is the provided name and x is an integer starting at 1
+    """
+    return (
+        EmailBranding.query.filter(EmailBranding.name.ilike(f"{name} (alternate %)")).order_by(EmailBranding.name).all()
+    )
+
+
 def dao_get_email_branding_options():
     return EmailBranding.query.all()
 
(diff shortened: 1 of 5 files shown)


app/dao/email_branding_dao.py
```diff
@@ -3,6 +3,17 @@
...
 
 
+def dao_get_existing_alternate_email_branding_for_name(name):
+    """
+    Returns any email branding with name of format `{name} (alternate {x})`
+
+    where name is the provided name and x is an integer starting at 1
+    """
+    return (
+        EmailBranding.query.filter(EmailBranding.name.ilike(f"{name} (alternate %)")).order_by(EmailBranding.name).all()
```
samuelhwilliams, line 13: nit: I feel like the implication here is the ordering should be based numerically on x but it'll actually be lexicographic/alphabetical.

app/email_branding/rest.py
```diff
@@ -73,3 +76,30 @@ def update_email_branding(email_branding_id):
...
+
+    alt_text = data["alt_text"]
+
+    existing_branding = dao_get_email_branding_by_name_case_insensitive(alt_text)
+    if not existing_branding:
+        chosen_name = alt_text
+    else:
+        existing_alternate_branding_options = [
+            x.name for x in dao_get_existing_alternate_email_branding_for_name(alt_text)
+        ]
```
samuelhwilliams, line 95: optional: this would be nicer as a set than a list

app/dao/email_branding_dao.py
leohemsted: Good point - in this case the ordering doesn't actually matter for our use case since we immediately chuck all the names in a set anyway, but that's definitely something to be aware of.

app/email_branding/rest.py
leohemsted: very uncharacteristic of me to miss an opportunity to use a set. squashed in to the commit

PR #3652, opened 2022-11-25
Persist reference provided by services in CSV

CrystalPea:
When a service sends notifications by uploading a spreadsheet with recipients and personalisation, so through sending a job, if that service provides a column called reference, we will persist the value of this column as:
- for SMS and email, as "reference" field
- for letters, as "client_reference" field

This can help services identify their notifications easier in the weekly reports they download. It can also help them identify the notifications better if we provide them the information from notification_history table.


app/celery/tasks.py
```diff
@@ -259,6 +262,7 @@ def save_email(self, service_id, notification_id, encrypted_notification, sender
             job_row_number=notification.get("row_number", None),
             notification_id=notification_id,
             reply_to_text=reply_to_text,
+            reference=notification.get("client_reference", None),
```
leohemsted, line 265: can we set `client_reference` rather than `reference` here (and in sms)? i dont think we return `reference` in any downloadable reports as it's an internal thing. Plus as katie pointed out in slack we use it for SES email references so it would just get overwritten immediately.

CrystalPea: Will do ☺️ . Good spot!

CrystalPea: Done now 👍🏼

leohemsted (approved):

PR #3653, opened 2022-11-28
Add guidance to the daily branding pool review ticket

quis:
image
image
image

***

image

### DIFF
--- a/templates/tasks/scheduled_tasks/new_email_brandings.html
+++ b/templates/tasks/scheduled_tasks/new_email_brandings.html
@@ -1,9 +1,9 @@
-<h2>
-    New email branding to review
-</h2>
-<p>These are the new email brands uploaded since {{ yesterday }}.</p><hr>
+{% if brands_by_organisation %}
+<h2>New email branding to review</h2>
+<p>Uploaded since {{ yesterday }}:</p>
+{% endif %}
 {%- for org, brands in brands_by_organisation.items() -%}
-    <p><a href="{{ domain }}/organisations/{{ org.id }}">{{ org.name }}</a>:</p><ul>
+    <p><a href="{{ domain }}/organisations/{{ org.id }}">{{ org.name }}</a>{% if not org.email_branding_id %} (no default){% endif %}:</p><ul>
     {%- for brand in brands -%}
         <li><a href="{{ domain }}/email-branding/{{ brand.id }}/edit">{{ brand.name }}</a></li>
     {%- endfor -%}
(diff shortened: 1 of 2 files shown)


templates/tasks/scheduled_tasks/new_email_brandings.html
```diff
@@ -1,14 +1,87 @@
...
+<ul>
+    <li>if it’s a duplicate</li>
+    <li>image quality</li>
+    <li>if it can be used by more than one service</li>
+    <li>branding name</li>
+</ul>
+
+<h3>1. Duplicates</h3>
+<p>
+    If the same logo is in the branding pool twice then keep whichever one looks best.
```
samuelhwilliams, line 28: If this happens, should any services with the 'worse' brand be moved over to the new one? Should we have guidance indicating to do or not do that?

templates/tasks/scheduled_tasks/new_email_brandings.html
```diff
@@ -1,14 +1,87 @@
...
+<p>Check the image has a transparent background and is:</p>
+<ul>
+    <li>landscape</li>
+    <li>not distorted or stretched</li>
+    <li>not low resolution</li>
+</ul>
+
+<h4>Text in a logo</h4>
+<p>
+    Some logos may contain text. If the logo contains the name of the service or organisation, it must be clear to read. Ideally, text size ideally should not be less than 14px.
```
samuelhwilliams, line 42: "Ideally, text size ideally" probably can drop one of those

quis:
```suggestion
    Some logos may contain text. If the logo contains the name of the service or organisation, it must be clear to read. Ideally, text size should be 14px or larger.
```

templates/tasks/scheduled_tasks/new_email_brandings.html
quis: That’s the kind of quality control of individual services branding that we’re trying to get away from I think

templates/tasks/scheduled_tasks/new_email_brandings.html
```diff
@@ -1,14 +1,87 @@
...
+</ul>
+
+<hr>
+
+{% endif %}
+
 {%- if brands_with_no_organisation -%}
-    <p>Unassociated brands:</p><ul>
-    {%- for brand in brands_with_no_organisation -%}
+    <p>These new brands are not associated with any organisation and do not need reviewing:</p><ul>
```
samuelhwilliams, line 83: nit: Vaguely an edge case but this sentence may look sort of weird by itself if there's no new brands attached to an org - there's no header and no "new since {{date}}" introduction as that's all gated in the "only if new org brands" condition

samuelhwilliams (approved):
Optional comments, but otherwise good 👍

PR #3654, opened 2022-11-28
Allow any job in scheduled state to be cancelled

samuelhwilliams:
The function used to find a job to cancel checks two things:

1) That it's in the SCHEDULED state
2) That it's scheduled for the future.

This means that a job which is scheduled to send, but past the scheduled
time, cannot be cancelled. But there are cases where users may want to
cancel these jobs still. We know that a job hasn't started processing if
the state is SCHEDULED, so it should still be safe to cancel even if the
ideal processing start time has passed. Let's remove the check on
scheduled_for and only require that a job's state is SCHEDULED to
determine if it's cancellable.

### DIFF
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -129,13 +129,19 @@ def dao_set_scheduled_jobs_to_pending():
     return jobs
 
 
-def dao_get_future_scheduled_job_by_id_and_service_id(job_id, service_id):
-    return Job.query.filter(
-        Job.service_id == service_id,
-        Job.id == job_id,
-        Job.job_status == JOB_STATUS_SCHEDULED,
-        Job.scheduled_for > datetime.utcnow(),
-    ).one()
+def dao_get_scheduled_job_by_id_and_service_id(job_id, service_id):
+    """Fetch a scheduled job from the DB, taking an exclusive lock on the row so that it can't be edited until this
+    transaction is committed or rolled back. This can be used to fetch the job in order to cancel it, while at the
+    same time preventing that job from being picked up and processing started by celery."""
+    return (
+        Job.query.filter(
+            Job.service_id == service_id,
+            Job.id == job_id,
+            Job.job_status == JOB_STATUS_SCHEDULED,
+        )
+        .with_for_update()
+        .one()
+    )
 
 
 def dao_create_job(job):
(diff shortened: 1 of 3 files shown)


migrations/versions/0388_populate_letter_branding.py
```diff
@@ -7,8 +7,8 @@
 """
 from alembic import op
 
-revision = '0388_populate_letter_branding'
-down_revision = '0387_migrate_alt_text'
+revision = "0388_populate_letter_branding"
```
klssmith, line 10: Is something flagging this formatting as invalid? Everything passed for me when I added this file because we exclude the migrations directory from black [here](https://github.com/alphagov/notifications-api/blob/c917176a9b6a1401a8541b52d813272d65852159/Makefile#L72) 🤔

samuelhwilliams: Ah, yeah, this change is coming from `pre-commit`, which doesn't have the `extend-exclude`. Looking at the best way to deal with it...

samuelhwilliams: Although I think we actually just shouldn't have migrations excluded, it seems like everything is mostly formatted correctly there - except the latest migrations.

So just going to drop that exclusion...

klssmith:
I was wondering about the risk of someone cancelling a scheduled job just as it's starting to run, which is now _very_ slightly more likely if we're letting you cancel past jobs. Is that something you reckon is just too low risk to worry about?

leohemsted:
There is technically a risk, however it's pretty low as it'd require 

cancel_job HTTP: fetch the scheduled job
run_scheduled_job task: runs, fetches the job too.
run_scheduled_job: updates to pending
process_job task: runs, selects the job (still in pending)
process_job task: sees status is and proceeds
cancel_job HTTP: update the task to cancelled
potential for wierdities here. most likely is we just end up with process_job overwriting the cancel status with one of in_progress and then finished

if we wanted to be super super secure, we could use a select `with_for_update` - which will keep an access exclusive lock on the row for the duration of the transaction, which means we can then update to cancelled safe in the knowledge no-one else will be able to fetch it until we've committed. we do this in a couple of other places

https://github.com/alphagov/notifications-api/blob/main/app/dao/jobs_dao.py#L108-L121

samuelhwilliams:
Good belt-and-braces 👍 Applied `with_for_update` in c8620c1d0a1697d91dfda26b224d43598dca5b79

leohemsted (approved):

PR #3655, opened 2022-11-28
Stop excluding migrations for formatting

samuelhwilliams:
99% of our migrations were reformatted for black but we are excluded the check on that directory. Let's not do that, and fix the few migrations that have gone in with iffy formatting since.


samuelhwilliams:
Nah you're not really, I think am for leaving the migrations directory excluded 😓

PR #3656, opened 2022-11-29
Weekly report on low free inbound sms numbers

samuelhwilliams:
We're rolling out some changes soon to allow services to enable inbound
SMS by themselves. At the moment they have to contact support and we'll
manually give them a number, which means we are constantly aware of how
many numbers we have left and can spot when we might need more.

If users can take from the pool themselves, without any oversight, we
need some monitoring in place to alert us automatically that we need to
ask for more numbers.

We don't know how quickly numbers will get used, but expect that a
weekly report should be fine, and that 50 numbers are unlikely to be
used up within the (roughly) week that it takes for new numbers to be
given to us.

Related credentials PR: https://github.com/alphagov/notifications-credentials/pull/306

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -433,6 +433,10 @@ def zendesk_new_email_branding_report():
 @notify_celery.task(name="check-for-low-available-inbound-sms-numbers")
 @cronitor("check-for-low-available-inbound-sms-numbers")
 def check_for_low_available_inbound_sms_numbers():
+    if (env := current_app.config["NOTIFY_ENVIRONMENT"].lower()) not in {"live", "production", "test"}:
+        current_app.logger.info(f"Skipping report run on in {env}")
+        return
+
     num_available_inbound_numbers = len(dao_get_available_inbound_numbers())
     current_app.logger.info(f"There are {num_available_inbound_numbers} available inbound SMS numbers.")
     if num_available_inbound_numbers > current_app.config["LOW_INBOUND_SMS_NUMBER_THRESHOLD"]:
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -21,6 +21,7 @@ from app.celery.tasks import (
 )
 from app.config import QueueNames, TaskNames
 from app.cronitor import cronitor
+from app.dao.inbound_numbers_dao import dao_get_available_inbound_numbers
 from app.dao.invited_org_user_dao import (
     delete_org_invitations_created_more_than_two_days_ago,
 )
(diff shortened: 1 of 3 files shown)


tests/app/celery/test_scheduled_tasks.py
```diff
@@ -882,17 +874,57 @@ def test_zendesk_new_email_branding_report_calculates_last_weekday_correctly(
...
 
 
 def test_zendesk_new_email_branding_report_does_not_create_ticket_if_no_new_brands(notify_db_session, mocker):
     mock_send_ticket = mocker.patch("app.celery.scheduled_tasks.zendesk_client.send_ticket_to_zendesk")
     zendesk_new_email_branding_report()
     assert mock_send_ticket.call_args_list == []
+
+
+def test_check_for_low_available_inbound_sms_numbers_logs_zendesk_ticket_if_too_few_numbers(notify_db_session, mocker):
+    mocker.patch.dict("app.celery.scheduled_tasks.current_app.config", {"LOW_INBOUND_SMS_NUMBER_THRESHOLD": 10})
```
klssmith, line 894: We have context managers to set a either a single config value or multiple in tests, for example
https://github.com/alphagov/notifications-api/blob/ac7791692ad488e99a92ecea12144c574f0598bb/tests/app/broadcast_message/test_utils.py#L338-L339

I'm trying to remember if there was a particular reason why we always tend to use that way of doing things instead of mocking the config

samuelhwilliams: Cool, will switch 👍  They both do the same thing, looks like we've just made our own mocker.patch.ditch 😅

samuelhwilliams: Patched in [28c348f](https://github.com/alphagov/notifications-api/pull/3656/commits/28c348ff4c00ecff450531c6076cd0e045142b69)

app/celery/scheduled_tasks.py
```diff
@@ -427,3 +428,32 @@ def zendesk_new_email_branding_report():
...
+        "Request more from our provider (MMG) and load them into the database.\n\n"
+        "Follow the guidance here: "
+        "https://github.com/alphagov/notifications-manuals/wiki/Support-Runbook#Add-new-inbound-SMS-numbers"
+    )
+
+    ticket = NotifySupportTicket(
+        subject="Request more inbound SMS numbers",
+        message=message,
+        ticket_type=NotifySupportTicket.TYPE_TASK,
+        technical_ticket=False,
```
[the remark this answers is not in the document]
samuelhwilliams: Patched in [28c348f](https://github.com/alphagov/notifications-api/pull/3656/commits/28c348ff4c00ecff450531c6076cd0e045142b69)

klssmith (approved):
Looks good to me - we can always adjust the value of `LOW_INBOUND_SMS_NUMBER_THRESHOLD` later once we've seen how quickly numbers are getting used

I don't think I have access to cronitor to double check things there, but it looks like it has been set up as expected here

samuelhwilliams:
@klssmith there are shared credentials accessible via `notify-pass credentials/cronitor`

PR #3657, opened 2022-11-29
don't send daily branding review on other envs (and refactor how we access config)

leohemsted:
Add the env check to the daily branding review to ensure we don't send it from local/from staging/etc. We already do this check in a bunch of places as seen in the other commits.
--------
I was fed up of us having sketchy config checks everywhere. `current_app` returns the flask app, so create a subclass of `Flask` that has a few helper methods to reduce calls to the config directly for checks like this. Over time we can add more checks elsewhere if we want to.

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -196,8 +196,8 @@ def raise_alert_if_letter_notifications_still_sending():
         message = "There are {} letters in the 'sending' state from {}".format(
             still_sending_count, sent_date.strftime("%A %d %B")
         )
-        # Only send alerts in production
-        if current_app.config["NOTIFY_ENVIRONMENT"] in ["live", "production", "test"]:
+
+        if current_app.should_send_zendesk_alerts:
             message += ". Resolve using https://github.com/alphagov/notifications-manuals/wiki/Support-Runbook#deal-with-letters-still-in-sending"  # noqa
 
             ticket = NotifySupportTicket(
@@ -278,7 +278,7 @@ def letter_raise_alert_if_no_ack_file_for_zip():
     zip_file_set.discard("")
 
     if len(zip_file_set - ack_file_set) > 0:
-        if current_app.config["NOTIFY_ENVIRONMENT"] in ["live", "production", "test"]:
+        if current_app.should_send_zendesk_alerts:
             ticket = NotifySupportTicket(
                 subject="Letter acknowledge error",
                 message=message,
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -244,7 +244,7 @@ def check_if_letters_still_pending_virus_check():
 
             Notifications: {sorted(letter_ids)}"""
 
-        if current_app.config["NOTIFY_ENVIRONMENT"] in ["live", "production", "test"]:
+        if current_app.should_send_zendesk_alerts:
             ticket = NotifySupportTicket(
                 subject=f"[{current_app.config['NOTIFY_ENVIRONMENT']}] Letters still pending virus check",
                 message=msg,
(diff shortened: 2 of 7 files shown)


samuelhwilliams (approved):
🤦

app/celery/scheduled_tasks.py
```diff
@@ -418,12 +418,13 @@ def zendesk_new_email_branding_report():
...
-    ticket = NotifySupportTicket(
-        subject="Review new email brandings",
-        message=message,
-        ticket_type=NotifySupportTicket.TYPE_TASK,
-        technical_ticket=False,
-        ticket_categories=["notify_no_ticket_category"],
-        message_as_html=True,
-    )
-    zendesk_client.send_ticket_to_zendesk(ticket)
+    if current_app.config["NOTIFY_ENVIRONMENT"] in ["live", "production", "test"]:
```
samuelhwilliams, line 421: nit: could be a set 😉

leohemsted: actually i might go and do the decent thing and make a nice model thing

leohemsted: fun fact: im pretty sure that `live` _and_ `production` being in the list is because we just weren't sure what the variable would be on prod paas 🐵

samuelhwilliams: yeah I figured that might be the case 😂

CrystalPea, line 421: optional: we have `["live", "production", "test"]` repeated like 6 times in this repo now. Maybe we could have a single source of truth for this, like a `PROD_OR_TEST` or something, set in init file or in utils?

CrystalPea (approved):
Nice refactor on top 👏🏼

samuelhwilliams:
I have just merged another instance of the env check in https://github.com/alphagov/notifications-api/pull/3656, so might need to update that (sorry/thanks) 🙏

app/celery/nightly_tasks.py
```diff
@@ -197,7 +197,7 @@ def raise_alert_if_letter_notifications_still_sending():
             still_sending_count, sent_date.strftime("%A %d %B")
         )
         # Only send alerts in production
```
samuelhwilliams, line 199: This comment semi lies - can probably be removed now

app/notify_api_flask_app.py
```diff
@@ -0,0 +1,15 @@
...
+    @property
+    def is_prod(self):
+        return self.config["NOTIFY_ENVIRONMENT"] in {"live", "production"}
+
+    @property
+    def is_test(self):
+        return self.config["NOTIFY_ENVIRONMENT"] == "test"
+
+    @property
+    def should_send_zendesk_alerts(self):
```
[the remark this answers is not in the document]
leohemsted: yeah i did get that, but then my vscode setup is a bit hit and miss on lots of inference too. we'll just have to use github's ai thing to sort it out for us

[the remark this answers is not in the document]
leohemsted: id much rather have this than something like the following tho

```py
from app.utils import should_send_zendesk_alerts
if should_send_zendesk_alerts(current_app):
  ...
```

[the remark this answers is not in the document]
leohemsted: an alternative would be something like the following:

but i dont really like it because we'd still need to use current_app for some things, and this new class for other things which feels messy

```py
# app/__init__.py (or wherever)
class NotifyApi:
  @property
  def is_prod(self):
    return current_app.config['NOTIFY_ENVIRONMENT'] == 'production'

notify_api = NotifyApi()
```

```py
# regular code
from app import notify_api
if notify_api.is_prod: 
  ...
```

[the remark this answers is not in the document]
samuelhwilliams: We don't need to change it, don't worry. Those other options are just complicating things more than necessary I think.

CrystalPea (approved):
Still good to go 👍🏼

PR #3658, opened 2022-11-30
include entrypoint script

playfair0:
- dropping in a single entrypoint file which hopefully makes it seamless to run various flask migrate, beats and various worker types
- minor additional changes to the dockerfile
- and docker-related makefile calls


idavidmcdonald (commented):
Nice. I like the entrypoint approach you've got here and think this could work well for us in production. It feels very similar to the purpose of https://github.com/alphagov/notifications-api/blob/main/scripts/paas_app_wrapper.sh but won't need a lot of the crud in there related to the PaaS.

I left a range of comments, feel free to put some time in my diary to chat through them or do some pairing on if that would be useful!

idavidmcdonald (approved):
Nice work David, especially on changing it so the needed environment variables are passed through to `docker run`.

As you are off ill and in the want to move this forward before you back in the new year, I made a few small adjustments and additions but all are minor.

I've also built and run this locally (for all of the different `make-x-with-docker`) and they work too which is awesome.

I couldn't figure out the answer to my questions about `chown` or `shift` without you so I k kept them in. When you are back, take a look and if you figure out either of them should be changed then feel free to follow up in a separate PR.

PR #3659, opened 2022-11-30
Run tests on recommended number of CPU cores

quis:
At the moment we parallelize the tests across 4 cores.

On some machines more cores are available, so using them makes the tests run faster.

pytest-xdist will automatically pick the best number of CPU cores when passed `auto` as the value for the `-n` flag<sup>1</sup>.

This matches what we do in the admin repo:
https://github.com/alphagov/notifications-admin/blob/46b2415fd2a96af4a6503d22f4bad80eaf496f98/Makefile#L67

***

# Before<sup>2</sup> 

image

# After<sup>2</sup>

image

***

1. https://pytest-xdist.readthedocs.io/en/latest/distribution.html
2. Run on a 16 inch Macbook Pro with an Apple M1 Pro chip

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ test: ## Run tests
 	flake8 .
 	isort --check-only ./app ./tests
 	black --check .
-	pytest -n4 --maxfail=10
+	pytest -n auto --maxfail=10
 
 .PHONY: freeze-requirements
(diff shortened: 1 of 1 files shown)


klssmith (approved):

leohemsted:
Out of curiosity: Tests still run with four cores on concourse, taking about 3m40s.

PR #3660, opened 2022-11-30
Add a GitHub Action check to block fixup commits

samuelhwilliams:
It can be useful/easy to use fixup commits to make changes to a PR based on review comments. They make it easy for a reviewer to see and only re-review what's changed since they last commented, but we don't want them to be merged to main.

This GitHub action will add a status check to the PR that can block merge if any fixup commits are present, forcing the dev to squash them down sensibly to ensure a cleaner git history. It feels easier than adding a concourse check...

At the moment we have GitHub actions fully disabled. We'd need to enable them, but we can do so in a restrictive manner - i.e. only enabling this github action (see below). The code behind this check is fairly simple and available for review here: https://github.com/13rac1/block-fixup-merge-action/blob/v2.0.0/entrypoint.sh. We pin a specific version of the action so won't eg be exposed to someone injecting a vulnerability in a new version.

image

### DIFF
--- a/.github/workflows/git.yml
+++ b/.github/workflows/git.yml
@@ -1,0 +1,12 @@
+name: Git Checks
+
+on: [pull_request]
+
+jobs:
+  block-fixup:
+    runs-on: ubuntu
+
+    steps:
+    - uses: actions/checkout@v2.0.0
+    - name: Block Fixup Commit Merge
+      uses: 13rac1/block-fixup-merge-action@v2.0.0


idavidmcdonald:
Sam and I spoke about this.

In short, I'm cautious of picking up another tool that people need to be aware of (Github actions) or learn how to use for potentially a small amount of benefit (fixup commits getting merged accidentally isn't that big a deal).

However, there is some benefit to this and open to seeing what the team think in one of our Tuesday catch ups and then Sam can make the final call on whether to merge or not.

PR #3661, opened 2022-12-01
Rename live to prod

leohemsted:
#### ["Calling `production` `live` is a remain of an old time that still makes our lives difficult today." - Sakis, Apr 2022](https://github.com/alphagov/notifications-cf-monitoring/commit/70264883f56e99acca9cc055bd2276a86efd70b6)

Confusingly, the OS environment variable `NOTIFY_ENVIRONMENT` is set to equal the paas space, so is `production`, however we then set `current_app.config["NOTIFY_ENVIRONMENT"]` to equal `live`. This has led to lots of confusion, as seen in all the places in code where we're checking if the environment is one of those two values. Not even we are certain which environment we're on!

This PR fixes that by renaming that variable to "production" and ensuring that wherever it is used we refer to `production` instead of `live`.

Note, the variable goes in to our statsd mappings as the `space` value. Then, our prometheus config replaces `live` with `production`.

Note, there's two other notable places that still talk about "live" with regards to the environment:

1. The celery queues. Migrating celery queues in a zero downtime way would be non-trivial, so lets not poke that beast right now.
2. The `live-notifications-csv-upload` bucket. All the other buckets are prefixed with `production`.

--------

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -531,9 +531,9 @@ class Staging(Config):
     CHECK_PROXY_HEADER = True
 
 
-class Live(Config):
+class Production(Config):
     NOTIFY_EMAIL_DOMAIN = "notifications.service.gov.uk"
-    NOTIFY_ENVIRONMENT = "live"
+    NOTIFY_ENVIRONMENT = "production"
     CSV_UPLOAD_BUCKET_NAME = "live-notifications-csv-upload"
     CONTACT_LIST_BUCKET_NAME = "production-contact-list"
     TEST_LETTERS_BUCKET_NAME = "production-test-letters"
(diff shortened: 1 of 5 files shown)


CrystalPea (commented):
Nice one, only do we expect some issues while deploying this? Like when some boxes have new environment name, and some have old environment name?

I guess other possible problems would be caught on preview.

leohemsted:
It should be fine and shouldn't affect anything - this is just a python variable in memory that is only used for

* statsd (here, statsd renames it back to `production`, so if we do that in advance things will just work fine)
* choosing the config to load (we're just removing some dead code and renaming classes here.

The places where `live` WOULD make a difference and require careful deployment, I haven't changed in this PR. To my knowledge the only places where that happens are:

* The queue prefix being `live` (this is defined by the os env var `NOTIFICATION_QUEUE_PREFIX`)
* The CSV bucket being prefixed by `live` (this is just a hardcoded string rather than being generated based on the environment).

CrystalPea (approved):

PR #3663, opened 2022-12-02
flask_env is deprecated

leohemsted:
FLASK_DEBUG=1 is the proper way to have the debug stack trace screen appear

(this is a very underwhelming code change - the key thing is I also renamed the environment variable in my `environment.sh`)


quis (approved):
https://flask.palletsprojects.com/en/2.2.x/config/#DEBUG

https://github.com/pallets/flask/issues/4714

PR #3664, opened 2022-12-07
Add new endpoint for adding inbound number to a service

klssmith:
We already have an endpoint to add an inbound number to a service - `.add_service_sms_sender` is used for both adding new sms senders and adding a particular inbound number to a service. We want to be able to add an inbound number to a service without specifying the exact number to add. This adds a new endpoint which handles adding any inbound number and a particular inbound number to the service.

`.add_service_sms_sender` will be changed later so that it doesn't deal with inbound numbers but only deals with adding sms senders.

### DIFF
--- a/app/inbound_number/inbound_number_schema.py
+++ b/app/inbound_number/inbound_number_schema.py
@@ -1,0 +1,9 @@
+from app.schema_validation.definitions import uuid
+
+add_inbound_number_to_service_request = {
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "POST add service inbound number",
+    "type": "object",
+    "title": "Assign inbound number to service",
+    "properties": {"inbound_number_id": uuid},
+}
--- a/app/inbound_number/rest.py
+++ b/app/inbound_number/rest.py
@@ -1,12 +1,22 @@
-from flask import Blueprint, jsonify
+from flask import Blueprint, jsonify, request
 
 from app.dao.inbound_numbers_dao import (
+    dao_allocate_number_for_service,
     dao_get_available_inbound_numbers,
     dao_get_inbound_number_for_service,
     dao_get_inbound_numbers,
     dao_set_inbound_number_active_flag,
 )
+from app.dao.service_sms_sender_dao import (
+    dao_add_sms_sender_for_service,
+    dao_get_sms_senders_by_service_id,
+    update_existing_sms_sender_with_inbound_number,
+)
 from app.errors import register_errors
+from app.inbound_number.inbound_number_schema import (
+    add_inbound_number_to_service_request,
+)
+from app.schema_validation import validate
 
 inbound_number_blueprint = Blueprint("inbound_number", __name__, url_prefix="/inbound-number")
 register_errors(inbound_number_blueprint)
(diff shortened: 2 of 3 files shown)


samuelhwilliams:
I'm wondering if we need this. Could the admin app just get a list of the inbound SMS numbers using the existing route, pick one at random, and POST back using the existing endpoint? Might save us some code. Am I missing something? :)

klssmith:
Admin could get a list of inbound sms and then use the existing route, but I was chatting to @leohemsted about this last week and we discussed having a new route just for setting the inbound number. I like the idea of having `add_service_sms_sender` just for adding a new SMS sender, while having a separate route for assigning an inbound SMS number to a service. The `add_service_sms_sender` will be simplified later, which will get rid of the duplication.

samuelhwilliams (commented):
One small comment, otherwise I think this looks good.

app/inbound_number/rest.py
```diff
@@ -37,3 +47,40 @@ def get_available_inbound_numbers():
...
+    form = validate(request.get_json(), add_inbound_number_to_service_request)
+
+    if form.get("inbound_number_id"):
+        inbound_number_id = form["inbound_number_id"]
+    else:
+        try:
+            available_number = dao_get_available_inbound_numbers()[0]
+            inbound_number_id = available_number.id
+        except IndexError:
+            raise InvalidRequest("There are no available inbound numbers", status_code=400)
```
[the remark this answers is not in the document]
klssmith: I've changed this just to be an `Exception` with a `500` status code and have confirmed you do see the message in the logs if there are no numbers

app/inbound_number/inbound_number_schema.py
```diff
@@ -0,0 +1,9 @@
+from app.schema_validation.definitions import uuid
+
+add_inbound_number_to_service_request = {
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "POST add service inbound number",
+    "type": "object",
+    "title": "Asign inbound number to service",
```
samuelhwilliams, line 7: nit: typo in assign

app/inbound_number/rest.py
```diff
@@ -37,3 +47,40 @@ def get_available_inbound_numbers():
...
+    form = validate(request.get_json(), add_inbound_number_to_service_request)
+
+    if form.get("inbound_number_id"):
+        inbound_number_id = form["inbound_number_id"]
+    else:
+        try:
+            available_number = dao_get_available_inbound_numbers()[0]
+            inbound_number_id = available_number.id
+        except IndexError:
+            raise Exception("There are no available inbound numbers")
```
samuelhwilliams, line 67: nit: we could store the IndexError in a variable (eg `e`) and add `from e` here to chain the exceptions together.

samuelhwilliams (approved):

app/inbound_number/inbound_number_schema.py
klssmith: Updated

app/inbound_number/rest.py
samuelhwilliams: There's not a huge benefit, but it's probably more semantically appropriate. It says says "we're handling the IndexError and purposefully throwing this new error because of it", rather than "we've tried to handle the IndexError and something else has gone wrong".

In a printed stacktrace it changes the line `During handling of the above exception, another exception occurred:` to `The above exception was the direct cause of the following exception:`, which is a bit clearer.

Quick primer: https://stefan.sofa-rockers.org/2020/10/28/raise-from/

Note on nits I leave: I use them in the way they're mentioned [here](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.): specifically, "Reviewers should _always_ feel free to leave comments expressing that something could be better, but if it’s not very important, prefix it with something like “Nit: “ to let the author know that it’s just a point of polish that they could choose to ignore."

PR #3665, opened 2022-12-07
stop pinning flask-sqlalchemy to git commit

leohemsted:
We pinned flask-sqlalchemy to a specific commit from a PR to bring in a specific option to disable count in `query.paginate` - that's now been released after _a while_, so we can unpin and go back to the pypi release. This version requires flask >= 2.2 so bump that too.

https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/changes/#changes
https://flask.palletsprojects.com/en/2.2.x/changes/


klssmith:
Optional, but we could now remove `SQLALCHEMY_RECORD_QUERIES` and `SQLALCHEMY_TRACK_MODIFICATIONS` from the config to simplify things - they're now both disabled by default so we don't need to set them to False any more

leohemsted:
good catch on the SQLALCHEMY_* vars - ive already got the FLASK_ENV in another PR (i looked at it across all our repos, we do use it elsewhere in other repos frustratingly but at least in api it's just in the readme and peoples environment.sh files

klssmith (approved):

PR #3668, opened 2022-12-12
Add separate rate limits per notification type

samuelhwilliams:
## Summary
Start tracking rate limits for services for each type of notification that can be sent.

We add the fields to the services table in the DB, and set the default values to be unreasonably high (such that they'll never be hit for now - a service would have to send 1B notifications in a day to hit the limit).

A follow-up PR on the API will backfill all of the null values to the default, and set the field as non-nullable.

A separate PR on admin will add the views to allow editing these limits, and to set the rate limits to our desired actual limits (once we've worked out what they are).

## Ticket
https://trello.com/c/EwdxmIQc/132-add-letter-day-rate-limits-per-service


app/notifications/validators.py
```diff
@@ -57,31 +57,43 @@ def check_service_over_api_rate_limit(service, key_type):
...
+                service.id, limit_group, int(service_stats), limit
             )
         )
-        raise TooManyRequestsError(service.message_limit)
+        raise TooManyRequestsError(f"{limit_group}: {limit}")
     return int(service_stats)
 
 
-def check_rate_limiting(service, api_key):
+def check_rate_limiting(service, api_key, notification_type=None):
```
leohemsted, line 90: i'd rather we didnt add this code yet if it's currrently unused (since we haven't updated how this function is called by post_notifications.py, and we haven't updated how `check_service_over_daily_message_limit` is called in tasks.py either)

app/notifications/validators.py
```diff
@@ -57,31 +57,43 @@ def check_service_over_api_rate_limit(service, key_type):
                 raise RateLimitError(rate_limit, interval, key_type)
 
 
-def check_service_over_daily_message_limit(service, key_type):
+def check_service_over_daily_message_limit(service, key_type, notification_type=None):
```
leohemsted, line 60: hmmm, is it better if we always require `notification_type` and then always check both cache keys (in the same way that we incremement both cache keys).

app/notifications/validators.py
```diff
@@ -57,31 +57,43 @@ def check_service_over_api_rate_limit(service, key_type):
...
         return 0
 
-    cache_key = daily_limit_cache_key(service.id)
+    limits = {
+        None: service.message_limit,
+        LETTER_TYPE: service.letter_message_limit,
+        EMAIL_TYPE: service.email_message_limit,
+        SMS_TYPE: service.sms_message_limit,
+    }
+    limit = limits[notification_type]
```
leohemsted, line 70: for the notification types, this will be null for old cached SerialisedService objects and also for those services that haven't been migrated yet - this'll lead to us comparing an int to None and getting errors due to that

tests/app/notifications/test_process_notification.py
```diff
@@ -249,7 +253,42 @@ def test_persist_notification_sets_daily_limit_cache_if_one_does_not_exists(
...
+            service=template.service,
+            personalisation={},
+            notification_type=notification_type,
+            api_key_id=api_key.id,
+            key_type=api_key.key_type,
+            reference="ref2",
+        )
+
+    assert (
+        mocker.call(str(service.id) + f"-{notification_type}-2016-01-01-count", 1, ex=86400) in mock_set.call_args_list
```
leohemsted, line 290: can we make this just one f string rather than a weird concatenation of a f string and a normal string 🙏

leohemsted: also we've now got three tests that kinda just interact weirdly.

for trial and live, make sure incr is called for the generic count
for trial and live, make sure set is called for the generic count
for the three notification types, make sure set is called for that specific count

im not entirely sure what would make these three tests cleaner, but i wonder if the notification type check can be parametrised in to the existing two tests? that might be overkill

also, i'd like to assert on the entire call args list, to ensure we're not setting additional incorrect keys anywhere

leohemsted (changes requested):
Migration won't work unfortunately. I've got a couple of other thoughts - mostly that I'd rather not add the actual checking code until we're actually going to use it, since it's only called synchronously and we're not actually invoking the new code paths yet.

migrations/versions/0389_split_rate_limits.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+    op.execute(
+        "UPDATE services_history "
+        "SET "
+        "letter_message_limit=999999999, "
+        "sms_message_limit=999999999, "
+        "email_message_limit=999999999"
+    )
+
+    op.alter_column("services", "letter_message_limit", existing_nullable=True, nullable=False)
```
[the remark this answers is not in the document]
samuelhwilliams: All great points 😱 

We can migrate the 'old' data - as in set the value of all columns - to 1B immediately can't we?

app/notifications/validators.py
samuelhwilliams: Hng, serialised services, yep.

migrations/versions/0389_split_rate_limits.py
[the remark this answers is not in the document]
leohemsted: we can, but we'll still need to set the default value for any services created after migration 1 is deployed but before code 1 containing the default prompt is deployed. so i'd suggest just doing all the setting of values in migration 2

app/notifications/validators.py
samuelhwilliams: Removed all of the functional code that was tracking the new rate limits - so this is just adding the columns and returning the fields now 👍

leohemsted (approved):

PR #3669, opened 2022-12-12
Add a make command to bump utils, using the tooling from utils version 59.3.0

quis:
59.3.0
---
* Add tooling to bump utils version in apps (which is then used by the makefile)

59.2.0
---
* Add support for creating redis cache keys for a specific notification type.

***

Complete changes: https://github.com/alphagov/notifications-utils/compare/59.1.1...59.3.0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,9 @@ CF_MANIFEST_PATH ?= /tmp/manifest.yml
 
 NOTIFY_CREDENTIALS ?= ~/.notify-credentials
 
+VIRTUALENV_ROOT := $(shell [ -z $$VIRTUAL_ENV ] && echo $$(pwd)/venv || echo $$VIRTUAL_ENV)
+PYTHON_EXECUTABLE_PREFIX := $(shell test -d "$${VIRTUALENV_ROOT}" && echo "$${VIRTUALENV_ROOT}/bin/" || echo "")
+
 
 ## DEVELOPMENT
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -77,6 +77,10 @@ freeze-requirements: ## Pin all requirements including sub dependencies into req
 	pip install --upgrade pip-tools
 	pip-compile requirements.in
 
+.PHONY: bump-utils
+bump-utils:  # Bump notifications-utils package to latest version
+	${PYTHON_EXECUTABLE_PREFIX}python -c "from notifications_utils.version_tools import upgrade_version; upgrade_version()"
```
samuelhwilliams, line 82: `PYTHON_EXECUTABLE_PREFIX` isn't defined in this makefile yet, so will need to add it. In admin it comes from:

```
VIRTUALENV_ROOT := $(shell [ -z $$VIRTUAL_ENV ] && echo $$(pwd)/venv || echo $$VIRTUAL_ENV)
PYTHON_EXECUTABLE_PREFIX := $(shell test -d "$${VIRTUALENV_ROOT}" && echo "$${VIRTUALENV_ROOT}/bin/" || echo "")
```

quis: Noted and added in [8cad60a](https://github.com/alphagov/notifications-api/pull/3669/commits/8cad60a89964b75a345d03fe721a076af1579a03)

samuelhwilliams (approved):

PR #3671, opened 2022-12-15
Ignore prompt_toolkit module for freezegun

samuelhwilliams:
Freezegun has an edge case interaction with `prompt_toolkit` that ends up breaking commands in ipdb (typed text doesn't appear on the prompt, which makes it really hard to use).

There's a PR on freezegun to ignore this module, which would mean we don't need to manually configure it, but for now it's not merged.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 import os
 from contextlib import contextmanager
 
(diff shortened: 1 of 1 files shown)


CrystalPea (approved):
:shipit:

PR #3673, opened 2022-12-16
Simplify `add_service_sms_sender` endpoint

klssmith:
We've added a new endpoint for adding an inbound number to a service (either a specific inbound number or a random one). This means that we can simplify the `add_service_sms_sender` endpoint so that it only deals with adding a "normal" sms sender to a service.


CrystalPea (approved):
Nice one 👏🏼 . Deleting code we no longer need is always satisfying :D

PR #3674, opened 2022-12-16
New rate limit fields are nullable

samuelhwilliams:


leohemsted (approved):
i’d ask why unit tests didnt pick this up, but i'm keen to unblock the pipeline first, and you’ll need to write unit tests when you do the admin side to actually make the fields editable anyway, so i'm happy to wait til then

probs some marshmallow bullllllshit

PR #3676, opened 2022-12-16
Make split rate limits non-nullable

samuelhwilliams:
We've rolled out the new columns, and defined defaults in SQLAlchemy, so now let's backfill all of the existing data and populate them with values, and then make them non-nullable.


klssmith:
The migration looks good to me, can you just commit the updated `.current-alembic-head` file too?

samuelhwilliams:
🤔 I ran `flask db upgrade` locally, not sure why that didn't get generated or committed.

I've added a pre-commit hook to help stop that happening again...

klssmith (approved):

PR #3677, opened 2022-12-19
Add columns for managing go live requests

quis:
This is so we can track:
- which organisations are allowed to approve their own go live requests (this is effectively a feature flag, but we don’t have permissions for organisations like we do for services)
- which services are currently waiting to go live, so we can filter on this attribute

Later we can:
- add default `False` values for these columns
- make them non-nullable 
- have the API return them for use in the admin app

### DIFF
--- a/migrations/versions/0391_add_go_live_columns.py
+++ b/migrations/versions/0391_add_go_live_columns.py
@@ -1,0 +1,26 @@
+"""
+
+Revision ID: 0391_add_go_live_columns
+Revises: 0390_require_rate_limit_value
+Create Date: 2022-12-19 08:01:58.001061
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0391_add_go_live_columns"
+down_revision = "0390_require_rate_limit_value"
+
+
+def upgrade():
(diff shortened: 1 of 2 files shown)


migrations/versions/0391_add_go_live_columns.py
```diff
@@ -0,0 +1,26 @@
...
+
+revision = "0391_add_go_live_columns"
+down_revision = "0390_require_rate_limit_value"
+
+
+def upgrade():
+    op.add_column("services", sa.Column("has_active_go_live_request", sa.Boolean(), nullable=True))
+    op.add_column("services_history", sa.Column("has_active_go_live_request", sa.Boolean(), nullable=True))
+
+    op.add_column("organisation", sa.Column("can_approve_own_go_live_requests", sa.Boolean(), nullable=True))
```
leohemsted, line 19: while we're here do we want to add some way to identify the org users that can approve go live requests? or were you planning to add that later

leohemsted (approved):

migrations/versions/0391_add_go_live_columns.py
quis: I guess we will start with all of them, since we don’t have a permissions model for organisations yet.

PR #3682, opened 2022-12-28
Add endpoint to notify organisation users of a go live request

quis:
This mostly copies the approach taken to Notify a service’s team members when a service goes live:

https://github.com/alphagov/notifications-api/blob/13ecbd888ee333204f4f6f85530bd45b28737f8e/app/service/rest.py#L292-L301

I’ve chosen to put this in the organisation blueprint because it’s ultimately emailing the team members for the whole organisation, even if it’s just about one service.

Merging this PR won’t do anything until the admin app starts calling the new endpoint.

This assumes that the admin app has done appropriate checks to make sure that the service is ready to go live, that it has an organisation, that the organisation can approve its own go live requests, etc.

***

Once triggered, the email will look like:

image


app/organisation/rest.py
```diff
@@ -284,6 +285,33 @@ def remove_letter_branding_from_organisation_pool(organisation_id, letter_brandi
     return {}, 204
 
 
+@organisation_blueprint.route("/notify-users-of-request-to-go-live/<uuid:service_id>", methods=["POST"])
+def notify_users_of_request_to_go_live(service_id):
+    template = dao_get_template_by_id(current_app.config["ORGANISATION_HAS_NEW_GO_LIVE_REQUEST_TEMPLATE_ID"])
+    service = dao_fetch_service_by_id(service_id)
+    organisation = service.organisation
+    notify_url = f"https://www.{current_app.config['NOTIFY_EMAIL_DOMAIN']}"
```
klssmith, line 293: `ADMIN_BASE_URL` can be used here instead of the email domain

app/organisation/sender.py
```diff
@@ -0,0 +1,34 @@
...
+    send_notification_to_queue,
+)
+
+
+def send_notification_to_organisation_users(
+    *, organisation, template, reply_to_text, personalisation=None, include_user_fields=None
+):
+    personalisation = personalisation or {}
+    include_user_fields = include_user_fields or []
+    active_users = dao_get_users_for_organisation(organisation.id)
```
klssmith, line 18: Just a question about how the feature will work - will it always be the case that any org user can approve requests? And likewise that you need to have a Notify account and be an org user in order to approve requests? I was wondering whether there might be org users who only want to view the billing, for example, who aren't interested in whether a service can go live or not

app/organisation/rest.py
quis: Changed in cad895b1

app/organisation/sender.py
quis: Yes, for any organisation which has the `can_make_own_services_live` flag set to `true`. In the future we will probably need an organisation permission model to support the different levels of access you are talking about.

migrations/versions/0393_add_go_live_template.py
```diff
@@ -0,0 +1,101 @@
...
+            current_timestamp,
+            '{current_app.config["NOTIFY_USER_ID"]}'
+        )
+        """
+    )
+
+
+def downgrade():
+    for table in ("notifications", "notification_history", "template_redacted", "templates", "templates_history"):
+        op.execute(f"DELETE FROM {table} WHERE template_id = '{template_id}'")
```
klssmith, line 101: This downgrade doesn't work for the `templates` and `templates_history` table where the column is just called `id` and not `template_id`.

migrations/versions/0393_add_go_live_template.py
```diff
@@ -0,0 +1,101 @@
...
+import textwrap
+
+from alembic import op
+from flask import current_app
+
+revision = "0393_add_go_live_template"
+down_revision = "0392_go_live_cols_non_nullable"
+
+
+template_id = current_app.config["ORGANISATION_HAS_NEW_GO_LIVE_REQUEST_TEMPLATE_ID"]
```
klssmith, line 17: In other migrations we've hardcoded values like these and the Notify service and user ID instead of importing them so that the migrations still work if we re-arrange the code. Could be worth doing here too (but I'll leave that decision up to you)

quis:
```suggestion
template_id = "5c7cfc0f-c3f4-4bd6-9a84-5a144aad5425"
```

migrations/versions/0393_add_go_live_template.py
quis: Good spot, fixed in [10e2576](https://github.com/alphagov/notifications-api/pull/3682/commits/10e2576faca9df3c04bede87b1e44f15e70c4f11)

klssmith (approved):

PR #3683, opened 2022-12-29
Update letter-checking task to ignore bank holidays

samuelhwilliams:
We've had the letter-checking task run the last few days, which opens a ticket with DVLA suggesting that they've fallen behind processing letters.

This isn't the case, as we've had two bank holidays this week. Let's respect these and properly look beyond them.

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -5,6 +5,10 @@ from flask import current_app
 from notifications_utils.clients.zendesk.zendesk_client import (
     NotifySupportTicket,
 )
+from notifications_utils.letter_timings import (
+    get_dvla_working_day_offset_by,
+    is_dvla_working_day,
+)
 from notifications_utils.timezones import convert_utc_to_bst
 from sqlalchemy import func
 from sqlalchemy.exc import SQLAlchemyError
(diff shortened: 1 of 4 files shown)


app/celery/nightly_tasks.py
```diff
@@ -216,16 +217,15 @@ def raise_alert_if_letter_notifications_still_sending():
 def get_letter_notifications_still_sending_when_they_shouldnt_be():
     today = datetime.utcnow().date()
 
-    # Do nothing on the weekend
-    if today.isoweekday() in {6, 7}:  # sat, sun
-        return 0, None
+    bank_holidays = BankHolidays()
```
quis, line 220: So we can change things in one place I’d use the definition from utils:

```suggestion
    bank_holidays = non_working_days_dvla
```

With `from notifications_utils.letter_timings import non_working_days_dvla`

samuelhwilliams: nice 💅

samuelhwilliams: Suggests I should probably be adding a `get_prev_dvla_working_day` to utils though. Love 2 PRs for the sake of 1 😓

quis (approved):

PR #3684, opened 2022-12-29
Select email brandings with ordering

samuelhwilliams:
It's good practice to provide specific ordering on queries. This also stops flakiness in the `test_zendesk_new_email_branding_report` test.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -398,6 +398,7 @@ def zendesk_new_email_branding_report():
             EmailBranding.created_at >= previous_weekday,
             User.platform_admin.is_(False),
         )
+        .order_by(EmailBranding.created_at)
         .all()
     )
 


CrystalPea (approved):

app/celery/scheduled_tasks.py
```diff
@@ -398,6 +398,7 @@ def zendesk_new_email_branding_report():
             EmailBranding.created_at >= previous_weekday,
             User.platform_admin.is_(False),
         )
+        .order_by(EmailBranding.created_at)
```
CrystalPea, line 401: we could order it first by user, and then by created at maybe? Then we would cluster brandings uploaded by the same user together 🤔

samuelhwilliams: I think, feel free to make that PR if you think it's useful, but I will not for now :) Just to get this in easily.

PR #3686, opened 2022-12-29
Change threshold for slow sms delivery and add more delivery metrics

idavidmcdonald:

## Required changes to other things

1. [Grafana panel for slow sms delivery](https://grafana-notify.cloudapps.digital/d/icjsQ-MWk2/notify-summary?viewPanel=51&orgId=1) will need relabelling to be 5 minutes rather than 4 minutes

2. New statsd mapping will be need adding to https://github.com/alphagov/notifications-aws/blob/main/paas/statsd/statsd-mapping.yml so the statsd metric can be picked up by prometheus and then added to some panels in Grafana

## 6 Dec slow delivery incident graph
image

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -118,7 +118,7 @@ def switch_current_sms_provider_on_slow_delivery():
     last ten minutes, then don't update them again either.
     """
     slow_delivery_notifications = is_delivery_slow_for_providers(
-        created_within_minutes=10,
+        created_within_minutes=15,
         delivered_within_minutes=5,
         threshold=0.15,
     )
@@ -136,7 +136,7 @@ def switch_current_sms_provider_on_slow_delivery():
 def generate_sms_delivery_stats():
     for delivery_interval in (1, 5, 10):
         providers_slow_delivery_ratios = get_ratio_of_messages_delivered_slowly_per_provider(
-            created_within_minutes=10, delivered_within_minutes=delivery_interval
+            created_within_minutes=15, delivered_within_minutes=delivery_interval
         )
 
         for provider, ratio in providers_slow_delivery_ratios.items():
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -117,9 +117,9 @@ def switch_current_sms_provider_on_slow_delivery():
     last ten minutes, then don't update them again either.
     """
     slow_delivery_notifications = is_delivery_slow_for_providers(
+        created_within_minutes=10,
+        delivered_within_minutes=5,
         threshold=0.15,
-        created_at=datetime.utcnow() - timedelta(minutes=10),
-        delivery_time=timedelta(minutes=5),
     )
 
     # only adjust if some values are true and some are false - ie, don't adjust if all providers are fast or
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -10,7 +10,7 @@ from notifications_utils.timezones import convert_utc_to_bst
 from sqlalchemy import between
 from sqlalchemy.exc import SQLAlchemyError
 
-from app import db, notify_celery, zendesk_client
+from app import db, notify_celery, statsd_client, zendesk_client
 from app.aws import s3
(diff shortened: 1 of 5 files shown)


app/dao/notifications_dao.py
```diff
@@ -448,28 +448,57 @@ def dao_timeout_notifications(cutoff_time, limit=100000):
...
+    last `created_within_minutes` minutes that took over
+    `delivered_within_minutes` minutes to be delivered
+
+    {
+        'mmg': 0.4,
+        'firetext': 0.12
+    }
     """
+    created_after = datetime.utcnow() - timedelta(minutes=created_within_minutes)
+    delivered_by = timedelta(minutes=delivered_within_minutes)
```
CrystalPea, line 492: For those two variables, I find the original names clearer.

app/celery/scheduled_tasks.py
```diff
@@ -129,6 +130,17 @@ def switch_current_sms_provider_on_slow_delivery():
                 dao_reduce_sms_provider_priority(provider_name, time_threshold=timedelta(minutes=10))
 
 
+@notify_celery.task(name="generate-sms-delivery-stats")
+def generate_sms_delivery_stats():
+    for delivery_interval in (1, 5, 10):
+        providers_slow_delivery_ratios = get_ratio_of_messages_delivered_slowly_per_provider(
+            created_within_minutes=10, delivered_within_minutes=delivery_interval
```
CrystalPea, line 137: I wonder, would it affect stats that we are looking for notifications that took longer than 10 minutes to deliver within a set of notifications that were created within last 10 minutes only?

CrystalPea (commented):
Interesting work, hopefully that will help us spot any future issues before our users do.

I left a couple comments.

app/dao/notifications_dao.py
idavidmcdonald: Ah thanks Pea, I did wonder if these were actually improved or not and wasn't quite happy with what I picked.

Out of interest, do you happen to have any alternatives for even better names? Equally I'm happy to change these back to what they were

app/celery/scheduled_tasks.py
idavidmcdonald: Oo great question Pea!

So I think you've spotted an issue there.

If `created_after` is 10 minutes ago, it means we only include notifications that were created less than or equal to 10 minutes ago.

There can't be a notification that is both
- created less than or equal to 10 minutes ago
- sent at equal to or older than 10 minutes ago

The only exception is a notification that is exactly 10 minutes old and still in sending.

I think there are two options here

1. Change the time period to consider notifications sent in say the last 15 minutes. For consistency, I'd suggest we do for all of the stats (1 min sending times, 5 min sending times, 10 min sending times)
2. Remove the 10 minute stat and just keep the 1 min and 5 min stat

I'm going to suggest we do option 1 for these.

What do you think Pea?

app/dao/notifications_dao.py
CrystalPea: I think `delivery_time` is very clear, and I wouldn't change it. `delivered_by` suggests a datetime, not a timedelta.

'created_after' is actually pretty alright. `created_since` might be even better, since we use `>=` operator, not ">"

app/celery/scheduled_tasks.py
CrystalPea: I also think suggestion 1 works best ☺️

app/dao/notifications_dao.py
idavidmcdonald: Thanks Pea, I've changed both to be your suggestions and squashed this change in as I don't think having distinct git history is useful for it

app/celery/scheduled_tasks.py
idavidmcdonald: Thanks Pea, I've added a commit on the end that implements option 1. Let me know what you think please!

app/dao/notifications_dao.py
CrystalPea: Thanks a lot!

CrystalPea (approved):
LGTM (after tests fixed ☺️  )

PR #3688, opened 2022-12-30
Add NewRelic to dev+preview

samuelhwilliams:
We add basic APM instrumentation for NewRelic. This will inject tracing
headers that filter through to any other instrumented apps and send them
to NewRelic so that we can see how requests travel through our services
and how long each span takes. This should help us highlight slower areas
of the codebase which we can then investigate and optimise.

We only send APM information - not logs - here, and we heavily limit the
information sent so that we hopefully aren't gathering much/any PII. We
also only deploy to dev and preview at the moment, so there's no risk of
this code gathering real user data for now. This will allow us to safely
investigate the tool and see whether it's something useful for us.

Required env vars:
`NEW_RELIC_LICENCE_KEY`

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,10 @@ export NOTIFICATION_QUEUE_PREFIX='YOUR_OWN_PREFIX'
 export FLASK_APP=application.py
 export FLASK_DEBUG=1
 export WERKZEUG_DEBUG_PIN=off
+
+export NEW_RELIC_ENABLED=[0|1]
+export NEW_RELIC_LICENSE_KEY=<ask for it>>
+export NEW_RELIC_APP_NAME=development.notify-api
 "> environment.sh
 ```
 
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -149,6 +149,9 @@ applications:
 
       DVLA_EMAIL_ADDRESSES: '{{ DVLA_EMAIL_ADDRESSES | tojson  }}'
 
+      NEW_RELIC_APP_NAME: '{{ environment }}.{{ CF_APP }}'
+      NEW_RELIC_LICENSE_KEY: '{{ NEW_RELIC_LICENSE_KEY }}'
+
       {% for key, value in app.get('additional_env_vars', {}).items() %}
       {{key}}: '{{value}}'
       {% endfor %}
--- a/newrelic.ini
+++ b/newrelic.ini
@@ -1,0 +1,215 @@
+# ---------------------------------------------------------------------------
+
+#
+# This file configures the New Relic Python Agent.
+#
+# The path to the configuration file should be supplied to the function
+# newrelic.agent.initialize() when the agent is being initialized.
+#
+# The configuration file follows a structure similar to what you would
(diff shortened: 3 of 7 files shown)


newrelic.ini
```diff
@@ -0,0 +1,220 @@
...
+[newrelic:preview]
+app_name = preview.notify-api
+monitor_mode = true
+
+[newrelic:staging]
+app_name = staging.notify-api
+monitor_mode = false
+
+[newrelic:production]
+app_name = production.notify-ai
```
leohemsted, line 217:
```suggestion
app_name = production.notify-api
```

manifest.yml.j2
```diff
@@ -149,6 +149,8 @@ applications:
 
       DVLA_EMAIL_ADDRESSES: '{{ DVLA_EMAIL_ADDRESSES | tojson  }}'
 
+      NEW_RELIC_LICENSE_KEY: '{{ NEW_RELIC_LICENSE_KEY }}'
```
leohemsted, line 152: could define this rather than relying on manually updating every line in paas_app_wrapper.sh

```suggestion
      NEW_RELIC_LICENSE_KEY: '{{ NEW_RELIC_LICENSE_KEY }}'
      NEW_RELIC_APP_NAME: '{{ CF_APP }}'
```

there's probably some thinking to do on if we care about having `notify` at the beginning (`notify-api`, `notify-delivery-worker-sender`, etc). i'd probably say yes so it's easier to match up with gunicorn, etc (and we've already defined the admin app as `notify-admin`, but it might just be boiler plate which makes it faffy to search on newrelic 🤔

samuelhwilliams: Have done this. We have to rip out `app_name` from newrelic.ini as the config file overrides environment variables (https://docs.newrelic.com/docs/apm/agents/python-agent/configuration/python-agent-configuration/#environment-variables)

README.md
```diff
@@ -34,6 +34,9 @@ export NOTIFICATION_QUEUE_PREFIX='YOUR_OWN_PREFIX'
 export FLASK_APP=application.py
 export FLASK_DEBUG=1
 export WERKZEUG_DEBUG_PIN=off
+
+export NEW_RELIC_LICENSE_KEY=<ask for it>>
+export NEW_RELIC_APP_NAME=development.notify-api
```
leohemsted, line 39: this'll mean your local celery instance will have the `development.notify-api` name 🤔 

i'd rather it had a separate app in newrelic for the tasks - not 100% sure of the best place to set that though, i'm not sure on how everyone calls celery locally

samuelhwilliams: Yeah I did realise this, but it's only locally, so ...

samuelhwilliams: It's going to be full of everyone's stuff anyway, I don't think - especially now when we're only exploring - we need to care about that local bucket much.

leohemsted: good point about it merging everyones. i guess if people want to enable it they should probably end up defining their own system `leo.notify-admin` etc etc

PR #3692, opened 2023-01-04
Find a record by UUID for admins

samuelhwilliams:
## Firebreak 'find a thing'

This endpoint will power a new page for platform admins to quickly jump to the best page for any* record in Notify with just a UUID, as in this little mock flow:


app/__init__.py
```diff
@@ -273,6 +274,9 @@ def register_blueprint(application):
     govuk_alerts_blueprint.before_request(requires_govuk_alerts_auth)
     application.register_blueprint(govuk_alerts_blueprint)
 
+    admin_blueprint.before_request(requires_admin_auth)
+    application.register_blueprint(admin_blueprint, url_prefix="/admin")
```
samuelhwilliams, line 278: Happy to mount this anywhere else.

I don't particularly like the layout of the API as it stands right now .. would love to have a chat around refactoring it. But that's for later. For now the `find-by-uuid` doesn't feel like it fits into any of our existing buckets, so we need a new one. We can either go for a generic 'admin' bucket, or we could have a `lookups` bucket? Or something else?

leohemsted: `/platform-admin`? i think `/admin` is super generic, and is immediately unclear about whether it refers to the notify-admin app or the notify team - who should rightfully be the only people who can use this endpoint

app/admin/rest.py
```diff
@@ -0,0 +1,103 @@
...
+}
+
+
+@admin_blueprint.route("/find-by-uuid", methods=["POST"])
+def find_by_uuid():
+    """Provides a simple interface for looking whether a UUID references any common DB objects"""
+    uuid_ = request.json["uuid"]
+
+    instance = None
+    for entity_name, model in FIND_BY_UUID_MODELS.items():  # noqa: B007 (entity name not used in loop)
```
leohemsted: im not a fan of us relying on the last value of an iterator after we've broken out of the loop - that's quite surprising control logic for me, and i easily missed the importance of using `.items` rather than `values`.

we could refactor this into a function instead? for example:

```py
def find_model_by_uuid(uuid_):
    for entity_name, model in FIND_BY_UUID_MODELS.items():
        # this is a nice alternative to except: pass even if we don't refactor the whole function
        with suppress(NoResultFound): 
            if instance := model.query.get(uuid_):
                return {'entity_name': entity_name, 'instance': instance}
    raise NoResultFound
```

app/__init__.py
samuelhwilliams: So, I had mounted it at platform-admin originally as well. However, it felt like a slight misnomer because it will be used by the admin app authenticating as the admin app (because the admin app doesn't pass through the user). So all of the authorization happens in the admin app invisible to the api. Mounting it at `platform-admin` makes me want to apply some permissions checks in the api that assert it's being called by a platform admin, but we can't do that (sidenote: I would like the admin app to pass the acting user through to the api eventually, so that the api can make assertions).

app/admin/rest.py
```diff
@@ -0,0 +1,103 @@
...
+    "service_callback_api": ServiceCallbackApi,
+    "complaint": Complaint,
+    "inbound_sms": InboundSms,
+}
+
+# The `context` here is actually information required by the admin app to build the redirect URL. This does mean
+# that this endpoint and the admin search page are fairly closely coupled. We could serialize the entire object
+# back, which might be more consistent with how other endpoints return representations of objects, but for many
+# objects we don't need any information at all, and our serialization of objects is already inconsistently between
+# `instance.serialize` and `some_schema.dump(instance)`. And not all of the serialized representations of objects
```
leohemsted: hmmm, i agree with this comment (that this is probably fine, but it would be nicer if it was less coupled). if we change a route in the admin app, how long will it be before we realise this endpoint has broken?

i'm surprised that not all the objects return a relevant service_id though! which ones are those? do they seem like an intentional decision or an oversight?

if we solve those issues, if we want to decouple the admin routes from this endpoint we could always just build a dict of the serialization functions.

```py
serialize_fns = {
  Service: Service.serialize,
  Template: template_schema.dump,
  ...
}
return serialize_fns[instance.__class__](instance)
```

don't feel too strongly though, and i'm sure that approach comes with its own set of foibles

tests/app/admin/test_rest.py
```diff
@@ -0,0 +1,112 @@
...
+    create_service_inbound_api,
+    create_service_with_inbound_number,
+    create_template_folder,
+)
+
+
+class TestFindByUUID:
+    def load_sample_data(self, request, model):
+        if model == Organisation:
+            request.getfixturevalue("sample_organisation")
```
leohemsted: ooh snazzy. not used this before

tests/app/platform_admin/test_rest.py
```diff
@@ -0,0 +1,115 @@
...
+        elif model == EmailBranding:
+            create_email_branding()
+        elif model == LetterBranding:
+            create_letter_branding()
+        elif model == User:
+            request.getfixturevalue("sample_user")
+        elif model == ProviderDetails:
+            request.getfixturevalue("sms_providers")
+        elif model == ServiceEmailReplyTo:
+            request.getfixturevalue("notify_service")
```
leohemsted, line 65: why's this returning the notify service? 🤔  i think i'd rather we created a sepcific reply to object on sample_service as we've done with some of the models below

app/admin/rest.py
samuelhwilliams: I think Template is the first thing I hit that didn't serialize service_id (explicitly!).

I did consider having a set of serialization functions, but I think that just obscures what's actually going on - suddenly we're just implicitly _expecting_ `service_id` to be sent, but the admin app will _require_ it. It feels more fragile.

This implementation is more verbose but also more explicit about the expectations and requirements, which I think is better.

app/admin/rest.py
samuelhwilliams: Yeah this is probably nicer, will change 👍

tests/app/platform_admin/test_rest.py
samuelhwilliams: TBH I just picked the first fixture I saw with a reply-to-email 🙄  Let's change it yes

leohemsted (approved):
dont forget to squash 😉

PR #3693, opened 2023-01-04
Notifications event log

samuelhwilliams:
This adds a new table to the DB for tracking 'events' that happen to notifications, so that we can eventually show this information in the admin UI (possibly only to platform admins) to help us understand how things are flowing and debug when things go wrong.

This patch mostly just tracks changes to notification statuses, as these are the key events that happen, but we also record detailed status codes when notifications fail, when a provider (and which provider) is chosen to deliver a given notification, and when we receive duplicate callbacks suggesting a conflicting state.

## Still to do (possibly in a separate PR)
* Nightly task to purge old events (from notifications created >x days ago). We definitely don't want to have this info forever, but it would be nice to keep for at least a few weeks. Possibly it could be the same amount of time as the service's data retention.
* API endpoints to allow admin to pull notification events for displaying.


app/dao/notifications_dao.py
```diff
@@ -118,6 +125,9 @@ def _update_notification_status(notification, status, detailed_status_code=None)
     )
     notification.status = status
     dao_update_notification(notification)
+    dao_record_notification_event(
+        notification, notes=f"detailed status code: {detailed_status_code}" if detailed_status_code else None
```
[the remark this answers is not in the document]
samuelhwilliams: Might be something to add in the future, but given firebreak is running short let's not scope creep into refactoring for now.

migrations/versions/0394_notification_event_log.py
```diff
@@ -0,0 +1,38 @@
...
+from sqlalchemy.dialects import postgresql
+
+revision = "0394_notification_event_log"
+down_revision = "0393_add_go_live_template"
+
+
+def upgrade():
+    op.create_table(
+        "notification_event_log",
+        sa.Column("id", sa.Integer(), nullable=False),
```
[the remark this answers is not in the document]
leohemsted: i'm concerned about the ramifications of integer overflow (and also performance in general suddenly 🤔). assuming 3 rows per email and 3.5 rows per sms (due to firetext having created->sending->pending->delivered)

in the last week according to performance stats we'll write ~72 million rows. we'll reach integer overflow (~2 bil) after 30 weeks.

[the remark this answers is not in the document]
samuelhwilliams: Oh yeah, we should at least use a big integer 🤔

leohemsted:
Places notifications are updated, and whether we capture those with a NotificationEventLog

- [x] notification creation
- [x] notification sending
- [x] sms callback code
- [ ] email callback code
- [ ] letter callback code
- [ ] timeout notifications

leohemsted (approved):
i'm happy with this code-wise but lets have a think about how we can run a performance test against this on staging before we merge

migrations/versions/0394_notification_event_log.py
```diff
@@ -0,0 +1,38 @@
...
+revision = "0394_notification_event_log"
+down_revision = "0393_add_go_live_template"
+
+
+def upgrade():
+    op.create_table(
+        "notification_event_log",
+        sa.Column("id", sa.BigInteger(), nullable=False),
+        sa.Column("notification_id", postgresql.UUID(as_uuid=True), nullable=True),
+        sa.Column("created_at", sa.DateTime(), nullable=True),
```
idavidmcdonald, line 21: I know we always call it `created_at` but I do wonder if we should instead have something like `event_datetime` instead to make it less confusing (but maybe not worth the divergence). One to ponder and seek other opinions if you were to consider

samuelhwilliams: That's a good shout 👍

idavidmcdonald: Maybe. I think the bigger decision is whether to diverge from consistency or not, for the name itself any of those seem like sensible options

samuelhwilliams: It's also definitely something we _could_ revisit at any point fairly easily so I'm not considering it a blocker on this PR for now.

leohemsted: i like the `_at` suffix for timestamps  and im fine with diverging from consistency considering that we may well be trying to match this timestamp with a timestamp given in an arbitrary field on the notification, so leaving it as created_at might be confusing. i like happened_at as a starter for ten

idavidmcdonald:
A small comment that is tangential to this. We have log messages for whenever we receive a delivery receipt for an SMS message. We don't have the equivalent for email. I've always wondered if we should (and pondered introducing that).

This new table might make that small gripe of inconsistent behaviour redundant!

samuelhwilliams:
Closing this down and will re-open in a fresh PR.

PR #3695, opened 2023-01-06
Apply new per-channel rate limits for new services

samuelhwilliams:
We're getting ready to apply the new per-notification-type rate limits. Let's start by applying them to any newly-created services. We'll update existing rate limits after this change has gone out.


samuelhwilliams:
Closing in favour of https://github.com/alphagov/notifications-admin/pull/4560

PR #3696, opened 2023-01-06
Migrate per-channel rate limits

samuelhwilliams:
We want to start applying the new per-channel rate limits soon. This migration will set the limits at the levels we want to apply:

For services in trial mode, 50 for each notification channel.

For live services, 250k for emails and SMS, and 20k for letters.

We only overwrite the limits if they're at the current default (insanely high) value, which means that any rate limits we've set manually for specific services won't be affected.

**Note**: need to merge+release https://github.com/alphagov/notifications-admin/pull/4560 first

### DIFF
--- a/migrations/versions/0395_migrate_rate_limits.py
+++ b/migrations/versions/0395_migrate_rate_limits.py
@@ -1,0 +1,96 @@
+"""
+
+Revision ID: 0395_migrate_rate_limits
+Revises: 0394_letter_branding_cols
+Create Date: 2023-01-06 09:27:38.125105
+
+"""
+from alembic import op
+
+revision = "0395_migrate_rate_limits"
+down_revision = "0394_letter_branding_cols"
+
+
+def upgrade():
+    op.execute(
+        """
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
Ready to review, but not ready to merge until we've got everything lined up.

klssmith:
The migration number will need to be updated since I've merged one with the same number, but this worked as expected when I tried it out 👍

migrations/versions/0395_migrate_rate_limits.py
```diff
@@ -0,0 +1,96 @@
+"""
+
+Revision ID: 0395_migrate_rate_limits
+Revises: 0393_add_go_live_template
```
klssmith, line 4:
```suggestion
Revises: 0394_letter_branding_cols
```

samuelhwilliams: such a useless comment

klssmith (approved):

PR #3697, opened 2023-01-06
Add new columns to letter branding table

klssmith:
This adds new nullable columns to the letter branding table for auditing and allows them to be used when updating and creating a brand:
* created_by_id
* created_at
* updated_by_id
* updated_at

### DIFF
--- a/tests/app/letter_branding/test_letter_branding_rest.py
+++ b/tests/app/letter_branding/test_letter_branding_rest.py
@@ -1,24 +1,19 @@
-import json
-import uuid
-
 import pytest
 
 from app.models import LetterBranding
-from tests import create_admin_authorization_header
 from tests.app.db import create_letter_branding
 
 
-def test_get_all_letter_brands(client, notify_db_session):
+def test_get_all_letter_brands(admin_request, notify_db_session):
     hm_gov = create_letter_branding()
     test_branding = create_letter_branding(
         name="test branding",
         filename="test-branding",
     )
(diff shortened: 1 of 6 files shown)


tests/app/letter_branding/test_letter_branding_rest.py
```diff
@@ -27,53 +22,110 @@ def test_get_all_letter_brands(client, notify_db_session):
...
     )
 
-    assert response.status_code == 201
-    json_response = json.loads(response.get_data(as_text=True))
-    letter_brand = LetterBranding.query.get(json_response["id"])
+    letter_brand = LetterBranding.query.get(response["id"])
     assert letter_brand.name == form["name"]
     assert letter_brand.filename == form["filename"]
+    assert letter_brand.updated_by_id is None
+    assert letter_brand.updated_at
```
samuelhwilliams, line 110: Should we assert the actual timestamp here (with freezegun)?

samuelhwilliams (approved):
One comment that'd be nice to address, otherwise looks good 👍

tests/app/letter_branding/test_letter_branding_rest.py
klssmith: I did consider that when I was writing this, but then decided I was happy with the tests as they are (checking the values have been set without testing the exact values). We mostly don't test these fields at all for other models and testing the exact values would require a change from the `default=datetime.datetime.utcnow` and `onupdate=datetime.datetime.utcnow` way we have of writing things in the models that I prefer.

PR #3698, opened 2023-01-11
Fix zendesk email branding - flaky test and BST bug

leohemsted:
The flaky test was solved by adding `tick=true` to ensure that the timestamps of the two email brandings are different, and thus ordering is respected.

Then I solved a BST bug we would have discovered in march when it would have picked up the wrong branding in the zendesk report and always been a day out.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -6,6 +6,7 @@ from flask import current_app
 from notifications_utils.clients.zendesk.zendesk_client import (
     NotifySupportTicket,
 )
+from notifications_utils.timezones import convert_utc_to_bst
 from sqlalchemy import between
 from sqlalchemy.exc import SQLAlchemyError
 
@@ -66,6 +67,7 @@ from app.models import (
     User,
 )
 from app.notifications.process_notifications import send_notification_to_queue
+from app.utils import get_london_midnight_in_utc
 
 
 @notify_celery.task(name="run-scheduled-jobs")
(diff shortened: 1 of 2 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -385,24 +387,26 @@ def delete_old_records_from_events_table():
 
 @notify_celery.task(name="zendesk-new-email-branding-report")
 def zendesk_new_email_branding_report():
-    previous_weekday = datetime.today().date() - timedelta(days=1)
+    previous_bst_weekday = convert_utc_to_bst(datetime.utcnow()).date() - timedelta(days=1)
```
samuelhwilliams, line 390: Don't think this is the best name, because it's wrong half of the year when we're not in BST.

I think `previous_weekday` is fine. If we need to add a comment saying that we're checking for daylight savings let's do that.

samuelhwilliams (approved):
Minor nit but otherwise looks good, nice fixing 👍

app/celery/scheduled_tasks.py
leohemsted: hmmm we do have a habit (of referring to `Europe/London` as "bst". See `bst_date` column in ft_notification_status, and indeed the `convert_utc_to_bst`. Perhaps bst isn't the best word in conjunction with date (rather than datetime) but i'll have a think.

samuelhwilliams: If we do it elsewhere, then ok.

Weird tho.

PR #3699, opened 2023-01-12
Remove research mode

leohemsted:
We no longer read this setting on the admin app[^1], we can remove it from the API now.

Process:

1. Make column nullable
1. Remove references to it in code
1. Drop column

This PR does the first two. Note there's still a bunch of references to "research mode`, mostly because there's a queues

This PR's also blocked based on us doing a quick audit to ensure we're not relying on research mode - not just on prod (lets check seeded functional test stuff too)

[^1]: https://github.com/alphagov/notifications-admin/pull/4565


tests/app/celery/test_process_ses_receipts_tasks.py
```diff
@@ -38,16 +38,6 @@ def test_process_ses_results_retry_called(sample_email_template, mocker):
     assert mocked.call_count != 0
 
 
-def test_process_ses_results_in_complaint(sample_email_template, mocker):
```
klssmith, line 41: I couldn't see another test for this scenario of a standard SES callback creating a complaint, so perhaps we'd want to just rewrite this instead of deleting it completely? The `process_ses_results` doesn't call `update_notification_status_by_reference`, so the check that that function isn't being called might be left over from some old code anyway

leohemsted: Good spot! I've tweaked that commit to keep this test in (and assert the notification isn't updated at all)

klssmith (approved):

PR #3700, opened 2023-01-13
Archive email branding

CrystalPea:
In this PR we:

- add "active" column for email_branding table
- check orgs and services associated with email branding to see if it can be archived
- remove branding from branding pools
- archive branding
- stop showing archived brands in lists of brandings


leohemsted (approved):
just need to change the migration number

i was a bit uncertain about if the migration would update all the old values to be `true` just based on the server default directive, but i tested locally and it works

PR #3701, opened 2023-01-17
Update SQLAlchemy engine configuration

samuelhwilliams:
In flask-sqlalchemy 2.4.0, the separate SQLALCHEMY_POOL_* options were deprecated and replaced with SQLALCHEMY_ENGINE_OPTIONS. In version 3.0.0 the deprecated options were removed, but we didn't update this configuration, so we've been running with only 5 connections (spiking up to 15 max) since. This fixes the configuration so that each instance will hold up to 15 connections in a pool, and then optionally draw up to 10 more (the default) under load.

https://github.com/pallets-eco/flask-sqlalchemy/blob/3.0.x/CHANGES.rst#version-240

---

Just throwing this up to track - there might be other stuff we need to fix from the move to 3.0.0


app/config.py
```diff
@@ -484,6 +488,15 @@ class Test(Development):
...
     SQLALCHEMY_RECORD_QUERIES = False
 
+    SQLALCHEMY_ENGINE_OPTIONS = {
+        "pool_size": 10,
+        "pool_timeout": 31,
+        "pool_recycle": 301,
+        "connect_args": {
+            "options": "-c statement_timeout=1200000",
+        },
+    }
```
idavidmcdonald, line 498: How come you wanted to override `SQLALCHEMY_ENGINE_OPTIONS` for the `Test` config class? `Test` inherits from `Config` where the originals where added to, couldn't we just let it use those?

I guess firstly, is there a particular benefit of overriding?

And secondly, is there a risk then that we misconfigure the values/dictionary for the `Config` class but won't know because we may not have also changed the `Test` class with a similar misconfiguration? ie should we let our tests check the default configuration for `SQLALCHEMY_ENGINE_OPTIONS` because it will give better coverage of the code that is going into production

idavidmcdonald (approved):
One comment to address but either nothing will change, or the change will be very minor so happy to give the green tick!

app/config.py
samuelhwilliams: So .. I wanted to override for test to make sure that we're not just testing the default values that sqlalchemy provides and the config doesn't actually do anything. For example, the [default `pool_timeout` is 30](https://docs.sqlalchemy.org/en/20/core/pooling.html#sqlalchemy.pool.QueuePool.params.timeout). Maybe that means we just shouldn't have it in config at all, though 🤔

I could definitely take it or leave.

samuelhwilliams: I did indeed go with a slight change. Removed the test config and made the test check sqlalchemy's internal settings against the config we've specified. This should catch the important stuff.

PR #3704, opened 2023-01-19
Fix some auto-generated migrations

samuelhwilliams:
When we make model changes it's nice to be able to use alembic to autogenerate the migration changes.

Unfortunately for a while now it's been generating instructions unrelated to the changes a dev picks up, because things have fallen out of sync over time.

This removes all of the differences except 2 indexes, which are slightly trickier to unpick because we haven't got naming conventions for indexes (see: https://alembic.sqlalchemy.org/en/latest/naming.html).

It also leaves alone the `research_mode` columns that are addressed here: https://github.com/alphagov/notifications-api/pull/3703

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2238,7 +2238,7 @@ class BroadcastMessage(db.Model):
     template = db.relationship("TemplateHistory", backref="broadcast_messages")
 
     _personalisation = db.Column(db.String, nullable=True)
-    content = db.Column(db.String, nullable=False)
+    content = db.Column(db.Text, nullable=False)
     # defaults to empty list
     areas = db.Column(JSONB(none_as_null=True), nullable=False, default=list)
 
--- a/app/models.py
+++ b/app/models.py
@@ -2531,6 +2531,12 @@ class BroadcastProviderTypes(db.Model):
     name = db.Column(db.String(255), primary_key=True)
 
 
+class BroadcastProviderMessageStatusType(db.Model):
+    __tablename__ = "broadcast_provider_message_status_type"
+
+    name = db.Column(db.String(), primary_key=True)
+
+
 class ServiceBroadcastProviderRestriction(db.Model):
     """
     TODO: Drop this table as no longer used
--- a/app/models.py
+++ b/app/models.py
@@ -1875,7 +1875,7 @@ class Rate(db.Model):
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     valid_from = db.Column(db.DateTime, nullable=False)
-    rate = db.Column(db.Float(asdecimal=False), nullable=False)
+    rate = db.Column(db.Numeric(asdecimal=False), nullable=False)
     notification_type = db.Column(notification_types, index=True, nullable=False)
 
     def __str__(self):
--- a/app/models.py
+++ b/app/models.py
@@ -1395,6 +1395,10 @@ class NotificationAllTimeView(db.Model):
 
     __tablename__ = "notifications_all_time_view"
 
+    # Tell alembic not to create this as a table. We have a migration where we manually set this up as a view.
+    # This is custom logic we apply - not built-in logic. See `migrations/env.py`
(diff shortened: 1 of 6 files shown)


app/models.py
```diff
@@ -1390,6 +1390,10 @@ class NotificationAllTimeView(db.Model):
 
     __tablename__ = "notifications_all_time_view"
 
+    # Tell alembic not to create this as a table. We have a migration where we manually set this up as a view.
+    # This is custom logic we apply - not built-in logic. See `migrations/env.py`
+    __table_args__ = {"info": {"managed": False}}
```
leohemsted, line 1395: i don't love `managed` here, i dont think it's immediately apparent that that refers to alembic. how about `managed_by_alembic`?

migrations/versions/0397_missed_steps.py
```diff
@@ -0,0 +1,21 @@
...
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0397_missed_steps"
+down_revision = "0396_make_research_mode_nullable"
+
+
+def upgrade():
+    # This was created as `sa.String` in the migration but defined in Python as `db.Text`.
+    op.alter_column("broadcast_message", "content", existing_type=sa.TEXT(), type_=sa.String(), existing_nullable=False)
```
leohemsted, line 17: given text and varchar without a length are functionally equivalent in postgres, i think i'd rather update the model to use `db.Text` and save on doing any migration 🤔. That feels less risky (not that I'm worried about this particular alter statement especially).

(really i'd rather we unified all our usage of `db.String` (~73), `db.String(255)` (~42), and `db.Text` (~22) and just chose one approach, probably `db.Text` for anything unbounded and `db.String(n)` for stuff we want to explicitly guard against, which at the mo is some but not all mobile number fields and the hex colour for meail branding

migrations/versions/0398_non_nullable_international.py
```diff
@@ -0,0 +1,20 @@
...
+"""
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0398_non_nullable_international"
+down_revision = "0397_missed_steps"
+
+
+def upgrade():
+    op.alter_column("notifications", "international", existing_type=sa.BOOLEAN(), nullable=False)
```
leohemsted, line 16: https://www.postgresql.org/docs/current/sql-altertable.html

> SET NOT NULL may only be applied to a column provided none of the records in the table contain a NULL value for the column. Ordinarily this is checked during the ALTER TABLE by scanning the entire table; however, if a valid CHECK constraint is found which proves no NULL can exist, then the table scan is skipped.

I'm worried that, as we don't have a matching check constraint, this will require alembic to scan the entire table to check for existance of null values.

Postgres acquires an `ACCESS EXCLUSIVE` lock for most alter statements, including this one. I'd definitely like us to ensure we can avoid any table scans while holding such a strict lock

side note: does alembic run each migration script in a separate transaction with commits between? I assume it does.

leohemsted (changes requested):
Think we need to figure out a plan for the `notifications` table if we do want to proceed with this

migrations/versions/0398_non_nullable_international.py
samuelhwilliams: Humm, yep, that one might be trickier.

re: side note - Alembic will run all of the (new) migrations in a single transaction: https://github.com/alphagov/notifications-api/blob/main/migrations/env.py#L64-L65

samuelhwilliams: Will take this migration out for now. We should be able to do something by adding an invalid check constraint, triggering a manual validation, and then doing the migration, but it's faffy...

leohemsted: good to know about the single transaction 👍.

but yeah, i agree with solving it in a separate PR. we have patterns in past migrations for adding a constraint with the `not valid` flag, then calling `validate constraint` separately which acquires a much weaker lock

leohemsted (approved):

PR #3707, opened 2023-01-24
Lookup notification by reference

samuelhwilliams:
Extend the 'find-by-uuid' platform admin endpoint to also lookup notifications by reference. We rename the endpoint to not specifically call out 'uuid', and change the expected data key from uuid->value so it's more generic as well.

We mount the endpoint at a new route, `platform-admin/find` to remove `-by-uuid` from the endpoint as well.


samuelhwilliams:
Actually not sure this immediately fits on this endpoint, references aren't unique ... will leave this extension for now

PR #3709, opened 2023-01-25
Dvla auth

leohemsted:
Not tested locally quite yet cos I'm not sure about local IAM permissions but i'll sort that out before merging

this key thing this adds is a `jwt_token` property that can be accessed to return the jwt token. i envisage this property in a future PR handling re-auth if needed. It will possibly (probably?) handle password and api key rotation if needed too. But this will unblock work on the print API

### DIFF
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -1,4 +1,7 @@
+import time
+
 import boto3
+import jwt
 import requests
 from flask import current_app
 
@@ -25,6 +28,7 @@ class DVLAClient:
     statsd_client = None
 
     _jwt_token = None
+    _jwt_expires_at = None
 
     def init_app(self, region, statsd_client):
         ssm_client = boto3.client("ssm", region_name=region)
@@ -41,8 +45,10 @@ class DVLAClient:
 
     @property
     def jwt_token(self):
-        if not self._jwt_token:
+        if not self._jwt_token or time.time() >= self._jwt_expires_at:
             self._jwt_token = self.authenticate()
+            jwt_dict = jwt.decode(self._jwt_token, options={"verify_signature": False})
+            self._jwt_expires_at = jwt_dict["exp"]
 
         return self._jwt_token
 
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -1,4 +1,6 @@
 import boto3
+import requests
+from flask import current_app
 
 
 class SSMParameter:
@@ -22,6 +24,8 @@ class DVLAClient:
 
     statsd_client = None
 
+    _jwt_token = None
+
     def init_app(self, region, statsd_client):
         ssm_client = boto3.client("ssm", region_name=region)
         self.dvla_username = SSMParameter(key="/notify/api/dvla_username", ssm_client=ssm_client)
@@ -29,10 +33,33 @@ class DVLAClient:
         self.dvla_api_key = SSMParameter(key="/notify/api/dvla_api_key", ssm_client=ssm_client)
 
         self.statsd_client = statsd_client
+        self.request = requests.Session()
 
     @property
     def name(self):
         return "dvla"
+
+    @property
+    def jwt_token(self):
+        if not self._jwt_token:
+            self._jwt_token = self.authenticate()
+
+        return self._jwt_token
+
+    def authenticate(self):
+        """
+        Fetch a JWT from the DVLA API that can be used in other DVLA API requests
+        """
+        response = self.request.post(
+            f"{current_app.config['DVLA_API_BASE_URL']}/thirdparty-access/v1/authenticate",
+            json={
+                "userName": self.dvla_username.get(),
+                "password": self.dvla_password.get(),
+            },
+        )
+        response.raise_for_status()
+
+        return response.json()["id-token"]
 
     def send_letter(self):
         pass
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -1,0 +1,38 @@
+import boto3
+
+
+class SSMParameter:
+    def __init__(self, key, ssm_client):
+        self.key = key
+        self.ssm_client = ssm_client
+
+    def get(self):
+        return self.ssm_client.get_parameter(Name=self.key, WithDecryption=True)["Parameter"]["Value"]
+
+    def set(self, value):
+        # this errors if the parameter doesn't exist yet in SSM as we haven't supplied a `Type`
+        # this is fine for our purposes, as we'll always want to pre-seed this data.
+        self.ssm_client.put_parameter(Name=self.key, Value=value, Overwrite=True)
+
+
+class DVLAClient:
+    """
+    DVLA HTTP API letter client.
+    """
+
+    statsd_client = None
+
+    def init_app(self, region, statsd_client):
+        ssm_client = boto3.client("ssm", region_name=region)
+        self.dvla_username = SSMParameter(key="/notify/api/dvla_username", ssm_client=ssm_client)
+        self.dvla_password = SSMParameter(key="/notify/api/dvla_password", ssm_client=ssm_client)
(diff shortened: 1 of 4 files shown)


leohemsted:
things to do in this pr:

* make sure terraform, ssm variable names, etc are good
* decide how we're going to ever run this stuff locally (options: use staging creds? use a new set of creds that also point to UAT? each dev creates a new set of creds for UAT?)

things to add in a future pr: 

* caching SSM values locally so we don't hit SSM every time (look at `TTLCache`)
* handling 401 (which is probably caused by the password being changed)

app/clients/letter/dvla.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def __set__(self, obj, value):
+        # this errors if the parameter doesn't exist yet in SSM as we haven't supplied a `Type`
+        # this is fine for our purposes, as we'll always want to pre-seed this data.
+        obj.ssm_client.put_parameter(Name=self.key, Value=value, Overwrite=True)
+
+
+class DVLAClient:
+    """
+    Amazon SES email client.
```
samuelhwilliams, line 24: 👀

app/clients/letter/dvla.py
```diff
@@ -0,0 +1,71 @@
...
+    dvla_api_key = SSMProperty(key="/notify/api/dvla_api_key")
+
+    def init_app(self, region, statsd_client):
+        self.ssm_client = boto3.client("ssm", region_name=region)
+        self.statsd_client = statsd_client
+        self.request = requests.Session()
+
+    @property
+    def name(self):
+        return "dvla"
```
samuelhwilliams, line 44: What's this for? 🤔

app/config.py
```diff
@@ -555,6 +561,8 @@ class Production(Config):
 
     CRONITOR_ENABLED = True
 
+    DVLA_API_BASE_URL = "https://driver-vehicle-licensing.api.gov.uk"
```
samuelhwilliams, line 564: These seem inconsistent with how we manage URLs generally. Should we just have `DVLA_API_BASE_URL` defined on the base config class with a default of the uat URL, and pass in an env var to each of deployed apps to set this correctly?

app/clients/letter/dvla.py
leohemsted: this was for consistency with the MMGClient, FiretextClient, and AWSSesClient: we have a dict where we look-up by name here:

https://github.com/alphagov/notifications-api/blob/075d0c92f4366da666893730f32ee6923434b7f4/app/clients/__init__.py#L22

I'm not 100% sold on this being necessary for letters though

samuelhwilliams:
Holding off on further review as I can't tell from the PR description if this is actually ready to be looked at or if you're still working on it?

app/config.py
leohemsted: hmmm yes, you're right, this is inconsistent. i thought it was how we did it elsewhere but checking now it's clearly not. 

With that said, I prefer having this in code - these aren't secret, it's overkill to hide them in a gpg encrypted pass store, and I think it's a nicer dev experience to be able to quickly check what URL our envs are using without having to access secrets. Thoughts?

samuelhwilliams: This is configuration, not code, so shouldn't require a deploy to change. I think, just because we happen to store all of our configuration gpg encrypted, it doesn't mean that non-secret config should be hard-coded instead.

I do agree it's nice to be able to see the URL easily, but I think that points to our gpg-encryption-for-all-config being the wrong thing rather than this being the right thing.

leohemsted: Hmmmm

* In the current setup we have two options: define in cred store or define in code. We've gone with define in cred store for other URLs, but we do have other flags defined in config.py: feature flags, default "from number", bucket names, etc. In some other repos we have some URLs defined in config (admin's ASSSET_DOMAIN, template preview's LETTER_LOGO_URL), but we also have ones set in the environment. document-download-api also has its hostname defined in the manifest.yml.j2 and then passed through as an environment variable.
* In the current setup, I prefer having it in the repo than in gpg for the reasons I listed above, but i appreciate that's a personal view
* This story is not about changing the current setup. We might want to consider that in the future but there's probably less incentive to look before we move off of paas?

Ultimately the decision at hand is between two flawed practices that we currently have available. I think that across our whole codebase we're not very consistent anyway so that pushes me away from storing this in credentials, but i'm not against it if you think it's better

leohemsted:
i can't respond to your comment in-line cos of github but yes, this is ready to be looked at (although not merged until the relevant terraform has gone in)

app/config.py
samuelhwilliams: Yep yep. As I say, just sharing my thoughts - not blocking you on this point.

app/clients/letter/dvla.py
```diff
@@ -0,0 +1,71 @@
...
+    @property
+    def jwt_token(self):
+        if not self._jwt_token or time.time() >= self._jwt_expires_at:
+            self._jwt_token = self.authenticate()
+            jwt_dict = jwt.decode(self._jwt_token, options={"verify_signature": False})
+            self._jwt_expires_at = jwt_dict["exp"]
+
+        return self._jwt_token
+
+    def send_letter():
```
samuelhwilliams, line 55:
```suggestion
    def send_letter(self):
```

app/clients/letter/dvla.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, key):
+        self.key = key
+
+    def __get__(self, obj, objtype=None):
+        return obj.ssm_client.get_parameter(Name=self.key, WithDecryption=True)["Parameter"]["Value"]
+
+    def __set__(self, obj, value):
+        # this errors if the parameter doesn't exist yet in SSM as we haven't supplied a `Type`
+        # this is fine for our purposes, as we'll always want to pre-seed this data.
+        obj.ssm_client.put_parameter(Name=self.key, Value=value, Overwrite=True)
```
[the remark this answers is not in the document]
leohemsted: i definitely think caching should come in, 100%, since this pr is hanging around a little longer than i intended i'll add it in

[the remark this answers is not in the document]
samuelhwilliams: I don't think there's a good naming solution here. Just use proper function/methods that clearly say "work is happening". Especially since SSM is (currently) not a thing that we use much, I think hiding the fact that we're read/writing to an actual secrets store is not the right thing to do.

samuelhwilliams (approved):
Couple of tiny fixes needed but then I think this looks good 👍 

The more I think about it the more I think the TTL on the cached info isn't particularly useful but we can revisit that in a future thing if we need to.

Test coverage looks good 👏 

```diff
diff --git a/app/clients/letter/dvla.py b/app/clients/letter/dvla.py
index 36ec045b7..8602eaa4b 100644
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -43,10 +43,6 @@ class DVLAClient:
     _jwt_token = None
     _jwt_expires_at = None
 
-    dvla_username = SSMProperty(key="/notify/api/dvla_username")
-    dvla_password = SSMProperty(key="/notify/api/dvla_password")
-    dvla_api_key = SSMProperty(key="/notify/api/dvla_api_key")
-
     def init_app(self, region, statsd_client):
         ssm_client = boto3.client("ssm", region_name=region)
         self.dvla_username = SSMParameter(key="/notify/api/dvla_username", ssm_client=ssm_client)
@@ -76,8 +72,8 @@ class DVLAClient:
         response = self.request.post(
             f"{current_app.config['DVLA_API_BASE_URL']}/thirdparty-access/v1/authenticate",
             json={
-                "userName": self.dvla_username,
-                "password": self.dvla_password,
+                "userName": self.dvla_username.get(),
+                "password": self.dvla_password.get(),
             },
         )
         response.raise_for_status()
```

leohemsted:
oops, changes got partially borked in rebase hell! squashed back out

PR #3711, opened 2023-01-26
Stop sending notifications to the ‘priority’ queue

quis:
Originally we had the idea that we could set certain templates (for example two factor codes) as ‘priority’ and try to deliver them more quickly.

We only ever let platform admins decide whether a template was priority.

Since then we discovered that:
- Celery 3 was slower to deliver messages from an often-empty queue (which the priority queue was) so this feature had the opposite effect to what was intended
- Celery 4 is so fast it obviates the need for a priority queue
- We have so many templates that it’s beyond the capacity of our team to review which ones should or shouldn’t be ‘priority’

By not sending any notifications to the priority queue, we can safely remove (in a followup PR):
- the queue
- the column on the templates table and model

This will affect about 10 templates in production which belong to real, live services.

### DIFF
--- a/app/notifications/rest.py
+++ b/app/notifications/rest.py
@@ -2,16 +2,9 @@ from flask import Blueprint, current_app, jsonify, request
 from notifications_utils import SMS_CHAR_COUNT_LIMIT
 
 from app import api_user, authenticated_service
-from app.config import QueueNames
 from app.dao import notifications_dao
 from app.errors import InvalidRequest, register_errors
-from app.models import (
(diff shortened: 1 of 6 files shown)


leohemsted (approved):
🚮

idavidmcdonald:
I haven't looked at the code but this is dreamy. Glad we are getting rid of this. Are we doing/done the equivalent removal in the admin app?

quis:
@idavidmcdonald yup – see https://github.com/alphagov/notifications-admin/pull/4567

PR #3713, opened 2023-02-01
Change email branding report page link

samuelhwilliams:
We've added useful information to the email branding view (not edit) page about whether it's an org default or in use by a service. This feels like more useful overview information to see when reviewing the branding as opposed to immediately seeing the edit branding form.

Probably optional and could be less useful if/when we merge https://github.com/alphagov/notifications-admin/pull/4608

## Old target page
image

## New target page
image

### DIFF
--- a/templates/tasks/scheduled_tasks/new_email_brandings.html
+++ b/templates/tasks/scheduled_tasks/new_email_brandings.html
@@ -5,7 +5,7 @@
 {%- for org, brands in brands_by_organisation.items() -%}
     <p><a href="{{ domain }}/organisations/{{ org.id }}/settings/email-branding">{{ org.name }}</a>{% if not org.email_branding_id %} (no default){% endif %}:</p><ul>
     {%- for brand in brands -%}
-        <li><a href="{{ domain }}/email-branding/{{ brand.id }}/edit">{{ brand.name }}</a></li>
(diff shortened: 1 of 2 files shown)


klssmith:
I prefer the original link destination because a branding that has just been uploaded can only be in use by the service that uploaded it and won't be in any branding pools, so I think the branding view page is only useful for existing brands.

samuelhwilliams:
Probably true that it won't be in use by any other services, although we do automatically put it into the pool when brands are created now, don't we? That's part of why we have the 'review email brands' report - to decide if we should take it _out_ of the pool.

samuelhwilliams:
(But definitely happy to drop this PR - threw it up very much to gauge opinion because it's only a tiny change)

klssmith:
Yep, I got it wrong about it not being in branding pools - it will go in the pool automatically, but it would only be the org default if there wasn't already an org default, so it wouldn't necessarily appear under the "Organisations using this branding as their default" section.

We link to the organisation from the Zendesk ticket, so can remove click there to remove it from the pool.

Let's see what others think about the change anyway

CrystalPea (approved):
Sounds good to me, also means we see a preview of the branding in an example email before we get to editing it.

PR #3714, opened 2023-02-01
Only expose port 6011 when running the API

samuelhwilliams:


klssmith:
The first commit worked and let me run both the Flask app and Celery with Docker at the same time 🎉
I get an error (with no message) when including the second commit though, maybe because I'm not using AWS with temporary sessions when running this

klssmith (approved):

PR #3717, opened 2023-02-01
Automate DWP report for zendesk

samuelhwilliams:
Adds a task which runs weekly to generate a report for DWP on SMS notifications sent in the last week. The query and zendesk ticket ID are provided through environment variables. We will (🤞) review this way of doing things in Q1 '23 as part of a wider look at reporting.

I've made it so that the entire query for the DWP reject is injected via an environment variable because:
1) It means we can tweak the query without doing a deploy
2) It felt odd to hard-code the query because it contains some service-specific info (the service ID). We could have just injected the service ID via env variables, but that also felt weird and we'd have to make code changes if we altered the query to include some other identifier.
3) Using a dict means we can also provide the report filename as the key

## Todo
 - [x] add the env vars to prod

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -492,13 +492,18 @@ def check_for_low_available_inbound_sms_numbers():
 
 @notify_celery.task(name="weekly-dwp-report")
 def weekly_dwp_report():
-    report_config = current_app.config["ZENDESK_REPORTING"]["weekly-dwp-report"]
+    report_config = current_app.config["ZENDESK_REPORTING"].get("weekly-dwp-report")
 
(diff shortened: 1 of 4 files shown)


klssmith (approved):
I haven't been able to test this locally, but we can see it in action on Monday

samuelhwilliams:
@klssmith sorry, I pushed a couple of tweaks. The config was on `Development` rather than the base class.

And after looking at how to get the config into the environment via credentials there's some funky decoding/encoding that needs to happen to unescape unicode. The corresponding creds PR is here: https://github.com/alphagov/notifications-credentials/pull/312

klssmith:
👍 I've tested building the file with the creds in the other PR and it all worked

PR #3719, opened 2023-02-10
Add `send_letter` method to the DVLA client

klssmith:
The `send_letter` method now calls the DVLA endpoint to create a print job. All the data needed by the method is passed to it as arguments to avoid the method needing to access the database or S3. International letters are not supported yet.

We raise custom errors based on the DVLA response, but these will be handled in a later story.

### DIFF
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -1,3 +1,4 @@
+import base64
 import secrets
 import string
 import time
@@ -142,5 +143,84 @@ class DVLAClient:
                 return password
         raise RuntimeError("Unable to generate sufficiently secure password")
 
-    def send_letter(self):
-        pass
+    def _get_auth_headers(self):
+        return {
+            "Accept": "application/json",
+            "Authorization": self.jwt_token,
+            "X-API-Key": self.dvla_api_key.get(),
+        }
+
+    def send_letter(
+        self,
+        *,
+        notification_id: str,
+        address: list[str],
+        postage: str,
+        service_id: str,
+        organisation_id: str,
+        pdf_file: bytes,
+    ):
+        """
+        Sends a letter to the DVLA for printing
+
+        address should be normalised address lines, e.g. ['A. User', 'London', 'SW1 1AA']
+
+        normalised address lines can be returned by calling:
+        `PostalAddress.from_personalisation(notification.personalisation).normalised_lines`
+        """
+        from app.models import INTERNATIONAL_POSTAGE_TYPES
+
+        if postage in INTERNATIONAL_POSTAGE_TYPES:
+            raise NotImplementedError
+
+        response = requests.post(
+            f"{current_app.config['DVLA_API_BASE_URL']}/print-request/v1/print/jobs",
+            headers=self._get_auth_headers(),
+            json=self._format_create_print_job_json(
+                notification_id, address, postage, service_id, organisation_id, pdf_file
+            ),
+        )
+
+        return response.json()
+
+    def _format_create_print_job_json(
+        self, notification_id, address_lines, postage, service_id, organisation_id, pdf_file
+    ):
+        from app.models import FIRST_CLASS
+
+        recipient_name = address_lines[0]
+        address_without_recipient = address_lines[1:]
+
+        json_payload = {
+            "id": notification_id,
+            "standardParams": {
+                "jobType": "NOTIFY",
+                "templateReference": "NOTIFY",
+                "businessIdentifier": notification_id,
+                "recipientName": recipient_name,
+                "address": {"unstructuredAddress": self._build_unstructured_address(address_without_recipient)},
+            },
+            "customParams": [
+                {"key": "pdfContent", "value": base64.b64encode(pdf_file).decode("utf-8")},
+                {"key": "organisationIdentifier", "value": organisation_id},
+                {"key": "serviceIdentifier", "value": service_id},
+            ],
+        }
+
+        # `despatchMethod` should not be added for second class letters
+        if postage == FIRST_CLASS:
+            json_payload["standardParams"]["despatchMethod"] = "FIRST"
+
+        return json_payload
+
+    @staticmethod
+    def _build_unstructured_address(address_without_recipient):
+        address_line_keys = ["line1", "line2", "line3", "line4", "line5"]
+
+        postcode = address_without_recipient[-1]
+        address_without_postcode = address_without_recipient[:-1]
+
+        unstructured_address = dict(zip(address_line_keys, address_without_postcode))
+        unstructured_address["postcode"] = postcode
+
+        return unstructured_address
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -9,6 +9,32 @@ import jwt
 import requests
 from flask import current_app
 
+from app.clients import ClientException
+
+
+class DvlaException(ClientException):
+    pass
+
+
+class DvlaRetryableException(DvlaException):
+    pass
+
+
+class DvlaNonRetryableException(DvlaException):
+    pass
+
+
+class DvlaDuplicatePrintRequestException(DvlaNonRetryableException):
+    pass
+
+
+class DvlaUnauthorisedRequestException(DvlaRetryableException):
+    pass
+
+
+class DvlaThrottlingException(DvlaRetryableException):
+    pass
+
 
 class SSMParameter:
     # cache properties for up to a day. note that if another process/app changes the value,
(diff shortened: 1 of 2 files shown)


app/clients/letter/dvla.py
```diff
@@ -80,5 +103,99 @@ def authenticate(self):
...
+                {"key": "serviceIdentifier", "value": service_id},
+            ],
+        }
+
+        if postage == FIRST_CLASS:
+            json_payload["standardParams"]["despatchMethod"] = "FIRST"
+
+        return json_payload
+
+    def build_unstructured_address(self, address_lines_without_recipient):
```
samuelhwilliams, line 194: Same with this (internal)

app/clients/letter/dvla.py
```diff
@@ -80,5 +103,99 @@ def authenticate(self):
...
+            elif e.response.status_code == 429:
+                raise DvlaThrottlingException(str(e))
+            elif e.response.status_code >= 500:
+                raise DvlaRetryableExcepton(str(e))
+            else:
+                raise DvlaException(str(e))
+        else:
+            return response.json()
+
+    def format_create_print_job_json(
```
samuelhwilliams, line 166: This can probably be an 'internal' method

app/clients/letter/dvla.py
samuelhwilliams, line 194: Same with this (internal)

app/clients/letter/dvla.py
```diff
@@ -80,5 +103,99 @@ def authenticate(self):
...
+        }
+
+        if postage == FIRST_CLASS:
+            json_payload["standardParams"]["despatchMethod"] = "FIRST"
+
+        return json_payload
+
+    def build_unstructured_address(self, address_lines_without_recipient):
+        address_line_keys = ["line1", "line2", "line3", "line4", "line5"]
+        postcode = address_lines_without_recipient.pop()
```
samuelhwilliams, line 196: This is mutating `address_lines_without_recipient` which was passed in - if anything in the caller tries to use it after calling this function it will (probably unexpectedly) not have the postcode in any more.

Might be nice to make a copy of this dict before modifying it so that the input isn't altered (ie it's a pure function).

We can also mark this as a `staticmethod` as it doesn't need `self`.

tests/app/clients/test_dvla.py
```diff
@@ -156,3 +171,232 @@ def test_jwt_token_calls_authenticate_if_expiry_time_passed(client, dvla_client,
         assert dvla_client._jwt_expires_at == one_hour_later
 
     assert mock_authenticate.called_once
+
+
+def test_format_create_print_job_json_builds_json_body_to_create_print_job(dvla_client):
```
samuelhwilliams, line 176: It might be nice to have these as class-based tests, eg:

```python
class TestFormatCreatePrintJobJson:
    def test_builds_expected_json(self):
        ...
```

Since the next few tests all have the same prefix. It's vaguely preference, but it does provide good and clear encapsulation of tests for a given thing

app/clients/letter/dvla.py
```diff
@@ -80,5 +103,99 @@ def authenticate(self):
...
+
+    def send_letter(
+        self,
+        notification_id: str,
+        address: list[str],
+        postage: str,
+        service_id: str,
+        organisation_id: str,
+        pdf_file: bytes,
+    ):
```
samuelhwilliams, line 121: I wonder if we should prefer to use keyword-only arguments here given we have quite a few things to pass in and calling it without keywords is going to make it hard to verify on-sight that the args in the right order. One way to do this:

```suggestion
    def send_letter(
        self,
        /,
        notification_id: str,
        address: list[str],
        postage: str,
        service_id: str,
        organisation_id: str,
        pdf_file: bytes,
    ):
```

What do you think?

app/clients/letter/dvla.py
```diff
@@ -80,5 +103,99 @@ def authenticate(self):
...
+                raise DvlaException(str(e))
+        else:
+            return response.json()
+
+    def format_create_print_job_json(
+        self, notification_id, address_lines, postage, service_id, organisation_id, pdf_file
+    ):
+        from app.models import FIRST_CLASS
+
+        recipient_name = address_lines.pop(0)
```
samuelhwilliams, line 171: Commented below, but again this is a mutation of incoming data that probably isn't ideal. Anything that calls `send_letter` will now have different data than they might expect.

app/clients/letter/dvla.py
```diff
@@ -6,6 +7,28 @@
...
+
+class DvlaDuplicatePrintRequestException(DvlaException):
+    pass
+
+
+class DvlaUnauthorisedRequestException(DvlaException):
+    pass
+
+
+class DvlaRetryableExcepton(DvlaException):
```
samuelhwilliams, line 25: Should we also have a `DvlaNonRetryableException` (to mirror `EmailClientNonRetryableException`) which we could use to handle 400s from the API? Then we can handle deal with those separately in celery.

Also might be fine to leave this for now and deal with it on the celery task story.

app/clients/letter/dvla.py
```diff
@@ -80,5 +103,99 @@ def authenticate(self):
...
+                    f'{e.response.status_code} {response.json()["errors"][0]["detail"]}'
+                )
+            elif e.response.status_code == 409:
+                raise DvlaDuplicatePrintRequestException(response.json()["errors"][0]["detail"])
+            elif e.response.status_code == 429:
+                raise DvlaThrottlingException(str(e))
+            elif e.response.status_code >= 500:
+                raise DvlaRetryableExcepton(str(e))
+            else:
+                raise DvlaException(str(e))
```
samuelhwilliams, line 162: Rather than `str(e)` it might be nice to use the syntax `raise Dvla[...]Exception() from e` to have it attached as a proper thing using [exception chaining](https://peps.python.org/pep-3134/). I think this should bubble up more clearly in logs, though it may still be fine to include `str(e)` in the message of the new exception.

We can probably add the chaining syntax to all of the errors in this except block

app/clients/letter/dvla.py
```diff
@@ -80,5 +103,99 @@ def authenticate(self):
...
         return response.json()["id-token"]
 
-    def send_letter(self):
-        pass
+    def _get_auth_headers(self):
+        return {
+            "Accept": "application/json",
+            "Authorization": self.jwt_token,
+            "X-API-Key": self.dvla_api_key.get(),
+        }
```
samuelhwilliams, line 111: Do we want to add a test for this?

app/clients/letter/dvla.py
```diff
@@ -80,5 +103,99 @@ def authenticate(self):
...
+            },
+            "customParams": [
+                {"key": "pdfContent", "value": base64.b64encode(pdf_file).decode("utf-8")},
+                {"key": "organisationIdentifier", "value": organisation_id},
+                {"key": "serviceIdentifier", "value": service_id},
+            ],
+        }
+
+        if postage == FIRST_CLASS:
+            json_payload["standardParams"]["despatchMethod"] = "FIRST"
```
samuelhwilliams, line 190: Do we leave this off altogether for 2nd class? I wonder if we should (if we can) be explicit about the despatchMethod in all cases.

samuelhwilliams: Just looked at the DVLA docs and I can see why you've done this as their enum for DespatchMethod doesn't include an entry for second. I'm a bit surprised by that...

klssmith: Yes, I did try passing in "SECOND" to see if worked despite not being documented but it gave an error message of "Invalid despatchMethod value"

samuelhwilliams: Maybe we can add a comment here just highlighting that detail?

app/clients/letter/dvla.py
klssmith: Now internal

app/clients/letter/dvla.py
klssmith: Now internal

app/clients/letter/dvla.py
klssmith: I've updated the places that were using `pop` to get values by using list slicing instead

app/clients/letter/dvla.py
klssmith: Agree this could help avoid confusion, particularly between the service_id and organisation_id, so I've made this change 👍

app/clients/letter/dvla.py
samuelhwilliams: Ah, that's plenty good enough 👍

app/clients/letter/dvla.py
klssmith: I was thinking of the `DvlaException` as the non-retryable one when I wrote this but can see that's not clear. I could maybe rename it to `DvlaNonRetryableException` and then just make the other errors inherit from `ClientException` directly?

Or the other option would be to keep the `DvlaException` (which all the errors inherit from but which is never used directly) and to also add a `DvlaNonRetryableException` exception type

app/clients/letter/dvla.py
klssmith: I've made this change and removed the `str(e)` since I don't think we need that now. I've also stopped manually adding the status code to the `DvlaUnauthorisedRequestException`, since that will now be visible as part of the original exception

app/clients/letter/dvla.py
klssmith: This is now updated, but in the case where we catch an unknown error I've kept the error type raised as `DvlaException` since we don't know if it's retryable or not so can't be more specific

samuelhwilliams (approved):
This looks good - nice work 👍

tests/app/clients/test_dvla.py
klssmith: I prefer non-classed based testing - I'll just stick with the current way for now since it's more consistent with how we write tests in the rest of the code base

samuelhwilliams: 👍

PR #3724, opened 2023-02-17
Add new `deliver_letter` task

klssmith:
The `DVLAClient` has a method to send a letter to the DVLA using their api. This adds a task, `deliver_letter`, to call that method. We need to get information about the letter and the file from S3 to pass as arguments to the client.

Some errors are retried using exponential back-off. If the error type can't be retried or we've reached the maximum number of retries, we put the letter into `technical-failure` status and raise a `NotificationTechnicalFailureException`.

If the API call to the DVLA is successful then we update the letter to `sending`.

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -127,7 +127,14 @@ def deliver_letter(self, notification_id):
         else:
             current_app.logger.exception(f"RETRY: Letter notification {notification_id} failed")
 
-        self.retry()
+        try:
+            self.retry()
+        except self.MaxRetriesExceededError as e:
+            update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
+            raise NotificationTechnicalFailureException(
+                "RETRY FAILED: Max retries reached. The task deliver_letter failed for notification "
+                f"{notification_id}. Notification has been updated to technical-failure"
+            ) from e
     except DvlaNonRetryableException as e:
         if isinstance(e, DvlaDuplicatePrintRequestException):
             current_app.logger.warning(f"Duplicate deliver_letter task called for notification {notification_id}")
@@ -135,12 +142,6 @@ def deliver_letter(self, notification_id):
 
         update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
         raise NotificationTechnicalFailureException(f"Error when sending letter notification {notification_id}") from e
-    except self.MaxRetriesExceededError as e:
-        update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
-        raise NotificationTechnicalFailureException(
-            f"RETRY FAILED: Max retries reached. The task deliver_letter failed for notification {notification_id}. "
-            "Notification has been updated to technical-failure"
-        ) from e
     except Exception as e:
         update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
         raise NotificationTechnicalFailureException(f"Error when sending letter notification {notification_id}") from e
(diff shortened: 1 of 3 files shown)


tests/app/celery/test_provider_tasks.py
```diff
@@ -180,3 +205,282 @@ def test_if_ses_send_rate_throttle_then_should_retry_and_log_warning(sample_noti
...
+        (
+            False,
+            "A. User",
+            {
+                "addressline1": "A. User",
+                "addressline2": "My Street",
+                "addressline3": "London",
+                "addressline4": "SW1 1AA",
+                "addressline5": None,
+                "addressline6": None,
```
samuelhwilliams, line 244: Why the difference in snake casing between these two parameterised examples? PostalAddress seems to want them with underscores.

edit: omg I've just read what InsensitiveDict does 🤯😱 That name is not very accurate. I really don't like how forgiving this is for address info.

app/celery/provider_tasks.py
```diff
@@ -70,3 +89,68 @@ def deliver_email(self, notification_id):
...
+            return
+
+        update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
+        raise NotificationTechnicalFailureException(f"Error when sending letter notification {notification_id}") from e
+    except self.MaxRetriesExceededError as e:
+        update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
+        raise NotificationTechnicalFailureException(
+            f"RETRY FAILED: Max retries reached. The task deliver_letter failed for notification {notification_id}. "
+            "Notification has been updated to technical-failure"
+        ) from e
```
samuelhwilliams, line 143: I believe this is in the wrong place - it needs to wrap the `self.retry` as that is what will throw this exception (see `deliver_email` and `deliver_sms`)

samuelhwilliams (commented):
This looks good 👍  Nice exception handling and good test coverage. I think there's one thing we need to update first.

app/celery/provider_tasks.py
klssmith: Yep, you're right - I've moved it and added the change as a new commit (instead of rebasing) to make it easier to review the difference

tests/app/celery/test_provider_tasks.py
klssmith: I realised when writing this code the address keys are different for CSV vs one-off letters and the `InsensitiveDict` is used to normalise things. Also that the address details we store in the `to` and the personalisation field vary depending on how the letter was sent, so it's all very inconsistent unfortunately

app/celery/provider_tasks.py
```diff
@@ -127,20 +127,21 @@ def deliver_letter(self, notification_id):
...
         raise NotificationTechnicalFailureException(f"Error when sending letter notification {notification_id}") from e
-    except self.MaxRetriesExceededError as e:
-        update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
-        raise NotificationTechnicalFailureException(
-            f"RETRY FAILED: Max retries reached. The task deliver_letter failed for notification {notification_id}. "
-            "Notification has been updated to technical-failure"
-        ) from e
     except Exception as e:
         update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
         raise NotificationTechnicalFailureException(f"Error when sending letter notification {notification_id}") from e
```
samuelhwilliams, line 147: Minor, but I think we could combine these two cases as they seem to be handled the same here?

samuelhwilliams (approved):
👍 

Minor optional refactor but otherwise I think this is good. Nice.

tests/app/celery/test_provider_tasks.py
samuelhwilliams: Oh no :(

app/celery/provider_tasks.py
klssmith: I've updated the code now to combine them

PR #3725, opened 2023-02-20
Support sending international letters via DVLA API

samuelhwilliams:
Updates the DVLA API client to accept and appropriately handle international letters.

Note: this meets the DVLA API spec (https://developer-portal.driver-vehicle-licensing.api.gov.uk/apis/printhub-request-api/printhub-request-api.json), but the API doesn't currently accept this request returning an error. Waiting for their clarification...

---

We'll probably be updating the interface for `dvla_client.send_letter` to take a PostalAddress object soon rather than `address_lines: list[str]`. But for now, we keep it as-is, and so rely on the postage to determine whether an address is international or not. If we revisit that decision (when PostalAddress.international is available) we may need to put some further thought into how to reconcile if notification.postage and PostalAddress.postage are in conflict. For now I'm leaving that.

### DIFF
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -192,11 +192,6 @@ class DVLAClient:
 
         address should be normalised address lines, e.g. ['A. User', 'London', 'SW1 1AA']
         """
-        from app.models import INTERNATIONAL_POSTAGE_TYPES
-
-        if postage in INTERNATIONAL_POSTAGE_TYPES:
-            raise NotImplementedError
-
         current_app.logger.info(f"Sending letter with id {notification_id}")
 
         try:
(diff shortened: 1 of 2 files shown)


app/clients/letter/dvla.py
```diff
@@ -269,3 +270,23 @@ def _build_unstructured_address(address_without_recipient):
...
         return unstructured_address
+
+    @staticmethod
+    def _build_international_address(address_without_recipient):
+        address_line_keys = ["line1", "line2", "line3", "line4", "line5"]
+
+        country = address_without_recipient[-1]
+        address_without_country = address_without_recipient[:-1]
+
+        unstructured_address = dict(zip(address_line_keys, address_without_country))
```
klssmith, line 281: Optional - this variable could be named `international_address` since it provides the value for the `internationalAddress` key

klssmith (approved):

app/clients/letter/dvla.py
samuelhwilliams: /doh 👍

PR #3726, opened 2023-02-22
Use gunicorn to serve flask app

jonathan-ryding:
Updates the Makefile, Dockerfile to use gunicorn instead of developmental flask when running locally. This is done for the `run-flask` and the Docker `run-flask-with-docker` Make commands.
There were some errors with the gunicorn config `gunicorn_config.py`. The logging directory and port settings fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ run-flask: ## Run flask
 
 .PHONY: run-flask-with-docker
 run-flask-with-docker: ## Run flask
-	./scripts/run_locally_with_docker.sh api
+	./scripts/run_locally_with_docker.sh gunicorn -c gunicorn_config.py -b 0.0.0.0:6011 application
 
 .PHONY: run-celery
 run-celery: ## Run celery
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ bootstrap-with-docker: generate-version-file ## Build the image to run the app i
 
 .PHONY: run-flask
 run-flask: ## Run flask
-	. environment.sh && flask run -p 6011
+	. environment.sh && gunicorn -b 127.0.0.1:6011 application
 
 .PHONY: run-flask-with-docker
 run-flask-with-docker: ## Run flask
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -30,4 +30,4 @@ RUN chown -R notify:notify /home/vcap/app \
 
 USER notify
 
-ENTRYPOINT [ "/home/vcap/app/entrypoint.sh" ]
+#ENTRYPOINT [ "/home/vcap/app/entrypoint.sh" ]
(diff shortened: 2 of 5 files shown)


entrypoint.sh
```diff
@@ -19,7 +19,7 @@ then
   shift 1
   # TODO: This is fine for local development but this file will soon need to change to have
```
idavidmcdonald, line 20: You will be able to remove this line 20 comment as it will now have gunicorn infront of it

Makefile
```diff
@@ -36,11 +36,11 @@ bootstrap-with-docker: generate-version-file ## Build the image to run the app i
 
 .PHONY: run-flask
 run-flask: ## Run flask
-	. environment.sh && flask run -p 6011
+	. environment.sh && gunicorn -b 127.0.0.1:6011 application
```
idavidmcdonald, line 39: I think we should keep this command as was before. This command is only used by developers to run the API locally. 99.9% of the time, they only need to run flask, they don't need to run gunicorn infront of it (they would only need to run gunicorn infront if they were editing gunicorn configuration which is done extremely rarely).

Being able to just run flask locally as before will help keep it simple for local development for anyone who doesn't run it in docker.

docker/Dockerfile
```diff
@@ -30,4 +30,4 @@ RUN chown -R notify:notify /home/vcap/app \
 
 USER notify
 
-ENTRYPOINT [ "/home/vcap/app/entrypoint.sh" ]
+#ENTRYPOINT [ "/home/vcap/app/entrypoint.sh" ]
```
idavidmcdonald, line 33: If we are going to use this dockerfile for both local development and also in production, it does make sense to get rid of this line. This will mean we can pass a much wider range of commands into our dockerfile to run.

gunicorn_config.py
```diff
@@ -10,8 +10,8 @@
 workers = 4
 worker_class = "eventlet"
 worker_connections = 256
-errorlog = "/home/vcap/logs/gunicorn_error.log"
-bind = "0.0.0.0:{}".format(os.getenv("PORT"))
+# errorlog = "/home/vcap/app/gunicorn_error.log"
```
idavidmcdonald, line 13: So by removing this setting, it will go back to the default for `errorlog` which appears to be `stderr` (https://docs.gunicorn.org/en/stable/settings.html#errorlog) which is what we want for ECS (because in ECS we will be sending all our logs to stdout and stderr so ECS can pick them up and ship them to cloudwatch.

So essentially, this is a good change for ECS (and I reckon you commented it out because you may ran into problems with trying to write logs to file locally when running too)

sakisv: Might be a good time to check with the rest of the dev team if they ever use the error log when debugging.

Most messages look like this:

Screenshot 2023-02-27 at 14 02 17

And even if we filter out the `INFO` messages and increase the time span to the last 12 weeks we see things like this:

Screenshot 2023-02-27 at 14 05 23

Makefile
leohemsted: definitely want to keep this - `flask` runs in development mode and offers hot code reloading when source files are modified

PR #3728, opened 2023-02-23
Dvla change password/api key scheduled tasks

leohemsted:
Take a look commit by commit cos there's a lot of nuance here that I'm not sure I've got right

### Authenticate

If we get a 401 when creating a jwt token, we should assume the password is out of date. Clear that password from the cache and throw a Retryable exception. Nice and simple.

### Change password

If we get a 401 when changing our password, we should assume that is because the old password is out of date. The problem is there are two cases here and we can't distinguish between each:

* it expired naturally - we'll need to manually reset it by email. This is a bad state. This should never happen ideally since we reset the password every month (and it normally expires every ninety days)
* it has already been reset by a different process from "old" to "new". Because we clear the cache within the lock, we know that we'll have the latest value from SSM. This means we have the "new" password in our request but dvla hasn't made consistent yet (it takes a few seconds for changes to reflect on dvla systems), so dvla gives us an error.

Since we can't distinguish between either of these, I've decided to throw a loud and noisy exception - we'll need to look at the logs and the updated at timestamp of the value in SSM to work out what's happened.

question: should we retry, maybe, once or twice, with a healthy delay (at least fifteen seconds), if we get a password invalid error?

### Change api key

Firstly, we create a JWT token. If the process doesn't have one cached, we need to generate one from the username/password combo. If the password is out of date then this will throw a Retryable as mentioned above - if so, then lets just raise that and the task will retry. The `authenticate` method clears down the password cache already so the next attempt should work.

If the API key has expired, the two potential reasons are exactly the same as the password logic above (either very old and in a bad state, or it's just been reset and we've got the new value.

### Questions

Retries: I copy and pasted the retry definition from deliver_letter but we almost certainly don't want to retry that much.

Clearing the cache _before_ sending a change password/api key request seems sensible, so I included it. but there might be some nuance I'm missing there around if tasks run soon after each other.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -15,7 +15,7 @@ from notifications_utils.timezones import convert_utc_to_bst
 from sqlalchemy import between
 from sqlalchemy.exc import SQLAlchemyError
 
-from app import db, notify_celery, statsd_client, zendesk_client
+from app import db, dvla_client, notify_celery, statsd_client, zendesk_client
 from app.aws import s3
 from app.celery.broadcast_message_tasks import trigger_link_test
 from app.celery.letters_pdf_tasks import get_pdf_for_templated_letter
@@ -533,3 +533,13 @@ def weekly_dwp_report():
         ),
         due_at=convert_utc_to_bst(datetime.utcnow()) + timedelta(days=7),
     )
+
+
+@notify_celery.task(name="change-dvla-password")
+def change_dvla_password():
+    dvla_client.change_password()
+
+
+@notify_celery.task(name="change-dvla-api-key")
+def change_dvla_api_key():
+    dvla_client.change_api_key()
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -12,6 +12,7 @@ from notifications_utils.clients.zendesk.zendesk_client import (
     NotifySupportTicketStatus,
 )
 from notifications_utils.timezones import convert_utc_to_bst
+from redis.exceptions import LockError
 from sqlalchemy import between
 from sqlalchemy.exc import SQLAlchemyError
 
@@ -25,6 +26,7 @@ from app.celery.tasks import (
     process_job,
     process_row,
 )
+from app.clients.letter.dvla import DvlaRetryableException
 from app.config import QueueNames, TaskNames
 from app.cronitor import cronitor
 from app.dao.inbound_numbers_dao import dao_get_available_inbound_numbers
@@ -535,11 +537,25 @@ def weekly_dwp_report():
     )
 
 
-@notify_celery.task(name="change-dvla-password")
-def change_dvla_password():
-    dvla_client.change_password()
+@notify_celery.task(bind=True, name="change-dvla-password", max_retries=3, default_retry_delay=60)
+def change_dvla_password(self):
+    try:
+        dvla_client.change_password()
+    except LockError:
+        # some other task is currently changing the password. let that process handle it and quietly exit
+        current_app.logger.info("change-dvla-password lock held by other process, doing nothing")
+    except DvlaRetryableException:
+        current_app.logger.info("change-dvla-password DvlaRetryableException - retrying")
+        self.retry()
 
 
-@notify_celery.task(name="change-dvla-api-key")
-def change_dvla_api_key():
-    dvla_client.change_api_key()
+@notify_celery.task(bind=True, name="change-dvla-api-key", max_retries=3, default_retry_delay=60)
+def change_dvla_api_key(self):
+    try:
+        dvla_client.change_api_key()
+    except LockError:
+        # some other task is currently changing the api key. let that process handle it and quietly exit
+        current_app.logger.info("change-dvla-api-key lock held by other process, doing nothing")
+    except DvlaRetryableException:
+        current_app.logger.info("change-dvla-api-key DvlaRetryableException - retrying")
+        self.retry()
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -55,6 +55,9 @@ class SSMParameter:
         self.ssm_client.put_parameter(Name=self.key, Value=value, Overwrite=True)
         self._value = value
 
+    def clear(self):
+        self._value = None
+
 
 class DVLAClient:
     """
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -118,10 +118,6 @@ class DVLAClient:
                     "x-api-key": self.dvla_api_key.get(),
                     "Authorization": self.jwt_token,
                 },
-                json={
-                    "userName": self.dvla_username.get(),
-                    "password": self.dvla_password.get(),
-                },
             )
             response.raise_for_status()
 
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -116,6 +116,10 @@ class DVLAClient:
             response.raise_for_status()
         except requests.HTTPError as e:
             if e.response.status_code == 401:
+                # likely the old password has already expired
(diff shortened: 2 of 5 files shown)


app/clients/letter/dvla.py
```diff
@@ -93,33 +93,44 @@ def jwt_token(self):
 
         return self._jwt_token
 
+    def _handle_common_dvla_errors(self, e: requests.HTTPError):
+        if e.response.status_code == 429:
+            raise DvlaThrottlingException() from e
+        elif e.response.status_code >= 500:
+            raise DvlaRetryableException() from e
+        else:
+            raise DvlaException() from e
```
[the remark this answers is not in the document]
leohemsted: yeah, i think its reasonable to be explicit in this generic one. if a 400 is retryable in your case, you should write that explicitly in your exception handler.

fa41fd4

app/config.py
```diff
@@ -360,6 +360,17 @@ class Config(object):
...
+            # first tuesday of every month
+            "change-dvla-api-key": {
+                "task": "change-dvla-api-key",
+                "schedule": crontab(hour=9, minute=0, day_of_week="tue", day_of_month="1-7"),
+                "options": {"queue": QueueNames.REPORTING},
+            },
+            "change-dvla-password": {
+                "task": "change-dvla-password",
+                "schedule": crontab(hour=9, minute=5, day_of_week="tue", day_of_month="1-7"),
+                "options": {"queue": QueueNames.REPORTING},
```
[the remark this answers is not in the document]
leohemsted: 53ac85d

app/config.py
```diff
@@ -366,9 +366,10 @@ class Config(object):
                 "schedule": crontab(hour=9, minute=0, day_of_week="tue", day_of_month="1-7"),
                 "options": {"queue": QueueNames.REPORTING},
             },
+            # first wednesday of every month
             "change-dvla-password": {
                 "task": "change-dvla-password",
-                "schedule": crontab(hour=9, minute=5, day_of_week="tue", day_of_month="1-7"),
+                "schedule": crontab(hour=9, minute=0, day_of_week="wed", day_of_month="1-7"),
```
samuelhwilliams, line 372: This could in theory run on the 1st of the month and then the api key changes on the 7th 🤔 

That is of course fine, but if we wanted to make sure they're consecutive I think we could change `day_of_month="2-8"`?

leohemsted: good point. thats probably less surprising

leohemsted: 🦶 squashed out

app/celery/scheduled_tasks.py
```diff
@@ -533,3 +535,29 @@ def weekly_dwp_report():
         ),
         due_at=convert_utc_to_bst(datetime.utcnow()) + timedelta(days=7),
     )
+
+
+@notify_celery.task(bind=True, name="change-dvla-password", max_retries=55, retry_backoff=True, retry_backoff_max=300)
```
samuelhwilliams, line 540: Yeah I agree this is probably excessive and might just interfere with us manually getting involved. Maybe just try 3 times in total and then fail and bubble up for manual intervention?

app/celery/scheduled_tasks.py
```diff
@@ -533,3 +535,29 @@ def weekly_dwp_report():
...
+def change_dvla_password(self):
+    try:
+        dvla_client.change_password()
+    except LockError:
+        # some other task is currently changing the api key. let that process handle it and quietly exit
+        current_app.logger.info("change-dvla-api-key lock held by other process, doing nothing")
+    except DvlaRetryableException:
+        # could be
```
samuelhwilliams, line 548: the possibilities are endless

samuelhwilliams: (but I'm not sure these comments are super useful)

app/clients/letter/dvla.py
```diff
@@ -87,44 +84,73 @@ def name(self):
...
                 json={
                     "userName": self.dvla_username.get(),
                     "password": self.dvla_password.get(),
                 },
             )
             response.raise_for_status()
+        except requests.HTTPError as e:
+            if e.response.status_code == 401:
+                # likely the old password has already expired
+                current_app.logger.exception("Failed to generate a DVLA jwt token")
```
samuelhwilliams, line 120: Do we want to log an exception here if we are going to retry the task? I don't know if this will trigger an alert/email but I don't think we want that if it's going to automatically recover.

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -1146,3 +1154,44 @@ def test_successful_run(self, mocker, mock_prod_notify_api, notify_db_session, m
         ]
         csv_file = mock_zendesk_update_ticket.call_args_list[0][1]["comment"].attachments[0].filedata
         assert csv_file.read() == 'result,text,comma\r\n1,something else,"quote,text"\r\n'
+
+
+# need to make sure this isnt prefixed with "Test" or pytest will collect and run the base class
+class DvlaScheduledTaskTests:
+    dvla_method_import_path = None
+    scheduled_task_function = None
```
samuelhwilliams, line 1162: Do we have this pattern anywhere else? It's fine although I suspect parameterising the tests is probably easier to read/grok and more consistent? And we don't have to comment about the specific prefix avoidance thing

app/celery/scheduled_tasks.py
leohemsted: ah but they _could be_ useful

(ill squash em out 🦶 🐛 )

app/clients/letter/dvla.py
leohemsted: good point - i think this will trigger an email if it happens in two consecutive minutes, but not otherwise (including if it errors multiple times in one minute) - as per https://github.com/alphagov/notifications-aws/pull/1141

samuelhwilliams: For now - with these email alerting rules.

If (when 🤞) we start using something like Sentry, this exception line and the raised error will be considered different errors/problems and alert separately. We can merge them but I think it seems like a not-very-useful error message to get when the exception below will either 1) retry automatically or 2) be the last retry and error out itself

app/config.py
```diff
@@ -360,6 +360,18 @@ class Config(object):
                 "schedule": crontab(hour=9, minute=0, day_of_week="mon"),
                 "options": {"queue": QueueNames.REPORTING},
             },
+            # first tuesday of every month
+            "change-dvla-api-key": {
+                "task": "change-dvla-api-key",
+                "schedule": crontab(hour=9, minute=0, day_of_week="tue", day_of_month="1-7"),
+                "options": {"queue": QueueNames.REPORTING},
```
samuelhwilliams, line 367: Very minor, but maybe `PERIODIC` is a better queue?

samuelhwilliams (approved):

app/config.py
leohemsted: yes absolutely should be and was my intention, sorry dont know how this got in

app/celery/scheduled_tasks.py
```diff
@@ -533,3 +535,27 @@ def weekly_dwp_report():
...
     )
+
+
+@notify_celery.task(bind=True, name="change-dvla-password", max_retries=3, default_retry_delay=60)
+def change_dvla_password(self):
+    try:
+        dvla_client.change_password()
+    except LockError:
+        # some other task is currently changing the api key. let that process handle it and quietly exit
+        current_app.logger.info("change-dvla-api-key lock held by other process, doing nothing")
```
klssmith, line 546: Just a tiny thing, but it looks like this log line (and the one a few lines below) should start with `change-dvla-password` and not `change-dvla-api-key`

leohemsted: good spot, squashed

app/config.py
leohemsted: fixed

PR #3729, opened 2023-02-23
Allow DVLA response files to have an extra field

klssmith:
The response files we receive from the DVLA have no header row and are in this format:
```
AAAAAAAAAJH3KUL1|Sent|2|Sorted
BBBBBBBBBJ31DT5Q|Sent|1|Sorted
CCCCCCCCCVM6C9GN|Sent|2|Sorted
```

We will soon be receiving the despatch date in the data too, so the data will start taking the following format:
```
AAAAAAAAAJH3KUL1|Sent|2|Sorted|2023-02-23
BBBBBBBBBJ31DT5Q|Sent|1|Sorted|2023-02-23
CCCCCCCCCVM6C9GN|Sent|2|Sorted|2023-02-23
```

This updates our code to not give an error if it receives the extra field. We will start using the extra field later.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,6 +1,8 @@
 import json
-from collections import defaultdict, namedtuple
+from collections import defaultdict
+from dataclasses import dataclass
 from datetime import datetime
+from typing import Optional
 
 from flask import current_app
 from notifications_utils.insensitive_dict import InsensitiveDict
@@ -486,8 +488,21 @@ def persist_daily_sorted_letter_counts(day, file_name, sorted_letter_counts):
     dao_create_or_update_daily_sorted_letter(daily_letter_count)
 
 
+@dataclass
+class NotificationUpdate:
+    """
+    A NotificationUpdate is used to wrap a row of a DVLA response file.
+    `despatch_date` is optional to support files which contain it and those that don't
+    """
+
+    reference: str
+    status: str
+    page_count: str
+    cost_threshold: str
+    despatch_date: Optional[str] = None
+
+
 def process_updates_from_file(response_file):
-    NotificationUpdate = namedtuple("NotificationUpdate", ["reference", "status", "page_count", "cost_threshold"])
     notification_updates = [NotificationUpdate(*line.split("|")) for line in response_file.splitlines()]
     return notification_updates
 
--- a/tests/app/celery/test_ftp_update_tasks.py
+++ b/tests/app/celery/test_ftp_update_tasks.py
@@ -1,4 +1,4 @@
-from collections import defaultdict, namedtuple
+from collections import defaultdict
 from datetime import date, datetime
 
 import pytest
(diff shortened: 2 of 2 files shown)


app/celery/tasks.py
```diff
@@ -486,8 +488,21 @@ def persist_daily_sorted_letter_counts(day, file_name, sorted_letter_counts):
...
+    """
+    A NotificationUpdate is used to wrap a row of a DVLA response file.
+    `despatch_date` is optional to support files which contain it and those that don't
+    """
+
+    reference: str
+    status: str
+    page_count: str
+    cost_threshold: str
+    despatch_date: Optional[str] = ""
```
samuelhwilliams, line 502: I think this should probably be

```suggestion
    despatch_date: Optional[str] = None
```

Otherwise you're setting the default value to an empty string, which means that unless someone explicitly provides `None` it will always be a string, ie the `Optional` part is not necessarily accurate.

I appreciate it's probably a bit difficult to grok some of this at the moment as we don't have any actual static type-checking being done to provide feedback.

klssmith: Thanks, I've made the suggested change

app/celery/tasks.py
```diff
@@ -486,8 +488,21 @@ def persist_daily_sorted_letter_counts(day, file_name, sorted_letter_counts):
...
+    """
+    A NotificationUpdate is used to wrap a row of a DVLA response file.
+    `despatch_date` is optional to support files which contain it and those that don't
+    """
+
+    reference: str
+    status: str
+    page_count: str
+    cost_threshold: str
+    despatch_date: Optional[str] = None
```
samuelhwilliams, line 502: I wonder if it might be nice to take the opportunity to have `page_count` and `despatch_date` as native objects (int and date) now? Possibly that means a few more changes to how they get used elsewhere - up to you whether it's worth it.

klssmith: I think `despatch_date` I'd prefer to leave for now - there will be an upcoming story to start using that field which might be a good time to change how it works, but I'd like to just make sure we are receiving the field in the expected format with no errors before trying to parse it or change its type.

Were you thinking that `page_count` could get changed to in an `int` in the dataclass, or before it's passed to it?

samuelhwilliams: So this is a bit of a quirk of dataclasses. If you define page_count as an int with `page_count: int`, it won't actually do any mutation of data as the instance is being made. If you pass in `"2"` to the field, you will end up with a string still. So you need to explicitly convert it to an integer before/as you pass it into the constructor.

So the line `notification_updates = [NotificationUpdate(*line.split("|")) for line in response_file.splitlines()]` would need to pull out the `page_count` and convert it to an int explicitly. It would be nice/more correct to use the right data types for each thing, but it may not be worth it right now as so if you want to keep it as-is, I'm OK with it.

klssmith: Thanks, I knew it wouldn't happen automatically in the dataclass but wasn't sure if you were thinking of [a solution like this](https://stackoverflow.com/a/54863771). I think I'd probably keep things as-is for now mainly because it keeps the code a little simpler and once we're fully migrated to sending letters via the API we may stop using response files.

tests/app/celery/test_ftp_update_tasks.py
```diff
@@ -39,9 +39,11 @@
 @pytest.fixture
 def notification_update():
     """
-    Returns a namedtuple to use as the argument for the check_billable_units function
+    Returns an instance of the  NotificationUpdate dataclass to use as the argument
+    for the check_billable_units function
     """
-    NotificationUpdate = namedtuple("NotificationUpdate", ["reference", "status", "page_count", "cost_threshold"])
+    from app.celery.tasks import NotificationUpdate
```
samuelhwilliams, line 45: these inline imports are probably a mild hint that this class isn't defined in a great place

I wonder if we should move it to a more central file like `dataclasses.py` (sibling to our `models.py`) so that these can be imported/used from anywhere without worrying about circular imports.

samuelhwilliams (approved):
I think this looks good 👍  Above comment optional

PR #3732, opened 2023-02-28
Run gunicorn in front of flask in Docker entrypoint

jonathan-ryding:
The command: make run-gunicorn-with-docker now uses the gunicorn server in front of the flask application in Docker.
Gunicorn error logs are no longer written to file, these are now output as stderr.
The server port variable was defined by the EXPOSED_PORTS parameter.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -17,9 +17,7 @@ then
 elif [ "$1" == "api" ]
 then
   shift 1
-  # TODO: This is fine for local development but this file will soon need to change to have
-  # gunicorn infront of flask
-  flask run --host 0.0.0.0 --port 6011
+  gunicorn -c /home/vcap/app/gunicorn_config.py application
 elif [ -n "$*" ]
 then
   $*
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -18,6 +18,10 @@ elif [ "$1" == "api" ]
 then
   shift 1
   gunicorn -c /home/vcap/app/gunicorn_config.py application
+elif [ "$1" == "api-local" ]
+then
+  shift 1
+  flask run --host 0.0.0.0 --port $PORT
 elif [ -n "$*" ]
 then
   $*
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -2,25 +2,20 @@
 
 if [ "$1" == "worker" ]
 then
-  shift 1
   # TODO: This is fine for local development but this file will soon need to change to have
   # many different celery workers like we currently do in production
   celery -A run_celery.notify_celery worker --pidfile="/tmp/celery.pid" --loglevel=INFO --concurrency=4
 elif [ "$1" == "beat" ]
 then
-  shift 1
   celery -A run_celery.notify_celery beat --loglevel=INFO
 elif [ "$1" == "migration" ]
 then
-  shift 1
   flask db upgrade
 elif [ "$1" == "api" ]
 then
-  shift 1
   gunicorn -c /home/vcap/app/gunicorn_config.py application
 elif [ "$1" == "api-local" ]
 then
-  shift 1
   flask run --host 0.0.0.0 --port $PORT
 elif [ -n "$*" ]
 then
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -10,7 +10,6 @@ from gds_metrics.gunicorn import child_exit  # noqa
 workers = 4
 worker_class = "eventlet"
 worker_connections = 256
-errorlog = "/home/vcap/logs/gunicorn_error.log"
 bind = "0.0.0.0:{}".format(os.getenv("PORT"))
 statsd_host = "{}:8125".format(os.getenv("STATSD_HOST"))
 gunicorn.SERVER_SOFTWARE = "None"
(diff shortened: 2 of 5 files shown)


gunicorn_config.py
```diff
@@ -10,7 +10,6 @@
 workers = 4
 worker_class = "eventlet"
 worker_connections = 256
-errorlog = "/home/vcap/logs/gunicorn_error.log"
```
idavidmcdonald, line 13: I think it is worth calling out that this will also change things for the PaaS deployed version and stop putting logs into the file (and then those will no longer be shipped to (probably Cloudwatch)).

Sakis pointed out that we pretty much never get any useful logs from gunicorn though so maybe this is tolerable to turn it off for the PaaS.

Alternatively, we could maybe try and override this by either

- keeping this line in the file and change the entrypoint.sh file to explicitely set `errorlog -` so that it goes to stdout
- removing this line in the file and change the command that is used to run the PaaS so that it specifies the file should be used for errorlog.

I'm not really sure what I think. @sakisv do you have an opinion?

entrypoint.sh
```diff
@@ -17,9 +17,7 @@ then
 elif [ "$1" == "api" ]
 then
   shift 1
-  # TODO: This is fine for local development but this file will soon need to change to have
-  # gunicorn infront of flask
-  flask run --host 0.0.0.0 --port 6011
+  gunicorn -c /home/vcap/app/gunicorn_config.py application
```
idavidmcdonald, line 20: So this is good but as you pointed out in PR message, this will change `make run-flask-with-docker` to actually run gunicorn infront of flask.

I think we should firstly change the name of the command so that it is `make run-gunicorn-with-docker` so it says what it does. 

And then secondly, I'm wondering if we should then add back in a command to `make run-flask-with-docker` so we can still have a way for just running flask...

scripts/run_locally_with_docker.sh
```diff
@@ -15,7 +15,7 @@ API_HOST_NAME=${API_HOST_NAME:-"http://host.docker.internal:6011"}
 # Only expose port 6011 if we're running the API - anything else is celery which shouldn't bind the port.
 # This lets us run celery via docker and the API locally.
 if [ "${@}" == "api" ]; then
-  EXPOSED_PORTS="-p 6011:6011"
+  EXPOSED_PORTS="-e PORT=6011 -p 6011:6011"
```
idavidmcdonald, line 18: I don't think this needs any change but I know we were discussing the other time how the PaaS version of this is configured.

I did a search in the api repo and the aws repo for `PORT` and 6011 but couldn't find anything relevant.

I couldn't figure out why. However, I've ssh'd onto one of the preview instances of `notify-api` and `notify-delivery-worker-sender` and then run `echo $PORT` which returns `8080`.

My guess is that this environment variable is set by the PaaS team automatically for us on the PaaS (and is value 8080 whereas on local we specify 6011, for reasons I don't remember).

So my thinking is that
- we require an environment variable for PORT when running gunicorn because our config file requires it (https://github.com/alphagov/notifications-api/blob/main/gunicorn_config.py#L14)
- runnning gunicorn on the PaaS we don't need to set a PORT because the PaaS set the environment variable for us and that is why we can't find any reference to it in our code
- running flask locally we don't actually need to set a port because flask has a default port that it exposes flask on (probably 8080), nor does it use the PORT environment variable anywhere in our python app. However we choose when running flask locally to override the default port flask is exposed on by calling `flask run -p 6011' in our makefile
- if we were to run gunicorn locally or via docker, either way we need to set the PORT environment variable for it to work

idavidmcdonald (commented):
Good work Jonathan. I think what this does is good. We might want to have a bit of discussion though (probably in person, maybe with Sakis too) so we can decide what we do about

- gunicorn error logs on the PaaS
- whether we offer a `make run-flask-with-docker` option and how we might go about doing that

gunicorn_config.py
sakisv: Personally I don't see any value in having these logs, and I would remove them (and also remove the relevant clouwatch-agent config bit from the run-app-paas script).

But, before doing that I'd like to confirm with the dev team that these logs are indeed useless and they haven't looked at them.

scripts/run_locally_with_docker.sh
sakisv: cloudfoundry does indeed expect the apps to run on 8080: https://docs.cloudfoundry.org/buildpacks/python/index.html#run-web-server.

When running things locally we need to have the apps listening on different ports so that they don't overlap.

The question I guess is, do we want to have the same ports used when we deploy things to AWS or we want to use a more homogenous approach where all the http apps listen on 8080?

I think I would prefer the later as it makes it simpler to reason about what's listening on which port and to have one less thing to keep in your mind, but then again developers may already know these ports by heart.

entrypoint.sh
sakisv: I think we should to avoid breaking devs workflows. As Leo [mentioned](https://github.com/alphagov/notifications-api/pull/3726#discussion_r1118666841) flask also runs with debug mode enabled

idavidmcdonald (approved):
Nice!

PR #3734, opened 2023-02-28
add dvla creds to preview

leohemsted:


app/config.py
```diff
@@ -553,6 +553,8 @@ class Preview(Config):
     API_RATE_LIMIT_ENABLED = True
     CHECK_PROXY_HEADER = False
 
+    DVLA_API_BASE_URL = "https://uat.driver-vehicle-licensing.api.gov.uk"
```
samuelhwilliams, line 556: This seems weird (but consistent I guess).

It feels like our config isn't actually configurable (by env vars), it's just code all in one place. This should be a default value on line 432 and then env vars in each deployed env can just override it.

But I guess we don't actually have a good answer for config because of the credentials repo thing. Maybe we can fix this after migrating off PaaS.

samuelhwilliams (approved):

app/config.py
leohemsted: no that's a fair point - i think i'll change this to be the default value on all environments, with prod specificially overriding the default

PR #3735, opened 2023-03-01
Update DVLA API to handle BFPO addresses

samuelhwilliams:
Blocked until https://github.com/alphagov/notifications-utils/pull/1022 is merged

---

At the moment we will just send BFPO addresses across as normal unstructuredAddress blocks, but the DVLA API has specific handling available for BFPO addresses. This uses the update to notifications_utils PostalAddress to detect and handle BFPO addresses appropriately. This will help our print provider sort and process those letters more efficiently.


CrystalPea (approved):
Looks good

PR #3736, opened 2023-03-01
Move constants to a central file

samuelhwilliams:
We should be able to access static strings from anywhere in the code base without having to worry about circular imports or side effects.


leohemsted (commented):
i'm fine with us moving to a separate constants file, however, there's one of two things i'd like to see:

light touch - go through the constants and ensure they're grouped semantically, clearly separated, with divider comments or similar

heavy touch - perhaps put [some of] the constants in classes
eg 

```python
class NotificationStatus:
  CREATED = "created"
  SENDING = "sending"
  ...
```

samuelhwilliams:
> heavy touch - perhaps put [some of] the constants in classes

I think this is a good idea and I'd like to do it - but separately.

leohemsted (approved):

PR #3737, opened 2023-03-02
Send letters via DVLA api on development, preview and staging

leohemsted:
bypass the ftp app on dev/preview/staging.

This doesn't include any selective enabling (for example per service). The entire env is either FTP or API for now.

also, perhaps slightly more contentious - add an optional print deadline to the collate letter task. This is necessary for running performance tests because when you call `collate-and-send` without any args, it runs for everything ready to send since the _last 5:30 we went past_. So if you run it at midday, it'll only look at letters sent up to 5:30pm yesterday. This makes it tough to manually trigger runs locally/on preview, since you often need to change the created_at in the database, and then move the file in s3, which is a pain.

By allowing us to pass in an optional arg to define when we want the print deadline to be, we can easily trigger this after loading files for load tests etc.

I've had some real headaches understanding which datetimes are UTC and which are local time, and obviously this is super critical to get right when we tell users about 5:30pm local being a cut-off, so for now rather than refactor everything and risk breaking something delicate, I thought I'd just be super explicit with the new variable name

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -2,6 +2,7 @@ from base64 import urlsafe_b64encode
 from datetime import datetime, timedelta
 from hashlib import sha512
 
+import dateutil
 from botocore.exceptions import ClientError as BotoClientError
 from flask import current_app
 from notifications_utils.letter_timings import LETTER_PROCESSING_DEADLINE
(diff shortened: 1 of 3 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -128,21 +130,30 @@ def update_validation_failed_for_templated_letter(self, notification_id, page_co
...
+    else:
+        print_run_date = convert_utc_to_bst(datetime.utcnow())
+        if print_run_date.time() < LETTER_PROCESSING_DEADLINE:
+            print_run_date = print_run_date - timedelta(days=1)
 
-    print_run_deadline = print_run_date.replace(hour=17, minute=30, second=0, microsecond=0)
+        print_run_deadline = print_run_date.replace(hour=17, minute=30, second=0, microsecond=0)
     _get_letters_and_sheets_volumes_and_send_to_dvla(print_run_deadline)
 
+    if current_app.config["NOTIFY_ENVIRONMENT"] in {"development", "preview", "staging"}:
```
samuelhwilliams, line 154: Can we do this with a env var/config instead?

app/celery/letters_pdf_tasks.py
```diff
@@ -128,21 +130,30 @@ def update_validation_failed_for_templated_letter(self, notification_id, page_co
 
 @notify_celery.task(name="collate-letter-pdfs-to-be-sent")
 @cronitor("collate-letter-pdfs-to-be-sent")
-def collate_letter_pdfs_to_be_sent():
+def collate_letter_pdfs_to_be_sent(print_run_deadline_utc=None):
```
samuelhwilliams, line 133: Good call with the `_utc` suffix - I would rather we start being this explicit more often 👍

app/celery/letters_pdf_tasks.py
```diff
@@ -283,6 +294,16 @@ def group_letters(letter_pdfs):
...
 
 
+def send_dvla_letters_via_api(print_run_deadline):
+    for postage in POSTAGE_TYPES:
+        current_app.logger.info(f"send-dvla-letters-for-day-via-api - starting processing for postage class {postage}")
+        for letter in dao_get_letters_to_be_printed(print_run_deadline, postage):
+            # TODO: send from new queue
+            deliver_letter.apply_async(kwargs={"notification_id": letter.id}, queue=QueueNames.SEND_LETTER)
+
+        current_app.logger.info(f"send-dvla-letters-for-day-via-api - finished processing for postage class {postage}")
```
samuelhwilliams, line 304: Tiny nit: maybe "finished queuing" instead of "finished processing"?

samuelhwilliams (approved):

tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -321,6 +322,31 @@ def test_get_key_and_size_of_letters_to_be_sent_to_print_handles_file_not_found(
     ]
 
 
+@mock_s3
+@pytest.mark.parametrize(
+    "print_run_deadline_utc, expected_time_called_with",
+    [
+        ("2021-06-01 12:00:00", datetime(2021, 6, 1, 13, 0, 0)),
```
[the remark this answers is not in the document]
leohemsted: added

PR #3738, opened 2023-03-03
Make volumes by service report return one row per service

klssmith:
This report was returning multiple rows for each service where the only difference between them was the free allowance data. We only want it to return one row per service and we only care about the free allowance at the time of the specified end date for the report.

The annual_billing subquery in the `fetch_volumes_by_service` was returning a row for each row in the annual billing table like this:

```
financial_year_start, service_id, free_allowance
(2022, service_1_id, 40000),
(2021, service_1_id, 150000),
(2022, service_2_id, 40000),
(2022, service_3_id, 10000),
(2020, service_3_id, 25000),
(2021, service_3_id, 150000),
```

We now use a window function in the subquery so that it also returns the latest (maximum) financial_year_start for each service ID:

```
latest_billing_year_for_service, financial_year_start, service_id, free_allowance
(2022, 2022, service_1_id, 40000),
(2022, 2021, service_1_id, 150000),
(2022, 2022, service_2_id, 40000),
(2022, 2022, service_3_id, 10000),
(2022, 2020, service_3_id, 25000),
(2022, 2021, service_3_id, 150000),
```

Then in the main query, we add a filter to only return rows where the `annual_billing.latest_billing_year_for_service` equals the `annual_billing.financial_year_start`, which acts to remove all rows for annual billing older than the end date for the report.

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -972,12 +972,14 @@ def fetch_volumes_by_service(start_date, end_date):
 
     annual_billing = (
         db.session.query(
-            func.max(AnnualBilling.financial_year_start).label("financial_year_start"),
+            func.max(AnnualBilling.financial_year_start)
+            .over(partition_by=AnnualBilling.service_id)
+            .label("latest_billing_year_for_service"),
+            AnnualBilling.financial_year_start.label("financial_year_start"),
             AnnualBilling.service_id,
             AnnualBilling.free_sms_fragment_limit,
         )
         .filter(AnnualBilling.financial_year_start <= year_end_date)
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Are we able to write a test for this as well?

klssmith:
The lines of billing data that I've added to the `set_up_usage_data` test function means that the `test_fetch_volumes_by_service` test now fails with the original code, but I can also add a separate and more comprehensive test.

samuelhwilliams (approved):

klssmith:
I've added another test so that we can be sure that the function is not only returning one row per service for the free allowance but that the free allowance returned is the free allowance for the end date of the query (which may not necessarily be the latest).

PR #3740, opened 2023-03-09
Add letter attachment table

samuelhwilliams:
Build out the DB model for letter attachments with basic fields. This table represents another PDF that has been uploaded to Notify and can be 'attached' to the end of a letter template. This will allow reusable formatted content (eg a form) to be attached to a letter.

---

~The orginal ticket suggests an `archived` field. I've left that off here as I don't think we need to denormalise this information - if there are no records with `template.letter_attachment_id == this_letter_attachment.id` then `this_letter_attachment` isn't in use. It saves us from a potentially-weird situation where we have `archived=true` on a letter attachment but it's still referred to by a live template.~ Added back archived_at/archived_by_id after some discussion

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -956,6 +956,26 @@ class TemplateBase(db.Model):
     def service_letter_contact(cls):
         return db.relationship("ServiceLetterContact", viewonly=True)
 
+    @declared_attr
+    def letter_attachment_id(cls):
+        return db.Column(UUID(as_uuid=True), db.ForeignKey("letter_attachment.id"), nullable=True)
+
+    @declared_attr
+    def letter_attachment(cls):
+        return db.relationship("LetterAttachment")
+
+    @declared_attr
+    def __table_args__(cls):
+        if cls.__name__ not in {"Template", "TemplateHistory"}:
+            raise RuntimeError("Make sure to manually add this CheckConstraint to the new migration")
+
+        return (
+            CheckConstraint(
+                "template_type = 'letter' OR letter_attachment_id IS NULL",
+                name=f"ck_{cls.__tablename__}_letter_attachments",
+            ),
+        )
+
     @property
     def reply_to(self):
         if self.template_type == LETTER_TYPE:
@@ -2364,3 +2384,17 @@ class WebauthnCredential(db.Model):
             "updated_at": get_dt_string_or_none(self.updated_at),
             "logged_in_at": get_dt_string_or_none(self.logged_in_at),
         }
+
+
+class LetterAttachment(db.Model):
+    __tablename__ = "letter_attachment"
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True, nullable=False, default=uuid.uuid4)
+
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    created_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey("users.id"), nullable=False)
+    archived_at = db.Column(db.DateTime, nullable=True)
+    archived_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey("users.id"), nullable=True)
+
+    original_filename = db.Column(db.String, nullable=False)
+    page_count = db.Column(db.SmallInteger, nullable=False)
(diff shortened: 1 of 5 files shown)


app/models.py
```diff
@@ -2364,3 +2384,18 @@ def serialize(self):
             "updated_at": get_dt_string_or_none(self.updated_at),
             "logged_in_at": get_dt_string_or_none(self.logged_in_at),
         }
+
+
+class LetterAttachment(db.Model):
```
CrystalPea, line 2389: is there a reason we have a singular form here? We have a plural for most our tables like that, for example `notifications`, `services`, `templates`.

app/models.py
```diff
@@ -2364,3 +2384,18 @@ def serialize(self):
...
+    __tablename__ = "letter_attachment"
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True, nullable=False, default=uuid.uuid4)
+
+    created_at = db.Column(db.DateTime, nullable=True, default=datetime.datetime.utcnow)
```
CrystalPea, line 2394: Maybe nullable should be False here? Since each attachment will have it?

app/models.py
```diff
@@ -2364,3 +2384,18 @@ def serialize(self):
...
+    created_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey("users.id"), nullable=True)
+    updated_at = db.Column(db.DateTime, nullable=True, onupdate=datetime.datetime.utcnow)
+    updated_by_id = db.Column(UUID(as_uuid=True), db.ForeignKey("users.id"), nullable=True)
+
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey("services.id"), primary_key=True)
```
CrystalPea, line 2399: Would be good to know why we went with service_id vs template_id here.

CrystalPea (commented):
Nice one, great work on constraints and relationships between tables.

Also good point about `archived` field, in current setup we don't need it, and if we begin to need it in the future, migrating that should be easy enough.

Still, I have a few questions, suggestions:

app/models.py
samuelhwilliams: Oh my, good spot

app/models.py
samuelhwilliams: We are actually, sadly, _really_ inconsistent.

For example we also have `organisation`, `domain`, `annual_billing`, `template_folder`.

And to make the inconsistency worse we even sometimes have the model be plural/singular and the table name be opposite, see ApiKey->api_keys. Sad times.

It's obviously an opinionated choice and fairly contentious (https://stackoverflow.com/questions/338156/table-naming-dilemma-singular-vs-plural-names). I guess I settle on the singular side of things here.

app/models.py
samuelhwilliams: So, `service_id` is definitely something we could drop.

I've excluded `template_id` because we've already got that relationship defined from Template.attachment_id so having the reverse isn't super useful.

The main reason I've added this is because it vastly simplifies us finding out which service the attachment was uploaded by. If the attachment is removed from a template, we make a new template version and then the attachment won't be related to any records in the `template` table - only `template_history`. It's a bit more mentally painful to then have to go attachment->template_history->service to answer that simple (and fairly reasonable I think) question of which service 'owns' the file.

CrystalPea: I don't think it would be that difficult to add if we need it later. It would be a matter of 1 SQL in the migration. And we definitely don't need it now. I think it would be useful if another dev, like @leohemsted , could weigh in.

I think the uses we need to think about for currently planned functions are:
- viewing the attachment on template page
- what happens when users deletes or replaces the attachment
- how easy it will be to make a rule to delete old attachments from S3. If we have an "archived" column, then when it is `archived`, `updated_at` is updated. We could also just have `archived_at` instead. Then it's easy to check which attachments have been archived a month or more ago and delete them from S3.
Without that the check becomes more awkward.
- how easy it is to re-assemble the letter in case of an issue with template preview, which happened in the past. So if attachment for the letter changes, how easy can we find the right attachment for when notification was sent and stitch it all together again.

app/models.py
CrystalPea: I see, I think plural makes more sense to me, but I guess it's a matter of preference.

app/models.py
samuelhwilliams: I have renamed to archived_at and archived_by_id :)

CrystalPea (approved):

PR #3742, opened 2023-03-13
Change test DB setup

samuelhwilliams:
Use a subprocess to run the flask db/alembic migration. Alembic does a lot of its own configuration, including logging setup. If we do this in-process then we override a lot of our own logging configuration, which breaks things like the `caplog` fixture from pytest.

Our alembic config actually disables our app logger, which means we aren't handling any log messages, which as we also see here, means we've got some log messages that are syntatically invalid but aren't picking them up (because log messages are lazily evaluated).


CrystalPea (approved):
Nice work 👏🏼

PR #3743, opened 2023-03-14
Inject user_id into log messages

samuelhwilliams:
If we're handling API requests from the admin app, it might be sending us a header that indicates the responsible user ID. Let's inject this into any log messages emitted so that it's easier to trace all of the actions a specific user has taken/is responsible for.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,8 +7,10 @@ import pytest
 import sqlalchemy
 
 from app import create_app, db
+from app.authentication.auth import requires_admin_auth, requires_no_auth
 from app.dao.provider_details_dao import get_provider_details_by_identifier
 from app.notify_api_flask_app import NotifyApiFlaskApp
+from tests.routes import admin_test_blueprint, test_blueprint
 
 # Freezegun has a negative interaction with prompt_toolkit that ends up suppressing text written on the prompt of ipdb
 # so let's ignore that module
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -22,6 +24,12 @@ def notify_api():
     app = NotifyApiFlaskApp("test")
     create_app(app)
 
+    # Attach some routes that can be used as helpers for specific tests.
+    test_blueprint.before_request(requires_no_auth)
+    app.register_blueprint(test_blueprint, url_prefix="/test")
+    admin_test_blueprint.before_request(requires_admin_auth)
+    app.register_blueprint(admin_test_blueprint, url_prefix="/admin-test")
```
leohemsted, line 31: i'm a little uncertain about these blueprints - i think i'd rather they were added just for a specific test rather than everything (also the names should probably more closely reflect the `before_request` auth functions being the interesting thing?)

but not sure if there's a nice way to do that in a separate fixture that we can just define in test_logging_filters or similar? 🤔

leohemsted (commented):
Also a merge conflict

samuelhwilliams:
not gonna bother sorting the reqs stuff until the utils PR is in

tests/conftest.py
leohemsted: yeah, for testing error handlers i created a new app entirely and then attached error handlers and had a route for each test, but that might not be quite as applicable/readable/useful for these tests 🤔  

https://github.com/alphagov/notifications-api/blob/ee50423c8c7386e0f11d3e682acc7d04cac3d092/tests/app/v2/test_errors.py#L15-L28

samuelhwilliams: Ah hmm yeah. I would rather test the real application here I think 🤔

leohemsted (approved):

PR #3744, opened 2023-03-14
Allow SQLALCHEMY_DATABASE_URI to override the database settings inside PaaS

corlettb:
What
----

Currently, when establishing a database connection, the VCAP_SERVICES environment variable is given priority over SQLALCHEMY_DATABASE_URI.

This proposed change would prioritise the SQLALCHEMY_DATABASE_URI variable over VCAP_SERVICES, provided that it is present. In the absence of SQLALCHEMY_DATABASE_URI, VCAP_SERVICES will be used as a fallback.

Why
----

We need to switch to fixed credentials for the database before we do the database cutover, while still using VCAP_SERVICES for Redis. This pull request will facilitate that change.

### DIFF
--- a/app/cloudfoundry_config.py
+++ b/app/cloudfoundry_config.py
@@ -6,8 +6,9 @@ def extract_cloudfoundry_config():
     vcap_services = json.loads(os.environ["VCAP_SERVICES"])
 
     # Postgres config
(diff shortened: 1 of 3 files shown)


idavidmcdonald (approved):
Looks great Ben. Good work at getting into the Notify codebase so quickly.

Feel free to make it into a proper PR, rather than in draft and I can merge it and deploy it out for you.

PR #3746, opened 2023-03-15
Restrict DVLA API TLS connections to specific ciphers 

samuelhwilliams:
The DVLA API we will connect to in prod supports a number of insecure
ciphers. It will take some time for them to retire support due to other
clients requiring them, but we want to make sure we aren't able to
connect with those insecure ciphers.

We will supply our own list of acceptable cipher suites via config.


CrystalPea (approved):
Looks great, the tests are really helpful, too 🙏🏼

PR #3748, opened 2023-03-17
Minor dvla API tweaks

leohemsted:
# set letter to technical-failure if we cant find it in s3

otherwise we'll just do nothing until the next day, when we'll try and send it again. i think the chance of it appearing in s3 overnight feels slim to unlikely - this was behaviour we added three years ago[^1] but the log message never fired due to a bug in how the function was called. it feels safer and more correct to set the letter to technical-failure. we'll get a proper exception log and it'll be easier to distinguish failed letters in technical-failure that we need to action from new letters in created that will go out as normal

[^1]: https://github.com/alphagov/notifications-api/pull/2760

# remove duplicate log line

we already log in provider_tasks.deliver_letter

### DIFF
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -240,7 +240,6 @@ class DVLAClient:
 
         address should be normalised address lines, e.g. ['A. User', 'London', 'SW1 1AA']
         """
-        current_app.logger.info(f"Sending letter with id {notification_id}")
 
         try:
             response = self.request.post(


app/clients/letter/dvla.py
```diff
@@ -240,7 +240,6 @@ def send_letter(
 
         address should be normalised address lines, e.g. ['A. User', 'London', 'SW1 1AA']
         """
-        current_app.logger.info(f"Sending letter with id {notification_id}")
```
samuelhwilliams, line 243: I'm not really against still having this line that says "we got to dvla_client.send_letter" 🤔  It's still a useful checkpoint I think.

samuelhwilliams (approved):

app/clients/letter/dvla.py
leohemsted: i did pause (because this was how i realised that we had a bug anyway) but i think as far as i can see, by changing the call to raise rather than return, we should never silently exit before that point any more so i think it's less useful now

PR #3751, opened 2023-03-22
Migrate to using caplog instead of mocking loggers

samuelhwilliams:
And fix a few more parts of the codebase where we just have broken log statements :sad-tear:

### DIFF
--- a/app/celery/process_ses_receipts_tasks.py
+++ b/app/celery/process_ses_receipts_tasks.py
@@ -54,7 +54,8 @@ def process_ses_results(self, response):
 
         if bounce_message:
             current_app.logger.info(
-                f"SES bounce for notification ID {notification.id}", extra=dict(bounce_message=bounce_message)
+                f"SES bounce for notification ID {notification.id}",
+                extra=dict(bounce_message=json.dumps(bounce_message)),
             )
 
         if notification.status not in [NOTIFICATION_SENDING, NOTIFICATION_PENDING]:
--- a/app/celery/process_ses_receipts_tasks.py
+++ b/app/celery/process_ses_receipts_tasks.py
@@ -53,7 +53,9 @@ def process_ses_results(self, response):
             return
 
         if bounce_message:
-            current_app.logger.info(f"SES bounce for notification ID {notification.id}: {bounce_message}")
+            current_app.logger.info(
+                f"SES bounce for notification ID {notification.id}", extra=dict(bounce_message=bounce_message)
+            )
 
         if notification.status not in [NOTIFICATION_SENDING, NOTIFICATION_PENDING]:
             notifications_dao._duplicate_update_warning(notification=notification, status=notification_status)
--- a/tests/app/celery/test_ftp_update_tasks.py
+++ b/tests/app/celery/test_ftp_update_tasks.py
@@ -229,28 +229,27 @@ def test_update_letter_notifications_to_error_updates_based_on_notification_refe
 
 
 def test_check_billable_units_when_billable_units_matches_page_count(
-    client, sample_letter_template, mocker, notification_update
+    client, sample_letter_template, caplog, notification_update
 ):
-    mock_logger = mocker.patch("app.celery.tasks.current_app.logger.error")
+    with caplog.at_level("ERROR"):
+        create_notification(sample_letter_template, reference="REFERENCE_ABC", billable_units=1)
 
-    create_notification(sample_letter_template, reference="REFERENCE_ABC", billable_units=1)
+        check_billable_units(notification_update)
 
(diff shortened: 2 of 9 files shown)


app/celery/process_ses_receipts_tasks.py
```diff
@@ -53,7 +53,9 @@ def process_ses_results(self, response):
             return
 
         if bounce_message:
-            current_app.logger.info(f"SES bounce for notification ID {notification.id}: {bounce_message}")
+            current_app.logger.info(
+                f"SES bounce for notification ID {notification.id}", extra=dict(bounce_message=bounce_message)
```
samuelhwilliams, line 57: `bounce_message` here is a dict that looks a bit like this (in the test): `{"bounce": {"bounceSubType": "General", "bounceType": "Permanent", "feedbackId": "0102015fc9e676fb-12341234-1234-1234-1234-9301e86a4fa8-000000", "remoteMtaIp": "123.123.123.123", "reportingMTA": "dsn; a7-31.smtp-out.eu-west-1.amazonses.com", "timestamp": "2017-11-17T12:14:05.131Z"}, "mail": {"headersTruncated": false, "messageId": "ref", "sendingAccountId": "12341234", "source": "\"TEST\" <TEST@notify.works>", "sourceArn": "arn:aws:ses:eu-west-1:12341234:identity/notify.works", "sourceIp": "0.0.0.1", "timestamp": "2017-11-17T12:14:03.000Z"}, "notificationType": "Bounce"}`.

This throws a "can't format" error from our JSON log formatter when we just inline the dict directly. I don't know if we actually need this info but by adding it to `extra` rather than inline into the message we:

1) Keep the message cleaner
2) Still have access to the raw bounce JSON as an attribute on the emitted log record.

samuelhwilliams: `extra` doesn't get emitted as a top-level but everything inside it does, yes. So we should get an attribute called `bounce_message` in kibana.

We get it in the JSON log output and I don't think anything would drop it along the way. It should auto-index into kibana.

samuelhwilliams: Example logs:

```json
{"name": "test", "levelname": "INFO", "message": "Logging configured", "pathname": "/Users/sam/work/gds/notifications-api/venv/lib/python3.9/site-packages/notifications_utils/logging.py", "lineno": 42, "time": "2023-03-22T11:46:00", "requestId": "no-request-id", "application": "test", "service_id": "no-service-id", "logType": "application"}
{"name": "app", "levelname": "INFO", "message": "Logging configured", "pathname": "/Users/sam/work/gds/notifications-api/venv/lib/python3.9/site-packages/notifications_utils/logging.py", "lineno": 42, "time": "2023-03-22T11:46:01", "requestId": "no-request-id", "application": "app", "service_id": "no-service-id", "logType": "application"}
{"name": "test", "levelname": "INFO", "message": "SES bounce for notification ID efb3e640-0545-4f92-b987-af2faef74f74: {'bounce': {'bounceSubType': 'General', 'bounceType': 'Permanent', 'feedbackId': '0102015fc9e676fb-12341234-1234-1234-1234-9301e86a4fa8-000000', 'remoteMtaIp': '123.123.123.123', 'reportingMTA': 'dsn; a7-31.smtp-out.eu-west-1.amazonses.com', 'timestamp': '2017-11-17T12:14:05.131Z'}, 'mail': {'headersTruncated': False, 'messageId': 'ref', 'sendingAccountId': '12341234', 'source': '\"TEST\" <TEST@notify.works>', 'sourceArn': 'arn:aws:ses:eu-west-1:12341234:identity/notify.works', 'sourceIp': '0.0.0.1', 'timestamp': '2017-11-17T12:14:03.000Z'}, 'notificationType': 'Bounce'}", "pathname": "/Users/sam/work/gds/notifications-api/app/celery/process_ses_receipts_tasks.py", "lineno": 56, "time": "2023-03-22T11:46:02", "requestId": null, "application": "test", "service_id": "no-service-id", "logType": "application"}
```

leohemsted: cool lets give it a try as is then - agree it'll be nice to just search for logs where `bounce_message` is populated, letting us filter easily on all eg permanents or transients if we need to yadda yadda

though i do wonder if that should be nested one furhter level to separate any arbitrary extra log bits from the generic stuff like levelname, lineno, time, etc? its probs okay as is but keen to get your thoughts. but something like:

```json
{
  "name": "test",
  "levelname": "INFO",
  "message": "something"
  "context": {
    "bounce_message": {...}
  },
  "time": "2023-...",
   ...
}
```

leohemsted (approved):

PR #3752, opened 2023-03-23
Run celery beat in UTC

samuelhwilliams:
We currently schedule celery beat tasks in Europe/London, which means they are scheduled in GMT/BST as appropriate. This means that the hour/minute set in the crontab is the wall-clock time that they will run in reality. This is nice for us as it's predictable and easy to reason about (we don't need to really consider timezones).

Unfortunately, that means celery-beat needs to consider timezones, and it hates doing that: `https://github.com/celery/celery/issues/6438`. This issue has been open on celery for 2.5 years as of writing, and talks about how the celery-beat scheduler can crap the bed when a timezone change occurs, leading to scheduled tasks being skipped. Restarting the celery beat scheduler once the new time zone happens is a workaround, but we don't have a good way to do that.

After some discussion we've agreed that we're happy to run tasks in UTC to get things running consistently.

We should be careful not to schedule anything between 11pm and midnight UTC, as this is 11pm-midnight GMT but midnight-1am BST - ie could run on different local days. For our reporting or processing tasks, this could be really awkward or broken, eg some notifications sent between 11pm and midnight could be counted in the next day's report. Let's add a test that makes sure nothing is scheduled to _only_ run between 11pm and midnight UTC.

### DIFF
--- a/tests/app/test_config.py
+++ b/tests/app/test_config.py
@@ -3,6 +3,7 @@ import os
 from unittest import mock
 
 import pytest
+from celery.schedules import crontab
 
 from app import config, db
 from app.config import QueueNames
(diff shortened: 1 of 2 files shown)


tests/app/test_config.py
```diff
@@ -86,6 +87,24 @@ def test_queue_names_all_queues_correct():
...
 
+def test_no_celery_beat_tasks_scheduled_over_midnight_between_timezones(notify_api):
+    badly_scheduled_tasks = []
+
+    for task_name, task_info in notify_api.config["CELERY"]["beat_schedule"].items():
+        schedule = task_info["schedule"]
+        if not isinstance(schedule, crontab):
+            continue
+
+        if not (schedule.hour - {23}):
```
leohemsted, line 98: took me a bit of local mucking around to realise that if a crontab is for something more frequent, it'll include every hour of the day in this. guess this makes sense, its hard to imagine a situation where we'd write, eg, crontab where it was {22, 23} or {11, 23} or something

leohemsted (approved):

tests/app/test_config.py
samuelhwilliams: Yeah sorry I could include a comment there, let me add one.

PR #3758, opened 2023-04-03
Re-configure Sentry; remove NewRelic

samuelhwilliams:
* Remove NewRelic; we won't be going ahead with it (😢)
* Re-configure Sentry to send errors and be explicitly disabled in production - except for a single instance, which we can use to monitor vitals as/when we trial sentry there.

---

Will eventually move the commonalities to notifications-utils, but for now just duplicating it.


app/performance.py
```diff
@@ -1,43 +1,34 @@
...
     environment = os.getenv("NOTIFY_ENVIRONMENT").lower()
     enable_apm = environment in {"development", "preview", "staging"}
 
-    if enable_apm and os.getenv("NEW_RELIC_ENABLED") == "1":
-        import newrelic.agent
-
-        # Expects NEW_RELIC_LICENSE_KEY set in environment as well.
-        newrelic.agent.initialize("newrelic.ini", environment=environment, ignore_errors=False)
+    if environment == "production" and os.getenv("CF_INSTANCE_INDEX", "-1") == "9":
+        # In production, turn it on for instance 9 (randomly chosen) only.
```
leohemsted, line 17: image

leohemsted (approved):

PR #3759, opened 2023-04-05
Replace flake8 and isort with ruff

samuelhwilliams:


requirements_for_test.txt
```diff
@@ -1,9 +1,8 @@
 -r requirements.txt
 black==22.8.0  # Also update `.pre-commit-config.yaml` if this changes
-flake8==5.0.4  # Also update `.pre-commit-config.yaml` if this changes
-flake8-bugbear==22.9.23
+ruff==0.0.256  # Also update `.pre-commit-config.yaml` if this changes
```
leohemsted, line 3: any reason you added 0.0.256 vs a more recent version? (i notice that they’re up to 261)

Makefile
```diff
@@ -82,9 +82,7 @@ generate-version-file: ## Generates the app version file
 
 .PHONY: test
 test: ## Run tests
-	flake8 .
-	isort --check-only ./app ./tests
-	black --check .
```
leohemsted, line 87: did we intend to remove `black` here?

leohemsted (approved):

requirements_for_test.txt
samuelhwilliams: 🥲 .256 is so much nicer

But no, I think 256 was the latest version when I made the branch a few weeks ago and hadn't bumped it. But will now.

Makefile
samuelhwilliams: No :) (I initially thought ruff could replace black but it doesn't - [yet](https://github.com/charliermarsh/ruff/issues/1904))

PR #3765, opened 2023-04-11
Fix a test locally for GMT/BST

samuelhwilliams:
time.time() returns a unix epoch timestamp (seconds since unix epoch).

datetime.fromtimestamp will turn a unix epoch timestamp into a naive local datetime. freezegun then assumes that's a utc/gmt datetime because we don't provide a tzoffset to the call.

Let's explicitly make a UTC datetime from the unix epoch so that freezegun freezes the right time.


whpearson (approved):
Tests are fixed locally

PR #3767, opened 2023-04-13
Update error content to remove contraction

samuelhwilliams:
This fixes an issue where the 'fancy' apostrophe also gets broken and displays as 'doesnâ€™t'


samuelhwilliams:
We are going to leave this as-is for now.

It is not fundamentally broken in the downloaded CSV report. It may display incorrectly for some users if the file is opened in the wrong encoding (eg windows 1252 rather than utf-8).

We will be looking at reporting in a more wholesale fashion in the near future and may address this then.

PR #3769, opened 2023-04-14
Add table for organisation permissions

whpearson:
This creates a new table with rows with

- organisation_permmission: A type that represents each permission
- enabled: Whether this organisation permission is enabled or not, I think this might be useful if the default behaviour might change. So even though it might default to off, it might be nice to make it explicitly off
- organisation_id: the organisation this permission applies to
- created_at, updated_at: standard timestamps

### DIFF
--- a/app/dao/organisation_permissions_dao.py
+++ b/app/dao/organisation_permissions_dao.py
@@ -1,0 +1,13 @@
+from app import db
+from app.dao.dao_utils import autocommit
+from app.models import OrganisationPermission
+
+
+@autocommit
+def set_organisation_permission(organisation, permissions):
+    query = OrganisationPermission.query.filter_by(organisation=organisation)
+    query.delete()
+    for p in permissions:
+        o_p = OrganisationPermission(organisation_id=organisation.id, permission=p)
+        o_p.organisation = organisation
+        db.session.add(o_p)
--- a/app/models.py
+++ b/app/models.py
@@ -59,6 +59,7 @@ from app.constants import (
     NOTIFICATION_STATUS_TYPES_COMPLETED,
     NOTIFICATION_STATUS_TYPES_FAILED,
     NOTIFICATION_TYPE,
+    ORGANISATION_PERMISSION_TYPES,
     PERMISSION_LIST,
     PRECOMPILED_TEMPLATE_NAME,
     SMS_AUTH_TYPE,
@@ -215,7 +216,6 @@ user_to_organisation = db.Table(
     UniqueConstraint("user_id", "organisation_id", name="uix_user_to_organisation"),
 )
 
-
 user_folder_permissions = db.Table(
     "user_folder_permissions",
     db.Model.metadata,
@@ -352,6 +352,21 @@ class OrganisationTypes(db.Model):
     annual_free_sms_fragment_limit = db.Column(db.BigInteger, nullable=False)
 
 
+class OrganisationPermission(db.Model):
+    __tablename__ = "organisation_permissions"
+    id = db.Column(UUID(as_uuid=True), primary_key=True, nullable=False, default=uuid.uuid4)
+
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    organisation = db.relationship("Organisation", backref="permissions")
+    organisation_id = db.Column(UUID(as_uuid=True), db.ForeignKey("organisation.id"), nullable=False)
+    permission = db.Column(
+        db.Enum(*ORGANISATION_PERMISSION_TYPES, name="organisation_permission_types"),
+        index=False,
+        unique=False,
+        nullable=False,
+    )
+
+
 class Organisation(db.Model):
     __tablename__ = "organisation"
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4, unique=False)
@@ -424,6 +439,11 @@ class Organisation(db.Model):
     def domain_list(self):
         return [domain.domain for domain in self.domains]
 
+    def set_permissions_list(self, permissions: list[str]):
+        from app.dao.organisation_permissions_dao import set_organisation_permission
+
+        set_organisation_permission(self, permissions)
+
     def serialize(self):
         return {
             "id": str(self.id),
@@ -448,6 +468,7 @@ class Organisation(db.Model):
             "billing_contact_email_addresses": self.billing_contact_email_addresses,
             "billing_reference": self.billing_reference,
             "can_approve_own_go_live_requests": self.can_approve_own_go_live_requests,
+            "permissions": [x.permission for x in self.permissions],
         }
 
     def serialize_for_list(self):
--- a/app/organisation/rest.py
+++ b/app/organisation/rest.py
@@ -78,7 +78,6 @@ def get_organisation_by_id(organisation_id):
 
 @organisation_blueprint.route("/by-domain", methods=["GET"])
 def get_organisation_by_domain():
-
     domain = request.args.get("domain")
 
     if not domain or "@" in domain:
@@ -117,7 +116,11 @@ def update_organisation(organisation_id):
         if not organisation.letter_branding_id:
             data["letter_branding_id"] = current_app.config["NHS_LETTER_BRANDING_ID"]
 
-    result = dao_update_organisation(organisation_id, **data)
+    if data.get("permissions") or data.get("permissions") == []:
+        organisation.set_permissions_list(data.get("permissions"))
+        result = True
+    else:
+        result = dao_update_organisation(organisation_id, **data)
 
     if data.get("agreement_signed") is True:
         # if a platform admin has manually adjusted the organisation, don't tell people
--- a/tests/app/organisation/test_rest.py
+++ b/tests/app/organisation/test_rest.py
@@ -92,6 +92,7 @@ def test_get_organisation_by_id(admin_request, notify_db_session):
         "billing_reference",
         "purchase_order_number",
         "can_approve_own_go_live_requests",
+        "permissions",
     }
     assert response["id"] == str(org.id)
     assert response["name"] == "test_org_1"
@@ -117,7 +118,6 @@ def test_get_organisation_by_id(admin_request, notify_db_session):
 
 
 def test_get_organisation_by_id_returns_domains(admin_request, notify_db_session):
-
     org = create_organisation(
         domains=[
             "foo.gov.uk",
@@ -275,7 +275,8 @@ def test_post_create_organisation_existing_name_raises_400(admin_request, sample
             },
             (
                 "organisation_type foo is not one of "
-                "[central, local, nhs_central, nhs_local, nhs_gp, emergency_service, school_or_college, other]"
+                "[central, local, nhs_central, nhs_local, nhs_gp, emergency_service, school_or_college, "
+                "other]"
             ),
         ),
     ),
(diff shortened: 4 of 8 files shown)


samuelhwilliams (commented):
Migration generally looks good 👍 

I think we might want to tweak the structure a bit so it's more in line with some of our other permission tables (although we're not super consistent by the look of it...)

I think the fields we need are: `organisation_id`, `created_at`, and `permission`. To remove the permission we currently just delete the record rather than disable it (for better or worse).

Permissions are a bit inconsistent too - an enum seems a good choice, but just so you know, in some other places (Eg service permissions), we have a separate table called `service_permission_types` and reference that through an FK.

I'm guessing you wrote this by hand rather than autogenerating it with alembic - do you know how to do that / want to know?

We will need an accompanying Python model class for the table as well - these are stored in `app/models.py` and generally go in at the same time as the migration file.

I can see this PR is in draft so maybe you know all of this/are working on it already, but let me know if you want to chat anything over :)

app/models.py
```diff
@@ -352,6 +352,21 @@ class OrganisationTypes(db.Model):
...
 
+class OrganisationPermission(db.Model):
+    __tablename__ = "organisation_permissions"
+    id = db.Column(UUID(as_uuid=True), primary_key=True, nullable=False, default=uuid.uuid4)
+
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    organisation = db.relationship("Organisation")
+    organisation_id = db.Column(UUID(as_uuid=True), db.ForeignKey("organisation.id"), nullable=False)
+    organisation_permission = db.Column(
+        db.Enum(*ORGANISATION_PERMISSION_LIST, name="organisation_permission_types"),
```
samuelhwilliams, line 363: Could we just call this `permission`? I think that matches what we have on other models, and otherwise we have `OrganisationPermission` as the table name as well as `organisation_permission` as a column name - feels a bit verbose/redundant.

whpearson: Sure.

app/dao/organisation_permissions_dao.py
```diff
@@ -0,0 +1,42 @@
+from app import db
+from app.dao import DAOClass
+from app.models import OrganisationPermission
+
+# Default permissions for a service
```
samuelhwilliams, line 5: This comment can probably be updated :)

app/dao/organisation_permissions_dao.py
```diff
@@ -0,0 +1,42 @@
...
+default_organisation_permissions = []
+
+
+class OrganisationPermissionDAO(DAOClass):
```
samuelhwilliams, line 9: Man, I didn't even know we had these. Well done for digging through and finding/reusing.

You may find that we've got some inconsistencies around some of the DAO stuff (I'm not totally sure why we even are using DAOs when we use sqlalchemy 🤔)

tests/app/organisation/test_rest.py
```diff
@@ -275,7 +275,8 @@ def test_post_create_organisation_existing_name_raises_400(admin_request, sample
             },
             (
                 "organisation_type foo is not one of "
-                "[central, local, nhs_central, nhs_local, nhs_gp, emergency_service, school_or_college, other]"
+                "[central, local, nhs_central, nhs_local, nhs_gp, emergency_service, school_or_college, "
+                "other]"
```
samuelhwilliams, line 279: Was this one of your auto-formatters kicking in or a manual choice? We leave most of our formatting decisions to `black` and linting to `ruff`.

If you haven't already: we provide a set of [pre-commit](https://pre-commit.com/) hooks that you can use in pretty much every relevant repo. Download/install it and then in each repo run `pre-commit install --install-hooks`. Then on every commit we'll run things like autoformatting, fixing line endings, and throwing linting errors etc.

app/models.py
```diff
@@ -352,6 +352,21 @@ class OrganisationTypes(db.Model):
     annual_free_sms_fragment_limit = db.Column(db.BigInteger, nullable=False)
 
 
+class OrganisationPermission(db.Model):
+    __tablename__ = "organisation_permissions"
+    id = db.Column(UUID(as_uuid=True), primary_key=True, nullable=False, default=uuid.uuid4)
+
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    organisation = db.relationship("Organisation")
```
samuelhwilliams, line 360: It would be nice to have a backref on this so that we can access `Organisation.permissions` directly (we do similar things for Service/ServicePermission)

```suggestion
    organisation = db.relationship("Organisation", backref="permissions")
```

app/dao/organisation_permissions_dao.py
```diff
@@ -0,0 +1,42 @@
...
+        model = OrganisationPermission
+
+    def add_default_service_permissions_for_organisation(self, organisation):
+        for name in default_organisation_permissions:
+            organisation_permission = OrganisationPermission(permission=name, organisation=organisation)
+            self.create_instance(organisation_permission, _commit=False)
+
+    def remove_user_service_permissions(self, organisation):
+        query = self.Meta.model.query.filter_by(organisation=organisation)
+        query.delete()
```
samuelhwilliams, line 20: I think we can remove references to `service` here - ie just call them `add_default_permissions_to_organisation` and `remove_organisation_permissions`

app/models.py
```diff
@@ -352,6 +352,21 @@ class OrganisationTypes(db.Model):
...
+    organisation_id = db.Column(UUID(as_uuid=True), db.ForeignKey("organisation.id"), nullable=False)
+    permission = db.Column(
+        db.Enum(*ORGANISATION_PERMISSION_LIST, name="organisation_permission_types"),
```
samuelhwilliams, line 363: Just for your awareness, there's a bit of history on Notify around the use of enums. We have some enums but in other places we use so called "type tables", see NotificationStatusType or OrganisationType.

This is because of some historical migration issues that were encountered when updating enums. I'm not sure of the exact historical reasons, but there are probably more type tables than enums.

There's a bit of low-level discussion around what we'd like to rationalise on going forwards. For now I'm happy for us to keep this as an enum but there may be other opinions on the team.

tests/app/organisation/test_rest.py
whpearson: I think it was one of the autoformatters kicking off. I'll make sure precommit hooks are working.

app/dao/organisation_permissions_dao.py
samuelhwilliams: I always leave behind stragglers like this... not to worry.

app/dao/organisation_permissions_dao.py
whpearson: Class is scrapped. There were a couple of other functions needed.

app/organisation/rest.py
```diff
@@ -353,3 +356,24 @@ def _send_notification(template_id, recipient, personalisation):
     _send_notification(
         current_app.config[signer_template_id], organisation.agreement_signed_by.email_address, personalisation
     )
+
+
+@organisation_blueprint.route("/<uuid:organisation_id>/organisation_permission/<string:permission>", methods=["POST"])
```
samuelhwilliams, line 361: This route could probably just be `/permission/` to remove some redundancy - what do you think?

app/organisation/rest.py
```diff
@@ -353,3 +356,24 @@ def _send_notification(template_id, recipient, personalisation):
...
+def add_organisation_permission(organisation_id, permission):
+    organisation = dao_get_organisation_by_id(organisation_id)
+    permissions = organisation.get_permissions_list()
+    # Doing it this way to prevent duplication of permissions
+    if permission not in permissions:
+        permissions.extend(permission)
+    permissions = [permission]
```
samuelhwilliams, line 368: I think we need to drop this line?

app/dao/organisation_permissions_dao.py
```diff
@@ -0,0 +1,34 @@
+from app import db
+from app.models import OrganisationPermission
+
+# Default permissions for an organisation
+default_organisation_permissions = []
+
+
+def delete_organisation_permission(permission):
```
samuelhwilliams, line 8: I think you should be able to wrap this function with the `@autocommit` decorator than than manually doing `db.session.commit()` - at least this is what we do in a lot of the other dao files?

Maybe not for the `set_organisation_permission` one below though if we need that `_commit` arg.

app/dao/organisation_permissions_dao.py
```diff
@@ -0,0 +1,34 @@
...
+        if _commit:
+            db.session.rollback()
+        raise e
+    else:
+        if _commit:
+            db.session.commit()
+
+
+def get_organisation_permissions_by_organisation_id(organisation_id):
+    return OrganisationPermission.query.filter_by(organisation_id=organisation_id).all()
```
samuelhwilliams, line 34: I don't think we need this - we should just be able to access `organisation.permissions` to get the same thing?

app/models.py
```diff
@@ -424,6 +439,18 @@ def live_services(self):
     def domain_list(self):
         return [domain.domain for domain in self.domains]
 
+    def get_permissions_list(self):
```
samuelhwilliams, line 442: If we have the `permissions` backref, we shouldn't need this - we can access `organisation.permissions` directly.

```
sam:~/work/gds/notifications-api [WP-perm-ask-to-join-service]$ flask shell
2023-04-24T09:05:22 app app INFO no-request-id "Logging configured" [in /Users/sam/work/gds/notifications-api/venv/lib/python3.9/site-packages/notifications_utils/logging.py:43]
Python 3.9.15 (main, Dec 22 2022, 19:49:52)
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
App: app
Instance: /Users/sam/work/gds/notifications-api/instance
>>> from app.models import *
>>> Organisation.query.first().permissions
2023-04-24T09:05:24 app app WARNING no-request-id "Checked out sqlalchemy connection from outside of request/task" [in /Users/sam/work/gds/notifications-api/app/__init__.py:409]
[]
```

app/models.py
```diff
@@ -424,6 +439,18 @@ def live_services(self):
...
+        from app.dao.organisation_permissions_dao import get_organisation_permissions_by_organisation_id
+
+        return [x.permission for x in get_organisation_permissions_by_organisation_id(self.id)]
+
+    def set_permissions_list(self, permissions):
+        from app.dao.organisation_permissions_dao import set_organisation_permission
+
+        permission_list = [OrganisationPermission(organisation_id=self.id, permission=p) for p in permissions]
+
+        set_organisation_permission(self, permission_list)
```
samuelhwilliams, line 452: This one I'm sort of OK with, but I think we could probably just bundle this all into `set_organisation_permission(organisation: Organisation, permissions: list[str])` and remove the extra helper layer

whpearson: Should this single function be in the dao?

samuelhwilliams: Yeah I think that's a good call

app/models.py
```diff
@@ -448,6 +468,7 @@ def serialize(self):
             "billing_contact_email_addresses": self.billing_contact_email_addresses,
             "billing_reference": self.billing_reference,
             "can_approve_own_go_live_requests": self.can_approve_own_go_live_requests,
+            "organisation_permissions": [x.permission for x in self.permissions],
```
samuelhwilliams, line 471: Can we just call this `permissions`? Having the `organisation_` prefix is still redundant

samuelhwilliams (approved):
I think after those final comments this looks good. Thanks for persevering and good work 🙌

PR #3772, opened 2023-04-17
Hourly ft_billing updates

samuelhwilliams:
Schedule a new task that updates `ft_billing` with the latest data for today every hour.

Includes a cronitor to make sure this is running hourly.

Ticket: https://trello.com/c/V6vkvxon/254-review-loading-time-for-organisation-page

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -28,23 +28,23 @@ def create_nightly_billing(day_start=None):
     for i in range(0, 10):
         process_day = (day_start - timedelta(days=i)).isoformat()
 
-        create_nightly_billing_for_day.apply_async(kwargs={"process_day": process_day}, queue=QueueNames.REPORTING)
+        create_or_update_ft_billing_for_day.apply_async(kwargs={"process_day": process_day}, queue=QueueNames.REPORTING)
         current_app.logger.info(
-            f"create-nightly-billing task: create-nightly-billing-for-day task created for {process_day}"
+            f"create-nightly-billing task: create-or-update-ft-billing-for-day task created for {process_day}"
         )
 
 
-@notify_celery.task(name="create-nightly-billing-for-day")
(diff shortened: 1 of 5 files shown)


app/celery/reporting_tasks.py
```diff
@@ -28,23 +28,30 @@ def create_nightly_billing(day_start=None):
...
+@notify_celery.task(name="update-ft-billing-for-today")
+@cronitor("update-ft-billing-for-today")
+def update_ft_billing_for_today():
+    process_day = convert_utc_to_bst(datetime.utcnow()).date().isoformat()
+    create_or_update_ft_billing_for_day.apply_async(kwargs={"process_day": process_day}, queue=QueueNames.REPORTING)
+
+
+@notify_celery.task(name="create-or-update-ft-billing-for-day")
+def create_or_update_ft_billing_for_day(process_day):
     process_day = datetime.strptime(process_day, "%Y-%m-%d").date()
```
whpearson, line 46: This just redefines the process_day (and so doesn't use the variable passed in). It might be worth combining with update_ft_billing_for_today, as that doesn't do too much, if I have not missed a reason for splitting them apart.

samuelhwilliams: It _does_ use the variable passed in, although I appreciate it's a bit contrived/confusing.

`process_day` passed in is a string; line 46 converts it to a `datetime.date`.

`update_ft_billing_for_today` is a very thin function/task but I think provides a clear context for what is being done.

I suppose we could update this to `create_or_update_ft_billing_for_day(process_day=None)` and then if `process_day` isn't provided, default it to today, but it's a bit less explicit and just calling `create_or_update_ft_billing_for_day()` without any args looks a little bit unusual.

I'm not overly attached, but I suppose that's the logic/reasoning I had.

whpearson (approved):
Changes make sense. I missed the process_day being used to generate itself. The type hints and variable name makes things clearer.

PR #3775, opened 2023-04-19
Stop on-demand calculation of ft_billing data

samuelhwilliams:
To merge after https://github.com/alphagov/notifications-api/pull/3772

Admin PR: https://github.com/alphagov/notifications-admin/pull/4685

---

Remove the on-demand calculations of ft_billing when the API endpoint is
hit, as this can be a set of heavy queries causing slow page loads. We
now have an hourly celery task that calculates this information for the
current day, so we can use the cached information instead.

To indicate when the data was last updated, we'll return the timestamp
for when the cache was last updated.

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -1,11 +1,12 @@
 from datetime import datetime, timedelta
 
+import pytz
 from flask import current_app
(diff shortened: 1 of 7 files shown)


whpearson (approved):
Changes look good. Ran tests locally and made sure the billing part of admin still functions.

PR #3779, opened 2023-04-21
change url to have dashes, not underscores

leohemsted:
whoops, missed this first time round. stick with existing convention.


whpearson (approved):
Looks good

PR #3784, opened 2023-04-27
Added queues to entrypoint.sh

robbrazil:
modified:   entrypoint.sh
Modified entrypoint.sh to cater for all queues


sakisv (commented):
Almost there, two small adjustments and it'd be good to 👍

PR #3785, opened 2023-04-27
Populate AnnualBilling automatically each year

samuelhwilliams:
At the start of each financial year, let's automatically populate AnnualBilling for all services that exist.

This will ensure that the org usage dashboard immediately reflects free allowances for all of its services.

Newly-created services get an AnnualBilling entry added upon creation.

Creds PR: https://github.com/alphagov/notifications-credentials/pull/338

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -591,7 +591,6 @@ def populate_annual_billing(year, missing_services_only):
         active_services = Service.query.filter(Service.active).all()
 
     for service in active_services:
-        print(f"update service {service.id} with default")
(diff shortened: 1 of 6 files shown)


whpearson:
Looks good. What was the reasoning behind leaving the tests to be of the populate_annual_billing_with_defaults and the test cli runner? Do we expect that to still be used at some point in the future?

samuelhwilliams:
Good point - we probably could get rid of some of those tests. I don't hate keeping the command around in case the task fails and we want to re-run it. But we could just emit the task again rather than run the command. I'll drop it out 👍

whpearson (approved):
Looks good 👍

PR #3791, opened 2023-05-03
Adding archive endpoint for letter-attachment

whpearson:
This endpoint is the one called when removing an attachment from a
template.

As there are multiple versions of the template, which might be reverted
to, the attachment is kept. So this archive endpoint just does a soft
delete.

It does this by removing it from template and updating the
archived_at and archived_by_id in the letter-attachment
db row.

### DIFF
--- a/app/letter_attachment/rest.py
+++ b/app/letter_attachment/rest.py
@@ -1,8 +1,10 @@
+import datetime
+
 from flask import Blueprint, jsonify, request
 
 from app.dao.templates_dao import dao_get_template_by_id, dao_update_template
 from app.errors import InvalidRequest, register_errors
-from app.letter_attachment.schema import post_create_letter_attachment_schema
+from app.letter_attachment.schema import post_archive_letter_attachment_schema, post_create_letter_attachment_schema
 from app.models import LetterAttachment
 from app.schema_validation import validate
 
@@ -41,3 +43,20 @@ def create_letter_attachment():
     dao_update_template(template)
 
     return jsonify(letter_attachment.serialize()), 201
+
+
+@letter_attachment_blueprint.route("/letter-attachment/<uuid:letter_attachment_id>/archive", methods=["POST"])
+def archive_letter_attachment(letter_attachment_id):
+    data = request.get_json()
+
+    validate(data, post_archive_letter_attachment_schema)
+
+    letter_attachment = LetterAttachment.query.get_or_404(letter_attachment_id)
+    if not (template := letter_attachment.template):
+        raise InvalidRequest("letter-attachment-already-archived", 400)
+    template.letter_attachment = None
+    letter_attachment.archived_at = datetime.datetime.utcnow()
+    letter_attachment.archived_by_id = data["archived_by"]
+    dao_update_template(template)
+
+    return "", 204
--- a/tests/app/letter_attachment/test_rest.py
+++ b/tests/app/letter_attachment/test_rest.py
@@ -103,3 +103,120 @@ def test_create_letter_attachment_returns_400_if_template_already_has_attachment
     }
 
     admin_request.post("letter_attachment.create_letter_attachment", _data=data, _expected_status=400)
+
+
+def test_archive_letter_attachment_creates_new_template_version(
+    admin_request, notify_db_session, sample_letter_template
+):
+    assert sample_letter_template.version == 1
+
+    upload_id = uuid.uuid4()
+
+    response = admin_request.post(
+        "letter_attachment.create_letter_attachment",
+        _data={
+            "upload_id": str(upload_id),
+            "created_by_id": str(sample_letter_template.created_by_id),
+            "original_filename": "filename.pdf",
+            "page_count": 2,
+            "template_id": str(sample_letter_template.id),
+        },
+        _expected_status=201,
+    )
+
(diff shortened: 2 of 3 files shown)


app/letter_attachment/rest.py
```diff
@@ -41,3 +43,21 @@ def create_letter_attachment():
...
+def archive_letter_attachment(letter_attachment_id):
+    data = request.get_json()
+
+    validate(data, post_archive_letter_attachment_schema)
+
+    letter_attachment = LetterAttachment.query.get_or_404(letter_attachment_id)
+
+    if (template := dao_get_template_by_id(data["template_id"])) is None:
+        raise InvalidRequest("template-not-found", 404)
+    template.letter_attachment = None
```
samuelhwilliams, line 58: I think we should rewrite this check so that we don't need to post the template ID across.

We can just use `letter_attachment.temple` to find the related template. If it exists then it's not been archived; if it doesn't then it's already been archived.

This protects us from another little potential edge case here - someone could send any old template_id across and we'll wipe the letter attachment from it. In this implementation we aren't enforcing that the letter_attachment being archived is attached to the template we're removing the attachment from.

Using the relationship takes care of this check for us.

I'd also be interested in your thoughts about 400 vs 404 here. I'm tempted to say that trying to archive an attachment that has already been archived should be a 400 rather than a 404, since we aren't _actually_ deleting it the resource does exist - but it's not valid to try to archive it again.

It depends how much we want to pretend that the underlying resource is being deleted rather than just hidden.

tests/app/letter_attachment/test_rest.py
```diff
@@ -103,3 +103,122 @@ def test_create_letter_attachment_returns_400_if_template_already_has_attachment
     }
 
     admin_request.post("letter_attachment.create_letter_attachment", _data=data, _expected_status=400)
+
+
+def test_archive_letter_attachment_creates_new_template_version(
```
samuelhwilliams, line 108: nice test 🙌

app/letter_attachment/rest.py
whpearson: I've updated to follow your suggestion. I seemed to have to look it up by the DAO, as It was causing this error.
`Can't record history for Template (something in your code has casued the database to flush the session early so there's nothing to copy into the history table)`
But it is using the id from the letter attachment relationship to do the lookup. I'll dig into the versioning thing a bit.

The 404 vs 400. I don't think we'll see many of these in practice. Unless someone is doing something hinkey to our API I'd expect it to only show up if someone was managing to press the button twice. I think 400 makes a little more sense, since you are making a bad request given the context. I've switched it over.

app/letter_attachment/rest.py
```diff
@@ -41,3 +43,24 @@ def create_letter_attachment():
...
+
+    validate(data, post_archive_letter_attachment_schema)
+
+    letter_attachment = LetterAttachment.query.get_or_404(letter_attachment_id)
+    # I found I had to do this to set up the versioning properly, rather than use the template directly.
+    if (
+        letter_attachment.template is None
+        or (template := dao_get_template_by_id(letter_attachment.template.id)) is None
+    ):
+        raise InvalidRequest("template-does-not-exist", 400)
```
samuelhwilliams, line 60:
```suggestion
    letter_attachment = LetterAttachment.query.get_or_404(letter_attachment_id)
    if not (template := letter_attachment.template):
        raise InvalidRequest("template-does-not-exist", 400)
```

I think we should be able to do this? But not sure exactly what problem you ran into with the comment as I'm not seeing it. 🤔

whpearson: It occurred when I assigned template before the if statement. Your version works fine. I don't know the intricacies of the template versioning system, I might dig if I get a chance.

samuelhwilliams: Unfortunately we have about 3 versioning systems.. :(

app/schemas.py
```diff
@@ -478,7 +478,7 @@ def get_reply_to_text(self, template):
         return template.get_reply_to_text()
 
     def get_letter_attachment(self, template):
-        return template.letter_attachment_id
+        return template.letter_attachment.serialize() if template.letter_attachment else None
```
samuelhwilliams, line 481: Do we need to worry about this return value changing (ie the raw ID already in use by admin) or are we free to change this still?

app/letter_attachment/rest.py
```diff
@@ -41,3 +43,20 @@ def create_letter_attachment():
...
+
+@letter_attachment_blueprint.route("/letter-attachment/<uuid:letter_attachment_id>/archive", methods=["POST"])
+def archive_letter_attachment(letter_attachment_id):
+    data = request.get_json()
+
+    validate(data, post_archive_letter_attachment_schema)
+
+    letter_attachment = LetterAttachment.query.get_or_404(letter_attachment_id)
+    if not (template := letter_attachment.template):
+        raise InvalidRequest("template-does-not-exist", 400)
```
samuelhwilliams, line 56: nit: maybe `letter-attachment-already-archived` is a better error? Optional

samuelhwilliams (commented):
I think this is basically good to go, provided the two comments below are OK 👍

samuelhwilliams (approved):
Nice 👍👍

PR #3793, opened 2023-05-04
Split api and document download host variables

klssmith:
This repo calls endpoints in admin, template-preview, document-download-api and in itself. When running on ECS, we want to use internal routes where possible and to only use public routes when necessary. The `ADMIN_BASE_URL` should always be the public route (since this is used to build links for users). `TEMPLATE_PREVIEW_API_HOST` can always be the internal route.

However, when calling document-download-api and api itself we want to use a mix of private and public routes. This splits up the environment variables for those hosts so that in a later PR we can change the code to use internal routes in places.


sakisv (approved):
👍

PR #3795, opened 2023-05-04
Use both internal and external URLs for api and document-download-api

klssmith:
When the apps are deployed on ECS we want to use internal hostnames when talking to other apps where possible, so have split out the environment variables in preparation for this https://github.com/alphagov/notifications-api/pull/3793

This PR changes the code to use the internal hostname where possible when calling itself or document-download-api. When deployed on PaaS both values will be the same, but when deployed on ECS the external and internal hostnames will be different.


samuelhwilliams (approved):
Looks good 👍

PR #3798, opened 2023-05-05
Return letter attachment data from template endpoint

leohemsted:
this is useful, as we typically pass the template json blob straight to template preview without modification when generating letters. If we return the full letter attachment object (note this is just id, metadata, etc, not the actual PDF), then that's one less thing we have to worry about handling separately elsewhere.

The template json now looks like. (note this only affects template json in endpoints where we already returned the letter_attachment id)

```json
{
...
	"id": "59e89e0d-a68e-4d67-9449-0f277042c524",
	"letter_attachment": {
		"archived_at": null,
		"archived_by_id": null,
		"created_at": "2023-05-05T14:25:48.524912Z",
		"created_by_id": "b97004cd-7ab3-4766-b1bd-c9552d4331d7",
		"id": "e0d139f5-0493-4eca-8ec0-da21164a7bed",
		"original_filename": "abc.pdf",
		"page_count": 1
	},
	"name": "letter Template Name",
...
}
```

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -392,7 +392,7 @@ class BaseTemplateSchema(BaseSchema):
         return template.get_reply_to_text()
 
     def get_letter_attachment(self, template):
-        return template.letter_attachment_id
+        return template.letter_attachment.serialize() if template.letter_attachment else None
 
     class Meta(BaseSchema.Meta):
         model = models.Template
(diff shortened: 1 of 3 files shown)


whpearson (approved):
Changes look good. Tested locally.

PR #3801, opened 2023-05-11
Add index to jobs.contact_list_id field

samuelhwilliams:
We do a lot of lookups on this field and having an index makes sense. This missing index has eventually caused a (poorly optimised) endpoint (get_contact_list) to start timing out for some users with a lot of contact lists.


whpearson (approved):
Checked the migration does what it says (and the downgrade works too).

Looks good.

samuelhwilliams:
Sorry @willp-bl  - have just switched this over to use `postgresql_concurrently` for the migration which should prevent a full lock on the table. I don't think it would be a problem because the table isn't that big - but good practice still I think.

PR #3802, opened 2023-05-15
Bump utils to 63.0.0

samuelhwilliams:
# 63.0.0

* Remove the `technical_ticket` parameter from NotifySupportTicket; replace with an optional `notify_ticket_type` that takes a NotifyTicketType enum value. If provided, this will maintain the existing behaviour where tickets are tagged as technical/non-technical on creation. If omitted, then tickets will have no ticket type, which may help us have a clearer process around triaging tickets.

 ## 62.4.0

* Add `class="page--first"` and `class="page--last"` to the first and last pages (respectively) of letters rendered with `LetterImageTemplate`

 ## 62.3.1

* Change `logger.exception` to `logger.warning` for log formatting error.

 ## 62.3.0

* Adds support for Chinese character sets in letters ***

Complete changes: https://github.com/alphagov/notifications-utils/compare/62.3.1...63.0.0

### DIFF
--- a/app/broadcast_message/utils.py
+++ b/app/broadcast_message/utils.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from flask import current_app
 from notifications_utils.clients.zendesk.zendesk_client import (
     NotifySupportTicket,
+    NotifyTicketType,
 )
 
 from app import zendesk_client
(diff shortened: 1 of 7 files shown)


whpearson (approved):
Looks good. Ticket types make sense for the different problems

PR #3804, opened 2023-05-17
Add an admin search endpoint for organisations

samuelhwilliams:
This will let us update the platform-admin search page to lookup organisations.


whpearson (approved):
Looks good. Is it worth testing the escaping of special characters for search?

samuelhwilliams:
Good call - I copied the tests we have for the similar services endpoint, but let me add a test for each covering special chars 👍

PR #3805, opened 2023-05-22
Remove use of `select *` in a query

idavidmcdonald:
This is done for two reasons.

Firstly, it is bad practice to use `select *`, backed up by https://stackoverflow.com/questions/321299/what-is-the-reason-not-to-use-select

Secondly, we've got a new database in our staging environment that has the same columns in the notifications table, but the order of the columns is slightly different. `select *` returns data in the default order and this order is different between the `notifications` and `notifications_history` tables meaning that this query currently errors. This column ordering difference was introduced in:
https://github.com/alphagov/notifications-aws/blob/main/scripts/remove-all-sensitive-and-personal-data/remove-all-sensitive-and-personal-data-part-2.sql

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -321,7 +321,10 @@ def insert_notification_history_delete_notifications(
     # Insert into NotificationHistory if the row already exists do nothing.
     insert_query = """
         insert into notification_history
-         SELECT * from NOTIFICATION_ARCHIVE
+         SELECT id, job_id, job_row_number, service_id, template_id, template_version, api_key_id,
+             key_type, notification_type, created_at, sent_at, sent_by, updated_at, reference, billable_units,
+             client_reference, international, phone_prefix, rate_multiplier, notification_status,
+              created_by_id, postage, document_download_count from NOTIFICATION_ARCHIVE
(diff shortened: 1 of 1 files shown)


samuelhwilliams (approved):
👍 

When I was looking at this I was half-tempted to extra the list of columns to a separate local var at the top of the function and then just `{}` format that into each of the strings - as there are now four places that use the same list of columns.

But up to you.

This could well be a separate thing, but I'd even be tempted to extract this column list to a separate global variable (list[str]) and make a fairly naive test asserting that `len(list) == len(Notifications.columns) - x`, so that if we ever add a column to Notifications we have some test that will definitely fail and make us check whether we need to update this task to include the field into notification_history. Feels like potentially there's a risk we could permanently lose some data we might want to keep. Unless we already have something like this?

idavidmcdonald:
Done as you suggested Sam, thanks for that.

Note, I am checking the list matches the number of columns in NotificationHistory which I think is a reliable way to determine how many columns of data should be transferred from Notifications into NotificationHistory (rather than as you suggested taking the count of Notifications and removing a hardcoded number of columns that contain personal data as).

samuelhwilliams (approved):
Nice one 👍

PR #3807, opened 2023-05-23
Record letter despatch dates

samuelhwilliams:
Adds a new table `NotificationLetterDespatch` that records when we receive confirmation that a letter has been despatched. In the future we will use this to issue billing reports to our print provider.

There are a few refactors in here so best to go commit-by-commit.

There is one WIP/fixme/todo comment that I'm open for thoughts on but also may leave to address separately. For now the implementation is technically fine but if we can tighten up the FK safely then that's also great.


leohemsted (approved):
i think this looks pretty good 👍

PR #3808, opened 2023-05-23
enable DVLA api to be toggled per env by postage type

leohemsted:
add a new `DVLA_API_POSTAGE_TYPE_EXCLUDE_LIST` (that defaults to empty list, as in exclude nothing, as in send everything via the API). if the API is disabled, it'll still do nothing so this is safe to merge now.

after this is in, we can merge credentials to enable api on prod but exclude all postage types (so in effect, no different to now)

then after that we can start removing postage types from the exclude list

- [ ] https://github.com/alphagov/notifications-credentials/pull/349


app/config.py
```diff
@@ -450,6 +450,7 @@ class Config(object):
     BROADCAST_ORGANISATION_ID = "38e4bf69-93b0-445d-acee-53ea53fe02df"
 
     DVLA_API_ENABLED = os.environ.get("DVLA_API_ENABLED") == "1"
+    DVLA_API_POSTAGE_TYPE_EXCLUDE_LIST = json.loads(os.environ.get("DVLA_API_POSTAGE_TYPE_EXCLUDE_LIST", "[]"))
```
samuelhwilliams, line 453: As long as we're super careful about making sure all the env vars are in place this is fine, but a more cautious approach might be to either have `DVLA_API_POSTAGE_TYPE_INCLUDE_LIST` or to specify all of the postage types as the default if the env var isn't set?

samuelhwilliams (approved):
Changes look good and should work.

One thing to maybe confirm with DVLA: at the start of the letters task we send an email over to DVLA saying how many letters/pages to expect. We should probably let them know/confirm to them that these will be split across the API and FTP - if their current process for reconciliation only takes FTP into account they might get confused.

PR #3812, opened 2023-06-01
Explicitly list columns for notification migration

samuelhwilliams:
And add a test to validate this works correctly.

In staging the `notifications_history` table may have columns in a different order because of how we migrate+anonymise the table from production. Our nightly process that moves notifications from the live table to the history table still expects those columns to be in the same order; this patch explicitly lists the columns to insert into


idavidmcdonald (approved):
Good spot Sam, thank you.

PR #3815, opened 2023-06-06
Add some indexes to notification history

samuelhwilliams:


app/models.py
```diff
@@ -1574,8 +1574,8 @@ class NotificationHistory(db.Model, HistoryModel):
     api_key = db.relationship("ApiKey")
     key_type = db.Column(db.String, db.ForeignKey("key_types.name"), unique=False, nullable=False)
     billable_units = db.Column(db.Integer, nullable=False, default=0)
-    notification_type = db.Column(notification_types, nullable=False)
-    created_at = db.Column(db.DateTime, unique=False, nullable=False)
+    notification_type = db.Column(notification_types, nullable=False, index=True)
```
idavidmcdonald, line 1577: We only expect this to be one of 3 things, right. I'm not sure how valuable an index might be in that case? I recall that indexes aren't so performant if they only cover very few values

samuelhwilliams: Yeah this is probably not a useful one to be honest, you're right 👍  Especially with the uneven distribution across the notification types.

PR #3822, opened 2023-06-14
Org user permissions

samuelhwilliams:
API PR 1 of 2 (maybe 3)

## Add organisation user permissions table
Adds the table which stores the actions a user can take for a given organisation. Right now we're only adding one permission (the ability to make services live) and everything else is just implicit if you're a member of the org.

## Update endpoint when adding a user to an org
So that the new permission is automatically set for new org members.

## Add admin endpoint for managing org user permissions
To prepare for the admin pages that will allow adding/removing permissions.

## Store permissions on invited org users
When inviting a user to an org, in the same way that we do with invited service users, store their eventual permissions on the invited user model.

---

To do in the future (after initial admin release):
- read permissions when adding user to the org (admin sends permissions across from the invite ... for some reason - being consistent with how service user invites work here).
- remove default for permissions field

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2,6 +2,7 @@ import datetime
 import enum
 import itertools
 import uuid
+from collections import defaultdict
 
 from flask import current_app, url_for
 from notifications_utils.insensitive_dict import InsensitiveDict
@@ -172,6 +173,15 @@ class User(db.Model):
             retval[service_id].append(x.permission)
         return retval
 
+    def get_organisation_permissions(self) -> dict[str, list[str]]:
+        from app.dao.organisation_user_permissions_dao import organisation_user_permissions_dao
+
+        retval = defaultdict(list)
+        for p in organisation_user_permissions_dao.get_permissions_by_user_id(self.id):
+            retval[str(p.organisation_id)].append(p.permission.value)
+
+        return retval
+
     def serialize(self):
         return {
             "id": self.id,
@@ -186,6 +196,7 @@ class User(db.Model):
             "organisations": [x.id for x in self.organisations if x.active],
             "password_changed_at": self.password_changed_at.strftime(DATETIME_FORMAT_NO_TIMEZONE),
             "permissions": self.get_permissions(),
+            "organisation_permissions": self.get_organisation_permissions(),
             "platform_admin": self.platform_admin,
             "services": [x.id for x in self.services if x.active],
             "can_use_webauthn": self.can_use_webauthn,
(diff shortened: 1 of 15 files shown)


migrations/versions/0415_org_invite_perms.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+revision = "0415_org_invite_perms"
+down_revision = "0414_org_user_permissions"
+
+
+def upgrade():
+    op.add_column("invited_organisation_users", sa.Column("permissions", sa.String(), nullable=True))
+    op.execute("UPDATE invited_organisation_users SET permissions = 'can_make_services_live'")
+    op.alter_column("invited_organisation_users", "permissions", existing_nullable=True, nullable=False)
```
samuelhwilliams, line 19: yes this is not strictly backwards-compatible but we can accept not being able to invite org users while the api is deploying

app/models.py
```diff
@@ -158,7 +160,7 @@ def password(self, password):
     def check_password(self, password):
         return check_hash(password, self._password)
 
-    def get_permissions(self, service_id=None):
+    def get_permissions(self, service_id=None, organisation_id=None):
```
whpearson, line 163: Why is the organisation_id included here? It doesn't appear to be used.

samuelhwilliams: Good spot - looks like a leftover from something I tried initially. Will remove :)

whpearson:
Is adding permissions to the organisation_user_permissions table going to be in another ticket?

app/models.py
```diff
@@ -172,6 +174,23 @@ def get_permissions(self, service_id=None):
...
+                    self.id, organisation_id
+                )
+            ]
+
+        retval = defaultdict(list)
+        for p in organisation_user_permissions_dao.get_permissions_by_user_id(self.id):
+            retval[str(p.organisation_id)].append(p.permission.value)
+
+        return retval
+
```
whpearson, line 193: Its a bit confusing that this changes the shape of what is returned based on the input parameters (either a list or a dict). But that may just be my prejudices.

samuelhwilliams: I don't really disagree, to be honest - my only thought process here was to copy what happens for `get_permissions` above (service permissions).

But in reality, `get_organisation_permissions` doesn't actually ever get called with an org_id at the moment, so I could pull this out for now?

whpearson (approved):
Looks good. Tested locally db migrations look fine.

PR #3829, opened 2023-06-22
Restrict org users that get 'go live' emails

samuelhwilliams:
We are rolling out an org-user level permission to say who can make services live. Given this, we should only send emails about services requesting to go live to users with the relevant permission.


whpearson (approved):
Changes look good

PR #3830, opened 2023-06-22
Respect invited org user permissions

samuelhwilliams:
Now that the admin app is always sending across permissions for an invited user, we can read and respect those permissions.

Note: the admin app is currently sending a hard-coded list of permissions, so it's not user-editable just with this PR. But this prepares us to respect dynamic permissions when the admin app has been updated.


app/organisation/rest.py
```diff
@@ -206,7 +206,8 @@ def get_organisation_services_usage(organisation_id):
 
 @organisation_blueprint.route("/<uuid:organisation_id>/users/<uuid:user_id>", methods=["POST"])
 def add_user_to_organisation(organisation_id, user_id):
-    new_org_user = dao_add_user_to_organisation(organisation_id, user_id)
+    permissions = request.get_json()["permissions"]
```
leohemsted, line 209: minor optional - does it make sense to do the transformation here rather than in the dao method (which can then just have type `permissions: list[str]`)

```suggestion
    permissions = [p["permission"] for p in request.get_json()["permissions"]]
```

leohemsted (approved):

app/organisation/rest.py
samuelhwilliams: I might change the dao method to `list[OrganisationUserPermissionType]` - I'd prefer to keep the stricter typing info until the last possible moment. But good point in general.

PR #3832, opened 2023-07-03
Use droplet for db-migration paas app

samuelhwilliams:
Let's use the built droplet when running db migrations, rather than making a new one from scratch (which involves pulling+installing dependencies etc). Should make deploys faster when we have migrations to run.


whpearson (approved):
Looks good

PR #3837, opened 2023-07-12
Don't restart app for db migrations

samuelhwilliams:
This app has 0 instances by default. `cf restart` on an app will bump it to 1 instance no matter what, which we don't want.


sakisv (approved):
Nothing is ever as straightforward as it seems, is it?

PR #3839, opened 2023-07-22
Add detect-secrets pre-commit hook

samuelhwilliams:
This integrates [detect-secret](https://github.com/Yelp/detect-secrets), which scans new commits to the repository for known/suspected secret values.

We integrate this using pre-commit, so that every new commit can be scanned (assuming pre-commit is enabled by the developer) and secrets flagged before they can be pushed and leaked.


samuelhwilliams:
we are using github push protection instead

PR #3843, opened 2023-08-03
Redis

robbrazil:
Added redis_url variable to allow the api app to call redis.
Modified two files:
1. cloudfoundry_config.py where we added and IF statement to check if the variable is in the environment. If not it will use the VCAP_SERVICES one
2. manifest.yml.j2 where we check if redis_url variable is defined and take its value from the env variable set

### DIFF
--- a/app/cloudfoundry_config.py
+++ b/app/cloudfoundry_config.py
@@ -11,4 +11,5 @@ def extract_cloudfoundry_config():
             "postgres", "postgresql"
         )
     # Redis config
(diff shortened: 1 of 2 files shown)


idavidmcdonald (approved):
Looks great.

Could benefit from squashing the two commits into a single commit. Although they touch different files, combined they are trying to solve one single problem which is to get the app using a REDIS_URL provided by environment variable (if it exists) rather than using the one from the PaaS.

PR #3849, opened 2023-08-15
QR code validation

samuelhwilliams:
Validate QR code length when:

* creating/updating a template
* sending a notification via the API

Importantly, this validation happens up front. If a notification with a
QR code that is 'too long' did manage to get accepted by the API,
template-preview *would* render a QR code for it and the letter would be
sent. This is so that we don't have to handle moving letters with QR
codes into `permanent-failure` and then explain why that has happened.

We may revisit this in the future, but for now we are preferring to stop
'bad QR codes' getting into the system at all, rather than handling
errors further downstream and having to bubble them back up.


leohemsted (approved):
looks good pending content work

PR #3851, opened 2023-08-18
Create functional test seeded users endpoint

samuelhwilliams:
Adds an endpoint to the API, only available in development and preview
environments, which can create users for the functional tests.

These users are created against a pre-seeded organisation/service, so
that is expected to exist in the DB already. But creating users on
demand allows us to run tests interacting with that org/service in
parallel, which should lead to much faster run times for our common
journeys

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -117,10 +117,15 @@ def create_app(application):
     return application
 
 
+def _should_register_functional_testing_blueprint(environment):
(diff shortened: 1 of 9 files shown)


leohemsted (commented):
im okay with this but i wouldn't mind chatting with the wider tech team to decide if this approach is the preferred way to solve this problem

CrystalPea (approved):

PR #3853, opened 2023-08-21
Bump utils to 70.0.1

samuelhwilliams:
70.0.0: updates validation messages for invalid phone numbers; we override these to retain API backwards compatibility
70.0.1: fixes `<strong>` tagging around QR codes during validation

---

We override the updated content for invalid phone errors so as to keep our API interface the same. We may unroll this in the future.


CrystalPea (approved):
A bit reluctant approve to unblock the QR code development - we need to ensure the remapping of the error messages is temporary - it's very awkward.

PR #3857, opened 2023-08-23
Chown /opt/venv faster

samuelhwilliams:
Copying the directory via a Dockerfile command is faster than running a command inside the build container to update ownership

```
 => [notify-api test 3/6] COPY --from=python_build --chown=notify:notify /opt/venv /opt/venv        2.8s
```

vs

```
 => [notify-api test 1/6] RUN chown -R notify:notify /opt/venv                                      9.2s
```


whpearson (approved):
Changes look good. Tried out locally too.

sakisv (approved):

PR #3858, opened 2023-08-25
Allow tests to run in or out of docker

samuelhwilliams:
When running inside docker we have different hostnames for our apps, so we can't hardcode them inside tests. We can add a fixture that pulls the hostnames for our apps from environment variables, allowing tests to pass either locally or inside docker.


whpearson (approved):
Tests run happily. Changes make sense.

PR #3860, opened 2023-08-29
Add celery task to delete one quarter of old notification history

idavidmcdonald:
https://trello.com/c/8qFHZC6a/353-delete-data-from-notificationhistory-older-than-april-2023

## For the reviewer
Please pay particular attention to:

- have I got the correct retention limit that matches what is expected in our decision record? If we get this wrong, we could delete a big bunch of data we actually want
- is the logic for knowing when not to delete records correct? If we get this wrong, we could delete a big bunch of data we actually want
- are you happy with the log statements? Will they make these easy to debug if we need to?

## What this does and additional context

This will run at 9pm (a quietish time but hopefully not going to conflict with the other nightly tasks that start around midnight, although if it does conflict it should still be OK).

This will start at 9pm on the day of merging and then should run for 14 days, each time deleting the oldest quarter of data. When it runs on the 15th day (and every day after that), it will see there are no notifications older than 1 April 2023 and do nothing. We could at that point, delete the celery beat configuration for it so we stop running a task that isn't going to do anything every night.

I assume that deleting one quarter of data at a time will be OK. The query will probably take quite a long time to run and delete a quarter of data, probably a few hours. One risk is that the celery work will get scaled down or restarted during this time. That shouldn't be the end of the world as the task will just run again tomorrow. Deleting notification_history isn't some urgent thing that must happen every night, if it gets delayed by a day that is fine. If we do start seeing issues with these queries taking several hours and erroring for some reason, we can take a similar approach to other mass deletion functions we have which run multiple queries, each one deleting 10,000 records at a time. A bit more complex python is all that is needed.

In july of 2024, we will be able to update the retention limit to April 2024 if we want so that this process will happen again for next years deletion.

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -11,7 +11,7 @@ from notifications_utils.letter_timings import (
     get_dvla_working_day_offset_by,
     is_dvla_working_day,
 )
-from notifications_utils.timezones import convert_utc_to_bst
+from notifications_utils.timezones import convert_bst_to_utc, convert_utc_to_bst
 from sqlalchemy import func
 from sqlalchemy.exc import SQLAlchemyError
 
@@ -340,32 +340,53 @@ def save_daily_notification_processing_time(bst_date=None):
 
 @notify_celery.task(name="delete-oldest-quarter-of-unneeded-notification-history")
 def delete_oldest_quarter_of_unneeded_notification_history():
-    # This retention limit is hardcoded for the moment. It is picked from
+    # This function will delete either 1 or 0 quarters of notifications from the notification_history
+    # table. It will pick the oldest quarter available before our retention limit and delete that.
+    # If there are no notifications older than our retention limit, then it won't do anything.
+    # By running this task several nights in a row, it will slowly delete a quarter of notification_history
+    # per night until we reach out retention limit
+    #
+    # We do everything in this function in BST up until the final point
+    # This is because we care about quarters in the BST timezone because this is
+    # how we bill our users and is easier to reason about.
+    # When we finally figure out which BST quarter we can delete, we convert the time
+    # to UTC so that we delete the correct notifications from our database
+    #
+    # This retention limit is hardcoded and picked from
     # https://github.com/alphagov/notifications-aws/blob/main/decisions/2022-12-01-notification-history-retention-period.md
     # and can be increased in the future when we move 3 months into the next financial year
-    retention_limit = datetime(2023, 4, 1)
+    retention_limit_bst = datetime(2023, 4, 1)
 
-    oldest_record = db.session.query(func.min(NotificationHistory.created_at)).one()[0]
-    if oldest_record >= retention_limit:
+    oldest_notification_bst = convert_utc_to_bst(db.session.query(func.min(NotificationHistory.created_at)).one()[0])
+    if oldest_notification_bst >= retention_limit_bst:
         current_app.logger.info(
-            "No notifications older than retention date of %s so nothing to delete", retention_limit
+            "No notifications older than retention date of %s so nothing to delete", retention_limit_bst
         )
         return
 
-    # Pick a potential deletion date to start, that we know is older than the oldest notification
-    # in notification_history at the time of writing this code
+    deletion_date_bst = pick_bst_quarter_to_delete_back_from(oldest_notification_bst)
+
+    # in case there is an error in our logic above, let us just double check that we aren't trying to delete for
+    # anything after our retention limit
+    if deletion_date_bst > retention_limit_bst:
+        current_app.logger.exception("Attempted to delete past our notification_history retention limit")
+        raise
+
+    deletion_date_utc = convert_bst_to_utc(deletion_date_bst)
+
+    delete_notification_history_older_than_datetime(deletion_date_utc)
+
+
+def pick_bst_quarter_to_delete_back_from(oldest_notification_bst):
+    # We've picked a potential deletion date to start that we know is older than the oldest notification
+    # in notification_history at the time of writing this code.
+    # The potential deletion date to start is intentionally the BST start of a new quarter
     deletion_date = datetime(2019, 10, 1)
-    while deletion_date <= oldest_record:
+    while deletion_date <= oldest_notification_bst:
         # There are no notifications older than our deletion date, lets add 3 months
         # and see if that is still true
         # When we finally find a deletion date that has notifications older than the deletion date
         # we can then delete anything older (ie the previous quarters data)
         deletion_date = deletion_date + dateutil.relativedelta.relativedelta(months=3)
 
-    # in case there is an error in our logic above, let us just double check that we aren't trying to delete for
-    # anything after our retention limit
-    if deletion_date > retention_limit:
-        current_app.logger.exception("Attempted to delete past our notification_history retention limit")
-        raise
-
-    delete_notification_history_older_than_datetime(deletion_date)
+    return deletion_date
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -352,10 +352,17 @@ def delete_oldest_quarter_of_unneeded_notification_history():
     # When we finally figure out which BST quarter we can delete, we convert the time
     # to UTC so that we delete the correct notifications from our database
     #
-    # This retention limit is hardcoded and picked from
+    # This retention limit is hardcoded and was originally picked from
     # https://github.com/alphagov/notifications-aws/blob/main/decisions/2022-12-01-notification-history-retention-period.md
-    # and can be increased in the future when we move 3 months into the next financial year
-    retention_limit_bst = datetime(2023, 4, 1)
+    # It was supposed to be 2023-4-1.
+    # However, at the time of writing this code we realised the FtBillingLetterDispatch table has been introduced
+    # which means we need to store letters older than 2023-4-1 in order to rebuild that table (because that table
+    # uses the date of dispatch, not the date of creation for which date to bill for). To keep it simple, we keep
+    # an extra quarters worth of data giving us plenty of buffer
+    #
+    # In the future, we will be able to update this retention_limit_bst value when we have progressed 3 quarters
+    # into the next financial year
+    retention_limit_bst = datetime(2023, 1, 1)
 
     oldest_notification_bst = convert_utc_to_bst(db.session.query(func.min(NotificationHistory.created_at)).one()[0])
     if oldest_notification_bst >= retention_limit_bst:
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -342,7 +342,7 @@ def delete_oldest_quarter_of_unneeded_notification_history():
     # This retention limit is hardcoded for the moment. It is picked from
     # https://github.com/alphagov/notifications-aws/blob/main/decisions/2022-12-01-notification-history-retention-period.md
     # and can be increased in the future when we move 3 months into the next financial year
-    retention_limit = datetime(2022, 4, 1)
+    retention_limit = datetime(2023, 4, 1)
 
     oldest_record = db.session.query(func.min(NotificationHistory.created_at)).one()[0]
     if oldest_record >= retention_limit:
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -376,8 +376,7 @@ def delete_oldest_quarter_of_unneeded_notification_history():
     # in case there is an error in our logic above, let us just double check that we aren't trying to delete for
     # anything after our retention limit
     if deletion_date_bst > retention_limit_bst:
-        current_app.logger.exception("Attempted to delete past our notification_history retention limit")
-        raise
+        raise Exception("Attempted to delete past our notification_history retention limit")
 
     deletion_date_utc = convert_bst_to_utc(deletion_date_bst)
 
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -32,6 +32,7 @@ from app.dao.jobs_dao import (
     dao_archive_job,
     dao_get_jobs_older_than_data_retention,
 )
+from app.dao.notification_history_dao import delete_notification_history_older_than_datetime
 from app.dao.notifications_dao import (
     dao_get_notifications_processing_time_stats,
     dao_timeout_notifications,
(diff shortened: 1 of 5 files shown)


klssmith (commented):
Nice work! I like the way you pick a deletion date

I have a couple of questions / comments about the exact point that we want to delete data from `notification_history` from. Is the plan to only keep data from the current financial year in case we need to ever rebuild billing? If we do want to be able to rebuild billing for the whole financial year, I think we'd need to keep the extra couple of weeks before the April 1st too. Partly because of timezones - notifications created between 31 March 23:00 and 1 April 00:00 will also be billed this year. But also because since around June we now bill for letters based on when they were _despatched_, which can be some time after they were created or sent to DVLA. We would need to keep some data from before the quarter in the `notifications_history` table in order to [populate the `ft_billing_letter_despatch` table](https://github.com/alphagov/notifications-api/blob/6935b240ed494f46361bcf2e322060bd773ca2da/app/dao/fact_billing_dao.py#L733) for the quarter.

app/celery/nightly_tasks.py
```diff
@@ -334,3 +335,42 @@ def save_daily_notification_processing_time(bst_date=None):
...
+    oldest_record = db.session.query(func.min(NotificationHistory.created_at)).one()[0]
+    if oldest_record >= retention_limit:
+        current_app.logger.info(
+            "No notifications older than retention date of %s so nothing to delete", retention_limit
+        )
+        return
+
+    # Pick a potential deletion date to start, that we know is older than the oldest notification
+    # in notification_history at the time of writing this code
+    deletion_date = datetime(2019, 10, 1)
```
klssmith, line 356: The `deletion_date` must be the start of quarter otherwise we risk deleting data that is still within the retention limit. Should we do a check somewhere in the `while` loop to check that `deletion_date` is earlier than the `retention_limit`? 

It's not a problem with the code as it is now, it would be more to stop someone accidentally changing `deletion_date` to `datetime(2022, 2, 1)` (for example) and then deleting everything up to 1st May 2022 instead of 1st April 2022.

app/celery/nightly_tasks.py
```diff
@@ -334,3 +335,42 @@ def save_daily_notification_processing_time(bst_date=None):
...
+        deletion_date = deletion_date + dateutil.relativedelta.relativedelta(months=3)
+
+    delete_notification_history_older_than_datetime(deletion_date)
+
+
+def delete_notification_history_older_than_datetime(older_than_datetime):
+    current_app.logger.info("Beginning to delete notification_history older than %s", older_than_datetime)
+
+    num_rows_deleted = NotificationHistory.query.filter(
+        NotificationHistory.sent_at < older_than_datetime,
```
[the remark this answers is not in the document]
idavidmcdonald: I kind of wasn't sure which to use out of the three of them. I think I very briefly thought, at which point does a notification become 'billable' and I guessed that would be when it is sent.

It definitely does make sense that it should be consistent with the value we pick in `oldest_record`. 

Giving it some more thought, I think I should make it `created_at`.

For our code that populates the `ft_billing` table, we decide which notifications are eligible for billing for a particular day based on their `created_at` - https://github.com/alphagov/notifications-api/blob/main/app/dao/fact_billing_dao.py#L552

Letters does appear to be the exception as you mention, but I think we should go for consistency with our ft_billing table because we are picking our retention date based on our billing needs (ie the ft billing table).

I'll address what we do about letters in a separate comment... 

As an aside, it looks like we still write into ft_billing for letters based on the Notification.created_at date (https://github.com/alphagov/notifications-api/blob/main/app/dao/fact_billing_dao.py#L618) but now are also writing into FactBillingLetterDespatch table based on the dispatch date for letters. I'm a bit confused about why we are still writing into ft_billing for letters if we have this new FactBillingLetterDespatch table which I'm guessing we are now using for billing. Feels like we are left with two sources of data which might get confused? Any idea if we are going to reduce these down to a single one in time?

app/celery/nightly_tasks.py
```diff
@@ -334,3 +335,42 @@ def save_daily_notification_processing_time(bst_date=None):
...
         )
     )
+
+
+@notify_celery.task(name="delete-oldest-quarter-of-unneeded-notification-history")
+def delete_oldest_quarter_of_unneeded_notification_history():
+    # This retention limit is hardcoded for the moment. It is picked from
+    # https://github.com/alphagov/notifications-aws/blob/main/decisions/2022-12-01-notification-history-retention-period.md
+    # and can be increased in the future when we move 3 months into the next financial year
+    retention_limit = datetime(2022, 4, 1)
```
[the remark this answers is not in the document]
idavidmcdonald: Yes, I think it should be 2023 and I must have picked 2022 because that is what the ADR referred to (but as you point out, the ADR was written last financial year).

I will update it (and same with the relevant tests that check the around the time of the retention limit) and rebase the change in.

app/celery/nightly_tasks.py
idavidmcdonald: Good suggestion.

I'll add a commit with this. Not only will it cover the case where someone decides to change the `deletion_date`, but also the case where there is some type of other bug in the while loop logic.

tests/app/celery/test_nightly_tasks.py
```diff
@@ -577,3 +580,71 @@ def test_delete_notifications_task_calls_task_for_services_that_have_sent_notifi
             ),
         ],
     )
+
```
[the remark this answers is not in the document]
idavidmcdonald: Good idea, new commit about to be added which does this. This gives me some more confidence that things are indeed working as expected (and helped me spot a bug in the code).

app/celery/nightly_tasks.py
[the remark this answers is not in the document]
idavidmcdonald: I didn't end up rebasing the commit in, but left it as a separate commit for easy readability of my change.

app/celery/nightly_tasks.py
```diff
@@ -334,3 +335,42 @@ def save_daily_notification_processing_time(bst_date=None):
...
+
+    delete_notification_history_older_than_datetime(deletion_date)
+
+
+def delete_notification_history_older_than_datetime(older_than_datetime):
+    current_app.logger.info("Beginning to delete notification_history older than %s", older_than_datetime)
+
+    num_rows_deleted = NotificationHistory.query.filter(
+        NotificationHistory.sent_at < older_than_datetime,
+    ).delete()
```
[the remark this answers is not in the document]
idavidmcdonald: Good point. Will add.

Out of interest, any suggestions for how I could have picked up on this earlier? Or how I could write my unit tests to pick up on the lack of this?

It also probably makes sense to move this function into one of the dao files for consistency with most other delete functions.

app/celery/nightly_tasks.py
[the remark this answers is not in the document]
idavidmcdonald: I've added a commit which changes it to use `created_at` rather than `sent_at`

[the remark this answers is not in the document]
klssmith: I agree that `created_at` sounds good, and letters can be considered separately. I haven't heard of any plans to not have letters included in both ft_billing and ft_billing_letter_despatch

app/celery/nightly_tasks.py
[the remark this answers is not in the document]
klssmith: I'm not too sure... I don't remember seeing this tested in other places. Moving it into a DAO file looks good

app/celery/nightly_tasks.py
```diff
@@ -334,3 +336,57 @@ def save_daily_notification_processing_time(bst_date=None):
...
+        )
+        return
+
+    deletion_date_bst = pick_bst_quarter_to_delete_back_from(oldest_notification_bst)
+
+    # in case there is an error in our logic above, let us just double check that we aren't trying to delete for
+    # anything after our retention limit
+    if deletion_date_bst > retention_limit_bst:
+        current_app.logger.exception("Attempted to delete past our notification_history retention limit")
+        raise
```
klssmith, line 373: This is really minor (since it does still raise), but calling `raise` outside of a an active Exception like this this gives the error of `RuntimeError: No active exception to reraise` so you could change it to `raise Exception("...")` instead.

app/celery/nightly_tasks.py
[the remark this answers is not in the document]
CrystalPea: > I'm a bit confused about why we are still writing into ft_billing for letters if we have this new FactBillingLetterDespatch table which I'm guessing we are now using for billing. Feels like we are left with two sources of data which might get confused? Any idea if we are going to reduce these down to a single one in time?

I think it's because we use `FactBillingLetterDespatch` for us being billed by DVLA, and `ft_billing` for us billing the service users. It's because we bill our users the same way we used to, but the way we get billed by DVLA has changed.

app/celery/nightly_tasks.py
```diff
@@ -334,3 +336,63 @@ def save_daily_notification_processing_time(bst_date=None):
...
+    # However, at the time of writing this code we realised the FtBillingLetterDispatch table has been introduced
+    # which means we need to store letters older than 2023-4-1 in order to rebuild that table (because that table
+    # uses the date of dispatch, not the date of creation for which date to bill for). To keep it simple, we keep
+    # an extra quarters worth of data giving us plenty of buffer
+    #
+    # In the future, we will be able to update this retention_limit_bst value when we have progressed 3 quarters
+    # into the next financial year
+    retention_limit_bst = datetime(2023, 1, 1)
+
+    oldest_notification_bst = convert_utc_to_bst(db.session.query(func.min(NotificationHistory.created_at)).one()[0])
```
CrystalPea, line 367: Maybe we could add log line that will tell us when the oldest notification was from? That could be useful for debugging. For example:

`delete_oldest_quarter_of_unneeded_notification_history: Oldest notification found in notification_history table was created_at {oldest_notification_bst} BST.`

CrystalPea (approved):
I think this is good to go, Katie's review was very thorough, I couldn't find anything more really, just a potential additional log line:

app/celery/nightly_tasks.py
```diff
@@ -334,3 +335,42 @@ def save_daily_notification_processing_time(bst_date=None):
...
+
+
+@notify_celery.task(name="delete-oldest-quarter-of-unneeded-notification-history")
+def delete_oldest_quarter_of_unneeded_notification_history():
+    # This retention limit is hardcoded for the moment. It is picked from
+    # https://github.com/alphagov/notifications-aws/blob/main/decisions/2022-12-01-notification-history-retention-period.md
+    # and can be increased in the future when we move 3 months into the next financial year
+    retention_limit = datetime(2022, 4, 1)
+
+    oldest_record = db.session.query(func.min(NotificationHistory.created_at)).one()[0]
```
samuelhwilliams: I think this query will read the entire notification history table because there's no index available on just `created_at`.

```
rdsbroker_6093d337_6918_4b97_9709_97529114eb90> explain select min(created_at) from notification_history;
+----------------------------------------------------------------------------------------------------------------+
| QUERY PLAN                                                                                                     |
|----------------------------------------------------------------------------------------------------------------|
| Finalize Aggregate  (cost=244986027.97..244986027.98 rows=1 width=8)                                           |
|   ->  Gather  (cost=244986027.75..244986027.96 rows=2 width=8)                                                 |
|         Workers Planned: 2                                                                                     |
|         ->  Partial Aggregate  (cost=244985027.75..244985027.76 rows=1 width=8)                                |
|               ->  Parallel Seq Scan on notification_history  (cost=0.00..237599580.80 rows=2954178780 width=8) |
+----------------------------------------------------------------------------------------------------------------+
EXPLAIN 5
Time: 0.061s
```

Given that, maybe just doing the delete query is fine? Otherwise we're scanning the table twice.

Alternatively let's put an index on `created_at`. I would have found one very useful for separate reasons over the last week.

---

But obviously we're doing this query to get the date of the oldest record so that we can work out what quarter of data to delete. An alternative to this would be to hard-code a start date (as you have below with 2019-10-01) and store/retrieve that date from redis in each run, adding 3 months each time. If the key isn't present, use 2019-10-01, otherwise whatever the value is (so that the first run works, but also so that if we lose redis, we recover - given a week or so).

Your call as to whether this optimisation is worth saving the double query. I think it could be, unless scanning notification history doesn't actually take as long as I fear it might.

app/config.py
```diff
@@ -262,6 +262,11 @@ class Config(object):
                 "options": {"queue": QueueNames.PERIODIC},
             },
             # app/celery/nightly_tasks.py
+            "delete-oldest-quarter-of-unneeded-notification-history": {
+                "task": "delete-oldest-quarter-of-unneeded-notification-history",
+                "schedule": crontab(hour=21),
+                "options": {"queue": QueueNames.PERIODIC},
+            },
```
samuelhwilliams, line 269: An alternative if we're worried about workers being rolled, or disabling this when it becomes less relevant, would be to schedule this as a task via concourse. In PaaS we can run via `cf run-task` and in ECS through a similar mechanism.

Then we can just pause the pipeline when it no longer needs to run nightly, and turn it back on when we want it to start kicking again. Not that we should really need to disable it, I think.

app/celery/nightly_tasks.py
samuelhwilliams: Possibly logging the time in UTC would be more useful, since that's what we would see in the DB and generally what we deal with behind the scenes, I think?

CrystalPea: I agree 👍🏼

app/celery/nightly_tasks.py
samuelhwilliams:
```
rdsbroker_6093d337_6918_4b97_9709_97529114eb90> explain select min(created_at) from notification_history;
+---------------------------------------------------------------------------------------------------------------------------->
| QUERY PLAN                                                                                                                 >
|---------------------------------------------------------------------------------------------------------------------------->
| Result  (cost=0.61..0.62 rows=1 width=8)                                                                                   >
|   InitPlan 1 (returns $0)                                                                                                  >
|     ->  Limit  (cost=0.58..0.61 rows=1 width=8)                                                                            >
|           ->  Index Only Scan using ix_notification_history_created_at on notification_history  (cost=0.58..204070599.34 ro>
|                 Index Cond: (created_at IS NOT NULL)                                                                       >
+---------------------------------------------------------------------------------------------------------------------------->
EXPLAIN 5
Time: 0.046s
rdsbroker_6093d337_6918_4b97_9709_97529114eb90> select min(created_at) from notification_history;
+----------------------------+
| min                        |
|----------------------------|
| 2019-06-20 00:00:06.372805 |
+----------------------------+
SELECT 1
Time: 0.039s
```

🎉

app/celery/nightly_tasks.py
idavidmcdonald: Will try and follow up with a subsequent PR to add in that logging...

I agree it is useful but this PR has taken me soo long to find time to merge that I'm going to merge as is (otherwise it won't happen for another few days)...

PR #3864, opened 2023-08-31
Support bilingual templates

CrystalPea:
## Work done:
1. Add new columns to Templates and TemplatesHistory tables:
- welsh_content
- welsh_subject
- languages
2. Add new enum for language options. Available options for now are `english` and `welsh_then_english`
3. Add new fields to schemas and to a check if template has anything to update


migrations/versions/0420_bilingual_templates.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("templates", sa.Column("welsh_content", sa.Text(), nullable=True))
+    op.add_column("templates", sa.Column("welsh_subject", sa.Text(), nullable=True))
+    template_language_options = postgresql.ENUM(
+        "english", "welsh_then_english", name="template_language_options", create_type=False
+    )
+    template_language_options.create(op.get_bind(), checkfirst=True)
```
samuelhwilliams, line 24: Ah nice, I wasn't super aware of the `create_type=False` and `checkfirst=True`. I assume that makes local testing a bit easier if you have to re-run the migrations a few times. We wouldn't expect to ever need those options in prod, right? (but they're fine to stay IMO)

app/constants.py
```diff
@@ -127,6 +127,12 @@
 NOTIFICATION_TYPES = [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]  # not broadcast
 NOTIFICATION_TYPE = [EMAIL_TYPE, SMS_TYPE, LETTER_TYPE]  # duplicate that can probably be cleaned up
 
+# Language options supported for bilingual letter templates
+ENGLISH_LANGUAGE = "english"
+WELSH_THEN_ENGLISH_LANGUAGE = "welsh_then_english"
+
+TEMPLATE_LANGUAGE_OPTIONS = [ENGLISH_LANGUAGE, WELSH_THEN_ENGLISH_LANGUAGE]
```
samuelhwilliams, line 134: question/optional suggestion: What do you think about using an actual enum? See `constants.OrganisationUserPermissionTypes` for an example. Although python enums have a bit of extra faff (having to use `eg LanguageOptions.ENGLISH.value` to get the actual string, it may be a bit clearer in terms of intent, and the intellisense support can be more helpful.

app/models.py
```diff
@@ -1001,6 +1003,16 @@ def __init__(self, **kwargs):
     postage = db.Column(db.String, nullable=True)
     broadcast_data = db.Column(JSONB(none_as_null=True), nullable=True)
 
+    welsh_content = db.Column(db.Text)
+    welsh_subject = db.Column(db.Text)
+    languages = db.Column(
```
samuelhwilliams, line 1008: question: do you think we should be more specific in naming these, eg `welsh_letter_content`, to make it readily apparent that these only apply to letters? 🤔 

optional suggestion:
```suggestion
    welsh_letter_content = db.Column(db.Text)
    welsh_letter_subject = db.Column(db.Text)
    letter_languages = db.Column(
```

samuelhwilliams (commented):
I think this looks good on the whole. A few comments for thought, but I don't see anything too contentious :) Nice.

app/models.py
CrystalPea: Ooh good idea, I will change this throughout.

app/constants.py
CrystalPea: I don't use intellisense much, but if it makes life easier for others, it's worth considering. I like the class syntax in it. `LanguageOptions.ENGLISH.value` does sound a bit faffy, I wonder why they implemented it like that 😅 .

Most of all I think having both types of syntax at once is non-consistent. So it would be good to discuss this on the tech catch up today, and decide which way we want to go, going forward.

migrations/versions/0420_bilingual_templates.py
CrystalPea: You used `create_type=False` in migration 0413, which was part of my inspiration 😁 . Other part was stackoverflow, as the auto-created migration file didn't work. Check my commit for this code for details ☺️

samuelhwilliams: Oh 😂  haha! Well that is proof that my memory is shot then ... 😬

samuelhwilliams (approved):
Nice work and good for persevering with the enum stuff. Once we got it typed correctly a lot of the faff does seem to have gone away :)

PR #3868, opened 2023-09-07
migrate old data and start using normalised_service_name

leohemsted:
1. ~~add new column as nullable~~ (https://github.com/alphagov/notifications-api/pull/3865)
2. ~~write new values to both new and old column~~ (https://github.com/alphagov/notifications-admin/pull/4824)
3. ~~migrate old data. add unique and not null constraints.~~ (https://github.com/alphagov/notifications-api/pull/3881)
4. **use new column**
5. delete references to old column in code
6. drop old column

the march onwards continues - now that the admin is sending back _both_ values, we're okay to:

* migrate old data.
* remove the constraint on the old column and add it to the new column
* update the code to expect checks on the new column instead


app/service/rest.py
```diff
@@ -167,14 +167,13 @@ def handle_integrity_error(exc):
         'duplicate key value violates unique constraint "{}"'.format(constraint) in str(exc)
         for constraint in {"services_name_key", "services_email_from_key", "services_normalised_service_name_key"}
     ):
+        dupe_name = (
```
CrystalPea, line 170: "dupe" also means to deceive or trick, is this not confusing?

CrystalPea (approved):
Looks good to me - just will need rebasing :D

app/service/rest.py
leohemsted: im pretty familiar with the use, especially given you've got the context from the constraint log line above, but happy to change it.

PR #3871, opened 2023-09-08
Revert "Bump gunicorn to 21.2.0"

samuelhwilliams:
This reverts the changes introduced by commit
6d80b03c577bb47cd09ac3e49a4b30284179832b.

In a recent load test we identified a huge drop-off in performance that has been traced back to this dependency bump. Until we can investigate further and work out exactly why - and potentially get it fixed up stream - we will go back to our previously-pinned version.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -9,7 +9,8 @@ Flask-Migrate==3.1.0
 flask-sqlalchemy==3.0.2
 Flask==2.3.2
 click-datetime==0.2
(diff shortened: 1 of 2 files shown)


idavidmcdonald (approved):
Looks good and thanks for a good code comment here. Useful.

I think you are missing one tiny bit of the code comment which is `21.y,z` where I'm guessing `y,z` should be replaced before merging?

PR #3873, opened 2023-09-12
update InvalidRequest error message for reply-to email address

BlessedDev:
This PR is for [this](https://trello.com/c/3JRlqM43/496-update-reply-to-email-address-already-in-use-validation-error-message) ticket.

It updates the InvalidRequest error message for reply-to email address validation.

### DIFF
--- a/tests/app/service/test_rest.py
+++ b/tests/app/service/test_rest.py
@@ -2732,7 +2732,7 @@ def test_verify_reply_to_email_address_doesnt_allow_duplicates(admin_request, no
     response = admin_request.post(
         "service.verify_reply_to_email_address", service_id=service.id, _data=data, _expected_status=409
     )
-    assert response["message"] == "Your service already uses ‘reply-here@example.gov.uk’ as an email reply-to address."
+    assert response["message"] == "‘reply-here@example.gov.uk’ is already a reply-to email address for this service."
 
 
 def test_add_service_reply_to_email_address(admin_request, sample_service):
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):
This change does what we need 👍

However in testing it I've noticed that the admin interface doesn't handle this error correctly, so we should raise a PR on admin to handle it and display the error properly. At the moment it just shows our default 500 error page.

tests/app/service/test_rest.py
```diff
@@ -2732,7 +2732,7 @@ def test_verify_reply_to_email_address_doesnt_allow_duplicates(admin_request, no
     response = admin_request.post(
         "service.verify_reply_to_email_address", service_id=service.id, _data=data, _expected_status=409
     )
-    assert response["message"] == "Your service already uses ‘reply-here@example.gov.uk’ as an email reply-to address."
+    assert response["message"] == "‘reply-here@example.gov.uk’ is already a reply-to address for this service."
```
karlchillmaid, line 2735:
```suggestion
    assert response["message"] == "‘reply-here@example.gov.uk’ is already a reply-to email address for this service."
```

tests/app/service/test_rest.py
```diff
@@ -2753,7 +2753,7 @@ def test_add_service_reply_to_email_address_doesnt_allow_duplicates(admin_reques
     response = admin_request.post(
         "service.add_service_reply_to_email_address", service_id=service.id, _data=data, _expected_status=409
     )
-    assert response["message"] == "Your service already uses ‘reply-here@example.gov.uk’ as an email reply-to address."
+    assert response["message"] == "‘reply-here@example.gov.uk’ is already a reply-to address for this service."
```
karlchillmaid, line 2756:
```suggestion
    assert response["message"] == "‘reply-here@example.gov.uk’ is already a reply-to email address for this service."
```

karlchillmaid (changes requested):
I left a couple of small edits – but otherwise the content looks 👍

tests/app/service/test_rest.py
BlessedDev: I have implemented the changes you requested.

samuelhwilliams (approved):

PR #3874, opened 2023-09-13
reduce concurrency of dvla api worker

leohemsted:
when sending 100k letters with 2 instances of 4 concurrency, we saw 177k requests (77k requests hit the rate limit).

reducing the number of processes from 8 to 6 should help manage this.

if we continue having issues we may want to look at a more sophisticated way of throttling

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -4,6 +4,8 @@
 if [ "$1" == "api-worker-periodic" ] || [ "$1" == "api-worker-broadcasts" ]
 then
   CONCURRENCY=2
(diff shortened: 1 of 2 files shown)


idavidmcdonald (commented):
Could you please make an equivalent change to the `entrypoint.sh` file please, which is the equivalent set up for ECS as otherwise this might get left and forgotten about and not copied over...

Apart from that, the change makes sense. Whether it will be enough is a different question...

CrystalPea (approved):

idavidmcdonald (approved):

PR #3875, opened 2023-09-13
fix syntax in entrypoint

leohemsted:


idavidmcdonald (approved):
Seems logical, lets see if it works!

PR #3876, opened 2023-09-13
Fix GET notifications by letter status

samuelhwilliams:
We tried to make it possible to filter notifications by some letter statuses (accepted, received) back in 2017, but the implementation was broken so our API actually rejected those statuses at the validation stage. We had "accepted, received" as a single status type rather than two separate "accepted" and "received" statuses.

This patch splits out the two statuses correctly, and refactors `Notification.substitute_status` to fix a related bug and simplify the implementation.

We were returning a mix of types: a string in some cases and a list of strings in others. We now consistently return a list of strings, which is what usage of the method expects anyway, as we pass the return value into a SQLAlchemy `in` clause, which will iterate over the return value. If it's a string, we end up checking if a notification status is in (eg) 'delivered', which checks against 'd,' 'e', 'l', ... We should instead check against ['delivered'].


tests/app/v2/notifications/test_get_notifications.py
```diff
@@ -339,6 +339,53 @@ def test_get_all_notifications_filter_by_single_status(api_client_request, sampl
...
+            True,
+            marks=pytest.mark.xfail(
+                reason=(
+                    "When we serialize a notification with status `returned-letter`, we display it as `received` "
+                    "instead. But we don't currently do the inverse when filtering on `received`. We probably should, "
+                    "but it would be a backwards-incompatible change and so may need flagging with API users before "
+                    "we fix this."
+                )
+            ),
+        ),
```
samuelhwilliams, line 360: Open question as to whether we just fix this, or want to notify API users as it would mean that anyone filtering on eg `notification_type=letter&status=received` is currently getting back only letters in the `delivered `status`, but would also get `returned-letter` if we "fixed" this.

Note also that we have an inconsistency around `returned-letter` now. We serialize notifications in that state as `received`, but on GET notifications they show up in filters for `status=failed` (but showing `status=received` in the response) 🙃

tests/app/v2/notifications/test_get_notifications.py
```diff
@@ -339,6 +339,51 @@ def test_get_all_notifications_filter_by_single_status(api_client_request, sampl
     assert json_response["notifications"][0]["status"] == "pending"
 
 
+@pytest.mark.parametrize(
+    "internal_status, filter_status, should_get",
```
CrystalPea, line 343: This `should_get` variable has a pretty confusing name, and I struggled to find a better one, and I thought - does this mean maybe we could approach this parametric in another way?

For example if you changed this param to `expected_num_of_notifications` instead, it would be either 0 or 1, and you could use it for your first assertion, too, and it would be easier to read.

And then the test would be also checking that there are no false positives ☺️ .

CrystalPea (commented):
Hi Sam, well done, lots of this is simpler than it was.

I think we need discussion with a team before we change what people get when they poll for `received` status though.

samuelhwilliams:
> I think we need discussion with a team before we change what people get when they poll for received status though.

I agree, which is why I've only flagged this and haven't changed the behaviour yet :)

app/models.py
```diff
@@ -1431,55 +1431,29 @@ def completed_at(self):
...
 
-        def _substitute_status_seq(_statuses):
-            return list(set(itertools.chain.from_iterable(_substitute_status_str(status) for status in _statuses)))
+            return [_status]
 
-        if isinstance(status_or_statuses, str):
-            return _substitute_status_str(status_or_statuses)
-        return _substitute_status_seq(status_or_statuses)
+        unique_substituted_statuses = {
+            substitute for status in status_or_statuses for substitute in _substitute_status(status)
```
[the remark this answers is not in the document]
samuelhwilliams: ah, it's actually a set comprehension, which may be some of the confusion. It is equivalent to this:

```
unique_substituted_statuses = set()
for status in status_or_statuses:
    for substitute in _substitute_status(status):
        unique_substituted_statuses.add(substitute)
```

app/v2/notifications/notification_schemas.py
```diff
@@ -85,7 +85,7 @@
             "type": "array",
             "items": {
                 "enum": NOTIFICATION_STATUS_TYPES
-                + [NOTIFICATION_STATUS_LETTER_ACCEPTED + ", " + NOTIFICATION_STATUS_LETTER_RECEIVED]
```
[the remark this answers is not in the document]
samuelhwilliams: Yeah ... no idea 😅

tests/app/v2/notifications/test_get_notifications.py
BlessedDev: I can see both sides of this issue. I however agree with @samuelhwilliams that the unique nature of a `returned-letter` as compared to undelivered emails or text messages, doesn't  wholly fit with `permanent-failure` and is probably best served by having a `returned` status.

BlessedDev:
This simplifies `Notification.substitute_status` and splitting `accepted` and `received` is a very positive move.
I don't have context regarding the question of reconciling the current inconsistencies mentioned.

PR #3877, opened 2023-09-13
Dvla error handling

leohemsted:
Couple of things that became apparent from load test:

* we don't catch connection errors or timeouts currently, they just sent a letter into temp failure
* we didn't have any code blocks for sending a letter that isn't created, or isn't a normal type. technically the collate task won't gather them, but thought it wise to put some extra checks in the deliver_letter task

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -16,7 +16,9 @@ from app.clients.letter.dvla import (
 from app.clients.sms import SmsClientResponseException
 from app.config import QueueNames
 from app.constants import (
+    KEY_TYPE_NORMAL,
     LETTER_TYPE,
+    NOTIFICATION_CREATED,
     NOTIFICATION_SENDING,
     NOTIFICATION_TECHNICAL_FAILURE,
 )
@@ -93,6 +95,18 @@ def deliver_letter(self, notification_id):
     current_app.logger.info("Start sending letter for notification id: %s", notification_id)
     notification = notifications_dao.get_notification_by_id(notification_id, _raise=True)
     postal_address = PostalAddress(notification.to, allow_international_letters=True)
+
+    if notification.status != NOTIFICATION_CREATED:
+        current_app.logger.warning(
+            "deliver_letter task called for notification %s in status %s", notification_id, notification.status
+        )
+        return
+
+    if notification.key_type != KEY_TYPE_NORMAL:
+        current_app.logger.error(
+            "deliver_letter task called for notification %s with key type %s", notification_id, notification.key_type
+        )
+        return
 
     try:
         file_bytes = find_letter_pdf_in_s3(notification).get()["Body"].read()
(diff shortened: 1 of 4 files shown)


app/clients/letter/dvla.py
```diff
@@ -125,19 +144,20 @@ def jwt_token(self):
...
+
+        def _handle_401(e: requests.HTTPError):
+            if e.response.status_code == 401:
+                # likely the old password has already expired
+                current_app.logger.exception("Failed to generate a DVLA jwt token")
+
+                self.dvla_password.clear()
+                raise DvlaRetryableException(e.response.json()[0]["detail"]) from e
+
+        with _handle_common_dvla_errors(custom_httperror_exc_handler=_handle_401):
```
leohemsted, line 160: interested to hear how people feel about this. is it obvious and understandable what this does and how we'd update/write another block in the future?

app/celery/provider_tasks.py
```diff
@@ -94,6 +96,18 @@ def deliver_letter(self, notification_id):
...
     postal_address = PostalAddress(notification.to, allow_international_letters=True)
 
+    if notification.status != NOTIFICATION_CREATED:
+        current_app.logger.warning(
+            "deliver_letter task called for notification %s in status %s", notification_id, notification.status
+        )
+        return
+
+    if notification.key_type != KEY_TYPE_NORMAL:
+        current_app.logger.warning(
```
samuelhwilliams, line 106: I might be tempted to log an error on this, so that it gets flagged in sentry and actually becomes visible. Otherwise is there a risk that we just emit warnings every day for the same notifications because their state doesn't change?

app/clients/letter/dvla.py
samuelhwilliams: Feels reasonable to me 👍  I think if I'd been doing it I'd have ended up adding a `post` method to the class (which takes an optional error handler) and centralising the logic there rather than using a context manager. Would remove a level of indentation, and mean that when adding future methods/calls to the client, we don't need to remember to add the context manager as all post requests would get the error handling natively.

But I don't think it is problematic enough to actually change what you've done.

Important part is that what you've got does what we need, so nice 👍

app/clients/letter/dvla.py
```diff
@@ -45,6 +47,23 @@ class DvlaThrottlingException(DvlaRetryableException):
...
+        raise DvlaRetryableException() from e
+    except requests.HTTPError as e:
+        custom_httperror_exc_handler(e)
+
+        if e.response.status_code == 429:
+            raise DvlaThrottlingException() from e
+        elif e.response.status_code >= 500:
+            raise DvlaRetryableException() from e
+        else:
+            raise DvlaNonRetryableException() from e
```
samuelhwilliams, line 64: optional: maybe we can add some context to each of these exceptions just to make them a bit easier to instantly read? I reckon in Sentry we could get the context from `e` and understand it, but having it surfaced immediately might still be nice. Something like (untested...)

```suggestion
            raise DvlaThrottlingException() from e
        elif e.response.status_code >= 500:
            raise DvlaRetryableException(f"Received {e.response.status_code} from {e.request.url}`) from e
        else:
            raise DvlaNonRetryableException(f"Received {e.response.status_code} from {e.request.url}`) from e
```

samuelhwilliams (approved):

app/celery/provider_tasks.py
leohemsted: added to this - not sure if it was obvious to you but i haven't changed  the status == created check since we'd expect that a letter not in created wouldnt get picked up again next time

PR #3879, opened 2023-09-14
Rename ‘Notify service user’ to ‘GOV.UK Notify’

quis:
So that if any actions are ever attributed to this user it’s clearer that they are being done by the platform itself, not some kind of generic/anonymous user.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0422_letter_languages_constraint
+0423_notify_user_name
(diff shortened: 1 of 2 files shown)


quis:
Nothing in the functional tests is relying on this user being called _Notify service user_: https://github.com/search?q=repo%3Aalphagov%2Fnotifications-functional-tests+%22notify+service+user%22&type=code

leohemsted (approved):

PR #3880, opened 2023-09-15
Add index concurrently on created_at to notification_history table

idavidmcdonald:
We want to add this index because at the moment it the table is too big to query based on created_at (we do have some indexes on this table already but those are multi column indexes so are not often able to be used when querying just on created_at).

We choose to create our index concurrently. This is backed by https://medium.com/paypal-tech/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#04b4.

Importantly it says:
This lower lock level allows reads and writes to continue against the table while the index is built.

You can also read the postgres documentation on it at: https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY

Creating an index concurrently needs to be done outside of a transaction. We can use an alembic feature to support the migration being done outside of a transaction:
https://alembic.sqlalchemy.org/en/latest/api/runtime.html#alembic.runtime.migration.MigrationContext.autocommit_block

Because it is done outside of a transaction, the sql should run very quickly and the migration should suceed, leaving the index to be built concurrently happening behind the scenes.

I will need to manually check that
the index is eventually created successfully. If it fails, then it will need to be removed manually, as per the postgres docs.

I'm not too worried about performance by adding this index. We don't care too much about insertion speed for the notification_history table, which an index might affect. It should however speed up our ability to query notification_history for particular queries.

## How to review
Read the above for context, check out the code and then run

```
source environment.sh
flask db upgrade
flask db downgrade
```
and check for no errors

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1670,6 +1670,7 @@ class NotificationHistory(db.Model, HistoryModel):
         Index(
             "ix_notification_history_service_id_composite", "service_id", "key_type", "notification_type", "created_at"
         ),
+        Index("ix_notification_history_created_at", "created_at", postgresql_concurrently=True),
     )
 
     @classmethod
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0423_notify_user_name
+0424_n_history_created_at
--- a/migrations/versions/0424_n_history_created_at.py
+++ b/migrations/versions/0424_n_history_created_at.py
@@ -1,0 +1,31 @@
+"""
+
+Revision ID: 0424_n_history_created_at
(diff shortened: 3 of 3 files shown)


samuelhwilliams (approved):

c4seyla (approved):

samuelhwilliams:
Grateful for adding this index 👍  There have been a few cases lately where this would have been useful. One example was auditing (recent) historical notifications sent with templates/template version when chasing down a rare [template version desync bug](https://github.com/alphagov/notifications-utils/pull/1064).

But this will more generally make it easier to run analysis queries on subsets of notification history (eg the last 3 months) in a much more performant way.

PR #3881, opened 2023-09-19
normalised_service_name migration

leohemsted:
Finally getting back at this!

Migrate the old email_from to new normalised_service_name column. 

1. ~~add new column as nullable~~ (https://github.com/alphagov/notifications-api/pull/3865)
2. ~~write new values to both new and old column~~ (https://github.com/alphagov/notifications-admin/pull/4824)
3. **migrate old data. add unique and not null constraints.**
4. delete references to old column in code
5. drop old column

### DIFF
--- a/migrations/versions/0425_unique_service_name_2.py
+++ b/migrations/versions/0425_unique_service_name_2.py
@@ -1,0 +1,37 @@
+"""
+
+Revision ID: 0425_unique_service_name_2
+Revises: 0424_n_history_created_at
+Create Date: 2023-09-01 10:10:19.011060
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "0425_unique_service_name_2"
+down_revision = "0424_n_history_created_at"
+
+
+def upgrade():
+    op.execute("UPDATE services SET normalised_service_name = email_from WHERE normalised_service_name IS NULL")
+    op.execute("UPDATE services_history SET normalised_service_name = email_from WHERE normalised_service_name IS NULL")
+    op.alter_column("services", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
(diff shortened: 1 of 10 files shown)


migrations/versions/0425_unique_service_name_2.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def upgrade():
+    op.execute("UPDATE services SET normalised_service_name = email_from WHERE normalised_service_name IS NULL")
+    op.execute("UPDATE services_history SET normalised_service_name = email_from WHERE normalised_service_name IS NULL")
+    op.alter_column("services", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("services_history", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("services", "email_from", existing_type=sa.VARCHAR(), nullable=True)
+    op.alter_column("services_history", "email_from", existing_type=sa.VARCHAR(), nullable=True)
+    op.drop_constraint("services_email_from_key", "services", type_="unique")
```
whpearson, line 23: This is a bit thorny. We may want to keep this constraint for now. if we drop it, we might get into a situation where we can't downgrade (if the email_from becomes duplicated in some way).

leohemsted: we need to drop this here as we're going to stop updating this value from the rest of the front end as part of our renaming step.

given that normalised_service_name already has a unique_constraint, by setting `email_from` to be the value from normalised_service_name in the downgrade script we should be covered i believe.

i might remove the where clause from the downgrade just to be safe.

whpearson: Okay that makes sense. Looks good

whpearson (approved):

PR #3882, opened 2023-09-19
Resolve an n+1 queries on endpoint

samuelhwilliams:
This relationship was set up to perform a SQL query for every template folder individually. By setting the relationship to load via a join we can instead emit a single SQL statement.

https://govuk-notify-gds.sentry.io/issues/4474807382/?project=4504949325824000


whpearson (approved):
Change makes sense and I've run the affected tests locally with app.config.Test =  True and they pass.

PR #3883, opened 2023-09-20
Switch the db migration task to use cflinuxfs4

corlettb:
What
----

Switch the db migration task to use cflinuxfs4

Why
----

cflinuxfs3 is currently the default on the PaaS. This is based off ubuntu 18.04. This stopped receiving updates after Apr 2023.

cflinuxfs4 will become the default on the PaaS after 27 Nov 2023.

The intention is to migrate all applications to cflinuxfs4. This change is the start of that process, where the db migration app is our canary.

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -47,6 +47,7 @@
       'staging': 0,
       'production': 0
     },
+    'stack': 'cflinuxfs4'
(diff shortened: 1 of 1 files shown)


sakisv (approved):
👍

PR #3886, opened 2023-09-20
Apply localised fix for n+1 queries

samuelhwilliams:
I tried to fix an n+1 queries on the `get_template_folders_for_service` endpoint with a generalised change to how the relationship on `TemplateFolder` joins to `User` (in https://github.com/alphagov/notifications-api/pull/3885), but this cascaded to cause massive queries on other endpoints when services have thousands of templates/users.

Let's fix this endpoint more specifically by loading the relationships we know we are going to use, in advance.

### DIFF
--- a/app/template_folder/rest.py
+++ b/app/template_folder/rest.py
@@ -1,5 +1,6 @@
 from flask import Blueprint, current_app, jsonify, request
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm import raiseload, selectinload
 from sqlalchemy.orm.exc import NoResultFound
 
 from app.dao.dao_utils import autocommit
@@ -7,7 +8,6 @@ from app.dao.service_user_dao import (
     dao_get_active_service_users,
     dao_get_service_user,
 )
-from app.dao.services_dao import dao_fetch_service_by_id
 from app.dao.template_folder_dao import (
     dao_create_template_folder,
     dao_delete_template_folder,
@@ -16,7 +16,7 @@ from app.dao.template_folder_dao import (
 )
 from app.dao.templates_dao import dao_get_template_by_id_and_service_id
 from app.errors import InvalidRequest, register_errors
-from app.models import TemplateFolder
+from app.models import Service, TemplateFolder
 from app.schema_validation import validate
 from app.template_folder.template_folder_schema import (
     post_create_template_folder_schema,
@@ -40,7 +40,11 @@ def handle_integrity_error(exc):
 
 @template_folder_blueprint.route("", methods=["GET"])
 def get_template_folders_for_service(service_id):
-    service = dao_fetch_service_by_id(service_id)
+    service = (
+        Service.query.filter_by(id=service_id)
+        .options(raiseload("users"), selectinload("all_template_folders").options(selectinload("users")))
+        .one()
+    )
 
     template_folders = [o.serialize() for o in service.all_template_folders]
     return jsonify(template_folders=template_folders)
--- a/tests/app/organisation/test_rest.py
+++ b/tests/app/organisation/test_rest.py
@@ -881,13 +881,17 @@ def test_get_organisation_services_usage(admin_request, notify_db_session, mocke
     assert response["updated_at"] == "2019-06-01T12:00:00+00:00"
 
 
-@pytest.mark.skip(
-    "Another test (`test_cbc_proxy_vodafone_send_link_test_invokes_function`) fails when we enable "
-    "SQLALCHEMY_RECORD_QUERIES, which is a requirement for this test. So we can't run this for now ... but "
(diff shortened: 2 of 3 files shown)


app/template_folder/rest.py
```diff
@@ -40,7 +40,11 @@ def handle_integrity_error(exc):
 
 @template_folder_blueprint.route("", methods=["GET"])
 def get_template_folders_for_service(service_id):
-    service = dao_fetch_service_by_id(service_id)
+    service = (
+        Service.query.filter_by(id=service_id)
+        .options(raiseload("users"), selectinload("all_template_folders").options(selectinload("users")))
```
leohemsted, line 45: i'm not familiar with these options - what do they do?

samuelhwilliams: `raiseload` basically will throw an error if it tries to access `users` - a way of saying we don't expect to join on this and it should error early instead of hitting the DB if it tries to. This endpoint doesn't access users, so it's a safety precaution. Not strictly necessary, and we could drop it if we want; would then default to lazy loading.

`selectinload` collects the related instances by using a `SELECT ... IN (pk1, pk2, pk3, ...)` - so will gather all related objects via a single additional query. It avoids doing a deep join which means we don't end up duplicating a load  service data via joined rows between service, template folder, and user permissions.

In the test here we end up with the following queries:

## Select service
```
SELECT services.id AS services_id, services.name AS services_name, services.normalised_service_name AS services_normalised_service_name, services.created_at AS services_created_at, services.updated_at AS services_updated_at, services.active AS services_active, services.letter_message_limit AS services_letter_message_limit, services.sms_message_limit AS services_sms_message_limit, services.email_message_limit AS services_email_message_limit, services.restricted AS services_restricted, services.email_from AS services_email_from, services.created_by_id AS services_created_by_id, services.prefix_sms AS services_prefix_sms, services.organisation_type AS services_organisation_type, services.crown AS services_crown, services.rate_limit AS services_rate_limit, services.contact_link AS services_contact_link, services.volume_sms AS services_volume_sms, services.volume_email AS services_volume_email, services.volume_letter AS services_volume_letter, services.consent_to_research AS services_consent_to_research, services.count_as_live AS services_count_as_live, services.go_live_user_id AS services_go_live_user_id, services.go_live_at AS services_go_live_at, services.has_active_go_live_request AS services_has_active_go_live_request, services.organisation_id AS services_organisation_id, services.notes AS services_notes, services.purchase_order_number AS services_purchase_order_number, services.billing_contact_names AS services_billing_contact_names, services.billing_contact_email_addresses AS services_billing_contact_email_addresses, services.billing_reference AS services_billing_reference, services.version AS services_version
FROM services
WHERE services.id = %(id_1)s
```

## Select template_folder
```
SELECT template_folder.service_id AS template_folder_service_id, template_folder.id AS template_folder_id, template_folder.name AS template_folder_name, template_folder.parent_id AS template_folder_parent_id
FROM template_folder
WHERE template_folder.service_id IN (%(primary_keys_1)s)
```

## Select template_folder user IDs
```
SELECT template_folder_1.id AS template_folder_1_id, user_to_service.user_id AS user_to_service_user_id, user_to_service.service_id AS user_to_service_service_id
FROM template_folder AS template_folder_1 JOIN user_folder_permissions AS user_folder_permissions_1 ON template_folder_1.id = user_folder_permissions_1.template_folder_id JOIN user_to_service ON user_to_service.user_id = user_folder_permissions_1.user_id AND user_to_service.service_id = user_folder_permissions_1.service_id
WHERE template_folder_1.id IN (%(primary_keys_1)s, %(primary_keys_2)s, %(primary_keys_3)s, %(primary_keys_4)s, %(primary_keys_5)s, %(primary_keys_6)s, %(primary_keys_7)s, %(primary_keys_8)s, %(primary_keys_9)s, %(primary_keys_10)s, %(primary_keys_11)s, %(primary_keys_12)s, %(primary_keys_13)s, %(primary_keys_14)s, %(primary_keys_15)s, %(primary_keys_16)s, %(primary_keys_17)s, %(primary_keys_18)s, %(primary_keys_19)s, %(primary_keys_20)s, %(primary_keys_21)s, %(primary_keys_22)s, %(primary_keys_23)s, %(primary_keys_24)s, %(primary_keys_25)s)
```

whpearson (approved):
Tests run fine locally with `SQLALCHEMY_RECORD_QUERIES = True`.

Changes makes sense. Good work on keeping the change smaller

samuelhwilliams:
Did a manual deploy of this to staging to test with a high impact service (bce47ec8-4041-43d4-8a43-5fcc5e6d61f8).


## Example - before
image

## Example - after
image

---

Not a _huge_ speedup in time, but it does remove the n+1 queries. I think a lot of the unmeasured time in these graphs are spent serialising/deserialising and rendering the template 🙃

PR #3890, opened 2023-09-21
run upgrades in separate transactions to remove deadlock

leohemsted:
we ran into deadlocks when running this script on production. adding a constraint acquires an access-exclusive lock. This is fine on its own, to add a not-null constraint is a relatively quick operation for a table of this size (27k services, 100k services_history). however, because the query was trying to acquire locks to both services and services_history, it ran in to a deadlock with another process where that process was also trying to acquire locks on both tables - each process acquired one of the locks, and was stuck waiting for the other to be released, so a deadlock was identified and the migration failed.

even if the other process wasn't doing anything big and dramatic, nothing can acquire a lock alongside an access-exclusive lock.

normally alembic wraps the entire upgrade script in a transaction (so that if there's an error, it can rollback safely to the previous version and ensure that the db state is consistent). but we can manually enter an autocommit block such that every statement commits separately outside of a transaction.

this has already been applied succesfully without incident to preview and staging so there's no action to do there.

we've mucked around in the past with multiple separate PRs and applying constraints and validating them separately, I don't think that's necessary here - https://github.com/alphagov/notifications-api/pull/2113

### DIFF
--- a/migrations/versions/0425_unique_service_name_2.py
+++ b/migrations/versions/0425_unique_service_name_2.py
@@ -14,24 +14,35 @@ down_revision = "0424_n_history_created_at"
 
 
 def upgrade():
-    op.execute("UPDATE services SET normalised_service_name = email_from WHERE normalised_service_name IS NULL")
-    op.execute("UPDATE services_history SET normalised_service_name = email_from WHERE normalised_service_name IS NULL")
-    op.alter_column("services", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
-    op.alter_column("services_history", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
-    op.alter_column("services", "email_from", existing_type=sa.VARCHAR(), nullable=True)
-    op.alter_column("services_history", "email_from", existing_type=sa.VARCHAR(), nullable=True)
-    op.drop_constraint("services_email_from_key", "services", type_="unique")
+    # run these commands with autocommit outside of a transaction - the main motivation for this is to
+    # ensure we don't try and acquire access-exclusive locks (which the alter column commands need) on both the
+    # services and services_history table at the same time, which can lead to deadlocks
+    with op.get_context().autocommit_block():
+        # update historical values
+        op.execute("UPDATE services SET normalised_service_name = email_from WHERE normalised_service_name IS NULL")
+        op.execute(
+            "UPDATE services_history SET normalised_service_name = email_from WHERE normalised_service_name IS NULL"
+        )
+
+        op.alter_column("services", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
+        op.alter_column("services_history", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
+
+        op.alter_column("services", "email_from", existing_type=sa.VARCHAR(), nullable=True)
+        op.alter_column("services_history", "email_from", existing_type=sa.VARCHAR(), nullable=True)
+
+        op.drop_constraint("services_email_from_key", "services", type_="unique")
 
 
 def downgrade():
-    op.alter_column("services_history", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=True)
-    op.alter_column("services", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=True)
+    with op.get_context().autocommit_block():
+        op.alter_column("services_history", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=True)
+        op.alter_column("services", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=True)
 
-    # back-fill any null email_from from normalised
-    op.execute("UPDATE services SET email_from = normalised_service_name")
-    op.execute("UPDATE services_history SET email_from = normalised_service_name")
+        # back-fill any null email_from from normalised
+        op.execute("UPDATE services SET email_from = normalised_service_name")
+        op.execute("UPDATE services_history SET email_from = normalised_service_name")
 
-    # now we can restore null constraint and unique index
-    op.alter_column("services", "email_from", existing_type=sa.VARCHAR(), nullable=True)
-    op.alter_column("services_history", "email_from", existing_type=sa.VARCHAR(), nullable=True)
-    op.create_unique_constraint("services_email_from_key", "services", columns=["email_from"])
+        # now we can restore null constraint and unique index
+        op.alter_column("services", "email_from", existing_type=sa.VARCHAR(), nullable=True)
+        op.alter_column("services_history", "email_from", existing_type=sa.VARCHAR(), nullable=True)
+        op.create_unique_constraint("services_email_from_key", "services", columns=["email_from"])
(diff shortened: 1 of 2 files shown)


BlessedDev (approved):

migrations/versions/0425_unique_service_name_2.py
```diff
@@ -14,24 +14,45 @@
...
+    # separate this from the service_history lock so that we reduce the chance of deadlocks
+    with op.get_context().autocommit_block():
+        op.alter_column("services", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
+        op.alter_column("services", "email_from", existing_type=sa.VARCHAR(), nullable=True)
+        op.drop_constraint("services_email_from_key", "services", type_="unique")
+
+    # this transaction needs an access-exclusive lock on service history table
+    with op.get_context().autocommit_block():
+        op.alter_column("services_history", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
+        op.alter_column("services_history", "email_from", existing_type=sa.VARCHAR(), nullable=True)
```
samuelhwilliams, line 34: I don't believe these will actually be treated as separate transactions FYI - `autocommit_block` puts the postgres connection into an autocommit mode, which means each individual statement is autocommitted. I don't think there's any difference from just having a single autocommit_block context manager around the whole lot.

```
sam:~/work/gds/notifications-api [fix-deadlock]$ flask db upgrade
2023-09-21T16:49:20 app app INFO no-request-id "Logging configured" [in /Users/sam/work/gds/notifications-utils/notifications_utils/logging.py:42]
INFO  [sqlalchemy.engine.Engine] select pg_catalog.version()
INFO  [sqlalchemy.engine.Engine] [raw sql] {}
INFO  [sqlalchemy.engine.Engine] select current_schema()
INFO  [sqlalchemy.engine.Engine] [raw sql] {}
INFO  [sqlalchemy.engine.Engine] show standard_conforming_strings
INFO  [sqlalchemy.engine.Engine] [raw sql] {}
INFO  [sqlalchemy.engine.Engine] BEGIN (implicit)
INFO  [sqlalchemy.engine.Engine] select relname from pg_class c join pg_namespace n on n.oid=c.relnamespace where pg_catalog.pg_table_is_visible(c.oid) and relname=%(name)s
INFO  [sqlalchemy.engine.Engine] [generated in 0.00006s] {'name': 'alembic_version'}
INFO  [sqlalchemy.engine.Engine] SELECT alembic_version.version_num
FROM alembic_version
INFO  [sqlalchemy.engine.Engine] [generated in 0.00007s] {}
INFO  [alembic.runtime.migration] Running upgrade 0424_n_history_created_at -> 0425_unique_service_name_2, Revision ID: 0425_unique_service_name_2
Revises: 0424_n_history_created_at
Create Date: 2023-09-01 10:10:19.011060
INFO  [sqlalchemy.engine.Engine] COMMIT
INFO  [sqlalchemy.engine.Engine] BEGIN (implicit; DBAPI should not BEGIN due to autocommit mode)
INFO  [sqlalchemy.engine.Engine] UPDATE services SET normalised_service_name = email_from WHERE normalised_service_name IS NULL
INFO  [sqlalchemy.engine.Engine] [generated in 0.00006s] {}
INFO  [sqlalchemy.engine.Engine] UPDATE services_history SET normalised_service_name = email_from WHERE normalised_service_name IS NULL
INFO  [sqlalchemy.engine.Engine] [generated in 0.00005s] {}
INFO  [sqlalchemy.engine.Engine] COMMIT using DBAPI connection.commit(), DBAPI should ignore due to autocommit mode
INFO  [sqlalchemy.engine.Engine] BEGIN (implicit)
INFO  [sqlalchemy.engine.Engine] COMMIT
INFO  [sqlalchemy.engine.Engine] BEGIN (implicit; DBAPI should not BEGIN due to autocommit mode)
INFO  [sqlalchemy.engine.Engine] ALTER TABLE services ALTER COLUMN normalised_service_name SET NOT NULL
INFO  [sqlalchemy.engine.Engine] [no key 0.00004s] {}
INFO  [sqlalchemy.engine.Engine] ALTER TABLE services ALTER COLUMN email_from DROP NOT NULL
INFO  [sqlalchemy.engine.Engine] [no key 0.00004s] {}
INFO  [sqlalchemy.engine.Engine] ALTER TABLE services DROP CONSTRAINT services_email_from_key
INFO  [sqlalchemy.engine.Engine] [no key 0.00004s] {}
INFO  [sqlalchemy.engine.Engine] COMMIT using DBAPI connection.commit(), DBAPI should ignore due to autocommit mode
INFO  [sqlalchemy.engine.Engine] BEGIN (implicit)
INFO  [sqlalchemy.engine.Engine] COMMIT
INFO  [sqlalchemy.engine.Engine] BEGIN (implicit; DBAPI should not BEGIN due to autocommit mode)
INFO  [sqlalchemy.engine.Engine] ALTER TABLE services_history ALTER COLUMN normalised_service_name SET NOT NULL
INFO  [sqlalchemy.engine.Engine] [no key 0.00003s] {}
INFO  [sqlalchemy.engine.Engine] ALTER TABLE services_history ALTER COLUMN email_from DROP NOT NULL
INFO  [sqlalchemy.engine.Engine] [no key 0.00003s] {}
INFO  [sqlalchemy.engine.Engine] COMMIT using DBAPI connection.commit(), DBAPI should ignore due to autocommit mode
INFO  [sqlalchemy.engine.Engine] BEGIN (implicit)
INFO  [sqlalchemy.engine.Engine] UPDATE alembic_version SET version_num='0425_unique_service_name_2' WHERE alembic_version.version_num = '0424_n_history_created_at'
INFO  [sqlalchemy.engine.Engine] [generated in 0.00005s] {}
INFO  [sqlalchemy.engine.Engine] COMMIT
```

see also https://alembic.sqlalchemy.org/en/latest/api/runtime.html#alembic.runtime.migration.MigrationContext.autocommit_block

my head currently doesn't want to tell me if this is going to be a problem or not .. but FYI

migrations/versions/0425_unique_service_name_2.py
```diff
@@ -14,24 +14,29 @@
...
 
 def upgrade():
-    op.execute("UPDATE services SET normalised_service_name = email_from WHERE normalised_service_name IS NULL")
-    op.execute("UPDATE services_history SET normalised_service_name = email_from WHERE normalised_service_name IS NULL")
-    op.alter_column("services", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
-    op.alter_column("services_history", "normalised_service_name", existing_type=sa.VARCHAR(), nullable=False)
-    op.alter_column("services", "email_from", existing_type=sa.VARCHAR(), nullable=True)
-    op.alter_column("services_history", "email_from", existing_type=sa.VARCHAR(), nullable=True)
-    op.drop_constraint("services_email_from_key", "services", type_="unique")
+    # run these commands with autocommit outside of a transaction - the main motivation for this is to
```
samuelhwilliams, line 17: is to WHAT I'm on tenterhooks here

leohemsted: oops

PR #3891, opened 2023-09-22
Allow clearing welsh letter subject and content

samuelhwilliams:
When a user changes a template from welsh then english back to just english, we should be able to clear the welsh letter data entirely by setting it to null. This updates the schema to allow that.

### DIFF
--- a/app/template/template_schemas.py
+++ b/app/template/template_schemas.py
@@ -40,8 +40,29 @@ post_update_template_schema = {
         "created_by": uuid,
         "archived": {"type": "boolean"},
         "current_user": uuid,
-        "letter_welsh_subject": {"type": "string"},
-        "letter_welsh_content": {"type": "string"},
         "letter_languages": {"type": "string", "enum": [i.value for i in LetterLanguageOptions]},
+        "letter_welsh_subject": {},
+        "letter_welsh_content": {},
+    },
+    "if": {
+        "properties": {"letter_languages": {"const": LetterLanguageOptions.english.value}},
+    },
+    "then": {
+        "properties": {
+            "letter_welsh_subject": {"type": "null"},
+            "letter_welsh_content": {"type": "null"},
+        }
+    },
+    "else": {
+        "if": {
+            "properties": {"letter_languages": {"const": LetterLanguageOptions.welsh_then_english.value}},
+        },
+        "then": {
+            "properties": {
+                "letter_welsh_subject": {"type": "string"},
+                "letter_welsh_content": {"type": "string"},
+            },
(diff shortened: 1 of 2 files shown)


app/template/template_schemas.py
```diff
@@ -40,8 +40,29 @@
...
+        "if": {
+            "properties": {"letter_languages": {"const": LetterLanguageOptions.welsh_then_english.value}},
+        },
+        "then": {
+            "properties": {
+                "letter_welsh_subject": {"type": "string"},
+                "letter_welsh_content": {"type": "string"},
+            },
+            "required": ["letter_welsh_subject", "letter_welsh_content"],
+        },
```
whpearson, line 66: Should we have an else statement for the case where letter_languages is not set?

samuelhwilliams: Hmm, we only have two options available in the `LetterLanguageOptions` enum, and those are both explicitly checked in the `if` statements so far. The only thing an `else` clause would process is if we add a new entry to `LetterLanguageOptions` but forget to update this. In that case I'm not sure what we should enforce through the schema since we don't know what that option would be.

What do you think we should add to an `else` statement? 🤔

CrystalPea: I think what Will means is that for non-letter templates, `letter_languages` field will be null.

samuelhwilliams: oh yeah 😆 I will add something ... 🤔

samuelhwilliams: I had a big of a play around, but I'm not convinced we want to do any further checks if `letter_languages` isn't set.

Unfortunately we do a lot of partial updates to templates and only pass across one or two attributes, so we can't do much in the way of cross-validation (eg only able to set letter languages if it's a letter template).

If we are setting `letter_languages`, then we make to make sure that welsh subject/content is also set or cleared appropriately.

But if we don't touch `letter_languages`, there are a lot of other things we could be doing:

* Updating a text or email - in which case we shouldn't be updating `letter_welsh_subject/content` - but we can't validate that here.
* Updating English letter subject/content
* Updating Welsh letter subject/content

Doing that validation correctly here is overly complex and restrictive, and probably means that we would have to start sending `letter_languages` all the time if we're touching English/Welsh subject/content. I think that's too likely to break our existing setup (or require us to change the admin side for little benefit), so I will leave it off for now.

whpearson: Makes sense. Thanks for thinking about it.

whpearson (approved):
Tests run locally fine. Changes make sense.

PR #3892, opened 2023-09-22
Enable the SQLAlchemy logger during migrations

samuelhwilliams:
This logs all of the actual SQL that runs on the DB and is useful to see. It can get a bit noisy, but since it's only during migrations that's fine - and probably more useful than not.


whpearson (approved):
Tested the change look good running locally.

PR #3903, opened 2023-10-02
lint alembic migrations with squawk

leohemsted:
squawk can take a sql file and lint it for common errors, such as renaming a column in an unsafe way or running a command that might acquire a wide-reaching long-lasting lock.

this commit wraps squawk in a little bash script that gets the alembic upgrades output as sql

there doesn't appear to be an obvious way to `#noqa` style "accept" a squawk warning if we know it's okay (for example, dropping a column we know is no longer written to or read from), so we can't immediately integrate this in to our CI pipeline or pre-commit hooks unfortunately.

typical output:

```
❯ ./scripts/lint_sql_migrations.sh
INFO  [alembic.runtime.migration] Generating static SQL
INFO  [alembic.runtime.migration] Running upgrade 0425_unique_service_name_2 -> 0426_drop_email_from, Revision ID: 0426_drop_email_from
Revises: 0425_unique_service_name_2
Create Date: 2023-09-29 18:47:48.703294
/tmp/upgrade.sql:2:2: warning: ban-drop-column

   2 | -- Running upgrade 0425_unique_service_name_2 -> 0426_drop_email_from
   3 |
   4 | ALTER TABLE services DROP COLUMN email_from;

  note: Dropping a column may break existing clients.

/tmp/upgrade.sql:6:2: warning: ban-drop-column

   6 | ALTER TABLE services_history DROP COLUMN email_from;

  note: Dropping a column may break existing clients.

find detailed examples and solutions for each rule at https://squawkhq.com/docs/rules
```

### DIFF
--- a/scripts/lint_sql_migrations.sh
+++ b/scripts/lint_sql_migrations.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+if ! [[ -x "$(command -v squawk)" ]]; then
+    echo "ERR: Squawk not found. Install with npm install -g squawk-cli" >&2
+    exit 1;
+fi
+
+
+# ensure main is up-to-date
+git fetch origin main:main
+
+FILES_COMMITTED_TO_BRANCH=$(git diff --name-only main -- migrations/versions/)
+UNADDED_FILES=$(git ls-files --others --exclude-standard migrations/versions)
+
+if [[ -z "${FILES_COMMITTED_TO_BRANCH}" && -z "${UNADDED_FILES}" ]]; then
+    # no edits to migrations, either added or not added
+    echo "No migrations"
+    exit 0;
+fi
+
+source environment.sh
+
+# what is the database currently pointing at on master?
+CURRENT_VERSION=$(git show main:migrations/.current-alembic-head)
+# what is the most recent version? note this depends on us only ever having one head
+NEWEST_VERSION=$(flask db heads | tail -n 1 | cut -d " " -f 1)
+
+if [[ "${CURRENT_VERSION}" == "${NEWEST_VERSION}" ]]; then
+    # there's changes to migrations, but no new version to test against?
+    # has someone modified an existing migration?
+    echo "Migrations found but can't work out what's different to master"
+    exit 0;
+fi
+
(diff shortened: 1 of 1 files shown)


leohemsted:
i wonder if integrating this in a way where it can just post a github comment and we can then review that and determine if it's appropriate is useful

leohemsted:
example squawk output script running against unsubscribe link

```
❯ scripts/lint_sql_migrations.sh
INFO  [alembic.runtime.migration] Generating static SQL
INFO  [alembic.runtime.migration] Running upgrade 0440_new_sms_allowance_n_rate -> 0441_add_unsubscribe_link, Revision ID: 0441_add_unsubscribe_link
Revises: 0440_new_sms_allowance_n_rate
Create Date: 2024-03-25 14:38:54.618674
/tmp/upgrade.sql:6:2: warning: constraint-missing-not-valid

   6 | ALTER TABLE notifications ADD CONSTRAINT ck_unsubscribe_link_is_null_if_notification_not_an_email CHECK (
   7 |         notification_type = 'email' OR unsubscribe_link is null
   8 |         );

  note: Requires a table scan to verify constraint and an ACCESS EXCLUSIVE lock which blocks reads.
  help: Add NOT VALID to the constraint in one transaction and then VALIDATE the constraint in a separate transaction.

find detailed examples and solutions for each rule at https://squawkhq.com/docs/rules```

File: scripts/lint_sql_migrations.sh [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+    # there's changes to migrations, but no new version to test against?
+    # has someone modified an existing migration?
+    echo "Migrations found but can't work out what's different to master"
+    exit 0;
+fi
+
+# tail -n +2 removes the first line of output - the stupid "logging configured" line
+flask db upgrade --sql $CURRENT_VERSION:$NEWEST_VERSION | tail -n +2 > /tmp/upgrade.sql
+
+squawk /tmp/upgrade.sql --pg-version="11.19"
```
whpearson, line 39: Should probably be --pg-version="15" I'm not sure the best way to keep this updated

leohemsted:
good point about documentation - i'll have a think, but hopefully if we get this starting to comment on every PR it'll be more obvious where and how to document it

PR #3904, opened 2023-10-02
Delete rows from notification_history in 50,000 chunks

idavidmcdonald:
This is because the query is currently taking over 20 minutes which is longer than our statement timeout limit and so is being cancelled.

We split it into smaller chunks so that we don't run into this problem. 50,000 has been chosen fairly arbirtrarily and is used as a query_limit elsewhere in our codebase.

We commit after each one so we commit progress regularly rather than just at the end.

We choose synchronize_session False as we do not need to access those notifications again that have been deleted.
https://docs.sqlalchemy.org/en/14/orm/session_basics.html#selecting-a-synchronization-strategy

### DIFF
--- a/app/dao/notification_history_dao.py
+++ b/app/dao/notification_history_dao.py
@@ -1,17 +1,35 @@
 from flask import current_app
+from sqlalchemy import select
 
-from app.dao.dao_utils import autocommit
+from app import db
 from app.models import NotificationHistory
 
 
-@autocommit
-def delete_notification_history_older_than_datetime(older_than_datetime):
+def delete_notification_history_older_than_datetime(older_than_datetime, query_limit=50000):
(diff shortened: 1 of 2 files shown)


robbrazil:
Wouldn't a simpler script that has only one iteration with the db make it lighter, we could even put a sleep betewwen each batch and make the batches smaller, like 
# Function to delete records in batches
def delete_records_in_batches():
    while True:
        with engine.connect() as connection:
            # Start a transaction
            with connection.begin():
                # Delete records older than OLDER_THAN_DATE, limited to BATCH_SIZE
                result = connection.execute(
                    text("DELETE FROM your_table_name WHERE created_at < :older_than_date LIMIT :batch_size"),
                    older_than_date=OLDER_THAN_DATE,
                    batch_size=BATCH_SIZE
                )
                # Get the number of deleted records
                deleted_records = result.rowcount

PR #3906, opened 2023-10-03
Fix typo in test assertion

klssmith:
This also adds the `notify_db_session` fixture to fix an error when the test is run as a single test.


idavidmcdonald (approved):
Good spot, should have used autospec or whatever that thing is!

PR #3908, opened 2023-10-04
Don't run notification_history purge task in staging

idavidmcdonald:
This will enable us to retain about 1 billion notifications in our staging notification_history table, which is roughly the same number that will be in production after this task has run against prod. This is important for our testing this week of migrating Postgres which will benefit from having staging look similar to production in size.

We did also consider taking a fresh copy of production, redacting the personal data and attaching it to the staging apps, but decided that was too much work and this is a quicker fix.


robbrazil (approved):

klssmith:
Might be nice to use [this approach](https://github.com/alphagov/notifications-api/blob/2f6e3a78af12ff5233c4c7c26bae165256675605/app/notify_api_flask_app.py) but that can be one for another time since this needs to go tonight

PR #3911, opened 2023-10-06
Let jobs be scheduled up to 7 days in the future 

quis:
4 days is sometimes not enough if there is a weekend with a bank holiday before and after it.

7 days is a sensible new limit because it:
- accounts for these longer weekends or holiday periods like Christmas
- without introducing any ambiguity about which ‘Thursday’ is meant (because only one of each weekday will ever be shown at once)

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -520,8 +520,8 @@ class JobSchema(BaseSchema):
         if value < datetime.utcnow():
             raise ValidationError("Date cannot be in the past")
 
-        if value > datetime.utcnow() + timedelta(days=4):
-            raise ValidationError("Date cannot be more than 4 days in the future")
+        if value > datetime.utcnow() + timedelta(days=7):
+            raise ValidationError("Date cannot be more than 7 days in the future")
 
     class Meta(BaseSchema.Meta):
         model = models.Job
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):
I think this could go in (checked out locally and looks good).

But if we're providing these extra dates, I start to feel like including the date+month would be useful. Some people definitely will be date oriented (I want these messages to send on the 1st of the month) rather than day oriented, and I think in either case it helps to provide extra immediate clarity that these are all in the future.

So probably I'm suggesting/pondering:

* Today
* Tomorrow
* Sunday 8th October
* Monday 9th October
* Tuesday 10th October
* Wednesday 11th October
* Thursday 12th October

quis:
The next page – once you have pressed the green button – presents the date like this:
image

quis:
Would look like:

image

***

image


***

image

karlchillmaid:
> But if we're providing these extra dates, I start to feel like including the date+month would be useful. Some people definitely will be date oriented (I want these messages to send on the 1st of the month) rather than day oriented, and I think in either case it helps to provide extra immediate clarity that these are all in the future.
> 
> So probably I'm suggesting/pondering:
> 
> * Today
> * Tomorrow
> * Sunday 8th October
> * Monday 9th October
> * Tuesday 10th October
> * Wednesday 11th October
> * Thursday 12th October
> 
> @karlchillmaid any thoughts? 🤔

I think this is a helpful addition, and Chris has already formatted the dates according to the style guide (no th or rd needed), so I’m happy 👍

PR #3914, opened 2023-10-09
Roll sentry out on more instances

samuelhwilliams:


app/performance.py
```diff
@@ -25,7 +25,7 @@ def init_performance_monitoring():
         send_request_bodies = "medium" if not_production else "never"
 
         # Enable tracing requests on only a single API instance in production while we check performance there.
```
klssmith, line 27: The comment is now out of date, but it looks like it's about to be deleted soon anyway so isn't too important

klssmith (approved):

PR #3915, opened 2023-10-09
Bump to psycopg2==2.9.6

samuelhwilliams:
After upgrading from pg11->pg15 locally, we started to see the following error: `sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) SCRAM authentication requires libpq version 10 or above`.

A suggested fix for this is to upgrade to psycopg2(-binary)==2.9.6, so that's what we do here.

---

If using `notifications-local`, you'll need to do a `make down`, `docker-compose build notify-api`, `make <opts> up` again to make sure your API container(s) get updated.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -16,7 +16,7 @@ itsdangerous==2.1.2
 jsonschema[format]==4.16.0
 marshmallow-sqlalchemy==0.28.1
 marshmallow==3.18.0
-psycopg2-binary==2.9.3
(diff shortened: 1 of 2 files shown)


leohemsted (approved):

leohemsted:
tested (briefly) and it seems to work fine with old pg11 as well as new pg15 so think this is good to go

PR #3917, opened 2023-10-10
remove periodic broadcast tasks

leohemsted:
none of these work anymore, and they'll just produce error logs


samuelhwilliams (approved):
👍 

It does seem like literally nothing else calls these tasks ... we _could_ drop the tasks and their tests right now.

But fine if you don't want to yet.

leohemsted:
🚮 ✅

samuelhwilliams (approved):
🎉 🔥

PR #3919, opened 2023-10-11
Random ordering of notification history

idavidmcdonald:
Review commit by commit

### DIFF
--- a/app/dao/notification_history_dao.py
+++ b/app/dao/notification_history_dao.py
@@ -6,13 +6,27 @@ from app import db
 def delete_notification_history_older_than_datetime(older_than_datetime, query_limit=50000):
     current_app.logger.info("Beginning to delete notification_history older than %s", older_than_datetime)
 
+    # We want to run this function concurrently, ie having several workers running this same task at the same time
+    # To ensure workers don't try and delete the same notifications (which will cause them to stop and exit)
+    # we get them to pick a random sample of notifications
+    # In production, to get 50,000 results we only need a random sample of about 0.05% of the table to give us
+    # 50,000 rows
+    # However, in other environments, including test, we don't have many notifications in the notification_history
+    # table and therefore need to get all of them (and don't need to worry about performance)
+    if current_app.config["NOTIFY_ENVIRONMENT"] == "production":
+        tablesample_percentage = 0.05
+    else:
+        tablesample_percentage = 100
+
     total_deleted = 0
     deleted_this_iteration = 1
     while deleted_this_iteration > 0:
         delete_query = f"""
             DELETE FROM notification_history
             where id in (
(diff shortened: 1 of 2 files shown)


app/dao/notification_history_dao.py
```diff
@@ -6,13 +6,27 @@
...
 
+    # We want to run this function concurrently, ie having several workers running this same task at the same time
+    # To ensure workers don't try and delete the same notifications (which will cause them to stop and exit)
+    # we get them to pick a random sample of notifications
+    # In production, to get 50,000 results we only need a random sample of about 0.05% of the table to give us
+    # 50,000 rows
+    # However, in other environments, including test, we don't have many notifications in the notification_history
+    # table and therefore need to get all of them (and don't need to worry about performance)
+    if current_app.config["NOTIFY_ENVIRONMENT"] == "production":
+        tablesample_percentage = 0.05
```
CrystalPea, line 17: will this value not change as notification_history shrinks?

CrystalPea (commented):
As a very temporary thing for a few days it could work, but as notification history table will shrink, the sample size will also shrink, no?

app/dao/notification_history_dao.py
idavidmcdonald: Good point Pea. I hadn't given this enough thought, or made it clear my thinking in the commit message and I should have.

After thinking it through, this value is OKish and would do the job but could maybe be better.

I checked and in our prod DB we have just over 1 billion notifications older than 2023-01-01 (ie these are the ones to not be deleted).

If we took 0.05 percent of 1 billion, it is 500,000. This means that tablesample will find *at least* 500,000 notifications. It will find more to start with whilst there are closer to 5 billion notifications, and at the end will find about 500,000 as we have removed the number of notifications. Crucially, we still have the `limit` clause on the end which will then pick only 50,000 of the ones picked by `tablesample`.

So 0.05 looks sufficiently large enough.

However, if we picked 0.005%, it might be more efficient. 0.005% of 1 billion is 50,000. So we would expect at least 50,000 notifications to be found each time still.

So the current value of 0.05% currently in the commit wouldn't encounter any problems, but we could potentially change it to 0.005% which I assume would result in quicker sql.

idavidmcdonald:
Closing this as
1. tablesample will return random rows from the table, but won't have applied the `created_at` filter before choosing these. This means our method of picking random rows is flawed. See https://dba.stackexchange.com/questions/258271/perform-tablesample-with-where-clause-in-postgresql#:~:text=tablesample%20is%20an%20%22attribute%22%20of,the%20table%20has%20been%20sampled

2. We are going to try a new approach of deleting hour by hour, instead of quarter by quarter. Pull request coming shortly!

PR #3921, opened 2023-10-12
Delete notification_history in hour chunks

idavidmcdonald:
This changes the approach in
delete_oldest_quarter_of_unneeded_notification_history which was too slow because it couldn't be run in parralell on multiple workers.

Instead, we queue up lots of small tasks that can be run in parraell which will each delete only a single hour of data.

I've tested some of these queries manually on production and for the 4 different 1 hour slots I tried, the queries took less than 10 seconds each.

The plan is to kick this task off once manually by logging into production and creating a task (rather than scheduling a task that we only want to run once which is a bit hard to do with cron).

If people are happy with this pull request, I'll merge it, run it and then come back to delete all the code for the old way of doing things.

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -32,7 +32,10 @@ from app.dao.jobs_dao import (
     dao_archive_job,
     dao_get_jobs_older_than_data_retention,
 )
-from app.dao.notification_history_dao import delete_notification_history_older_than_datetime
+from app.dao.notification_history_dao import (
+    delete_notification_history_between_two_datetimes,
+    delete_notification_history_older_than_datetime,
+)
 from app.dao.notifications_dao import (
     dao_get_notifications_processing_time_stats,
     dao_timeout_notifications,
@@ -409,3 +412,54 @@ def pick_bst_quarter_to_delete_back_from(oldest_notification_bst):
         deletion_date = deletion_date + dateutil.relativedelta.relativedelta(months=3)
 
     return deletion_date
+
+
+@notify_celery.task(name="delete_unneeded_notification_history_by_hour")
+def delete_unneeded_notification_history_by_hour():
+    # This task will delete all of the notification_history table older than 1 Jan 2023 BST
+    #
+    # This task will create lots of tasks, each one responsible for deleting a particular hour of
+    # notification_history that is no longer needed
+    #
+    # This retention limit is hardcoded and was originally picked from
+    # https://github.com/alphagov/notifications-aws/blob/main/decisions/2022-12-01-notification-history-retention-period.md
+    # It was supposed to be 2023-4-1.
+    # However, at the time of writing this code we realised the FtBillingLetterDispatch table has been introduced
+    # which means we need to store letters older than 2023-4-1 in order to rebuild that table (because that table
+    # uses the date of dispatch, not the date of creation for which date to bill for). To keep it simple, we keep
+    # an extra quarters worth of data giving us plenty of buffer
+    #
+    # In the future, we will be able to update this retention_limit value when we have progressed 3 quarters
+    # into the next financial year
+    #
+    # Arbitrary start_datetime, just slightly older than the oldest notification in the notification_history
+    # table at the time of writing
+    start_datetime = datetime(2020, 8, 1, 0, 0, 0)
+    retention_limit = datetime(2023, 1, 1, 0, 0, 0)
+
+    while start_datetime < retention_limit:
+        end_datetime = start_datetime + timedelta(hours=1)
+        delete_unneeded_notification_history_for_specific_hour.apply_async(
+            [start_datetime, end_datetime],
+            # We use the broadcasts queue temporarily as it pulled from by a worker doing no work
+            # We don't want to put the tasks on the periodic queue because they make take up all
+            # the workers capacity, stopping other important tasks from happening
+            queue=QueueNames.BROADCASTS,
+        )
+        current_app.logger.info(
+            "Created delete_unneeded_notification_history_for_specific_hour task between %s and %s",
+            start_datetime,
+            end_datetime,
+        )
+        start_datetime = end_datetime
+
+
+@notify_celery.task(name="delete_unneeded_notification_history_for_specific_hour")
+def delete_unneeded_notification_history_for_specific_hour(start_datetime, end_datetime):
+    current_app.logger.info(
+        "Beginning delete_unneeded_notification_history_for_specific_hour between %s and %s",
+        start_datetime,
+        end_datetime,
+    )
+
+    delete_notification_history_between_two_datetimes(start_datetime, end_datetime)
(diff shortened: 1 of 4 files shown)


app/celery/nightly_tasks.py
```diff
@@ -409,3 +412,49 @@ def pick_bst_quarter_to_delete_back_from(oldest_notification_bst):
...
+    # an extra quarters worth of data giving us plenty of buffer
+    #
+    # In the future, we will be able to update this retention_limit value when we have progressed 3 quarters
+    # into the next financial year
+    #
+    # Arbitrary start_datetime, just slightly older than the oldest notification in the notification_history
+    # table at the time of writing
+    # retention_limit is 1 Jan 2023 in UTC
+    start_datetime = datetime(2020, 8, 1, 0, 0, 0)
+    retention_limit = datetime(2022, 12, 31, 23, 0, 0)
```
klssmith, line 439: UTC and BST are the same in January, so we don't need to subtract an hour
```suggestion
    retention_limit = datetime(2023, 1, 1)
```

app/celery/nightly_tasks.py
```diff
@@ -409,3 +412,48 @@ def pick_bst_quarter_to_delete_back_from(oldest_notification_bst):
         deletion_date = deletion_date + dateutil.relativedelta.relativedelta(months=3)
 
     return deletion_date
+
+
+@notify_celery.task(name="delete_unneeded_notification_history_by_hour")
+def delete_unneeded_notification_history_by_hour():
```
samuelhwilliams, line 418: What's our plan for triggering this? A manual step, or something else?

idavidmcdonald: The plan is to kick this task off once manually by logging into production and creating a task. 

I did think about scheduling a task for a particular day of the year but it felt a bit ick and I thought there was a chance it would get left in the code (not to mention picking a day to schedule it for that is as soon as possible but is also after the time it takes to get this reviewed and approved).

Open to options though.

samuelhwilliams: I think the only reason I ask is that, in the case it's a manual thing, maybe we want the option to be able to provide a smaller daterange in the beginning as a test run. As it stands this will generate all of the hours across the entire period, so if something goes wrong from the get go then it's a bit more cleanup.

But I suspect it's probably manageable.

As a test run we could just manually log some of the child tasks with our own hour ranges.

And then later if it goes wrong we can just scale down the broadcast worker and purge the queue.

So probably not worth changing it after all.

app/celery/nightly_tasks.py
```diff
@@ -409,3 +412,48 @@ def pick_bst_quarter_to_delete_back_from(oldest_notification_bst):
...
+    # In the future, we will be able to update this retention_limit value when we have progressed 3 quarters
+    # into the next financial year
+    #
+    # Arbitrary start_datetime, just slightly older than the oldest notification in the notification_history
+    # table at the time of writing
+    start_datetime = datetime(2020, 8, 1, 0, 0, 0)
+    retention_limit = datetime(2023, 1, 1, 0, 0, 0)
+
+    while start_datetime < retention_limit:
+        delete_unneeded_notification_history_for_specific_hour.apply_async(
```
samuelhwilliams, line 441: I wouldn't hate a log message here so that we can see a clear record of all the hours we're generating, just as an audit trail of what we've made and a clear trail that the start/stop is correct when we do a prod run.

samuelhwilliams (approved):

app/dao/notification_history_dao.py
```diff
@@ -33,3 +33,18 @@ def delete_notification_history_older_than_datetime(older_than_datetime, query_l
     current_app.logger.info(
         "Deleted %s rows from notification_history older than %s", total_deleted, older_than_datetime
     )
+
+
+def delete_notification_history_between_two_datetimes(start, end):
```
samuelhwilliams, line 38:
```suggestion
def delete_notification_history_between_two_datetimes(start: datetime, end: datetime):
```

Also wouldn't hate some type hints on these. Not super sure if these _are_ datetimes or they end up as strings after the celery serialization...

idavidmcdonald: I've added this in and rebased it into my original commit

app/celery/nightly_tasks.py
idavidmcdonald: I've added this in and rebased it as part of my original commit

app/celery/nightly_tasks.py
idavidmcdonald: I'm happy you've talked yourself out of it :) It would be fairly quick to scale down the broadcasts worker if we want to stop things.

idavidmcdonald:
FYI both, I've made some slight tweaks and rebased in based on your comments.

Given this code, if something goes really wrong, could delete a lot of data that it shouldn't... could one of you please just give another approval just to make sure things nothing looks amiss. 

Thanks!

app/celery/nightly_tasks.py
```diff
@@ -409,3 +412,54 @@ def pick_bst_quarter_to_delete_back_from(oldest_notification_bst):
...
+            # We use the broadcasts queue temporarily as it pulled from by a worker doing no work
+            # We don't want to put the tasks on the periodic queue because they make take up all
+            # the workers capacity, stopping other important tasks from happening
+            queue=QueueNames.BROADCASTS,
+        )
+        current_app.logger.info(
+            "Created delete_unneeded_notification_history_for_specific_hour task between %s and %s",
+            start_datetime,
+            end_datetime,
+        )
```
samuelhwilliams, line 453: Something to bear in mind (I know I asked for this) is that this will make the task take quite a bit longer to run. Just something to be aware of - so we should try to avoid rolling the API shortly after kicking off this job. I don't think we need to remove this though.

app/dao/notification_history_dao.py
```diff
@@ -33,3 +35,18 @@ def delete_notification_history_older_than_datetime(older_than_datetime, query_l
     current_app.logger.info(
         "Deleted %s rows from notification_history older than %s", total_deleted, older_than_datetime
     )
+
+
+def delete_notification_history_between_two_datetimes(start: datetime, end: datetime):
```
samuelhwilliams, line 40: 💀 just tested this locally and it does end up getting called with strings. when we call `.apply_async` with the datetimes as args, celery calls `isoformat` (or something similar) on them automatically. But it doesn't reverse that on the way out. So probably worth updating this to `str` actually 🥲

samuelhwilliams (approved):
Reviewed this again, and run it again, locally - still looks good to me.

One comment on the type hints which, after checking now, are actually wrong unfortunately. But logically things look good.

app/dao/notification_history_dao.py
idavidmcdonald: Thanks Sam. I've removed the type hints for now. What I'm going to do is merge this and get it tested in prod and then going to add back in the type hints this afternoon (and adapt my tests to also being receiving strings) when I have no meetings!

PR #3923, opened 2023-10-16
drop email-from

leohemsted:
it's no longer used anywhere 🎉 

this has taken a while to get to

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0425_unique_service_name_2
+0426_drop_email_from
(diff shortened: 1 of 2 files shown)


idavidmcdonald:
Worth a read/scan of https://medium.com/paypal-tech/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#5d48 to make sure that this migration is zero downtime safe!

samuelhwilliams:
Just the latter.

leohemsted:
agree that there's a slight issue, but i'd like this in before the schema freeze 🤔

PR #3926, opened 2023-10-16
Make conversion from datetimes to strings clearer

idavidmcdonald:
It turns out that when we pass datetimes as args to the next task, celery behind the scenes must be converting them into strings in what appears to be using `isoformat()`.

This makes it explicit that
delete_unneeded_notification_history_for_specific_hour expects and requires strings because this is what it will actually receive from celery.

This updates the tests to correspond as well.


samuelhwilliams (approved):
Thanks for coming back to do this 👍

PR #3927, opened 2023-10-16
add new email sender columns to service and history table

leohemsted:
the `custom_email_sender_name` will be set by users with manage service permission on the front-end, to reflect a name they'd like their emails to appear from.

the `email_sender_local_part` will be set by lowercasing that value, and replacing spaces and punctuation with the period character. if the service hasn't set `custom_email_sender_name` then the local part should be set to be equal to normalised_service_name. eventually this column will have a not-null constraint. (we could also add a db constraint to check the equivalence of normalised_service_name too if we want)

On max data length:

The RFC3696[^1] states that local parts may be up to 64 characters. However, in practice, we have seen notify services succesfully sending emails with local parts longer than that - the key limitation is that the entire email address must be under 320 characters.

The entire from address is `"Service Name" <service.name@notifications.service.gov.uk>` (and this may change in the future if we add custom from domains). To be safe, set these to have a maximum limit of 100 characters, since the name and local part will typically be the same length (so ~200 chars total) plus ~34 characters for the domain, angle brackets, @, etc. That gives a buffer to ensure that we won't go over the max email address length.

There are four services with a name longer than 100 characters now (143 and 104 characters long). Two of them look like they were created in error but the other two are real and live. Both of those two send emails currently. i don't know what we should do with them

-----
### Todo before merge:

- [x] https://github.com/alphagov/notifications-api/pull/3923
- [x] decide on max length, work out what to do with teams that are currently over
- [ ] decide if we like the naming? i went for longer and more explicit given some of the grief i've had lately understanding terse column and variable names, but i'm open to ideas
- [ ] wait for next week for db migration
- [ ] merge with infra team so we can monitor migrations etc
 
[^1]: https://datatracker.ietf.org/doc/html/rfc3696#:~:text=In%20addition%20to%20restrictions,they%20are%20rarely%0A%20%20%20encountered.


idavidmcdonald:
When you want to merge this, can we do it together please (doesn't need to be me specifically, but someone on the infra team)? This will be our first schema migration in production since migration and I want us to just double check it as it goes out through every environment (including making sure the alembic table gets updated correctly).

risicle:
Only issue I see is having to backfill the services table if you ever change `make_string_safe_for_email_local_part(..)`, but i can understand you wanting to keep it as a stored field for the history table.

samuelhwilliams (approved):
This seems good code-wise, func tests pass 👍

PR #3928, opened 2023-10-17
Use Sentry to monitor celery beat tasks

samuelhwilliams:
Trial using Sentry to automatically monitor our scheduled tasks. At the moment this is _in addition_ to Cronitor, so we can just test whether Sentry meets all of our needs without interrupting our current workflow with Cronitor.

https://govuk-notify-gds.sentry.io/crons/?environment=development&project=4504949325824000&statsPeriod=7d

## Screenshots
image

image

### When tasks are missed, they are raised as native issues in Sentry
image

### DIFF
--- a/app/performance.py
+++ b/app/performance.py
@@ -1,6 +1,8 @@
 import os
 from functools import partial
 
+from sentry_sdk.integrations.celery import CeleryIntegration
+
 
 def sentry_sampler(sampling_context, sample_rate: float = 0.0):
     if sampling_context["parent_sampled"]:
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
@idavidmcdonald how do you feel about getting this in to trial it? For now, in addition to cronitor. I've muted sentry alerts for cron jobs, so it shouldn't generate any extra noise. Just a chance for us to contrast and compare with cronitor safely.

idavidmcdonald (approved):
Sure, happy for you to give this a try. Please do report back in a tech catch up some time on your findings so we can make a decision on it.

PR #3930, opened 2023-10-18
Raise sentry issue in slow text message delivery

samuelhwilliams:
Although we have nice graphs for monitoring slow text message delivery, they currently require proactive manual monitoring to spot slow delivery issues across the board.

Let's try to be a bit more proactive on this. We can check whether >10% or more of text messages have been slow (>5 minutes) in the last 15 minutes, and raise an error if slow. This should let us get a jump on investigating problems here.

10% is a _bit_ of a finger in the air from looking at the recent slow delivery graphs and trying to work out a reasonable point to be alerted at that is a balance between not being too noisy, and detecting levels of user impact that aren't tolerable.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -155,8 +155,8 @@ def _check_slow_text_message_delivery_reports_and_raise_error_if_needed(reports:
     if percent_slow_notifications >= 10:
         with sentry_sdk.push_scope() as scope:
             error_context = {
-                "Dashboard URL": (
-                    "https://grafana-notify.cloudapps.digital/d/icjsQ-MWk2/notify-summary?var-env=preview"
+                "Support runbook": (
+                    "https://github.com/alphagov/notifications-manuals/wiki/Support-Runbook#slow-sms-delivery"
                 ),
                 "Slow text messages - #": slow_notifications,
                 "Slow text messages - %": percent_slow_notifications,
@@ -168,8 +168,10 @@ def _check_slow_text_message_delivery_reports_and_raise_error_if_needed(reports:
                 error_context[f"provider.{report.provider}.slow_notifications"] = report.slow_notifications
                 error_context[f"provider.{report.provider}.total_notifications"] = report.total_notifications
 
-            scope.set_context("Slow delivery data", error_context)
-            current_app.logger.error(">10%% of text messages are taking longer than 5 minutes to deliver.")
+            scope.set_context("Slow SMS delivery", error_context)
+            current_app.logger.error(
+                "Over 10% of text messages sent in the last 15 minutes have taken over 5 minutes to deliver."
+            )
 
 
 @notify_celery.task(name="generate-sms-delivery-stats")
--- a/tests/app/celery/test_scheduled_tasks.py
+++ b/tests/app/celery/test_scheduled_tasks.py
@@ -214,7 +214,10 @@ def test_check_slow_text_message_delivery_reports_and_raise_error_if_needed(capl
             ),
         ]
     )
-    assert ">10%% of text messages are taking longer than 5 minutes to deliver." not in caplog.messages
+    assert (
+        "Over 10% of text messages sent in the last 15 minutes have taken over 5 minutes to deliver."
+        not in caplog.messages
+    )
 
     _check_slow_text_message_delivery_reports_and_raise_error_if_needed(
         [
@@ -224,7 +227,9 @@ def test_check_slow_text_message_delivery_reports_and_raise_error_if_needed(capl
             ),
         ]
     )
-    assert ">10%% of text messages are taking longer than 5 minutes to deliver." in caplog.messages
+    assert (
+        "Over 10% of text messages sent in the last 15 minutes have taken over 5 minutes to deliver." in caplog.messages
+    )
 
 
 def test_check_job_status_task_calls_process_incomplete_jobs(mocker, sample_template):


idavidmcdonald (commented):
So I've had a quick glance through to see the approach you've taken here.

I'm fine with the approach of alerting via sentry.

What I'd like to see, that you aren't doing in this PR, is the creation of a new statsd metric for slow delivery percentage across both providers (as in, the thing our users care about).

In your PR, I believe you calculate this number once per minute and if it is above a certain rate then we alert via sentry. But I don't think it will be so good to have an alert go off for a metric that you can't see a graph for. How will we tell if the metric is improving or resolved? How will we compare it to a baseline for what is normal? Potentially having to look at errors that come in once per minute to sentry with differing numbers of slow messages and see the pattern. I'd prefer that people can see this on a graph somewhere. At the moment, the 'easiest' way to do that is by setting up a new statsd metric. 

Maybe it would be `slow-delivery.sms.delivered-within-minutes.5.ratio` ie a provider called `sms` which is agnostic to which supplier sent the message. Bonus if you added a 1 minute and 10 minute interval for equivalence to the statsd metrics per provider.

Would be happy to chat this through in person if you want to chat through ideas/approach. Let me know!

samuelhwilliams:
I was going to go in and add that as a follow-up (as requires other PRs to scrape the stat as well etc), but will roll it in here since you've prompted 👍

idavidmcdonald:
Thanks. I think worth adding in this PR as I wonder if you would structure your code slightly differently to support that additional metric(s). Maybe/maybe not. I think I was initially expecting us to run the queries for getting the slow messages sent per provider once more, but making a slight tweak to them so they don't filter by provider. I wasn't expecting the approach of using data from the queries per provider to help calculate the total. Seems smarter and quicker to do what you've suggested but maybe introduces a bit more code complexity/a new pattern that I wasn't expecting. Maybe is still the right way to go, but wasn't what I initially had in mind (for what that is worth).

samuelhwilliams:
@idavidmcdonald have added some commits to generate the new stat and rename the function since it's got a slightly wider scope re: slow delivery. Happy to tweak approach if you think it'd be particularly better with your approach.

app/celery/scheduled_tasks.py
```diff
@@ -143,15 +145,57 @@ def switch_current_sms_provider_on_slow_delivery():
...
+    slow_notifications = sum(report.slow_notifications for report in reports)
+    percent_slow_notifications = (slow_notifications / total_notifications) * 100
+
+    # If over 10% of all text messages sent over the period have taken longer than 5 minutes to deliver, let's flag a
+    # sentry error for us to investigate.
+    if percent_slow_notifications >= 10:
+        with sentry_sdk.push_scope() as scope:
+            error_context = {
+                "Dashboard URL": (
+                    "https://grafana-notify.cloudapps.digital/d/icjsQ-MWk2/notify-summary?var-env=preview"
```
idavidmcdonald, line 159: Discuss: Good work on linking out directly to something to support with the investigation (Grafana). I wonder if we should link to Grafana or if we should link to a runbook entry in our team wiki that explains what an investigator should do (and also links to the Grafana graph too)

app/celery/scheduled_tasks.py
```diff
@@ -143,15 +145,57 @@ def switch_current_sms_provider_on_slow_delivery():
...
+                "Total text messages": total_notifications,
+            }
+
+            for report in reports:
+                error_context[f"provider.{report.provider}.slow_ratio"] = report.slow_ratio
+                error_context[f"provider.{report.provider}.slow_notifications"] = report.slow_notifications
+                error_context[f"provider.{report.provider}.total_notifications"] = report.total_notifications
+
+            scope.set_context("Slow delivery data", error_context)
+            current_app.logger.error(">10%% of text messages are taking longer than 5 minutes to deliver.")
```
idavidmcdonald, line 172: Typo: You've got two `%` symbols in here

idavidmcdonald, line 172: Optional: I wonder if we want to match this text to how our Grafana describes this

Grafana says 'This graph shows the percentage of text messages sent in the last 15 minutes that have taken over X minutes to be delivered.'

To write something that matches that language closely, you could try:

'Over 10% of text messages sent in the last 15 minutes have taken over 5 minutes to deliver'.

app/celery/scheduled_tasks.py
```diff
@@ -143,15 +145,57 @@ def switch_current_sms_provider_on_slow_delivery():
...
+                "Slow text messages - %": percent_slow_notifications,
+                "Total text messages": total_notifications,
+            }
+
+            for report in reports:
+                error_context[f"provider.{report.provider}.slow_ratio"] = report.slow_ratio
+                error_context[f"provider.{report.provider}.slow_notifications"] = report.slow_notifications
+                error_context[f"provider.{report.provider}.total_notifications"] = report.total_notifications
+
+            scope.set_context("Slow delivery data", error_context)
```
idavidmcdonald, line 171: Optional: Our Grafana panel for these is called 'Slow SMS delivery'. Maybe we should match it for consistency?

app/celery/scheduled_tasks.py
```diff
@@ -143,15 +145,57 @@ def switch_current_sms_provider_on_slow_delivery():
                 dao_reduce_sms_provider_priority(provider_name, time_threshold=timedelta(minutes=10))
 
 
+def _check_slow_text_message_delivery_reports_and_raise_error_if_needed(reports: list[SlowProviderDeliveryReport]):
+    total_notifications = sum(report.total_notifications for report in reports)
+    slow_notifications = sum(report.slow_notifications for report in reports)
+    percent_slow_notifications = (slow_notifications / total_notifications) * 100
+
+    # If over 10% of all text messages sent over the period have taken longer than 5 minutes to deliver, let's flag a
```
idavidmcdonald, line 153: No action needed: I think this is a good starting metric and threshold. This is because 5 minutes is the same period that we use for our sms load balancing. 

We may decide to increase the threshold but we can see what happens.

https://grafana-notify.cloudapps.digital/d/icjsQ-MWk2/notify-summary?orgId=1&viewPanel=51&from=now-7d&to=now

Looking at the graph for sms delivery slow than 5 minutes over the last 7 days, a very very rough eyeball gives me about 7 points in the past 7 days where both sms providers were both above 10%. Assuming traffic was split equally then that would breach this alert.

idavidmcdonald (approved):
Thanks Sam, this is looking pretty good.

I've left a bunch of small (and hopefully not too pedantic) comments for potential small tweaks that can make this alert even easier to understand and investigate. 

Given those are all fairly stylistic, happy to approve this as is and leave you to tweak any wording as you see fit.

My reckoning (without looking at the statsd code or the grafana graphs) is that when we merge this code to production, we will get a new metric on our 3 'slow delivery' panels in grafana for a new 'provider' called `sms`. Do you think we should do anything about that?

Thoughts include
- update the description on the left of the panels to make it clear what this is and that all the lines on the graphs aren't per provider
- could even split the provider agnostic line into a separate graph. Pros and cons to this. More panels on an already busy dashboard. But then maybe a panel that can be entirely user focussed and you don't need to care about what is happening behind the scenes when looking at it. I'm undecided on what is best (maybe leaning towards no new panel).
- talking the team through the new metric (not just the tech team) and alert

app/celery/scheduled_tasks.py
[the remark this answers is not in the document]
samuelhwilliams: Oops, I think I meant to go and inject the environment using NOTIFY_ENVIRONMENT env var. Good spot.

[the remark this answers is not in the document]
samuelhwilliams: That's true, yes. I think that's OK, but yeah we'll have to remember to do it (or else do it when the alert is raised and we find the URL doesn't work any more 😓)

samuelhwilliams: That might be a better idea, you're right - avoids the grafana URL issue as well. I will set one up and link to it 👍

app/celery/scheduled_tasks.py
```diff
@@ -143,15 +145,57 @@ def switch_current_sms_provider_on_slow_delivery():
...
+    # sentry error for us to investigate.
+    if percent_slow_notifications >= 10:
+        with sentry_sdk.push_scope() as scope:
+            error_context = {
+                "Dashboard URL": (
+                    "https://grafana-notify.cloudapps.digital/d/icjsQ-MWk2/notify-summary?var-env=preview"
+                ),
+                "Slow text messages - #": slow_notifications,
```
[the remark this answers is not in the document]
samuelhwilliams: I add it here because we have a percentage and a raw number with the same prefix, unlike with total text messages 🤔

app/celery/scheduled_tasks.py
samuelhwilliams: Good idea 👍  Updated.

app/celery/scheduled_tasks.py
samuelhwilliams: Yep, definitely something we'll need to keep an eye on if it's looking flappy or noisy. Possibly we need to only alert on it if there's 2 or 3 slow errors in a row (ie there's prolonged delivery issues)

PR #3933, opened 2023-10-23
Set `application_name` on pg connections

samuelhwilliams:
Set `application_name` on postgres connections

Setting this attribute may help us narrow down/debug which apps are running
queries that hold locks in contention.

See also:

* https://postgresqlco.nf/doc/en/param/application_name/
* https://wiki.postgresql.org/wiki/Lock_Monitoring

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -74,6 +74,9 @@ def create_app(application):
     application.config.from_object(configs[notify_environment])
 
     application.config["NOTIFY_APP_NAME"] = application.name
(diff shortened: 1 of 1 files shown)


risicle:
Yes please :heart:

app/__init__.py
```diff
@@ -74,6 +74,9 @@ def create_app(application):
     application.config.from_object(configs[notify_environment])
 
     application.config["NOTIFY_APP_NAME"] = application.name
```
samuelhwilliams, line 76: Separately, I think this is a really weird thing we're setting and I might look at updating this. I think it's always just "app" 🙃

CrystalPea (approved):

PR #3934, opened 2023-10-24
Reduce number of slow sms delivery errors

samuelhwilliams:
Let's only log an error (and thereby raise a Sentry error) if we've seen delivery be slow for >10% of text messages for the last 15 minutes consecutively. This should help increase the signal of the error, indicating that every time it happens we actually need to take a look and investigate properly. It's been a bit noisy and flappy so far.


app/celery/scheduled_tasks.py
```diff
@@ -153,25 +154,36 @@ def _check_slow_text_message_delivery_reports_and_raise_error_if_needed(reports:
...
+                    error_context[f"provider.{report.provider}.slow_notifications"] = report.slow_notifications
+                    error_context[f"provider.{report.provider}.total_notifications"] = report.total_notifications
 
-            scope.set_context("Slow SMS delivery", error_context)
-            current_app.logger.error(
-                "Over 10% of text messages sent in the last 15 minutes have taken over 5 minutes to deliver."
-            )
+                scope.set_context("Slow SMS delivery", error_context)
+                current_app.logger.error(
+                    "Over 10% of text messages sent in the last 15 minutes have taken over 5 minutes to deliver."
```
whpearson, line 182: Should we update this wording? Am I right in thinking it is the last 45 minutes 10% of the text messages have taken over 5 minutes to deliver?

samuelhwilliams: Not quite. Exactly what this means is a bit tricky to pin down.

However this task runs every 5 minutes, so 3 consecutive failures will take 10 minutes to happen (first, then 5 mins later, then another 5 mins later).

Exactly what this necessarily means is a bit hard for me to pin down mentally even in my own head. There's quite a few ways we could hit this issue...

We could have consistent delivery issues where 10% of _all_ messages are taking >5 mins to deliver. We could have a little burst where, in some 5 minute period, 30% of _all_ messages get delayed for >5 mins and then nothing else is delayed at all. That 5 minute period will show up in 3 of these checks, and averaged over 15 minutes will show up as 10% of all messages in a 15 minute period. etc

samuelhwilliams: Actually ... this task runs every minute, which I didn't realise 🙃  Which makes a lot of my above 'logic' incorrect.

I might change this to require 5 consecutive detections.

whpearson: Makes sense. I think the error message should be updated to be truthful. So if it looks back 15 minutes, and runs every minute then for 5 runs it would be 19 minutes? Does that make sense? Or could we just extend the window we look back and not have it have to have multiple failures?

samuelhwilliams: I want to keep the message fairly succinct and I believe the current error message _is_ truthful, it's just not _exacting_.

Every time we record a failure, it must be the case that within the last 15 minutes, 10% or more of text messages have taken 5 minutes to deliver.

We just only actually error on this if that is the case 5x in a row, so as to provide us some certainty that it's not just a blip in delivery but a persistent issue we need to investigate.

---

I'm open to alternative suggestions but I don't want to get too blocked on this as we can always change the message in the future.

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -205,31 +205,54 @@ def test_generate_sms_delivery_stats(environment, expect_check_slow_delivery, mo
...
+                    ),
+                ]
+            )
+            assert (
+                "Over 10% of text messages sent in the last 15 minutes have taken over 5 minutes to deliver."
+                not in caplog.messages
+            )
+            assert mock_set.call_args_list == [mocker.call("slow-sms-delivery:number-of-times-over-threshold", 0)]
+
+        # At 10%+, should increment redis and log an error when it's the third consecutive call.
```
leohemsted, line 237:
```suggestion
        # At 10%+, should increment redis and log an error when it's the fifth consecutive call.
```

app/celery/scheduled_tasks.py
```diff
@@ -153,25 +154,36 @@ def _check_slow_text_message_delivery_reports_and_raise_error_if_needed(reports:
...
-                ),
-                "Slow text messages - #": slow_notifications,
-                "Slow text messages - %": percent_slow_notifications,
-                "Total text messages": total_notifications,
-            }
+        count = redis_store.incr(CacheKeys.NUMBER_OF_TIMES_OVER_SLOW_SMS_DELIVERY_THRESHOLD)
+
+        # If this is the third consecutive time we've seen the threshold breached, then we log an error to Sentry.
+        # This tells us that for at least 15 minutes we've seen delivery take longer than 5 minutes for >10% of
+        # requests. At that point it's probably worth investigating. By only checking for == 3, we don't log any
```
leohemsted, line 161:
```suggestion
        # If this is the fifth consecutive time we've seen the threshold breached, then we log an error to Sentry.
        # This tells us that for at least 15 minutes we've seen delivery take longer than 5 minutes for >10% of
        # requests. At that point it's probably worth investigating. By only checking for == 5, we don't log any
```

i think the sentry msg is obvious enough (given that the first action i'd probably take after seeing this would be to check grafana and see the provider graphs) but if we do want to capture the nuance that @whpearson raised around it reflecting ~20 minutes then this comment is a reasonable spot

leohemsted (approved):

PR #3935, opened 2023-10-25
Correct the new postage rates for first class

CrystalPea:
as we received corrected rates from DVLA.


migrations/versions/0428_letter_rates_errata.py
```diff
@@ -0,0 +1,107 @@
...
+                start_date=start_date,
+                sheet_count=sheet_count,
+                rate=rate,
+                crown=crown,
+                post_class=post_class,
+            )
+
+
+def downgrade():
+    INCORRECT_RATES = [
```
leohemsted, line 67: im now wondering if there's even a need for this downgrade script.

like, i cant see any value to inserting incorrect rates we know have never been used by dvla. if we don't do anything on downgrade (literally just `pass`), and run upgrade again, it should still delete everything and insert the correct rates without duplicates, which is the key thing i think, to keep the db in a consistent state

leohemsted (approved):
One question about the downgrade butt think that's non-essential

migrations/versions/0428_letter_rates_errata.py
CrystalPea: I agree the only need for the downgrade script would be if DVLA did double take and said the rates were right in the end 😅 .

I can remove it if you think that's better? I'm fine either way ☺️

leohemsted: hah if they did a double take i think i'd cry

(but even then, we'd want a _new_ upgrade script, rather than downgrading and then removing this script entirely)

yeah i think removing is better, signifies intent. and having `incorrect_rates` gives me the heebie jeebies 😅.

we use `def downgrade(): pass` a bunch in other migrations that just insert data by the looks of things

leohemsted (approved):

PR #3937, opened 2023-10-30
Add 'is_precompiled_letter' to template history

samuelhwilliams:
We now rely on this attribute when rendering letter template history versions.

Fixes: https://govuk-notify-gds.sentry.io/issues/4586197393/?end=2023-10-30T23%3A59%3A59&project=4504949328838656&referrer=issue-stream&start=2023-10-01T00%3A00%3A00&stream_index=0&utc=true

Tested locally and does fix the admin pages.


leohemsted (approved):
can we add a unit test?
just sticking a line in test_template_history_version in tests/app/template/test_rest_history.py is probs fine

PR #3939, opened 2023-10-31
Add endpoints for org service approval flow

samuelhwilliams:
The updated flow for org members to approve/reject services needs to send two new emails:

If the service is outright rejected, the user that submitted the go-live request receives an email confirming the rejection. We include an (org) user-supplied reason for the rejection.

If we believe the org user needs to talk to the service user, then we send an email to the org user with details about what to do next and how to pick up the journey afterwards. This email contains details about what they need to find out (if it's unique, and whether the name is clear enough). The content for this email is not simple - we adjust it dynamically based on answers the org user provides. Therefore the template is simply a personalisation variable, and we will generate the appropriate markdown in the admin app and pass it through to the API. This is because the content will closely reflect what we show to the org user on the admin page, and so co-locating these things should make keeping them in sync easier.

---

Admin PR: https://github.com/alphagov/notifications-admin/pull/4854


tests/app/conftest.py
```diff
@@ -893,6 +893,58 @@ def organisation_has_new_go_live_request_template(notify_service):
...
+    )
+
+    return create_custom_template(
+        service=notify_service,
+        user=notify_service.users[0],
+        template_config_name="ORGANISATION_REJECTED_GO_LIVE_REQUEST_TEMPLATE_ID",
+        content=template_content,
+        subject="Your request to go live has been rejected",
+        template_type="email",
+    )
```
samuelhwilliams, line 945: This all makes me a bit sad but it follows what other templates do and I'm not changing that for now...

migrations/versions/0429_go_live_templates.py
```diff
@@ -0,0 +1,158 @@
...
+reject_template_id = "507d0796-9e23-4ad7-b83b-5efbd9496866"
+reject_template_content = textwrap.dedent(
+    """
+Hi ((name))
+
+You sent a request to go live for the following GOV.UK Notify service:
+
+((service_name))
+
+The request was rejected by ((organisation_team_member_name)) at ((organisation_name)). They gave the following reason:
```
karlchillmaid, line 32:
```suggestion
# Your request to go live was rejected

You sent a request to go live for a GOV.UK Notify service called ‘((service_name))’.

((organisation team member)) at ((organisation name)) rejected the request for the following reason:
```

tests/app/conftest.py
```diff
@@ -893,6 +893,57 @@ def organisation_has_new_go_live_request_template(notify_service):
...
+def organisation_reject_go_live_request_template(notify_service):
+    template_content = textwrap.dedent(
+        """\
+        Hi ((name))
+
+        You sent a request to go live for the following GOV.UK Notify service:
+
+        ((service_name))
+
+        The request was rejected by ((organisation_team_member_name)) at ((organisation_name)). They gave the following reason:
```
karlchillmaid, line 924:
```suggestion
# Your request to go live was rejected

You sent a request to go live for a GOV.UK Notify service called ‘((service_name))’.

((organisation team member)) at ((organisation name)) rejected the request for the following reason:
```

app/config.py
```diff
@@ -183,6 +183,8 @@ class Config(object):
     MOU_SIGNED_ON_BEHALF_SIGNER_RECEIPT_TEMPLATE_ID = "c20206d5-bf03-4002-9a90-37d5032d9e84"
     MOU_SIGNED_ON_BEHALF_ON_BEHALF_RECEIPT_TEMPLATE_ID = "522b6657-5ca5-4368-a294-6b527703bd0b"
     ORGANISATION_HAS_NEW_GO_LIVE_REQUEST_TEMPLATE_ID = "5c7cfc0f-c3f4-4bd6-9a84-5a144aad5425"
+    ORGANISATION_CONTINUE_GO_LIVE_REQUEST_TEMPLATE_ID = "62f12a62-742b-4458-9336-741521b131c7"
+    ORGANISATION_REJECTED_GO_LIVE_REQUEST_TEMPLATE_ID = "507d0796-9e23-4ad7-b83b-5efbd9496866"
```
CrystalPea, line 187: [naming suggestion] I would rename those constants a bit, so:
- domain comes first, here domain is the go-live-request
- we know who rejected the request
- we know who gets the email about next steps.

So the structure would be [domain]-[action]-[how org user relates to this action]

`GO_LIVE_REQUEST_REJECTED_BY_ORG_USER_TEMPLATE_ID`

`GO_LIVE_REQUEST_NEXT_STEPS_FOR_ORG_USER_TEMPLATE_ID`

app/organisation/rest.py
```diff
@@ -324,6 +326,71 @@ def notify_users_of_request_to_go_live(service_id):
...
 
+@organisation_blueprint.route(
+    "/notify-org-member-about-continuation-of-go-live-request/<uuid:service_id>", methods=["POST"]
+)
+def notify_org_member_about_continuation_of_go_live_request(service_id):
+    data = request.get_json()
+    validate(data, post_notify_org_member_about_continuation_of_go_live_request)
+
+    template = dao_get_template_by_id(current_app.config["ORGANISATION_CONTINUE_GO_LIVE_REQUEST_TEMPLATE_ID"])
+    service = dao_fetch_service_by_id(service_id)
```
CrystalPea, line 337: [suggestion] Should we rename `service` to `subject_service` or `trial_mode_service`, or `go_live_service` or something, since we already also get and use GOV.UK Notify service in this view, and it's a bit confusing?

Similar in the other view.

migrations/versions/0429_go_live_templates.py
```diff
@@ -0,0 +1,158 @@
...
+            current_timestamp,
+            '{current_app.config["NOTIFY_USER_ID"]}'
+        )
+        ON CONFLICT DO NOTHING
+        """
+    )
+
+
+def downgrade():
+    # I don't consider there to be a safe downgrade path as other tables could theoretically have FK dependencies
```
CrystalPea, line 155: [comment, non-actionable] I agree once the feature is live, the downgrade would not be recommended to avoid losing notifications. I think downgrades are much more useful if something goes wrong during deployment. However, since this migration doesn't change schema, but just adds data, I don't think we need one.

tests/app/organisation/test_rest.py
```diff
@@ -1290,3 +1290,133 @@ def test_notify_org_users_of_request_to_go_live_requires_org_user_permission(
...
+        service_id=sample_service.id,
+        _data={"to": "first-org-user@example.gov.uk", "service_name": "test service", "body": "blah blah blah"},
+        _expected_status=204,
+    )
+
+    assert mock_persist_notification.call_count == 1
+    assert mock_persist_notification.call_args_list[0][1]["recipient"] == "first-org-user@example.gov.uk"
+
+    assert mock_persist_notification.call_args_list[0][1] == dict(
+        template_id=uuid.UUID("62f12a62-742b-4458-9336-741521b131c7"),
```
CrystalPea, line 1340: [question] we prefer to hardcode the template id here rather than access it from config? (same for the other test)

CrystalPea (commented):
Hi there, good work, this is a complex piece of work and it took me of back and forth between here and admin PR to understand what is going on.

I suggested some naming/wording changes - let me know what you think.

app/organisation/rest.py
```diff
@@ -324,6 +326,71 @@ def notify_users_of_request_to_go_live(service_id):
     return {}, 204
 
 
+@organisation_blueprint.route(
+    "/notify-org-member-about-continuation-of-go-live-request/<uuid:service_id>", methods=["POST"]
+)
+def notify_org_member_about_continuation_of_go_live_request(service_id):
```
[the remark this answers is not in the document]
samuelhwilliams: Good suggestion, will change

app/config.py
samuelhwilliams: I will make that change. Will need to change the one above also, which is about go live requests too.

tests/app/organisation/test_rest.py
samuelhwilliams: yep, I'm generally fairly OK hard-coding values in tests, especially in these cases where they should never change anyway and if they do change I'd quite like the test to break to require extra validation of intent.

app/organisation/rest.py
samuelhwilliams: I'm fairly ambivalent on this. I think it's fairly standard for `service` to refer to the service that the request is primarily about. Calling it `go_live_service` might be more clear, but when working across the API if the subject of every API action had a different name I think I'd find that to be more mental overhead.

I maybe consider this to be a bit like `x` and `i` variables, where the non-descriptive name has some broader understanding behind it. But I also didn't think about it _that_ much...

CrystalPea: Ok, that's fair 👍🏼

app/organisation/rest.py
```diff
@@ -39,6 +39,8 @@
 from app.organisation.organisation_schema import (
     post_create_organisation_schema,
     post_link_service_to_organisation_schema,
+    post_notify_org_member_about_next_steps_of_go_live_request,
+    post_notify_service_member_of_rejected_request_to_go_live,
```
CrystalPea, line 43: [naming nitpick, optional]  change to `post_notify_service_member_of_rejected_go_live_request` so both functions have `go_live_request` wording.

CrystalPea (approved):

migrations/versions/0429_go_live_templates.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+Hi ((name))
+
+# Your request to go live was rejected
+
+You sent a request to go live for a GOV.UK Notify service called ‘((service_name))’.
+
+((organisation_team_member_name)) at ((organisation_name)) rejected the request for the following reason:
+
+((reason))
```
karlchillmaid, line 34: @samuelhwilliams Can we only add the inset text (`^`) formatting once this exists in the Notify admin interface?

samuelhwilliams: `reason` comes through to us here with the inset text formatting baked in by admin.

However I might tweak that so that it happens here instead, so will make a small change...

PR #3942, opened 2023-11-02
Add "ses-callbacks" to list of queues

klssmith:
In order to run the email-provider-stub locally, the `ses-callbacks` queue needs to be added to the list of queues. The worker watches all queues defined here when running locally.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -29,6 +29,7 @@ class QueueNames(object):
     CALLBACKS = "service-callbacks"
     CALLBACKS_RETRY = "service-callbacks-retry"
     LETTERS = "letter-tasks"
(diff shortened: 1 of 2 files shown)


idavidmcdonald (approved):
The bit I was wondering about was how have we got away without this all this time?

I think the answer is that when one of our local dev environments sends an email, it is sent by SES in the AWS Tools account. When SES receives a delivery receipt, it gets sent to an SNS subscription.

However, unlike preview, staging, production, we don't have a lambda subscribed to that SNS subscription to go put the task on a relevant queue (it wouldn't know which of our queues it should put it on).

So we've never been able to receive delivery receipts for SES to our local dev environments. That is why we've been able to survive without explicitly listening to this queue.

PR #3943, opened 2023-11-03
email sender column defaults and history cleanup

leohemsted:
move the decision making about what `normalised_service_name` and `email_sender_local_part` should be set to from the admin app to the model, to ensure that whenever services are updated, their email address is kept consistent.

Also, clean up a lot of the `Versioned` class that automatically creates history tables for some of our models. The history_meta was pulled from sqlalchemy's examples[^1] without modification - however, this has a lot of complexity to account for things we never use, such as polymorphic sql table inheritance. We're able to rip out a lot of this, which makes it a lot simpler - while this makes the file less futureproof if we want to be able to support those things in the future, it means we have a lot less actual code to read if we just want to understand what's going on. (It's worth saying that that example has already diverged since we copied it verbatim, so I think it's better to have shorter code we're more likely to understand, especially if we don't know why the example changed, if they fixed bugs with it etc).

Review note: a lot of the history_meta changes were also whitespace, so review with whitespace off for a slightly easier time. Commit by commit recommended.

- [x] https://github.com/alphagov/notifications-api/pull/3927

[^1]: https://docs.sqlalchemy.org/en/20/_modules/examples/versioned_history/history_meta.html


samuelhwilliams (approved):
So, I think this all looks good. The name/normalised service name cleanup is nice too.

I'm _reasonably_ confident this all works as expected. I've had a play around, a dig through, and run functional tests locally and all seems to be as expected.

Cleaning up some of the history_meta stuff is nice. 🔥

PR #3946, opened 2023-11-07
Remove PaaS Redis and PaaS Postgres

idavidmcdonald:

I started by removing the code for Redis as part of https://trello.com/c/9XD1L1cR/430-delete-old-paas-redis-and-references-to-it, but it turns out the postgres code is essentially the same and so I also took the opportunity to remove the postgres side of things too!


c4seyla (approved):
Nice :)

PR #3947, opened 2023-11-09
Fix POST template schema for bilingual letters

samuelhwilliams:
The previous schema had a flaw: if `letter_languages` was not passed, the schema was unintentionally requiring `letter_welsh_subject` and `letter_welsh_content` to be null values, if present.

The new schema allows any of the following options:

* The three "letter" language keys are missing
* `letter_languages` to be missing, in which case the pair of `letter_welsh_subject` and `letter_welsh_content` can be present, together, with matching data types (both null or both strings).
* `letter_languages` to be `english`, then both `letter_welsh_subject` and `letter_welsh_content` must be present and null
* `letter_languages` to be `welsh_then_english`, then both `letter_welsh_subject` and `letter_welsh_content` must be present and strings

### DIFF
--- a/app/template/template_schemas.py
+++ b/app/template/template_schemas.py
@@ -41,28 +41,47 @@ post_update_template_schema = {
         "archived": {"type": "boolean"},
         "current_user": uuid,
         "letter_languages": {"type": "string", "enum": [i.value for i in LetterLanguageOptions]},
-        "letter_welsh_subject": {},
-        "letter_welsh_content": {},
     },
+    # read as: "if `letter_languages` is present
     "if": {
-        "properties": {"letter_languages": {"const": LetterLanguageOptions.english.value}},
+        "required": ["letter_languages"],
     },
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
> letter_languages to be missing, in which case the pair of letter_welsh_subject and letter_welsh_content can be present, together, with matching data types (both null or both strings).

In practice, it's likely that we'd only send `letter_welsh_subject` and `letter_welsh_content` together as strings - when editing the Welsh version of the template. I don't think it's necessarily likely that we'd ever send those two fields as `null` without also setting `letter_languages` to `english` at the same time - but I think I'm ok with leaving that theoretical option open since it's potentially valid, as far as the JSON validator is concerned.

I may come back and add some additional python logic on the endpoint to make sure that we aren't setting/clearing `letter_welsh_subject/content` if the `letter_languages` field is in the wrong state for that, though.

PR #3950, opened 2023-11-14
Require 10 consecutive minutes of slow delivery

samuelhwilliams:
After some more discussion, we want to settle on requiring 10 consecutive breaches before triggering an error to reduce potential false positives.


idavidmcdonald (approved):
Nice

PR #3952, opened 2023-11-15
migrations: add `0438_min_numeric_scl_bulk`

risicle:
Sits on top of #3949

This updates the numerics in `notifications`, `notification_history` and `ft_billing` to their min_scale following the DMS migration.

Performs updates in blocks (partitioned on a pseudorandom uuid) to avoid locking any whole tables (or even one specific time-period) at once.

#3949 should be merged & released well before this to avoid the situation where any old values picked out of the database before this was applied are inserted back into the db while this is running, escaping the update.

### DIFF
--- a/migrations/versions/0438_min_numeric_scl_bulk.py
+++ b/migrations/versions/0438_min_numeric_scl_bulk.py
@@ -1,0 +1,95 @@
+"""
+
+Revision ID: 0438_min_numeric_scl_bulk
+Revises: 0437_min_numeric_scl_aux_tbls
+
+"""
+
+revision = "0438_min_numeric_scl_bulk"
+down_revision = "0437_min_numeric_scl_aux_tbls"
+
+import uuid
+
+from alembic import op
+from sqlalchemy import (
+    Numeric,
+    and_,
+    case,
+    cast,
+    column,
+    func,
+    table,
+    update,
+)
+from sqlalchemy.dialects.postgresql import UUID
+
+
+def _get_cases(var, max_scale=7):
+    # values used in types must be constants so we need
+    # to do this slightly ridiculous case statement covering
+    # each scale we expect to encounter
+    return case(
+        {i: cast(var, Numeric(1000, i)) for i in range(max_scale)},
+        value=func.min_scale(var),
+        else_=var,
+    )
+
+
+def upgrade():
+    # the following operations should be idempotent, so in the case of a failure,
+    # re-applying all in a retry should cause no issues
+
+    with op.get_context().autocommit_block():
+        conn = op.get_bind()
+
+        # apply in blocks to avoid locking whole table at once
+        n_blocks = 16
+        block_step = (1 << 128) // n_blocks
+
+        for table_name in (
+            "notifications",
+            "notification_history",
+        ):
+            _table = table(
+                table_name,
+                column("id", UUID(as_uuid=True)),
+                column("rate_multiplier"),
+            )
+
+            for block_start in range(0, 1 << 128, block_step):
+                block_end = block_start + block_step
+                conn.execute(
+                    update(_table)
+                    .values(rate_multiplier=_get_cases(_table.c.rate_multiplier))
+                    .where(
+                        and_(
+                            _table.c.id >= uuid.UUID(int=block_start),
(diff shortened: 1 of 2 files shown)


migrations/versions/0438_min_numeric_scl_bulk.py
```diff
@@ -0,0 +1,95 @@
...
+            "notification_history",
+        ):
+            _table = table(
+                table_name,
+                column("id", UUID(as_uuid=True)),
+                column("rate_multiplier"),
+            )
+
+            for block_start in range(0, 1 << 128, block_step):
+                block_end = block_start + block_step
```
samuelhwilliams, line 60: Do we have any idea how long this will take? Slightly scares me. Also wondering if it might be faster to do it based on created_at ranges instead for notifications_history? 🤔

risicle: Are you worried about total time taken or how long locks are held? Because the latter should be addressable by increasing the number of blocks.

I'd have to redesign the logic a bit to use `created_at` - the beauty of using UUIDs is they have a uniform range. With `created_at` I'd need to find the earliest timestamp, then would we do it day at a time? week at a time? until we reach the latest timestamp..

risicle: You raise an interesting point though, I'm curious how strongly `created_at` correlates with heap order in `notification_history` after all this time...

samuelhwilliams: > Are you worried about total time taken or how long locks are held? Because the latter should be addressable by increasing the number of blocks.

Both, to a degree. I'm more curious than worried at the moment, but this does broadly remind me of this PR (or really, the ones preceding it): https://github.com/alphagov/notifications-api/pull/3921

David was writing a job to delete notifications from the history table. We started doing it in quarter-by-quarter chunks but that was too slow and timed out. I think then we considered some table sampling process. And then we settled on hour-by-hour chunks, which was v fast.

I'm not sure we need to go hour-by-hour, but there was a significant benefit to using small date-based increments. Still took about ~4 hours to delete everything in scope. Obviously those were deletes rather than updates, so slightly different profile, but I think worth a cursory investigate first. Splitting the table into only 16 chunks feels quite coarse though. Based on an estimate of the NH table size, each chunk would affect ~78,551,087 rows.

risicle: The correlation for `created_at` is `0.6957699` so yes, it's quite ordered.

However I've got to wonder how ordered it's going to be after I've rewritten every row :blush:

risicle: Is there any concern with potentially locking a whole day/week at a time for `notification_history`?

samuelhwilliams: A day at a time should hopefully be fairly uncontentious, as long as it's not running overnight (when we generate billing) - and then, only for the most recent days.

I can't think of any other large scale queries that would be in conflict during the day.

risicle: I would expect updates to be slower than deletes.

samuelhwilliams: Me also.

I wouldn't be against you running some sample queries to get some realistic idea of performance.

This could also be something we do outside of a migration, since migrations block deploys. It could just be a script or command that we trigger manually.

samuelhwilliams: In fact I think it will need to be, as I've just remembered that we have a timeout on migrations. I'm pretty sure the pipeline auto-fails them (or at least considers them to have failed) if they take longer than 10 minutes. 

https://github.com/alphagov/notifications-aws/blob/main/concourse/templates/api.yml.j2#L197-L212

risicle: I take it we don't have a way of overriding this on a per-migration basis.

It might be more useful to have a per-transaction timer but that would require tinkering with alembic/sqlalchemy.

risicle:
Going to re-open this as a new PR because it's going to be so different.

PR #3956, opened 2023-11-20
use email_sender_local_part to send messages

leohemsted:
also clean up todos of removing normalised_service_name - it's no longer in the admin app so is safe to not expect in json

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -517,6 +517,7 @@ class Service(db.Model, Versioned):
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     _name = db.Column("name", db.String(255), nullable=False, unique=True)
+    # this isn't intended to be accessed, just used for checking service name uniqueness. See `email_sender_local_part`
     _normalised_service_name = db.Column("normalised_service_name", db.String, nullable=False, unique=True)
 
     @hybrid_property  # a hybrid_property enables us to still use it in queries
@@ -532,14 +533,6 @@ class Service(db.Model, Versioned):
         if not self.custom_email_sender_name:
             self._email_sender_local_part = self._normalised_service_name
 
-    @hybrid_property
-    def normalised_service_name(self):
-        return self._normalised_service_name
-
-    @normalised_service_name.setter
-    def normalised_service_name(self, value):
-        raise NotImplementedError("normalised_service_name can only be written to via `name`")
-
     _custom_email_sender_name = db.Column("custom_email_sender_name", db.String(255), nullable=True)
     # TODO: once data is migrated this should be not nullable
     _email_sender_local_part = db.Column("email_sender_local_part", db.String(255), nullable=True)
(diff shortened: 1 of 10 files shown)


app/serialised_models.py
```diff
@@ -78,10 +78,10 @@ class SerialisedService(SerialisedModel):
         "name",
         "active",
         "contact_link",
+        # TODO: do we still need this?
         "normalised_service_name",
```
quis, line 82: I think this will already be missing in the case where the admin app has populated the cache

app/serialised_models.py
```diff
@@ -78,10 +78,10 @@ class SerialisedService(SerialisedModel):
...
-        # TODO: add these fields when data migrated. check how these interact with hybrid properties on model.
-        # "custom_email_sender_name",
-        # "email_sender_local_part",
+        "custom_email_sender_name",
+        "email_sender_local_part",
```
quis, line 84: I wonder if it’s worth adding these in a separate PR first and then clearing the cache

app/serialised_models.py
leohemsted, line 82: i dont know if we need this at all - the only use of this now is to be a more robust uniqueness constraint than just a column default, and to set `email_sender_local_part` to when `custom_email_sender_name` is cleared, but that could be calculated on the spot

this could probs be entirely replaced by a custom check constraint that strips whitespace, lowercases, and does a few tweaks to normalise dashes, quotes, etc.

app/models.py
```diff
@@ -633,9 +623,6 @@ def from_json(cls, data):
         # validate json with marshmallow
         fields = data.copy()
 
-        # TODO: remove this line once admin stops passing through normalised_service_name:
-        fields.pop("normalised_service_name", None)
```
samuelhwilliams, line 637: Have we cleared all the service caches since this happened? I don't think anything passes it through even if it's in the cache, but just in case.

leohemsted: yup, i cleared all the caches at the end of last wk 👍

tests/app/service/test_archived_service.py
```diff
@@ -57,7 +57,7 @@ def test_deactivating_service_changes_name_and_email(client, sample_service):
     archived_service = dao_fetch_service_by_id(sample_service.id)
 
     assert archived_service.name == "_archived_2018-07-07_Sample service"
-    assert archived_service.normalised_service_name == "archived20180707sample.service"
+    assert archived_service.email_sender_local_part == "archived20180707sample.service"
```
samuelhwilliams, line 60: ew i don't like this 🙃

leohemsted: i hadn't really thought about this. the _motivation_ for changing `_normalised_service_name` is to make sure the old service name doesn't conflict with the old service name now that it's inactive.

it's a bit weird that this'll get updated if they use their service name, but will get left untouched if they have a custom email sender defined.

couple of options (there may be other things we could do)

* change _normalised_service_name by accessing the column directly and bypassing the property - that way `email_sender_local_part` would be unchanged
* accept this as a quirk. we'd need to make sure that if/when we unarchive, the script we use correctly restores `email_sender_local_part` (which it will do automatically if it uses the `service.name` property to unarchive the service)

samuelhwilliams: > change _normalised_service_name by accessing the column directly and bypassing the property - that way email_sender_local_part would be unchanged

I think this is sort of OK.

I'd also be pro us thinking about just not changing the service name/various attributes other than `state` at all when archiving. It feels like a weird mutation for us to do 🤔 There are probably UI considerations though if we remove that semi-obvious marker. We should probably have an "archived services" org view for example, rather than just listing them in the live/trial services view.

samuelhwilliams: (can be followed-up separately)

samuelhwilliams (approved):
There's one related todo left that I see, but I assume you'll follow up on that separately.

```
    # TODO: once data is migrated this should be not nullable
    _email_sender_local_part = db.Column("email_sender_local_part", db.String(255), nullable=True)
```

This looks good 👍

leohemsted:
ah good spot - that should have gone in the prev PR (with the six migrations) 👍

PR #3957, opened 2023-11-21
use custom email sender name if provided

leohemsted:
the point of this whole piece of work :facepalm:


quis (approved):
Checked and we have a test for the non-custom case here:
https://github.com/alphagov/notifications-api/blob/3b686ee3958c2eae8309bcb827f4ffef1dfacca1/tests/app/delivery/test_send_to_providers.py#L162

PR #3959, opened 2023-11-22
commands: add commands for updating numeric scale for large tables

risicle:
Replaces #3952, an attempt to do this in migrations.

These commands update the numerics in `Notification`, `NotificationHistory` and `FactBilling` to their `min_scale` following the DMS migration.

For `FactBilling` and `Notification`, performs updates in blocks (partitioned on a pseudorandom uuid) to avoid locking any whole tables (or even one specific time-period) at once.

For `NotificationHistory`, partition the blocks on `created_at` as that is more closely correlated to the physical layout of the table, and accesses to it (outside the periodic update tasks) aren't particularly clustered around any one point in time that we should avoid locking.

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -6,6 +6,7 @@ import os
 import random
 import uuid
 from datetime import date, datetime, timedelta
+from itertools import accumulate, repeat
 from time import monotonic
 from unittest import mock
 
@@ -17,6 +18,13 @@ from flask import current_app, json
 from notifications_utils.recipients import RecipientCSV
 from notifications_utils.statsd_decorators import statsd
 from notifications_utils.template import SMSMessageTemplate
+from sqlalchemy import (
+    Numeric,
+    case,
+    cast,
+    func,
+    select,
+)
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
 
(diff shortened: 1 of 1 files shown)


app/commands.py
```diff
@@ -818,6 +829,115 @@ def local_dev_broadcast_permissions(user_id):
...
+                {
+                    "rate": _get_min_scale_cases(FactBilling.rate),
+                }
+            )
+
+
+@click.option("-h", "--block-hours", type=float, default=1.0)
+@notify_command(name="update-notification-history-numerics-min-scale")
+def update_notification_history_numerics_min_scale(block_hours):
+    block_period = timedelta(microseconds=block_hours * 60 * 60 * 1e3 * 1e3)
```
samuelhwilliams, line 900:
```suggestion
    block_period = timedelta(hours=block_hours)
```

?

risicle: `hours=` needs to be an integer, no?

risicle: ...no they do not!

TIL

samuelhwilliams (approved):
These look good to me 👍  All work as expected locally and I think the strategies for grouping are sensible.

samuelhwilliams:
You _may_ want to install the `pre-commit` hooks so you don't get CI test failures for formatting things.

PR #3963, opened 2023-11-27
Revert "commands: add commands for updating numeric scale for large tables"

risicle:

This was too slow to be worth it, so let's not bother and save ourselves fragmenting the tables in the process.


samuelhwilliams (approved):

samuelhwilliams:
🚢

PR #3966, opened 2023-11-30
migrations: use advisory lock to prevent concurrent migration processes

risicle:
https://trello.com/c/35GEi5aB/548-remove-risk-of-api-code-being-deployed-to-production-before-the-alembic-migrations-have-been-applied

See inline comment for more detail.

The `lock_timeout` setting needs to be moved to after this lock is taken as we'd rather this didn't apply to the advisory lock itself - there's no harm in making a migration process that hasn't taken any other locks out yet wait.

I've tested this by deliberately spawning a duplicate migration container running a dummy slow migration (`sleep(60)`) and ensuring that the two attempts queue up after each other and the second one doesn't find it needs to do anything.

Any other way of testing anyone can think of?

### DIFF
--- a/migrations/env.py
+++ b/migrations/env.py
@@ -1,5 +1,6 @@
 from collections.abc import Callable, Iterable, Mapping
 import psycopg2
+import struct
 import time
 from contextlib import contextmanager
 from logging.config import fileConfig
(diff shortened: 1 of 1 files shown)


migrations/env.py
```diff
@@ -68,28 +69,38 @@ def run_migrations_online():
...
+        # big-endian representation of the string "alembic"
+        lock_id = struct.unpack(">q", struct.pack("8s", b"alembic"))[0]
+        connection.execute("SELECT pg_advisory_lock(%s)", lock_id)
+
+        # abort any migrations if a lock (other than the above advisory lock)
+        # cannot be acquired after one second.
+        #
+        # if we see issues with this lock timeout failing, we should try running
+        # again when there are no locks on that table, perhaps at a quieter time.
+        connection.execute("SET lock_timeout = 1000")
```
idavidmcdonald, line 102: As per https://www.postgresql.org/docs/15/sql-set.html, the default for SET is that it applies for the session only. I wanted to make sure that it wasn't going to apply indefinitely and it therefore doesn't.

migrations/env.py
```diff
@@ -68,28 +69,38 @@ def run_migrations_online():
...
+        )
+
+        # take a *session-level* advisory lock to prevent multiple migration
+        # processes attempting to run concurrently. being a session-level lock,
+        # it should safely cover our few migrations that need to perform multiple
+        # transactions.
+        # advisory lock ids are 64b (signed) integers, so use the null-padded,
+        # big-endian representation of the string "alembic"
+        lock_id = struct.unpack(">q", struct.pack("8s", b"alembic"))[0]
+        connection.execute("SELECT pg_advisory_lock(%s)", lock_id)
```
idavidmcdonald, line 95: Cool, I wasn't aware of pg_advisory_locks. Kind of interesting. Seems a reasonable way of solving this problem.

idavidmcdonald (approved):
Nice. Seems like a good solution.

When this is merged, it will enable us to run the migration service in ECS as well as PaaS, as we know they won't be able to clash.

As part of your ticket, are you also able to get that set up please? I think in ECS it already runs in preview, I'm not sure about staging, and it doesn't run in production. That is probably configured in the API concourse pipeline itself. Let me know if you have any questions and I can give more context

migrations/env.py
risicle: Nope, it's not persistent. The connection (and its session) will be dropped when the flask command that runs the migrations terminates.

PR #3970, opened 2023-12-06
Remove dependence on Emergency Alerts code in `notifications_utils`

quis:
So we can remove the emergency-alerts-related code from utils without having to do the bigger job of removing emergency alerts from the API codebase entirely.


leohemsted (changes requested):
we'll need to remove from this old migration as well

https://github.com/alphagov/notifications-api/blob/6899d627932d824f1d1f2b24d1f01e876e1e8770/migrations/versions/0336_broadcast_msg_content_2.py#L10

leohemsted (approved):

PR #3971, opened 2023-12-08
run the letter still pending virus scan check every ten minutes

leohemsted:
currently we run every hour, and wait for a letter to be stuck for 90 minutes before we kick it off again.

We don't think this is necessary, and is causing issues where, for example, a letter passed virus scan and was sent to template preview for sanitising at 4:30pm. the template preview task crashed, and we didn't pick up the letter to scan/sanitise it again until 7pm, causing it to miss the cutoff.

it isn't the end of the world if a letter is processed/scanned twice as this should be relatively idempotent, so we can just check every ten minutes if something has been stuck for ten minutes

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -788,11 +788,11 @@ def letters_missing_from_sending_bucket(seconds_to_subtract):
 
 
 def dao_precompiled_letters_still_pending_virus_check():
-    ninety_minutes_ago = datetime.utcnow() - timedelta(seconds=5400)
+    ten_minutes_ago = datetime.utcnow() - timedelta(seconds=600)
 
     notifications = (
         Notification.query.filter(
-            Notification.created_at < ninety_minutes_ago, Notification.status == NOTIFICATION_PENDING_VIRUS_CHECK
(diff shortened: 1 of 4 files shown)


app/dao/notifications_dao.py
```diff
@@ -788,7 +788,7 @@ def letters_missing_from_sending_bucket(seconds_to_subtract):
 
 
 def dao_precompiled_letters_still_pending_virus_check():
-    ninety_minutes_ago = datetime.utcnow() - timedelta(seconds=5400)
+    ninety_minutes_ago = datetime.utcnow() - timedelta(seconds=600)
```
idavidmcdonald, line 791: variable name will need updating

leohemsted: yeah i saw that, will update the content. have we ever had this zendesk ticket get triggered? i cant find any example of it in my inbox

klssmith: It used to get triggered because that's all the task did - open a ticket. But I'm not sure it does now that it only gets opened if the letter can't be automatically fixed

leohemsted: @klssmith / @idavidmcdonald i've updated variable names, tests, etc

klssmith (approved):

PR #3976, opened 2023-12-14
Remove emergency alerts code

leohemsted:
This doesn't deal with any database migration, but it does get rid of all code.

TODO before we merge:

- [ ] check nothing in admin app will break
- [ ] check nothing with redis will break


quis:
At the moment if you try to view a broadcast template in the admin app you will get an exception (the app doesn’t know how to preview one of these templates).

But the only way you can get to that page is by manually entering the `id` of the template in the URL. You can’t see a list of those templates on the templates page for example. I don’t think this PR will change that behaviour, so I don’t think handling it needs to be a blocker to merging.

We could put in a workaround to return a `404`. Or we could do a migration later on to delete all `broadcast` templates.

idavidmcdonald:
FYI, we can't delete any of the broadcast data out of our database for 1 year please. Either that or we need to take a snapshot of that data and store it elsewhere for a year. This is because emergency alerts need it as an audit trail for any security incidents (precedent is to store it for a year, just like Notify does).

The 1 year is from the time they migrated to their own system (was that like early November maybe?)

leohemsted:
thanks, that's useful to know david. I'd probably be keen on snapshotting and storing it in a pgdump in s3 ideally

klssmith:
This has been checked and rebased to get rid of merge conflicts. [The data has been deleted](https://github.com/alphagov/notifications-api/pull/4516), so the comments relating to that no longer apply.

The failure of the `migration-isolation-check` ("Please separate migration changes from other code changes into separate PRs. This helps ensure safer deployments and easier rollbacks if needed") is fine - we're only removing a couple of unused imports from the migration files, not making any database changes.

PR #3977, opened 2023-12-15
Email selected team members when someone asks to join a service

BlessedDev:
This PR adds the endpoint to support the work being done in this [PR](https://github.com/alphagov/notifications-admin/pull/4912), as described in [this](https://trello.com/c/rYMOdCYZ/578-email-selected-team-members-when-someone-asks-to-join-a-service) card.

An alteration has been made to the expected JSON content, `from_user_ids ` has now been replaced with `service_managers_ids` which is a better description.
Also the new endpoint will only return a status code.

Also for now all request invite emails are being sent to `notify-join-service-request@digital.cabinet-office.gov.uk` as prescribed in the ticket, instead of service managers.

### DIFF
--- a/app/service_invite/rest.py
+++ b/app/service_invite/rest.py
@@ -137,17 +137,14 @@ def request_user_invite(service_id, user_to_invite_id):
     invite_link_host = request_json["invite_link_host"]
     accept_invite_request_url = f"{invite_link_host}/services/{service.id}/users/invite/{user_requesting_invite.id}"
 
-    # # Ensure that the user making the request is already not part of the service
     if user_requesting_invite.services and service in user_requesting_invite.services:
-        message = f"You are already a member of {service.name}"
+        message = "user-already-in-service"
         raise BadRequestError(message=message)
 
-    # Send the user's service invite request to the service managers listed
     send_service_invite_request(
         user_requesting_invite, recipients_of_invite_request, service, reason_for_request, accept_invite_request_url
     )
 
-    # Send a receipt email to the user that requested the invite
     send_receipt_after_sending_request_invite_letter(user_requesting_invite)
 
     return {}, 204
@@ -165,7 +162,6 @@ def send_service_invite_request(
             saved_notification = persist_notification(
                 template_id=template.id,
                 template_version=template.version,
-                # TODO change recipient to actual email address of service managers when the testing phase completes
                 recipient="notify-join-service-request@digital.cabinet-office.gov.uk",
                 service=notify_service,
                 personalisation={
@@ -189,12 +185,12 @@ def send_service_invite_request(
             # In a scenario were multiple service managers are listed, and the list contains an
             # invalid service manager, we would rather log the errors and not raise an exception so
             # that notifications can still be sent to the valid service managers
-            message = f"Can’t create notification - {recipient.name} is not part of the {service.name}"
+            message = "not-a-valid-service-manager"
             current_app.logger.error(message)
 
     if number_of_notifications_generated == 0:
         # If no notification is sent we want to raise an exception
-        message = f"Can’t create notification as the service manager listed is not part of the {service.name}"
+        message = "no-valid-service-managers"
         raise BadRequestError(message=message)
 
 
@@ -206,7 +202,6 @@ def send_receipt_after_sending_request_invite_letter(user_requesting_invite):
     saved_notification = persist_notification(
         template_id=template.id,
         template_version=template.version,
-        # TODO change recipient to actual email address of service managers when the testing phase completes
         recipient=user_requesting_invite.email_address,
         service=notify_service,
         personalisation={"name": user_requesting_invite.name},
--- a/app/service_invite/rest.py
+++ b/app/service_invite/rest.py
@@ -20,6 +20,7 @@ from app.notifications.process_notifications import (
     send_notification_to_queue,
 )
 from app.schemas import invited_user_schema
+from app.v2.errors import BadRequestError
 
 service_invite = Blueprint("service_invite", __name__)
 
@@ -102,20 +103,6 @@ def get_invited_user(invited_user_id):
     return jsonify(data=invited_user_schema.dump(invited_user)), 200
 
 
-@service_invite.route("/service/<uuid:service_id>/invite/request-for/<uuid:user_to_invite_id>", methods=["POST"])
-def request_user_invite(service_id, user_to_invite_id):
-    request_json = request.get_json()
-    get_user_by_id(user_to_invite_id)
-    Service.query.get(current_app.config["NOTIFY_SERVICE_ID"])
-    dao_fetch_service_by_id(service_id)
-    data = {
-        "from_user_ids": request_json[""],
-        "reason": request_json["reason"],
-        "invite_list_host": request_json["oops"],
-    }
-    return jsonify(data=data), 200
-
-
 @service_invite.route("/invite/service/<token>", methods=["GET"])
 @service_invite.route("/invite/service/check/<token>", methods=["GET"])
 def validate_service_invitation_token(token):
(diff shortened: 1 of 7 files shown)


app/service_invite/rest.py
```diff
@@ -121,3 +124,95 @@ def validate_service_invitation_token(token):
...
+def send_receipt_after_sending_request_invite_letter(user_requesting_invite):
+    template_id = current_app.config["RECEIPT_FOR_REQUEST_INVITE_TO_SERVICE_TEMPLATE_ID"]
+    template = dao_get_template_by_id(template_id)
+    notify_service = Service.query.get(current_app.config["NOTIFY_SERVICE_ID"])
+
+    saved_notification = persist_notification(
+        template_id=template.id,
+        template_version=template.version,
+        # TODO change recipient to actual email address of service managers when the testing phase completes
+        recipient="notify-join-service-request@digital.cabinet-office.gov.uk",
```
quis, line 210: This email should go to the user for real, not the shared inbox

quis (changes requested):

app/service_invite/rest.py
```diff
@@ -121,3 +124,95 @@ def validate_service_invitation_token(token):
...
+    recipients_of_invite_request_ids = request_json["service_managers_ids"]
+    recipients_of_invite_request = [get_user_by_id(recipient_id) for recipient_id in recipients_of_invite_request_ids]
+    service = dao_fetch_service_by_id(service_id)
+    reason_for_request = request_json["reason"]
+    invite_link_host = request_json["invite_link_host"]
+    accept_invite_request_url = f"{invite_link_host}/services/{service.id}/users/invite/{user_requesting_invite.id}"
+
+    # # Ensure that the user making the request is already not part of the service
+    if user_requesting_invite.services and service in user_requesting_invite.services:
+        message = f"You are already a member of {service.name}"
```
samuelhwilliams, line 142: I think we should prefer to return a code here rather than a 'nice' bit of text. Since this is an internal API endpoint, we can keep the user-facing content in the admin app, which will make updating it easier.

I'm aware there's probably not many examples of this in the API right now, unfortunately, so not surprised you went for this. But a code would be better. 🙏 

```suggestion
        message = "user-already-in-service"
```

app/service_invite/rest.py
```diff
@@ -121,3 +124,95 @@ def validate_service_invitation_token(token):
...
+        raise BadRequestError(message=message)
+
+    # Send the user's service invite request to the service managers listed
+    send_service_invite_request(
+        user_requesting_invite, recipients_of_invite_request, service, reason_for_request, accept_invite_request_url
+    )
+
+    # Send a receipt email to the user that requested the invite
```
samuelhwilliams, line 150: [nit]: I don't think we need these comments because you've given the functions nice sensible names :)

app/service_invite/rest.py
```diff
@@ -121,3 +124,95 @@ def validate_service_invitation_token(token):
...
+@service_invite.route("/service/<service_id>/invite/request-for/<user_to_invite_id>", methods=["POST"])
+def request_user_invite(service_id, user_to_invite_id):
+    request_json = request.get_json()
+    user_requesting_invite = get_user_by_id(user_to_invite_id)
+    recipients_of_invite_request_ids = request_json["service_managers_ids"]
+    recipients_of_invite_request = [get_user_by_id(recipient_id) for recipient_id in recipients_of_invite_request_ids]
+    service = dao_fetch_service_by_id(service_id)
+    reason_for_request = request_json["reason"]
+    invite_link_host = request_json["invite_link_host"]
+    accept_invite_request_url = f"{invite_link_host}/services/{service.id}/users/invite/{user_requesting_invite.id}"
```
samuelhwilliams, line 138: I can't remember if we/Leo talked about this URL being sent in the request body rather than the API building it? 🤔

app/service_invite/rest.py
```diff
@@ -121,3 +124,95 @@ def validate_service_invitation_token(token):
...
+        else:
+            # In a scenario were multiple service managers are listed, and the list contains an
+            # invalid service manager, we would rather log the errors and not raise an exception so
+            # that notifications can still be sent to the valid service managers
+            message = f"Can’t create notification - {recipient.name} is not part of the {service.name}"
+            current_app.logger.error(message)
+
+    if number_of_notifications_generated == 0:
+        # If no notification is sent we want to raise an exception
+        message = f"Can’t create notification as the service manager listed is not part of the {service.name}"
```
samuelhwilliams, line 197: Again, I think let's come up with an error code for this and use it instead. Maybe something like `no-valid-service-managers`?

app/service_invite/rest.py
```diff
@@ -121,3 +124,95 @@ def validate_service_invitation_token(token):
...
+
+def send_receipt_after_sending_request_invite_letter(user_requesting_invite):
+    template_id = current_app.config["RECEIPT_FOR_REQUEST_INVITE_TO_SERVICE_TEMPLATE_ID"]
+    template = dao_get_template_by_id(template_id)
+    notify_service = Service.query.get(current_app.config["NOTIFY_SERVICE_ID"])
+
+    saved_notification = persist_notification(
+        template_id=template.id,
+        template_version=template.version,
+        # TODO change recipient to actual email address of service managers when the testing phase completes
```
samuelhwilliams, line 209: This comment can be removed now :)

```suggestion
```

app/service_invite/rest.py
BlessedDev: I take your point on this and will adjust it accordingly.

app/service_invite/rest.py
BlessedDev: I erred on the side of leaving in the comments for the review, I will take them out.

app/service_invite/rest.py
```diff
@@ -189,12 +185,12 @@ def send_service_invite_request(
             # In a scenario were multiple service managers are listed, and the list contains an
             # invalid service manager, we would rather log the errors and not raise an exception so
             # that notifications can still be sent to the valid service managers
-            message = f"Can’t create notification - {recipient.name} is not part of the {service.name}"
+            message = "not-a-valid-service-manager"
             current_app.logger.error(message)
```
samuelhwilliams, line 189: Given these are log messages rather than responses, we can keep them information rich so it helps us investigate :)



```suggestion
            current_app.logger.error("Ignoring request-to-join-service email to user %(recipient.id) - they are not part of service %(service.id)", dict(recipient=recipient, service=service))
```

samuelhwilliams (approved):
Nice tests 👍

app/service_invite/rest.py
```diff
@@ -121,3 +124,93 @@ def validate_service_invitation_token(token):
...
+    template = dao_get_template_by_id(template_id)
+    notify_service = Service.query.get(current_app.config["NOTIFY_SERVICE_ID"])
+    number_of_notifications_generated = 0
+    invalid_service_manager_ids = []
+    for recipient in recipients_of_invite_request:
+        if service in recipient.services:
+            saved_notification = persist_notification(
+                template_id=template.id,
+                template_version=template.version,
+                recipient="notify-join-service-request@digital.cabinet-office.gov.uk",
```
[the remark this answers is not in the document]
quis: We’re using a Google group while we do user research with the feature.

migrations/versions/0438_request_invite_templates.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+)
+
+
+receipt_for_request_invite_to_a_service_template_id = "38bcd263-6ce8-431f-979d-8e637c1f0576"
+receipt_for_request_invite_to_a_service_template_content = textwrap.dedent(
+    """\
+    Hi ((name))
+
+    …
```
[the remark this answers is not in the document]
quis: https://docs.google.com/document/d/1arlgMdIQ0RGwDqL_g9WC6unpUyoTTuNzz_2f3m-6E_Q/edit#heading=h.2do9edspgf62

migrations/versions/0438_request_invite_templates.py
```diff
@@ -0,0 +1,167 @@
...
+from flask import current_app
+
+
+revision = "0438_request_invite_templates"
+down_revision = "0437_min_numeric_scl_aux_tbls"
+
+
+request_invite_to_a_service_template_id = "77677459-f862-44ee-96d9-b8cb2323d407"
+request_invite_to_a_service_template_content = textwrap.dedent(
+    """\
```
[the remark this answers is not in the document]
BlessedDev: The slash is there after the first line to to avoid the empty line as suggested in the `textwrap` docs https://docs.python.org/3/library/textwrap.html#:~:text=For%20example%3A-,def,-test()%3A%0A%20%20%20%20%23%20end

app/service_invite/rest.py
```diff
@@ -180,18 +180,21 @@ def send_service_invite_request(
             )
             send_notification_to_queue(saved_notification, queue=QueueNames.NOTIFY)
             number_of_notifications_generated += 1
+            invalid_service_manager_ids.append(recipient.id)
```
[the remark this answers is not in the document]
BlessedDev: Good catch. Thanks, I have taken it out.

migrations/versions/0438_request_invite_templates.py
```diff
@@ -0,0 +1,167 @@
...
+
+GOV.​UK Notify team
+https://www.gov.uk/notify
+    """
+)
+
+
+receipt_for_request_invite_to_a_service_template_id = "38bcd263-6ce8-431f-979d-8e637c1f0576"
+receipt_for_request_invite_to_a_service_template_content = textwrap.dedent(
+    """\
```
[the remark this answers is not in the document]
BlessedDev: The slash is there after the first line to to avoid the empty line as suggested in the` textwrap` docs https://docs.python.org/3/library/textwrap.html#:~:text=For%20example%3A-,def,-test()%3A%0A%20%20%20%20%23%20end

BlessedDev:
> Nice tests 👍

Thank you.

PR #3980, opened 2023-12-18
WIP send a file with filenames

samuelhwilliams:


app/v2/notifications/notification_schemas.py
```diff
@@ -133,7 +133,23 @@
...
     },
+    "allOf": [
+        {
+            "anyOf": [
+                {"not": {"required": ["is_csv"]}},
+                {"not": {"required": ["filename"]}},
+            ],
+            "validationMessage": (
+                "`filename` and `is_csv` are exclusive. "
+                "If you want to set `filename` explicitly, do not send a value for `is_csv`."
```
samuelhwilliams, line 149: @karlchillmaid  there's also this error message 🙃

idavidmcdonald (commented):
So generally looks good

3 things to think about validation (mostly due to my lack of recent experience with how the hell we do validation, in particular for document download).

1. A user could provide a filename of `.`, is that OK?

2. I see on the document download API, you've got a pull request up to restrict filenames to max 100 characters. What happens if someone makes a request to the Notify API for 101 characters. Will the Notify API expose the document download API error back to the user?

3. Do we need to be more restrictive generally about characters that are valid? https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names says that some characters are invalid for windows for example, meaning that if someone put a filename of 'hello?sam' then they may run into some sort of issue when downloading (I'm not sure what that could hypothetically be if anything)?

app/v2/notifications/notification_schemas.py
```diff
@@ -123,7 +123,7 @@
         },
         "is_csv": {
             "format": "send_a_file_is_csv",
-            "required": False,
+            # "required": False,
```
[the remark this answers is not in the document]
samuelhwilliams: Accidental :)

samuelhwilliams:
> A user could provide a filename of ., is that OK?

Here, yes. It will get rejected by document-download-api and surface an appropriate error message. There's a weird bit of contention here around what should own the error messages, as they're definitely split/duplicated across both APIs. I hope to do a broad cleanup on the doc-dl-api after this is all merged to make it just return error codes, and shift user-facing error messages to the responsibility of this API wholly.

> I see on the document download API, you've got a pull request up to restrict filenames to max 100 characters. What happens if someone makes a request to the Notify API for 101 characters. Will the Notify API expose the document download API error back to the user?
Yes.

app/v2/notifications/notification_schemas.py
```diff
@@ -123,7 +123,6 @@
         },
         "is_csv": {
             "format": "send_a_file_is_csv",
-            "required": False,
```
idavidmcdonald, line 126: How come we want to remove this? We have required False on the other ones? Maybe we should be keeping this and also adding required False to `filename` too?

idavidmcdonald (approved):
Minor bit to confirm around `"required": False` but should be good when sorted.

Happy for the interesting validation to move into the document download API as you suggest.

app/v2/notifications/notification_schemas.py
samuelhwilliams: `required: False` is the default. There are required statements further down now for both `is_csv` and `filename`, so I thought having them just there might be clearer. Otherwise they say `required: false` here, but then further down there's another (inverted) `require` on them.

I think having `required: false` here looks like it's doing something independent, but the rules around this parameter are now tightly bound to the rules around `filename`, so I thought co-locating them might be clearer...

JSON Schema requireds are a bit weird though so I'm not sure what's best. Happy to put them back here.

idavidmcdonald: thanks for explaining. Pros and cons either way. Pretty minor impact either way.

Your choice! I'm happy either way.

PR #3985, opened 2023-12-20
Add API key name to notifications serialisation for CSV

CrystalPea:
So we can put that field in the report our users download.


idavidmcdonald (approved):
I've run this locally, with your corresponding admin pull request and downloaded a report successfully. It had both cases (sent using the API and not via the API) and both came out as expected.

As you hadn't mentioned it explicitly, I assume you will need to deploy the API out to prod before you can deploy the admin change out?

PR #3986, opened 2023-12-21
Fix delivery callbacks sometimes updating complaints callbacks instead

klssmith:
We have a bug where editing the delivery status callback for a service
can actually overwrite the complaint callback instead. We don't have a
proper UI for changing or adding complaing callbacks - there is only one
team that has this set up.

The `get_service_callback_api` function wasn't filtering by callback
type, so could return the delivery status callback or could return the
complaint callback, leading to the wrong type being updated. This adds
an extra filter to the function so that we have to specify the type of
callback we are updating.

### DIFF
--- a/app/dao/service_callback_api_dao.py
+++ b/app/dao/service_callback_api_dao.py
@@ -27,8 +27,10 @@ def reset_service_callback_api(service_callback_api, updated_by_id, url=None, be
     db.session.add(service_callback_api)
 
 
-def get_service_callback_api(service_callback_api_id, service_id):
-    return ServiceCallbackApi.query.filter_by(id=service_callback_api_id, service_id=service_id).first()
+def get_service_callback_api(service_callback_api_id, service_id, callback_type):
+    return ServiceCallbackApi.query.filter_by(
+        id=service_callback_api_id, service_id=service_id, callback_type=callback_type
+    ).first()
 
 
 def get_service_delivery_status_callback_api_for_service(service_id):
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -229,6 +229,14 @@ class ServiceSchema(BaseSchema, UUIDsAsStringsMixin):
     custom_email_sender_name = fields.String(allow_none=True)
     # this can only be set via custom_email_sender_name or name
     email_sender_local_part = fields.String(dump_only=True)
+    service_callback_api = fields.Method("service_delivery_status_callback_api")
+
+    def service_delivery_status_callback_api(self, service):
+        return [
+            callback.id
+            for callback in service.service_callback_api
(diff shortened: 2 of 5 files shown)


samuelhwilliams (commented):
When testing this locally I've been getting 500s on `fetch_service_callback_api` because it can't find an instance. This is because admin is requesting with the ID of the complaint callback, which then doesn't match the callback type we're filtering on.

I think we need to also exclude complaint callbacks from being returned on the service `service_callback_api` field as well (and then clear the service cache, but I don't think we need to worry about splitting this PR out given it's only one service with a complaint callback - we can just deploy and clear quickly).

migrations/versions/0439_callback_type_non_nullable.py
```diff
@@ -0,0 +1,23 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "0439_callback_type_non_nullable"
+down_revision = "0438_request_invite_templates"
+
+
+def upgrade():
+    op.alter_column("service_callback_api", "callback_type", existing_type=sa.VARCHAR(), nullable=False)
```
idavidmcdonald, line 17: Am I right in thinking that this is doing `ALTER TABLE service_callback_api ALTER COLUMN callback_type SET NOT NULL`? Do correct me if wrong though!

If so, this would be covered in https://medium.com/paypal-tech/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#09eb

Which says `adding a not-null constraint to an existing column requires an exclusive lock on the table while a full table scan verifies that no null values exist`. 

Taking a exclusive table lock sounds bad on that table (may not be able to create service callbacks which technically is part of the processing of delivery receipts tasks).

Worth confirming if this is the case and looking into the alternative approach mentioned in the link above for how to do this migration without taking an exclusive table lock.

klssmith:
@samuelhwilliams I've added a commit which makes sure we only return the delivery status callback instead of both types

migrations/versions/0439_callback_type_non_nullable.py
klssmith: Yes, it does do `ALTER TABLE service_callback_api ALTER COLUMN callback_type SET NOT NULL`. The tables affected are small (433 rows in `service_callback_api` and 1700 in `service_callback_api_history`) but I could just leave the migration out or do it separately in stages if you think it would be better not to do that - it's not really required, it's more of a "nice to have".

klssmith: @idavidmcdonald I've dropped the migration from this PR

klssmith:
I think this should be ready for another look now @samuelhwilliams and @idavidmcdonald

samuelhwilliams (approved):
Nice 👍

PR #3988, opened 2023-12-27
Remove dvla ftp code

leohemsted:
note that we still have response files coming via ftp so a find in files might still show some usages of the word


idavidmcdonald (approved):
You've got a failing unit test but apart from that this looks reasonable

PR #3989, opened 2023-12-29
Remove a bunch of broadcasts code

samuelhwilliams:
Leaves the DB models in place, to be cleaned up separately.


leohemsted (commented):
we can remove app/xml_schemas/CAP-v1.2.xsd too

leohemsted (approved):
did some basic ctrl+f for obvious phrases and seems good

PR #3991, opened 2023-12-30
Use api image for tests

robbrazil:
Dockerfile
Create Concourse Test Image
A new layer has been created to run the API tests using its own image. 
The problem before was that the API did not have POSTGRES in its build.  
The new layer, called CONCOURSE_TESTS builds this extra layer based in the TEST layer and prepares the image to run all the tests in concourse.

Instructions to build and run the tests locally have been also amended in the README.

API can now run all the tests using postgres installed in the new layer.


c4seyla (approved):

README.md
```diff
@@ -49,15 +49,28 @@ notify-pass credentials/mmg
...
+2. Run the container:
+make run-local-db-with-docker
+
+3. Exec into the container:
+docker exec -it 0f7f2e4718f9 /bin/bash (change container ID)
+
+4. Once inside the container, run:
+service postgresql start
+make bootstrap
+make test
```
samuelhwilliams, line 70: I think we should be able to improve these docs, especially line 65 where there's just some random container ID.

If the run scripts sets a container name, then you can replace the container ID with the container name and have a command that doesn't need to be edited on every use. At least,

Makefile
```diff
@@ -42,6 +42,10 @@ run-flask: ## Run flask
 run-flask-with-docker: ## Run flask
 	./scripts/run_locally_with_docker.sh api-local
 
+.PHONY: run-local-db-with-docker
+run-local-db-with-docker: ##
+	./scripts/run_local_db_with_docker.sh api-local
```
samuelhwilliams, line 47: Do we need a whole new script or can we just add a new target to the `run_locally_with_docker` script, eg `api-local-db`?

docker/Dockerfile
```diff
@@ -83,3 +83,43 @@ COPY --chown=notify:notify Makefile requirements.txt requirements_for_test.txt .
...
+    && apt-get install -y --no-install-recommends \
+        postgresql-$POSTGRES_VERSION \
+    && rm -rf /var/lib/apt/lists/* /tmp/*
+
+RUN \
+    service postgresql start \
+    && su - postgres -c "psql -c \"alter user postgres with password 'postgres';\"" \
+    && su - postgres -c "psql -c \"create database notification_api;\"" 
+
+ENV SQLALCHEMY_DATABASE_URI=postgresql://postgres:postgres@localhost/notification_api
```
samuelhwilliams, line 125: Small one but there's no line ending on this file 😢 

If you can, installing the `pre-commit` hooks would be nice (see the README in each repo, but it's essentially installing the command running `pre-commit install --install-hooks` in each repo).

robbrazil: done, thanks

Makefile
robbrazil: @samuelhwilliams i have made smallish changes to the run_locally_with_docker script. Please see if these are acceptable, in that case, we can run with just that one script

README.md
robbrazil: I re-wrote it as I made a new make command to encompass all the steps needed to run the tests locally. Please review it

Makefile
```diff
@@ -42,6 +42,14 @@ run-flask: ## Run flask
 run-flask-with-docker: ## Run flask
 	./scripts/run_locally_with_docker.sh api-local
 
+.PHONY: run-local-tests-with-docker
+run-local-tests-with-docker:
+	docker build -f docker/Dockerfile --target concourse_tests -t notifications-api .
```
samuelhwilliams, line 47: This will force a slow rebuild every time the command is run. In Concourse that's probably fine; locally it's potentially annoying.

However I don't think we should be running this locally, so it's also probably fine to ignore.

Makefile
```diff
@@ -42,6 +42,14 @@ run-flask: ## Run flask
...
+	./scripts/run_locally_with_docker.sh api-local
+	docker exec -it notifications_api /bin/bash -c "service postgresql start"
+	docker exec -it notifications_api make test
```
samuelhwilliams, line 50: I think there's theoretically a race condition here between the DB coming online and the tests running. I don't hit it locally and I think the risk of hitting it is fairly small given our slow-ish unit test startup times.

Maybe just one to be aware of and address if we start seeing issues.

README.md
```diff
@@ -49,15 +49,19 @@ notify-pass credentials/mmg
...
 
-The API (on Notify) works with PostgreSQL 15. After installation, open the Postgres app, open the sidebar, and update or replace the default server with a compatible version.
-
-**Note:** you may need to add the following directory to your PATH in order to bootstrap the app.
+#### How to build and run tests locally:
 
 ```
-export PATH=${PATH}:/Applications/Postgres.app/Contents/Versions/15/bin/
+1. run the make command
+    make run-local-tests-with-docker
```
samuelhwilliams, line 59:
```suggestion
1. run the make command
    `make run-local-tests-with-docker`
```

README.md
```diff
@@ -49,15 +49,19 @@ notify-pass credentials/mmg
 
 ### Postgres
 
-Install [Postgres.app](http://postgresapp.com/).
+Postgres is built in the CONCOURSE_TESTS as an extra layer derived from the TEST layer.
+This is built to allow the tests to run using the API PRODUCTION layer.
```
samuelhwilliams, line 53: I don't know if this comment change is that helpful.

It used to highlight the fact that this app has a dependency on Postgres which needs to be managed manually. Now that context is deleted and replaced with a comment about a very specific setup for testing via Concourse (which I think is the main thing we're talking about in this PR?).

```suggestion
This app requires Postgres to run.

If you are using [notifications-local](https://github.com/alphagov/notifications-local), the correct Postgres version will be provided automatically by the docker-compose file.

If you are running this app manually, you should will need to manage Postgres yourself. Install [Postgres.app](http://postgresapp.com/). Check the docker-compose file above to find the correct Postgres version to use.

When running our test suite, Postgres is baked into the test image under the `concourse_tests` step of `docker/Dockerfile`.
```

Makefile
robbrazil: you do not need to run make...if building locally. This was done to simplify the tests locally.

Makefile
robbrazil: maybe, but i tried a couple (of a hundred) times....LOL no race condition has happened. we could add a sleep if desired.

README.md
robbrazil: postgres is built within the image as an extra layer. Running locally or in concourse, postgres will be present in the "concourse_tests" layer. 
There is no need to manually install postgres either way.
Not sure I understand your comments.

Makefile
robbrazil: Please have a review of the original task on https://trello.com/c/SRDI5dT8/576-get-api-unit-tests-to-run-in-the-api-docker-image-not-the-generic-unit-test-image-on-concourse

README.md
robbrazil: please have a look at the original task:
https://trello.com/c/SRDI5dT8/576-get-api-unit-tests-to-run-in-the-api-docker-image-not-the-generic-unit-test-image-on-concourse

samuelhwilliams: Yes, but you're thinking about testing only.

The readme is documenting the general app altogether - ie what you need to run the app, not what you need to _test_ the app.

The readme currently says that if you want to _run_ the app, you need to install Postgres manually. That's still true so I don't think we should change the readme to only talk about the testing side of things.

README.md
```diff
@@ -49,15 +49,19 @@ notify-pass credentials/mmg
...
+#### How to build and run tests locally:
 
 ```
-export PATH=${PATH}:/Applications/Postgres.app/Contents/Versions/15/bin/
+1. run the make command
+    `make run-local-tests-with-docker`
 ```
+The step above will build the image, run the container, start postgresql and run the tests.
+
+Note: There is a file called "version.py.dist" that is used for "healthcheck.py". As it stands, this file is unusable and the local build will fail. You must, either rename it to "version.py" or simply create a local file with that name and copy the contents to it.
```
samuelhwilliams, line 63: This shouldn't really need to be done manually. We have a `make generate-version-file` step that creates this file. This step also automatically runs when `make bootstrap` is run, which is what devs and Concourse use to prepare the app to run or be tested.

I'm also not sure that `version.py.dist` _is_ actually used anywhere. I think it's just a leftover file maybe - can't see any references to it at least...

scripts/run_locally_with_docker.sh
```diff
@@ -21,10 +20,9 @@ else
   EXPOSED_PORTS=""
 fi
 
-docker run -it --rm \
+docker run -it -d --name notifications_api --rm \
```
samuelhwilliams, line 23: I'm not sure this `-d` flag is what most people would expect ... suggests we're doing something a bit unusual here.

Makefile
idavidmcdonald: So ironically, I don't think we should ever be running `make run-local-tests-with-docker` on Concourse (Sam, you referenced an expectation that it would be run in concourse which I don't think should be the plan).

The reason we won't be running this command in Concourse is because Concourse will enter the docker image using bash, and then will then run whichever commands are specified in the concourse job. The image will already be built by this point.

See https://github.com/alphagov/notifications-aws/blob/main/concourse/templates/api.yml.j2#L64-L66 for how we do this already for the API.

So `run-local-tests-with-docker` would only be useful if developers want/need to run that command from their local machines.


I think you have 2 options here.

1. Remove the `run-local-tests-with-docker` command entirely. If we just want to focus on CONCOURSE running the unit tests in a docker image, then we don't need this make command because it would only be used by devs on their local machines. (Yes, it would be nice to put this functionality in place but could always come in a separate PR or later)

2. Continue with the approach of letting devs run the unit tests in a docker image using Make, however we might want to consider a few tweaks e.g.

- if you run the command twice, it passes the first time and then fails the second
- if you run the command once, and then run `make run-flask-with-docker` that has also broken
- I'm not sure if `./scripts/run_locally_with_docker.sh api-local` is the right thing to run. That will leave the app running in the container after the tests have finished

README.md
idavidmcdonald: I agree with Sam here. I think this section should remain about what you need to do to get Postgres so you can run the app locally.

Any additional context about running the unit tests in Concourse isn't too important for the README. If we want to talk about running the unit tests locally in docker, then that would probably be best in the section 'To test the application' which is below.

README.md
idavidmcdonald: Agree that we should remove the need for manual action like this please.

scripts/run_locally_with_docker.sh
idavidmcdonald: Yep, I'd expect this to run in the foreground I think

scripts/run_locally_with_docker.sh
```diff
@@ -21,10 +20,9 @@ else
...
   -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
   -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
-  -e SQLALCHEMY_DATABASE_URI=$SQLALCHEMY_DATABASE_URI \
```
idavidmcdonald, line 27: Are we sure we want to remove this? Worth double checking

idavidmcdonald (commented):
So the dockerfile itself looks good.

However this PR is trying to do 2 things. 

The first is to set up the dockerfile so we can use it on Concourse. 
The second is to get devs to run unit tests in docker in our new docker image.

I think the second is the tricky one and your dockerfile looks good already. I'd strongly suggest splitting this into 2 separate PRs and focussing for the moment on getting the API unit tests running in Concourse as that is the core bit of the ticket. The rest can come as a bonus!

Makefile
samuelhwilliams: Hm, interesting.

I got that expectation from the Dockerfile step being called `concourse_tests`, I suppose, and since this new setup probably makes testing a bit of a painfully slow loop (since it'll rebuild the image every time).

I think focusing this on just what's needed for concourse for now makes sense 👍

scripts/run_locally_with_docker.sh
robbrazil: ok

README.md
robbrazil: ok

Makefile
```diff
@@ -42,6 +42,14 @@ run-flask: ## Run flask
 run-flask-with-docker: ## Run flask
 	./scripts/run_locally_with_docker.sh api-local
 
+.PHONY: run-local-tests-with-docker
+run-local-tests-with-docker:
```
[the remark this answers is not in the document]
robbrazil: ok

scripts/run_locally_with_docker.sh
robbrazil: ok, since we are not using the make command i created, i reverted the script to its original format

robbrazil:
@idavidmcdonald Thanks for your comments. very much appreciated the level of detail and guidance.
I have concetrated this PR exclusivelly in the building of the image.

docker/Dockerfile
```diff
@@ -83,3 +83,43 @@ COPY --chown=notify:notify Makefile requirements.txt requirements_for_test.txt .
...
+    && apt-get install -y --no-install-recommends \
+        postgresql-$POSTGRES_VERSION \
+    && rm -rf /var/lib/apt/lists/* /tmp/*
+
+RUN \
+    service postgresql start \
+    && su - postgres -c "psql -c \"alter user postgres with password 'postgres';\"" \
+    && su - postgres -c "psql -c \"create database notification_api;\""
+
+ENV SQLALCHEMY_DATABASE_URI=postgresql://postgres:postgres@localhost/notification_api
```
idavidmcdonald, line 125: This is slightly different to what we have in our `db-unit-tests` dockerfile - https://github.com/alphagov/notifications-aws/blob/main/concourse/docker/db-unit-tests-image.Dockerfile#L35

That might be OK though... happy to find out and change it to what it was if we need to

idavidmcdonald (approved):
Thanks Roberto.

I think this looks good. I've left two small comments for potential changes and two comments just thinking about the bigger picture.

I think when you merge this, you'll soon find out if there are any problems with the database set up commands and then you can come back and easily tweak.

README.md
```diff
@@ -49,15 +49,13 @@ notify-pass credentials/mmg
...
-The API (on Notify) works with PostgreSQL 15. After installation, open the Postgres app, open the sidebar, and update or replace the default server with a compatible version.
+If you are using [notifications-local](https://github.com/alphagov/notifications-local), the correct Postgres version will be provided automatically by the docker-compose file.
 
-**Note:** you may need to add the following directory to your PATH in order to bootstrap the app.
+If you are running this app manually, you will need to manage Postgres yourself. Install [Postgres.app](http://postgresapp.com/). Check the docker-compose file above to find the correct Postgres version to use.
 
-```
-export PATH=${PATH}:/Applications/Postgres.app/Contents/Versions/15/bin/
-```
+When running our test suite, Postgres is baked into the container via the `concourse_tests` step of `docker/Dockerfile`.
```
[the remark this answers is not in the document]
robbrazil: ok

docker/Dockerfile
```diff
@@ -83,3 +83,43 @@ COPY --chown=notify:notify Makefile requirements.txt requirements_for_test.txt .
 RUN make bootstrap
 
 COPY --chown=notify:notify . .
+
+##### Concourse Test Image ##############################################################
+FROM test as concourse_tests
+
+USER root
+
+ENV POSTGRES_VERSION=15 \
```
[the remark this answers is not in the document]
robbrazil: I am not sure minor version upgrades would play a signifcant part here. We could probably discuss this with the Dev team and come up with an approach that suit them as well.

docker/Dockerfile
```diff
@@ -83,3 +83,43 @@ COPY --chown=notify:notify Makefile requirements.txt requirements_for_test.txt .
...
+    echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main $POSTGRES_VERSION" > /etc/apt/sources.list.d/pgdg.list \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        postgresql-$POSTGRES_VERSION \
+    && rm -rf /var/lib/apt/lists/* /tmp/*
+
+RUN \
+    service postgresql start \
+    && su - postgres -c "psql -c \"alter user postgres with password 'postgres';\"" \
+    && su - postgres -c "psql -c \"create database notification_api;\""
```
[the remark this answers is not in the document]
robbrazil: David,
After analysing the test layer, where `make bootstrap` is run (on the assumption that the db is already installed, I decided to build the `concourse_tests` layer based on production and bringing in all the necessary requirements to it. 

- `MAKE BOOTSTRAP` is gone.
- The duplicated references are gone (`generate version files` and `python packages`).
- I have also added the `flask db upgrade` just before the tests start in the pipeline

docker/Dockerfile
robbrazil: that should be ok either way

idavidmcdonald (approved):
Let us give it a try!

PR #3997, opened 2024-01-03
Adjust values in Redis if letters are cancelled

klssmith:
We store a Redis key for the total daily notifications sent by a service and for each notification type so that we can ensure that services don't exceed their rate limits. If a large batch of letters is cancelled, we should adjust these values in Redis so that cancelled letters do not count as part of the rate limit.

This adds a function to update the values in Redis for total daily notifications sent and total daily letters sent by service if a letter job is cancelled. This could be added to the endpoint to cancel a single letter, but cancelling individual letters is unlikely to cause a service to falsely appear to go over its limit.

It makes use of a new `decrby` method in the `RedisClient`. It would be possible to just use `set` to re-set the value, however the benefit of using `decrby` is that the TTL stays the same - 1 day after the key was first stored.

**Relies on**
- [x] https://github.com/alphagov/notifications-utils/pull/1086

You'll need to install this locally to test

### DIFF
--- a/app/letters/utils.py
+++ b/app/letters/utils.py
@@ -6,13 +6,16 @@ from enum import Enum
 
 import boto3
 from flask import current_app
+from notifications_utils.clients.redis import daily_limit_cache_key
 from notifications_utils.letter_timings import LETTER_PROCESSING_DEADLINE
 from notifications_utils.pdf import pdf_page_count
 from notifications_utils.s3 import s3upload
 from notifications_utils.timezones import convert_utc_to_bst
 
+from app import redis_store
 from app.constants import (
     KEY_TYPE_TEST,
+    LETTER_TYPE,
     NOTIFICATION_VALIDATION_FAILED,
     RESOLVE_POSTAGE_FOR_FILE_NAME,
     SECOND_CLASS,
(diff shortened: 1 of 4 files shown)


app/letters/utils.py
```diff
@@ -238,3 +241,27 @@ def get_billable_units_for_letter_page_count(page_count):
...
     return billable_units
+
+
+def adjust_daily_service_limits_for_cancelled_letters(service_id, no_of_cancelled_letters):
+    """
+    Updates the Redis values for the daily letters sent and total number of notifications sent
+    by a service. These values should be decreased if letters are cancelled.
+
+    There shouldn't be a case where these keys don't exist or where they would be changed to a
+    negative number, but this checks for those conditions anyway.
```
samuelhwilliams, line 252: I'm not sure this is true. My understanding of our rate limiting cache keys is that they don't all reset at a specific time; they last for 24 hours from created.

So if a service sends 1 letter on Monday at 1pm, it creates the cache key. Then if they send 10k letters at 10am the next day, those would be added to the existing cache key. At 1pm that cache key would expire, and they could cancel the job which would either hit a missing cache key or potentially go negative.

So a good thing you have checked those conditions.

But if you agree that this is correct then maybe good to remove or update the comment. I can't, at least, see anywhere where we clean out cache keys at a specific time.

It does make me think that, for letters at least, it probably does make sense to clear the cache key at 5:30, when we process all the letters for that day, though...

samuelhwilliams (approved):
Works nicely 👍 

It'd be nice to also do the rate limiting fix for single letter sends so that we don't have an unusual edge case, but can do that separately. This looks good to go (once utils merged and bumped here) 👍

app/letters/utils.py
klssmith: While updating the comment I realised that I'll need to make some further changes to this PR. We use a different cache key for each day, so need to make sure we are changing the value for the day when the letters were created. For example, if I send letters yesterday they will be counted under `"<service_id>-letter-<yesterday>-count"`. But if I then try to cancel them today I will be changing `"<service_id>-letter-<today>-count"` instead.

samuelhwilliams: Yeah, I think that's where my last sentence comes in a bit.

I think for letters it may make sense to align the cache keys to the 5:30pm deadline. So either clear all the (letter limit) cache keys when that tasks starts, or do some maths on the cache key TTL to make it always expire at 5:30pm

klssmith: Even if cache keys are expired at 5.30pm (which I think it makes sense to do) this PR would still have the issue where cancelling letters sent after 5.30pm yesterday changes the wrong key - today's instead of yesterday's.

One quick solution would be to add in some logic so that we only adjust the Redis values if we're cancelling letters sent the same day. If we're cancelling letters sent yesterday there would be no point in trying to update the key from yesterday since it won't be used. I'll see how simple that would be to add.

samuelhwilliams: Ah - I didn't think of the case where you are cancelling letters from the day before (or more) - ie scheduled jobs? Good catch.

klssmith: It won't be scheduled jobs (you can only schedule texts and emails) but someone could still be cancelling a job that was created yesterday evening. I've pushed a new commit which makes sure that we only adjust the cache key if you're cancelling letters sent today.

It's still a bit weird that the letter sending period is 17.30 to 17.30 and the cache period is midnight to midnight, but that might be a bigger bit of work so could be done separately.

samuelhwilliams: Good catch. I totally missed that the cache key had the date it in - somehow 🙃

PR #4001, opened 2024-01-08
paas manifest: handle -ecs-fixup suffixed apps

risicle:
This is one possible implementation for https://trello.com/c/TSJ74HqV/533-create-a-deployment-in-paas-of-the-workers-that-is-scaled-down-to-0-which-has-the-queue-prefix-for-ecs-production-instead-of-paa - the approach that creates copies of each app listening on the ECS queue.

Not tested yet (how?).

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -82,7 +82,22 @@
   'notify-delivery-worker-save-api-notifications': {'disk_quota': '4G'},
 } -%}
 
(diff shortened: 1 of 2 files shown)


risicle:
Arg. This approach doesn't actually work because of course jinja is not python https://jinja.palletsprojects.com/en/3.1.x/templates/#expressions

c4seyla (approved):

PR #4002, opened 2024-01-09
paas manifest: handle notify-delivery-worker-ecs-fixup app

risicle:
This is the "combined app" approach to https://trello.com/c/TSJ74HqV/533-create-a-deployment-in-paas-of-the-workers-that-is-scaled-down-to-0-which-has-the-queue-prefix-for-ecs-production-instead-of-paa

### DIFF
--- a/manifest.yml.j2
+++ b/manifest.yml.j2
@@ -57,6 +57,19 @@
       'production': 1
     },
   },
+  'notify-delivery-worker-ecs-fixup': {
+    'disk_quota': '10G',
+    'memory': '4G',
+    'additional_env_vars': {
+      'NOTIFICATION_QUEUE_PREFIX': NOTIFICATION_QUEUE_PREFIX + '-ecs',
+      'CELERYD_PREFETCH_MULTIPLIER': 1,
+    },
+    'instances': {
+      'preview': 0,
+      'staging': 0,
+      'production': 0
+    },
+  },
   'notify-delivery-worker-jobs': {'memory': '2G'},
   'notify-delivery-worker-research': {'disk_quota': '10G'},
   'notify-delivery-worker-sender': {'disk_quota': '4G', 'memory': '4G'},
--- a/scripts/paas_app_wrapper.sh
+++ b/scripts/paas_app_wrapper.sh
@@ -4,6 +4,10 @@ case $NOTIFY_APP_NAME in
     unset GUNICORN_CMD_ARGS
     exec scripts/run_app_paas.sh gunicorn -c /home/vcap/app/gunicorn_config.py application
     ;;
+  delivery-worker-ecs-fixup)
+    exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=4 \
+    -Q broadcast-tasks,create-letters-pdf-tasks,database-tasks,job-tasks,letter-tasks,notify-internal-tasks,periodic-tasks,reporting-tasks,research-mode-tasks,retry-tasks,save-api-email-tasks,save-api-sms-tasks,send-email-tasks,send-letter-tasks,send-sms-tasks,service-callbacks,service-callbacks-retry,ses-callbacks,sms-callbacks 2> /dev/null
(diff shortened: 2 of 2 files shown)


manifest.yml.j2
```diff
@@ -57,6 +57,21 @@
...
     },
   },
+  'notify-delivery-worker-ecs-fixup': {
+    'disk_quota': '10G',
+    'memory': '4G',
+    'additional_env_vars': {
+      'NOTIFICATION_QUEUE_PREFIX': NOTIFICATION_QUEUE_PREFIX + '-ecs',
+      'CELERYD_PREFETCH_MULTIPLIER': 1,
+      'CBC_PROXY_AWS_ACCESS_KEY_ID': CBC_PROXY_AWS_ACCESS_KEY_ID,
+      'CBC_PROXY_AWS_SECRET_ACCESS_KEY': CBC_PROXY_AWS_SECRET_ACCESS_KEY,
```
idavidmcdonald, line 67: You can get rid of these 2 lines. I can see these have come from the `notify-delivery-worker-broadcasts` which is actually an app no longer in use that is ready to be deleted! We aren't running any tasks through that app anymore as it was all to do with emergency alerts.

manifest.yml.j2
```diff
@@ -57,6 +57,21 @@
       'production': 1
     },
   },
+  'notify-delivery-worker-ecs-fixup': {
+    'disk_quota': '10G',
+    'memory': '4G',
+    'additional_env_vars': {
+      'NOTIFICATION_QUEUE_PREFIX': NOTIFICATION_QUEUE_PREFIX + '-ecs',
+      'CELERYD_PREFETCH_MULTIPLIER': 1,
```
idavidmcdonald, line 65: No action needed:

This will make our worker a bit slower... but it will mean that if it has to pull down a bunch of reporting tasks (usually run by the `reporting-worker` then it will process them well (because those tasks can take longer than 5 minutes to run and the prefetch_multiplier reduces the chance of duplicates).

I think this is worth it

scripts/paas_app_wrapper.sh
```diff
@@ -4,6 +4,10 @@ case $NOTIFY_APP_NAME in
     unset GUNICORN_CMD_ARGS
     exec scripts/run_app_paas.sh gunicorn -c /home/vcap/app/gunicorn_config.py application
     ;;
+  delivery-worker-ecs-fixup)
+    exec scripts/run_app_paas.sh celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=4 \
+    -Q broadcast-tasks,create-letters-pdf-tasks,database-tasks,job-tasks,letter-tasks,notify-internal-tasks,periodic-tasks,reporting-tasks,research-mode-tasks,retry-tasks,save-api-email-tasks,save-api-sms-tasks,send-email-tasks,send-letter-tasks,send-sms-tasks,service-callbacks,service-callbacks-retry,ses-callbacks,sms-callbacks 2> /dev/null
```
idavidmcdonald, line 9: You can remove `broadcast-tasks` as these are no longer used as they were for emergency alerts and are due to be removed...

manifest.yml.j2
```diff
@@ -57,6 +57,21 @@
       'production': 1
     },
   },
+  'notify-delivery-worker-ecs-fixup': {
+    'disk_quota': '10G',
+    'memory': '4G',
+    'additional_env_vars': {
+      'NOTIFICATION_QUEUE_PREFIX': NOTIFICATION_QUEUE_PREFIX + '-ecs',
```
idavidmcdonald, line 64: This will mean we get 2 environments with this name defined in our manifest file. I hope/assume it takes the later one (which would be this one you've defined here).

idavidmcdonald (approved):
Looks good. A couple of small comments to remove some emergency alerts related stuff but nothing substantial.

manifest.yml.j2
risicle: gone

idavidmcdonald (approved):

PR #4016, opened 2024-02-05
config: set `NOTIFY_RUNTIME_PLATFORM` and `NOTIFY_REQUEST_LOG_LEVEL`

risicle:
For some reason this was never set for this app, so on PaaS the app just has an empty value.

This configuration should continue to produce `app.request` post-request log messages on PaaS.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -118,6 +118,9 @@ class Config(object):
     DEBUG = False
     NOTIFY_LOG_PATH = os.getenv("NOTIFY_LOG_PATH")
 
+    NOTIFY_RUNTIME_PLATFORM = os.getenv("NOTIFY_RUNTIME_PLATFORM", "paas")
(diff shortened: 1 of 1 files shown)


manifest.yml.j2
```diff
@@ -128,6 +128,9 @@ applications:
       FLASK_APP: application.py
       NOTIFY_ENVIRONMENT: {{ environment }}
 
+      NOTIFY_RUNTIME_PLATFORM: paas
+      NOTIFY_REQUEST_LOG_LEVEL: INFO
```
idavidmcdonald, line 132: These are being set as environment variables here in the manifest, but I don't think they get set on the `app.config` object. I don't see anywhere in the API repo doing that, or in the utils directory.

Therefore, my guess is that this won't take affect as you intend?

risicle: Yup, changed.

klssmith (approved):

PR #4020, opened 2024-02-15
docker entrypoint: use `--quiet` flag & `--logfile` option to send celery worker logs to `/dev/null`

risicle:
For whatever reason, a lot of celery's worker logs are being sent to stdout despite what documentation says.

~No easy way to pre-merge test, so :shrug:~

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -12,7 +12,7 @@ else
 fi
 
 # Define a common command prefix
-WORKER_CMD="celery -A run_celery.notify_celery worker --logfile=/dev/null --concurrency=$CONCURRENCY"
+WORKER_CMD="celery --quiet -A run_celery.notify_celery worker --logfile=/dev/null --concurrency=$CONCURRENCY"
 COMMON_CMD="$WORKER_CMD -Q"
 
 if [ "$1" == "worker" ]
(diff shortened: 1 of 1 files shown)


entrypoint.sh
```diff
@@ -12,12 +12,12 @@ else
 fi
 
 # Define a common command prefix
-WORKER_CMD="celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=$CONCURRENCY"
+WORKER_CMD="celery -A run_celery.notify_celery worker --logfile=/dev/null --concurrency=$CONCURRENCY"
```
idavidmcdonald, line 15: I guess no point in having the loglevel if it is all going to /dev/null, right?

idavidmcdonald (approved):

idavidmcdonald (approved):

risicle:
A bit more info on this - I've tested this a little bit locally (and read a lot of celery code :roll_eyes: )

The `--quiet` flag should get rid of the vast majority of unstructured logs and random nonsense going to stdout.

Using `--logfile` seems better to me than `2>` because the former leaves open the possibility of critical (e.g. startup) errors from getting through to _some_ logger (albeit in a horrible unstructured form)

entrypoint.sh
risicle: Indeed

idavidmcdonald (approved):

PR #4022, opened 2024-02-19
Bump utils to 74.12.0

risicle:
## 74.12.0

* Remove celery.* structured logging for now. It can return when we have more certainty over what values it will log or can rule out that it is ever given sensitive values.

 ## 74.11.0

* Add option to ignore list of web paths from request logging. Defaults to /_status and /metrics.
* Add new fields to web request log messages (user_agent, host, path)

 ## 74.9.1
***

Complete changes: https://github.com/alphagov/notifications-utils/compare/74.9.1...74.12.0

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -28,7 +28,7 @@ notifications-python-client==8.0.1
 # PaaS
 awscli-cwlogs==1.4.6
 
-notifications-utils @ git+https://github.com/alphagov/notifications-utils.git@74.9.1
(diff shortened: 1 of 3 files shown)


risicle:
Tests broken presumably by https://github.com/alphagov/notifications-utils/pull/1099

quis (approved):

leohemsted (approved):

quis (approved):

PR #4023, opened 2024-02-19
Bumping utils for crown

whpearson:
**74.12.1**

email template: use the new crown and associated styling
**74.12.0**
Remove celery.* structured logging for now. It can return when we have more certainty over what values it will log or can rule out that it is ever given sensitive values.
**74.11.0**
Add option to ignore list of web paths from request logging. Defaults to /_status and /metrics.
Add new fields to web request log messages (user_agent, host, path)
**74.9.1**


leohemsted:
superceded by https://github.com/alphagov/notifications-api/pull/4022

PR #4033, opened 2024-03-12
Remove `service_has permission` function

quis:
Using the method on the service model itself is less verbose.

Some detail of specific changes in individual commits.

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -34,7 +34,7 @@ from app.dao.templates_dao import (
 from app.errors import InvalidRequest, register_errors
 from app.letters.utils import get_letter_pdf_and_metadata
 from app.models import Template
-from app.notifications.validators import check_reply_to, service_has_permission
+from app.notifications.validators import check_reply_to
 from app.schema_validation import validate
 from app.schemas import (
     template_history_schema,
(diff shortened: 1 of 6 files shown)


app/template/rest.py
```diff
@@ -80,12 +80,12 @@ def validate_parent_folder(template_json):
 def create_template(service_id):
     fetched_service = dao_fetch_service_by_id(service_id=service_id)
     # permissions needs to be placed here otherwise marshmallow will interfere with versioning
-    permissions = [p.permission for p in fetched_service.permissions]
+    [p.permission for p in fetched_service.permissions]
```
BlessedDev, line 83: This line seems to be an  accidental holdover

BlessedDev (dismissed):
The changes make sense and makes the code not just less verbose but more readable.
I have flagged up one issue for your consideration.

app/template/rest.py
quis: Weirdly when I remove that line the tests fail. Probably related to the comment on the line above. Will see if I can investigate and work out why this is happening

quis: I have done some digging into why this was happening.

Here’s a fix, which should be merged first: https://github.com/alphagov/notifications-api/pull/4048

Then in this PR we can remove the line which is seemingly doing nothing.

leohemsted (approved):

PR #4034, opened 2024-03-12
Remove all things PaaS

corlettb:
What
----

Remove all things PaaS

Why
----

We are decommissioning PaaS


klssmith:
Just a few more things I think can be removed:
- `jinja2-cli[yaml]==0.8.2` from requirements_for_test.text
- runtime.txt
- there are instructions in the README for `cf run-task...` which are now out of date
- The Makefile doesn't use `NOTIFY_CREDENTIALS` any more and I also couldn't find anywhere that we use the preview, staging and production targets (though worth double checking)

corlettb:
Thanks. Updated.

klssmith (approved):

PR #4035, opened 2024-03-12
sql change to speed it up

c4seyla:
Changes the data access object to streamline the background sql for the live_services.csv, this should make it faster than the gunicorn timeout of 30s


leohemsted (approved):
i think this looks sensible

i copied the queries out of test runs, ran them against prod, and the resultant CSVs are identical 👨‍🍳 💋 

even if the query isn't significantly different, not merging the rows in python makes a lot of sense and it's simpler to read 👍

PR #4037, opened 2024-03-13
Simplify how we update contact blocks on letter templates

quis:
See individual commits.

The aim of this is to get rid of the separate method with its hand-crafted `dict` of properties, and prevent a re-occurrence of bugs like https://github.com/alphagov/notifications-api/pull/4032

I would have liked to go further and re-use the same code path as other updates but was defeated by schema weirdness (something to do with `reply_to`, `service_letter_contact_block` and `service_letter_contact_block_id` being sort of the same things but sort of different).


leohemsted (approved):
nice change 👍

PR #4044, opened 2024-03-20
Add new sms rate and allowances into the database

CrystalPea:
through a migration.

New sms rate: 2.27p

New allowances:

- 30,000 free text messages for `central`, `nhs_central`
- 10,000 free text messages for `local`, `nhs_local`, `emergency_service`
- 5,000 free text messages for `school_or_college`, `other`
- 0 free text messages for `nhs_gp`

My local db after migration:
Screenshot 2024-03-20 at 16 50 24

### DIFF
--- a/migrations/versions/0440_new_sms_allowance_n_rate.py
+++ b/migrations/versions/0440_new_sms_allowance_n_rate.py
@@ -1,0 +1,45 @@
+"""
+
+Revision ID: 0440_new_sms_allowance_n_rate
+Revises: 0439_intl_letters_jan_24
+Create Date: 2023-03-06 11:32:20.588364
+
+"""
+
+import uuid
+
+from alembic import op
+
+
+revision = "0440_new_sms_allowance_n_rate"
+down_revision = "0439_intl_letters_jan_24"
+
+new_allowances = {
+    "central": 30000,
+    "nhs_central": 30000,
+    "local": 10000,
+    "nhs_local": 10000,
+    "emergency_service": 10000,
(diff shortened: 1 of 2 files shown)


migrations/versions/0440_new_sms_allowance_n_rate.py
```diff
@@ -0,0 +1,47 @@
...
+
+    for org_type, allowance in new_allowances.items():
+        op.execute(
+            "INSERT INTO default_annual_allowance(id, valid_from_financial_year_start, organisation_type, allowance, notification_type) "
+            f"VALUES('{uuid.uuid4()}', 2024, '{org_type}', {allowance}, 'sms')"
+        )
+
+
+def downgrade():
+    op.execute(f"DELETE FROM rates WHERE id = '{new_sms_rate_id}'")
```
whpearson, line 45: This didn't delete the row with a price of 0.0227 for me running flask db downgrade I think it forgets the uuid between different invocations. Perhaps using the valid from to delete the row instead would work.

whpearson (changes requested):

migrations/versions/0440_new_sms_allowance_n_rate.py
CrystalPea: Ooh good catch, you are absolutely right!

CrystalPea: I fixed it now 😌

whpearson (approved):
Changes look good.

PR #4048, opened 2024-03-22
Late-assign IDs to objects with history

quis:
Objects that have a corresponding `…History` object must have a value for their `id` property. This is how we know they are versions of the same object, because they have the same value for their primary key.

We assign the value of the `id` property in the DAO layer, for example:

https://github.com/alphagov/notifications-api/blob/492143312cd5ad75775f49c8fd7d3037370115fd/app/dao/templates_dao.py#L17

This is assumed to be the last code that is run before the object is persisted. If we didn’t do this then SQLAlchemy would generate the ID for us at the time we `commit`.

However there is a problem if – earlier in the process – the REST layer creates other objects which have a foreign key relationship to an object which has not been persisted.

What happens in this case is:
1. primary object gets created (without an ID)
2. secondary object gets created, forcing SQLAlchemy to give an ID to the primary object to define the foreign key relationship
3. SQLAlchemy autoflushes and commits the objects
3. primary object is passed to the DAO layer, where it is given a new ID
4. DAO layer re-commits the object, violating a foreign key constraint because the secondary object now refers to an ID which doesn’t exist

The way we have got around this is by triggering unrelated queries on a different object, which seems to prevent step 3 from happening, for example:
https://github.com/alphagov/notifications-api/blob/492143312cd5ad75775f49c8fd7d3037370115fd/app/template/rest.py#L82-L83

If the above lines are moved below this one then the `IntegrityError`s start happening, even though the two bits of code seem unrelated:
https://github.com/alphagov/notifications-api/blob/492143312cd5ad75775f49c8fd7d3037370115fd/app/template/rest.py#L86

***

This pull request moves the assigning of `id`s into the code which creates the history object.

This means that:
- it is done consistently in one place
- it is only done where objects don’t already have a `id`, so we never change an objects `id`

***

A less general purpose proof of concept – which might aid understanding – is shown in e03e5afb3fec75a967451fa1ded03b40f0ecb4af

### DIFF
--- a/app/dao/api_key_dao.py
+++ b/app/dao/api_key_dao.py
@@ -11,8 +11,6 @@ from app.models import ApiKey
 @autocommit
 @version_class(ApiKey)
 def save_model_api_key(api_key):
-    if not api_key.id:
-        api_key.id = uuid.uuid4()  # must be set now so version history model can use same id
     api_key.secret = uuid.uuid4()
     db.session.add(api_key)
 
--- a/app/dao/service_callback_api_dao.py
+++ b/app/dao/service_callback_api_dao.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 
-from app import create_uuid, db
+from app import db
 from app.constants import COMPLAINT_CALLBACK_TYPE, DELIVERY_STATUS_CALLBACK_TYPE
 from app.dao.dao_utils import autocommit, version_class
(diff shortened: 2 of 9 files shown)


leohemsted (approved):
just needs a lil comment i think

leohemsted (approved):

PR #4051, opened 2024-03-27
Add unsubscribe link to our models and database

CrystalPea:
As a new column in notifications table. Also add a constraint that forces the link to be null for non-email notifications.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -114,7 +114,7 @@ class User(db.Model):
     take_part_in_research = db.Column(db.Boolean, nullable=False, default=True)
 
     # either email auth or a mobile number must be provided
-    CheckConstraint("auth_type in ('email_auth', 'webauthn_auth') or mobile_number is not null")
+    __table_args__ = (CheckConstraint("auth_type in ('email_auth', 'webauthn_auth') or mobile_number is not null"),)
 
     services = db.relationship("Service", secondary="user_to_service", backref="users")
     organisations = db.relationship("Organisation", secondary="user_to_organisation", backref="users")
@@ -1440,9 +1440,6 @@ class Notification(db.Model):
 
     unsubscribe_link = db.Column(db.String, nullable=True)
 
-    # unsubscribe_link value should be null for non-email notifications
-    CheckConstraint("notification_type = 'email' OR unsubscribe_link is null")
-
     __table_args__ = (
         db.ForeignKeyConstraint(
             ["template_id", "template_version"],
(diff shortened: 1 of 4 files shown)


app/models.py
```diff
@@ -1438,6 +1438,11 @@ class Notification(db.Model):
 
     postage = db.Column(db.String, nullable=True)
 
+    unsubscribe_link = db.Column(db.String, nullable=True)
+
+    # unsubscribe_link value should be null for non-email notifications
+    CheckConstraint("notification_type = 'email' OR unsubscribe_link is null")
```
leohemsted, line 1444: i think this should be in the `__table_args__` below?

https://docs.sqlalchemy.org/en/14/orm/declarative_tables.html#declarative-table-configuration

might be a good idea to check the other constraints in the file - most are in __table_args__ but there's a couple of stragglers.

i dont think has a huge impact on us (it's more used when sqlalchemy is used to generate db schemas, whereas we just use alembic), but would be nice to get it consistent

migrations/versions/0441_add_unsubscribe_link.py
```diff
@@ -0,0 +1,29 @@
...
+down_revision = "0440_new_sms_allowance_n_rate"
+
+
+def upgrade():
+    op.add_column("notifications", sa.Column("unsubscribe_link", sa.String(), nullable=True))
+    op.create_check_constraint(
+        "ck_unsubscribe_link_is_null_for_letters_and_sms",
+        "notifications",
+        """
+        notification_type = 'email' OR unsubscribe_link is null
```
leohemsted, line 22: additionally, can we add a test that asserts that the constraint is correct (by manually allowing both null and non-null emails, while rejecting non-null unsub links?).

this'll also mean we update create_notification fixtures in the db which would be good to do

app/models.py
CrystalPea: Yeah, I think the syntax was half-like-this, half-like-that. I don't mind either way, so I put them all in table args variables. Ready for another look ☺️

migrations/versions/0441_add_unsubscribe_link.py
CrystalPea: I'm sorry, could you elaborate? I am not sure I understand what you're asking for.

Do you want me to add a unit test that expects error when trying to create an sms/letter with an unsubscribe link?

> this'll also mean we update create_notification fixtures in the db

Do you mean test fixtures here, or something in real database?

I intentionally kept this PR small, as it's a migration PR. I wanted to do more codey things in a next, separate PR. But if you think the constraint needs testing, I could add that?

leohemsted: yeah, i think a test where we just make sure the constraint works would be good. it'd be awkward if we accidentally made a mistake in this constraint and haven't tested the code and need to do another migration to fix it.

spitballing, but could do something like
```python
@pytest.mark.parametrise('template_type, unsubscribe_link', [
 ('email', None),
 ('email', 'https://unsub.com'),
 pytest.mark.xfail('sms', 'https://unsub.com', raises=SomeException),
]
def test_constraint_works():
  notification = create_notifications(type=type, unsubscribe_link=unsubscribe_link)
  db.commit()
  assert notification.unsubscribe_link = unsubscribe_link
```

CrystalPea: Ok, this is done now :)

leohemsted (approved):

PR #4055, opened 2024-04-03
Save and return email notification with unsubscribe_link

CrystalPea:
if that link is provided, via V2 API integration.


whpearson (approved):
Changes make sense

PR #4056, opened 2024-04-03
try and improve template delete performance

leohemsted:
Background: When deleting templates, we find out when it was most recently used so that we can inform the user.

Problem: This query can be extremely slow, sometimes timing out. A couple of random picks showed it to be slow when templates haven't been used recently at all (so it needs to search through lots of notifications to see if there's a match so it can find the most recent created_at timestamp)

Reckon: templates that people want to delete probably haven't been used recently most of the time (or they wouldn't want to delete them).

Theory: if we search using the `ix_notifications_template_id` index just to see if any row exists at all, we can bypass this query for a decent chunk of template delete queries, thus improving performance.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -87,18 +87,22 @@ FIELDS_TO_TRANSFER_TO_NOTIFICATION_HISTORY = [
 
 
 def dao_get_last_date_template_was_used(template_id, service_id):
-    last_date_from_notifications = (
-        db.session.query(functions.max(Notification.created_at))
-        .filter(
-            Notification.service_id == service_id,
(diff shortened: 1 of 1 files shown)


CrystalPea (approved):
The code looks good 👍🏼 . Nice performance gain 🙌🏼 

> Reckon: templates that people want to delete probably haven't been used recently most of the time (or they wouldn't want to delete them).

I think the main reason we do this check/warning is because many services have lots of similarly-named templates and it's easy to make a mistake. Just two days ago when I was on support a user sent over a 1000 letters using a wrong template, and wrote to us thinking we mixed something up. So it's good we retain this check, and well done for making it more efficient 👏🏼

PR #4062, opened 2024-04-12
send inbound sms tasks to service-callback queue

leohemsted:
they were prev on the catch all internal queue but they fit better here (and we can scale sms-callback queue better to deal with potential spikes in volumes based on 3rd party inputs)

### DIFF
--- a/app/notifications/receive_notifications.py
+++ b/app/notifications/receive_notifications.py
@@ -65,7 +65,7 @@ def receive_mmg_sms():
         provider_name="mmg",
     )
 
-    tasks.send_inbound_sms_to_service.apply_async([str(inbound.id), str(service.id)], queue=QueueNames.NOTIFY)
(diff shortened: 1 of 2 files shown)


risicle (approved):
Guess I should take some blame for this too if it goes wrong

risicle:
Actually wouldn't the `service-callbacks` queue make more sense? Seeing as the work that needs doing is making a request back to a user's endpoint?

leohemsted:
no wait you're right, i got confused because the DR receipts call the sms-callback queue to internally process the updating the db etc, whereas we process the inbound ones synchronously in the web request. probably something to do in the future but not just a "quick lets do this on a fri afternoon" jobby

leohemsted:
squashed to change the queue to hide my mistakes

risicle (approved):

PR #4070, opened 2024-04-24
Remove platform admin when a user is archived

quis:
This means we won’t need to access the database to remove people’s access when they leave GDS.


idavidmcdonald (approved):
Smart

CrystalPea (approved):

PR #4071, opened 2024-04-24
Resolve Dependabot alerts

quis:
Closes #4064

Closes #4065


whpearson (approved):
Both changes work locally bootstrapped

DNSpython has no worrying changes in https://dnspython.readthedocs.io/en/stable/whatsnew.html

Eventlet has no worrying changes in 
https://github.com/eventlet/eventlet/blob/master/NEWS

PR #4074, opened 2024-05-03
Drop unused index on letter despatch cost threshold

CrystalPea:
We did an audit of indexes in our database and found that some of them have not been used at all.

Here is the list of all unused insexes: https://docs.google.com/spreadsheets/d/1Eqfze5WTXM4LDIFDqPcfQ_7axnaodC0zzOz5eFmCU-U/edit?usp=sharing (look at 2nd tab)

In this PR I am removing one of those indexes, just to cautiously dip a toe. If that goes well, we will remove all the rest of them.

More context about decision to remove unused indexes is explained in this blog post: https://www.cybertec-postgresql.com/en/get-rid-of-your-unused-indexes/

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0442_add_new_service_permission
+0443_drop_ix_letter_cost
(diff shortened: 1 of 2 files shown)


idavidmcdonald (approved):
Looks good.

You are using concurrently as suggested by https://medium.com/paypal-tech/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680 for DROP INDEX which means this should be zero downtime.

PR #4077, opened 2024-05-07
Remove 9 biggest remaining unused indexes

CrystalPea:
I divided them into two migration files for better readability.

Info on droping many indexes at once concurrently: https://dba.stackexchange.com/questions/319720/multiple-drop-index-concurrently-running-on-the-same-table-in-postgres

Here is the list of all unused indexes: https://docs.google.com/spreadsheets/d/1Eqfze5WTXM4LDIFDqPcfQ_7axnaodC0zzOz5eFmCU-U/edit?usp=sharing

Why it's good to remove unused indexes: https://www.cybertec-postgresql.com/en/get-rid-of-your-unused-indexes/

### DIFF
--- a/migrations/versions/0446_drop_unused_indexes_2.py
+++ b/migrations/versions/0446_drop_unused_indexes_2.py
@@ -1,0 +1,77 @@
+"""
+
+Revision ID: 0446_drop_unused_indexes_2
+Revises: 0445_drop_unused_indexes_1
+Create Date: 2023-09-17 15:17:58.545277
+
+"""
+
+from alembic import op
+
+
+revision = "0446_drop_unused_indexes_2"
+down_revision = "0445_drop_unused_indexes_1"
+
+
+def upgrade():
+    with op.get_context().autocommit_block():
+        # 6
+        op.drop_index(
+            "ix_users_auth_type",
+            table_name="users",
+            postgresql_concurrently=True,
+        )
+        # 7
+        op.drop_index(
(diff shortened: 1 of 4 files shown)


klssmith:
Could we remove these from the models too? Otherwise when running `flask db migrate`, the migration that gets generated attempts to add them back

klssmith:
I've realised we can't always adjust the models to match the indexes we have, since some of the "normal tables" and history tables share a model and we're only removing the index from one, but it might be possible to do it in some cases

migrations/versions/0446_drop_unused_indexes_2.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        # 7
+        op.drop_index(
+            "ix_api_keys_history_created_by_id",  # can't remove index from model as this is based on api_keys table
+            table_name="api_keys_history",
+            postgresql_concurrently=True,
+        )
+        # 8
+        op.drop_index(
+            "ix_api_keys_history_key_type",  # can't remove index from model as this is based on api_keys table
```
klssmith, line 32: Really minor, but this index has been removed from the model (but it's only a comment in a migration so it's not too important)

klssmith (approved):

migrations/versions/0446_drop_unused_indexes_2.py
CrystalPea: Ahh good spot, I have deleted that comment now.

PR #4078, opened 2024-05-08
Remove enivronment knowledge from application

corlettb:
What
----

Remove environment knowledge from application. Environment configuration will be passed in during deployments.

Replace environment name checks with feature flags. New feature flags added by this change are:

- REGISTER_FUNCTIONAL_TESTING_BLUEPRINT
- SEND_ZENDESK_ALERTS_ENABLED
- CHECK_SLOW_TEXT_MESSAGE_DELIVERY
- SEND_LETTERS_ENABLED
- SENTRY_ALLOW_PII

Previously each of these were performance environment name checks to determine whether to enable or disable certain functionality.

Why
----

We are updating our notify deployment model to allow for many more environments. As such environment settings will be kept in as few places as possible (eventually just one place).

The main benefit of this is to be able to add new environments without having to modify source code

Note
----

This change should not be merged until the deployment configuration has been updated. See [the following pr](https://github.com/alphagov/notifications-aws/pull/2202)


app/config.py
```diff
@@ -435,6 +435,24 @@ class Config(object):
...
+    S3_BUCKET_LETTERS_SCAN = os.environ.get("S3_BUCKET_LETTERS_SCAN")
+    S3_BUCKET_INVALID_PDF = os.environ.get("S3_BUCKET_INVALID_PDF")
+    S3_BUCKET_TRANSIENT_UPLOADED_LETTERS = os.environ.get("S3_BUCKET_TRANSIENT_UPLOADED_LETTERS")
+    S3_BUCKET_LETTER_SANITISE = os.environ.get("S3_BUCKET_LETTER_SANITISE")
+    FROM_NUMBER = os.environ.get("FROM_NUMBER")
+    API_RATE_LIMIT_ENABLED = os.environ.get("API_RATE_LIMIT_ENABLED", "1") == "1"
+
+    SEND_LETTERS_ENABLED = os.environ.get("SEND_LETTERS_ENABLED", "0") == "1"
+    REGISTER_FUNCTIONAL_TESTING_BLUEPRINT = os.environ.get("REGISTER_FUNCTIONAL_TESTING_BLUEPRINT", "0") == "1"
+    SEND_ZENDESK_ALERTS_ENABLED = os.environ.get("SHOULD_SEND_ZENDESK_ALERTS", "0") == "1"
```
klssmith, line 453:
```suggestion
    SEND_ZENDESK_ALERTS_ENABLED = os.environ.get("SEND_ZENDESK_ALERTS_ENABLED", "0") == "1"
```

klssmith (commented):
Changes looked good, but there were just a couple of small things I noticed

`CRONITOR_ENABLED` is hardcoded to `False` in the default config, so won't be overridden by an environment variable.

The default config contains `NOTIFY_ENVIRONMENT = "development"`, so even though the `create_app` function gets `NOTIFY_ENVIRONMENT` from an environment variable it always gets overwritten as `development`.

app/config.py
corlettb: Updated

corlettb:
I have updated the CRONITOR_ENABLED and NOTIFY_ENVIRONMENT settings

app/config.py
```diff
@@ -120,7 +120,7 @@ class Config(object):
     NOTIFY_REQUEST_LOG_LEVEL = os.getenv("NOTIFY_REQUEST_LOG_LEVEL", "INFO")
 
     # Cronitor
-    CRONITOR_ENABLED = False
+    CRONITOR_ENABLED = os.environ.get("CRONITOR_ENABLED", False)
```
risicle, line 123: Other settings are comparing with `0`/`1` to get their boolean result, this isn't.

corlettb: Good spot! I've updated.

risicle (approved):
I *think* this is right.

PR #4079, opened 2024-05-08
Move to using SES v2 API

leohemsted:
This is probably worth a separate pull request to anything around custom headers, as it makes a big change - using SESv2 (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sesv2.html), which has a new format.

Also do some refactoring to make it clearer what is being passed in and around this function


CrystalPea (approved):
LGTM 👍🏼

BlessedDev (approved):

PR #4081, opened 2024-05-09
Add https validation for unsusbscribe links

BlessedDev:
This is to enforce https urls for unsubscribe links posted to the post_notification endpoint.

### DIFF
--- a/app/v2/notifications/notification_schemas.py
+++ b/app/v2/notifications/notification_schemas.py
@@ -4,7 +4,7 @@ from app.constants import (
     NOTIFICATION_STATUS_TYPES,
     NOTIFICATION_TYPES,
 )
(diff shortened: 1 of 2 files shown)


CrystalPea (commented):
Nice one, it would be also cool to have a test for this validation, no?

I think it would go in this file: https://github.com/alphagov/notifications-api/blob/main/tests/app/v2/notifications/test_notification_schemas.py

BlessedDev:
> Nice one, it would be also cool to have a test for this validation, no?
> 
> I think it would go in this file: https://github.com/alphagov/notifications-api/blob/main/tests/app/v2/notifications/test_notification_schemas.py

Yes, I have added a test for https validation

CrystalPea (approved):

PR #4082, opened 2024-05-09
Send unsubscribe header

leohemsted:
Requires the following to be merged (split into two PRs to keep PRs focused, and risk lower)

- [ ] #4079 

-----------

[bump boto3 (and celery) to bring in latest ses API](https://github.com/alphagov/notifications-api/commit/3bc9a759ca44f4e9e93c6acd98681fbcf94fdeac) 

we need to bump boto3 to bring in the changes to the sesv2 API to add
support for custom headers.

I went through the boto3 changelogs with a fine toothcomb - nothing
there is particularly alarming. Here's a filtered list of just the lines
i think will affect our apps. Note that we only use boto3 for
interacting with things programatically, so while there may have been
changes to, eg, boto3's ecs API, since we don't interact with ECS from
our apps we don't need to worry about that. See the filtered list[^1].

Notably there were some changes to the boto3 sqs protocol, and celery
needed to be bumped to support that - see the kombu PR[^2] if you're
interested in learning more.

[^1]: https://gist.github.com/leohemsted/0572ad15c57fdbaef8b9697cc6e5eaa4
[^2]: https://github.com/celery/kombu/pull/1807
-------
[send unsubscribe headers](https://github.com/alphagov/notifications-api/commit/5909d048f842c2267709b11bcf5be4fbb30480bc) 

send unsubscribe links and one-click directive in line with the RFC[^3]
make a list that we can add further headers to in the future if we want
to

[^3]: https://www.rfcreader.com/#rfc8058_line278

### DIFF
--- a/app/clients/email/aws_ses.py
+++ b/app/clients/email/aws_ses.py
@@ -1,5 +1,5 @@
 from time import monotonic
-from typing import Optional
+from typing import Dict, List, Optional
 
 import boto3
 import botocore
@@ -75,6 +75,7 @@ class AwsSesClient(EmailClient):
         body: str,
         html_body: str,
         reply_to_address: Optional[str],
+        headers: List[Dict[str, str]],
     ) -> str:
         reply_to_addresses = [punycode_encode_email(reply_to_address)] if reply_to_address else []
         to_addresses = [punycode_encode_email(to_address)]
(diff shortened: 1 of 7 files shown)


CrystalPea:
We have been testing this with Leo, the headers get passed, but the button doesn't show up. We are not sure why yet, I will look into it more tomorrow.

BlessedDev (approved):

CrystalPea:
I rebased against latest main to enable smooth merge on Monday 👍🏼

PR #4084, opened 2024-05-14
Rename `unsubscribe_link` to `one_click_unsubscribe_url`

quis:
This is:
- more correct (it’s a URL designed for machine consumption, not a link in a web page)
- less likely to be confused with the manual unsubscribe links we talk about in our guidance<sup>1</sup>

This will mean a bit of work to rename the parameter in all the clients, but easier to do it now than after we’ve released them (when it would be a breaking change)

***

Internally we retain `unsubscribe_link`. We will make it consistent later, in order to avoid any need for updates to the `notifications` table (which would require locks, multiple migrations, etc).

***

1. https://www.notifications.service.gov.uk/using-notify/unsubscribe-links or
   snapshotted at https://github.com/alphagov/notifications-admin/blob/20fd4424d613d256e02284cc7c523392c18e8b8c/app/templates/views/guidance/using-notify/unsubscribe-links.html

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1614,7 +1614,7 @@ class Notification(db.Model):
             "completed_at": self.completed_at(),
             "scheduled_for": None,
             "postage": self.postage,
-            "unsubscribe_link": self.unsubscribe_link,
+            "one_click_unsubscribe_url": self.unsubscribe_link,
         }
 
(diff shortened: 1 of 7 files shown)


BlessedDev (approved):
I have tested this locally in tandem with the python api client and it works as expected.

PR #4085, opened 2024-05-15
Add test coverage for getting a user

klssmith:
This makes minor changes to test all the fields that are returned when a User is serialized. It also tests that more user properties can be updated.

### DIFF
--- a/tests/app/user/test_rest.py
+++ b/tests/app/user/test_rest.py
@@ -51,7 +51,7 @@ def test_get_user_list(admin_request, sample_service):
 
 def test_get_user(admin_request, sample_service, sample_organisation):
     """
-    Tests GET endpoint '/<user_id>' to retrieve a single service.
+    Tests GET endpoint '/<user_id>' to retrieve a single user.
     """
     sample_user = sample_service.users[0]
     sample_user.organisations = [sample_organisation]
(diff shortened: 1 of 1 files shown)


tests/app/user/test_rest.py
```diff
@@ -62,15 +62,24 @@ def test_get_user(admin_request, sample_service, sample_organisation):
...
+    assert sorted(fetched["permissions"][str(sample_service.id)]) == sorted(expected_permissions)
+    assert fetched["organisation_permissions"] == {str(sample_organisation.id): []}
+    assert fetched["platform_admin"] is False
     assert fetched["services"] == [str(sample_service.id)]
-    assert fetched["organisations"] == [str(sample_organisation.id)]
     assert fetched["can_use_webauthn"] is False
-    assert sorted(fetched["permissions"][str(sample_service.id)]) == sorted(expected_permissions)
+    assert fetched["state"] == sample_user.state
+    assert fetched["take_part_in_research"] is True
+    assert fetched["receives_new_features_email"] is True
```
idavidmcdonald, line 82: Optional: You could consider adding an assertion on here about the number of keys in `fetched`. That way if we added any additional properties to a user, it would fail this test and we would be prompted to update it

idavidmcdonald (approved):

tests/app/user/test_rest.py
klssmith: Good idea, that's now added (you will need to re-approve)

idavidmcdonald (approved):

PR #4090, opened 2024-05-21
first draft of unsubscribe_request table

leohemsted:
Looking for comments on this please!

the table links to notification, service, templates_history (and templates). it has a batch id that links to - see chris's mockups for info.

notably, just using a view for the report. I'm not sure if we want to do this or not - it's got advantage that we normalise the data and avoid needing to keep different sources of data in check. the disadvantage is that if in the future we want to add data that we dont want to store on the individual unsubscribe_request

Viewing chris's slides about how we currently think we'll batch up unsubscribe requests into reports might be useful for understanding.

https://docs.google.com/presentation/d/1L9pTMQW-4d6z2LfOdBll7qZ3qXxC_gufk8q83zAM3nY/edit#slide=id.g2da06c6c272_0_35


app/models.py
```diff
@@ -2551,3 +2551,90 @@ def serialize(self):
...
+    template = db.relationship(Template, backref=db.backref("unsubscribe_requests"))
+    template_history = db.relationship(TemplateHistory, backref=db.backref("unsubscribe_requests"))
+
+    # nullable as at some point we may want to delete these for data retention purposes
+    email_address = db.Column(db.Text, nullable=True)
+
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+
+    # is this useful or is a boolean enough? if we create a table for a report, we could remove from this table entirely
+    downloaded_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
```
BlessedDev, line 2617: I think `downloaded_at` should be removed from this table entirely as we already have `has_been_downloaded` in the unsubscribe_request_report table and we know the unsubscribe request will be downloaded as part of the report.

app/models.py
```diff
@@ -2551,3 +2551,90 @@ def serialize(self):
...
+
+    id = db.Column(UUID(as_uuid=True), primary_key=True)
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey("services.id"), nullable=False)
+    service = db.relationship(Service, backref=db.backref("unsubscribe_requests"))
+
+    earliest_timestamp = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    latest_timestamp = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    has_been_downloaded = db.Column(db.Boolean, nullable=False)
+    count = db.Column(db.Integer, nullable=False)
+
```
BlessedDev, line 2583: We would need a boolean `processed` attribute and a `processed_at` . On slide 4 of the proposed user interface design, a key part of the journey is the user marking that the report has been actioned and that the contained unsubscribe requests have been actioned.

app/models.py
```diff
@@ -2551,3 +2551,90 @@ def serialize(self):
...
+    # is it useful to
+    template_id = db.Column(UUID(as_uuid=True), db.ForeignKey("templates.id"), nullable=False)
+    template_version = db.Column(db.Integer)
+    template = db.relationship(Template, backref=db.backref("unsubscribe_requests"))
+    template_history = db.relationship(TemplateHistory, backref=db.backref("unsubscribe_requests"))
+
+    # nullable as at some point we may want to delete these for data retention purposes
+    email_address = db.Column(db.Text, nullable=True)
+
```
BlessedDev, line 2613: I assume that will be dictated by the service's data retention policy?
Since the best practice for one click unsubscribe uri is to have an opaque identifier that doesn't expose PII, could we have for instance an `opaque_identifier` which would be a hash of the email address of some sort?

app/models.py
```diff
@@ -2551,3 +2551,90 @@ def serialize(self):
...
+    # this is denormalised but might still be useful to have as a separate column?
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey("services.id"), nullable=False)
+    service = db.relationship(Service, backref=db.backref("unsubscribe_requests"))
+
+    # is it useful to
+    template_id = db.Column(UUID(as_uuid=True), db.ForeignKey("templates.id"), nullable=False)
+    template_version = db.Column(db.Integer)
+    template = db.relationship(Template, backref=db.backref("unsubscribe_requests"))
+    template_history = db.relationship(TemplateHistory, backref=db.backref("unsubscribe_requests"))
+
```
BlessedDev, line 2610: I think  they are all useful to have.

app/models.py
```diff
@@ -2551,3 +2551,90 @@ def serialize(self):
...
+    id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+
+    # WIP:
+    # Ignoring a strict foreign key relationship here for now. Notifications are archived to the NotificationHistory
+    # table by a nightly job and I haven't investigated whether that might break a strict FK yet or if it would
+    # work smoothly. We can still have a relationship using an explicit join condition.
+    notification_id = db.Column(UUID(as_uuid=True), index=True, nullable=False)
+    notification = db.relationship(
+        "NotificationAllTimeView",
+        primaryjoin="NotificationLetterDespatch.notification_id == foreign(NotificationAllTimeView.id)",
```
BlessedDev, line 2597: Should this be `primaryjoin="UnsubscribeRequest.notification_id == foreign(NotificationAllTimeView.id)"`?

leohemsted: oops, typo. squashed out

app/models.py
leohemsted: i've renamed `downloaded_at` to `processed_at` - i'd slightly misinterpreted the slides and had got downloading vs processing mixed up in my head. i dont think its useful to record when something is downloaded, only when it's processed.

leohemsted: so my approach involved using a view for `unsubscribe_request_report` rather than a table.

So the report doesn't actually exist, and there's no value there to be set - it's just derived by looking at this value

app/models.py
leohemsted: changed `downloaded_at` and `has_been_downloaded` to `processed_at` and `has_been_processed` to help resolve this.

but i'm not sure if `processed_at` is too generic a term. Who has processed it? What does processing mean?

I guess processing itself and what that means is up to the service user - but maybe your `processed_by_service` idea has legs 🦵 🦵

leohemsted:
questions for the room:

* is a view okay or should we use a table?
* do we want to delete old records, or just remove the email address PII?
* should email address be synchronously encrypted?

app/models.py
leohemsted: need to check with rest of team this afternoon to answer this.

unsure about data retention. in chris's slides it refers to deleting processed reports after 7 days. we need to decide what that means (do we keep the rows in the db? do we delete the email address PII from the database? do we just hide from the UI? etc)

leohemsted: we couldn't have a hash though, because when the user downloads the report they need the original user email address - so we need it to be either be plain text, or two-way encrypted rather than hashed

idavidmcdonald:
A few random thoughts. I don't think any of these are deal breakers.

1. I think I'd like some form of retention limit on the data in the table. If a service team didn't come and download and process their report for 2 years, I'd say they've left it too late and we shouldn't be obligated to keep this data for that long. I think we should pick less than 2 years, that was to pick an extreme. I'm not sure what the right number should be. 90 days? a data retention period set by the service of up to 90 days?

2. Kind of the same point as above but to emphasise it separately. We only give services 7 days (by default) to download their inbound text messages or they will be lost. Why should this feature be any different? 

3. You mentioned we could make the email address nullable, so we could remove that data if it was processed/reached a retention limit. I'm cautious on the use of nullable. I think for this type of data our precedent is to go for a `_history` table which has the personal data removed and can be kept indefinitely. Is the deviation from this pattern worth it? Or another option you suggested is just delete all of the data - I would be fine with us doing that I think. I don't have a strong opinion on which the best approach is.

4. The point in the meeting about who processed the unsubscriptions was interesting. I really liked the idea of adding an audit event for whenever someone downloads a CSV report out of notify with personal data in! However, whether we should use the `events` table to record who 'processed' the unsubscriptions - I'm not so sure about that. I think I see the purpose of the events table as a security audit table, that we rarely use and interogate. Maybe it also has some purpose for debugging? Maybe? But if we wanted to show something in the user interface to our users like 'Leo Hemsted processed these on Tuesday' then I'd probably encourage us to store this not in the events table as it feels less 'audit' and more 'feature'.

5. Bigger idea and unlikely to be helpful given the time pressure... but if we managed subscribes and not just unsubscribes in Notify, this might be easier because we've just be removing an email from a subscription...

leohemsted:
1. Yes, I agree. Something to pick up with ahmed and chris. I like 90 days as a starting point.
2. Good point. I think the chance of inbound messages happening randomly without _recent_ prompting is a bit lower in my mind, at least for services that dont have api integrations, but thats a pretty weakly held assumption and a fair challenge.
3. fair point. we have an `inbound_sms_history`. i think for now i will make email non-nullable, and then build a separate history table in the same pattern
4. interesting. i'll have a think about it.
5. noooooooo

app/models.py
CrystalPea: Maybe it could be `actioned_at` and `has_been_actioned` ?

leohemsted: i think i like gozie's idea - `processed_by_service_at`. gonna go with that for now

leohemsted:
Pea's comments:

1. 90 day retention - sounds good. i've got a history table now
2. view vs table - i ended up going for a table - i think it just makes it a bit easier to tweak. the main thing that bugged me about the view is that every single unsubscribe request has a "processed_at" timestamp. But really the "processed_at" timestamp belongs to a report itself, and so having a timestamp duplicated across all the rows felt a bit weird.
3. synchronously encrypted - pretty sure i discussed this with david and we decided its fine to have it unencrypted. makes it easier to search for a specific email address too. we dont have anything in place for encryption yet anyway so that'd just make development longer

app/models.py
leohemsted: resolving as no longer a view. also changed downloaded_at, thanks for the suggestion!

BlessedDev (approved):

app/models.py
```diff
@@ -2551,3 +2556,92 @@ def serialize(self):
...
+        primaryjoin="Template.id == UnsubscribeRequest.template_id",
+        backref=db.backref("unsubscribe_requests"),
+    )
+
+    email_address = db.Column(db.Text, nullable=False)
+
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+
+    # is this useful or is a boolean enough? if we create a table for a report, we could remove from this table entirely
+    processed_at = db.Column(db.DateTime, nullable=True)
```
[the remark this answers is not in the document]
leohemsted: ive actually removed this sorry! i thought about it because for now we're just thinking about reports, and the "what if these timestamps end up different" question was bugging me a bit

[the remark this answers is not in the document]
BlessedDev: Fair enough, I don't feel too strongly either way as it seems we have to ship this and make adjustments to the tables if or as needed.

BlessedDev (approved):

leohemsted:
the squawk failures are all of type:

-----
```
/tmp/upgrade.sql:2:2: warning: prefer-timestamptz

  note: A timestamp field without a timezone can lead to data loss, depending on your database session timezone.
  help: Use timestamptz instead of timestamp for your column type.
```
think this is fine as it's in line with all of the rest of our timestamps. pending a larger look at our timestamps cc @CrystalPea 
-----
```
/tmp/upgrade.sql:15:2: warning: prefer-bigint-over-int

  note: Hitting the max 32 bit integer is possible and may break your application.
  help: Use 64bit integer values instead to prevent hitting this limit.
```
this is a `template_version` field so i dont want to change the type as it's a foreign key.

we may want to in time look at how to filter these out

PR #4092, opened 2024-05-22
Add a way to lint migrations when running on Concourse

klssmith:
**Add script to lint migrations from Concourse**
The existing script to lint migrations using Squawk won't work when run
from Concourse. This renames it (with no changes) and creates a new
script for running Squawk on Concourse.

When running from Concourse, we can't use Git to compare the migrations
on the branch to the migrations on main. The pull-request resource we
use clones main at the latest commit and merges the pull request at
into main, so the branch no longer exists. The script for Concourse
checks the latest version from GitHub instead.

**Update Dockerfile to include Squawk**
This updates the `concourse_tests` target of the Dockerfile to install
the squawk-cli for checking migrations. When running on Concourse, Node
was not getting added to the path automatically, so needed to be
manually added (though didn't when running locally).


whpearson (commented):
Looks good the only thing that might be nice is splitting out the shared common code  into a function or script that can be run.

klssmith:
Yeah, I don't love the duplication but then I did think keeping it might be clearer since you can see the whole script in one go instead of needing to look for shared code in another place. I did also wonder about having one file with some kind of argument to determine whether to run the CI vs the local version, but again it seemed a bit more complicated since you'd need to know to pass the right argument through. I can make some kind of shared function though if that seems best.

klssmith:
@whpearson On second thoughts, after having a go at creating a shared function I realised that there are only really a few shared lines to pull out and it starts to look a little more confusing, because it's not immediately clear why there's an extra file with Squawk in the name etc. So I think I'm now leaning towards keeping things as they are

whpearson (approved):

leohemsted (approved):

PR #4094, opened 2024-05-24
Add unsubscribe url endpoint

BlessedDev:
This PR based builds the one-click unsubscribe endpoint to unsubscribe end users as described in [this](https://trello.com/c/qFVyAmMk/32-build-notificationsservicegovuk-unsubscribe) card.
It stops short of "batching" unsubscribe requests as that is a separate piece of work.


tests/app/one-click-unsubscribe/test_one_click_unsubscribe.py
```diff
@@ -0,0 +1,31 @@
...
+    token = generate_token(
+        sample_email_notification.to, current_app.config["SECRET_KEY"], current_app.config["DANGEROUS_SALT"]
+    )
+    response = unsubscribe_url_post(client, sample_email_notification.id, token)
+    response_json_data = response.get_json()
+    created_unsubscribe_request = get_unsubscribe_request_by_notification_id_dao(sample_email_notification.id)
+    assert response.status_code == 200
+    assert response_json_data["message"] == "Unsubscribe successful"
+    assert response_json_data["result"] == "success"
+    assert created_unsubscribe_request.notification_id == sample_email_notification.id
```
leohemsted, line 23: i think id like a couple of more thorough assertions here on the other values of the unsubscribe request

app/one_click_unsubscribe/rest.py
```diff
@@ -0,0 +1,35 @@
...
+    notification = get_notification_by_id(notification_id)
+    create_unsubscribe_request_dao(notification)
+
+    current_app.logger.debug("Received unsubscribe request for notification_id: %s", notification_id)
+
+    return jsonify(result="success", message="Unsubscribe successful"), 200
+
+
+def generate_unsubscribe_url_token(data):
+    generate_token(str(data), current_app.config["SECRET_KEY"], current_app.config["DANGEROUS_SALT"])
```
leohemsted, line 35: this should return - probably need some more test if we didnt catch this yet

leohemsted (commented):
a few comments - i also wonder if we need a bit more test coverage here, mostly around making sure the unsubscribe request is stored in the db correctly, and the relationships work. eg: does it still work if theres a more recent template version, or the template is archived, etc? Are the created_at etc timestamps correct?

app/one_click_unsubscribe/rest.py
```diff
@@ -0,0 +1,35 @@
...
+    # Create unsubscribe request from the notification object
+    notification = get_notification_by_id(notification_id)
+    create_unsubscribe_request_dao(notification)
+
+    current_app.logger.debug("Received unsubscribe request for notification_id: %s", notification_id)
+
+    return jsonify(result="success", message="Unsubscribe successful"), 200
+
+
+def generate_unsubscribe_url_token(data):
```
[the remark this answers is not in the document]
BlessedDev: This is an error, It was part of a half thought process I had abandoned. The tests didn't catch it as it is not being used anywhere. I had intended to take it out and forgot.
I will add a `generate_unsubscribe_link_url` function as part of the work in [this](https://trello.com/c/6H3orCLk/33-add-list-unsubscribe-headers-to-emails-if-the-template-is-unsubscribable) ticket.

tests/app/one-click-unsubscribe/test_one_click_unsubscribe.py
BlessedDev: Good point. I will add a few more including ` template_version`

app/one_click_unsubscribe/rest.py
```diff
@@ -0,0 +1,35 @@
...
+
+    try:
+        check_token(token, current_app.config["SECRET_KEY"], current_app.config["DANGEROUS_SALT"], max_age_seconds)
+
+    except BadData as e:
+        errors = {"unsubscribe request": "This is not a valid unsubscribe link."}
+        raise InvalidRequest(errors, status_code=400) from e
+
+    # Create unsubscribe request from the notification object
+    notification = get_notification_by_id(notification_id)
```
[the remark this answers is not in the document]
BlessedDev: Good catch, I will deal with that scenario.

BlessedDev:
> a few comments - i also wonder if we need a bit more test coverage here, mostly around making sure the unsubscribe request is stored in the db correctly, and the relationships work. eg: does it still work if theres a more recent template version, or the template is archived, etc? Are the created_at etc timestamps correct?

I have added tests for two test cases:
i. the template is updated
ii. the template is redacted

I have made some assumptions in the tests and hopefully they are accurate.

tests/app/one_click_unsubscribe/test_one_click_unsubscribe.py
```diff
@@ -0,0 +1,116 @@
...
+    notification = create_notification(template=test_template, to_field="foo@bar.com")
+    initial_template_version = test_template.version
+    token = generate_token(notification.to, current_app.config["SECRET_KEY"], current_app.config["DANGEROUS_SALT"])
+
+    response = unsubscribe_url_post(client, notification.id, token)
+
+    # update template content to generate new template version
+    test_template.content = "New content"
+    test_template.process_type = "priority"
+    dao_update_template(test_template)
```
leohemsted, line 47: should this step should happen before we call `unsubscribe_url_post`?

so in order:

* create notification
* update template
* unsubscribe endpoint is called

tests/app/one_click_unsubscribe/test_one_click_unsubscribe.py
```diff
@@ -0,0 +1,116 @@
...
+    initial_template_version = test_template.version
+    notification = create_notification(template=test_template, to_field="foo@bar.com")
+    token = generate_token(notification.to, current_app.config["SECRET_KEY"], current_app.config["DANGEROUS_SALT"])
+
+    response = unsubscribe_url_post(client, notification.id, token)
+
+    # archive template
+    test_template.archived = True
+    dao_update_template(test_template)
+    subsequent_template_version = test_template.version
```
leohemsted, line 74: same point about this happening out of order as above

leohemsted (commented):
looks pretty good! just a couple of minor points about the order of some events in the tests and then i think this is good to go

tests/app/one_click_unsubscribe/test_one_click_unsubscribe.py
BlessedDev: Thanks for catching that. It was a mistake and I have corrected the order.

leohemsted (approved):

PR #4096, opened 2024-05-31
Delete unused `raise-alert-if-no-letter-ack-file` nightly task

klssmith:
We stopped receiving ack files from the DVLA when we switched to sending letters using their API, so no longer need the nightly task to check that we have them.

We will need to update Cronitor too to remove this task.

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -1,6 +1,5 @@
 from datetime import datetime, timedelta
 
-import pytz
 from flask import current_app
 from notifications_utils.clients.zendesk.zendesk_client import (
     NotifySupportTicket,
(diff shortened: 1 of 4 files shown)


app/celery/nightly_tasks.py
```diff
@@ -252,70 +251,6 @@ def get_letter_notifications_still_sending_when_they_shouldnt_be():
...
-                ticket_categories=["notify_letters"],
-            )
-            zendesk_client.send_ticket_to_zendesk(ticket)
-
-        current_app.logger.error(
-            "Letter ack file does not contain all zip files sent.",
-            extra=dict(
-                pdf_bucket=current_app.config["S3_BUCKET_LETTERS_PDF"],
-                subfolder=datetime.utcnow().strftime("%Y-%m-%d") + "/zips_sent",
-                ack_bucket=current_app.config["S3_BUCKET_DVLA_RESPONSE"],
```
idavidmcdonald, line 310: I did a search for `S3_BUCKET_DVLA_RESPONSE` in our code. I can see we use it in this file but also in research_mode_tasks.py in `create_fake_letter_response_file`... 

I could approve this PR as is... but if you wouldn't mind also seeing if you could tidy up `create_fake_letter_response_file` too as I think we might be able to get rid of it? It is doing some fun stuff where it creates fake letter response files in dev and preview and I'm not sure we want that anymore at all...

That might enable us to remove the `S3_BUCKET_DVLA_RESPONSE` key from our config entirely

klssmith: I think we do still need the `create_fake_letter_response_file` task, because the `S3_BUCKET_DVLA_RESPONSE` bucket is / was used for two things.

1. It was being used for ack files when we sent letters using FTP (the code that's being deleted here).
2. It is still being used for response files. These get put in the bucket via SNS, so the code isn't in this repo apart from when we want to fake that in preview and dev using `create_fake_letter_response_file`

idavidmcdonald: Oh yes, good point. Sorry for my confusion.

Let me have a proper look at this PR and will approve shortly

idavidmcdonald (approved):

PR #4097, opened 2024-05-31
Make dao_get_last_date_template_was_used only search within last year

CrystalPea:
To reduce time it takes to run the query and prevent delete_template page from timing out.

The query to check when template was last used would often time out (Sentry has record of 10 000 errors this caused - and it's a manual process to delete a template, so it's a lot of users seeing error page :< ).

One measure to prevent that is for us to only search within last year - that should be good enough for the user to be warned if they are deleting a well-used template by mistake.

Should go out after this PR: https://github.com/alphagov/notifications-admin/pull/5101

Story ticket: https://trello.com/c/RyEMRpQw/719-find-out-why-template-last-use-query-is-slow-and-fails

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -4,6 +4,7 @@ from itertools import groupby
 from operator import attrgetter
 
 from botocore.exceptions import ClientError
+from dateutil.relativedelta import relativedelta
 from flask import current_app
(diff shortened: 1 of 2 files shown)


whpearson (approved):
Test and changes make sense.

whpearson (approved):

PR #4098, opened 2024-05-31
Update code to use new Zendesk form 

klssmith:
Every ticket type is now a "task" and should have the `notify_task_type`
drop down menu populated instead of the `ticket_categories` dropdown,
which has been removed.

Needs https://github.com/alphagov/notifications-utils/pull/1121 to be merged first


whpearson (approved):
Looks like it makes sense with new utils

PR #4101, opened 2024-06-04
Make service callback sending consistent

idavidmcdonald:
Best reviewed commit by commit.

See https://trello.com/c/KQ6vP4EK/770-incident-action-review-if-we-can-make-inbound-sms-callbacks-more-similar-to-how-we-do-delivery-notification-callbacks for context

Very open to critique on my test coverage here please

### DIFF
--- a/app/celery/service_callback_tasks.py
+++ b/app/celery/service_callback_tasks.py
@@ -74,56 +74,15 @@ def send_inbound_sms_to_service(self, inbound_sms_id, service_id):
         "date_received": inbound_sms.provider_date.strftime(DATETIME_FORMAT),
     }
 
-    try:
-        response = request(
-            method="POST",
-            url=inbound_api.url,
-            data=json.dumps(data),
-            headers={"Content-Type": "application/json", "Authorization": "Bearer {}".format(inbound_api.bearer_token)},
-            timeout=60,
-        )
-        current_app.logger.debug(
-            "send_inbound_sms_to_service sending %s to %s, response %s",
-            inbound_sms_id,
-            inbound_api.url,
-            response.status_code,
-        )
-        response.raise_for_status()
-    except RequestException as e:
-        current_app.logger.warning(
-            "send_inbound_sms_to_service failed for service_id: %s for inbound_sms_id: %s and url: %s. exception: %s",
-            service_id,
-            inbound_sms_id,
-            inbound_api.url,
-            e,
-        )
-        if not isinstance(e, HTTPError) or e.response.status_code >= 500:
-            try:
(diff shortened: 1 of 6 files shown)


app/celery/tasks.py
```diff
@@ -527,71 +522,6 @@ def check_billable_units(notification_update):
...
-        current_app.logger.debug(
-            "send_inbound_sms_to_service sending %s to %s, response %s",
-            inbound_sms_id,
-            inbound_api.url,
-            response.status_code,
-        )
-        response.raise_for_status()
-    except RequestException as e:
-        current_app.logger.warning(
-            "send_inbound_sms_to_service failed for service_id: %s for inbound_sms_id: %s and url: %s. exception: %s",
```
klssmith, line 564: I find having the `service_id` in the log message like we had here helpful because it means that we can easily identify the services with failing callbacks from the logs. Without this, we'd have to go and look up the service in the database. Is there a way we can add this back if others also find it useful?

app/celery/service_callback_tasks.py
```diff
@@ -68,34 +94,34 @@ def _send_data_to_service_callback_api(self, data, service_callback_url, token,
...
             service_callback_url,
             e,
         )
         if not isinstance(e, HTTPError) or e.response.status_code >= 500 or e.response.status_code == 429:
             try:
-                self.retry(queue=QueueNames.CALLBACKS_RETRY)
-            except self.MaxRetriesExceededError:
+                self.retry(queue=retry_queue)
+            except self.MaxRetriesExceededError as e:
                 current_app.logger.warning(
```
klssmith, line 114: After this change we would get warnings not errors if the for `MaxRetriesExceededError`. That's probably fine, but just worth double checking. We don't consistently act on these logs, but I'm not sure if we consider Inbound SMS data "more important" than delivery status data since it does consist of actual messages and not just information about the messages.

app/celery/tasks.py
idavidmcdonald: Just pushed up a commit that should help with this

app/celery/service_callback_tasks.py
idavidmcdonald: Good spot.

I'm really undecided but I decide to make them all error level. It might mean more sentry errors, but we don't have a good way of dealing with this for the moment so I'd rather we keep getting told about it and so incentivised to come up with a better solution than just ignore the fact callbacks might be broken for service teams...

I've pushed up a new commit

idavidmcdonald:
Good spot again. Yep, definitely good to move it down to 5 seconds. 60 second timeouts are not nice as leave our apps hanging for ages if a service teams server is down

PR #4104, opened 2024-06-07
Add config option to set receipt callback urls for firetext and mmg

corlettb:
What
----

Add config option to set receipt callback urls for firetext and mmg.

FIRETEXT_RECEIPT_URL : Set firetext parameter 'receipts' when sending sms messages
MMG_RECEIPT_URL : Set mmg parameter 'devurl' when sending sms messages

These options override account wide callback urls for Firetext and MMG.

Why
----

Having the option to override these settings allows us to share accounts for development environments.

These options are not intended to be used in non-development accounts.

### DIFF
--- a/tests/app/conftest.py
+++ b/tests/app/conftest.py
@@ -11,6 +11,7 @@ from sqlalchemy.orm.session import make_transient
 
 from app import db
 from app.clients.sms.firetext import FiretextClient
+from app.clients.sms.mmg import MMGClient
 from app.constants import (
     EMAIL_TYPE,
     KEY_TYPE_NORMAL,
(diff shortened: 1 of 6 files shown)


tests/app/conftest.py
```diff
@@ -552,16 +553,47 @@ def mmg_provider():
...
+
+@pytest.fixture(scope="function")
+def mock_firetext_client_with_receipts(mocker):
+    additional_config = {"FIRETEXT_RECEIPT_URL": "https://www.example.com/notifications/sms/firetext"}
+    mock_config = create_mock_firetext_config(mocker, additional_config)
+    return create_mock_firetext_client(mocker, mock_config)
+
+
+@pytest.fixture(scope="function")
+def mock_mmg_client(mocker):
```
robinjam, line 589: Should this be:

```suggestion
def mock_mmg_client_with_receipts(mocker):
```

?

robinjam (approved):
LGTM, non-blocking suggestion

robinjam (approved):

PR #4105, opened 2024-06-07
Return billing data to API users via get_notification_by_id()

CrystalPea:
I recommend reviewing commit-by-commit, although there is some refactoring happening along the way.

We return following information:
cost_in_pounds: overall cost of the notification, sent as a string to avoid floating point issues. Alternatively we could return a dictionary with pounds and pence as separate key-value pairs.

I decided to go with a string, as I saw paypal doing the same.
I think either would be ok in the end. String is more human-readable I think, as this is how we would see prices in receipts, and it should be easy to process across different languages.

cost_details: extra info about the cost, which is:
* for letters: sheets_of_paper, postage
* for sms: sms_fragments, rate_multiplier, rate
* for emails: empty dict

For cost_details, we are getting rate information from other tables - rates and letter_rates tables.

I followed a pattern we have on User model, where we check permissions table in User.get_permissions().

These rates tables are very small, so hopefully checking them shouldn't cost us too much, but if you think it's a problem, I'm happy to think of alternative ways - maybe connect those two tables with notifications table, so they are readily available?

For letters in "pending-virus-check" status, that information won't be ready yet. So there is a boolean `"is_cost_data_ready"` field to indicate if cost information is ready. If it's not, we return None for cost, and empty dict for cost details.

SOME QUESTIONS:
- for each notification, we get date from [sms]`rates`/`letter_rates` table - that is another database call. Is there a way to make this more performant? Maybe connecting `letter/sms rates` tables to `notifications` table with a relationship?
- does the naming make sense?

Related docs PR: https://github.com/alphagov/notifications-tech-docs/pull/232


app/models.py
```diff
@@ -1643,6 +1643,57 @@ def serialize(self):
...
+    def _get_cost_info_for_sms(self, serialized):
+        serialized["cost_details"]["sms_fragments"] = self.billable_units
+        serialized["cost_details"]["rate_multiplier"] = self.rate_multiplier
+        sms_rate = self._get_sms_rate()
+        serialized["cost_details"]["rate"] = str(sms_rate)
+        serialized["cost_in_pounds"] = str(self.billable_units * self.rate_multiplier * sms_rate)
+
+        return serialized
+
+    def _get_cost_info_for_letter(self, serialized):
```
klssmith, line 1668: Templated letters start off with the `billable_units` set to 0 until the `create-pdf-for-templated-letter` task in template preview has been called and then has called `update-billable-units-for-letter` in api. So the cost data will not be ready immediately for any letters, not just precompiled ones.

app/models.py
```diff
@@ -1643,6 +1643,57 @@ def serialize(self):
...
+        if self.status == "pending-virus-check":  # are there any other not-ready states?
+            serialized["is_cost_data_ready"] = False
+            serialized["cost_details"] = {}
+            serialized["cost_in_pounds"] = None
+        else:
+            serialized["cost_details"]["sheets_of_paper"] = self.billable_units
```
idavidmcdonald, line 1674: So for letters we will expose `serialized["cost_details"]["sheets_of_paper"]` but for text messages there won't be a `sheets_of_paper` key in `cost_details`.

Is this OK?

Obviously sheets of paper isn't relevant for text messages, but does that mean we should not include the key or we should include the key but set the value as null?

Our precedent for the top level keys in a notification is to return all the keys possible, with them set as null if they aren't appropriate (see https://github.com/alphagov/notifications-api/blob/main/app/models.py#L1601).

I don't know why we do that though... One to find out the reason for I think. It might be that that pattern is 'better API design' according to the hive mind. Maybe it is because some API users might struggle to process responses if they don't have all the keys all the time (I'm thinking maybe .net or java here). Maybe it is because if we don't always include the keys then our users have to do logic like

`if key exists and get value of key` rather than just `get value of key` everytime they want to access a key in `cost_details`.

idavidmcdonald (commented):
A few comments that came to my mind when you walked us through this yesterday (thanks for doing that).

tests/app/test_model.py
```diff
@@ -223,6 +223,47 @@ def test_letter_notification_serializes_with_subject(client, sample_letter_templ
     assert res["subject"] == "Template subject"
 
 
+def test_notification_serialize_with_cost_info_for_sms(client, sample_template, sms_rate):
+    notification = create_notification(sample_template, billable_units=2)
+
+    response = notification.serialize_with_cost_info()
+
+    assert response["is_cost_data_ready"] is True
+    assert response["cost_details"] == {"sms_fragments": 2, "rate_multiplier": 1, "rate": "0.0227"}
```
klssmith, line 232: I haven't looked into what's behind the difference, but when trying this out locally I got the value of rate_multiplier as `1.0` instead of `1`

app/models.py
```diff
@@ -1643,6 +1643,57 @@ def serialize(self):
...
+            return self._get_cost_info_for_letter(serialized)
+
+        return serialized
+
+    def _get_cost_info_for_sms(self, serialized):
+        serialized["cost_details"]["sms_fragments"] = self.billable_units
+        serialized["cost_details"]["rate_multiplier"] = self.rate_multiplier
+        sms_rate = self._get_sms_rate()
+        serialized["cost_details"]["rate"] = str(sms_rate)
+        serialized["cost_in_pounds"] = str(self.billable_units * self.rate_multiplier * sms_rate)
```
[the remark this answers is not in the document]
CrystalPea: I don't know how we arrived at the initial decision, but I think it was a good one - pounds seem more intuitive - it's like we see the prices written usually.
And having pence instead still wouldn't solve floating point issue - because sms rates have resolution below one pence 😳

app/models.py
CrystalPea: ahh you are right - I will update the code to reflect this 👍🏼

CrystalPea: Done ☺️ . Together with an extended commit message to document it ☺️

app/models.py
```diff
@@ -1643,6 +1643,57 @@ def serialize(self):
...
+
+        return serialized
+
+    def _get_cost_info_for_letter(self, serialized):
+        if self.status == "pending-virus-check":  # are there any other not-ready states?
```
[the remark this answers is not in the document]
CrystalPea: I added that test now ☺️

app/models.py
CrystalPea: Ooh, good point - I will try to find out if my current approach works with different languages (I hope it does).

As to users having to do extra work to see if key doesn't exist - I'm afraid for cost info they might need to do some checks anyway:
- they will need to check the value for `"is_cost_data_ready"` - if this is False, the  "cost_in_pounds" is None instead of a string with number inside. This is on purpose - I didn't want users to assume the cost is 0.00 for notifications that are not ready.
- they will have to check if the notification is a letter or an sms to get the correct cost information (but they should know this beforehand as they are the ones sending the GET request, right?)

Anyway, that's a really interesting point and I will check about .net and java and about any API design guidelines on this 🙌🏼

tests/app/test_model.py
CrystalPea: Ahh, yes, I saw this, too, thanks for pointing out that it's not consistent - I will make it consistent 👍🏼

CrystalPea: I added a commit for this, with my reasoning in the commit message.

app/models.py
CrystalPea: I struggled to find blog posts on this, so I looked into the Paypal API again, to see if they have examples either way. Since they are such a huge company, their users use all sorts of languages.

I saw they for example have a `"supplementary_data"` key, where the value is a dictionary that can have a variety of key-value pairs. Look for it on this page if you're interested: https://developer.paypal.com/docs/api/payments/v2/

We already also return dictionaries with variety of key-value pairs inside already. We do this for personalisation - as some templates have no placeholders, and others have some and they are always different. Here is an example:
https://docs.notifications.service.gov.uk/java.html#get-a-template-by-id-response

So since the top level key is always there, and the value is always a dict, I vote that this is good enough, and actually imo neater than the alternative, and less likely to result in users mistakenly counting the cost of notification as 0 if they confuse the notification type.

app/models.py
```diff
@@ -1643,6 +1643,62 @@ def serialize(self):
...
+
+    def _get_letter_cost(self):
+        from app.dao.letter_rate_dao import dao_get_letter_rates_for_timestamp
+
+        rates = dao_get_letter_rates_for_timestamp(self.created_at)
+        letter_rate = next(
+            (rate for rate in rates if rate.sheet_count == self.billable_units and rate.post_class == self.postage),
+            None,
+        )
+        if self.billable_units == 0:
```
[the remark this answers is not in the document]
CrystalPea: Ahh, great catch, I pushed it now :3

tests/app/db.py
```diff
@@ -502,7 +505,7 @@ def create_email_branding(
     return email_branding
 
 
-def create_rate(start_date, value, notification_type):
+def create_sms_rate(start_date, value, notification_type):
```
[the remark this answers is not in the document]
CrystalPea: ohh you are right, I will change it back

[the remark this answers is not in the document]
CrystalPea: done

tests/app/test_model.py
```diff
@@ -223,6 +223,61 @@ def test_letter_notification_serializes_with_subject(client, sample_letter_templ
     assert res["subject"] == "Template subject"
 
 
+def test_notification_serialize_with_cost_info_for_sms(client, sample_template, sms_rate):
```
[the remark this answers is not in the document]
CrystalPea: Sure thing, I added a parametrize ☺️

tests/app/test_model.py
```diff
@@ -223,6 +223,61 @@ def test_letter_notification_serializes_with_subject(client, sample_letter_templ
...
+    notification = create_notification(sample_letter_template, billable_units=None, postage="first", status=status)
+
+    response = notification.serialize_with_cost_info()
+
+    assert response["is_cost_data_ready"] is False
+    assert response["cost_details"] == {}
+    assert response["cost_in_pounds"] is None
+
+
+@pytest.mark.parametrize("status", ["validation-failed", "permanent-failure", "technical-failure", "cancelled"])
```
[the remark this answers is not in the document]
CrystalPea: Done ☺️

app/models.py
```diff
@@ -1643,6 +1643,62 @@ def serialize(self):
...
         return serialized
 
+    def serialize_with_cost_info(self):
+        serialized = self.serialize()
+        serialized["cost_details"] = {}
+        serialized["cost_in_pounds"] = "0.00"
+        serialized["is_cost_data_ready"] = True
+
+        if self.notification_type == "sms":
+            return self._get_cost_info_for_sms(serialized)
```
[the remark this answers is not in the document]
klssmith: Thanks, I think that new naming makes things clearer for me 🙂

klssmith:
Thanks for making all the changes, there are a couple of final things I spotted that don't necessarily need changing but that we might want to check

1. It would be worth just double checking that we're 100% happy with the data that gets returned and its format. For example, for letters we return `postage` in `cost_details`, but there is an existing `postage` field that already gets returned. And the `cost_details` dict will return a rate for SMS but not for letters.

2. I noticed that for SMS we do need the `deliver_sms` task to have run before we have the final billable units. I think that's _probably_ fine - unlike with letters, that should happen pretty much instantly, so there would only be an issue if we had an incident or big delay in sending to a provider. Do you have thoughts on that?

3. Letters in a `cancelled` status still have billable units set, so there will be a `cost_in_pounds` for them that gets returned from the API. Letters in `technical-failure` may have billable units set - it depends how far the code had got when their status changed. If we don't want these statuses to show a cost, then we'll need to tweak the code

CrystalPea:
Hey @klssmith 

> It would be worth just double checking that we're 100% happy with the data that gets returned and its format. For example, for letters we return postage in cost_details, but there is an existing postage field that already gets returned. 

Yeah, that's a conscious decision - I know it's also a top level key, but decided to also put it in he cost_info, as it's part of how the cost is decided for letters

> And the cost_details dict will return a rate for SMS but not for letters.
That's because for letters the rate is always the same as the cost - happy to return it if we think it would make things clearer, but I feel it could make things muddier - people could start asking us if the rate and the cost is always the same or not for example.

> I noticed that for SMS we do need the deliver_sms task to have run before we have the final billable units. I think that's probably fine - unlike with letters, that should happen pretty much instantly, so there would only be an issue if we had an incident or big delay in sending to a provider. Do you have thoughts on that?

Good spot - I could make a similar guard statement for SMS like we have for letters just in case - that is if SMS is in "created" or "sending" status and the billable units are 0, we say cost data not ready yet?

> Letters in a cancelled status still have billable units set, so there will be a cost_in_pounds for them that gets returned from the API. Letters in technical-failure may have billable units set - it depends how far the code had got when their status changed. If we don't want these statuses to show a cost, then we'll need to tweak the code

Good spot again - I will amend the code and write a test for this.

app/models.py
leohemsted: i am in favour of not including keys for fields that do not make semantic sense

how we handle this in client libraries of strongly typed languages is a separate issue, but probably solved by polymorphism. eg java client could return a `CostData` object which is of type `LetterCostData` or `SMSCostData`

leohemsted: i'm interested in whether `cost_in_pounds` should be nested within `cost_data`. i like the idea of being able to retrieve the entire cost data from the json blob in one line, and the actual value _is_ data about the cost

app/models.py
```diff
@@ -1642,6 +1644,86 @@ def serialize(self):
...
+
+    def _add_cost_data_for_sms(self, serialized):
+        if not self._is_cost_data_ready_for_sms():
+            serialized["is_cost_data_ready"] = False
+            serialized["cost_details"] = {}
+            serialized["cost_in_pounds"] = None
+        else:
+            serialized["cost_details"]["billable_sms_fragments"] = self.billable_units
+            serialized["cost_details"]["international_rate_multiplier"] = self.rate_multiplier
+            sms_rate = self._get_sms_rate()
```
leohemsted, line 1668: are we worried about extra DB queries - should this be cached?

(i'm mostly wondering what happens if someone makes a get request to get the status of all notifications, or if we end up using this code in other bulk ways). 

might be something we just note in a TODO comment and do nothing about for now

app/models.py
```diff
@@ -1642,6 +1644,86 @@ def serialize(self):
...
+            serialized["cost_details"]["rate"] = str(sms_rate)
+            serialized["cost_in_pounds"] = str(self.billable_units * self.rate_multiplier * sms_rate)
```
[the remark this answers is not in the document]
CrystalPea: I don't think you explained badly, I see your point. 

I still think we should send the value as string, since we are sending it via json and the value is more stable that way.

Here is a thread on this in Stackoverflow: https://stackoverflow.com/questions/35709595/why-would-you-use-a-string-in-json-to-represent-a-decimal-number

(I should probably include it in the commit message if we go with this 😳 )

Happy to have a call about this (tomorrow?)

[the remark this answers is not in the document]
CrystalPea: Hmm... but you have a point we send rather simple and small numbers, and we do no operations on them, so maybe float is ok, too - let's chat.

app/models.py
CrystalPea: I was worried a little bit - I think it would be a good idea to cache it actually - what we'd do as a cache key, `created_at` cast to date (without time)?

And for letters we'd have to do the date, postage, and sheets, but also could work? I'm happy to add this today.

Just let's work together to keep review cycle short and get it out before I go on holiday hahah :D

CrystalPea: Following conversation:
- for now add get_cache for redis, check if it also automatically sets when value not there.
- in the future think about moving to serialized models

app/models.py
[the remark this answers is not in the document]
CrystalPea: Outcome: 

Both string and float are valid options, and both have been used by big players APIs - for example Twilio uses a float, while Paypal uses a string.

Leo has slight preference for float, since it means fewer conversions on both ours and our users' sides, so let's go with that.

app/models.py
CrystalPea: I placed `cost_in_pounds` at the top level to make it's easily accessible to the users. My assumption here is that most users will be only interested in the cost itself, while `cost_details` is more like supplementary information in case they need to do some deeper digging.

I would prefer to receive it like that if I were a user.

app/models.py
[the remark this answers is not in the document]
CrystalPea: This is changed now, and in the docs, too.

app/models.py
CrystalPea: Caching is added for both getting sms rate, and getting letter rates now.

app/models.py
joybytes: curious minor question: why explicitly `cost_in_pounds`, do we calculate in any other currency?

klssmith (approved):
Looks like all the comments are addressed (I can see the comment about data structure was discussed elsewhere). Good to go! 🎉

PR #4108, opened 2024-06-12
Bump requests to the latest version

quis:
Versions 2.32.0 and 2.32.1 were yanked and are no longer available from PyPI


whpearson (approved):
Tested locally. Changelog doesn't look concerning

PR #4109, opened 2024-06-12
Adding protected Sender ids, scripts, DB tbl endpoint and test

whpearson:
This has

1) A script for processing csvs from BT to give insert statements 
2) A database table called protected_sender_ids that holds a list of
   sender_ids.
3) A protected_sender_id/check get endpoint that can be used by admin to
   check whether a sender_id is protected (true, i.e. in the list) or not.
4) Tests for that endpoint

### DIFF
--- a/app/protected_sender_id/rest.py
+++ b/app/protected_sender_id/rest.py
@@ -1,0 +1,27 @@
+from flask import Blueprint, jsonify, request
+
+from app.errors import InvalidRequest, register_errors
+from app.models import ProtectedSenderId
+from app.protected_sender_id.protected_sender_schema import protected_sender_request
+from app.schema_validation import validate
+
+protected_sender_id_blueprint = Blueprint("protected-sender-id", __name__)
+
+register_errors(protected_sender_id_blueprint)
+
+
+@protected_sender_id_blueprint.route("/check")
+def check_if_sender_id_is_protected():
+    if request.args:
+        validate(request.args, protected_sender_request)
+
+    sender_id = request.args.get("sender_id", "")
+    if sender_id == "":
+        raise InvalidRequest(message="sender_id must be passed in", status_code=400)
+    normalised_sender_id = "".join(sender_id.split()).lower()
+    result = ProtectedSenderId.query.filter(ProtectedSenderId.sender_id == normalised_sender_id).scalar()
+    if result:
+        is_protected_sender_id = True
+    else:
+        is_protected_sender_id = False
+    return jsonify(is_protected_sender_id)
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,3 @@
+The scripts in this directory are used to populate prod like environments in the following fashion:
+
+- `cat $CSV_DOWNLOADED_FROM_BT | python process_sender_names.py`  - this script is run by hand. It takes a list of protected sender names that we get from BT and uses that list to populate a db table. A sanity check of the insert statement is needed before running. The table is used to check if a sender name our user wants to set is protected. If it's one of the protected ones, we don't set it, and we also raise an alert in a form of a Zendesk ticket (code for that is in admin repo). The script filters out all government and healthcare sender names as we want our service users to be able to use them.
--- a/scripts/process_sender_names.py
+++ b/scripts/process_sender_names.py
@@ -1,0 +1,29 @@
+# ruff: noqa: T201
+import csv
+import sys
+
+print("Processing header names")
+
+
+reader = csv.reader(sys.stdin)
+headers = next(reader)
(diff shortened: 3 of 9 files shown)


scripts/README.md
```diff
@@ -0,0 +1,3 @@
+The scripts in this directory are used to populate prod like environments in the following fashion:
+
+- `cat $CSV_DOWNLOADED_FROM_BT | python process_sender_names.py`  - emits an insert statement that has a list of senderids to populate a table that lists those that are protected from being set. Admin will also raise a zendesk ticket if someone tries to set their senderid to one of those protected. It filters out all government and healthcare sender ids as we want those to be able to be set by our users.
```
CrystalPea, line 3:
```suggestion
- `cat $CSV_DOWNLOADED_FROM_BT | python process_sender_names.py`  - this script takes a list of protected sender names that we get from BT and uses that list to populate a db table. We then use that table to check if a sender name our user wants to set is protected. If it's one of the protected ones, we don't set it, and we also raise an alert in a form of a Zendesk ticket (code for that is in admin repo). The script filters out all government and healthcare sender names as we want our service users to be able to use them.
```

scripts/process_sender_names.py
```diff
@@ -0,0 +1,29 @@
...
+            new_sender_ids = line[column_for_sender_id].lower().split(",")
+            # Do the split/join dance to remove whitespace
+            new_sender_ids = map(lambda x: "".join(x.split()), new_sender_ids)
+            # Sometimes there are trailing commas remove them
+            new_sender_ids = filter(lambda x: x != "", new_sender_ids)
+            sender_ids = sender_ids + list(new_sender_ids)
+
+
+joined_sender_ids = "'),('".join(sender_ids)
+print(f"INSERT INTO protected_sender_ids VALUES ('{ joined_sender_ids}') ON CONFLICT DO NOTHING;")
```
CrystalPea, line 29: do we use print to do a database insert? That's quite unusual, no?

scripts/README.md
CrystalPea: I edited the script description for ease of reading, I hope it's ok.

Also are we going to run this script manually or programmatically? It's not mentioned.

CrystalPea (changes requested):
Hey, nice one, I left some comments - mainly things about readability/naming.

Another thing is - everything is a single commit, but it would be good to have migration and model as a separate commit at the start. I remember in the past we wanted migration to go in as a separate PR entirely - in case we need to roll back, it makes things easier. I'm not sure if that's still the case - tagging @leohemsted and @idavidmcdonald to check if we still require this.

app/protected_sender_id/rest.py
```diff
@@ -0,0 +1,29 @@
...
+
+register_errors(protected_sender_id_blueprint)
+
+
+@protected_sender_id_blueprint.route("/check")
+def check_if_sender_id_is_protected():
+    if request.args:
+        validate(request.args, protected_sender_request)
+
+    # If start and end date are not set, we are expecting today's stats.
```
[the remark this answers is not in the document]
whpearson: Leftover from copy paste. Thanks

scripts/process_sender_names.py
whpearson: I'm thinking the query might want to be seen before it is run in production. In case there is something wrong with the processing of the data.

whpearson:
I've added some documentation here. https://github.com/alphagov/notifications-manuals/wiki/Protected-sender-ids

scripts/README.md
```diff
@@ -0,0 +1,3 @@
+The scripts in this directory are used to populate prod like environments in the following fashion:
+
+- `cat $CSV_DOWNLOADED_FROM_BT | python process_sender_names.py`  - this script is run by hand takes a list of protected sender names that we get from BT and uses that list to populate a db table. A sanity check of the insert statement is needed before running. The table is used to check if a sender name our user wants to set is protected. If it's one of the protected ones, we don't set it, and we also raise an alert in a form of a Zendesk ticket (code for that is in admin repo). The script filters out all government and healthcare sender names as we want our service users to be able to use them.
```
CrystalPea, line 3:
```suggestion
- `cat $CSV_DOWNLOADED_FROM_BT | python process_sender_names.py`  - this script is run by hand. It takes a list of protected sender names that we get from BT and uses that list to populate a db table. A sanity check of the insert statement is needed before running. The table is used to check if a sender name our user wants to set is protected. If it's one of the protected ones, we don't set it, and we also raise an alert in a form of a Zendesk ticket (code for that is in admin repo). The script filters out all government and healthcare sender names as we want our service users to be able to use them.
```

CrystalPea (approved):

PR #4111, opened 2024-06-17
Generate unsubscribe links

BlessedDev:
This PR adds List-Unsubscribe headers to emails where `template.has_unsubscribe_link` is `True` as specified in [this](https://trello.com/c/6H3orCLk/33-add-list-unsubscribe-headers-to-emails-if-the-template-is-unsubscribable) card.

I have added logic to the `persist_notifications` function to generate an `unsubscribe_link` if `template.has_unsubscribe_link` is `True` and an `unsubscribe_link` has not been provided, ie via an API call.

**Subscription template**
subscription template

**Email metadata**
email metadata


app/celery/tasks.py
```diff
@@ -294,11 +295,15 @@ def save_api_email_or_sms(self, encoded_notification):
     provider_task = (
         provider_tasks.deliver_email if notification["notification_type"] == EMAIL_TYPE else provider_tasks.deliver_sms
     )
+    # If a template.has_unsubscribe_link is True but an unsubscribe link is not provided via the API, we will
+    # generate an unsubscribe link for the notification.
+    template = dao_get_template_by_id(notification["template_id"], notification["template_version"])
```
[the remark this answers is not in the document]
BlessedDev: Good point. I will update `save_email_or_sms_to_queue` to add `template_has_unsubscribe_link` to the `encoded notification`

quis:
This also needs to work for one-off emails, which get persisted here:
https://github.com/alphagov/notifications-api/blob/2a14ae85a9dd4e94c81a9b9b331325650caf5870/app/notifications/rest.py#L102-L104

…and emails sent by uploading spreadsheets, which get persisted here:
https://github.com/alphagov/notifications-api/blob/2a14ae85a9dd4e94c81a9b9b331325650caf5870/app/celery/tasks.py#L249-L251

BlessedDev:
> This also needs to work for one-off emails, which get persisted here:
> 
> https://github.com/alphagov/notifications-api/blob/2a14ae85a9dd4e94c81a9b9b331325650caf5870/app/notifications/rest.py#L102-L104
> 
> …and emails sent by uploading spreadsheets, which get persisted here:
> 
> https://github.com/alphagov/notifications-api/blob/2a14ae85a9dd4e94c81a9b9b331325650caf5870/app/celery/tasks.py#L249-L251

I have already updated both calls to `persist_notification` to include `template_has_unsubscribe_link=template.has_unsubscribe_link`

leohemsted (commented):
A few ideas on how to change some things for readability, and tidy up the tests a bit - at the moment im not in love with those two new fixtures

tests/app/notifications/test_process_notification.py
```diff
@@ -529,3 +531,60 @@ def test_persist_notification_for_international_letter(sample_letter_template, p
...
+        template_id=job.template.id,
+        template_version=job.template.version,
+        template_has_unsubscribe_link=job.template.has_unsubscribe_link,
+        recipient=test_email_address,
+        service=job.service,
+        personalisation=None,
+        notification_type=EMAIL_TYPE,
+        api_key_id=sample_api_key.id,
+        key_type=sample_api_key.key_type,
+        job_id=job.id,
```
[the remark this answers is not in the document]
BlessedDev: I had followed the pattern from `test_persist_email_notification_stores_normalised_email`. I will adjust both tests to have only the necessary parameters.

app/notifications/process_notifications.py
```diff
@@ -214,3 +219,9 @@ def simulated_recipient(to_address, notification_type):
         return to_address in formatted_simulated_numbers
     else:
         return to_address in current_app.config["SIMULATED_EMAIL_ADDRESSES"]
+
+
+def generate_unsubscribe_link(notification_id: str, email_address: str) -> str:
+    token = generate_token(email_address, current_app.config["SECRET_KEY"], current_app.config["DANGEROUS_SALT"])
+    notify_base_url = current_app.config["API_HOST_NAME"]
+    return f"{notify_base_url}/unsubscribe/{notification_id}/{token}"
```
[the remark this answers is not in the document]
BlessedDev: This is a good idea, I will tidy up `url_with_token` and use that instead.

app/v2/notifications/post_notifications.py
```diff
@@ -302,7 +302,9 @@ def save_email_or_sms_to_queue(
     encoded = signing.encode(data)
```
leohemsted, line 302: just realised we could write something like:
```suggestion
    # include template_has_unsubscribe_link in the call to the task but dont try and return it from the function in the Notification object
    encoded = signing.encode(data | {'"template_has_unsubscribe_link": template.has_unsubscribe_link})
```

tests/app/notifications/test_process_notification.py
```diff
@@ -529,3 +531,60 @@ def test_persist_notification_for_international_letter(sample_letter_template, p
...
+    unsubscribe_link,
+    template_has_unsubscribe_link,
+):
+    """
+    If an unsubscribe link is provided, no unsubscribe link will be generated and the provided
+    unsubscribe link is set on the notification.
+    An unsubscribe link will only be generated if no unsubscribe link has been provided
+    and template.has_unsubscribe_link is True
+    """
+    if template_has_unsubscribe_link:
```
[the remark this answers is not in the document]
BlessedDev: I have split this out into two separate tests one to cover test cases for when `template.has_unsubscribe_link` is True and the other for when it is false.

app/celery/tasks.py
```diff
@@ -294,11 +295,13 @@ def save_api_email_or_sms(self, encoded_notification):
     provider_task = (
         provider_tasks.deliver_email if notification["notification_type"] == EMAIL_TYPE else provider_tasks.deliver_sms
     )
+
     try:
         persist_notification(
             notification_id=notification["id"],
             template_id=notification["template_id"],
             template_version=notification["template_version"],
+            template_has_unsubscribe_link=notification["template_has_unsubscribe_link"],
```
leohemsted, line 304: i just realised, we can reuse the SerialisedTemplate code from `save_email` above - and then we can just use `template.has_unsubscribe_link` here and we won't need to pass it through the encoded notification stuff! that'll make this a lot simpler and save us editing a whole bunch of tests 🙏

https://github.com/alphagov/notifications-api/blob/8353ed0fd15549407ad45b6c491bfcbb2b2e5f41/app/celery/tasks.py#L238-L242

```suggestion
            template_has_unsubscribe_link=template.has_unsubscribe_link,
```

leohemsted: also this is the code that i'm proposing to remove so hopefully this whole function can go away soon and then we dont have to change this 🤞

tests/app/celery/test_tasks.py
```diff
@@ -1733,7 +1733,7 @@ def test_save_api_email_or_sms(mocker, sample_service, notification_type):
         data.update({"to": "+447700900855"})
         expected_queue = QueueNames.SEND_SMS
 
-    encoded = signing.encode(data)
+    encoded = signing.encode(data | {"template_has_unsubscribe_link": False})
```
leohemsted, line 1736: if we do need to keep these tests, then i think it'd be a bit more readable to just include `"template_has_unsubscribe_link": False` in the main data block 🤔

leohemsted (commented):
just realised we migiht not need to deal with the encoded notification stuff at all! 🤞

app/celery/tasks.py
BlessedDev: @leohemsted 
I had considered it previously, at the time it seemed more efficient to add the `template.has_unsubscribe_link` value to `encoded_notification` rather than pulling in the `SerialisedTemplate` just to extract the same value.
Since `save_api_email` is about to be removed, I think I will remove this code and all the changes made to support it and not deploy the work done in this PR until after `save_api_email` has been removed.

tests/app/celery/test_tasks.py
BlessedDev: I will remove them entirely as they will be redundant soon any way, as `save_api_email` is about to be removed.

tests/app/notifications/test_process_notification.py
```diff
@@ -529,3 +529,89 @@ def test_persist_notification_for_international_letter(sample_letter_template, p
...
+        ("https://please-unsubscribe-me.com/unsubscribe", "https://please-unsubscribe-me.com/unsubscribe"),
+        (None, "https://notify-generated-super-unsubscribe-link.com/unsubscribe"),
+    ],
+)
+def test_persist_notification_when_template_has_unsubscribe_link_is_true(
+    unsubscribe_link, mocker, sample_api_key, sample_service, expected_unsubscribe_link
+):
+    """
+    Tests that if template.has_unsubscribe_link is True that an unsubscribe link is generated by persist_notification
+    only if one hasn't already been provided.
```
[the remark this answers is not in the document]
BlessedDev: Thank you.

app/notifications/process_notifications.py
```diff
@@ -120,6 +122,12 @@ def persist_notification(
     notification_created_at = created_at or datetime.utcnow()
     if not notification_id:
         notification_id = uuid.uuid4()
+
+    if template_has_unsubscribe_link and not unsubscribe_link:
+        base_url = current_app.config["API_HOST_NAME"]
+        url = f"/unsubscribe/{str(notification_id)}/"
+        unsubscribe_link = url_with_token(recipient, url=url, base_url=base_url)
```
CrystalPea, line 129: ohh, so recipient data will be hashed, but available to us later, correct?

app/utils.py
```diff
@@ -49,12 +51,10 @@ def get_prev_next_pagination_links(current_page, next_page_exists, endpoint, **k
     return links
 
 
-def url_with_token(data, url, config, base_url=None):
-    from notifications_utils.url_safe_token import generate_token
-
-    token = generate_token(data, config["SECRET_KEY"], config["DANGEROUS_SALT"])
-    base_url = (base_url or config["ADMIN_BASE_URL"]) + url
-    return base_url + token
+def url_with_token(data, url, base_url=None):
```
CrystalPea, line 54: nice refactor 👏🏼

tests/app/notifications/test_process_notification.py
```diff
@@ -529,3 +529,89 @@ def test_persist_notification_for_international_letter(sample_letter_template, p
...
     assert persisted_notification.international
+
+
+@pytest.mark.parametrize(
+    "unsubscribe_link",
+    ["https://please-unsubscribe-me.com/unsubscribe", None],
+)
+def test_persist_notification_when_template_has_unsubscribe_link_is_false(
+    unsubscribe_link, mocker, sample_job, sample_api_key
+):
```
CrystalPea, line 540: maybe we could have a short docstring for this test, too, like we do for one below - to explain what w're testing? For example:
```
If template_has_unsubscribe_link is False, we either use the unsubscribe_link provided by the user, or set it to None.
```

CrystalPea (approved):
Looks good to me 👍🏼

app/notifications/process_notifications.py
BlessedDev: Encoding the recipient data is primarily to ensure the validity of an unsubscribe request. The endpoint checks that it is a valid token.
The recipient data would also be available on the `notification` object.

app/utils.py
BlessedDev: > nice refactor 👏🏼

Thank you.

PR #4114, opened 2024-06-18
Pin requests due to bug with connecting to DVLA

idavidmcdonald:
We updated requests in https://github.com/alphagov/notifications-api/pull/4108 but we saw errors in production connecting to DVLA as per https://github.com/psf/requests/issues/6730.

It looks like moving back to 2.32.2 should be unaffected, at least according to the github issue above.

Once a new version is released that fixes it, we can drop this pin. If it doesn't get fixed we could look at an alternative approach like https://github.com/alphagov/notifications-api/pull/4113

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -16,6 +16,8 @@ itsdangerous==2.1.2
 jsonschema[format]==4.16.0
 marshmallow-sqlalchemy==0.28.1
 marshmallow==3.18.0
+# Requests pinned to 2.32.2 until https://github.com/psf/requests/issues/6730 is fixed. Once so, we can remove this pin
+requests==2.32.2
 psycopg2-binary==2.9.6
 PyJWT==2.5.0
 SQLAlchemy==1.4.41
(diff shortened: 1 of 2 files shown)


leohemsted:
have we been able to test this locally to confirm it's a solution?

i assume we should be able to test this by just setting up the SSL settings the same as we do in api and making a request to `https://driver-vehicle-licensing.api.gov.uk/thirdparty-access/v1/authenticate` and making sure it doesn't error

corlettb (approved):

idavidmcdonald:
Good point.

Verified it using the following script with versions 2.32.2 and 2.32.3 of requests. I replicate the error in 2.32.3 and get no exception in 2.32.2

```
from requests.adapters import HTTPAdapter
from app.clients.letter.dvla import _SpecifiedCiphersAdapter
import requests

ciphers = "TLS_AES_128_GCM_SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-SHA256"

session = requests.Session()
session.mount("https://driver-vehicle-licensing.api.gov.uk", _SpecifiedCiphersAdapter(ciphers=ciphers))
response = session.get("https://driver-vehicle-licensing.api.gov.uk/thirdparty-access/v1/authenticate")
print(response)
```

PR #4115, opened 2024-06-18
Remove high volume worker

leohemsted:
Here are some reasons I think we should remove it:

1. There’s two queues (save-api-email-tasks and save-api-sms-tasks and a worker (api-worker-save-api-notifications)  that we pay money for
1. Code complexity - this was raised when @BlessedDev tried to update when we save notifications to handle unsubscribe links (which we will need to make sure work for [GOV.UK](http://gov.uk/) Email!), and it’s more nuanced because this json blob is passed on a celery task.
1. We have other services we consider as large such as NHS login, NHS Notify, DWP Prod, that we do not have on this queue and we have been fine
1. I don’t think we’ve ever tested that this actually increased performance of Notify. Our current DB connection usage is pretty low and even at peak times we’re no-where near our hard limit of 5000 concurrent connections.
We return a 200 OK, and give a user a notification ID, but don’t actually create the notification in the database. If something goes wrong (eg the task is lost off the queue/worker) then we won’t have created the notification but will have told the user that we did! that’s really bad.
1. Even if nothing goes wrong, in that period while the task is on the queue, if the service does a GET notification by id request, they’ll receive a 404 not found because the notification won’t be in the DB. This is super weird and confusing.

Deploy process:

- [x] remove services from creds so path isnt used https://github.com/alphagov/notifications-credentials/pull/461
- [x] remove code path https://github.com/alphagov/notifications-api/pull/4124
- [x] terraform delete ecs task, ssm creds. delete from concourse pipeline at the same time? https://github.com/alphagov/notifications-aws/pull/2276
- [x] delete the cred files completely https://github.com/alphagov/notifications-credentials/pull/463
- [ ] clickops remove sqs queues
- [ ] remove worker definition from API <-- YOU ARE HERE
- [ ] figure out what to do with high-volume vs not-high-volume statsd metrics

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -25,8 +25,6 @@ class QueueNames:
     SMS_CALLBACKS = "sms-callbacks"
     ANTIVIRUS = "antivirus-tasks"
     SANITISE_LETTERS = "sanitise-letter-tasks"
-    SAVE_API_EMAIL = "save-api-email-tasks"
-    SAVE_API_SMS = "save-api-sms-tasks"
     BROADCASTS = "broadcast-tasks"
     GOVUK_ALERTS = "govuk-alerts"
 
@@ -49,8 +47,6 @@ class QueueNames:
             QueueNames.LETTERS,
             QueueNames.SES_CALLBACKS,
             QueueNames.SMS_CALLBACKS,
-            QueueNames.SAVE_API_EMAIL,
-            QueueNames.SAVE_API_SMS,
             QueueNames.BROADCASTS,
         ]
 
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -56,9 +56,6 @@ case "$1" in
   api-worker-service-callbacks)
     exec $COMMON_CMD service-callbacks,service-callbacks-retry
     ;;
-  api-worker-save-api-notifications)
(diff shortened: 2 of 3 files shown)


app/config.py
```diff
@@ -25,8 +25,6 @@ class QueueNames:
     SMS_CALLBACKS = "sms-callbacks"
     ANTIVIRUS = "antivirus-tasks"
     SANITISE_LETTERS = "sanitise-letter-tasks"
-    SAVE_API_EMAIL = "save-api-email-tasks"
-    SAVE_API_SMS = "save-api-sms-tasks"
```
joybytes, line 29: not sure if this code still in use but it seems there is mention of these Queues [here](https://github.com/alphagov/notifications-api/blob/d0f90b70a8ac83f00df581a484e5eea0dd309983/app/v2/notifications/post_notifications.py#L304)

leohemsted: think this branch could be rebased but those lines of code were removed already here:

https://github.com/alphagov/notifications-api/pull/4124/files#diff-cc69ccd3b3b0ba4e44f73c182412190d6af4e52c1bc0680fdc745a60c2eced22L305

leohemsted: i've now rebased 👍

joybytes (approved):

PR #4117, opened 2024-06-20
add unsubscribe link to save_api_email_or_sms

BlessedDev:
This PR fixes a bug that was noticed by a user as described in [this](https://trello.com/c/VRzsixUp/59-add-unsubscribelink-to-saveapiemailorsms) card, by adding `unsubscribe_link` to the persisted notification in `save_api_email_or_sms`.

I have added an assertion to `test_save_api_email_or_sms` to test that a provided `unsubscribe_link` is set on the notification.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -305,6 +305,7 @@ def save_api_email_or_sms(self, encoded_notification):
             reply_to_text=notification["reply_to_text"],
             status=notification["status"],
             document_download_count=notification["document_download_count"],
+            unsubscribe_link=notification["unsubscribe_link"],
         )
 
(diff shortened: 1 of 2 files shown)


BlessedDev:
> It looks like we don't add unsubscribe links to saving emails sent not via the API at the moment. https://github.com/alphagov/notifications-api/pull/4117/files#diff-cca5917c620894fe770faa88835e9e186285d5b91a55fca992eb94d5dc87e635L263 I'm guessing that is another card. This makes sense and I'm happy to approve if save_email is another card.

@whpearson Yes, there is [WIP](https://trello.com/c/6H3orCLk/33-add-list-unsubscribe-headers-to-emails-if-the-template-is-unsubscribable) to add unsubscribe links to emails not sent by the API. Adding Unsubscribe links to email headers are an API only feature at the moment.

whpearson (approved):

PR #4118, opened 2024-06-21
fix(utils): change archived field nomanclature from date to datetime

joybytes:
## PR Summary
- by archiving service fields with a date suffix, this means that if we archive more than one service with the same name in one day, we get a constraint error
- here, this is being change to datetime to avoid the issue

### Related Tasks
[Change archive name format from date to datetime](https://trello.com/c/xAjLBYtE/794-change-archive-name-format-from-date-to-datetime)

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -134,8 +134,8 @@ def email_address_is_nhs(email_address):
 
 
 def get_archived_db_column_value(column):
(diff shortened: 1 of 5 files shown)


idavidmcdonald (approved):
Looks good, Joy. Nice work!

Try and grab someone after a standup to talk you through how our deployments work and how to deploy this. They can probably pair with you on clicking the merge button.

PR #4120, opened 2024-06-25
Migration to add new first and second class letter rates

klssmith:
https://trello.com/c/LTcHxYV9/819-update-new-letter-prices-on-notify

This adds the new rates for first and second class letters that will be valid from 1 July 2024. There will be no changes to the international letter prices.


klssmith:
I'm not sure why the Squawk check is complaining with `Postgres failed to parse query`. This has run fine locally, and we've run similar migrations in the past. This migration should be very safe since we're only inserting 20 rows.

CrystalPea (approved):
Everything looks good, the rates match the spreadsheet, and it runs well both on upgrade and downgrade.

:shipit:

PR #4121, opened 2024-06-25
Remove some dependencies which don’t need to be in requirements files

quis:
If a dependency is already:
- specified as a subdependency of one of our other dependencies
- spcified in the dependencies of `notifications-utils`

…then we don’t need to also specify it in `requirements.in`, unless we are trying to pin it to a specific version

***

A bunch of our test dependencies are now specified in [`notifications-utils/requirements_for_test_common.txt`](https://github.com/alphagov/notifications-utils/blob/f59616870b5457d5441863366d8c99af8507ca28/requirements_for_test_common.txt) so can be removed from `requirements_for_test.txt`


docker/Dockerfile
```diff
@@ -117,9 +117,11 @@ COPY --chown=notify:notify . .
 
 ENV SQLALCHEMY_DATABASE_URI=postgresql://postgres:postgres@localhost/
 
-RUN pip3 install -r requirements_for_test.txt
+RUN pip3 install -r requirements.txt
```
leohemsted, line 120: we shouldn't need this line (as we copy the prod requirements across on line 112, which includes utils etc)

quis: It needs both to:
- install app requirements (including utils)
- use utils to copy `requirements_for_test_common.txt`
- install test requirements including those in `requirements_for_test_common.txt`

leohemsted (approved):

PR #4122, opened 2024-06-26
Remove org permissions when archiving a user

klssmith:
When archiving a user, we were removing them from their services and removing all service permissions. We were also removing them from any organisations that they were a member of, but we were not removing their organisation permissions which gave an error when trying to serialize the user.

This change ensures we now remove an archived user's organisation permissions too.

### DIFF
--- a/tests/app/dao/test_users_dao.py
+++ b/tests/app/dao/test_users_dao.py
@@ -7,7 +7,8 @@ from sqlalchemy.exc import DataError
 from sqlalchemy.orm.exc import NoResultFound
 
 from app import db
-from app.constants import EMAIL_AUTH_TYPE
+from app.constants import EMAIL_AUTH_TYPE, OrganisationUserPermissionTypes
+from app.dao.organisation_user_permissions_dao import organisation_user_permissions_dao
 from app.dao.service_user_dao import (
     dao_get_service_user,
     dao_update_service_user,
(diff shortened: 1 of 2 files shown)


tests/app/dao/test_users_dao.py
```diff
@@ -221,8 +222,18 @@ def test_dao_archive_user(sample_user, sample_organisation, fake_uuid):
     create_permissions(sample_user, service_2, "view_activity")
     create_permissions(service_2_user, service_2, "manage_settings")
 
-    # make sample_user an org member
+    # make sample_user an org member with permissions
     sample_organisation.users = [sample_user]
+    new_permissions = [
+        OrganisationUserPermissions(user=sample_user, organisation=sample_organisation, permission=p)
+        for p in [OrganisationUserPermissionTypes.can_make_services_live.value]
+    ]
```
quis, line 230: Don’t think you need the `for` with just 1 item:
```suggestion
    new_permissions = [
        OrganisationUserPermissions(user=sample_user, organisation=sample_organisation, permission=OrganisationUserPermissionTypes.can_make_services_live.value)
    ]
```

quis (approved):
One optional simplification if you want it

tests/app/dao/test_users_dao.py
klssmith: That's much nicer, so I've rebased the change in (you'll need to re-approve please)

quis (approved):

PR #4123, opened 2024-07-02
Send inbound sms callback tasks to service-callback queue

klssmith:
The service-callbacks worker now has database access: https://github.com/alphagov/notifications-aws/pull/2284

We can now make the change to put inbound SMS tasks on the service-callbacks queue, with retries on the service-callbacks-retry queue for consistency with other callback tasks.


whpearson (approved):
Change looks good

PR #4124, opened 2024-07-02
remove high volume queues

leohemsted:
we have no high volume services defined in our creds at all so we can simply remove the entire codebase without worrying about deprecation orders of task definitions

Here are some reasons I think we should remove it:

1. There’s two queues (save-api-email-tasks and save-api-sms-tasks and a worker (api-worker-save-api-notifications)  that we pay money for
1. Code complexity - this was raised when @BlessedDev tried to update when we save notifications to handle unsubscribe links (which we will need to make sure work for [GOV.UK](http://gov.uk/) Email!), and it’s more nuanced because this json blob is passed on a celery task.
1. We have other services we consider as large such as NHS login, NHS Notify, DWP Prod, that we do not have on this queue and we have been fine
1. I don’t think we’ve ever tested that this actually increased performance of Notify. Our current DB connection usage is pretty low and even at peak times we’re no-where near our hard limit of 5000 concurrent connections.
1. We return a 200 OK, and give a user a notification ID, but don’t actually create the notification in the database. If something goes wrong (eg the task is lost off the queue/worker) then we won’t have created the notification but will have told the user that we did! that’s really bad.
1. Even if nothing goes wrong, in that period while the task is on the queue, if the service does a GET notification by id request, they’ll receive a 404 not found because the notification won’t be in the DB. This is super weird and confusing.

Deploy process:

- [x] remove services from creds so path isnt used https://github.com/alphagov/notifications-credentials/pull/461
- [ ] remove code path <-- YOU ARE HERE
- [ ] terraform delete ecs task, ssm creds. delete from concourse pipeline at the same time? https://github.com/alphagov/notifications-aws/pull/2276
- [ ] delete the cred files completely https://github.com/alphagov/notifications-credentials/pull/463
- [ ] clickops remove sqs queues
- [ ] remove worker definition from API https://github.com/alphagov/notifications-api/pull/4115
- [ ] figure out what to do with high-volume vs not-high-volume statsd metrics

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -7,7 +7,7 @@ from notifications_utils.insensitive_dict import InsensitiveDict
 from notifications_utils.recipient_validation.postal_address import PostalAddress
 from notifications_utils.recipients import RecipientCSV
 from notifications_utils.timezones import convert_utc_to_bst
-from sqlalchemy.exc import IntegrityError, SQLAlchemyError
+from sqlalchemy.exc import SQLAlchemyError
 
 from app import create_random_identifier, create_uuid, notify_celery, signing
 from app.aws import s3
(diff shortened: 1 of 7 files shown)


whpearson:
Looks good. Might need to modify it to kick off concourse again.

klssmith:
Really minor thing, but you can also remove the queue names from config.py and the `api-worker-save-api-notifications` app from entrypoint.sh

BlessedDev (approved):
Looks good to me. Nice One, there was a lot to untangle.

PR #4125, opened 2024-07-05
Add flask task to create the functional test fixtures

corlettb:
[Trello Card](https://trello.com/c/GA0r5iXd/766-get-functional-tests-to-work-in-dev-env-bootstrap-an-empty-database)

What
----

Upgrade moto to 5.0.11.

Add flask task to create the functional test fixtures

It will output a file (and/or write to ssm) at the end of the task with all the variables required to run the functional tests.

Why
----

This will allow us the automate the dev environments setup so they are ready for the functional tests.

It will also work for local environment setup to run the functional tests.

### DIFF
--- a/app/functional_tests_fixtures/__init__.py
+++ b/app/functional_tests_fixtures/__init__.py
@@ -1,0 +1,476 @@
+import os
+from uuid import uuid4
+
+import boto3
+from flask import current_app
+from sqlalchemy.exc import NoResultFound
+
+from app.constants import (
+    EDIT_FOLDER_PERMISSIONS,
+    EMAIL_AUTH,
+    EXTRA_LETTER_FORMATTING,
+    INBOUND_SMS_TYPE,
+    LETTER_TYPE,
+    SECOND_CLASS,
+    SEND_EMAILS,
+    SEND_LETTERS,
+    SEND_TEXTS,
+    VIEW_ACTIVITY,
+)
+from app.dao.annual_billing_dao import (
+    set_default_free_allowance_for_service,
+)
+from app.dao.api_key_dao import (
+    get_model_api_keys,
+    save_model_api_key,
+)
+from app.dao.inbound_numbers_dao import (
+    dao_get_inbound_number_for_service,
+    dao_set_inbound_number_to_service,
+)
+from app.dao.organisation_dao import (
+    dao_create_organisation,
+    dao_get_organisation_by_id,
+    dao_get_organisations_by_partial_name,
+    dao_update_organisation,
+)
+from app.dao.permissions_dao import permission_dao
+from app.dao.service_email_reply_to_dao import (
+    add_reply_to_email_address_for_service,
+    dao_get_reply_to_by_service_id,
+)
+from app.dao.service_inbound_api_dao import get_service_inbound_api_for_service, save_service_inbound_api
+from app.dao.service_permissions_dao import (
+    dao_add_service_permission,
+    dao_fetch_service_permissions,
+)
+from app.dao.service_sms_sender_dao import (
+    dao_add_sms_sender_for_service,
+    dao_get_sms_senders_by_service_id,
+)
+from app.dao.services_dao import (
+    dao_add_user_to_service,
+    dao_create_service,
+    dao_fetch_service_by_id,
+    get_services_by_partial_name,
+)
+from app.dao.templates_dao import dao_create_template, dao_get_all_templates_for_service
+from app.dao.users_dao import get_user_by_email, save_model_user
+from app.models import (
+    InboundNumber,
+    Organisation,
+    Permission,
+    Service,
+    ServiceEmailReplyTo,
+    ServiceInboundApi,
+    User,
+)
+from app.schemas import api_key_schema, template_schema
+
+"""
+This function creates a set of database fixtures that functional tests use to run against.
+
+It is intended for use in non-production environments only.
+
+The function will create all the database rows required for the functional tests and output
+an environment file that the functional tests can use to execute against the environment.
+
+The environment file can be outputted to a file or uploaded to AWS SSM. The file is intended
+for local testing, and SSM is for the pipeline.
+"""
+
+
+def apply_fixtures():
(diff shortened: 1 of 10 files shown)


app/functional_tests_fixtures/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+            Value=functional_test_config,
+            Type="SecureString",
+            Overwrite=True,
+        )
+
+        if response["ResponseMetadata"]["HTTPStatusCode"] != 200:
+            raise Exception("Failed to upload to SSM")
+
+
+def _create_user(name, email_address, password, auth_type="sms_auth", organisations=None, mobile_number=None):
```
leohemsted, line 195: all the functions in this file are really similar to `tests/app/db.py`, which is reliably kept up to date with model changes since it's tested via unit tests.

can we reuse those functions? (it's probs ok if we need to move files around to avoid circular imports)

app/functional_tests_fixtures/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+        "name": "Functional Tests - CSV SMS Template with Build ID",
+        "template_type": "sms",
+        "content": "The quick brown fox jumped over the lazy dog. Build id: ((build_id)).",
+        "created_by": user_id,
+    }
+
+    new_template = template_schema.load(data)
+
+    if not new_template.postage and new_template.template_type == LETTER_TYPE:
+        new_template.postage = SECOND_CLASS
```
leohemsted, line 366: we shouldn't need this guard in `_create_sms_template` right? is this a copy+paste error?

app/functional_tests_fixtures/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+    ServiceInboundApi,
+    User,
+)
+from app.schemas import api_key_schema, template_schema
+
+functional_tests_fixtures_blueprint = Blueprint("functional_test_fixtures", __name__)
+
+
+@functional_tests_fixtures_blueprint.cli.command("apply")
+def apply():
```
leohemsted, line 74: i think we could do with a docstring here with details on:

* how this function relates to the functional tests
* when this function gets invoked, and how it's used
* how and when we need to keep this up to date, with regards to all of:
  * functional test changes
  * env var changes
  * db schema changes

and anyhting else worth mentioning

leohemsted (commented):
i guess i'm a little unclear on what exactly's going on and i do not know what i would be expected to do to keep this up to date, nor how i would find out if it is not up to date (which is largely the same problem as our existing sql scripts)

app/functional_tests_fixtures/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+
+    _create_service_sms_senders(service.id, "07700900500", True, inbound_number_id)
+    _create_service_sms_senders(service.id, "func tests", False, None)
+
+    _create_service_inbound_api(service.id, service_admin_user.id)
+
+    functional_test_config = f"""
+
+export FUNCTIONAL_TESTS_API_HOST={current_app.config['API_HOST_NAME']}
+export FUNCTIONAL_TESTS_ADMIN_HOST={current_app.config['ADMIN_BASE_URL']}
```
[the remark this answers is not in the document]
corlettb: The flow in terms of the concourse pipelines will be the following:

- Terraform will execute this flask task via ecs optionally based on environment. It will run them for the initial install and after each api version update.
- The flask task will write the functional tests config file to SSM. This will change each time as we have generated passwords in there and will be different between environments as guids will change.
- Before the functional tests run they will download the environment file from ssm and source it. The output matches our current functional tests environment file.

Currently the file or ssm are optional based on environment variables passed in. The pipeline will just use ssm. The unit tests in this PR use a file output. I am currently testing it by running the flask command in a different interpreter so can't fake an ssm call. I could try to change that.

I included all the secrets here to just simplify the configuration needed for the functional tests. We could only include environment variables that were generated by the fixtures if you felt that was a better direction to go in.

This [is the pr for the terraform side](https://github.com/alphagov/notifications-aws/pull/2292/files)

app/functional_tests_fixtures/__init__.py
corlettb: As discussed these functions are idempotent vs the db.py ones. We also need to move the db.py out of tests and into the app. We decided it wasn't worth it in the end.

app/functional_tests_fixtures/__init__.py
corlettb: Yeah copy+paste issue. Fixed.

app/functional_tests_fixtures/__init__.py
corlettb: I've added a docstring.

app/functional_tests_fixtures/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+    FROM_NUMBER
+
+and the following environment variables:
+    REQUEST_BIN_API_TOKEN - request bin token to be used by functional tests
+    FUNCTIONAL_TEST_ENV_FILE - (optional) output file for the environment variables
+    SSM_UPLOAD_PATH - (optional) path to upload the environment variables to AWS SSM
+"""
+
+
+@functional_tests_fixtures_blueprint.cli.command("apply")
```
leohemsted, line 105: one thing i've realised - we use the library `click` in `app/commands.py` to manage our commands - i think it make sense to move this decorated wrapper function there (we then dont need the blueprint at all so can remove the changes to `test_route_authentication`. (then we could invoke it via `flask command apply-functional-test-fixtures` or similar.

click will automatically pull in the docstring from a function to be used as the cli `--help` flag too so we could move the docstring to there - however im less sure about that as i think while running the command most people will probs have this file open in a code editor so the docstring being alongside the code seems more helpful to me intuitively

(i think the actual `apply_fixtures` func and all the create funcs should still live in this file)

corlettb: Good spot. Updated.

leohemsted (approved):

leohemsted (approved):

PR #4127, opened 2024-07-08
Delete cache when processing new unsubscribe request

quis:
We want to cache a summary of unsubscribe requests for each service to make the dashboard load quickly.

We need to invalidate this cache whenever:
1. a new requst comes in
2. some requests are marked as completed
3. some old requests are cleaned up according to our data retention policy

This commit does 1., as a start


BlessedDev (commented):
Looks good to me.

BlessedDev (approved):

PR #4128, opened 2024-07-09
Add unsubscribe request reports summary endpoints

BlessedDev:
This PR builds the endpoints needed for the work done in this [PR](https://github.com/alphagov/notifications-admin/pull/5139).

### DIFF
--- a/app/one_click_unsubscribe/rest.py
+++ b/app/one_click_unsubscribe/rest.py
@@ -1,5 +1,5 @@
 import uuid
-from datetime import datetime, timedelta
+from datetime import datetime
 
 from flask import Blueprint, current_app, jsonify
 from itsdangerous import BadData
@@ -63,30 +63,33 @@ def get_unsubscribe_request_data(notification, email_address):
 
 
 def create_unsubscribe_request_reports_summary(service_id):
-    reports_summary = {"batched_reports_summaries": [], "unbatched_report_summary": {}}
-    existing_unsubscribe_reports_summaries = []
+    reports_summary = []
+    batched_unsubscribe_reports_summary = []
+    unbatched_unsubscribe_report_summary = []
     # Check for existing unsubscribe reports and create their summaries
     if unsubscribe_request_reports := get_unsubscribe_request_reports_dao(service_id):
-        existing_unsubscribe_reports_summaries = _create_existing_unsubscribe_request_reports_summary(
+        batched_unsubscribe_reports_summary = _create_batched_unsubscribe_request_reports_summary(
             unsubscribe_request_reports,
         )
-        reports_summary["batched_reports_summaries"] = existing_unsubscribe_reports_summaries
+        batched_unsubscribe_reports_summary = batched_unsubscribe_reports_summary
     # Check for unsubscribe requests and create a summary for them
     if unbatched_unsubscribe_requests := get_unbatched_unsubscribe_requests_dao(service_id):
-        if existing_unsubscribe_reports_summaries:
-            earliest_timestamp = existing_unsubscribe_reports_summaries[0]["latest_timestamp"] + timedelta(seconds=3)
+        if batched_unsubscribe_reports_summary:
+            earliest_timestamp = batched_unsubscribe_reports_summary[0]["latest_timestamp"]
         else:
             service = dao_fetch_service_by_id(service_id)
             earliest_timestamp = service.created_at
 
-        unbatched_unsubscribe_requests_report_summary = _create_unbatched_unsubscribe_request_report_summary(
-            unbatched_unsubscribe_requests, earliest_timestamp
+        unbatched_unsubscribe_report_summary.append(
+            _create_unbatched_unsubscribe_request_report_summary(unbatched_unsubscribe_requests, earliest_timestamp)
         )
-        reports_summary["unbatched_report_summary"] = unbatched_unsubscribe_requests_report_summary
+
+    reports_summary = unbatched_unsubscribe_report_summary + batched_unsubscribe_reports_summary
+
     return reports_summary
 
 
-def _create_existing_unsubscribe_request_reports_summary(unsubscribe_request_reports: list) -> list:
+def _create_batched_unsubscribe_request_reports_summary(unsubscribe_request_reports: list) -> list:
     report_summaries = []
     for report in unsubscribe_request_reports:
         report_summary = {
--- a/app/one_click_unsubscribe/rest.py
+++ b/app/one_click_unsubscribe/rest.py
@@ -99,6 +99,7 @@ def _create_batched_unsubscribe_request_reports_summary(unsubscribe_request_repo
             "latest_timestamp": report.latest_timestamp,
             "processed_by_service_at": report.processed_by_service_at,
             "is_a_batched_report": True,
+            "status": "Completed" if report.processed_by_service_at else "Downloaded",
         }
         report_summaries.append(report_summary)
     return report_summaries
@@ -115,5 +116,6 @@ def _create_unbatched_unsubscribe_request_report_summary(
         "latest_timestamp": datetime.utcnow(),
         "processed_by_service_at": None,
         "is_a_batched_report": False,
+        "status": "Not downloaded",
     }
     return report_summary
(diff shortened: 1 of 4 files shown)


app/one_click_unsubscribe/rest.py
```diff
@@ -52,3 +60,57 @@ def get_unsubscribe_request_data(notification, email_address):
         "service_id": notification.service_id,
         "email_address": email_address,
     }
+
+
+def create_unsubscribe_request_reports_summary(service_id):
+    reports_summary = {"batched_reports_summaries": [], "unbatched_report_summary": {}}
```
quis, line 66: I’m now wondering why we are returning 2 fields here when we just join them into one list in the admin app. Would it not be better to return one list with the contents of `unbatched_report_summary` as the first item?

We can always use the `is_a_batched_report` field to tell them apart

app/one_click_unsubscribe/rest.py
```diff
@@ -52,3 +60,57 @@ def get_unsubscribe_request_data(notification, email_address):
...
+    # Check for existing unsubscribe reports and create their summaries
+    if unsubscribe_request_reports := get_unsubscribe_request_reports_dao(service_id):
+        existing_unsubscribe_reports_summaries = _create_existing_unsubscribe_request_reports_summary(
+            unsubscribe_request_reports,
+        )
+        reports_summary["batched_reports_summaries"] = existing_unsubscribe_reports_summaries
+    # Check for unsubscribe requests and create a summary for them
+    if unbatched_unsubscribe_requests := get_unbatched_unsubscribe_requests_dao(service_id):
+        if existing_unsubscribe_reports_summaries:
+            earliest_timestamp = existing_unsubscribe_reports_summaries[0]["latest_timestamp"] + timedelta(seconds=3)
```
quis, line 77: Why are we adding 3 seconds here? Feels like we might miss some requests in the gap

app/one_click_unsubscribe/rest.py
quis: This would also let you move a lot of the Python logic below into a single SQL query. Not sure of the SQLAlchemy syntax but something like this would be faster and easier to follow:
```sql
select
    batch_id, count(*)
from
    unsubscribe_request
group by
    batch_id
```
Would have to figure out the `join` where `batch_id` is none but I’m sure someone will know the SQL wizardry.

BlessedDev: I had started off merging both batched and un-batched reports and later unpicked them to give flexibility but it is seems to be more trouble than it's worth, so I have merged them again.

Regarding the remaining logic. One slight area of complication has been that we have batched `unsubscribe_request_reports` that exist in the database and we have to present un-batched `unsubscribe_requests` as an unbatched report whose `count` and `latest_timestamp` are dynamic and potentially change every time the user views them until they are converted to a batched `unsubscribe_request_report `once the user initiates a download.
The premise I worked on was creating summaries for all reports, batched or unbatched, which are presented to the user. The summaries are not persisted in the database and are crafted by this method. Once a user initiates a download, an un-batched report summary is converted to an `unsubscribe_request_report`, (the` batch_id` it was assigned will become the batched `unsubscribe_request_report` `id` and stored in the data base.

app/one_click_unsubscribe/rest.py
BlessedDev: That was a mistake, It was my intention to use 3 microseconds. I had wanted some kind of demarcation between an `unsubscribe_request_report` `earliest_timestamp` and the previous `unsubscribe_request_report` `latest_timestamp`.
I have removed it.

app/one_click_unsubscribe/rest.py
BlessedDev: DId you mean `get_unbatched_unsubscribe_requests_dao` should return the `count` as well?

BlessedDev: I now see your point, were you referring to something like this? https://github.com/alphagov/notifications-api/blob/1c591702c3c59771bdb1d53e5902f8295f56ee33/app/dao/returned_letters_dao.py#L78

BlessedDev:
I am rethinking some of the work done here and converting it back to a draft.

app/one_click_unsubscribe/rest.py
```diff
@@ -52,3 +60,62 @@ def get_unsubscribe_request_data(notification, email_address):
...
+            unsubscribe_request_reports,
+        )
+        batched_unsubscribe_reports_summary = batched_unsubscribe_reports_summary
+    # Check for unsubscribe requests and create a summary for them
+    if unbatched_unsubscribe_requests := get_unbatched_unsubscribe_requests_dao(service_id):
+        if batched_unsubscribe_reports_summary:
+            earliest_timestamp = batched_unsubscribe_reports_summary[0]["latest_timestamp"]
+        else:
+            service = dao_fetch_service_by_id(service_id)
+            earliest_timestamp = service.created_at
```
quis, line 81: I think it would be better if this was the timestamp of the oldest unbatched request. Otherwise it’s going to look confusing when there are dates from several years ago being displayed, which don’t then appear in the report

app/one_click_unsubscribe/rest.py
```diff
@@ -52,3 +60,62 @@ def get_unsubscribe_request_data(notification, email_address):
...
+    report_summaries = []
+    for report in unsubscribe_request_reports:
+        report_summary = {
+            "batch_id": str(report.id),
+            "count": report.count,
+            "earliest_timestamp": report.earliest_timestamp,
+            "latest_timestamp": report.latest_timestamp,
+            "processed_by_service_at": report.processed_by_service_at,
+            "is_a_batched_report": True,
+            "status": "Completed" if report.processed_by_service_at else "Downloaded",
```
quis, line 102: I think we should keep this logic in the admin app because it’s presentational, not an actual field in the database
```suggestion
```

app/one_click_unsubscribe/rest.py
```diff
@@ -52,3 +60,62 @@ def get_unsubscribe_request_data(notification, email_address):
...
+) -> dict:
+    count = len(unbatched_unsubscribe_requests)
+    report_summary = {
+        "batch_id": str(uuid.uuid4()),
+        "count": count,
+        "earliest_timestamp": earliest_timestamp,
+        "latest_timestamp": datetime.utcnow(),
+        "processed_by_service_at": None,
+        "is_a_batched_report": False,
+        "status": "Not downloaded",
```
quis, line 119: Same here
```suggestion
```

app/one_click_unsubscribe/rest.py
```diff
@@ -52,3 +60,62 @@ def get_unsubscribe_request_data(notification, email_address):
...
+        report_summaries.append(report_summary)
+    return report_summaries
+
+
+def _create_unbatched_unsubscribe_request_report_summary(
+    unbatched_unsubscribe_requests: list, earliest_timestamp: str = None
+) -> dict:
+    count = len(unbatched_unsubscribe_requests)
+    report_summary = {
+        "batch_id": str(uuid.uuid4()),
```
quis, line 113: I’m a bit unclear how this will work, because the API will return a different ID on each request. That means the admin app won’t have a fixed way to ask for the unbatched requests. I think the admin app can work that out by looking at `is_a_batched_report` instead, but it’s confusing to return a UUID here instead of `None`:
```suggestion
        "batch_id": None,
```

app/one_click_unsubscribe/rest.py
BlessedDev: I understand your point. My premise was that since the summary already contained a non  database field `is_a_batched_report, however I am happy to move it to the admin app.

app/one_click_unsubscribe/rest.py
BlessedDev: Good catch. You are right, it should be None.

app/one_click_unsubscribe/rest.py
quis: Let’s say we wanted to change the wording in the future. To me it feels more natural to make that change in the admin app. So that’s why ’d suggest it lives there

app/one_click_unsubscribe/rest.py
quis: Yeah, timestamp of the earliest unbatched request. I think the `created_at` date of the earliest request in the report makes more sense.

quis (approved):

PR #4129, opened 2024-07-10
Fix failing test - pass in date dynamically into value in test

CrystalPea:
Instead of hard-coding it.


quis (approved):
Approving this because it matches the pattern on lines 316–318, but I would prefer that we use plain strings in test assertions in combination with an `@freeze_time` decorator on the function

CrystalPea:
@quis thanks for approving - I could do that later, as a refactor?

PR #4132, opened 2024-07-10
When serializing cost info for notification, rename rate to sms_rate

CrystalPea:
This is following missed piece of content review. We are renaming it so it's more specific, and so users do not get confused.


joybytes (approved):

joybytes (approved):
lgtm!

PR #4135, opened 2024-07-15
Dockerfile: pre-compile bytecode for app in production image

risicle:
This should produce a small improvement in startup time, which is useful when scaling up in response to sudden loads.

Bytecode for the libraries in the venv is already generated during pip install.


robinjam (approved):
Nice

PR #4136, opened 2024-07-16
use new phone number validation code

rparke:
Update to API to use utils 80.1.1, makes changes to v2 schema validation and tests to support breaking changes to how exception objects are instantiated, and makes changes to tests using invalid (but previously uncaught) numbers.

### DIFF
--- a/app/v2/errors.py
+++ b/app/v2/errors.py
@@ -2,7 +2,7 @@ import json
 
 from flask import current_app, jsonify, request
 from jsonschema import ValidationError as JsonSchemaValidationError
-from notifications_utils.recipient_validation.errors import InvalidRecipientError
+from notifications_utils.recipient_validation.errors import InvalidPhoneError, InvalidRecipientError
 from sqlalchemy.exc import DataError
 from sqlalchemy.orm.exc import NoResultFound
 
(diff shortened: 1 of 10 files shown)


app/v2/errors.py
```diff
@@ -85,14 +85,15 @@ def to_dict_v2(self):
 def register_errors(blueprint):
     @blueprint.errorhandler(InvalidRecipientError)
```
leohemsted, line 86: this also catches `InvalidEmailError` and `InvalidAddressError` - they won't have the `get_legacy_v2_api_error_message` function on them 🤔 

does this work correctly for invalid emails? or does this now crash

i think we'll want a separate error handler for `InvalidPhoneError` (and then decorate this function twice - once for InvalidLetter and one for InvalidEmail?)

rparke: No tests have failed ... (assuming we have tests for invalid emails)

leohemsted:
```python
    @blueprint.errorhandler(InvalidPhoneError)
    def invalid_phone_format(error):
        return .... (get_legacy_v2_api_error_message)

    @blueprint.errorhandler(InvalidRecipientError)
    def invalid_format(error):
        return .... (str(error))
    ```

rparke: Added the InvalidPhoneError handling.

Do we want to think about tests for emails now or later?

leohemsted: lets worry later

PR #4139, opened 2024-07-22
Count all unsubscribe requests from last 7 days on dashboard

quis:
The dashboard says:
> ## In the last 7 days
>
> **1,234** email unsubscribe requests
> Latest report 3 weeks ago

At the moment the 1,234 number will go down when requests are marked as completed. This doesn’t feel right, because those requests have still been received in the last 7 days.

This pull request changes the query to count all requests from the last 7 days, whether or not they have been marked as completed.

This is consistent with received text messages and returned letters (although we don’t have a similar concept of marking these are completed at the moment).

### DIFF
--- a/tests/app/dao/test_unsubscribe_request_dao.py
+++ b/tests/app/dao/test_unsubscribe_request_dao.py
@@ -91,6 +91,19 @@ def test_get_unsubscribe_requests_statistics_dao(sample_service):
         }
     )
 
+    notification_5 = create_notification(template=template_2)
+    # This request should not be counted because it’s more than 7 days ago
+    create_unsubscribe_request_dao(
+        {
+            "notification_id": notification_5.id,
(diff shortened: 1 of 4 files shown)


tests/app/dao/test_unsubscribe_request_dao.py
```diff
@@ -91,6 +91,36 @@ def test_get_unsubscribe_requests_statistics_dao(sample_service):
         }
     )
 
+    notification_5 = create_notification(template=template_2)
```
BlessedDev, line 94: The doc string is no longer accurate and should be removed

BlessedDev (commented):
Looks good to me. I have suggested removing the `docstring` (or altering it) for 
`test_get_unsubscribe_requests_statistics_dao` as it is no longer an accurate description.

BlessedDev (approved):

PR #4145, opened 2024-07-25
Fix letter callback endpoint

klssmith:
This is only temporary code, since we have a separate story to handle callbacks properly. However, this change stops the endpoint from giving an error while testing.

### DIFF
--- a/app/notifications/notifications_letter_callback.py
+++ b/app/notifications/notifications_letter_callback.py
@@ -71,5 +71,7 @@ def process_letter_callback():
         notification_id = signing.decode(token)
     except BadSignature:
         current_app.logger.info("Letter callback with invalid token of %s received", token)
+    else:
+        current_app.logger.info("Letter callback for notification id %s received", notification_id)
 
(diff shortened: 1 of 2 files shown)


app/notifications/notifications_letter_callback.py
```diff
@@ -71,5 +71,7 @@ def process_letter_callback():
         notification_id = signing.decode(token)
     except BadSignature:
         current_app.logger.info("Letter callback with invalid token of %s received", token)
+    else:
+        current_app.logger.info("Letter callback for notification id %s received", notification_id)
 
-    current_app.logger.info("Letter callback for notification id %s received", notification_id)
+    return "ok"
```
leohemsted, line 77: might be nice to say `return jsonify(result="success"), 200` in line with our SMS callbacks

klssmith: Sure, I was assuming that everything in this function would be overwritten when the code for it is written "properly" but have changed it

leohemsted (approved):

PR #4146, opened 2024-07-26
Add process unsubscribe request report endpoint

BlessedDev:
This PR adds a `process_unsubscribe_request_report` endpoint which supports the work done in [this](https://github.com/alphagov/notifications-admin/pull/5152) PR to facilitate a user being able to mark an `unsubscribe_request_report` as completed by updating the `processed_by_service_at` value of an `unsubscribe_request_report` as described in [this](https://trello.com/c/5hJS5Oia/54-let-users-mark-email-unsubscribe-requests-as-completed) card.

### DIFF
--- a/app/dao/unsubscribe_request_dao.py
+++ b/app/dao/unsubscribe_request_dao.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from sqlalchemy import desc, func
 
 from app import db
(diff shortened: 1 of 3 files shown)


quis:
If we made this into a more generic `update` endpoint would that make it easier for the user to undo marking the report as processed?

BlessedDev:
> If we made this into a more generic `update` endpoint would that make it easier for the user to undo marking the report as processed?

Presently it can update the `processed_by_service_at` value to a new one but not a value of `None`. I can alter the form validation on the Front End to accept a cleared checkbox as a valid submission so that the such that the `processed_by_service_at` value can be set back to `None`.

quis:
Yes, that would also work

BlessedDev:
> Yes, that would also work

I will update the `process_unsubscribe_request_report` endpoint accordingly.

quis (approved):

PR #4147, opened 2024-07-26
Add unsubscribe link to body of emails

quis:
Best reviewed commit-by-commit.

# Generate unsubscribe link at time of sending

When  a user has provided us an unsubscribe link through an API call we need to store it. But where we are generating the link we don’t need to store it, because we already store all the information needed to generate it.

By not storing it we make it easy to differentiate which notifications have user-supplied unsubscribe links and which have ones which we are generating.

# Add unsubscribe link to body of emails

We have started sending list-unsubscribe headers if a user marks a template as unsubscribeable, for teams who aren’t using the API.

However we are already telling these teams to add unsubscribe links to the body of their emails.

This means:

- there are two different ways to add unsubscribe links
- users will have to manage requests to unsubscribe coming from two different places

This will be confusing, and make it harder for us to write good guidance that users will be able to follow, and harder to explain what the ‘let users unsubscribe from these emails’ checkbox does.

If we have a confusing product proposition and users cant follow our guidance then we risk our spam score rising.

This commit adds a way to generate a separate unsubscribe link which we can add to the footer of every email (using the argument added in https://github.com/alphagov/notifications-utils/pull/1136)

We can then report unsubscribes back to the team in the same way, no matter if the recipient has clicked the button in their email client or the link at the end of the email.

***

If the user has checked the checkbox then we will always show the link in the body of the email (this feels like the most predictable behaviour). If they have also passed in a one-click URL for the headers then we will respect that too.

The only combination which isn’t possible is using Notify’s generated URL for one-click, and your own in the body of the email. This feels like the most unlikely combination.

***

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/1136

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1751,15 +1751,22 @@ class Notification(db.Model):
 
         return letter_rate
 
+    def _generate_unsubscribe_link(self, base_url):
+        return url_with_token(
+            self.to,
+            url=f"/unsubscribe/{str(self.id)}/",
+            base_url=base_url,
+        )
+
     def get_unsubscribe_link_for_headers(self, *, template_has_unsubscribe_link):
         if self.unsubscribe_link:
             return self.unsubscribe_link
         if template_has_unsubscribe_link:
-            return url_with_token(
-                self.to,
-                url=f"/unsubscribe/{str(self.id)}/",
-                base_url=current_app.config["API_HOST_NAME"],
-            )
+            return self._generate_unsubscribe_link(current_app.config["API_HOST_NAME"])
+
+    def get_unsubscribe_link_for_body(self, *, template_has_unsubscribe_link):
+        if template_has_unsubscribe_link:
+            return self._generate_unsubscribe_link(current_app.config["ADMIN_BASE_URL"])
 
 
 class NotificationHistory(db.Model):
--- a/app/models.py
+++ b/app/models.py
@@ -1759,12 +1759,19 @@ class Notification(db.Model):
         )
 
     def get_unsubscribe_link_for_headers(self, *, template_has_unsubscribe_link):
+        """
+        Generates a URL on the API domain, which accepts a POST request from an email client
+        """
         if self.unsubscribe_link:
             return self.unsubscribe_link
         if template_has_unsubscribe_link:
             return self._generate_unsubscribe_link(current_app.config["API_HOST_NAME"])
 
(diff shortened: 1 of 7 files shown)


app/models.py
```diff
@@ -1748,6 +1751,23 @@ def _get_letter_cost(self):
 
         return letter_rate
 
+    def _generate_unsubscribe_link(self, base_url):
```
CrystalPea, line 1754: It would be useful to add a docstring/comment here explaining why we have admin base URL for body, and api base URL for header maybe?

I gather it's because in header we have one-click POST link, and in body we need a GET link so the email previews do not auto-unsubscribe the user?

CrystalPea (commented):
Hi,

Code looks good, two little comments.

I was unable to make my local apps run unfortunately, so I couldn't test if the code does what it should. Shall we go on a meet tomorrow and you show me how it works on your machine?

tests/app/notifications/test_process_notification.py
```diff
@@ -640,7 +635,7 @@ def test_persist_notification_when_template_has_unsubscribe_link_is_false(
     "unsubscribe_link, expected_unsubscribe_link",
     [
         ("https://please-unsubscribe-me.com/unsubscribe", "https://please-unsubscribe-me.com/unsubscribe"),
-        (None, "https://notify-generated-super-unsubscribe-link.com/unsubscribe"),
+        (None, None),
```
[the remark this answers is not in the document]
quis:
```suggestion
        # We don’t persist template-level unsubscribe links – they are generated at time of sending
        (None, None),
```

app/models.py
```diff
@@ -1748,6 +1751,23 @@ def _get_letter_cost(self):
...
+            base_url=base_url,
+        )
+
+    def get_unsubscribe_link_for_headers(self, *, template_has_unsubscribe_link):
+        if self.unsubscribe_link:
+            return self.unsubscribe_link
+        if template_has_unsubscribe_link:
+            return self._generate_unsubscribe_link(current_app.config["API_HOST_NAME"])
+
+    def get_unsubscribe_link_for_body(self, *, template_has_unsubscribe_link):
```
quis, line 1767:
```suggestion
    def get_unsubscribe_link_for_body(self, *, template_has_unsubscribe_link):
        """
        Generates a URL on the admin domain, which serves a page telling the user they
        have been unsubscribed
        """
```

CrystalPea (approved):
:shipit:

PR #4152, opened 2024-07-30
Enable download of unsubscribe request reports

BlessedDev:
This PR adds the endpoints to support downloading an `unsubscribe-request-report` as described in [this](https://trello.com/c/BeZGTW0E/53-let-users-download-email-unsubscribe-requests) card.

### DIFF
--- a/app/dao/unsubscribe_request_dao.py
+++ b/app/dao/unsubscribe_request_dao.py
@@ -1,11 +1,10 @@
 from datetime import datetime
 
 from sqlalchemy import desc, func
-from sqlalchemy.sql.operators import is_
 
 from app import db
 from app.dao.dao_utils import autocommit
-from app.models import UnsubscribeRequest, UnsubscribeRequestReport
+from app.models import UnsubscribeRequest, UnsubscribeRequestReport, Notification, NotificationHistory, Template, Job
 from app.utils import midnight_n_days_ago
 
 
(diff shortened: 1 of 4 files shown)


app/dao/unsubscribe_request_dao.py
```diff
@@ -80,3 +105,20 @@ def create_unsubscribe_request_reports_dao(unsubscribe_request_report):
...
+    """
+    unbatched_unsubscribe_requests = UnsubscribeRequest.query.filter(
+        UnsubscribeRequest.unsubscribe_request_report_id.is_(None),
+        UnsubscribeRequest.service_id == service_id,
+        UnsubscribeRequest.created_at >= earliest_timestamp,
+        UnsubscribeRequest.created_at <= latest_timestamp,
+    ).all()
+    for unsubscribe_request in unbatched_unsubscribe_requests:
+        unsubscribe_request.unsubscribe_request_report_id = report_id
+        db.session.add(unsubscribe_request)
```
quis, line 124: I’m wondering if it’s better to batch up all the unbatched requests, rather than leave any that arrive after `latest_timestamp` dangling

Pro: you will get a download of everything you need to action
Con: the `count` might not quite match up to what you see on the page

BlessedDev: I wondered about that as well, the `latest_timestamp` is also the `earliest_timestamp` of the prospective next report so the un-batched reports that arrive soon after that will go into that next report.
One other con is that we may have an inconsistency where reports either have requests in them that are beyond their `latest_timestamps` or not have requests that should have been in them considering their `earliest_timestamps`.
My thinking on the issue is that this is one we can revisit once we get an idea of how users get on with the feature. It would be a simple fix whichever way we go.

quis: Yeah I’ve had a deeper think about this and I think the way you have it is right. Admin `POST`s `earliest_timestamp` and `latest_timestamp` and the API returns what’s between those two and nothing more.

app/dao/unsubscribe_request_dao.py
```diff
@@ -80,3 +105,20 @@ def create_unsubscribe_request_reports_dao(unsubscribe_request_report):
...
+    """
+    This method retrieves all the un-batched unsubscribe requests received before or on
+    the report's latest_timestamp and sets their unsubscribe_request_report_id to the report_i
+    """
+    unbatched_unsubscribe_requests = UnsubscribeRequest.query.filter(
+        UnsubscribeRequest.unsubscribe_request_report_id.is_(None),
+        UnsubscribeRequest.service_id == service_id,
+        UnsubscribeRequest.created_at >= earliest_timestamp,
+        UnsubscribeRequest.created_at <= latest_timestamp,
+    ).all()
```
quis, line 121: I don’t know if this is the exact SQLAlchemy syntax but a single update query should be much more efficient than a for loop
```suggestion
    ).update(
        report_id=report_id
    )
```

quis (approved):

PR #4153, opened 2024-08-01
Tweak start and end dates of unsubscribe request reports

quis:
Originally we thought it would be nice and clean to extend the end dates of the reports so there were no gaps between them.

On second thought I think this will be confusing where the latest report claims to run up to the current time. In this case:
- the report, when downloaded might not even contain requests from today so it could look like requests are missing from the report
- the report might claim to have requests up until now, but if any new ones come in between the user clicking on the page and generating the report then they wouldn’t be included

In could also be confusing when a report is cleaned up by data retention. Adjacent reports would then appear to expand to fill the gap, even though they gain no new requests.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2676,6 +2676,27 @@ class UnsubscribeRequestReport(db.Model):
     processed_by_service_at = db.Column(db.DateTime, nullable=True)
     count = db.Column(db.BigInteger, nullable=False)
 
+    def serialize(self):
+        return {
+            "batch_id": str(self.id),
+            "count": self.count,
+            "earliest_timestamp": self.earliest_timestamp.isoformat(),
+            "latest_timestamp": self.latest_timestamp.isoformat(),
+            "processed_by_service_at": self.processed_by_service_at.isoformat(),
+            "is_a_batched_report": True,
+        }
+
+    @staticmethod
+    def serialize_unbatched_requests(self, unbatched_unsubscribe_requests):
+        return {
+            "batch_id": None,
+            "count": len(unbatched_unsubscribe_requests),
+            "earliest_timestamp": unbatched_unsubscribe_requests[-1].created_at.isoformat(),
+            "latest_timestamp": unbatched_unsubscribe_requests[0].created_at.isoformat(),
+            "processed_by_service_at": None,
+            "is_a_batched_report": False,
+        }
+
 
 class UnsubscribeRequest(db.Model):
     __tablename__ = "unsubscribe_request"
--- a/app/models.py
+++ b/app/models.py
@@ -2682,7 +2682,9 @@ class UnsubscribeRequestReport(db.Model):
             "count": self.count,
             "earliest_timestamp": self.earliest_timestamp.isoformat(),
             "latest_timestamp": self.latest_timestamp.isoformat(),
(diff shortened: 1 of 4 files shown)


app/models.py
```diff
@@ -2676,6 +2676,27 @@ class UnsubscribeRequestReport(db.Model):
     processed_by_service_at = db.Column(db.DateTime, nullable=True)
     count = db.Column(db.BigInteger, nullable=False)
 
+    def serialize(self):
+        return {
+            "batch_id": str(self.id),
+            "count": self.count,
+            "earliest_timestamp": self.earliest_timestamp.isoformat(),
+            "latest_timestamp": self.latest_timestamp.isoformat(),
+            "processed_by_service_at": self.processed_by_service_at.isoformat(),
```
BlessedDev, line 2685: This would raise an error for an unprocessed `unsubscribe_request_report` which would have a `latest_timestamp` of `None`.
Maybe something along the lines of:
`"processed_by_service_at":
                self.processed_by_service_at.isoformat() if self.processed_by_service_at else None`.

It would also mean adding test coverage for an `unsubscribe_request_report` with no `processed_by_service_at` value.

app/models.py
```diff
@@ -2676,6 +2676,27 @@ class UnsubscribeRequestReport(db.Model):
...
+    def serialize_unbatched_requests(unbatched_unsubscribe_requests):
+        return {
+            "batch_id": None,
+            "count": len(unbatched_unsubscribe_requests),
+            "earliest_timestamp": unbatched_unsubscribe_requests[-1].created_at.isoformat(),
+            "latest_timestamp": unbatched_unsubscribe_requests[0].created_at.isoformat(),
+            "processed_by_service_at": None,
+            "is_a_batched_report": False,
+        }
+
```
BlessedDev, line 2700: `earliest_timestamp` and `latest_timestamp` of the report matching the `created_at` values of the earliest and latest  un-batched `unsubscribe_requests` is more logically consistent in my opinion and I think this is a good change 👍. 
It also makes things a lot simpler and saves a lot of headache down the road. 
Also shifting this into the model is nice as well.

tests/app/service/test_rest.py
```diff
@@ -3533,8 +3533,8 @@ def test_get_unsubscribe_request_report_summary_for_initial_unsubscribe_requests
     expected_unbatched_unsubscribe_request_summary = {
         "batch_id": None,
         "count": 2,
-        "earliest_timestamp": notification_2.created_at.strftime(date_format) + " GMT",
-        "latest_timestamp": (datetime.utcnow()).strftime(date_format) + " GMT",
+        "earliest_timestamp": notification_1.created_at.isoformat(),
+        "latest_timestamp": notification_2.created_at.isoformat(),
```
BlessedDev, line 3537: 👍

BlessedDev (changes requested):
I agree with the changes made here and the refactoring is great. I have flagged up the issue of `isoformat()` being potentially applied  to a `processed_by_service_at` value of `None` which needs to be addressed and maybe covered in a test case.

app/models.py
quis: Good spot - have added this in 2b8d9ea

BlessedDev (approved):

PR #4157, opened 2024-08-08
feat: implement dvla callback processing for letter notifications

joybytes:
## Summary
Introduces the `process_letter_callback` function to handle letter notification callbacks by verifying the token, matching notification IDs, and processing callback data. 

Main functionalities for this callback are::
- Token validation, we raise errors for invalid tokens or mismatched notification IDs.
- Updating notification statuses and handling duplicates to avoid duplicate updates.
- Checking billable units against the DVLA page count and logging discrepancies.

### Related tasks:
[Make DVLA callback API update NotificationLetterDespatch](https://trello.com/c/rpy5t1Dh/887-make-dvla-callback-api-update-notificationletterdespatch)


app/celery/process_letter_client_response_tasks.py
```diff
@@ -0,0 +1,56 @@
...
+
+    dao_update_notification(notification)
+
+    if new_status == NOTIFICATION_TECHNICAL_FAILURE:
+        raise NotificationTechnicalFailureException(
+            f"Letter status received as REJECTED for notification id: {notification.id}"
+        )
+
+
+def check_billable_units_by_id(notification, dvla_page_count):
```
ipdigital24, line 38: method name with suffix __by_id_ suggests that the method relies on _id_ whereas it actually relies on _notification_ entity

app/celery/process_letter_client_response_tasks.py
```diff
@@ -0,0 +1,56 @@
...
+            notification_id,
+            new_status,
+        )
+        return
+
+    notification.status = new_status
+
+    dao_update_notification(notification)
+
+    if new_status == NOTIFICATION_TECHNICAL_FAILURE:
```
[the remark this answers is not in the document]
joybytes: no, we do want to update the notification even on the event of "TECHINICAL_FAILURE"
It just means that from the DVLA side, the letter was for one reason or another rejected (example: no fixed abode) and we need to take action on our side. Logging the error is a way of letting us know that something went wrong, but we still want the database to reflect the true state of the letter at this given time.

app/celery/process_letter_client_response_tasks.py
joybytes: well spotted, thanks
https://github.com/alphagov/notifications-api/pull/4157/commits/e0eb522c33e3aa06c7a69e951fb4b25643a9116d

app/celery/process_letter_client_response_tasks.py
```diff
@@ -0,0 +1,56 @@
...
+from app.constants import DVLA_NOTIFICATION_DISPATCHED, NOTIFICATION_DELIVERED, NOTIFICATION_TECHNICAL_FAILURE
+from app.dao.notifications_dao import (
+    dao_get_notification_or_history_by_id,
+    dao_update_notification,
+)
+from app.exceptions import NotificationTechnicalFailureException
+
+
+@notify_celery.task(bind=True, name="process-letter-callback")
+def process_letter_callback_data(self, notification_id, page_count, dvla_status):
```
ipdigital24, line 13: is there a requirement to ensure dvla_status is one of DISPATCHED and REJECTED before processing?

app/celery/process_letter_client_response_tasks.py
```diff
@@ -0,0 +1,56 @@
...
+    if notification.billable_units != int(dvla_page_count):
+        current_app.logger.error(
+            "Notification with id %s has %s billable_units but DVLA says page count is %s",
+            notification.id,
+            notification.billable_units,
+            dvla_page_count,
+        )
+
+
+def determine_new_status(dvla_status):
```
ipdigital24, line 48: looks like it is a map function. is there a separate place (else where in the code) for such mapping?

app/constants.py
```diff
@@ -98,6 +98,10 @@
 NOTIFICATION_STATUS_LETTER_RECEIVED = "received"
 DVLA_RESPONSE_STATUS_SENT = "Sent"
 
+# DVLA letter callback status
```
ipdigital24, line 101: could these be wrapped like `NOTIFICATION_STATUS_TYPES = []`? ideally these should have been enums

app/celery/process_letter_client_response_tasks.py
joybytes: yes, we do this as part of the schema validation here: https://github.com/alphagov/notifications-api/blob/91ee6fa3410d912f1c91c05465bd3952d9244bc6/app/notifications/notifications_letter_callback.py#L88

app/dao/notifications_dao.py
```diff
@@ -234,6 +234,13 @@ def get_notification_by_id(notification_id, service_id=None, _raise=False):
     return query.one() if _raise else query.first()
 
 
+def dao_get_notification_or_history_by_id(notification_id):
```
ipdigital24, line 237: when I saw the method (with _dao_get_ prefix) I wasn't sure if it was returning a dao or an entity. it is now clearer that it returns a notification object and I wonder what is the distinction between _get__ and _dao_get__

app/celery/process_letter_client_response_tasks.py
joybytes: no, but I liked the suggestion so I did the following refactoring
https://github.com/alphagov/notifications-api/pull/4157/commits/cab242e899a314bdbeb289b9ffa84a7c90ad5234

app/constants.py
joybytes: addressed on the comment above: https://github.com/alphagov/notifications-api/pull/4157/commits/cab242e899a314bdbeb289b9ffa84a7c90ad5234

app/dao/notifications_dao.py
joybytes: this is a nomenclature that it seems to be existing within the code

so if we separate the layers of the application we would get something like
handler (endpoint entry where validation occurs)
service (business logic)
dao (database operations)

so because this function is part of a database operation, the prefix dao seems to be what is done in general.
I might be wrong, but that is what I have observed.

tests/app/notifications/test_notifications_letter_callbacks.py
```diff
@@ -102,3 +104,186 @@ def _sample_sns_s3_callback(filename):
...
+        # missing totalSheets field
+        (
+            [
+                {"key": "postageClass", "value": "1ST"},
+                {"key": "mailingProduct", "value": "MM UNSORTED"},
+                {"key": "Print Date", "value": "2024-08-01T09:15:14.456Z"},
+            ],
+            "data [{key: postageClass, value: 1ST}, {key: mailingProduct, value: MM UNSORTED}, {key: Print Date, "
+            "value: 2024-08-01T09:15:14.456Z}] is too short",
+        ),
```
ipdigital24, line 210: should validation of `mailingProduct` also be included?

app/celery/process_letter_client_response_tasks.py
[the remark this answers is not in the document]
ipdigital24: I now got the purpose however that raises further queries:
* what should celery task do when NotificationTechnicalFailureException is thrown? should it ignore, retry, etc.
* what if `dao_update_notification` faised when the status is TECHINICAL_FAILURE?
* could this be just logged to indicate the status? 
* could this be just returned as a flag to indicate the status?

app/notifications/notifications_letter_callback.py
```diff
@@ -64,14 +132,37 @@ def process_letter_response():
...
         notification_id = signing.decode(token)
     except BadSignature:
         current_app.logger.info("Letter callback with invalid token of %s received", token)
-    else:
-        current_app.logger.info("Letter callback for notification id %s received", notification_id)
+        raise InvalidRequest("A valid token must be provided in the query string", 403) from None
+
+    request_data = request.get_json()
+
+    if notification_id != request_data["id"]:
```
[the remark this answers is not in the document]
joybytes: I don't think that will bring much value

tests/app/notifications/test_notifications_letter_callbacks.py
joybytes: https://github.com/alphagov/notifications-api/pull/4157/commits/912f29565533198d925088bc026697a73eb2ee13

app/notifications/notifications_letter_callback.py
```diff
@@ -64,14 +132,37 @@ def process_letter_response():
...
+        current_app.logger.exception(
+            "Notification ID %s in letter callback data does not match token ID %s",
+            request_data["id"],
+            notification_id,
+        )
+        raise InvalidRequest("Notification ID in letter callback data does not match ID in token", 400)
+
+    current_app.logger.info("Letter callback for notification id %s received", notification_id)
```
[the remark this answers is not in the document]
joybytes: I quite like it after the validation

app/celery/process_letter_client_response_tasks.py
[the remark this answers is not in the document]
joybytes: - I believe that currently, we do updates to TECHNICAL_FAILURE, I don't believe this is something we're introducing specifically in this PR. I think this is the normal flow in place. @klssmith can you confirm?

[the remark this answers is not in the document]
klssmith: We are not expecting to receive "REJECTED" statuses from DVLA, so if we do we would need to get in touch and to find out what was wrong with the letters and what happened to them. Updating a notification to technical failure and also logging an error is a pattern we use elsewhere, such as [here](https://github.com/alphagov/notifications-api/blob/6f79e083cfe3cf6075c92013a23b8bd027a80dd5/app/celery/letters_pdf_tasks.py#L276-L277).

Updating the status (as well as logging) means that we can easily query for affected letters in the database and then update them. If we just log, we would need to parse the logs to get all the notification IDs out which isn't so simple.

app/celery/process_letter_client_response_tasks.py
```diff
@@ -0,0 +1,56 @@
...
+    return DVLA_TO_NOTIFICATION_STATUS_MAP[dvla_status]
+
+
+def is_duplicate_update(current_status, new_status):
```
klssmith, line 55: Do we want to also check for the case where we receive a second, different, callback status for a notification? It _shouldn't_ ever happen, but that would be something that we'd want to know about. E.g. when updating other notifications, we [do it this way](https://github.com/alphagov/notifications-api/blob/6f79e083cfe3cf6075c92013a23b8bd027a80dd5/app/dao/notifications_dao.py#L177-L185)

joybytes: https://github.com/alphagov/notifications-api/pull/4157/commits/57d86ab061e8012cca513b6f6cd4de8afe56ea0e

I wanted to reuse the method `_duplicate_update_warning` - however, I noticed that if we have the underscore at the beginning it means it is a private method.

I think it would be nice to use this method for consistency in logs, do you recommend me renaming the method to remove the underscore or to duplicate the function?

joybytes: ok I will address the refactoring at a separate PR and just leave the current implementation as is

app/notifications/notifications_letter_callback.py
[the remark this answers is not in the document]
ipdigital24: I was thinking from readability point of view and this method could read like:
```
notification_id = parse_token(token)
check_token(or notification_id)_matches_payload(notification_id, request_data)
extract_properties_from_request(...)
process_letter_callback_data(...)
```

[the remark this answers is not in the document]
joybytes: https://github.com/alphagov/notifications-api/pull/4157/commits/5cd3a521088c819e9a1fd9486445bc75ac04da10

https://github.com/alphagov/notifications-api/pull/4157/commits/2312dfd55b8795200bd36b1e3c62a08957822a73

https://github.com/alphagov/notifications-api/pull/4157/commits/08842f711e6a8baa69869b3d83e2749afc5985b8

ipdigital24 (approved):

PR #4160, opened 2024-08-14
fix: manual csv download without row limit

joybytes:
## Summary
- there is an issue with manual csv download
- for customers with extended retention days of data, if they try to do a manual csv download it creates latency on our database

- To fix the issue:
- notifications-api : create an endpoint which provides the number of notifications for the intended data being retrieved
- notifications-admin : call the endpoint created here prior to allowing the CSV download and if notifications_count exceed the number of (? - still to be decided) - we direct the customer to do these CSV downloads via the API

### In this PR:
- schema validation for the request which includes (template_type and limit_days) using `notifications_filter_schema`
- `notifications_sent_count` response (int)
- the value of notifications_sent_count is retrieved from the fact(ft)_billing table : 
- - the table is filtered by template type, vst_date and service_id
- - values for notifications_sent column are then summed to provide the notifications_sent_count

```
SELECT SUM(ft_billing.notifications_sent) AS sum_1
FROM   ft_billing
WHERE  ft_billing.service_id = '20028f41-0f26-4cbd-a7c4-ff06ccbbca15' :: uuid
       AND ft_billing.bst_date >= '2023-08-02'
       AND ft_billing.notification_type IN ( 'sms', 'email' ) 
```
## Related task:
[advise service users to use api instead of using admin page to download notifications if volume is very large](https://trello.com/c/VYDXkKxu/911-advise-service-users-to-use-api-instead-of-using-admin-page-to-download-notifications-if-volume-is-very-large)

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -516,7 +516,7 @@ def count_notifications_for_service(service_id):
     template_types = multidict.getlist("template_type")
 
     # set default values for template_types and limit_days
-    if limit_days is None or limit_days == 0:
+    if not limit_days:
         limit_days = 7
 
     if not template_types:
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -506,6 +506,22 @@ def get_all_notifications_for_service(service_id):
     )
 
 
+@service_blueprint.route("/<uuid:service_id>/notifications/count", methods=["GET"])
+def count_notifications_for_service(service_id):
+    data = notifications_filter_schema.load(request.args)
+
+    multidict = MultiDict(data)
+
+    notification_count = notifications_dao.count_notifications_for_service(
+        service_id=service_id,
+        status=multidict.getlist("status"),
+        template_type=multidict.getlist("template_type"),
+        limit_days=data.get("limit_days"),
+    )
+
+    return jsonify({"count": notification_count}), 200
+
+
 @service_blueprint.route("/<uuid:service_id>/notifications/<uuid:notification_id>", methods=["GET"])
 def get_notification_for_service(service_id, notification_id):
     notification = notifications_dao.get_notification_with_personalisation(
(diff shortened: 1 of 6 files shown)


app/service/rest.py
```diff
@@ -506,6 +506,21 @@ def get_all_notifications_for_service(service_id):
...
+@service_blueprint.route("/<uuid:service_id>/notifications/count", methods=["GET"])
+def count_notifications_for_service(service_id):
+    data = notifications_filter_schema.load(request.args)
+
+    multidict = MultiDict(data)
+
+    notification_count = fact_billing_dao.get_count_of_notifications_sent(
+        service_id=service_id,
+        template_types=multidict.getlist("template_type"),
+        limit_days=data.get("limit_days"),
```
leohemsted, line 518: do we want to default these two query params to sensible defaults? or error if they're not supplied?

maybe limit_days=7, template_type=['sms', 'email', 'letter'] but i'm open if you have other thoughts (or if the schema already does this by default and i didnt realise)

leohemsted (approved):
This looks like good stuff 👍

app/service/rest.py
joybytes: https://github.com/alphagov/notifications-api/pull/4160/commits/db48dee691df1fc0d41abd2e5aa2e14f839e1bfe

app/service/rest.py
```diff
@@ -509,13 +510,22 @@ def get_all_notifications_for_service(service_id):
...
 def count_notifications_for_service(service_id):
     data = notifications_filter_schema.load(request.args)
-
+    limit_days = data.get("limit_days")
     multidict = MultiDict(data)
 
+    template_types = multidict.getlist("template_type")
+
+    # set default values for template_types and limit_days
+    if limit_days is None or limit_days == 0:
```
leohemsted, line 519: cos `0` is falsey, we can just streamline this if statement like we do the template_types below
```suggestion
    if not limit_days:
```

joybytes: https://github.com/alphagov/notifications-api/pull/4160/commits/bc6b677fe2132c9506a042caa0b838ee80b63df9

leohemsted (approved):

PR #4165, opened 2024-08-20
API clients: make thread-local, allowing use of persistent requests sessions

risicle:
https://trello.com/c/ORGrd1jn/498-upgrade-docker-images-for-our-ecs-apps-to-debian-bookworm

Individual commit messages contain more details - note caveat about service callback session reuse.

The use of requests sessions should reduce CPU usage resulting from continual re-parsing of the CA bundle, which in turn should allow us to upgrade to OpenSSL 3.x with little trouble.

Ran the FTs against this on dev-a with `merged_api_workers` disabled: https://concourse.notify.tools/teams/dev-a/pipelines/deploy-notify/jobs/deploy/builds/144

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -3,6 +3,8 @@ import random
 import string
 import time
 import uuid
+from contextvars import ContextVar
+from functools import partial
 from time import monotonic
 
 from celery import current_task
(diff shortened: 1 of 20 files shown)


risicle:
#4179 is my preferred approach to this now

PR #4170, opened 2024-08-28
Only return unsubscribe reports with requests

quis:
When we start to delete old unsubscribe requests we will leave behind the reports to which they were joined.

We don’t want to delete these reports completely, since it’s useful for us to keep a record of what happened when, and how many requests there were. We do a similar thing for `Job` objects, where they stay in the database for all time.

However it’s not useful to show these old reports on the dashboard, because if a user clicks through to them they won’t be able to access the underlying requests.

So this pull request modifies the function which returns the list of reports to the admin app, excluding those which no longer have any requests attached to them.

### DIFF
--- a/tests/app/dao/test_unsubscribe_request_dao.py
+++ b/tests/app/dao/test_unsubscribe_request_dao.py
@@ -1,3 +1,5 @@
+from uuid import uuid4
+
 from sqlalchemy import desc
 
 from app.constants import EMAIL_TYPE
@@ -8,6 +10,7 @@ from app.dao.unsubscribe_request_dao import (
     get_latest_unsubscribe_request_date_dao,
     get_unsubscribe_request_by_notification_id_dao,
     get_unsubscribe_request_report_by_id_dao,
+    get_unsubscribe_request_reports_dao,
     get_unsubscribe_requests_data_for_download_dao,
     get_unsubscribe_requests_statistics_dao,
 )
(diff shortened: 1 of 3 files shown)


tests/app/dao/test_unsubscribe_request_dao.py
```diff
@@ -293,6 +295,50 @@ def test_assign_unbatched_unsubscribe_requests_to_report_dao(sample_service):
...
 
+def test_unsubscribe_reports_only_includes_those_with_requests(sample_service):
+    unsubscribe_request_report_1 = UnsubscribeRequestReport(
+        id=uuid4(),
+        count=123,
+        earliest_timestamp=midnight_n_days_ago(4),
+        latest_timestamp=midnight_n_days_ago(0),
+        service_id=sample_service.id,
+    )
+    create_unsubscribe_request_reports_dao(unsubscribe_request_report_1)
```
CrystalPea, line 306: creating of an unsub report is quite lengthy, it would be good to make it into a function in `https://github.com/alphagov/notifications-api/blob/1f8e5b1bc2c966352ef7f13a082ffa5f3e0e8465/tests/app/db.py` to keep our tests easier to take in at a glance.

CrystalPea (approved):
Looks good, I left a comment on test set-up 🙌🏼

CrystalPea (approved):

PR #4171, opened 2024-08-28
Updated phonenumber validation code to use allow_landline flag

rparke:
Two changes:

1) Adds permissions (international and landline) as flags to the phonenumbers validation method so that it can be used universally for all validation.

2) Separates out validation from parsing so that an instance of a phonenumber object represents a possible number, and validation is called explicitly to check whether this possible number can be sent to by a service. This makes the class more flexible and improves code readability.


quis (approved):
One optional formatting change, but generally good to go

PR #4173, opened 2024-08-29
Rewrite unsubscribe requests tests to use a helper

quis:
For lots of objects we create reguarly in tests we have a little helper function that:
- makes the code less verbose
- handles creating the associated objects automatically

We didn’t have this for unsubscribe requests. Now we do.

I’ve also modified some of the tests to assert against actual hard values, rather than results of queries. This makes them less brittle and harder to hide bugs in.

### DIFF
--- a/tests/app/dao/test_unsubscribe_request_dao.py
+++ b/tests/app/dao/test_unsubscribe_request_dao.py
@@ -1,12 +1,11 @@
-from uuid import uuid4
+from datetime import datetime
 
-from sqlalchemy import desc
+from freezegun import freeze_time
 
 from app.constants import EMAIL_TYPE
 from app.dao.unsubscribe_request_dao import (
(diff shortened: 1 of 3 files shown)


CrystalPea (commented):
Nice one, great to see these tests shortened :')

Should we do the same for the unsubscribe reports? I know there is less to shorten there, but it feels more right - it follows a pattern.

Also we had a Slack chat yesterday about some new tests not cleaning up after themselves - don't we have to introduce `notify_db_session` to test definition to avoid that, or did you find another way?

quis:
@CrystalPea I’ve added that helper now too.

Not sure what was going on with my tests yesterday but these all use `sample_service` which inherits from `notify_db_session` so should be all good.

rparke (approved):

PR #4174, opened 2024-08-29
connection pool: effectively disable parallel query by default

risicle:
https://trello.com/c/AGXtyaAe/887-disable-parallel-query-by-default

See extensive inline comment.

This has absolutely no test coverage, same as all of the sqlalchemy engine hooks in the unit tests.

I made an attempt at making the hooks work for the unit tests, but the result is ugly and has a number of caveats which would add to the complexity of the fixtures.


risicle:
I've now also successfully deployed this to dev-a and run the functional tests against it.

robinjam (approved):

PR #4178, opened 2024-09-02
Add `created_at` column to unsubscribe request reports

quis:
So we can perform actions a certain nuber of days after a report is first downloaded, we need to know when it was first downloaded. This is also the same time it gets created. Therefore we can use the pattern of a `created_at` column.

Need to deploy this as a migration first before adding the column to the model.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0461_user_research_email
+0462_unsubscribe_created_at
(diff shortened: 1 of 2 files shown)


quis:
Squawk is failing with this error, which isn’t relevant to us:
```
  note: A timestamp field without a timezone can lead to data loss, depending on your database session timezone.
  help: Use timestamptz instead of timestamp for your column type.
```

rparke (approved):

rparke (approved):

PR #4179, opened 2024-09-02
API clients: make thread-local, allowing use of persistent requests sessions (app context version)

risicle:
https://trello.com/c/ORGrd1jn/498-upgrade-docker-images-for-our-ecs-apps-to-debian-bookworm

~Depends on https://github.com/alphagov/notifications-utils/pull/1147~

An alternative to https://github.com/alphagov/notifications-api/pull/4165 - the principal difference being that this references `current_app` in the lazy-construction of the clients, which means:

 - we can do away with the weird error-throwing early-factory thing that otherwise needs to be replaced in `current_app`
 - however, the clients can only ever be accessed from within an app context
 - it has sensible(ish) behaviour when multiple app contexts are pushed

This last point might not seem hugely relevant on this app, but when it comes to the `admin` app, the test fixtures do all sorts of funny things with app contexts.

The main purpose of following this pattern on `api` is consistency with what I expect to do on `admin`.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -3,6 +3,8 @@ import random
 import string
 import time
 import uuid
+from collections.abc import Callable
+from contextvars import ContextVar
 from time import monotonic
 
 from celery import current_task
@@ -25,6 +27,7 @@ from notifications_utils.clients.redis.redis_client import RedisClient
 from notifications_utils.clients.signing.signing_client import Signing
 from notifications_utils.clients.statsd.statsd_client import StatsdClient
 from notifications_utils.clients.zendesk.zendesk_client import ZendeskClient
+from notifications_utils.local_vars import LazyLocalGetter
 from sqlalchemy import event
 from werkzeug.exceptions import HTTPException as WerkzeugHTTPException
 from werkzeug.local import LocalProxy
@@ -42,20 +45,12 @@ db = SQLAlchemy()
 migrate = Migrate()
 ma = Marshmallow()
 notify_celery = NotifyCelery()
-firetext_client = FiretextClient()
-mmg_client = MMGClient()
-aws_ses_client = AwsSesClient()
-aws_ses_stub_client = AwsSesStubClient()
-dvla_client = DVLAClient()
 signing = Signing()
 zendesk_client = ZendeskClient()
 statsd_client = StatsdClient()
 redis_store = RedisClient()
 cbc_proxy_client = CBCProxyClient()
-document_download_client = DocumentDownloadClient()
 metrics = GDSMetrics()
-
-notification_provider_clients = NotificationProviderClients()
 
 api_user = LocalProxy(lambda: g.api_user)
 authenticated_service = LocalProxy(lambda: g.authenticated_service)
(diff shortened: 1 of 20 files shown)


risicle:
-> draft only because it contains a TMP commit for bumping the utils to an as-yet-unmerged version

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -651,9 +651,9 @@ def test_post_email_notification_validates_personalisation_send_a_file_values(
     expected_filename,
 ):
     mocker.patch("app.celery.provider_tasks.deliver_email.apply_async")
-    document_download_mock = mocker.patch("app.v2.notifications.post_notifications.document_download_client")
-    document_download_mock.upload_document.side_effect = (
-        lambda service_id, content, is_csv, confirmation_email, **kwargs: f"{content}-link"
+    document_download_upload_document_mock = mocker.patch(
+        "app.document_download_client.upload_document",
```
robinjam, line 655: Bit of a nitpick, but couldn't you have achieved the same effect with a smaller/easier to read diff by just changing `"app.v2.notifications.post_notifications.document_download_client"` to `"app.document_download_client"` (leaving the rest as-is)?

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1131,8 +1131,10 @@ def test_post_notification_with_document_upload(
     )
 
     mocker.patch("app.celery.provider_tasks.deliver_email.apply_async")
-    document_download_mock = mocker.patch("app.v2.notifications.post_notifications.document_download_client")
-    document_download_mock.upload_document.side_effect = lambda service_id, content, is_csv, **kwargs: f"{content}-link"
+    document_download_upload_document_mock = mocker.patch(
+        "app.document_download_client.upload_document",
```
robinjam, line 1138: Ditto here

app/__init__.py
```diff
@@ -65,6 +59,91 @@
...
+            for getter in (
+                get_firetext_client,
+                get_mmg_client,
+            )
+        },
+        email_clients={
+            getter.expected_type.name: LocalProxy(getter)
+            for getter in ((get_aws_ses_stub_client,) if current_app.config["SES_STUB_URL"] else (get_aws_ses_client,))
+        },
+    ),
```
robinjam, line 125: I think I need an explanation of what's happening here, I don't really understand

risicle: Two dictionary comprehensions, generating dicts of the `expected_type`'s `.name` mapped against a `LocalProxy` to its getter.

The `email_clients` one switches its getter based on `SES_STUB_URL`.

tests/app/v2/notifications/test_post_notifications.py
risicle: It wouldn't be _exactly_ the same effect, no.

Patching the `document_download_client` will cause the name binding _in the target module_ to be switched out to a mock object. Any code that accesses the `document_download_client` from somewhere else will get the original unmocked client.

Patching `document_download_client.upload_document` will patch the referenced object's method, which should have an effect globally.

When I see the former it gives me anxiety, when I see the latter I have more confidence it's been mocked adequately.

robinjam (approved):
I _think_ this LGTM. I'm not thrilled about the `memo_resetters` mechanism but I also can't think of a better way to do it off the top of my head.

(Actually I'm not thrilled about any of this but I think it's the best we can do within the constraints of Flask 😅)

PR #4185, opened 2024-09-11
Make DVLA callback API update `NotificationLetterDespatch`

klssmith:
https://trello.com/c/rpy5t1Dh/887-make-dvla-callback-api-update-notificationletterdespatch

For each letter notification, we store in the `notifications_letter_despatch` table the notification ID, the despatchDate (ie: the date when it was printed, which is the date that DVLA will bill us for in their invoice), and the costThreshold which is one of sorted or unsorted - these refer to two different pricing models that DVLA apply based on volumes of letters that we send.

Best reviewed commit by commit, but change now ensures that we write to the `notifications_letter_despatch` table when we get a letter callback from the DVLA.

Still to do: We need to think more about error handling and what happens if the `process_letter_callback_data` task fails part way through - what do we do about retries, how do we identify which tasks failed etc. I think that can be split out into a separate story.

### DIFF
--- a/app/celery/process_letter_client_response_tasks.py
+++ b/app/celery/process_letter_client_response_tasks.py
@@ -8,13 +8,14 @@ from app.constants import (
 from app.dao.notifications_dao import (
     _duplicate_update_warning,
     dao_get_notification_or_history_by_id,
+    dao_record_letter_despatched_on_by_id,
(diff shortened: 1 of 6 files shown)


joybytes (approved):
lgtm, well done!

PR #4189, opened 2024-09-13
Update requester invite approved email

spatel033:
When a user successfully requests to join a service they should get an email request has been approved. More details on [this card](https://trello.com/c/FHhogwve/904-send-a-different-invite-email-if-the-user-has-requested-to-join-a-service).


migrations/versions/0464_requester_invite_approved.py
```diff
@@ -0,0 +1,89 @@
...
+            false,
+            current_timestamp,
+            '{current_app.config["NOTIFY_USER_ID"]}'
+        )
+        ON CONFLICT DO NOTHING
+        """
+    )
+
+
+def downgrade():
```
rparke, line 88: Shouldn't DB migrations always have a downgrade?

migrations/versions/0464_requester_invite_approved.py
```diff
@@ -0,0 +1,89 @@
...
+Sign in to GOV.UK Notify to get started.
+
+Thanks
+
+GOV.​UK Notify
+https://www.gov.uk/notify"""
+)
+
+
+def upgrade():
```
rparke, line 31: If we are just adding DB records, and not changing the DB schema, (unless there's some subtlety I'm not getting which ... there may be), adding records should just be done _via_ DAO's.

app/service_invite/rest.py
```diff
@@ -31,36 +31,25 @@
...
-            ),
-        },
-        notification_type=EMAIL_TYPE,
-        api_key_id=None,
-        key_type=KEY_TYPE_NORMAL,
-        reply_to_text=invited_user.from_user.email_address,
-    )
-
-    send_notification_to_queue(saved_notification, queue=QueueNames.NOTIFY)
+    if is_join_request:
```
leohemsted, line 44: given this flow's so different

"create an invite in the database, send that user an email with a link to join service"

"vs person has already been added to service, give them a confirmation email saying so"

i feel like these could be separate endpoints?

that would make this a lot tidier and much easier to reason about when looking at eg access logs

leohemsted (commented):
does this work? it used to send an invite, now it just tells them "you've been added to the service" but does it actually add them? does the request to join a service flow still work after this code goes in?

i need to have a think about this story overall - sorry but i think theres been some confusion, probably my fault, about whether this story should have been picked up at all.

spatel033:
Closing because of requirement changes

PR #4196, opened 2024-09-23
TMP notifications-utils: 84.0.0 -> ???, update requirement freezing accordingly

risicle:
This is an example app update to correspond with changes in https://github.com/alphagov/notifications-utils/pull/1155

`copy_config()` is now done during `freeze-requirements` and the results should be committed so we no longer have any bootstrapping issues. using `pip-sync` means we have to freeze our test requirements too as `pip-sync` will only install explicitly requested packages, not second-tier dependencies.


quis:
Superceded by https://github.com/alphagov/notifications-api/pull/4256

PR #4206, opened 2024-10-02
fetch_billing_data_for_day: omit service id filter clause if `None`

risicle:
...rather than performing a query mentioning every `service_id` by uuid. This might not speed the query up but will at least make it more digestible in rds insights.


app/dao/fact_billing_dao.py
```diff
@@ -552,7 +550,7 @@ def _email_query():
                 NotificationAllTimeView.created_at >= start_date,
                 NotificationAllTimeView.created_at < end_date,
                 NotificationAllTimeView.notification_type == notification_type,
-                NotificationAllTimeView.service_id.in_(service_ids),
+                *(() if service_ids is None else (NotificationAllTimeView.service_id.in_(service_ids),)),
```
leohemsted, line 553: nice little one-liner that i hadn't thought about doing before ✅

leohemsted (approved):

PR #4208, opened 2024-10-02
Create ServiceJoinRequest in db when user requests to join a service

spatel033:
Create the join service request in the background for the existing flow. More details on [this card](https://trello.com/c/zHU4O9SL/974-create-servicejoinrequest-in-db-when-user-requests-to-join-a-service).


CrystalPea (approved):
LGTM

PR #4209, opened 2024-10-03
use `EventletTimeoutMiddleware` if using eventlet

risicle:
See https://github.com/alphagov/notifications-utils/pull/1157

~This successfully raises an exception after 30s, which is what we previously expected our gunicorn timeouts to be set to everywhere (and is what they're set to on our non-eventlet gunicorn apps - template-preview-web & antivirus-web). Perhaps we should be setting it to 60s though - what our cloudfront and ALBs supposedly use - because this will cause the same visible behaviour as we currently get.~

:point_up: Changed my mind - switched this to 60s for the reasons suggested above.

I can demonstrate this working on dev-a to anyone interested.

Note that even though it will stop this app processing a request after N seconds, that doesn't mean this cancellation will necessarily be propagated to whatever this app is waiting on (database, another http api..). That sort of thing would require more plumbing.

_It would be useful if a developer would check this doesn't break their local development workflow before I merge this_. It should have no effect for them because most development modes don't use eventlet by default.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -27,6 +27,7 @@ from notifications_utils.clients.redis.redis_client import RedisClient
 from notifications_utils.clients.signing.signing_client import Signing
 from notifications_utils.clients.statsd.statsd_client import StatsdClient
 from notifications_utils.clients.zendesk.zendesk_client import ZendeskClient
+from notifications_utils.eventlet import EventletTimeout
 from notifications_utils.local_vars import LazyLocalGetter
 from sqlalchemy import event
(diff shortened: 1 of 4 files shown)


risicle:
Chris raised a point that made me think harder about this.. one thing about this middleware is it's a naive timeout - not a "read" timeout and won't necessarily do the right thing for a streaming request.

But we have to decide what the Right Thing _is_ for a streaming request.

risicle:
Actually, looking deeper, I think this middleware will only have an effect until the initial response is returned. After that, the request can still take as long as it wants to iterate through the remaining chunks.

PR #4210, opened 2024-10-03
nightly notification deletion: stagger sub-task invocations over 1h

risicle:
To hopefully avoid saturating database iops when thousands of these tasks are requested simultaneously.

Using a `stagger_total_period` instead of an individual step size to avoid a huge number of sub-tasks being spread over an excessively long time. Making this configurable should allow us to trigger ad-hoc invocations with different periods.

~Untested :muscle:~

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -132,10 +132,13 @@ def delete_letter_notifications_older_than_retention():
     _delete_notifications_older_than_retention_by_type("letter")
 
 
-def _delete_notifications_older_than_retention_by_type(notification_type):
+def _delete_notifications_older_than_retention_by_type(
+    notification_type,
+    stagger_total_period=timedelta(hours=1),
+):
     flexible_data_retention = fetch_service_data_retention_for_all_services_by_notification_type(notification_type)
 
-    for f in flexible_data_retention:
+    for i, f in enumerate(flexible_data_retention):
         day_to_delete_backwards_from = get_london_midnight_in_utc(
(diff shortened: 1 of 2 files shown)


app/celery/nightly_tasks.py
```diff
@@ -160,14 +164,15 @@ def _delete_notifications_older_than_retention_by_type(notification_type):
 
     service_ids_to_purge = service_ids_that_have_sent_notifications_recently - service_ids_with_data_retention
 
-    for service_id in service_ids_to_purge:
+    for i, service_id in enumerate(service_ids_to_purge):
```
joybytes, line 167: very minor, non-blocker:

would it be useful to only stagger if
```
if len(service_ids_to_purge) > 1:
```
check otherwise it becomes a bit redundant

joybytes (approved):
lgtm

app/celery/nightly_tasks.py
risicle: Well, in the single service id case, `i` is only ever `0`, so the countdown gets set to 0 which afaik is the default.

Though you do raise an interesting point. If it's called with e.g. 3, it will still spread them over the hour. Which is perhaps a little absurd 🤷

risicle:
(😭 formatter)

ipdigital24 (approved):
discussed with Rob on the working of the code

risicle:
Looks like this has introduced a flaky test because the order we iterate through the services is non-deterministic. Fixing.

PR #4211, opened 2024-10-04
manually create and delete enum in existing alembic migration

leohemsted:
there's a bug/design flaw in alembic where if you autogenerate a migration script, it'll create the enum implicitly, but won't include it in the downgrade. This is due to complex reasons that I kinda glazed over.[^1][^2]

With this migration, if you run downgrade and then upgrade again it fails due to the enum not being dropped by the downgrade, and thus already existing for the upgrade.

Lets drop the enum manually in the downgrade script. create the enum manually (not because we need to, but just to give a bit of clarity as to what's going on behind the scenes)

[^1]: https://github.com/sqlalchemy/alembic/issues/89
[^2]: https://github.com/sqlalchemy/alembic/issues/868#issuecomment-867914756

### DIFF
--- a/migrations/versions/0464_create_svc_join_requests.py
+++ b/migrations/versions/0464_create_svc_join_requests.py
@@ -5,8 +5,11 @@ from sqlalchemy.dialects import postgresql
 revision = "0464_create_svc_join_requests"
 down_revision = "0463_backfill_created_at"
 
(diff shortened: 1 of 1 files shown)


joybytes (approved):

joybytes (approved):
thanks for fixing this, I will keep this PR saved as a reminder for types and downgrades!

PR #4215, opened 2024-10-08
nightly notification deletion: make tests more tolerant of iteration order

risicle:
Fixup from #4210


quis:
Shouldn’t this line mean the tests don’t need the order to be deterministic?
https://github.com/alphagov/notifications-api/blob/c55790e45d62fe26f96707943e0d58c9eabfdd9f/tests/app/celery/test_nightly_tasks.py#L566

risicle:
The issue is that in the tested code, the `countdown=` numbers are assigned based on iteration order, and without special attention getting the correct `countdown=` value paired up with the right service uuid is 50/50

quis (approved):
Ah, got it, so it doesn’t matter which service gets which countdown value.

risicle:
Nope.

PR #4217, opened 2024-10-10
Update get service-join-request to include requester and updated_by

spatel033:
We have requirement to display the name of service, requester and status_changed_by. Updated db query to include the user and service and also updated response payload.
https://trello.com/c/pQovKftH/976-create-new-approver-flow-for-request-to-join-a-service


joybytes (approved):
nice !

PR #4219, opened 2024-10-11
control timeout configuration using config/env vars

risicle:
Doing a few things here to make some of the database tweaks we do and timeouts we set more consistent and controllable via environment variables (which will become useful when we e.g. want to use different `statement_timeout` values for different daemons.

Replacement for  #4218

See commits for individual detail..

Successfully tested on dev-a https://concourse.notify.tools/teams/dev-a/pipelines/deploy-notify/jobs/test/builds/52

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -138,11 +138,15 @@ class Config:
     AWS_REGION = "eu-west-1"
     INVITATION_EXPIRATION_DAYS = 2
     NOTIFY_APP_NAME = "api"
+
     SQLALCHEMY_ENGINE_OPTIONS = {
         "pool_size": int(os.environ.get("SQLALCHEMY_POOL_SIZE", 5)),
         "pool_timeout": 30,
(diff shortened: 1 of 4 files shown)


app/config.py
```diff
@@ -146,6 +147,15 @@ class Config:
             "options": "-c statement_timeout=1200000",
```
robinjam, line 147: Is it worth also removing this? I know the `SET statement_timeout` in `connect()` will overwrite it but it seems confusing to configure the same thing in two places

risicle: Indeed perhaps I should remove it... I'm just a little unsure whether its replacement gets run for things like our alembic migrator.

risicle:
Yup it looks like alembic etc don't heed our sqlalchemy event hooks - perhaps they're not using a connection pool at all?

So we should leave the connection options in place, but I'll add a comment

robinjam (approved):

PR #4222, opened 2024-10-16
feat: update service join request

joybytes:
### Summary
This PR introduces a new endpoint that allows updating the status of a service join request. The endpoint also handles assigning permissions to users when the request is approved.

### Endpoint:
POST /update-service-join-request-status/<uuid:request_id>

### Changes:
**Service Join Request Status Update:**

- Accepts a request to update the status of a service join request (pending, approved, rejected).
- Handles validation of the input payload using service_join_request_update_schema.
- If the service join request is not found, returns a 404 response.

**Add user to service:**

- When the status is set to approved, it adds the user to a service and sets the permission
- This is applied by dao_add_user_to_service

**Send approved email**
- If the user request is approved, an email is sent to the user 
- This is handled by send_service_join_request_decision_email

**Cancel Pending requests for the same service by the same user**
- Pending service join requests are cancelled for a user when one of their requests is approved, ensuring the system reflects the most up-to-date request. 
- The cancellation is service-specific, meaning only requests for the same service as the approved request will be canceled. Any requests the user has for other services remain unaffected

### Ticket
[Update the request status when request to join service approved/rejected](https://trello.com/c/7jW26Z4s/969-update-the-request-status-when-request-to-join-service-approved-rejected)

### DIFF
--- a/app/dao/service_join_requests_dao.py
+++ b/app/dao/service_join_requests_dao.py
@@ -1,6 +1,9 @@
+from datetime import datetime
+from typing import Literal
 from uuid import UUID
 
 from app import db
+from app.constants import SERVICE_JOIN_REQUEST_STATUS_TYPES
 from app.dao.dao_utils import autocommit
 from app.models import ServiceJoinRequest, User
 
@@ -27,3 +30,27 @@ def dao_get_service_join_request_by_id(request_id: UUID) -> ServiceJoinRequest |
         .options(db.joinedload("contacted_service_users"))
         .one_or_none()
     )
+
+
+@autocommit
+def dao_update_service_join_request(
+    request_id: UUID,
+    status: Literal[*SERVICE_JOIN_REQUEST_STATUS_TYPES],
+    status_changed_by_id: UUID,
+    reason: str = None,
+) -> ServiceJoinRequest | None:
+    service_join_request = dao_get_service_join_request_by_id(request_id)
+
+    if not service_join_request:
+        return None
+
+    if status:
+        service_join_request.status = status
+        service_join_request.status_changed_by_id = status_changed_by_id
+        service_join_request.status_changed_at = datetime.utcnow()
+
+    if reason is not None:
+        service_join_request.reason = reason
+
+    db.session.add(service_join_request)
+    return service_join_request
--- a/app/dao/service_join_requests_dao.py
+++ b/app/dao/service_join_requests_dao.py
@@ -3,7 +3,11 @@ from typing import Literal
 from uuid import UUID
 
 from app import db
-from app.constants import SERVICE_JOIN_REQUEST_STATUS_TYPES
+from app.constants import (
+    SERVICE_JOIN_REQUEST_CANCELLED,
+    SERVICE_JOIN_REQUEST_PENDING,
+    SERVICE_JOIN_REQUEST_STATUS_TYPES,
+)
 from app.dao.dao_utils import autocommit
 from app.models import ServiceJoinRequest, User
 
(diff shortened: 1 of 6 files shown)


app/service/rest.py
```diff
@@ -1322,3 +1328,38 @@ def get_service_join_request(request_id: uuid.UUID):
         raise InvalidRequest(message=f"Service join request with ID {request_id} not found.", status_code=404)
 
     return jsonify(service_join_request.serialize()), 200
+
+
+@service_blueprint.route("/update-service-join-request-status/<uuid:request_id>", methods=["PUT"])
```
spatel033, line 1333: Any reason of using `PUT`? :thinking: We have been using mainly `POST` in notify and this API is required to update partial data

app/schema_validation/service_join_request.py
```diff
@@ -10,3 +10,39 @@
...
+                    "send_emails",
+                    "send_letters",
+                    "manage_api_keys",
+                    "platform_admin",
+                    "view_activity",
+                ],
+            },
+            "description": "List of permissions being granted or modified",
+        },
+        "status_changed_by": {"type": "string", "format": "uuid"},
```
spatel033, line 37:
```suggestion
        "status_changed_by_id": {"type": "string", "format": "uuid"},
```

app/service/rest.py
joybytes: changed to POST

app/schema_validation/service_join_request.py
```diff
@@ -10,3 +10,39 @@
...
+                "type": "string",
+                "enum": [
+                    "manage_users",
+                    "manage_templates",
+                    "manage_settings",
+                    "send_texts",
+                    "send_emails",
+                    "send_letters",
+                    "manage_api_keys",
+                    "platform_admin",
```
klssmith, line 31: Optional, but we might want to remove `platform_admin` from here for safety, since it should never be possible to set that this way and we don't want users to ever have it.

app/schema_validation/service_join_request.py
```diff
@@ -10,3 +10,39 @@
...
+                    "view_activity",
+                ],
+            },
+            "description": "List of permissions being granted or modified",
+        },
+        "status_changed_by_id": {"type": "string", "format": "uuid"},
+        "status": {
+            "type": "string",
+            "enum": ["pending", "approved", "rejected"],
```
klssmith, line 40: Looks like `cancelled` [has been added](https://github.com/alphagov/notifications-api/pull/4221) as an option since this was originally written.

app/service/rest.py
```diff
@@ -1322,3 +1328,38 @@ def get_service_join_request(request_id: uuid.UUID):
         raise InvalidRequest(message=f"Service join request with ID {request_id} not found.", status_code=404)
 
     return jsonify(service_join_request.serialize()), 200
+
+
+@service_blueprint.route("/update-service-join-request-status/<uuid:request_id>", methods=["POST"])
+def update_service_join_request(request_id: uuid.UUID):
+    data = request.get_json()
+
+    try:
```
klssmith, line 1337: Minor thing, but we don't need to put error handling round `validate` because it throws a error which we already handle with an error handler. So we will already return a `400` with the messages.

app/dao/service_join_requests_dao.py
```diff
@@ -27,3 +34,44 @@ def dao_get_service_join_request_by_id(request_id: UUID) -> ServiceJoinRequest |
...
+    if pending_requests:
+        for request in pending_requests:
+            dao_update_service_join_request(
+                request_id=request.id,
+                status=SERVICE_JOIN_REQUEST_CANCELLED,
+                status_changed_by_id=approver_id,
+                reason="system update due to cancellation",
+            )
+
+    return
```
spatel033, line 77: Is this intentional `return`?

app/service/rest.py
```diff
@@ -1322,3 +1328,72 @@ def get_service_join_request(request_id: uuid.UUID):
...
+        "dashboard_url": f"{admin_base_url}/services/{service_id}",
+    }
+
+
+def send_service_join_request_decision_email(requester_email_address: str, template_id: uuid, personalisation: dict):
+    notify_service = dao_fetch_service_by_id(current_app.config["NOTIFY_SERVICE_ID"])
+
+    saved_notification = persist_notification(
+        template_id=template_id,
+        template_version=1,
```
klssmith, line 1389: You'll need to get the current version of the template to send, which means that we need to get the template out of the database. There's an example of where we've done that in this endpoint you could copy https://github.com/alphagov/notifications-api/blob/98ade3b931eb4d2f4aa232802aa134ee48f18b59/app/organisation/rest.py#L332

klssmith (approved):

PR #4224, opened 2024-10-16
add script that tests downgrade scripts for new migrations

leohemsted:
if this script finds new migration files, it will

* create a blank DB
* upgrade to the latest head (ie: all new migration files)
* downgrade to the current main to make sure the downgrade scripts run
* upgrade again to make sure the downgrade scripts left things in a sensible state
* fail if there were any sql errors etc

as this just runs the whole branch's diff, this _doesn't_ cover cases where an individual downgrade script is bad, but all when combined with the other downgrade scripts in that branch it works out ok (imagine if you have two versions in a PR, but the downgrade for both alembic versions only happens in the function for one of them).

But i think this is okay as a first step

next steps:

- [ ] create a new task in the PR concourse job
  - fail concourse 
- [ ] maybe make a lil jinja template to format the output into a nice github comment?

### DIFF
--- a/scripts/test_sql_migration_downgrade.sh
+++ b/scripts/test_sql_migration_downgrade.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+BOLDGREEN="\033[1;32m"
+ENDCOLOR="\033[0m"
+
+export FLASK_APP=application.py
+export NOTIFY_ENVIRONMENT="development"
+
+# what is the database currently pointing at on main?
+CURRENT_VERSION=$(curl -s https://raw.githubusercontent.com/alphagov/notifications-api/main/migrations/.current-alembic-head)
+
+# what is the most recent version? note this depends on us only ever having one head
+NEWEST_VERSION=$(flask db heads | tail -n 1 | cut -d " " -f 1)
+
+if [[ "${CURRENT_VERSION}" == "${NEWEST_VERSION}" ]]; then
+    echo "No migrations to check"
+    exit 0;
+fi
+
+
+# delete any existing test DB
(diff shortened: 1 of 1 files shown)


scripts/test_sql_migration_downgrade.sh
```diff
@@ -0,0 +1,40 @@
...
+
+SQLALCHEMY_DATABASE_URI="postgresql://localhost/migration_test" flask db upgrade > /dev/null 2>&1
+
+# make sure downgrade can run succesfully
+echo -e "${BOLDGREEN}============== running downgrade ==============${ENDCOLOR}"
+SQLALCHEMY_DATABASE_URI="postgresql://localhost/migration_test" flask db downgrade
+
+# make sure downgrade has left everything okay
+echo -e "${BOLDGREEN}============ running upgrade again ============${ENDCOLOR}"
+SQLALCHEMY_DATABASE_URI="postgresql://localhost/migration_test" flask db upgrade $CURRENT_VERSION
```
klssmith, line 36: It's not clear to me why we're doing `flask db upgrade $CURRENT_VERSION` here and not `flask db upgrade`? With this line as it is, we're not re-running the migration that was added, it's just trying to make the db match the version on main, so I'm seeing this script pass with a migration file I would expect to fail

leohemsted: oops that's a typo - i meant to add this line to `flask db downgrade` (so that if theres more than one downgrade script in a PR, it'll test all of them)

leohemsted: fixed in fe6f9bf14

klssmith (approved):

PR #4226, opened 2024-10-17
Update get service-join-request to include requester and updated_by

spatel033:
We have requirement to display the name of service, requester and status_changed_by. Updated db query to include the user and service and also updated response payload.

https://trello.com/c/pQovKftH/976-create-new-approver-flow-for-request-to-join-a-service

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2825,14 +2825,14 @@ class ProtectedSenderId(db.Model):
 @dataclass
 class SerializedServiceJoinRequest:
     service_join_request_id: str
-    requester_id: str
     service_id: str
     created_at: str
     status: str
     status_changed_at: str | None
-    status_changed_by_id: str | None
     reason: str | None
     contacted_service_users: list[str]
+    status_changed_by: User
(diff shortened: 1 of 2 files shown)


app/models.py
```diff
@@ -2873,12 +2873,24 @@ class ServiceJoinRequest(db.Model):
     def serialize(self) -> SerializedServiceJoinRequest:
```
joybytes, line 2873: I like the User type on both Requester and Status Changed By
However, I think we should remove the requester_id from the model if we are using requester:User type

This means we would need refactoring on the table schema and on the `dao_create_service_join_request`. 
On `dao_create_service_join_request` we can leave the signature the same (requester_id) but change the logic of the function to get the User.

I just think that having both requester and requester_id in the class `ServiceJoinRequest` seems redundant.

joybytes: ignore the above, my brain clearly was not functioning properly  - sorry about that!

joybytes (approved):
nice!

PR #4228, opened 2024-10-22
nightly tasks: don't use a `countdown` greater than `visibility_timeout`

risicle:
Fixups for #4207 and  #4210

Celery/kombu's SQS support is imperfect and the [caveats note](https://docs.celeryq.dev/en/v5.4.0/getting-started/backends-and-brokers/sqs.html#caveats) that using a `countdown` greater than the `visibility_timeout` will likely cause duplicate tasks. It looks like this is happening to us (though luckily it hasn't caused drastic ill effects so far).

So reduce the max `countdown` time we're using to less than 5 minutes as we're currently using a 5 minute 10 second `visibility_timeout`.

Going further it may be nice if we found a way to issue a warning if trying to use a task `countdown`/`eta` beyond the `visibility_timeout`, but let's get this fixed in the meantime.

### DIFF
--- a/app/celery/reporting_tasks.py
+++ b/app/celery/reporting_tasks.py
@@ -19,15 +19,19 @@ from app.dao.notifications_dao import get_service_ids_with_notifications_on_date
 
 @notify_celery.task(name="create-nightly-billing")
 @cronitor("create-nightly-billing")
-def create_nightly_billing(day_start=None):
-    # day_start is a datetime.date() object. e.g.
-    # up to 4 days of data counting back from day_start is consolidated
+def create_nightly_billing(
+    day_start=None,
+    n_days=10,
+    stagger_total_period_seconds=timedelta(minutes=5).seconds,
+):
+    # day_start is a datetime.date() object. i.e. up to n_days days of data counting
+    # back from day_start is consolidated
     if day_start is None:
         day_start = convert_utc_to_bst(datetime.utcnow()).date() - timedelta(days=1)
     else:
         # When calling the task its a string in the format of "YYYY-MM-DD"
         day_start = datetime.strptime(day_start, "%Y-%m-%d").date()
-    for i in range(10):
+    for i in range(n_days):
         process_day = (day_start - timedelta(days=i)).isoformat()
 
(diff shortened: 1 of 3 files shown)


leohemsted (dismissed):

app/celery/reporting_tasks.py
```diff
@@ -19,23 +19,27 @@
 
 @notify_celery.task(name="create-nightly-billing")
 @cronitor("create-nightly-billing")
-def create_nightly_billing(day_start=None):
+def create_nightly_billing(
+    day_start=None,
+    n_days=10,
+    stagger_total_period=timedelta(minutes=5),
```
leohemsted, line 25: is timedelta serialisable? or will this throw an error when it tries to put it on sqs?

app/celery/reporting_tasks.py
```diff
@@ -19,23 +19,27 @@
 
 @notify_celery.task(name="create-nightly-billing")
 @cronitor("create-nightly-billing")
-def create_nightly_billing(day_start=None):
+def create_nightly_billing(
+    day_start=None,
+    n_days=10,
```
ipdigital24, line 24: what does `n_days` mean?

app/celery/reporting_tasks.py
risicle: This... is a good question.

risicle: Simpler to just use seconds innit

app/celery/reporting_tasks.py
risicle: Number of days.

risicle: If we had mathematical notation available to us this would be $N_{days}$ but we don't.

ipdigital24: :-)
probably not for this pr but I was confused with `n_days = 10 (number of days to process)` vs the comment `up to 4 days of data counting back from day_start is consolidated`

risicle: I'll update it.

leohemsted (approved):

PR #4233, opened 2024-10-29
models: add `pg_trgm` indexes for `Notification`'s `normalised_to` & `client_reference` fields

risicle:
https://trello.com/c/sHzruWY5/929-add-pgtrgm-index

These fields are searched by `dao_get_notifications_by_recipient_or_reference` with `LIKE`/`ILIKE` operators, which are unable to use a btree index as the search terms aren't left-anchored. A `pg_trgm` index can be used by these queries to massively reduce the number of rows that need to be scanned.

Using a GIN index as, despite the documentation suggesting it have slightly faster insertion performance than GiST (which is probably more critical to us), most peoples' real world experiences seem to favour GIN by a significant margin for all operations.

One poster even claims they've *never* managed to get GiST to outperform GIN for `pg_trgm` in any scenario.

Migration adds it with a `CREATE INDEX CONCURRENTLY` command to avoid locking tables.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1466,6 +1466,22 @@ class Notification(db.Model):
             "notification_type = 'email' OR unsubscribe_link is null",
             name="ck_unsubscribe_link_is_null_if_notification_not_an_email",
         ),
+        Index(
+            "ix_notifications_normalised_to_trgm",
+            "normalised_to",
+            postgresql_using="gin",
+            postgresql_ops={
+                "normalised_to": "gin_trgm_ops",
+            },
+        ),
+        Index(
+            "ix_notifications_client_reference_trgm",
+            "client_reference",
+            postgresql_using="gin",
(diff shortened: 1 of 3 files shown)


risicle:
Ohh I need to ensure the test database has pg_trgm installed now...

risicle:
@leohemsted opinion on squawk migration failure? I think we need to deliberately not do these in a transaction.

risicle:
Here I've added the extension to the migration for the sake of unit tests & local development, even though the RDS database(s) should already have the `pg_trgm` extension by default due to its presence in `db-setup.sh`.

This passes dev-a.

leohemsted (approved):
lets yolo deploy it before the 5pm letter deadline

CrystalPea (approved):
Ooh excited for this

PR #4236, opened 2024-10-31
Simplify collate task

leohemsted:
Tidy up the task a little bit - there's a bunch of complexity that is left over from when we used to group letters and zip them up

* processing postages separately
* ordering by service and timestamp
* returning entire notification object from the DB rather than just the ID

also added a lil logging and commenting.

ideally i'd like to reduce this to <2 minute runtime so that we don't have to pause deploys in the evening - i dont think this'll get us quite there but it's a nice to have


quis (approved):
Looks sensible to me 👍🏻

PR #4238, opened 2024-10-31
look at jobId instead of id in dvla letter callbacks

leohemsted:
* it was a bit confusing which ID was which, so i made the log line clearer
* i fixed us using `id` rather than `jobId` to refer to the notification ID
* i ensured celery task is called properly (some json serialisation issues, plus an incorrectly named kwarg)
* i improved logging so that if we have more issues while testing callbacks its easier for us to debug. these shouldnt contain PII so no issues there


klssmith (approved):

klssmith (approved):
Looks good to me. We could stop casting to an int here now, but it's not causing any issues to keep it https://github.com/alphagov/notifications-api/blob/da2ebfdc5947425a90d2eaff8504b7e0bfcfcf99/app/celery/process_letter_client_response_tasks.py#L44

leohemsted:
Good spot - might tidy that up in an upcoming PR that does a bit of playing around with that task

PR #4239, opened 2024-10-31
rename migration file to just 0468.py

leohemsted:
unfortunately we accidentally merged a PR with a migration that just had "0468" as the reference, rather than a name with a description. This has been deployed through to production now, so rather than updating the alembic_version field in the database, we should just change the filename and the .current-alembic-head file so we (and our unit tests and migration CI checks) don't get confused about which file is the most recent migration head

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0468_add_notifications_trgm_normto_cref_idxs
+0468
--- a/migrations/versions/0468.py
+++ b/migrations/versions/0468.py
@@ -6,6 +6,7 @@ from alembic import op
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql
 
+# was meant to be 0468_add_notifications_trgm_normto_cref_idxs but accidentally committed as this
 revision = "0468"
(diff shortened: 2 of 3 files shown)


klssmith (approved):

risicle (approved):
:cry: doh

leohemsted:
added a new test that fails on main but doesn't fail on this branch
```
    def test_revision_matches_filename(notify_db_session):
        migrations = sorted(glob.glob("migrations/versions/*.py"))
        head_migration_filename = pathlib.Path(migrations[-1]).stem
        head_migration_id = head_migration_filename.split(".")[0]

        results = notify_db_session.execute("select version_num from alembic_version").fetchall()
        assert len(results) == 1
>       assert (
            results[0].version_num == head_migration_id
        ), 'alembic head filename {head_migration_filename} does not match "revision" {results[0].version}'
E       AssertionError: alembic head filename {head_migration_filename} does not match "revision" {results[0].version}
E       assert '0468' == '0468_add_not...mto_cref_idxs'
E         - 0468_add_notifications_trgm_normto_cref_idxs
E         + 0468
```

risicle (approved):

PR #4242, opened 2024-11-01
Create new celery mock and fix some tests

leohemsted:
I was getting frustrated that we had some unit tests that were saying

```
@celery.task
def some_task(arg1, arg2):
  pass

def foo():
  some_task.apply_async(args=[1, 2, 3], queue='some-queue-name')

def test_foo(mocker)
  mock_task = mocker.patch('some_task.apply_async')
  mock_task.assert_called_with(args=[1, 2, 3], queue=ANY)
```

There's a bug here - This code'll never work as `some_task` has two arguments but we're trying to call it with three args!

So, I wrote a `mock_celery_task` fixture that does some clever bits. It will:

* make sure your args and kwargs match up with the underlying task definition
* make sure your arguments are json serializable (so that they could be sent to SQS)
* make sure your args and kwargs match the type hints in the task definition (if there are any)

----

This isn't perfect. Notable things I've found so far:

As it tries to json serialize all your inputs, if you have a Mock you're passing around it'll throw an error. I think this is okay, and maybe we just move away from throwing mocks about and move towards the actual values we expect things to return etc

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -549,7 +549,7 @@ class Test(Development):
     # but the database name is set in the _notify_db fixture
     SQLALCHEMY_RECORD_QUERIES = True
 
-    CELERY = {**Config.CELERY, "broker_url": "you-forgot-to-mock-celery-in-your-tests://"}
+    CELERY = {**Config.CELERY, "broker_url": "you-forgot-to-mock-celery-in-your-tests://", "broker_transport": None}
 
     ANTIVIRUS_ENABLED = True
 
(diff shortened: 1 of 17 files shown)


CrystalPea (approved):
Nice one 👏🏼 

I assume you made the `celery_mocker` fail to see it does the checks, right?

I wonder - it's the case of "who watches the watchers" - where we don't have tests checking that complex fixtures like this work as expected 😅

risicle:
Did you mean to include the letter-callbacks commit in this branch?

leohemsted:
@risicle @CrystalPea i _did_ mean to include those commits originally as i was testing with them as mervi noticed - they've now been squashed out and it should work 

however i've converted this to a draft pr as i have to tidy up some commit messages 🧹

rparke (approved):

PR #4244, opened 2024-11-01
in task, expect values deserialised from json

leohemsted:
being a bit more explicit with type hints should hopefully remind us that the values in the task are just json dumped, so aren't rich enums or date objects

----

note: i've tested this manually locally by using `curl` and sending an example payload they provided us, but with my own db's UUID swapped. seems to work end-to-end (as well as my local setup works, which is barely at all)

### DIFF
--- a/app/celery/process_letter_client_response_tasks.py
+++ b/app/celery/process_letter_client_response_tasks.py
@@ -20,11 +20,10 @@ from app.models import LetterCostThreshold
 
 @notify_celery.task(bind=True, name="process-letter-callback")
 def process_letter_callback_data(
-    self, notification_id: str, page_count: int, dvla_status: str, cost_threshold: str, despatch_date: str
+    self, notification_id: str, page_count: int, dvla_status: str, cost_threshold: str, despatch_date: date
 ):
     notification_id = uuid.UUID(notification_id)
     cost_threshold = LetterCostThreshold(cost_threshold)
-    despatch_date = date.fromisoformat(despatch_date)
 
     notification = dao_get_notification_or_history_by_id(notification_id)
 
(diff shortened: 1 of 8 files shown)


joybytes (approved):
nice!

klssmith:
With your second commit message of using the callback queue in order to be the same as email/sms, isn't it only when sending the callbacks to _services_ that we use the callbacks queue? We use "ses-callbacks" and "sms-callbacks" when processing receipts.

leohemsted:
oops, you're right, good spot. i'll need a new queue and to update the terraform for the worker to read from it 😩

app/celery/process_letter_client_response_tasks.py
```diff
@@ -12,10 +15,16 @@
...
 from app.exceptions import NotificationTechnicalFailureException
+from app.models import LetterCostThreshold
 
 
 @notify_celery.task(bind=True, name="process-letter-callback")
-def process_letter_callback_data(self, notification_id, page_count, dvla_status, cost_threshold, despatch_date):
+def process_letter_callback_data(
+    self, notification_id: str, page_count: int, dvla_status: str, cost_threshold: str, despatch_date: date
+):
+    notification_id = uuid.UUID(notification_id)
```
klssmith, line 25: Very minor, but you could update the type hint for `dao_record_letter_despatched_on_by_id` now
https://github.com/alphagov/notifications-api/blob/cdf76491a01e8ef8c242e31a6ebf246a8447e9cc/app/dao/notifications_dao.py#L901

klssmith (approved):

app/celery/process_letter_client_response_tasks.py
leohemsted: done 👍

klssmith (approved):

PR #4247, opened 2024-11-05
feat: Add folder permissions for service join update

joybytes:
## Summary
- Include folder permissions update during service join request flow
- We do this in the current invite-user journey, we just need to make sure the same is done for the service request journey

### Ticket:
[Let folder permissions be selected when approving a service join request](https://trello.com/c/70y7hQOW/1021-let-folder-permissions-be-selected-when-approving-a-service-join-request)


whpearson (approved):
Looks good

PR #4250, opened 2024-11-06
Notification: add multi-variate extended statistics including basic declarative/testing mechanism

risicle:
Individual commits have more detail.

The more interesting thing here is the addition of `__extended_statistics__` on a model, which is something that's only consumed by the new `test_extended_statistics_presence` test to ensure this "declaration" of extended statistics matches those as provisioned by the migrations.

Running these migrations should take negligible time - the new statistics aren't actually collected until the database does its next `ANALYZE` run for the table.


leohemsted (approved):

whpearson (approved):
Change makes sense. Passes locally. select * from pg_statistic_ext looks sensible and the test seems useful.

PR #4263, opened 2024-11-15
Update user's auth_type when service join request is approved 

klssmith:
The form to choose user permissions in admin will now show radio buttons
to select the user's auth type if the service has email auth enabled.
This change ensures we save the option that was selected.

In the case where a service does not have email auth enabled but the
user has no phone number, we can't update their auth type so keep it as
"sms_auth".


joybytes (approved):
⭐

kr8n3r (approved):
works a treat

PR #4264, opened 2024-11-15
`Notification`: add `ix_notifications_failed_service_id_composite`

risicle:
Notifications with failed status are comparatively rare and tend to sit in cold regions of the table. Added to them being composed of several different individual status types (which, searched for together, rules out a lot of standard index operations), postgres finds it hard to query these efficiently.

This is a partial index only covering `NOTIFICATION_STATUS_TYPES_FAILED` (which conveniently at time of writing is the same set of status types requested together by admin), so it should be a small index that should allow such a query to completely disregard the `notification_status` clause and start reading out the results in `created_at` order with very little more filtering.

This is aimed at speeding up pages such as `/services/.../notifications/email?status=failed` which can be very slow.

I should probably add a comment to admin alongside this making a similar note to check api's indexes if changing the status type groups on that end.

### DIFF
--- a/app/constants.py
+++ b/app/constants.py
@@ -18,6 +18,7 @@ NOTIFICATION_RETURNED_LETTER = "returned-letter"
 
 # Raw notification status values grouped into types
 NOTIFICATION_STATUS_TYPES_FAILED = [
+    # if changing this group, update ix_notifications_failed_service_id_composite to match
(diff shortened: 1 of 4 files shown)


risicle:
Squawk expected to fail because of `CREATE INDEX CONCURRENTLY`

whpearson (approved):
Looks good

PR #4265, opened 2024-11-18
Refactor: create table service_join_request_contacted_users

joybytes:
## Summary
- Created a new table service_join_request_contacted_users.
- Copied data from soon to be deprecated contacted_users to newly created table (PR that [follows this one](https://github.com/alphagov/notifications-api/pull/4270))
- Refactored ServiceJoinRequest model to reference new table.

## Tested
- downgrade [script passes](https://github.com/alphagov/notifications-api/blob/fe6f9bf14e97261344ba21fc2b207ff8b105a657/scripts/test_sql_migration_downgrade.sh) with success

## Ticket:
https://trello.com/c/kgwSzxw2/1051-rename-contactedserviceusers-table

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2908,7 +2908,7 @@ class SerializedServiceJoinRequest:
 
 
 contacted_users = db.Table(
-    "contacted_users",
+    "service_join_request_contacted_users",
     db.Model.metadata,
     db.Column(
         "service_join_request_id", UUID(as_uuid=True), db.ForeignKey("service_join_requests.id"), primary_key=True
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0476_notifications_failed_idx
+0477_create_sjr_contacted_users
--- a/migrations/versions/0477_create_sjr_contacted_users.py
+++ b/migrations/versions/0477_create_sjr_contacted_users.py
@@ -1,0 +1,36 @@
+"""
+Create Date: 2024-11-22
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = "0477_create_sjr_contacted_users"
+down_revision = "0476_notifications_failed_idx"
+
+
+def upgrade():
+    # 1. Create service_join_request_contacted_users
+    op.create_table(
+        "service_join_request_contacted_users",
+        sa.Column(
+            "service_join_request_id",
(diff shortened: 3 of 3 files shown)


migrations/versions/0476_rename_contacted_users.py
```diff
@@ -0,0 +1,59 @@
...
+    # 2. Copy data from the old table to the new
+    op.execute(
+        """
+        INSERT INTO service_join_request_contacted_users (service_join_request_id, user_id)
+        SELECT service_join_request_id, user_id FROM contacted_users;
+        """
+    )
+
+    # 3. Drop the old table
+    op.drop_table("contacted_users")
```
leohemsted, line 35: i think these two lines (and the corresponding migrations) should be in a separate migration.

otherwise, while this PR is deploying, we run migrations first, and then in the ~10 minutes or so it takes to roll instances, Notify will break and serve error messages as it will try and read from `contacted_users` but the table no longer exists

leohemsted: this means that the data will be temporarily empty between the two migrations - but i think that's acceptable to do while this feature isn't used. but i'd rather ensure we never return 5xx errors

joybytes: Hi Leo, actually after talking with Katie and Saurabh the following has been suggested:

https://github.com/alphagov/notifications-api/pull/4265/commits/3adc11f21ea0ad045fa72fe4e8f0654f8f560c2b

Renaming the table instead of the dropping approach - this is the resource @spatel033 gave me: https://petegraham.co.uk/rename-postgres-table-with-alembic/

Let me know what you think

spatel033: Concourse lint-migration is failing due to renaming the table. There is an extra step is required to rename table with minor lock: https://squawkhq.com/docs/renaming-table

leohemsted: i guess the advantage of the view method is that it handles data transfer and ensuring reads/writes are uninterrupted. but it doesn't solve the issues from that blog where we'd want to ensure we rename indexes/sequences/etc

i'm easy on which one we choose - but either way i think two PRs feels inevitable

leohemsted: wait, you can insert to a view if its only got one from table 🤯 

every day's a school day!

spatel033 (approved):

PR #4266, opened 2024-11-18
`get_notifications_for_service`: add ability to omit `_personalisation` from query

risicle:
This involves a bit of juggling of options. Firstly this renames the existing argument `personalisation` (which dates back to 2016) to `with_template`, which better describes what it actually does.

Then this adds a new argument `with_personalisation` that, if set `False` (not the default) will defer the loading of the `_personalisation` db column.

`_personalisation` is generally the biggest field in a `notifications` row. The field is rarely used in bulk requests and including it (significantly) inflates the size of intermediate results in the query plan, increasing the size of e.g. external sorts. Omitting it will probably also reduce the amount postgres needs to fetch from its TOAST table.

I'd also be interested in changing the `service.get_all_notifications_for_service` view to omit the personalisation in responses to deliver a similar optimization for a greater number of calls. However that would involve an interface change so I've not done it here.

FWIW, this may not deliver a significant visible speedup for this view itself, but it should reduce the amount it contributes to database cache pollution.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -16,7 +16,7 @@ from notifications_utils.recipient_validation.phone_number import try_validate_a
 from notifications_utils.timezones import convert_bst_to_utc, convert_utc_to_bst
 from sqlalchemy import and_, asc, desc, func, literal, or_, union
 from sqlalchemy.dialects.postgresql import insert
-from sqlalchemy.orm import joinedload
+from sqlalchemy.orm import defer, joinedload, undefer
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.sql import functions
 from sqlalchemy.sql.expression import case
@@ -251,6 +251,7 @@ def get_notifications_for_service(
     limit_days=None,
     key_type=None,
     with_template=False,
+    with_personalisation=True,
     include_jobs=False,
     include_from_test_key=False,
     older_than=None,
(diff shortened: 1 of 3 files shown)


CrystalPea (commented):
Uhh, this didn't trip up any unit tests, then? 😳 

Would we like a unit test for that?

risicle:
Arguably the unit tests didn't catch this because they were written in "test one thing" style and didn't happen to write a test for this "thing".

CrystalPea:
@risicle would it be ok to add these missing tests? I think these would be:
test_get_notifications_for_service_includes_personalisation_by_default
test_get_notifications_for_service_without_personalisation
test_get_notifications_for_service_excludes_template_data_by_default
get_notifications_for_service_with_template

risicle:
Come to think, these aren't particularly easy things to test because sqlalchemy hides the result of these flags - a deferred field will be lazily loaded on access, similar to a relationship if it has not been `joinedload`ed. I'd have to poke around with sqlalchemy internals to check it's doing anything.

quis:
@risicle something like this seems to work, at least for the `personalisation` flag:

```python
@pytest.mark.parametrize("with_personalisation", (True, False))
def test_get_notifications_query_options(sample_template, with_personalisation):
    query = get_notifications_for_service(sample_template.service_id, with_personalisation=with_personalisation)
    sql = str(query._query_args["query"].options())
    assert ("personalisation" in sql.lower()) is with_personalisation
```

Think it’s worth having a test because otherwise it would be easy for someone else to subtly re-introduce a performance problem.

risicle:
I think I know how i can do it based on https://github.com/sqlalchemy/sqlalchemy/issues/5709#issuecomment-729689097

CrystalPea (approved):
Looks great, and good sleuthing on the tests :shipit:

PR #4268, opened 2024-11-20
feat: Extend protected_sender_ids to include organisation_id column

joybytes:
## Summary
- Added organisation_id column with foreign key reference to Organisation table
- Refactored endpoint to receive organisation as an argument - this is intentional so this field remains optional and it also prevents a breaking change
- If a sender_id is found but the organisation does not match, we return false

## Ticket:
https://trello.com/c/6jXOAfRJ/1034-incident-action-put-in-place-restrictions-on-sender-ids-that-can-be-used-by-an-organisation

### DIFF
--- a/app/protected_sender_id/rest.py
+++ b/app/protected_sender_id/rest.py
@@ -1,6 +1,6 @@
 from flask import Blueprint, jsonify, request
 
-from app.errors import InvalidRequest, register_errors
+from app.errors import register_errors
 from app.models import ProtectedSenderId
 from app.protected_sender_id.protected_sender_schema import protected_sender_request
 from app.schema_validation import validate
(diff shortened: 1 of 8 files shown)


app/protected_sender_id/rest.py
```diff
@@ -25,3 +25,27 @@ def check_if_sender_id_is_protected():
...
+    sender_id = payload["sender_id"].strip()
+    organisation_id = payload["organisation_id"].strip()
+
+    normalised_sender_id = "".join(sender_id.split()).lower()
+
+    # does not seem like we use index so maybe it's not necessary at all
+    result = ProtectedSenderId.query.filter(ProtectedSenderId.sender_id == normalised_sender_id).first()
+
+    if not result:
+        return jsonify({"protected": False}), 404  # sender_id does not exist
```
joybytes, line 46: I think this should be true

klssmith:
I haven't checked the code itself yet, but it would be good to go through how this would work with a PM (Lisa or Katie might know perhaps?) just to check that each protected ID would only belong to one organisation. If not, we'll have to take a different approach to the database tables

joybytes:
> I haven't checked the code itself yet, but it would be good to go through how this would work with a PM (Lisa or Katie might know perhaps?) just to check that each protected ID would only belong to one organisation. If not, we'll have to take a different approach to the database tables

thanks Katie,
I put this up on slack and it turns out it's best not to move forward with the story. So this was a very important point, thanks again  ⭐

PR #4271, opened 2024-11-25
Install packages using `uv`

quis:
[uv](https://github.com/astral-sh/uv) is:
- between 10× and 100× faster than pip
- a drop-in replacement for `pip install`, `pip compile`, etc

This PR gets it working both for local development and on CI.

Also includes a few additional tidy-ups for our dependency-management steps.


Makefile
```diff
@@ -15,7 +15,7 @@ PYTHON_EXECUTABLE_PREFIX := $(shell test -d "$${VIRTUALENV_ROOT}" && echo "$${VI
 
 .PHONY: bootstrap
 bootstrap: generate-version-file ## Set up everything to run the app
-	pip3 install -r requirements_for_test.txt
+	uv pip sync requirements_for_test.txt
 	createdb notification_api || true
 	(. environment.sh && flask db upgrade) || true
```
leohemsted, line 20: in my new `uv` managed world i dont have an accessible venv - rather uv wants you to use `uv run` to access things from within the venv. So if i run `make bootstrap` locally i now get `/bin/bash: flask: command not found. i can't even run `which python` or `which pip` on my machine - a python dev with no available python! 🤯 

The change would be:

```suggestion
	(. environment.sh && uv run --no-project flask db upgrade) || true
```

(we'd also want to update all the other bits like when we run python `python -c "...  copy_config()"` in freeze-reqs)

not sure if we need to do that now? i can work around it/deal with it. but just something to think about. somewhat surprised it works in docker 🤔

Makefile
```diff
@@ -99,10 +99,9 @@ watch-tests: ## Watch tests and run on change
 
 .PHONY: freeze-requirements
 freeze-requirements: ## Pin all requirements including sub dependencies into requirements.txt
-	pip install --upgrade pip-tools
-	pip-compile requirements.in
+	uv pip compile requirements.in
```
leohemsted, line 102: https://docs.astral.sh/uv/pip/compile/#locking-requirements

> Note by default the uv pip compile output is just displayed and --output-file / -o argument is needed to write to a file.

```suggestion
	uv pip compile requirements.in -o requirements.txt
```

(and the same for the requirements_for_test line below)

Makefile
```diff
@@ -15,7 +15,7 @@ PYTHON_EXECUTABLE_PREFIX := $(shell test -d "$${VIRTUALENV_ROOT}" && echo "$${VI
 
 .PHONY: bootstrap
 bootstrap: generate-version-file ## Set up everything to run the app
-	pip3 install -r requirements_for_test.txt
+	uv pip sync requirements_for_test.txt
```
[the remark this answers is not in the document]
quis: Yeah I’m open to having `install` locally and `sync` on CI
```suggestion
	uv pip install -r requirements_for_test.txt
```

Makefile
leohemsted: you know what, i can work around this by just running `source .venv/bin/activate` - i'll stick that in my direnv and leave it at that

requirements.txt
```diff
@@ -25,18 +21,18 @@ blinker==1.6.2
     #   flask
     #   gds-metrics
     #   sentry-sdk
-boto3[crt]==1.34.100
+boto3==1.34.100
```
quis, line 24: The `[crt]` here refers to these optional dependencies (`bototcore[crt]`):
https://github.com/boto/boto3/blob/2c3b93f41f145a828e999bd9da748ba4c2ef282c/setup.cfg#L11

`bototcore[crt]` itself refers to this optional dependency (`awscrt`):
https://github.com/boto/botocore/blob/1ad32855c799456250b44c2762cacd67f5647a6e/setup.cfg#L12

`awscrt` is already baked into this requirements file on line 13, so what uv is doing here by removing the brackets is fine.

leohemsted: yeah i did a lil digging myself yesterday

https://github.com/astral-sh/uv/issues/1595

choice quotes:

>  packages don't have access to what extras they were installed only, and the only thing they see is "is package X in my extras installed".

looks like pip-tools might be moving that way too anyway

PR #4273, opened 2024-11-25
`notifications_all_time_view`: use `UNION ALL` to combine tables

risicle:
It shouldn't be possible for a notification to be in both tables at once, and using `UNION ALL` allows the plan to omit a sorting/deduplication step that can be quite expensive.


robinjam (approved):
Good idea

leohemsted (approved):

PR #4276, opened 2024-11-26
Add fixtures for API client integration tests to functional test fixtures

robinjam:
[Trello card](https://trello.com/c/D1m5uleD/952-run-api-client-integration-tests-as-part-of-new-pipeline)

We're working to update the API client integration tests to use the same set of fixtures as the functional tests, so that they can be run in every environment (instead of being tied specifically to preview), without needing to manually "clickops" the required services and templates into each environment.

This will be useful when testing changes in a dev environment, and will also be necessary before we can achieve our goal of retiring the preview environment.

This PR adds the necessary additional fixtures to get the API clients working. It'll be followed up by PRs in each of the API client repos to make the necessary adjustments to have them use these new fixtures.

### DIFF
--- a/app/functional_tests_fixtures/__init__.py
+++ b/app/functional_tests_fixtures/__init__.py
@@ -39,6 +39,7 @@ from app.dao.service_email_reply_to_dao import (
     dao_get_reply_to_by_service_id,
 )
 from app.dao.service_inbound_api_dao import get_service_inbound_api_for_service, save_service_inbound_api
+from app.dao.service_letter_contact_dao import add_letter_contact_for_service, dao_get_letter_contacts_by_service_id
 from app.dao.service_permissions_dao import (
     dao_add_service_permission,
     dao_fetch_service_permissions,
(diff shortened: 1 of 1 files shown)


app/functional_tests_fixtures/__init__.py
```diff
@@ -472,3 +509,21 @@ def _create_service_inbound_api(service_id, user_id):
     inbound_api.bearer_token = "1234567890"
 
     save_service_inbound_api(inbound_api)
+
+
+def _create_inbound_sms(service, count):
+
+    num_existing = len(dao_get_inbound_sms_for_service(service.id, limit=count))
+
+    for _ in range(num_existing, count):
```
joybytes, line 518: Minor - feel free to ignore:
If I understand correctly we are trying to create inbound sms if num_existing is less than count with `range(num_existing, count)`, is that right?

we can do something like `num_to_create = max(0, count - num_existing)` to make it clear how many we’re iterating over and avoids any potential confusion with range. 

Then you could do `range(num_to_create)`

joybytes (approved):
lgtm with minor comment but it's a non-blocker

PR #4282, opened 2024-11-29
skip old letters in still_pending_virus_check task

leohemsted:
previously, we were just selecting all letters more than 10 mins old with status pending_virus_check. However, this involves checking statuses all the way back through time - given this task runs every ten minutes, we don't need to check the status of letters created a day ago as they'll already have been fixed for example.

For now, just start with 30 minutes as the maximum time, in case the scheduled task doesn't run for any reason.

If it doesn't run multiple times in a row, then we'd need to run the task manually with a longer margin at least once - so I've made it an optional parameter so we can manually trigger the task

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -815,12 +815,16 @@ def letters_missing_from_sending_bucket(seconds_to_subtract):
     return notifications
 
 
-def dao_precompiled_letters_still_pending_virus_check():
-    ten_minutes_ago = datetime.utcnow() - timedelta(seconds=600)
+def dao_precompiled_letters_still_pending_virus_check(max_minutes_ago_to_check):
+    earliest_timestamp_to_check = datetime.utcnow() - timedelta(minutes=max_minutes_ago_to_check)
+    ten_minutes_ago = datetime.utcnow() - timedelta(minutes=10)
 
     notifications = (
         Notification.query.filter(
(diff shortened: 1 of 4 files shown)


app/dao/notifications_dao.py
```diff
@@ -815,12 +815,16 @@ def letters_missing_from_sending_bucket(seconds_to_subtract):
     return notifications
 
 
-def dao_precompiled_letters_still_pending_virus_check():
-    ten_minutes_ago = datetime.utcnow() - timedelta(seconds=600)
+def dao_precompiled_letters_still_pending_virus_check(max_minutes_ago_to_check):
+    thirty_minutes_ago = datetime.utcnow() - timedelta(minutes=max_minutes_ago_to_check)
```
risicle, line 819: Can we not call this `thirty_minutes_ago` if it's actually just `max_minutes_ago_to_check`? ;)

leohemsted: oops yes we definitely can, good spot

leohemsted: Squashed

risicle (approved):
LGTM

Perhaps you'll want a variant of the check that also runs for a longer time period once every 24h?

leohemsted:
@risicle good thinking. i added one at 8pm as not much happens on notify at that time but we also dont have the nightly billing queries running then

risicle (approved):

PR #4283, opened 2024-11-29
`dao_get_uploads_by_service_id`: use literal `ServiceDataRetention` for letters_subquery

risicle:
Using literal values allows the planner to see them and here it's easy for us to do because we already know the `service_id` and `notification_type`.

### DIFF
--- a/app/dao/uploads_dao.py
+++ b/app/dao/uploads_dao.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from flask import current_app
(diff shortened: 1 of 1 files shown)


leohemsted (approved):
is there any merit at all to doing the same with the above query? we could remove the coalesce but we'd end up replacing it with something like

`case when notification_type = 'email' then 3 when notification_type = 'sms' then 7 end`

not sure if that offers any benefits

risicle:
Thought about it but wasn't sure.. didn't think the complication warranted it (for now at least).

Also.. the notification type is a property of the template table so it needs a join with that anyway.

PR #4284, opened 2024-12-02
job tasks: introduce `shatter_job_rows` to create individual row tasks

risicle:
In `process_job` and `process_incomplete_job` we're partly bottlenecked by the rate at which we can put SQS messages on the queue, causing the main job-processing task to take longer than it should. this is a problem because it holds a lock on the job row and it can sometimes get killed by ECS scale-in halfway through processing, leaving a zombie lock on the job row amongst possibly other things.

`shatter_job_rows` requires no database connection and its' sole responsibility is to turn a sequence of task args/kwargs pairs into individual tasks.

`DEFAULT_SHATTER_JOB_ROWS_BATCH_SIZE` is chosen with an eye to avoid nearing the SQS message limit for these shatter tasks.

This imports the "example equivalent" of `itertools.batched` which is introduced in python 3.12. we should remove our copy when we upgrade past that.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -22,10 +22,11 @@ from app import db, dvla_client, notify_celery, redis_store, statsd_client, zend
 from app.aws import s3
 from app.celery.letters_pdf_tasks import get_pdf_for_templated_letter
 from app.celery.tasks import (
+    get_id_task_args_kwargs_for_job_row,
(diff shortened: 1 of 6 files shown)


risicle:
Just realized this would probably require a two-stage rollout.

Will worry about that once it's reviewed.

leohemsted (approved):

leohemsted:
looks good (agree about two-stage rollout tho)

PR #4285, opened 2024-12-02
Notification: add `ix_notifications_service_id_ntype_created_at`

risicle:
Create this index alongside `ix_notifications_service_id_composite` to see which gets more extended use over time, then we can consider which to keep.

Most notifications queries are for more than one status type, which means the existing index almost never uses its ability to scan in `created_at` order.


risicle:
Squawk expected to fail due to index concurrently/transaction.

spatel033 (approved):

PR #4296, opened 2024-12-05
Fix join-live-service functional tests

spatel033:
Functional tests for join-a-live-service failing because of the can_ask_to_join_a_service flag is not enabled initially and domain does not exist. Update the org fixture to create org with required flag and domain.

### DIFF
--- a/app/functional_tests_fixtures/__init__.py
+++ b/app/functional_tests_fixtures/__init__.py
@@ -7,6 +7,7 @@ from flask import current_app
 from sqlalchemy.exc import NoResultFound
 
 from app.constants import (
+    CAN_ASK_TO_JOIN_SERVICE,
     EDIT_FOLDER_PERMISSIONS,
     EMAIL_AUTH,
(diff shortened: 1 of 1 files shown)


app/functional_tests_fixtures/__init__.py
```diff
@@ -279,7 +280,8 @@ def _create_organiation(org_name="Functional Tests Org"):
 
         dao_create_organisation(org)
 
-    dao_update_organisation(org.id, can_approve_own_go_live_requests=True)
+    org.set_permissions_list([CAN_ASK_TO_JOIN_SERVICE])
+    dao_update_organisation(org.id, domains=["digital5.cabinet-office.gov.uk"], can_approve_own_go_live_requests=True)
```
leohemsted, line 284: we define the email domain already on line 93 of this file - can we pass that email domain in to this function?

it'll avoid typos like this one :)

leohemsted (approved):

PR #4298, opened 2024-12-06
Handling duplicate unsubscribe requests

BlessedDev:
Currently duplicate requests can be made using the same unsubscribe link, hence there are occurrences of multiple instances of unsubscribe requests with the same `notification_id` in the `unsubscribe_request` table.

The changes proposed in this PR would prevent duplicate unsubscribe requests from being entered into the `unsubscribe_request` and being added to an unsubscribe request report, as described in this [card](https://trello.com/c/2rwsxzZd/165-duplication-of-one-click-unsubscribe-links-issue)

For the purpose of the work, a duplicate unsubscribe request is being defined as an unsubscribe_request that has
the same notification_id of a previously received unsubscribe request that has not yet been processed by the service that initiated the notification.

### DIFF
--- a/app/one_click_unsubscribe/rest.py
+++ b/app/one_click_unsubscribe/rest.py
@@ -8,7 +8,8 @@ from app.dao.notifications_dao import get_notification_by_id
 from app.dao.unsubscribe_request_dao import (
     create_unsubscribe_request_dao,
     get_unbatched_unsubscribe_requests_dao,
-    get_unsubscribe_request_reports_dao,
+    get_unsubscribe_request_reports_dao, get_unsubscribe_request_by_notification_id_dao,
+    get_unsubscribe_request_report_by_id_dao,
 )
 from app.errors import InvalidRequest, register_errors
 from app.models import UnsubscribeRequestReport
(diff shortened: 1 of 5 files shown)


tests/app/one_click_unsubscribe/test_one_click_unsubscribe.py
```diff
@@ -119,3 +136,39 @@ def test_invalid_one_click_unsubscribe_url_notification_id(client, sample_email_
...
+    assert result is True
+
+
+def test_is_duplicate_unsubscribe_request_for_batched_unprocessed_request(sample_service):
+    notification_id = create_unsubscribe_request_and_return_the_notification_id(
+        sample_service, is_batched=True, processed_by_service=False
+    )
+    # Simulate a duplicate unsubscribe request with the same notification_id
+    result = is_duplicate_unsubscribe_request(notification_id)
+    assert result is True
```
quis, line 174: Think this would be easier to follow if you rewrote it as one test with 
```python
@pytest.mark.parametrize("is_batched, proccessed_by_service, expected_duplicate", (…))
```

app/one_click_unsubscribe/rest.py
```diff
@@ -69,3 +75,17 @@ def create_unsubscribe_request_reports_summary(service_id):
...
+    A duplicate unsubscribe request is being defined as an unsubscribe_request that has
+    the same notification_id of a previously received unsubscribe request that has not yet been processed
+    by the service that initiated the notification.
+    """
+    if unsubscribe_request := get_unsubscribe_request_by_notification_id_dao(notification_id):
+        report_id = unsubscribe_request.unsubscribe_request_report_id
+        if not report_id:
+            return True
+        return False if get_unsubscribe_request_report_by_id_dao(report_id).processed_by_service_at else True
+    return False
```
quis, line 91: This is easier to follow with less nesting:
```suggestion
unsubscribe_request = get_unsubscribe_request_by_notification_id_dao(notification_id)

if not unsubscribe_request:
    return False

report_id = unsubscribe_request.report_id

if report_id and get_unsubscribe_request_report_by_id_dao(report_id).processed_by_service_at:
    return False 

return True
```

quis (changes requested):

app/one_click_unsubscribe/rest.py
BlessedDev: I take your point on deploying less nesting.

tests/app/one_click_unsubscribe/test_one_click_unsubscribe.py
BlessedDev: Agreed.

quis (approved):
Looks good 👍🏻

PR #4300, opened 2024-12-08
`dao_get_last_date_template_was_used`: restrict `notifications` table search to midnight

risicle:
Beyond last midnight we should have a record of any notifications in `ft_notification_status`, which should be faster to search (if it isn't, we can address that easily). As long as we ensure we can search notifications in `created_at` order (hence targeting it at a specific index), we can avoid scanning old, cache-cold rows.

Taking the template instance as an argument allows us to extract all the field values we need to fill out these queries while maintaining a convenient calling interface.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -88,17 +88,26 @@ FIELDS_TO_TRANSFER_TO_NOTIFICATION_HISTORY = [
 ]
 
 
-def dao_get_last_date_template_was_used(template_id, service_id):
+def dao_get_last_date_template_was_used(template):
+    uniform_now = datetime.now()
+
     # first, just check if there are any rows present for this template in the notification table.
     # we can use the ix_notifications_template_id. If there are rows, then lets check to find out exactly
     # when the most recent created date was (also checking key type test too)
-    if db.session.query(Notification.query.filter(Notification.template_id == template_id).exists()).scalar():
+    if db.session.query(Notification.query.filter(Notification.template_id == template.id).exists()).scalar():
         last_date_from_notifications = (
             db.session.query(functions.max(Notification.created_at))
             .filter(
-                Notification.service_id == service_id,
-                Notification.template_id == template_id,
+                Notification.template_id == template.id,
                 Notification.key_type != KEY_TYPE_TEST,
+                # beyond last midnight we should have a record of the notification in FactNotificationStatus
+                # which is faster to query
+                Notification.created_at >= get_london_midnight_in_utc(uniform_now),
+                # filtering by notification_type and service_id is technically redundant, but postgres can't
+                # be certain of this and specifying them allows ix_notifications_service_id_ntype_created_at
+                # to be used
+                Notification.notification_type == template.template_type,
+                Notification.service_id == template.service_id,
             )
             .scalar()
         )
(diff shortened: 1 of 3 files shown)


app/dao/notifications_dao.py
```diff
@@ -107,14 +116,14 @@ def dao_get_last_date_template_was_used(template_id, service_id):
...
+        FactNotificationStatus.query.filter(FactNotificationStatus.template_id == template.id).exists()
     ).scalar():
         last_date = (
             db.session.query(functions.max(FactNotificationStatus.bst_date))
             .filter(
-                FactNotificationStatus.template_id == template_id,
+                FactNotificationStatus.template_id == template.id,
                 FactNotificationStatus.key_type != KEY_TYPE_TEST,
-                FactNotificationStatus.bst_date > datetime.now() - relativedelta(years=1),
+                FactNotificationStatus.bst_date > uniform_now - relativedelta(years=1),
```
quis, line 126: Not a change needed in this PR, but if searching `FactNotificationStatus` is quick enough I wonder if we still need this line (and therefore could revert https://github.com/alphagov/notifications-admin/pull/5101)

risicle: It's never "quick enough" ;)

Srsly, I haven't done anything (yet) to make searching `FactNotificationStatus` any faster, so this PR probably doesn't improve the Worst Case Scenario very much.

I was considering adding an index on `(template_id, bst_date)` to accelerate this query.

risicle:
Searching for a use of a nonexistent template id by our biggest volume service:

Old query: I got bored of waiting for an `EXPLAIN ANALYZE` result

New query:

```
QUERY PLAN                                                                                                                     
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Result  (cost=17628.16..17628.17 rows=1 width=8) (actual time=0.997..0.998 rows=1 loops=1)
   Buffers: shared hit=2 read=2
   I/O Timings: shared/local read=0.978
   InitPlan 1 (returns $0)
     ->  Limit  (cost=0.56..17628.16 rows=1 width=8) (actual time=0.995..0.996 rows=0 loops=1)
           Buffers: shared hit=2 read=2
           I/O Timings: shared/local read=0.978
           ->  Index Scan Backward using ix_notifications_service_id_ntype_created_at on notifications  (cost=0.56..1974291.03 rows=112 width=8) (actual time=0.994..0.995 rows=0 loops=1)
                 Index Cond: ((service_id = '<our-biggest-service>'::uuid) AND (notification_type = 'email'::notification_type) AND (created_at IS NOT NULL) AND (created_at >= '2024-12-10 00:00:00'::timestamp without time zone))
                 Filter: (((key_type)::text <> 'test'::text) AND (template_id = '<nonexistent-template>'::uuid))
                 Buffers: shared hit=2 read=2
                 I/O Timings: shared/local read=0.978
 Planning:
   Buffers: shared hit=9
 Planning Time: 0.279 ms
 Execution Time: 1.014 ms
```

app/dao/notifications_dao.py
risicle: https://github.com/alphagov/notifications-api/pull/4305

CrystalPea (approved):

PR #4301, opened 2024-12-09
Remove code related to ftp and response files

klssmith:
This deletes the code related to receiving updates about letters via response files in S3. We now always get the updates via HTTPS callback instead, which means we can also stop checking the `LETTER_DELIVERY_CALLBACKS_ENABLED ` environment variable.

We can't just delete code - there's some that needs to be replaced (the middle commit). If using a test API key with `SEND_LETTERS_ENABLED` set to False, we were mocking the delivery receipt by putting a file in S3. This has now been updated to instead send a callback.


app/celery/research_mode_tasks.py
```diff
@@ -114,43 +175,15 @@ def firetext_callback(notification_id, to):
...
-    # on development we can't trigger SNS callbacks so we need to manually hit the DVLA callback endpoint
-    if current_app.config["NOTIFY_ENVIRONMENT"] == "development":
-        make_request("letter", "dvla", _fake_sns_s3_callback(upload_file_name), None)
-
-
-def _fake_sns_s3_callback(filename):
-    message_contents = '{"Records":[{"s3":{"object":{"key":"%s"}}}]}' % (filename)  # noqa
-    return json.dumps({"Type": "Notification", "MessageId": "some-message-id", "Message": message_contents})
+@notify_celery.task(bind=True, name="create-fake-letter-callback", max_retries=3, default_retry_delay=60)
+def create_fake_letter_callback(self, notification_id: uuid.UUID, billable_units: int, postage: str):
```
leohemsted, line 179: unfortunately i think we need to create this celery task in a separate PR first, or during the deploy the celery workers running the old code will error if they pick up one of these tasks without having a function to call for it

klssmith: I've split the code which adds the new task into [a separate PR](https://github.com/alphagov/notifications-api/pull/4303), and will rebase this PR after the new code has been added

klssmith: The code related to adding and using the new task for test letters has been added in https://github.com/alphagov/notifications-api/pull/4303 and https://github.com/alphagov/notifications-api/pull/4306, so this PR now solely deals with removing code.

risicle:
Hmm I think this has broken fresh deployments because we wait for the SNS subscription to be confirmed before the terraform completes.

PR #4305, opened 2024-12-10
`FactNotificationStatus`: add index on `template_id`, `bst_date`

risicle:
Aimed at improving search time of `dao_get_last_date_template_was_used`. See also https://github.com/alphagov/notifications-api/pull/4300

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2218,6 +2218,14 @@ class FactNotificationStatus(db.Model):
     created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
     updated_at = db.Column(db.DateTime, nullable=True, onupdate=datetime.datetime.utcnow)
 
+    __table_args__ = (
+        Index(
+            "ix_ft_notification_status_template_id_bst_date",
+            "template_id",
+            "bst_date",
+        ),
+    )
+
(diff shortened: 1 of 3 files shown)


leohemsted (approved):

risicle:
Might need to force-push to unstick the CI?

risicle:
Squawk expected to fail for normal index concurrently reasons

leohemsted:
@risicle

squawk is failing on prefer-robust-stmts - which specifically refers to the fact this runs outside a transaction, and thus the migration could theoretically fail after applying this statement and not be able to roll it back. (in practice this would be unlikely to happen in this script since we're not making any statements after this (except for the update alembic_version itself?)

in that case you'd end up with the alembic version NOT updated but the index created, so further attempts to re-apply would error

it wants you to set `IF NOT EXISTS`, which it looks like we can do via `if_not_exists=True`

https://squawkhq.com/docs/prefer-robust-stmts/
https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.create_index

risicle:
Ahhhh. I should probably actually read the failure shouldnt I

leohemsted (approved):

PR #4306, opened 2024-12-11
Start using new task to generate fake letter callbacks

klssmith:
The `create-fake-letter-callback` task was [previously added to the code](https://github.com/alphagov/notifications-api/pull/4303), this change starts using it to generate fake letter responses.

The exception it catches has been changed from to `HTTPError`, since `Exception` was not specific enough.

### DIFF
--- a/app/celery/research_mode_tasks.py
+++ b/app/celery/research_mode_tasks.py
@@ -221,7 +221,7 @@ def _fake_sns_s3_callback(filename):
 def create_fake_letter_callback(self, notification_id: uuid.UUID, billable_units: int, postage: str):
     try:
         send_letter_response(notification_id, billable_units, postage)
-    except Exception:
+    except requests.HTTPError:
         try:
             self.retry()
         except self.MaxRetriesExceededError:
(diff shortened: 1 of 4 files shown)


klssmith:
@leohemsted this is the PR split out from https://github.com/alphagov/notifications-api/pull/4303. There were a couple of comments you left there for when the task gets added here.
1. I've changed the task to only catch `HTTPError` instead of `Exception`
2. I've left the function that makes the API request as it is, so it just re-raises for every status. I think this is ok because it [matches how the SMS requests](https://github.com/alphagov/notifications-api/blob/728e422e97eac6f62c98bf095df190c529371982/app/celery/research_mode_tasks.py#L128-L139) are handled. I don't think we will get `4xx` statuses, so all error statuses will be `5xx` statuses. The JSON we're creating is always valid according to our `dvla_letter_callback_schema` schema, and if anything were to change, the unit tests should pick that up. I'm happy to change that though if you think it should be differentiating between the different error statuses.

leohemsted (approved):

PR #4308, opened 2024-12-12
`dao_get_notification_outcomes_for_job`: omit redundant `service_id` condition

risicle:
This is called repeatedly by `get_uploads_by_service`.

Filtering by `service_id` makes postgres use two indexes rather than one, resulting in a more complex query that touches more data.

(incidentally, it does this because multi-column statistics have limitations with nullable columns - otherwise I'd hope postgres' extended statistics for the notifications table would see that a simple filter in an index scan would probably remove 0 rows)

The added safety of filtering these results by service is a nice-to-have but I'd argue here we're only returning aggregated counts, not including any sensitive data, which hugely limits the worst-case-scenario.

It's hard to get exact performance figures for queries like this that "warm up" very quickly, but an `EXPLAIN (ANALYZE, BUFFERS) ...` showed that the new query touches about half the number of storage blocks than the previous did (cached or uncached).

### DIFF
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -30,10 +30,10 @@ from app.models import (
 from app.utils import midnight_n_days_ago
 
 
-def dao_get_notification_outcomes_for_job(service_id, job_id):
+def dao_get_notification_outcomes_for_job(job_id):
     notification_statuses = (
         db.session.query(func.count(Notification.status).label("count"), Notification.status)
-        .filter(Notification.service_id == service_id, Notification.job_id == job_id)
+        .filter(Notification.job_id == job_id)
         .group_by(Notification.status)
         .all()
     )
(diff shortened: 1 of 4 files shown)


app/dao/jobs_dao.py
```diff
@@ -30,10 +30,10 @@
 from app.utils import midnight_n_days_ago
 
 
-def dao_get_notification_outcomes_for_job(service_id, job_id):
+def dao_get_notification_outcomes_for_job(job_id):
     notification_statuses = (
```
leohemsted, line 34: idly thinking, we could be smarter and instead of just saying "look in notifications, if it's empty look in ft_notification_status", we could say `if job.created_at > last_midnight` before needing to query

actually never mind i just realised that this would mean the stats wouldn't be guaranteed up to date for jobs from yesterday with things still in sending, which is presumably why we did it this way

leohemsted (approved):

app/dao/jobs_dao.py
risicle: I am strongly considering another change where we detect jobs whose statuses are all "final", and aggressively redis-cache the result.

PR #4309, opened 2024-12-13
`Notification`: remove `ix_notifications_service_id_composite` index

risicle:
The vast majority of uses of this index have been taken over by `ix_notifications_service_id_ntype_created_at` - let's see if we can do without it.


risicle:
Yes, I've added `if_not_exists=True` to the reverse-migration even though we don't think it has any effect on our alembic/sqlalchemy version.. in case at some point we upgrade we can get the added robustness.

BlessedDev (approved):

PR #4310, opened 2024-12-13
fix SES rate limit exception handling

leohemsted:
the class is actually TooManyRequestsException, and we don't need to check for the message content

lil thing i noticed when we hit these errors the other day

### DIFF
--- a/app/clients/email/aws_ses.py
+++ b/app/clients/email/aws_ses.py
@@ -103,7 +103,7 @@ class AwsSesClient(EmailClient):
         except botocore.exceptions.ClientError as e:
             self.statsd_client.incr("clients.ses.error")
 
-            # http://docs.aws.amazon.com/ses/latest/DeveloperGuide/api-error-codes.html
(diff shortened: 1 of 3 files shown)


app/clients/email/aws_ses.py
```diff
@@ -106,14 +106,11 @@ def send_email(
             # http://docs.aws.amazon.com/ses/latest/DeveloperGuide/api-error-codes.html
```
quis, line 106: This URL doesn’t exist any more – equivalent seems to be [this one](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_SendEmail.html#API_SendEmail_Errors) (which does document both `TooManyRequestsException` and (via the link to [common errors](https://docs.aws.amazon.com/ses/latest/APIReference-V2/CommonErrors.html)) `ThrottlingException`)
```suggestion
            # https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_SendEmail.html#API_SendEmail_Errors
```

quis (approved):

PR #4312, opened 2024-12-15
feat: Remove inbound sms from service

joybytes:
1. Created endpoint to remove inbound sms for service(`remove_inbound_sms_for_service`)
This process is currently done manually with this query https://github.com/alphagov/notifications-manuals/wiki/Support-Runbook#query-to-remove-the-ability-to-receive-text-messages
This endpoint: 
- removes inbound_sms permission for the service
- removes service_sms_senders when inbound number id is present
- in `inbound_number` table disassociates service id from the number and based on the payload (archive: bool) we archive the number by setting the `active` field to false or release by leaving the `active ` field as true

2. Created endpoint to get the most recent usage date for the service inbound number (`get_most_recent_inbound_usage_date`):
 - This checks the following tables: notification, sms inbound, sms inbound history
 - This was created to provide a date of last usage on admin, when platform admins are looking to archive or release the number, they have this information promptly which can help with the decision

### Ticket:
https://trello.com/c/jC9B8l27/1056-build-platform-admin-endpoint-to-remove-inbound-sms-from-service


app/inbound_sms/rest.py
```diff
@@ -60,3 +71,33 @@ def get_inbound_by_id(service_id, inbound_sms_id):
...
+def remove_inbound_sms_for_service(service_id):
+    data = request.get_json()
+    validate(data, remove_inbound_sms_for_service_schema)
+
+    archive = data["archive"]
+
+    try:
+        dao_remove_service_permission(service_id, INBOUND_SMS_TYPE)
+        dao_remove_inbound_sms_senders(service_id)
+        archive_or_release_inbound_number_for_service(service_id, archive)
```
spatel033, line 86: Here are three database operations. If the second or last transaction fails, data inconsistency may arise. For instance, if a user attempts to archive a number and dao_remove_inbound_sms_senders encounters an exception, archive_or_release_inbound_number_for_service will not be executed. Consequently, only dao_remove_service_permission will be successful, leading to a partial update and potential data corruption.

Transaction allows series of database operations are either all committed to the database or all rolled back if an error occurs. We can follow similar approach https://github.com/alphagov/notifications-api/blob/81af15c03ca0ec7557101752291d8305e658217f/app/dao/services_dao.py#L310C1-L331C1

joybytes: that's a good point actually! well spotted

joybytes: I added transaction method and refactored tests
https://github.com/alphagov/notifications-api/pull/4312/files#:~:text=service_sms_sender_dao%20import%20dao_remove_inbound_sms_senders-,with%20transaction()%3A,-dao_remove_service_permission(service_id

app/dao/service_sms_sender_dao.py
```diff
@@ -106,3 +108,15 @@ def _raise_when_no_default(old_default):
     # check that the update is not updating the only default to false
     if not old_default:
         raise Exception("You must have at least one SMS sender as the default.", 400)
+
+
+def dao_remove_inbound_sms_senders(service_id: UUID, commit=True):
```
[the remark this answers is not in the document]
joybytes: This was done intentionally
I noticed while testing the transactions through tests and manually with `@autocommit` that these were not working as expected. 

[This test in my PR](https://github.com/alphagov/notifications-api/blob/3c78ed613209c80f31d57a0701139bc778a4730d/tests/app/dao/test_inbound_numbers_dao.py#L169) in an example of it. -`test_dao_remove_inbound_sms_for_service_failure`

If you leave the autocommits, you will see that the state before and after the transaction run are different which is not what we would want.

spatel033 (approved):

PR #4314, opened 2024-12-18
Drop `daily_sorted_letter` table

klssmith:
We stopped writing to this table when we started receiving letter delivery status via HTTPS callback instead of via response files in S3. However, since 2023 we've been using the `ft_billing_letter_despatch` table for letter billing and to store more detailed information about which letters were classed as `sorted` vs `unsorted` so this data is not needed.


whpearson (approved):
changes look good. Nothing in the database since november.

PR #4320, opened 2024-12-27
Update join a service templates

spatel033:
The contents for join a service templates have been updated in prodution and current versions of dev are out dated. Update the migration files so dev templates are updated when creating new environment.


quis (commented):
Editing the existing migrations won’t have any effect. If they have already been run (for example in production) then they won’t get run again

spatel033:
> Editing the existing migrations won’t have any effect. If they have already been run (for example in production) then they won’t get run again

We need to fix templates when setting up new dev environment, all existing environments would remain unchanged.

quis (approved):

PR #4323, opened 2025-01-02
jobs: add new job status `JOB_STATUS_FINISHED_ALL_NOTIFICATIONS_CREATED`

risicle:
When `check_for_missing_rows_in_completed_jobs` first accounts for all rows of a `JOB_STATUS_FINISHED` job, it now transitions it to `JOB_STATUS_FINISHED_ALL_NOTIFICATIONS_CREATED`, allowing further runs of `check_for_missing_rows_in_completed_jobs` to avoid counting notifications for it again.

This change will be visible to internal-api users, so https://github.com/alphagov/notifications-admin/pull/5340 needs to be released before this.


risicle:
🟢 https://concourse.notify.tools/teams/dev-a/pipelines/deploy-notify/jobs/test/builds/184 🎉

leohemsted (approved):

PR #4331, opened 2025-01-08
Remove nvm

spatel033:
The change is part of the removing nvm as part of ITHC report. More details on [this card](https://trello.com/c/jFpYYlGh/1087-remove-nvm-from-docker-images-verify-shell-script-before-executing). We use node only for `squawk-cli` so removing the installation via npm and installing via pip/uv.

squawk-migration-check has been successful with this change.
image

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -162,16 +162,3 @@ RUN \
 RUN \
     service postgresql start \
     && su - postgres -c "psql -c \"alter user postgres with password 'postgres';\""
-
-
-ENV NODE_VERSION=20.10.0
-ENV NVM_VERSION=0.39.7
-
-RUN \
-    echo "Installing NVM & Node" \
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -162,16 +162,3 @@ RUN \
...
-
-ENV NODE_VERSION=20.10.0
-ENV NVM_VERSION=0.39.7
-
-RUN \
-    echo "Installing NVM & Node" \
-    && curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.sh | bash \
-    && source ~/.nvm/nvm.sh \
-    && nvm install $NODE_VERSION \
-    && npm install -g squawk-cli@1.1.0
```
klssmith, line 175: We have `squawk-cli` installed here for running the `lint-migrations` PR check, so do still need it

spatel033: Thank you for highlighting this. Installing `squawk-cli` via pip.

klssmith (approved):

PR #4332, opened 2025-01-09
Row numbers in Returned Letters Report to match row numbers

CrystalPea:
in the spreadsheet with recipient data uploaded by service users when they sent the letters.

This far, there has been a mismatch:

When our service users send letters in bulk ia Notify page, they do so by uploading a spreadsheet with recipient data and persinalisation. Each row in the spreadsheet becomes a letter.

In such a spreadsheet, first row is for column headers - for example address_line_1, address_line_2, address_line_3, reference. Then recipient rows start from row 2.

When we save the rows in notifications table in our db, we start from index 0 for first recipient. That means that there is a mismatch of 2 rows between our db and the spreadsheet that has been uploaded.

When collating data for Returned Letters Report, for some reason we have been adding 1 to the db row number - which meant there was still a mismatch - this time by one row.

This commit fixes the mismatch - from now on, the row number we present in returned letters report will match the row number in the uploaded spreadsheet.

### DIFF
--- a/app/dao/returned_letters_dao.py
+++ b/app/dao/returned_letters_dao.py
@@ -107,7 +107,9 @@ def fetch_returned_letters(service_id, report_date):
                 User.name.label("user_name"),
                 User.email_address,
                 Job.original_file_name,
-                (table.job_row_number + 1).label("job_row_number"),  # row numbers start at 0
+                # row numbers in notifications db table start at 0, but in spreadsheet uploaded by service user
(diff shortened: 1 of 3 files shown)


quis (approved):
Fine to merge once comms are out

rparke (approved):

quis:
Needs to be merged for Wednesday 29 January (what time?)

PR #4337, opened 2025-01-13
Remove redudant db commit when getting serialised service

spatel033:
We have investigated P1 504 outage and came across following possible issue. This PR is to remove db commit which was added previously to close the  transaction to improve the performance.

First we get redis error - [kibana](https://kibana.logit.io/s/9423a789-282c-4113-908d-0be3b1bc9d1d/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(message,request_time,status),filters:!(),index:'8ac115c0-aac1-11e8-88ea-0383c11b333a',interval:auto,query:(language:kuery,query:'%22redis%20error%22'),sort:!('@timestamp',desc)))

then pg connection closed error - [kibana](https://kibana.logit.io/s/9423a789-282c-4113-908d-0be3b1bc9d1d/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(message,request_time,status),filters:!(),index:'8ac115c0-aac1-11e8-88ea-0383c11b333a',interval:auto,query:(language:kuery,query:'%22connection%20already%20closed%22'),sort:!('@timestamp',desc)))

Possible scenario we found during the P1 504 outage:
- Redis get disconnected
-  We have [this service](https://kibana.logit.io/s/9423a789-282c-4113-908d-0be3b1bc9d1d/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(message,request_time,status),filters:!(),index:'8ac115c0-aac1-11e8-88ea-0383c11b333a',interval:auto,query:(language:kuery,query:'%22api%20key%20not%20found%22%20and%20%227f266c24-ff07-474c-ab73-fe43745d3711%22'),sort:!('@timestamp',desc))) bombarding auth request (200-450 auth request per min) with incorrect token
- All these requests now go to postgres
- When getting serialised service to get api_key for auth, it also does unnecessary db.session.commit() ??
- At the same time RDS has 66% commit queries

image
image
image
image

### DIFF
--- a/app/serialised_models.py
+++ b/app/serialised_models.py
@@ -12,7 +12,7 @@ from notifications_utils.serialised_model import (
 )
 from werkzeug.utils import cached_property
 
-from app import db, redis_store
+from app import redis_store
 from app.dao.api_key_dao import get_model_api_keys
 from app.dao.services_dao import dao_fetch_service_by_id
 
@@ -67,7 +67,6 @@ class SerialisedTemplate(SerialisedModel):
         )
 
         template_dict = template_schema.dump(fetched_template)
-        db.session.commit()
 
         return {"data": template_dict}
 
@@ -99,7 +98,6 @@ class SerialisedService(SerialisedModel):
         from app.schemas import service_schema
 
         service_dict = service_schema.dump(dao_fetch_service_by_id(service_id))
-        db.session.commit()
 
         return {"data": service_dict}
 
@@ -127,5 +125,4 @@ class SerialisedAPIKeyCollection(SerialisedModelCollection):
         keys = [
             {k: getattr(key, k) for k in SerialisedAPIKey.__annotations__} for key in get_model_api_keys(service_id)
         ]
-        db.session.commit()
         return cls(keys)


leohemsted:
https://github.com/alphagov/notifications-api/pull/2887/commits/46dbcb7e3659989648f91e87d855811a95ae9547

concurrent db connection count is not as big a concern for us right this second as it used to be

quis:
If it makes sense to remove this we should also remove the one for `SerialisedTemplate` here:
https://github.com/alphagov/notifications-api/blob/84e798118a2141feb4d7f6c364de913e8ce95bee/app/serialised_models.py#L70

…and `SerialisedAPIKeys` here:
https://github.com/alphagov/notifications-api/blob/84e798118a2141feb4d7f6c364de913e8ce95bee/app/serialised_models.py#L130

But (without a description in your PR) I’m not understanding the benefit/rationale for doing this now.

quis (approved):

leohemsted:
i'm still a little skeptical about this:

1. i dont understand why redis would fail - we havent had redis errors before/elsewhere
1. the redis errors we see are EventletTimeoutErrors. There are two reasons an eventlet could time out:
  * the IO takes over thirty seconds
  * the IO completed on time but the eventlet process wasn't able to carry on with the code execution in under 30 seconds 
1. i dont see why this would be contained to one specific api-web instance
1. 450 requests per min sounds like a lot but it's only 7.5 requests per second - whereas we have 35 concurrent instances with four processes each that can each handle 250 requests

spatel033:
> i'm still a little skeptical about this:
> 
> 1. i dont understand why redis would fail - we havent had redis errors before/elsewhere
> 2. the redis errors we see are EventletTimeoutErrors. There are two reasons an eventlet could time out:
> 
> * the IO takes over thirty seconds
> * the IO completed on time but the eventlet process wasn't able to carry on with the code execution in under 30 seconds
> 
> 1. i dont see why this would be contained to one specific api-web instance
> 2. 450 requests per min sounds like a lot but it's only 7.5 requests per second - whereas we have 35 concurrent instances with four processes each that can each handle 250 requests

Yes this PR will not solve the actual redis issue but it may stop 504 error when redis connection is closed

spatel033:
> > i'm still a little skeptical about this:
> > 
> > 1. i dont understand why redis would fail - we havent had redis errors before/elsewhere
> > 2. the redis errors we see are EventletTimeoutErrors. There are two reasons an eventlet could time out:
> > 
> > 
> > * the IO takes over thirty seconds
> > * the IO completed on time but the eventlet process wasn't able to carry on with the code execution in under 30 seconds
> > 
> > 
> > 1. i dont see why this would be contained to one specific api-web instance
> > 2. 450 requests per min sounds like a lot but it's only 7.5 requests per second - whereas we have 35 concurrent instances with four processes each that can each handle 250 requests
> 
> Yes this PR will not solve the actual redis issue but it may stop 504 error when redis connection is closed

Recent P1 stared with [eventlet timeout](https://kibana.logit.io/s/9423a789-282c-4113-908d-0be3b1bc9d1d/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:'2025-01-14T15:49:07.607Z',mode:absolute,to:'2025-01-14T16:27:58.786Z'))&_a=(columns:!(message,request_time,status),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'8ac115c0-aac1-11e8-88ea-0383c11b333a',key:application_name,negate:!f,params:(query:api-web,type:phrase),type:phrase,value:api-web),query:(match:(application_name:(query:api-web,type:phrase))))),index:'8ac115c0-aac1-11e8-88ea-0383c11b333a',interval:auto,query:(language:kuery,query:%22eventlet%22),sort:!('@timestamp',desc))) for `SerialisedAPIKeyCollection.from_service_id(self.id)`

PR #4338, opened 2025-01-14
auth: move "API authorised.." log message to after assigning `service_id`

risicle:
This means the log should be annotated with the `service_id`, which is useful.

Ideally we'd be using a logging system where "request-global" parameters like this could be easily queried, but we're not (yet), so this tweak should make life a little easier in the meantime.


quis (approved):
Makes sense. I can see the logging code gets the service ID from the global request object here: 
https://github.com/alphagov/notifications-utils/blob/7322d3573eb54285166cae84da22edfae8067f02/notifications_utils/logging/flask.py#L315

PR #4342, opened 2025-01-15
Add new international letter rates for 6 Feb 2025 onwards

klssmith:
International letter rates are changing on 6 February 2025, so this adds a migration to update them.

See https://trello.com/c/qZb1RQL8/1123-update-the-prices-in-the-notify-database-royal-mail-price-increase for details of the date and price changes to double check


klssmith:
Failing the Squawk check because Squawk can't understand the parameterized insert statements, but this should be safe and we've done migrations just like this before

PR #4346, opened 2025-01-17
remove functional test data

spatel033:
There are lots of test data in preview environment. This PR is for the removing the test data ([card](https://trello.com/c/bH9JnDjZ/1117-clear-functional-test-data-in-preview)) and re-creating the some of the fixture data.

Steps to removing the test data in preview environment: 
1. Merge this PR
2. Take the snapshot of notify-db in rds and also take database dump locally
3. ssh into ecs `api-web` service
4. Run `flask command purge-functional-test-data -u notify-tests-preview` to delete user data objects.
  * note: takes a while! multiple hours!

> Additional steps for Preview:
a. Remove https://www.notify.works/organisations/63b9557c-22ea-42ac-bcba-edaa50e3ae51/settings/ and https://www.notify.works/organisations/9b37efda-5855-4b18-bc37-6752f95994cb/settings/ Orgs

6. Run `flask command functional-test-fixtures` to create fixture data, this will create tmp environment file for functional-test service
7. `cd /tmp` and `cat functional_test_env.sh` to get the new environment file
8. `PASSWORD_STORE_DIR=$HOME/gds/notifications-credentials pass edit credentials/functional-tests/preview-functional ` and update the new environment file. This will auto commit and require manual PR process.
9. after merging, run `notifications-aws/scripts/upload-credentials/upload-credentials.sh preview` locally to push new environment file to SSM
10. Trigger the admin functional-test-preview pipeline/stage

In case of reverting the process:
1. Revert the PR for credentials
2. Rollback db from local dump, if local dump does not work then follow [these steps](https://docs.publishing.service.gov.uk/manual/howto-backup-and-restore-in-aws-rds.html) to restore from snapshot
3. Trigger admin functional-test-preview pipeline/stage

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -135,7 +135,12 @@ def purge_functional_test_data(user_email_prefix):
             services = dao_fetch_all_services_by_user(usr.id)
             for service in services:
                 print(f"Deleting service {service.id=} {service.name=} that {usr.id} belongs to")
-                delete_service_and_all_associated_db_objects(service)
+                # This also deletes the functional if the test users are associated
+                # It may require to create essential db object to run functional tests
+                if str(service.id) != current_app.config["NOTIFY_SERVICE_ID"]:
+                    delete_service_and_all_associated_db_objects(service)
+                else:
+                    print(f"Skipping service {service.id=} {service.name=}")
 
             services_created_by_this_user = dao_fetch_all_services_created_by_user(usr.id)
             for service in services_created_by_this_user:
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -42,14 +42,20 @@ from app.models import (
     Permission,
     Service,
     ServiceContactList,
+    ServiceDataRetention,
     ServiceEmailReplyTo,
     ServiceInboundApi,
+    ServiceJoinRequest,
     ServiceLetterContact,
     ServicePermission,
     ServiceSmsSender,
     Template,
+    TemplateFolder,
     TemplateHistory,
     TemplateRedacted,
+    UnsubscribeRequest,
+    UnsubscribeRequestHistory,
+    UnsubscribeRequestReport,
     User,
     VerifyCode,
 )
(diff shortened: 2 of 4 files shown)


app/commands.py
```diff
@@ -135,6 +135,8 @@ def purge_functional_test_data(user_email_prefix):
             services = dao_fetch_all_services_by_user(usr.id)
             for service in services:
                 print(f"Deleting service {service.id=} {service.name=} that {usr.id} belongs to")
+                # This also deletes the functional and notify services if the test users are associated
```
leohemsted, line 138: do we want to include an assertion here to make sure we don't delete the notify service just to be on the safe side?

app/commands.py
```diff
@@ -108,7 +108,7 @@ def wrapper(*args, **kwargs):
         return wrapper
 
 
-@notify_command()
+@notify_command("delete-functional-test-data")
```
leohemsted, line 111: we dont need to add this arg - flask click commands can be called with the function name (with hyphens) - so we can just run `flask command purge-functional-test-data -u notify-test-preview`

app/dao/services_dao.py
```diff
@@ -372,8 +386,18 @@ def _delete(query):
     _delete(NotificationHistory.query.filter_by(service=service))
     _delete(Notification.query.filter_by(service=service))
     _delete(Job.query.filter_by(service=service))
-    _delete(Template.query.filter_by(service=service))
-    _delete(TemplateHistory.query.filter_by(service_id=service.id))
+
+    templates = Template.query.filter_by(service=service)
+    for template in templates:
+        template.folder = None
+        _delete(TemplateHistory.query.filter_by(id=template.id))
```
leohemsted, line 393: would it be easier to just delete these template history objects in one row after the for loop? means we only issue one query rather than lots, and also just in case we end up with templates being deleted but the history still being around, then it prevents any weird issues.

so

```
    templates = Template.query.filter_by(service=service)
    for template in templates:
        template.folder = None
    _delete(templates)
    _delete(TemplateHistory.query.filter_by(service_id=service.id))
```

app/commands.py
spatel033: Nofify service should be created when we run fixtures 🤔, adding flag on the safe side.

app/commands.py
```diff
@@ -135,7 +135,12 @@ def purge_functional_test_data(user_email_prefix):
             services = dao_fetch_all_services_by_user(usr.id)
             for service in services:
                 print(f"Deleting service {service.id=} {service.name=} that {usr.id} belongs to")
-                delete_service_and_all_associated_db_objects(service)
+                # This also deletes the functional if the test users are associated
+                # It may require to create essential db object to run functional tests
+                if service.name != "GOV.UK Notify":
```
leohemsted, line 140: we can check `NOTIFY_SERVICE_ID` in the config - feels a bit more robust to check that rather than the service name (that UUID is hard coded from an alembic migration)

app/commands.py
```diff
@@ -135,7 +135,12 @@ def purge_functional_test_data(user_email_prefix):
             services = dao_fetch_all_services_by_user(usr.id)
             for service in services:
                 print(f"Deleting service {service.id=} {service.name=} that {usr.id} belongs to")
-                delete_service_and_all_associated_db_objects(service)
+                # This also deletes the functional if the test users are associated
+                # It may require to create essential db object to run functional tests
+                if service.name != current_app.config["NOTIFY_SERVICE_ID"]:
```
leohemsted, line 140: this won't work - needs to be service.id. probably cast to a string too

problem with working on untested code 😢

app/commands.py
spatel033: It won't create as it is part of the migration

leohemsted (approved):

PR #4347, opened 2025-01-17
Add new callback types

BlessedDev:
This PR adds a new callback type returned_letters to the service_callback_api table described in [this](https://trello.com/c/5SuwTyC1/175-create-a-new-type-of-callback-in-service-callback-api-table-returnedletter) card.
It also adds an inbound_sms callback_type in preparation for future work to merge the service_inbound_api into the service_callback_api table.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0486_letter_rates_feb_2025
+0847_insert_new_callback_types
--- a/migrations/versions/0847_insert_new_callback_types.py
+++ b/migrations/versions/0847_insert_new_callback_types.py
@@ -1,0 +1,19 @@
+"""
+Create Date: 2025-01-17 12:30:50.050539
+"""
+
+from alembic import op
+
+revision = '0847_insert_new_callback_types'
+down_revision = '0486_letter_rates_feb_2025'
+
+
+def upgrade():
+    insert_returned_letter_callback_type = "INSERT INTO service_callback_type VALUES ('returned_letters')"
+    insert_inbound_sms_callback_type = "INSERT INTO service_callback_type VALUES ('inbound_sms')"
+    op.execute(insert_returned_letter_callback_type)
+    op.execute(insert_inbound_sms_callback_type)
+
+
+def downgrade():
+    pass


app/constants.py
```diff
@@ -268,7 +268,16 @@ class OrganisationUserPermissionTypes(enum.StrEnum):
 # Service callbacks
 DELIVERY_STATUS_CALLBACK_TYPE = "delivery_status"
 COMPLAINT_CALLBACK_TYPE = "complaint"
-SERVICE_CALLBACK_TYPES = [DELIVERY_STATUS_CALLBACK_TYPE, COMPLAINT_CALLBACK_TYPE]
+RETURNED_LETTERS_CALLBACK = "returned_letters"
+SERVICE_CALLBACK_TYPES = [DELIVERY_STATUS_CALLBACK_TYPE, COMPLAINT_CALLBACK_TYPE, RETURNED_LETTERS_CALLBACK]
```
quis, line 272: Are we going to replace these with the `enum` at some point? Can we do that in this PR?

app/constants.py
```diff
@@ -268,7 +268,16 @@ class OrganisationUserPermissionTypes(enum.StrEnum):
...
+
+
+class ServiceCallbackTypes(enum.StrEnum):
+    delivery_status = "delivery_status"
+    complaint = "complaint"
+    returned_letters = "returned_letters"
+    inbound_sms = "inbound_sms"
```
quis, line 279: I feel like it’s confusing to add this now. It means anyone looking at the code needs to understand that this is not used. Why not wait until we are going to use it to add it?

migrations/versions/0847_insert_new_callback_types.py
```diff
@@ -0,0 +1,22 @@
...
+def upgrade():
+    insert_returned_letter_callback_type = "INSERT INTO service_callback_type VALUES ('returned_letters')"
+    insert_inbound_sms_callback_type = "INSERT INTO service_callback_type VALUES ('inbound_sms')"
+    op.execute(insert_returned_letter_callback_type)
+    op.execute(insert_inbound_sms_callback_type)
+
+
+def downgrade():
+    delete_returned_letter_callback_type = "DELETE FROM service_callback_type WHERE name = 'returned_letters'"
+    delete_inbound_sms_callback_type = "DELETE FROM service_callback_type WHERE name = 'inbound_sms'"
```
quis, line 20: This downgrade will start to fail as soon as we have any callbacks using these types. Should we also delete those first?

rparke: Related, is the downgrade test script currently working?

app/constants.py
BlessedDev: I see your point on this and will remove the enum from this PR.

app/constants.py
BlessedDev: That is the plan. However I will remove the code for the enum from this PR and deal with it in a separate PR.

migrations/versions/0847_insert_new_callback_types.py
BlessedDev: > This downgrade will start to fail as soon as we have any callbacks using these types. Should we also delete those first?

Good catch. We definitely don't want that, so I have updated the downgrade script to do nothing.

BlessedDev: > Related, is the downgrade test script currently working?

It works locally currently because no callbacks are using those types at the moment. However I have now updated the script.

migrations/.current-alembic-head
```diff
@@ -1 +1 @@
-0486_letter_rates_feb_2025
+0847_insert_new_callback_types
```
quis, line 1:
```suggestion
0487_insert_new_callback_types
```
Same in the `revision` definition and the filename.

BlessedDev:
@quis 
On reflection I am going to close this PR. I will focus solely on returned_letters and making the service_callback methods generic to accommodate returned letters.
The intended migration will then be part of another PR.

PR #4348, opened 2025-01-19
gunicorn: use `psutil` to emit post-request log output for long requests

risicle:
https://trello.com/c/M4vBaTPb/971-instrument-eventlet-to-get-to-the-bottom-of-0900-glitches

This is ugly and nasty and should be removed as soon as we've finished using it. we have to perform our own request timing because we can't reasonably access the flask-level annotations from the gunicorn side.

Performing this from gunicorn has the advantage of being executed *after* a response is sent, and so the 100ms delay this causes (or unexpected errors this results in for that matter) shouldn't prevent a user from getting their response.

### DIFF
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -12,3 +12,47 @@ worker_connections = 256
 statsd_host = "{}:8125".format(os.getenv("STATSD_HOST"))
 keepalive = 90
 timeout = int(os.getenv("HTTP_SERVE_TIMEOUT_SECONDS", 30))  # though has little effect with eventlet worker_class
+
+debug_post_threshold = os.getenv("NOTIFY_GUNICORN_DEBUG_POST_REQUEST_LOG_THRESHOLD_SECONDS", None)
+if debug_post_threshold:
+    debug_post_threshold_float = float(debug_post_threshold)
+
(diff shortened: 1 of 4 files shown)


gunicorn_config.py
```diff
@@ -12,3 +12,47 @@
...
+        elapsed = os.times().elapsed - req._pre_request_elapsed
+        if elapsed > debug_post_threshold_float:
+            import json
+            import time
+
+            import psutil
+
+            # consume this iterator to give cpu_percent calculation a "start time" to work with
+            list(psutil.process_iter(["cpu_percent"]))
+            time.sleep(0.1)  # period over which to calculate cpu_percent
```
leohemsted, line 41: to clarify if i've understood what is happening here - the intention here is to say "we have just had a slow request finish - now pause for 100ms and see what else is still going on on the box at the moment even after that request finished?"

which is why the example log you showed has 0% CPU for every PID?

leohemsted (approved):
🤞

gunicorn_config.py
risicle: Not all of the logs are all-0.0.

But yes, this is effectively what it's doing.

PR #4350, opened 2025-01-21
gunicorn: use local copy of `child_exit` instead of importing from `gds_metrics`

risicle:
https://trello.com/c/M4vBaTPb/971-instrument-eventlet-to-get-to-the-bottom-of-0900-glitches

This is something that contributes to the error splurge we get every time our applications start - fixing this rules out the possibility of the early-import-of-unpatched-io-routines being related to our issues.

(I have also explicitly tested killing a worker process unexpectedly, hence discovering we need to set `PROMETHEUS_MULTIPROC_DIR` more aggressively)

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+export PROMETHEUS_MULTIPROC_DIR="/tmp"
+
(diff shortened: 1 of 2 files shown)


leohemsted (approved):
lovely stuff

PR #4352, opened 2025-01-22
 gunicorn: add post-request profiling to `NOTIFY_GUNICORN_DEBUG_POST_REQUEST_LOG_THRESHOLD_SECONDS`

risicle:
This runs eventlet's adapted profiling code for the 100ms post-request wait to sample what this particular python process is doing - during a degradation there are usually other requests in flight being slowed down by _something_. So while this will not capture useful information about why the triggering request was slowed down, it should hopefully provide information about the cause of the following requests being held up.


leohemsted (approved):

risicle:
Merging this even though I've disabled debug output on prod - can revert later.

PR #4366, opened 2025-01-31
Add endpoints for returned letter callback

BlessedDev:
This PR adds the necessary endpoints for the returned letter callback feature as described in [this](https://trello.com/c/6qK45O5W/182-when-new-returned-letters-added-to-the-database-send-callbacks-to-services-that-have-callbacks-set-up-for-returned-letters) card.
It also implements the migration to add a `returned_letter` callback_type to the `service_callback_type` table.

The celery tasks will be addressed in another PR.

### DIFF
--- a/migrations/versions/0847_returned_letter_callback.py
+++ b/migrations/versions/0847_returned_letter_callback.py
@@ -4,8 +4,8 @@ Create Date: 2025-01-24 12:05:58.010551
 
 from alembic import op
 
-revision = '0847_returned_letter_callback'
-down_revision = '0486_letter_rates_feb_2025'
+revision = "0847_returned_letter_callback"
+down_revision = "0486_letter_rates_feb_2025"
 
 
 def upgrade():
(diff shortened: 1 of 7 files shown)


migrations/versions/0847_returned_letter_callback.py
```diff
@@ -0,0 +1,17 @@
...
+down_revision = '0486_letter_rates_feb_2025'
+
+
+def upgrade():
+    insert_returned_letter_callback_type = "INSERT INTO service_callback_type VALUES ('returned_letter')"
+    op.execute(insert_returned_letter_callback_type)
+
+
+def downgrade():
+    pass
```
CrystalPea, line 19: wouldn't we want to remove any returned letter callbacks, and then remove the callback type upon downgrade?

CrystalPea (commented):
Nice one! One small comment - let me know what you think 🙌🏼 

Might be worth squashing each test commit with corresponding add endpoint commit? But not a necessity 🙌🏼

migrations/versions/0847_returned_letter_callback.py
BlessedDev: @CrystalPea Good catch, I have updated the script.

CrystalPea (approved):

PR #4371, opened 2025-02-04
Update service free allowance when organisation type change

spatel033:
Update the free allowance when updating organisation type. If the service free allowance has been updated manually then it should keep the manual allowance, this requirement should be handled as part of [this story](https://trello.com/c/i67GtAIi/1113-figure-out-how-to-automatically-handle-high-volume-service-free-allowances). More details on [this card](https://trello.com/c/31ksdlS9/1147-if-you-update-organisation-type-it-doesnt-update-services-free-allowances-for-that-org).
This PR needs to be merged after https://github.com/alphagov/notifications-api/pull/4375

### DIFF
--- a/app/dao/organisation_dao.py
+++ b/app/dao/organisation_dao.py
@@ -1,5 +1,4 @@
 from flask import current_app
-from sqlalchemy import and_
 from sqlalchemy.sql.expression import func
 
 from app import db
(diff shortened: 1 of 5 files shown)


app/dao/organisation_dao.py
```diff
@@ -158,6 +160,11 @@ def _update_organisation_services(organisation, attribute, only_where_none=True)
         db.session.add(service)
 
 
+def _update_organisation_services_free_allowance(organisation):
+    for service in organisation.services:
+        set_default_free_allowance_for_service(service, year_start=None)
```
klssmith, line 165: This function only keeps the manual free allowances if they're 0, but not if they're any other value - it sounds like this isn't what we want to happen?

spatel033: Tested the changes with https://github.com/alphagov/notifications-api/pull/4375. `set_default_free_allowance_for_service` has been refactored so allowance would be updated for any value.

klssmith (approved):

PR #4372, opened 2025-02-04
Stop using `uv` to install packages when running `make bootstrap`

klssmith:
This is currently giving errors, so this change stops us using `uv` until we can investigate further.


quis (approved):

klssmith:
Closing in favour of pinning UV

PR #4375, opened 2025-02-05
Automate setting the free allowance for high volume services

klssmith:
See individual commits for details.

Our current code to set the free allowance for a service either gives the service the default its organisation type or keeps the free allowance as zero if it was zero for previous years. This has a couple of issues:
- services with a different custom allowance didn't have this persisted to future years
- services which a free allowance of 0 due to being a high volume sender never had this reset if they stopped sending so many messages

Two extra columns have been added to the annual billing table, `high_volume_service_last_year` and `has_custom_allowance`. These let us automatically set the free allowance to 0 for high volume senders and to persist custom free allowances.

### DIFF
--- a/app/billing/rest.py
+++ b/app/billing/rest.py
@@ -82,9 +82,7 @@ def create_or_update_free_sms_fragment_limit(service_id):
     return jsonify(form), 201
 
 
-def update_free_sms_fragment_limit_data(service_id, free_sms_fragment_limit, financial_year_start=None):
-    # TODO: `financial_year_start` parameter can be removed, but has been kept temporarily so nothing breaks
-    # during deployment
+def update_free_sms_fragment_limit_data(service_id, free_sms_fragment_limit):
     current_year = get_current_financial_year_start_year()
 
     dao_create_or_update_annual_billing_for_year(service_id, free_sms_fragment_limit, current_year)
--- a/app/billing/rest.py
+++ b/app/billing/rest.py
@@ -7,6 +7,7 @@ from app.billing.billing_schemas import (
 )
 from app.dao.annual_billing_dao import (
     dao_create_or_update_annual_billing_for_year,
+    dao_get_default_annual_allowance_for_service,
     dao_get_free_sms_fragment_limit_for_year,
     set_default_free_allowance_for_service,
 )
@@ -15,6 +16,7 @@ from app.dao.fact_billing_dao import (
     fetch_usage_for_service_annual,
     fetch_usage_for_service_by_month,
 )
+from app.dao.services_dao import dao_fetch_service_by_id
 from app.errors import register_errors
 from app.models import Service
 from app.schema_validation import validate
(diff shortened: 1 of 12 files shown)


app/dao/annual_billing_dao.py
```diff
@@ -44,38 +85,29 @@ def set_default_free_allowance_for_service(service, year_start=None):
...
+
+        if last_years_allowance and last_years_allowance.has_custom_allowance:
+            # carry over the allowance from last year
+            free_sms_fragment_allowance = last_years_allowance.free_sms_fragment_limit
+            has_custom_allowance = True
+        else:
+            # get the default for the org
+            default_free_sms_fragment_allowance = dao_get_default_annual_allowance_for_service(service, year_start)
+            free_sms_fragment_allowance = default_free_sms_fragment_allowance.allowance
+            has_custom_allowance = False
```
leohemsted, line 110: i think this if/else is the crux of the complexity here and i spent a while thinking how to frame it, haven't really come up with anything great.... maybe this is marginally more readable by separating the `has_custom_allowance` concern from the "what actually is the free allowance 

```python
last_year = AnnualBilling.query(....)
if  last_year:
    has_custom_allowance = last_year.has_custom_allowance

if has_custom_allowance:
    next_allowance = last_year.free_sms_fragment_limit:
else:
    default = dao_get_default_annual_allowance_for_service(service, year_start)
    next_allowance = default.allowance
```

maybe this makes no difference at all to legibility? 

(I do like being explicit about renaming the variable to `next_allowance` or `next_free_sms_fragment_allowance` or similar to make sure we dont confuse it with any existing values, given that we are thinking about those at the same time)

leohemsted (approved):
This looks like good stuff, the commits are really sensible and well laid out too.

I've got a couple of comments/thoughts/ideas but they shouldn't block merging this and I'm not sure if they're fundamentally better suggestions than what you've already implemented, or if they're just different

app/billing/rest.py
```diff
@@ -82,9 +84,25 @@ def create_or_update_free_sms_fragment_limit(service_id):
...
+    org type, we set `has_custom_allowance` to False, otherwise it is True.
+    """
     current_year = get_current_financial_year_start_year()
+    service = dao_fetch_service_by_id(service_id, with_users=False)
+    default_free_allowance = dao_get_default_annual_allowance_for_service(service, current_year)
+
+    if default_free_allowance == free_sms_fragment_limit:
+        has_custom_allowance = False
+    else:
+        has_custom_allowance = True
```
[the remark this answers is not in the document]
klssmith: I've made this change and rebased it in to the relevant commit.

app/dao/annual_billing_dao.py
klssmith: The idea above doesn't quite work as is because `has_custom_allowance` only gets defined if there was an allowance last year. By adjusting it to work, it then starts to look not too different from the version I already have. So instead I've just renamed the variable to `next_free_sms_fragment_allowance` (added in a new commit).

leohemsted (approved):

PR #4378, opened 2025-02-11
Update service_callback_api field in ServiceSchema

BlessedDev:
service_callback_api now also includes returned_letter callback type, so this field is no longer a list of ids filtered on the delivery_status callback type.
The choice of representing the data in dicts, with callback_id and callback_type as keys, is to aid filtering in the front end as to what type of callback is being considered ie
delivery_status or returned_letter.

Needs to wait until this PR is deployed: https://github.com/alphagov/notifications-admin/pull/5382


quis (commented):
I think we need to re-think the approach a bit here. We can’t just change the API response and rely on the admin PR going out at exactly the same time – this needs to work for old and new versions of the admin code.

CrystalPea (approved):

PR #4380, opened 2025-02-12
Delete test letter files from S3

klssmith:
When deleting the test letters from the Notifications table as part of the `delete-notifications-for-service-and-type` task we weren't deleting the test files from S3. We do delete the files in S3 for real letters, so not deleting the files for test letters appears to have been an oversight.

This copies the approach we take for real letters and their files, and deletes the test letter files in batches.

https://trello.com/c/iHHjm2fs/1149-investigate-nightly-job-for-deleting-test-letter-objects-from-s3-after-the-retention-period-expires

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -438,16 +438,31 @@ def move_notifications_to_notification_history(
     )
 
 
-def delete_test_notifications(notification_type, service_id, timestamp_to_delete_backwards_from):
-    # Deleting test Notifications, test notifications are not persisted to NotificationHistory
-    Notification.query.filter(
-        Notification.notification_type == notification_type,
-        Notification.service_id == service_id,
-        Notification.created_at < timestamp_to_delete_backwards_from,
-        Notification.key_type == KEY_TYPE_TEST,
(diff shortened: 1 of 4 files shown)


app/dao/notifications_dao.py
```diff
@@ -438,17 +438,32 @@ def move_notifications_to_notification_history(
...
-        Notification.service_id == service_id,
-        Notification.created_at < timestamp_to_delete_backwards_from,
-        Notification.key_type == KEY_TYPE_TEST,
-    ).delete(synchronize_session=False)
+    subquery = (
+        db.session.query(Notification.id)
+        .filter(
+            Notification.notification_type == notification_type,
+            Notification.service_id == service_id,
+            Notification.created_at < timestamp_to_delete_backwards_from,
```
risicle, line 454: Could we have some sort of limit of earliest notification to consider? Otherwise we're probably doing the whole sequential-scan-of-the-table-back-to-the-beginning again...

risicle: Actually come to think, it might not be such a big deal given it's restricted to a single service id

spatel033 (approved):

PR #4383, opened 2025-02-17
feat: Create Users List Endpoint for Platform Admins

joybytes:
## Summary:
This PR introduces a new API endpoint (`/users-list`) for platform admins to fetch a filtered list of users based on various criteria. The filtered data is used to generate a CSV export.

Users can be filtered by:
- Creation date (created_start, created_end)
- Last login date (logged_in_start, logged_in_end)
- Research opt-in status (take_part_in_research)
- User state (defaults to "active")

### Related Ticket:
[Add getting list  users and emails for yearly survey to platform admin](https://trello.com/c/qeSQJ5XB/1138-add-getting-list-users-and-emails-for-yearly-survey-to-platform-admin)

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -77,6 +77,7 @@ from app.history_meta import Versioned
 from app.utils import (
     DATETIME_FORMAT,
     DATETIME_FORMAT_NO_TIMEZONE,
+    dict_filter,
     get_dt_string_or_none,
     get_london_midnight_in_utc,
     get_uuid_string_or_none,
@@ -179,7 +180,13 @@ class User(db.Model):
 
         return retval
 
-    def serialize(self):
+    def serialize(self, service_filter_keys=None):
+        if service_filter_keys is None:
+            services_data = [x.id for x in self.services if x.active]
+        else:
+            service_filter_keys = list(set(service_filter_keys) | {"id"})
+            services_data = [dict_filter(x, service_filter_keys) for x in self.services if x.active]
+
         return {
             "id": self.id,
             "name": self.name,
@@ -196,7 +203,7 @@ class User(db.Model):
             "permissions": self.get_permissions(),
             "organisation_permissions": self.get_organisation_permissions(),
             "platform_admin": self.platform_admin,
-            "services": [x.id for x in self.services if x.active],
+            "services": services_data,
             "can_use_webauthn": self.can_use_webauthn,
             "state": self.state,
             "take_part_in_research": self.take_part_in_research,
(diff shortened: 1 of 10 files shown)


app/models.py
```diff
@@ -196,7 +225,11 @@ def serialize(self):
             "permissions": self.get_permissions(),
             "organisation_permissions": self.get_organisation_permissions(),
             "platform_admin": self.platform_admin,
-            "services": [x.id for x in self.services if x.active],
+            "services": (
+                [{"id": x.id, "name": x.name} for x in self.services if x.active]
```
spatel033, line 229: We could do something like this to allow to extend more parameters in future.
```
def serialize(self, service_filter_keys=["id"]):
dict_filter = lambda x, y: dict([ (i, x[i]) for i in x if i in y ]) #utils function
```
```
"services": [dict_filter(x, service_filter_keys)for x in self.services if x.active]
```

tests/app/conftest.py
```diff
@@ -168,7 +168,7 @@ def sample_service_with_email_branding(sample_service):
 
 
 @pytest.fixture(scope="function", name="sample_service_full_permissions")
-def _sample_service_full_permissions(notify_db_session):
+def sample_service_full_permissions(notify_db_session):
```
spatel033, line 171: `_` is intentional to stop using this function directly. Is there any specific reason to remove `_`.

klssmith:
One for future perhaps, but it would be good to change your editor settings so that it doesn't reformat all the files or so that the formatting rules match the standard ones we use (e.g. line length of 120). It just makes it hard to see what has actually been changed in the PR and searching the git history is made harder if there are lots of formatting changes included

app/models.py
joybytes: good idea, take a look on the implementation now and let me know if you have any suggestions

tests/app/conftest.py
joybytes: thanks, that was by mistake!

app/models.py
```diff
@@ -179,7 +179,17 @@ def get_organisation_permissions(self) -> dict[str, list[str]]:
 
         return retval
 
-    def serialize(self):
+    @staticmethod
+    def dict_filter(data_obj, keys):
```
spatel033, line 183: Looks good! It would be better if we move `dict_filter` to `app.utils` so we can use for other models in future.

joybytes: true!
done

spatel033 (approved):

PR #4385, opened 2025-02-20
Add get request report

spatel033:
PR is part of making csv download more reliable. More details on [this card](https://trello.com/c/QCQAOt9O/1169-create-report-requests-table-and-get-endpoint).
- Added new table `report_requests`
- Endpoint to GET report request by id


spatel033:
lint migrations failing because of timezone, open for any suggestion

app/constants.py
```diff
@@ -363,3 +363,27 @@ class CacheKeys:
     SERVICE_JOIN_REQUEST_PENDING,
     SERVICE_JOIN_REQUEST_CANCELLED,
 ]
+
+# Report request report types
+REPORT_REQUEST_NOTIFICATIONS_STATUS_CSV = "notifications_status_csv"
```
joybytes, line 368: I think maybe the report type for notifications can be just "notifications" or something like "service_notifications" just to be more specific
I think the `notifications_status_csv` seems a little weird - sorry I know this was on the document.

app/models.py
```diff
@@ -2962,3 +2965,49 @@ def serialize(self) -> SerializedServiceJoinRequest:
...
+    report_type = db.Column(
+        db.Enum(*REPORT_REQUEST_REPORT_TYPES, name="report_request_report_type"),
+        nullable=False,
+    )
+    status = db.Column(
+        db.Enum(*REPORT_REQUEST_STATUS_TYPES, name="report_request_status_type"),
+        nullable=False,
+        default=REPORT_REQUEST_REQUESTED,
+    )
+    parameter = db.Column(JSONB, nullable=False, default={})
```
joybytes, line 2997: could we add a constraint between report_type and the expected fields in parameter?
For example, for report_type = notifications, we would expect the following fields to be present

```
{
  "notification_type": "sms",
  "notification_status": "delivered"
}

```
notification_type can be 'all', 'letters', 'sms', 'emails'

and 
notification_status can be 'all', 'sending', 'delivered', 'failed'

app/models.py
```diff
@@ -2962,3 +2965,49 @@ def serialize(self) -> SerializedServiceJoinRequest:
...
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    updated_at = db.Column(
+        db.DateTime, nullable=False, default=datetime.datetime.utcnow, onupdate=datetime.datetime.utcnow
```
klssmith, line 3000: `updated_at` should be nullable, because it won't be set when the row is first created.

joybytes (approved):

app/dao/report_requests_dao.py
```diff
@@ -0,0 +1,14 @@
...
+from app.dao.dao_utils import autocommit
+from app.models import ReportRequest
+
+
+@autocommit
+def dao_create_report_request(report_request: ReportRequest):
+    db.session.add(report_request)
+
+
+def dao_get_report_request_by_id(service_id: UUID, report_id: UUID) -> ReportRequest | None:
```
klssmith, line 13: Optional, but this won't return `None` if no result is found it will raise an error so `None` could be removed from the type hint just to avoid confusion. Raising an error is fine though and works well with the frontend code.

klssmith (commented):
I left a couple of small optional things, but think this looks good

app/constants.py
```diff
@@ -363,3 +363,27 @@ class CacheKeys:
...
+
+# Report request report types
+REPORT_REQUEST_NOTIFICATIONS = "notifications"
+
+REPORT_REQUEST_REPORT_TYPES = [
+    REPORT_REQUEST_NOTIFICATIONS,
+]
+
+# Report request report statuses
+REPORT_REQUEST_REQUESTED = "requested"
```
joybytes, line 375: maybe we can remove requested

cc: @klssmith

klssmith: It would make things more consistent with `jobs` to have two statuses to cover the point from creating a row to starting to process the data - for jobs we have `pending` then `in progress`

joybytes (approved):

PR #4388, opened 2025-02-25
Shatter letter delivery tasks

quis:
# Before

When delivering letters to DVLA we are bottlenecked by the speed at which we can create SQS messages. This is a problem because we have one task which delivers all the letters, and it creates one SQS message at a time. This task runs for a long time, and we can’t deploy while it’s running.

image

# After 

This change makes it so that one new task (and SQS message) is created for each 100 letters. Each of these new tasks – which can be picked up asynchronously – then create individual tasks for each of its 100 letters.

The overall time taken will probably be about the same, but the maximum time taken by any one task should be greatly reduced.

image

# Estimated increase in performance

  | Number of  letters<sup>1</sup>  | Database time | Time talking to SQS | Total task time
-- | -- | -- | -- | -- 
**Before** | 166,936  | 11 seconds<sup>2</sup> | 29 minutes 15 seconds | 29 minutes 26 seconds<sup>1</sup>
**After** | 166,936  |  11 seconds | 17 seconds | 28 seconds 💥 

1. Taken from last night’s print run
2. Estimated by running the same query manually against production and scaling up based on number of results

***

This is a similar approach taken by `shatter_jobs` in https://github.com/alphagov/notifications-api/pull/4284

### DIFF
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -235,10 +235,7 @@ def send_dvla_letters_via_api(print_run_deadline_local, batch_size=100):
         (row.id for row in dao_get_letters_to_be_printed(print_run_deadline_local)),
         batch_size,
     ):
-        current_app.logger.info("triggered tasks for %i letters", batch_size)
         shatter_deliver_letter_tasks.apply_async([batch], queue=QueueNames.PERIODIC)
-
-    current_app.logger.info("send-dvla-letters-for-day-via-api - finished queuing")
 
 
 @notify_celery.task(name="shatter-deliver-letters-tasks")
--- a/app/celery/letters_pdf_tasks.py
+++ b/app/celery/letters_pdf_tasks.py
@@ -49,6 +49,7 @@ from app.letters.utils import (
     move_scan_to_invalid_pdf_bucket,
 )
 from app.models import Service
+from app.utils import batched
 
 
 @notify_celery.task(bind=True, name="get-pdf-for-templated-letter", max_retries=15, default_retry_delay=300)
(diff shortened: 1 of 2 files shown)


app/celery/letters_pdf_tasks.py
```diff
@@ -228,16 +229,27 @@ def send_letters_volume_email_to_dvla(letters_volumes, date):
         send_notification_to_queue(saved_notification, queue=QueueNames.NOTIFY)
 
 
-def send_dvla_letters_via_api(print_run_deadline_local):
+def send_dvla_letters_via_api(
+    print_run_deadline_local,
+    batch_size=32,  # can this be larger because each task only takes 1 arg?
```
leohemsted, line 234: some quick testing, i could send a task invocation with 180 args - 182 failed

100 could be reasonable. worth a comment saying that this is tied to the number and size of arguments to ensure we go back to this in case we add other args in future

quis: Also worth considering if it’s good/bad/irrelevant for the batch size to be a multiple of the query limit here:
https://github.com/alphagov/notifications-api/blob/46ac04b8547a3a95941d84a7d87908caca8d6027/app/dao/notifications_dao.py#L789

leohemsted: my head leans towards irrelevant - if we go to the database in-between sending shatter tasks or half-way through doesn't really matter in my head cos the extra data we'll hold in memory (100 UUIDs) is so piddly

app/celery/letters_pdf_tasks.py
```diff
@@ -228,16 +229,27 @@ def send_letters_volume_email_to_dvla(letters_volumes, date):
...
     current_app.logger.info("send-dvla-letters-for-day-via-api - starting queuing")
-    for i, row in enumerate(dao_get_letters_to_be_printed(print_run_deadline_local)):
-        if i % 10000 == 0:
-            current_app.logger.info("triggered tasks for %i letters", i)
-        deliver_letter.apply_async(kwargs={"notification_id": row.id}, queue=QueueNames.SEND_LETTER)
+    for batch in batched(
+        (row.id for row in dao_get_letters_to_be_printed(print_run_deadline_local)),
+        batch_size,
+    ):
+        current_app.logger.info("triggered tasks for %i letters", batch_size)
```
leohemsted, line 238: this'll log a lot (100x more than it logs atm!). i think we don't really need this as we'll still have the "send-dvla-letters-for-day-via-api - finished queuing" log below (tbh we can remove that log line too and just rely on the normal "Celery task x took y seconds" log line) as the collate task doesn't do anything after triggering all these, and just finishes
```suggestion
```

quis: f782fe2

quis:
Date | Number of letters | Time taken
---|---|---
25th February | 166,936 | 29 minutes 27 Seconds
26th February | 167,764 | 23 seconds

78&times; faster

PR #4393, opened 2025-03-04
Update letter to "sending" on `DvlaDuplicatePrintRequestException`

klssmith:
If we receive a `DvlaDuplicatePrintRequestException` error when sending a letter to the DVLA, it means that they already have the letter in their system. We were logging a warning, leaving the letter in "created" state and then trying to send the letter again the following evening, repeating this process until we got a delivery receipt for the letter.

There's a risk that DVLA could delete the letter data from their system at some point, which would then result in the letter being sent twice. To avoid this happening we can update the letter status to "sending", since DVLA clearly have a reference to it.


app/celery/provider_tasks.py
```diff
@@ -145,6 +145,7 @@ def deliver_letter(self, notification_id):
             ) from e
     except Exception as e:
         if isinstance(e, DvlaDuplicatePrintRequestException):
```
leohemsted, line 147: can make this even tidier by moving it to a `except DvlaDuplicatePrintRequestException` handler

leohemsted (approved):

PR #4398, opened 2025-03-10
Bump notify-utils to 97.0.3

spatel033:
# CHANGELOG

## 97.0.3

* Bump minimum jinja2 version to 3.1.6

## 97.0.2

* Fix external link redirect

## 97.0.1

* Fix QR code URL with '&' encoded as '&amp;'

## 97.0.0

* for bilingual letters preview, only show barcodes on the first page (patch).
* rename `include_notify_tag` argument to `includes_first_page` (major).

## 96.0.0

* BREAKING CHANGE: the `gunicorn_defaults` module has been moved to `gunicorn.defaults` to make space for gunicorn-related utils that don't have the restricted-import constraints of the gunicorn defaults. Imports of `notifications_utils.gunicorn_defaults` should be changed to `notifications_utils.gunicorn.defaults`.
* Added `ContextRecyclingEventletWorker` custom gunicorn worker class.

## 95.2.0

* Implement `InsensitiveSet.__contains__`

## 95.1.2

* Fix to the number of arguments the `on_retry` of the `NotifyTask` class takes.

## 95.1.1

* Add `RUF100` rule to linter config (checks for inapplicable uses of `# noqa`)

## 95.1.0

* Adds log message and statsd metric for retried celery tasks

## 95.0.0

* Reverts 92.0.0 to restore new validation code

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-# This file was automatically copied from notifications-utils@95.1.2
+# This file was automatically copied from notifications-utils@97.0.3
 
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -1,6 +1,6 @@
 import os
 
-from notifications_utils.gunicorn_defaults import set_gunicorn_defaults
+from notifications_utils.gunicorn.defaults import set_gunicorn_defaults
 
 set_gunicorn_defaults(globals())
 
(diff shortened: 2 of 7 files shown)


klssmith:
When this is unblocked, you'll need to check the changelog for breaking changes. For example, if we accidentally don't make the changes required in 96.0.0 the app won't start (which is something that happened when we tried to deploy template preview without adjusting for that).

kr8n3r (dismissed):

kr8n3r (approved):

PR #4404, opened 2025-03-19
shatter_job_rows: recursively subdivide batches if SQS message too big

risicle:
See https://trello.com/c/umdHKBfA/1053-dynamic-shatter-job-rows-batch-size

This introduces a wrapper function, `_shatter_job_rows_with_subdivision` which will catch an exception caused by an oversized SQS message and split the batch in half before trying again, recursing if it too throws the same error.

Uses this in both `process_job` and `process_incomplete_jobs`. In the latter case, catch the exception raised when a batch can't be subdivided any further, log an exception and continue with the next incomplete job. this is unusual in that it's the only exception we treat like this - perhaps we should be catching more exceptions like this and proceeding with the next job(s) to avoid a single bad job stopping all incomplete jobs being re-processed, but in this specific case it became clear to me that allowing this exception to stop everything would be a silly behaviour.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -44,7 +44,7 @@ from app.service.utils import service_allowed_to_send_to
 from app.utils import batched
 from app.v2.errors import TooManyRequestsError
 
-DEFAULT_SHATTER_JOB_ROWS_BATCH_SIZE = 8
+DEFAULT_SHATTER_JOB_ROWS_BATCH_SIZE = 32
 
 
 class UnprocessableJobRow(Exception):
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,6 +1,7 @@
 from collections.abc import Sequence
 from datetime import datetime
 
+from botocore.exceptions import ClientError as BotoClientError
 from flask import current_app
 from notifications_utils.insensitive_dict import InsensitiveDict
 from notifications_utils.recipient_validation.postal_address import PostalAddress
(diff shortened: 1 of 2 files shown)


app/celery/tasks.py
```diff
@@ -83,15 +88,42 @@ def process_job(job_id, sender_id=None, shatter_batch_size=DEFAULT_SHATTER_JOB_R
...
+                args_kwargs_seq,
             ),
             queue=QueueNames.JOBS,
         )
+        if not top_level:
+            log_context = {"shatter_batch_size": len(args_kwargs_seq)}
+            current_app.logger.info(
+                "Job shatter batch sent with reduced size of %(shatter_batch_size)s", log_context, extra=log_context
+            )
+    except BotoClientError as e:
```
quis, line 110: Very optional because I’m not sure if this is actually more readable or not, but consider:
```suggestion
    except BotoClientError as e:
        ...
    else:
        if not top_level:
            log_context = {"shatter_batch_size": len(args_kwargs_seq)}
            current_app.logger.info(
                "Job shatter batch sent with reduced size of %(shatter_batch_size)s", log_context, extra=log_context
            )
```

app/celery/tasks.py
```diff
@@ -46,6 +47,10 @@
 DEFAULT_SHATTER_JOB_ROWS_BATCH_SIZE = 8
 
 
+class JobRowTooBig(Exception):
```
quis, line 50: I’d consider renaming this to something more generic:
```suggestion
class CantProcessJobRow(Exception):
```

We are sniffing `InvalidParameterValue` which is quite generic. If we messed up something else we’d end up with an exception like:
```
JobRowTooBig: The specified queue does not exist or you do not have access to it.
```

This might make send whoever is on support down a bit of a rabbit hole trying to figure out which row is too big (when it’s nothing to do with the row size).

quis (approved):
Tests are really thorough, nice work

Left a couple of comments which you can take or leave depending on whether you agree with them

app/celery/tasks.py
risicle: Yeah maybe

quis (approved):

PR #4406, opened 2025-03-26
Add notifications request report

spatel033:
Add logic for processing notification request report. Get the request and notifications, convert them in csv format data and upload to S3 via multipart upload. More information on the [card](https://trello.com/c/sRiFZcNB/1197-function-report-data-processing).


app/report_requests/process_notifications_report.py
```diff
@@ -0,0 +1,127 @@
...
+                part_number=page,
+                bucket_name=s3_bucket,
+                filename=filename,
+                upload_id=upload_id,
+                data_bytes=data_bytes,
+            )
+            parts.append({"PartNumber": page, "ETag": upload_response["ETag"]})
+            page += 1
+
+            if len(serialized_notifications) < page_size:
```
klssmith, line 77: If I make `page_size` less than the number of notifications available then I get an error. I haven't dug into exactly what is failing, but we'll need to check that everything works as expected when the number of notifications is either higher than or equal to `page_size`.

Mocking `REPORT_REQUEST_NOTIFICATIONS_CSV_BATCH_SIZE` to be a low number in the tests will help with this - check out how we use the `set_config` here for example: https://github.com/alphagov/notifications-api/blob/18bd7eb06eb974627ee9f8c5c0f26e47c613779c/tests/app/dao/test_inbound_sms_dao.py#L304

app/report_requests/process_notifications_report.py
```diff
@@ -0,0 +1,127 @@
...
+from app.dao.notifications_dao import get_notifications_for_service
+from app.dao.report_requests_dao import dao_get_report_request_by_id
+from app.dao.service_data_retention_dao import fetch_service_data_retention_by_notification_type
+
+
+def process_report_request(service_id, report_request_id):
+    report_request = dao_get_report_request_by_id(service_id, report_request_id)
+    notification_type = report_request.parameter["notification_type"]
+    notification_status = report_request.parameter["notification_status"]
+    service_retention = fetch_service_data_retention_by_notification_type(service_id, notification_type)
```
klssmith, line 21: If `notification_type` is `all`, then this line gives an error.

app/constants.py
```diff
@@ -143,6 +181,8 @@
 EMAIL_TYPE = "email"
 LETTER_TYPE = "letter"
 BROADCAST_TYPE = "broadcast"
+ALL_TYPE = "all"
+ALL_NOTIFICATION_TYPES = [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]
```
klssmith, line 185: The `NOTIFICATION_TYPES` constant 2 lines down covers all three notification types so can be used instead of adding a new constant

app/report_requests/process_notifications_report.py
```diff
@@ -0,0 +1,161 @@
...
+        "Job",
+        "Status",
+        "Time",
+        "API key name",
+    ]
+    csv_writer.writerow(headers)
+
+    try:
+        for substituted_type in substituted_types:
+            service_retention = fetch_service_data_retention_by_notification_type(service_id, substituted_type)
```
klssmith, line 55: `fetch_service_data_retention_by_notification_type` returns `None` if the service doesn't have custom data retention for the specified notification type. So for that case, the default value of 7 days should be used instead of `service_retention.days_of_retention`.

app/report_requests/process_notifications_report.py
```diff
@@ -0,0 +1,154 @@
...
+    s3_multipart_upload_part,
+)
+
+from app.constants import NOTIFICATION_REPORT_REQUEST_MAPPING
+from app.dao.notifications_dao import get_notifications_for_service
+from app.dao.report_requests_dao import dao_get_report_request_by_id
+from app.dao.service_data_retention_dao import fetch_service_data_retention_by_notification_type
+
+
+def process_report_request(service_id, report_request_id):
```
joybytes, line 19: should this have the celery.task decorator on it so I am able to do `apply_async`?

spatel033: `process_report_request` only contains the logic to get the request and generate csv. Celery task would be part of another card to find the next request, process_report_request, update the status of the request, handle any error and retry.

app/report_requests/process_notifications_report.py
```diff
@@ -0,0 +1,154 @@
...
+                page=page,
+                page_size=page_size,
+                limit_days=limit_days,
+            )
+
+            csv_data = convert_notifications_to_csv(serialized_notifications)
+            csv_writer.writerows(csv_data)
+
+            # upload when the minimum upload part size exceeds
+            if csv_buffer.tell() >= S3_MULTIPART_UPLOAD_MIN_PART_SIZE:
```
joybytes, line 69: Noticed we're using csv_buffer.tell() to check size, but since tell() gives character count (not bytes), it might not match the actual size we're comparing it to:

```
S3_MULTIPART_UPLOAD_MIN_PART_SIZE = 5 * 1024 * 1024  # 5MB minimum multi part upload size

```

To make sure we're checking the real byte size, it might be better to do:

```
data_bytes = csv_buffer.getvalue().encode("utf-8")
if len(data_bytes) >= S3_MULTIPART_UPLOAD_MIN_PART_SIZE:
```

spatel033: Good spot!

joybytes (approved):

app/report_requests/process_notifications_report.py
```diff
@@ -0,0 +1,159 @@
...
+                service_id=service_id,
+                template_type=notification_type,
+                notification_status=notification_status,
+                page=page,
+                page_size=page_size,
+                limit_days=limit_days,
+            )
+
+            csv_data = convert_notifications_to_csv(serialized_notifications)
+            csv_writer.writerows(csv_data)
```
leohemsted, line 66: we need to re-use the buffer because we're using the same `csvwriter`. but since we're not using a more advanced writer (ie: we don't need to keep track of headers), we could just create a new csv writer each time, and with that create a new buffer each time so it's simpler to manage and we would never need to `seek` at all

app/report_requests/process_notifications_report.py
```diff
@@ -0,0 +1,170 @@
...
+            self._upload_csv_part_if_needed()
+            page += 1
+
+        # Upload any remaining data
+        self._upload_remaining_data()
+
+    def _fetch_serialized_notifications(self, page: int, limit_days: int) -> list[dict[str, Any]]:
+        statuses = NOTIFICATION_REPORT_REQUEST_MAPPING[self.notification_status]
+
+        notifications = get_notifications_for_service(
```
leohemsted, line 89: this needs to `include_jobs=True` (and a relevant test)

tests/app/report_requests/test_process_notifications_report.py
```diff
@@ -1,157 +1,616 @@
...
+        assert line_count == 3  # two rows + header
+
+        reader = csv.DictReader(io.StringIO(content))
+
+        headers = reader.fieldnames
+        assert ",".join(headers) == expected_headers
+    
+        references = [row.get("Reference") for row in reader]
+        assert references[0] == "job-reference"
+        assert references[1] == "no-job-reference"
```
leohemsted, line 616: can we assert the contents of the "Job" column here too?

leohemsted (approved):

PR #4407, opened 2025-03-26
Add new letter rates valid from April 2025

klssmith:
First and second class letter rates are increasing from 1st April 2025, so this adds the new rates to the database.

[Trello card](https://trello.com/c/Mk5arZ2z/1232-letter-price-changes)


klssmith:
This is failing the Squawk linting check because Squawk does not seem to recognise parameterized queries, but this migration is one that has been run multiple times before with different values

spatel033 (approved):

PR #4413, opened 2025-04-01
Add create report-request POST endpoint

spatel033:
POST report-request endpoint deals with different types of reports. For now it would only handle the necessary logic for 'notifications_report'. More details on [this card](https://trello.com/c/sgeh0FH8/1183-post-endpoint-for-report-request).
At this stage we are not looking to implement the queue function (that’s part of a task) or a lock mechanism.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -21,6 +21,8 @@ from app.constants import (
     MOBILE_TYPE,
     NOTIFICATION_CANCELLED,
     NOTIFICATION_TYPES,
+    REPORT_REQUEST_NOTIFICATIONS,
+    REPORT_REQUEST_PENDING,
     SERVICE_JOIN_REQUEST_APPROVED,
 )
 from app.dao import fact_billing_dao, fact_notification_status_dao, notifications_dao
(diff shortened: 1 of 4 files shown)


app/service/rest.py
```diff
@@ -1419,3 +1423,31 @@ def _fetch_returned_letter_data(service_id, report_date):
...
+    try:
+        user_id = req_json.get("user_id")
+        report_request = ReportRequest(
+            user_id=user_id,
+            service_id=service_id,
+            report_type=report_type,
+            status=REPORT_REQUEST_PENDING,
+            parameter=parameter,
+        )
+        dao_create_report_request(report_request)
```
joybytes, line 1449: This call for `dao_create_request` is a placeholder because we need to handle the lock and call the queueing process - I think you already know this but putting here just in case you had other thoughts.

joybytes (approved):
looks, thanks for the work!

app/service/rest.py
spatel033: Yes there will be lock mechanism left which will be implemented as part of different card.

PR #4414, opened 2025-04-04
Add predefined queue names so celery will not try to create queues

corlettb:
https://trello.com/c/joBsPfMJ/955-move-sqs-queue-creation-into-terraform-and-clean-up-old-queues

What
----

Add predefined queue names so celery will not try to create queues

Why
----

We want all queues to be created and managed through terraform

Note
----

This change should not be merged before [this pr](https://github.com/alphagov/notifications-aws/pull/2564).

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -54,6 +54,20 @@ class QueueNames:
             QueueNames.REPORT_REQUESTS_NOTIFICATIONS,
         ]
 
+    def external_queues():
+        return [
+            QueueNames.ANTIVIRUS,
+            QueueNames.SANITISE_LETTERS,
+            QueueNames.GOVUK_ALERTS,
+        ]
+
+    @staticmethod
+    def predefined_queues(prefix, aws_region, aws_account_id):
+        return {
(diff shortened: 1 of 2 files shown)


app/config.py
```diff
@@ -54,6 +54,20 @@ def all_queues():
             QueueNames.REPORT_REQUESTS_NOTIFICATIONS,
         ]
 
+    def non_worker_queues():
```
risicle, line 57: I think a better name for this would be either `non_api_queues` or `external_queues`, because we still call the ecs tasks that consume them "workers".

corlettb: I did call it non_api_queues to begin with... but then I saw there is code that publishes to those queues.

app/config.py
```diff
@@ -220,6 +235,7 @@ class Config:
             "visibility_timeout": 310,
```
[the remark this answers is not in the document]
corlettb: Removed. I've removed it in the other PRs as well. Thanks Leo.

app/config.py
corlettb: Renamed to external_queues

risicle:
Rebased for testing...

risicle (approved):

PR #4415, opened 2025-04-04
feat: Add duplication check and stale timeout logic to report-requests 

joybytes:
## PR Summary:

- Duplicate detection logic: When a request is submitted, we now check for an existing request with the same user, service, report type, and parameter values (notification_type, notification_status). If a matching request is already in a non-final state (pending or in progress) and is within the timeout window, we return that instead of creating a new one.

- Stale request handling: Requests are considered stale if updated_at is older than the configured timeout. If updated_at is null, we fall back to using created_at.

- Validation-first approach: Only notifications_report is supported at the moment. Any unsupported or invalid inputs will be rejected during schema validation.

- Test coverage: Added and extended tests

### Ticket:
[Lock mechanism to prevent duplicate requests](https://trello.com/c/UuqvwE7H/1219-lock-mechanism-to-prevent-duplicate-requests)

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -1,4 +1,5 @@
 import itertools
+import json
 import uuid
 from datetime import datetime
 
@@ -43,7 +44,11 @@ from app.dao.fact_notification_status_dao import (
     fetch_stats_for_all_services_by_date_range,
 )
 from app.dao.organisation_dao import dao_get_organisation_by_service_id
-from app.dao.report_requests_dao import dao_create_report_request, dao_get_report_request_by_id
+from app.dao.report_requests_dao import (
+    dao_create_report_request,
+    dao_get_oldest_ongoing_report_request,
+    dao_get_report_request_by_id,
+)
 from app.dao.returned_letters_dao import (
     fetch_most_recent_returned_letter,
     fetch_recent_returned_letter_count,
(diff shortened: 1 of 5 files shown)


leohemsted (commented):
still need to look through the tests closely but heres some comments while i do that

app/service/rest.py
```diff
@@ -1429,25 +1434,73 @@ def get_report_request_by_id(service_id, request_id):
...
+        if existing_request:
+            current_app.logger.info(
+                "Report request %s for user %s (service %s) with params %s",
+                existing_request.id,
+                existing_request.user_id,
+                existing_request.service_id,
+                json.dumps(existing_request.parameter, separators=(",", ":")),
+            )
+
+            return jsonify(data=existing_request.serialize()), 200
```
spatel033, line 1468: We don’t currently have a design for handling existing requests that are still in progress. For example, if a user requests an email report and it takes around 20 minutes, then tries to request it again, they’ll just get the existing request. It would be helpful to show a message on the UI letting the user know that a request is already in progress, so they’re aware. It would be better to create card for this.

joybytes: we don't need a design for handling duplicates, 
if a request is duplicate from there we return the data to the UI and the flow follows normally.
Normally = polling the get_request_report and checking for the "completed" status

spatel033 (approved):

PR #4416, opened 2025-04-04
Add unique constraint to OrganisationPermission model

klssmith:
At the moment, there's nothing to stop an organisation having lots of rows for the same permission in the database. While the frontend code shouldn't be adding duplicate rows let's make sure that scenario doesn't occur and give unexpected behaviour.

If creating the table from scratch this could have worked like the `service_permissions` table, with the `organisation_id` and `permission` columns forming a composite primary key. However, since the table exists it's a lot of effort to change it to that format.

There is only one type of organisation permission available, `can_ask_to_join_a_service`, and it's for a feature that has not launched, so making this change now is safe - the organisation_permission table only has a couple of rows of data.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -369,6 +369,7 @@ class OrganisationPermission(db.Model):
         unique=False,
         nullable=False,
     )
+    __table_args__ = (UniqueConstraint("organisation_id", "permission", name="uix_organisation_permission"),)
 
 
 class OrganisationUserPermissions(db.Model):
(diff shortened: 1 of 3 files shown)


klssmith:
Squawk is warning about this, but I can't find a way of removing all warnings (fixing one warning, results in the opposite warning). I think this a relatively safe migration since the table affected is essentially empty apart from a few rows of test data.

risicle:
What's more, I'm not completely convinced `IF NOT EXISTS` is really a wonderful idea here. If an index by that name does exist there's nothing guaranteeing that it matches what you expect it to be.

spatel033 (approved):

PR #4420, opened 2025-04-09
add DATABASE_TRANSACTION_TIMEOUT_MS config variable

risicle:
https://trello.com/c/hPalBorA/956-short-term-mitigation-for-timeouts-on-long-running-celery-tasks

With same absurd default value as `DATABASE_STATEMENT_TIMEOUT_MS` to maintain status quo. We will likely set this to a "real" value through environment variables in terraform.


risicle:
I'm not convinced this is going to be useful - to allow the overnight celery jobs to work we'll need to set it to >900s, which is exactly what we're trying to avoid

PR #4421, opened 2025-04-09
Fix default value for `ServiceJoinRequest.created_at column`

klssmith:
The default value should be a function, but shouldn't _call_ the function. Fixing this means that the `created_at` column now has the datetime that the join request was created, not when the app started up.


leohemsted (approved):
oops, good catch

PR #4423, opened 2025-04-11
Serialize timestamps for admin app with subsecond precision

quis:
Otherwise the admin app won’t always be able to sort things in the correct order.

Looking at examples of `def serialize()` in `models.py` we seem to prefer the timestamp with the timezone. So let’s go with that.

The admin app uses the forgiving [`dateutil.parser`](https://dateutil.readthedocs.io/en/stable/parser.html) (via [`notifications_utils.utc_string_to_aware_gmt_datetime`](https://github.com/alphagov/notifications-utils/blob/e4b6f3f3ab8944f760987af3f3039dfeb2910ff2/notifications_utils/timezones.py#L9C5-L18)) to deserialize the strings it gets from the API back into `datetime` objects. So it doesn’t need the string to be in an exact format.

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -24,7 +24,7 @@ import app.constants
 from app import db, ma, models
 from app.dao.permissions_dao import permission_dao
 from app.models import ServicePermission
-from app.utils import DATETIME_FORMAT_NO_TIMEZONE, parse_and_format_phone_number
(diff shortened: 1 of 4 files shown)


BlessedDev (approved):

BlessedDev:
> Looking at examples of `def serialize()` in `models.py` we seem to prefer the timestamp with the timezone. So let’s go with that.

This makes sense to me.

PR #4427, opened 2025-04-14
Add economy letter feature flag

spatel033:
We have added a service-level permission economy letter feature flag to make sure other teams do not accidentally use the economy rates or options while we are still developing the feature. [Card](https://trello.com/c/k2L5QcH2/1249-service-level-permission-for-economy-rates) for this PR.


leohemsted (commented):
can you add the value to `SERVICE_PERMISSION_TYPES` in constants.py too? i think this PR is a sensible time to add (rather than in future PRs that are trying to use the permission)

leohemsted (approved):

PR #4430, opened 2025-04-15
Write to both callback tables

BlessedDev:
This PR refactors `create_service_callback_api`, `update_service_callback_api`, `remove_service_callback_api` to write to the `service_callback_api` table alongside the `service_inbound_api` for `inbound_sms` callback.

This is to facilitate the merge of the `service_inbound_api` table data into the `service_callback_api` table.

`fetch_service_callback_api` remains unchanged and retrieves `inbound_sms` callback data from the `service_inbound_api` table. This has ramifications for the updating `inbound_sms` callback data stored in the `service_callback_api` table as `update_service_callback_api` utilises `service_inbound_api` `id` which would be different from the `id` of the same callback data in the `service_callback_api` table.
Additionally there will be callback data that is present in the `service_inbound_api` that would not be present in the `service_callback_api` until the table merge is completed.

`service_callback_api` and `service_inbound_api` can't have the same `ids`

***

[Overall plan](https://docs.google.com/document/d/1Oo8VbB3tdJMLPM6WslJlbbwANA33_lJcN3a6AKkCL7A/edit?tab=t.0):
- [x] [Add inbound_sms type](https://github.com/alphagov/notifications-api/pull/4431#pullrequestreview-2772182640) to service_callback_type table
- [x] Start writing to existing `service_callback_api`  and `service_inbound_api` sms tables for inbound SMS callbacks (need to consider updating and deleting too)&ensp;**👈 you are here**
- [ ] Migrate the data from `service_inbound_api` to `service_callback_api` table and from `service_inbound_api_history` to `service_callback_api_history`
- [ ] Start reading from `service_callback_api` table for inbound SMS callbacks
- [ ] Remove the else statements in the API view endpoints and DAO methods, and everything related to the `service_inbound_api` table – in other words stop updating/deleting the old table
- [ ] Migration to remove the `service_inbound_api` and `service_inbound_api_history` tables

This PR is dependent on https://github.com/alphagov/notifications-api/pull/4431

### DIFF
--- a/app/dao/service_callback_api_dao.py
+++ b/app/dao/service_callback_api_dao.py
@@ -32,6 +32,10 @@ def get_service_callback_api(service_callback_api_id, service_id, callback_type)
     ).first()
 
 
+def get_service_callback_api_by_callback_type(service_id, callback_type):
+    return ServiceCallbackApi.query.filter_by(service_id=service_id, callback_type=callback_type).first()
+
+
 def get_delivery_status_callback_api_for_service(service_id):
     return ServiceCallbackApi.query.filter_by(
         service_id=service_id, callback_type=ServiceCallbackTypes.delivery_status.value
--- a/app/service/callback_rest.py
+++ b/app/service/callback_rest.py
@@ -115,18 +115,26 @@ REMOVE_SERVICE_CALLBACK_ERROR_MESSAGES = {
 def remove_service_callback_api(callback_api_id, service_id):
     callback_type = request.args.get("callback_type")
     if callback_type == ServiceCallbackTypes.inbound_sms.value:
-        callback_api = get_service_inbound_api(callback_api_id, service_id)
-        delete_callback_api_method = delete_service_inbound_api
+        service_inbound_api = get_service_inbound_api(callback_api_id, service_id)
+        if not service_inbound_api:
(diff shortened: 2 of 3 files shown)


app/dao/service_callback_api_dao.py
```diff
@@ -32,6 +32,10 @@ def get_service_callback_api(service_callback_api_id, service_id, callback_type)
     ).first()
 
 
+def get_service_callback_api_by_callback_type(service_id, callback_type):
```
CrystalPea, line 35: do we need both `get_service_callback_api` and `get_service_callback_api_by_callback_type`? They are very similar.

CrystalPea (commented):
Hey, great work on this, well done.

One little comment.

app/dao/service_callback_api_dao.py
BlessedDev: It is a temporary addition just to solve the issue where an update is being applied to an inbound_sms callback which has been written to both tables. The callback will be have different ids in both tables, and because we are fetching the data from the `service_inbound_api` table, the id presented to the update_service_callback_api endpoint would be that of the row in the `service_inbound_api` table. This new dao method helps us to retrieve an `inbound_sms` callback present for the service by callback type. It would be removed once the table merge is completed.

CrystalPea (approved):

PR #4434, opened 2025-04-16
Rename join a service routes

klssmith:
This moves code around and temporarily duplicates it so that we can rename the join a service URLs before the feature goes live. See individual commits for full details.

**The route to create a join request**
We had a 2 routes to create a join request
1. `/service/<uuid:service_id>/service-join-request` - this is the URL we want, but didn't contain much functionality
2. `/service/<service_id>/invite/request-for/<user_to_invite_id>` this contained the functionality, but we decided not to use this URL

This moves the functionality from 2. into 1.

**The routes to get and update a join request**
We decided we wanted to change the URLs from 
* `GET /service/service-join-request/<uuid:request_id>` to `GET /service/<service_id>/service-join-request/<uuid:request_id>`
* `POST /service/update-service-join-request-status/<uuid:request_id>` to `POST /service/<service_id>/service-join-request/<uuid:request_id>`

This adds the new routes (the old ones will be removed later), which required new DAO functions which take a service id

### DIFF
--- a/app/dao/service_join_requests_dao.py
+++ b/app/dao/service_join_requests_dao.py
@@ -68,6 +68,30 @@ def dao_update_service_join_request(
     return service_join_request
 
 
+@autocommit
+def dao_update_service_join_request_by_id(
+    request_id: UUID,
+    service_id: UUID,
+    status: Literal["approved", "rejected", "pending", "cancelled"],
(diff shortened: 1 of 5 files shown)


joybytes (approved):
⭐ 🥇 👸

PR #4435, opened 2025-04-17
feat: Persist  reason field on service join request creation

joybytes:
## PR Summary
- Extended service join requests to persist the `reason` field when we create the request
- Previously we persisted the data during the update of a request, but not during the creation. Now we do it for both.
- The field is by default None if no value is available.

### Ticket:
[Join a service dev work](https://trello.com/c/ozzxcO3o/1221-join-a-service-dev-work)

### DIFF
--- a/app/dao/service_join_requests_dao.py
+++ b/app/dao/service_join_requests_dao.py
@@ -14,11 +14,15 @@ from app.models import ServiceJoinRequest, User
 
 @autocommit
 def dao_create_service_join_request(
-    requester_id: UUID, service_id: UUID, contacted_user_ids: list[UUID]
+    requester_id: UUID,
+    service_id: UUID,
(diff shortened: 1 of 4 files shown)


klssmith (approved):
Tested locally and all works nicely 🎉

app/dao/service_join_requests_dao.py
```diff
@@ -103,5 +107,4 @@ def dao_cancel_pending_service_join_requests(requester_id: UUID, approver_id: UU
                 request_id=request.id,
                 status=SERVICE_JOIN_REQUEST_CANCELLED,
                 status_changed_by_id=approver_id,
-                reason="system update due to cancellation",
```
spatel033, line 106: The reason was used only during cancellation, now it is only for creation. Not needed anything but it's for reference.

PR #4439, opened 2025-04-17
Send economy postage to dvla

spatel033:
There is new "economy" postage option available for letters. Updating the DVLA API request payload to send "ECONOMY" in the despatchMethod. All other logic for the existing postage classes will stay the same.
For the callback the mailingProduct will be shown as 'UNSORTEDE' for UNSORTED ECONOMY and 'MM ECONOMY' for Mailmark Economy.  The postage class will be '2ND' for both.
More details on the [card](https://trello.com/c/zY3ijvpi/1245-send-economy-postage-to-dvla)

### DIFF
--- a/app/notifications/notifications_letter_callback.py
+++ b/app/notifications/notifications_letter_callback.py
@@ -52,7 +52,16 @@ dvla_letter_callback_schema = {
                                 "properties": {
                                     "key": {"const": "mailingProduct"},
                                     "value": {
-                                        "enum": ["UNCODED", "MM UNSORTED", "UNSORTED", "MM", "INT EU", "INT ROW"]
+                                        "enum": [
+                                            "UNCODED",
+                                            "MM UNSORTED",
+                                            "UNSORTED",
+                                            "MM",
+                                            "INT EU",
+                                            "INT ROW",
+                                            "UNSORTEDE",
+                                            "MM ECONOMY",
+                                        ]
                                     },
                                 }
                             },
@@ -180,9 +189,8 @@ def extract_properties_from_request(request_data) -> LetterUpdate:
 
 
 def _get_cost_threshold(mailing_product: str, postage: str) -> LetterCostThreshold:
-    if mailing_product == "MM" and postage == "2ND":
+    if (mailing_product == "MM" or mailing_product == "MM ECONOMY") and postage == "2ND":
         return LetterCostThreshold("sorted")
-
     return LetterCostThreshold("unsorted")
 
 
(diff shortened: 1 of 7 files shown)


app/schema_validation/__init__.py
```diff
@@ -135,6 +135,15 @@ def validate_letter_production_run_date(instance):
     raise ValidationError("Datetime format is invalid. It must be in the format %Y-%m-%d %H:%M:%S.%f")
```
leohemsted, line 135: we can remove this old unused validator

app/notifications/notifications_letter_callback.py
```diff
@@ -94,8 +87,9 @@
                 "jobType": {"type": "string"},
                 "jobStatus": {"type": "string", "enum": [DVLA_NOTIFICATION_DISPATCHED, DVLA_NOTIFICATION_REJECTED]},
                 "templateReference": {"type": "string"},
+                "transitionDate": {"type": "string", "format": "letter_transition_date"},
```
leohemsted, line 90: can we just use the built in `"format": "date-time"` rather than having a custom format? it looks like it's just an iso timestamp

app/notifications/notifications_letter_callback.py
```diff
@@ -177,8 +171,8 @@ def extract_properties_from_request(request_data) -> LetterUpdate:
     postage = next(item["value"] for item in despatch_properties if item["key"] == "postageClass")
     cost_threshold = _get_cost_threshold(mailing_product, postage)
 
-    despatch_datetime = next(item["value"] for item in despatch_properties if item["key"] == "productionRunDate")
-    despatch_date = _get_despatch_date(despatch_datetime)
+    despatch_datetime = request_data["data"]["transitionDate"]
+    despatch_date = datetime.datetime.strptime(despatch_datetime, "%Y-%m-%dT%H:%M:%SZ").date()
```
leohemsted, line 177: if it's an iso timestamp we can use `datetime.fromisoformat`

however, we then need to make sure we get a date from it - we'll probs need to use `convert_utc_to_bst` from notifications_utils.timezone

(and make sure you add a test first to prove that it's currently buggy if we get a callback at 23:30 utc during summertime

app/notifications/notifications_letter_callback.py
spatel033: I tried `"format": "datetime"` which did not work so created custom validator. Thanks.

leohemsted: ah no worries, that's definitely thrown me off in the past too!

there's a list here (we use draft 7 to validate, so should get all of these)

https://json-schema.org/understanding-json-schema/reference/type#built-in-formats

leohemsted (approved):

PR #4445, opened 2025-04-23
feat: Support Economy Mail in Notify

joybytes:
## PR Summary:

Extended tests and logic to support Economy Mail. 
Most of our DB supports any string for postage, tests were added mirroring DB functionality by adding "economy" postage to the following key areas:

- Notifications and Notification History
- Fact Billing (including report billing and letter despatch)
- Letter Rates, Template, Template History  

Additionally, schema_validation was updated to include economy too.

### Ticket:
- [Allow postage to be economy in Notify](https://trello.com/c/Y95RhqNO/1244-allow-postage-to-be-economy-in-notify)


app/schema_validation/__init__.py
```diff
@@ -58,7 +58,7 @@ def validate_schema_postage_including_international(instance):
     For validating postage sent by admin when sending a precompiled letter, where postage can include international
     """
     if isinstance(instance, str):
-        if instance not in ["first", "second", "europe", "rest-of-world"]:
+        if instance not in ["first", "second", "economy", "europe", "rest-of-world"]:
             raise ValidationError("invalid. It must be first, second, europe or rest-of-world.")
```
klssmith, line 62: You can change this validation error message to include "economy" now - this validator is only used by the admin app, so we won't affect what API users see

app/schema_validation/__init__.py
```diff
@@ -47,7 +47,7 @@ def validate_schema_postage(instance):
     For validating postage on templates and user requests, where postage can only be `first` or `second`
     """
     if isinstance(instance, str):
-        if instance not in ["first", "second"]:
+        if instance not in ["first", "second", "economy"]:
```
klssmith, line 50: Can we hold off adding this in for now? It will start letting API users send economy letters - my PR to stop them being created will need to go in first, and I just want to test a few things with that

app/constants.py
```diff
@@ -113,18 +113,20 @@
...
-UK_POSTAGE_TYPES = [FIRST_CLASS, SECOND_CLASS]
+POSTAGE_TYPES = [FIRST_CLASS, SECOND_CLASS, ECONOMY_CLASS, EUROPE, REST_OF_WORLD]
+UK_POSTAGE_TYPES = [FIRST_CLASS, SECOND_CLASS, ECONOMY_CLASS]
 INTERNATIONAL_POSTAGE_TYPES = [EUROPE, REST_OF_WORLD]
 
 # Map postage zones to abbreviations for our print provider
 RESOLVE_POSTAGE_FOR_FILE_NAME = {
     FIRST_CLASS: 1,
     SECOND_CLASS: 2,
+    ECONOMY_CLASS: 3,
```
klssmith, line 129: This was some discussion about whether we could use "ECONOMY" here instead of 3 https://gds.slack.com/archives/C06S65BV0K1/p1745507145097329?thread_ts=1745326611.307059&cid=C06S65BV0K1. We'd need to test if that's ok when sending letters, but perhaps we can try that out, and if there's a problem go back to using 3

app/schema_validation/__init__.py
joybytes: Like we mentioned on the call:
1) I will wait for your PR to be merged first
2) I did not update the error message so we don't let users know that "economy" is available until release

app/schema_validation/__init__.py
joybytes: done, thanks

app/constants.py
joybytes: I have changed from 3 -> "ECONOMY"

@spatel033 during your testing,would you be able to test letters with the ECONOMY and see if they provide any feedback on it?
I am keen on the word economy on the pdf too, but let's see if DVLA complains about it before we make this change.

joybytes: I am resolving this as I don't think it's a blocker for a merge and it's an easy change if something unexpected happens with the DVLA testing

klssmith:
There's just a little merge conflict to fix now, otherwise this is good to go

klssmith (approved):

PR #4448, opened 2025-04-23
Attempt number 2: Enforce daily international sms limit for services

CrystalPea:
This is a de-revert of a revert of this PR: https://github.com/alphagov/notifications-api/pull/4433

Why the deploy went wrong the first time round:
Two things went wrong: 
- first - ServiceSchema didn't know about the `international_sms_message_limit` key
- second - all cached Service objects in Redis didn't have  `international_sms_message_limit` key.

What we did to protect against it this time:
- we added the `international_sms_message_limit` key to ServiceSchema: https://github.com/alphagov/notifications-api/pull/4441
- Tested with Leo: we managed to replicate the issue - by having a cache dict for service without the key for the new limit. The issue can be remediated by clearing the Service cache ahead of deploy.
- TODO: clear Service cache to get rid of stale entries without the `international_sms_message_limit` key


leohemsted (approved):
🙏

PR #4450, opened 2025-04-24
Stop economy mail letters being created by services without the service permission

klssmith:
Services without the economy mail feature flag should not be able to create a letter with economy postage.

This change stops letters from being created, and is adding an extra layer of protection, since the api should not be given the data to create economy mail letters for unauthorised services in the first place. Users of a service without permission won't see the economy mail option in admin, and we also have API schema validation which currently blocks economy postage.


joybytes (approved):
⭐

PR #4454, opened 2025-04-29
Set default limits for existing services

CrystalPea:
First get services that sent international SMS in the last 12 months, and how many they sent on their busiest day.
Then for each of those services, if they sent more than 50 on their busiest day, and if they are still at default limit, set their new international_sms_message_limit to busiest day * 2.

Then get all remaining services that have their international_sms_message_limit set to default value (250,000) and set their limit to 100 per day.

This has been tested locally:
- for service with default limit that did not send any intl sms (updates to 100)
- for service with default limit that sent 200 intl sms on busiest day (updates to 400)
- for service with default limit that sent 49 intl sms on busiest day (updates to 100)
- for service not on default limit (it shouldn't update)

The ticket: https://trello.com/c/BmYkI2tB/272-update-default-international-sms-daily-limit-based-on-historical-sending-data-for-each-service

Needs to be merged after: https://github.com/alphagov/notifications-admin/pull/5446

Squawk migration linting check fails, it doesn't like `.fetchall()`, I tested on a case where that query returns no results though, and it all works fine.

### DIFF
--- a/migrations/versions/0500_update_intl_limit_values.py
+++ b/migrations/versions/0500_update_intl_limit_values.py
@@ -1,0 +1,84 @@
+"""
+Create Date: 2025-04-29 12:36:49.360512
+"""
+
+import textwrap
+
+from alembic import op
+from flask import current_app
+from sqlalchemy import text
+
+revision = '0500_update_intl_limit_values'
+down_revision = '0499_merge_callback_tables'
+
+
+def upgrade():
+    conn = op.get_bind()
+
+    # First get services that sent international SMS in the last 12 months, and how many they sent on their busiest day.
+    # Then for each of those services, if they sent more than 50 on their busiest day,
+    #  set their new international_sms_message_limit to busiest day * 2.
+
+    # Then get all remaining services that have their international_sms_message_limit set to default value (250,000)
+    # and set their limit to 100 per day.
+
+    services_with_intl_sms = conn.execute(
+        text(
+            textwrap.dedent(
+                """
+                SELECT intl_billing.service_id, MAX(intl_billing.intl_sent) AS max_daily_intl
+                FROM (SELECT
+                    service_id, bst_date, SUM(notifications_sent) as intl_sent FROM ft_billing WHERE bst_date > '2024-04-21' AND notification_type = 'sms' AND international = true GROUP BY service_id, bst_date
+                ) AS intl_billing
+
+                GROUP BY intl_billing.service_id ORDER BY max_daily_intl DESC;
+                """
+            )
+        ),
+    ).fetchall()
+
+    services_with_custom_limit = []
+
+    for service_id, max_daily_intl in services_with_intl_sms:
+        new_limit = max_daily_intl * 2 if max_daily_intl > 50 else 100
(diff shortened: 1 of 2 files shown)


migrations/versions/0500_update_intl_limit_values.py
```diff
@@ -0,0 +1,84 @@
...
+    # Then get all remaining services that have their international_sms_message_limit set to default value (250,000)
+    # and set their limit to 100 per day.
+
+    services_with_intl_sms = conn.execute(
+        text(
+            textwrap.dedent(
+                """
+                SELECT intl_billing.service_id, MAX(intl_billing.intl_sent) AS max_daily_intl
+                FROM (SELECT
+                    service_id, bst_date, SUM(notifications_sent) as intl_sent FROM ft_billing WHERE bst_date > '2024-04-21' AND notification_type = 'sms' AND international = true GROUP BY service_id, bst_date
```
rparke, line 31: Should this date be hardcoded?

Do we want the upgrade to target services in the last 12 months from whenever the upgrade is run or do we want it to target services for a fixed date range `2024-04-21` to `2025-04-20`?

migrations/versions/0500_update_intl_limit_values.py
```diff
@@ -0,0 +1,84 @@
...
+                SET
+                    international_sms_message_limit = 100,
+                    updated_at = now()
+                WHERE
+                    international_sms_message_limit = 250000
+                """
+            )
+        ),
+        services_with_custom_limit,
+    )
```
rparke, line 80: Edge case that may have been tested and that we won't hit on the upgrade:

if somebody has a custom limit of 125,000 set, the query above will set them to 250,000 correctly and this second query will downgrade them to 100 unexpectedly.

Looping over the set of service_ids not in `services_with_custom_limit` would solve this even if it's a bit slower (there are probably faster options) but it's only a migration that's running once and means we know we won't hit this edge case.

migrations/versions/0500_update_intl_limit_values.py
CrystalPea: I think the start date doesn't matter too much, and the query should look until the present moment at the time of running.

My reasoning: we want to get services busiest day in recent time - it doesn't really matter if it's in last 12 or 13 months, but it does matter that we don't skip ant recent days.

So this query meets both these conditions. I don't feel very strongly about it though. I just like the simplicity of it.

migrations/versions/0500_update_intl_limit_values.py
CrystalPea: I'm not sure I understand - if somebody has a custom limit of 125,000 set, that service will not be caught by any of these queries. As for both of them starting limit needs to be the default (as per the WHERE clauses).

migrations/versions/0500_update_intl_limit_values.py
rparke: Tbh then that's fine with me. It's just about catching _enough_ people who are using it somewhat recently.

migrations/versions/0500_update_intl_limit_values.py
rparke: Ah **yes** you're right. I missed that they all need to be 250,000

PR #4460, opened 2025-05-07
Remove economy mail feature flag and permission checks

klssmith:
This stops using the economy mail permission feature flag now that it should be rolled out to all services, undoing the work in https://github.com/alphagov/notifications-api/pull/4450

Database permissions will need to be removed separately

### DIFF
--- a/app/notifications/process_letter_notifications.py
+++ b/app/notifications/process_letter_notifications.py
@@ -1,9 +1,8 @@
 from notifications_utils.recipient_validation.postal_address import PostalAddress
 
 from app import create_random_identifier
-from app.constants import ECONOMY_LETTER_SENDING, LETTER_TYPE
+from app.constants import LETTER_TYPE
 from app.notifications.process_notifications import persist_notification
-from app.notifications.validators import check_service_has_permission
 
 
 def create_letter_notification(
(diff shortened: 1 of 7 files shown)


spatel033:
We added `economy_letter_sending` in [db](https://github.com/alphagov/notifications-api/pull/4427/files#diff-9a30052457e1b767d1f64a483ccdd3ec883ee6c4aed04a7ff4ab38cb9e0a252c). Do we need to change this back?

klssmith:
@spatel033 Yes, it does need to be removed from the database but that will be done in a separate PR after this (along with removing all the [other unused service permissions](https://github.com/alphagov/notifications-api/pull/4464) from the database that I'm tidying up). When deploying the apps the database migrations get run first before the code is deployed, which is why I'm splitting this change in two - so we don't remove the permissions in the database while they're still referenced in code.

spatel033 (approved):

klssmith:
@joybytes @spatel033 I've removed the last commit from this - the one which removed the unused constants. That part will need to be deployed later as part of tidying the code because otherwise you can get errors if trying to update service that still has the economy mail permission

spatel033 (approved):

PR #4461, opened 2025-05-07
Treat crown dependency numbers as domestic when checking limits

CrystalPea:
When checking and adding to daily international sms limit.

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -7,9 +7,7 @@ from notifications_utils.clients.redis import (
 )
 from notifications_utils.recipient_validation.email_address import validate_and_format_email_address
 from notifications_utils.recipient_validation.errors import InvalidPhoneError
-from notifications_utils.recipient_validation.phone_number import (
-    PhoneNumber,
-)
+from notifications_utils.recipient_validation.phone_number import PhoneNumber
 from notifications_utils.recipient_validation.postal_address import PostalAddress
 from sqlalchemy.orm.exc import NoResultFound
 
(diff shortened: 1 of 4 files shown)


app/notifications/validators.py
```diff
@@ -159,7 +157,8 @@ def validate_and_return_extended_phone_number_info(service, send_to, key_type, c
 
         recipient_data = _get_extended_phone_number_info(phone_number, send_to)
 
-        if recipient_data["international"] and check_intl_sms_limit:
+        # UK_PREFIX includes standard domestic numbers, and crown dependency numbers
+        if str(phone_number.number.country_code) != UK_PREFIX and check_intl_sms_limit:
```
quis, line 161: We have [a method](https://github.com/alphagov/notifications-utils/blob/9154f4b5d0ff5d9ad212cb6f08b31297bdeeb63a/notifications_utils/recipient_validation/phone_number.py#L224-L232) for this already which reads a bit nicer:
```suggestion
        if not phone_number.is_uk_phone_number() and check_intl_sms_limit:
```

app/notifications/validators.py
```diff
@@ -159,7 +157,8 @@ def validate_and_return_extended_phone_number_info(service, send_to, key_type, c
 
         recipient_data = _get_extended_phone_number_info(phone_number, send_to)
 
-        if recipient_data["international"] and check_intl_sms_limit:
+        # UK_PREFIX includes standard domestic numbers, and crown dependency
+        if (not phone_number.is_uk_phone_number()) != UK_PREFIX and check_intl_sms_limit:
```
quis, line 161: `!= UK_PREFIX` is wrong 

Swap the order and no brackets are needed:
```suggestion
        if check_intl_sms_limit and not phone_number.is_uk_phone_number():
```

app/notifications/validators.py
```diff
@@ -7,9 +7,7 @@
 )
 from notifications_utils.recipient_validation.email_address import validate_and_format_email_address
 from notifications_utils.recipient_validation.errors import InvalidPhoneError
-from notifications_utils.recipient_validation.phone_number import (
-    PhoneNumber,
-)
+from notifications_utils.recipient_validation.phone_number import UK_PREFIX, PhoneNumber
```
quis, line 10:
```suggestion
from notifications_utils.recipient_validation.phone_number import PhoneNumber
```

kr8n3r (approved):

PR #4463, opened 2025-05-08
only increment international SMS limit cache for SMS notifications

leohemsted:
we were inadvertently doing this for all notifications. previously we used `notification.international` which resolves to False for emails, but we switched to checking if the prefix is NOT uk - the prefix for emails is None so this was resolving as true.

(fun fact: it would have continued to resolve false for international letters)


quis (approved):

leohemsted:
@quis sensible idea, i added to this PR

PR #4465, opened 2025-05-09
Refactor `increment_daily_limit_cache` to be easier to follow

quis:
By splitting the code which talks to Redis from the code which figures out which caches to set it should be easier to follow what is going on.

It also reduces duplication of the Redis-talking code and lets us put all the logic for international SMS in one place, rather than double-checking and using [flag arguments](https://martinfowler.com/bliki/FlagArgument.html).

### DIFF
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -161,11 +161,7 @@ def persist_notification(
     # if simulated create a Notification model to return but do not persist the Notification to the dB
     if not simulated:
         dao_create_notification(notification)
-        increment_daily_limit_caches(
-            service,
-            notification,
-            key_type,
-        )
+        increment_daily_limit_caches(service, notification, key_type)
 
     return notification
 
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -161,7 +161,7 @@ def persist_notification(
     # if simulated create a Notification model to return but do not persist the Notification to the dB
     if not simulated:
         dao_create_notification(notification)
-        increment_daily_limit_cache(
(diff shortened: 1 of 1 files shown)


app/notifications/process_notifications.py
```diff
@@ -161,37 +161,24 @@ def persist_notification(
...
-        cache_key = redis.daily_limit_cache_key(service_id, notification_type=notification_type_)
-        if redis_store.get(cache_key) is None:
-            # if cache does not exist set the cache to 1 with an expiry of 24 hours,
-            # The cache should be set by the time we create the notification
-            # but in case it is this will make sure the expiry is set to 24 hours,
-            # where if we let the incr method create the cache it will be set a ttl.
-            redis_store.set(cache_key, 1, ex=86400)
-        else:
-            redis_store.incr(cache_key)
+    increment_daily_limit_cache(service.id)
```
CrystalPea, line 173: I think we haven't added this type of cache key before the change?

quis: We do. Because of this line we set caches for notification types of `None` and `notification_type`:
https://github.com/alphagov/notifications-api/blob/7a91fb85183457fa02fa6457c300e79c9babb856/app/notifications/process_notifications.py#L178

I don’t this existing way of doing this (with a 2-item loop) is very clear to read.

You can see it’s covered by the tests here:
https://github.com/alphagov/notifications-api/blob/7a91fb85183457fa02fa6457c300e79c9babb856/tests/app/notifications/test_process_notification.py#L288-L291

CrystalPea: Ahh, you are right 😮

PR #4466, opened 2025-05-12
Decouple service inbound api

BlessedDev:
This PR is in preparation for dropping the `service_inbound_api` and `service_inbound_api_history` tables. It removes most references to both tables.


quis (approved):
Good to go

It might be worth raising a separate PR to beef up some of the tests in https://github.com/alphagov/notifications-api/blob/main/tests/app/dao/test_service_callback_api_dao.py which assume that `callback_type` will always be `delivery_status`. Feels a bit like the rate limit thing where the tests only covered the `sms` case.

BlessedDev:
> Good to go
> 
> It might be worth raising a separate PR to beef up some of the tests in https://github.com/alphagov/notifications-api/blob/main/tests/app/dao/test_service_callback_api_dao.py which assume that `callback_type` will always be `delivery_status`. Feels a bit like the rate limit thing where the tests only covered the `sms` case.

That is a good point. I will raise a separate PR as suggested to address the issue.

PR #4467, opened 2025-05-12
Drop service inbound api table

BlessedDev:
This PR contains a migration to drop both the `service_inbound_api` and `service_inbound_api_history` tables.
It also updates the DetailedServiceSchema.
There are a few sqauwk-migration-checks warnings that may need to be silenced.
This PR is dependant on https://github.com/alphagov/notifications-api/pull/4466 being merged in first. It is also linked to this PR https://github.com/alphagov/notifications-admin/pull/5465

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -861,40 +861,6 @@ class ServiceGuestList(db.Model):
         return f"Recipient {self.recipient} of type: {self.recipient_type}"
 
 
-class ServiceInboundApi(db.Model, Versioned):
-    __tablename__ = "service_inbound_api"
-    id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
-    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey("services.id"), index=True, nullable=False, unique=True)
-    service = db.relationship("Service", backref="inbound_api")
-    url = db.Column(db.String(), nullable=False)
-    _bearer_token = db.Column("bearer_token", db.String(), nullable=False)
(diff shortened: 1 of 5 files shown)


CrystalPea (commented):
Hi Gozie,

I looked at the code and it looks good, but I run into some issues on my local - not sure if they are related or not.

Let's meet tomorrow and try to check it together?

Maybe let's try

klssmith:
Might be the same issues that @CrystalPea was referring to, but this will cause an error when deployed on its own because admin is expecting to there to be an `inbound_api` property on the service, but that no longer exists. So [this line](https://github.com/alphagov/notifications-admin/blob/fcf7f2ca189a317e5ed43bc3b25fe35fd654e7d5/app/models/service.py#L660) nows gives an error of  `AttributeError: 'Service' object has no attribute 'inbound_api'`.

If this error doesn't appear for you locally, it could be the service you are using for testing is cached in Redis so still has the `inbound_api` attribute. So you will also need to consider cached values in the order of rolling out and think about at what stage Redis should be cleared.

klssmith:
I should have added that this wouldn't be a problem long term, because I can see that there is an admin PR to update the service model. But the problems can occur after this change has been merged but before the admin one has, so as the changes are rolled out.

CrystalPea (approved):
This all works when following [the deployment plan](https://docs.google.com/document/d/1Oo8VbB3tdJMLPM6WslJlbbwANA33_lJcN3a6AKkCL7A/edit?tab=t.0#heading=h.xh5nxy9lqjny)

So I'm happy to approve, as long as that is followed.

PR #4469, opened 2025-05-13
Add economy mail postage to letters volume email

klssmith:
The automated email we send to the DVLA each night needs updating to
include economy mail letters, otherwise these will be counted in the
international letters section.

The templates will need to updated manually. Updating the template in
the migration ensures that new environments that are created will have
the correct content.


klssmith:
Force pushed to correct a typo in the migration

spatel033 (approved):

PR #4472, opened 2025-05-16
Add task to generate report requests

klssmith:
We have an endpoint to create report requests, and a class which generates the data. This connects the two bits by creating a task which, when called from the endpoint, uses our existing code to generate the data.

The task does nothing if the report request is already in "pending" status. Otherwise it, calls the code to generate the data and updates the status of the report request appropriately.

https://trello.com/c/ubo67aY3/1265-celery-task-report-data-processing

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -26,6 +26,10 @@ from app.constants import (
     NOTIFICATION_CREATED,
     NOTIFICATION_RETURNED_LETTER,
     NOTIFICATION_VALIDATION_FAILED,
+    REPORT_REQUEST_FAILED,
+    REPORT_REQUEST_IN_PROGRESS,
+    REPORT_REQUEST_PENDING,
+    REPORT_REQUEST_STORED,
     SMS_TYPE,
 )
 from app.dao.jobs_dao import dao_get_job_by_id, dao_update_job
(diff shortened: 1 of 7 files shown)


app/celery/tasks.py
```diff
@@ -526,3 +536,31 @@ def _check_and_queue_returned_letter_callback_task(notification_id, service_id):
...
+    report_request = dao_get_report_request_by_id(service_id=service_id, report_id=report_request_id)
+
+    current_app.logger.info(
+        "Starting process-report-request task for report request id %s and status %s",
+        report_request_id,
+        report_request.status,
+    )
+
+    if report_request.status != REPORT_REQUEST_PENDING:
+        return
```
joybytes, line 552: might be an overkill but could we log something here too, since we are not returning errors, just in case.

app/celery/tasks.py
```diff
@@ -56,6 +62,10 @@ class UnprocessableJobRow(Exception):
     pass
 
 
+class ProcessReportRequestException(Exception):
```
joybytes, line 65: how does this work?

app/celery/tasks.py
klssmith: Discussed offline and we decided not to add an extra log line here since the log line above shows if the task was called when the request is in "pending" state

app/celery/tasks.py
klssmith: Discussed offline

PR #4482, opened 2025-05-23
Fix report request size

spatel033:
Without seek(0), the CSV ends up with a lot of extra empty space at the end. This makes the file much larger than it should be and slows down report generation. Adding self.csv_buffer.seek(0) moves the cursor to the start of the buffer before reading from it. This makes sure the buffer is fully cleared so the next batch of data starts clean.


app/report_requests/process_notifications_report.py
```diff
@@ -130,6 +130,7 @@ def _upload_csv_part_if_needed(self) -> None:
         data_bytes = self.csv_buffer.getvalue().encode("utf-8")
         if len(data_bytes) >= S3_MULTIPART_UPLOAD_MIN_PART_SIZE:
             self._upload_part(data_bytes)
+            self.csv_buffer.seek(0)
```
joybytes, line 133: maybe let's add a comment here just so we can remember three months from now why this is important?
I know I'd forget 😄 
something like:
Make sure truncation starts from the beginning to avoid residual data

joybytes: Also, should we include this on `_fetch_and_upload_notifications` too?
```
self._upload_part(data_bytes)
self.csv_buffer.seek(0)
self.csv_buffer.truncate(0)
```

joybytes (approved):

app/report_requests/process_notifications_report.py
```diff
@@ -130,6 +130,10 @@ def _upload_csv_part_if_needed(self) -> None:
         data_bytes = self.csv_buffer.getvalue().encode("utf-8")
         if len(data_bytes) >= S3_MULTIPART_UPLOAD_MIN_PART_SIZE:
             self._upload_part(data_bytes)
+
+            # Reset the buffer for the next part
+            # truncate(0) does not reset the cursor so seek(0) is needed to reset the cursor
+            self.csv_buffer.seek(0)
```
CrystalPea, line 136: Should we add a unit test for this?

PR #4485, opened 2025-06-02
Fix empty test assertions

quis:
`called_with` and `called_once` are not methods of `unittest.Mock`.

Instead they are mocked functions which will:
- accept any arguments
- always return something `True`-ish

This means the assertions in these tests are doing nothing, and changing their arguements or the arguements of the methods they are mocking does not cause the tests to fail.

This commit switches to using methods which do exist.


klssmith (approved):
Nice fix

PR #4492, opened 2025-06-04
Set utils to v99.8 & rebrand flag to False

tombye:
Brings in the version of utils with the new GOV.UK branding and sets the flag that turns it on in our emails to off.

## Notes for reviewers

Please try sending an email when your email branding is set to GOV.UK. Then change the `rebrand` argument to `True`, send another and check it changes to the new brand.


kr8n3r (approved):

kr8n3r (commented):
emails are sent with correct branding

PR #4505, opened 2025-06-25
Add initial otel spans and metrics

corlettb:
https://trello.com/c/mmrH8QbC/1154-turn-second-otel-spike-into-a-pr-we-can-merge

### DIFF
--- a/app/celery/process_ses_receipts_tasks.py
+++ b/app/celery/process_ses_receipts_tasks.py
@@ -4,6 +4,7 @@ from datetime import datetime, timedelta
 import iso8601
 from celery.exceptions import Retry
 from flask import current_app, json
+from opentelemetry import metrics
 from sqlalchemy.orm.exc import NoResultFound
 
 from app import notify_celery, statsd_client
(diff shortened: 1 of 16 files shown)


app/celery/process_ses_receipts_tasks.py
```diff
@@ -78,12 +94,24 @@ def process_ses_results(self, response):
...
             statsd_client.timing_with_dates(
                 f"callback.ses.{notification_status}.elapsed-time", datetime.utcnow(), notification.sent_at
             )
 
+            otel_ses_notification_processing_histogram.record(
+                (datetime.utcnow() - notification.sent_at).total_seconds(),
+                {
+                    "notification_status": notification_status,
+                },
+            )
```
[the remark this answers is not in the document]
corlettb: I've not looked at trying to aggregate span events into metrics. I've just used span events for additional logging so far. Obviously here we can't use a span as an sqs queue would have been involved at some point.

app/otel/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+
+    instrument_map = {
+        "wsgi": (_instrument_wsgi, "WSGI"),
+        "celery": (_instrument_celery, "Celery"),
+        "flask": (_instrument_flask, "Flask"),
+        "redis": (_instrument_redis, "Redis"),
+        "requests": (_instrument_requests, "Requests"),
+        "sqlalchemy": (_instrument_sqlalchemy, "SQLAlchemy"),
+        "psycopg2": (_instrument_psycopg2, "Psycopg2"),
```
risicle, line 87: Do we need both sqlalchemy and psycopg2?

app/otel/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+
+def _instrument_app(app: Flask) -> None:
+    """
+    Apply OpenTelemetry instrumentation
+    """
+
+    instrument_map = {
+        "wsgi": (_instrument_wsgi, "WSGI"),
+        "celery": (_instrument_celery, "Celery"),
+        "flask": (_instrument_flask, "Flask"),
```
risicle, line 83: Do we need both flask and wsgi?

PR #4507, opened 2025-06-26
Set the status of outdated report requests to deleted

spatel033:
Generated report request files are stored in an S3 bucket with a 1-day lifecycle policy. After the files are deleted from S3, the related database entries become redundant, so we mark them as 'deleted'. We keep these deleted report requests for a while to support stats or investigations. More info on this [card](https://trello.com/c/qsDngRr6/1349-daily-job-to-update-s3-files-as-deleted-phase-1b).

First need to merge https://github.com/alphagov/notifications-api/pull/4506

### DIFF
--- a/app/celery/nightly_tasks.py
+++ b/app/celery/nightly_tasks.py
@@ -40,6 +40,7 @@ from app.dao.notifications_dao import (
     get_service_ids_with_notifications_before,
     move_notifications_to_notification_history,
 )
+from app.dao.report_requests_dao import update_report_requests_status_to_deleted
 from app.dao.service_data_retention_dao import (
     fetch_service_data_retention_for_all_services_by_notification_type,
 )
@@ -394,3 +395,14 @@ def delete_unneeded_notification_history_for_specific_hour(start_datetime: str, 
     )
 
     delete_notification_history_between_two_datetimes(start_datetime, end_datetime)
+
+
+@notify_celery.task(name="update-report-status-to-deleted")
+@cronitor("update-report-status-to-deleted")
+def update_report_status_to_deleted():
+    try:
+        update_report_requests_status_to_deleted()
+        current_app.logger.info("Successfully updated report status to deleted.")
+    except SQLAlchemyError as e:
+        current_app.logger.error("Failed to update report status to deleted: %s", str(e))
+        raise
--- a/app/config.py
+++ b/app/config.py
@@ -343,6 +343,11 @@ class Config:
                 "schedule": crontab(hour=2, minute=0),
                 "options": {"queue": QueueNames.PERIODIC},
             },
+            "update-report-status-to-deleted": {
+                "task": "update-report-status-to-deleted",
+                "schedule": crontab(hour=2, minute=00),
+                "options": {"queue": QueueNames.PERIODIC},
+            },
             "remove_sms_email_jobs": {
                 "task": "remove_sms_email_jobs",
                 "schedule": crontab(hour=4, minute=0),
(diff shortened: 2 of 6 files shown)


app/config.py
```diff
@@ -383,6 +383,11 @@ class Config:
                 "schedule": crontab(hour=19, minute=00),
                 "options": {"queue": QueueNames.PERIODIC},
             },
+            "update-report-status-to-deleted": {
```
klssmith, line 386: This is a really minor thing, but can we move this in the list so that it's next to `save-daily-notification-processing-time`? The nightly tasks are roughly listed in time order, which makes it easy to tell what's running at what time of night

app/dao/report_requests_dao.py
```diff
@@ -63,3 +76,14 @@ def dao_get_oldest_ongoing_report_request(
...
     return report_request
+
+
+def update_report_requests_status_to_deleted():
+    cutoff = datetime.utcnow() - timedelta(days=1)
+
+    ReportRequest.query.filter(
+        ReportRequest.status != REPORT_REQUEST_DELETED,
+        ReportRequest.created_at < cutoff,
+    ).update({"status": REPORT_REQUEST_DELETED}, synchronize_session=False)
```
klssmith, line 87: The `updated_at` time should be changed too if we're updating the status, like what we do for `dao_timeout_notifications` https://github.com/alphagov/notifications-api/blob/3a24edd661bcde6aede1fbd5e61581bddbaa943d/app/dao/notifications_dao.py#L544-L546

app/celery/nightly_tasks.py
```diff
@@ -394,3 +395,13 @@ def delete_unneeded_notification_history_for_specific_hour(start_datetime: str,
     )
 
     delete_notification_history_between_two_datetimes(start_datetime, end_datetime)
+
+
+@notify_celery.task(name="update-report-status-to-deleted")
```
klssmith, line 400: Looking at the other nightly tasks, they all tend to have Cronitor enabled, so perhaps good to do here too?

app/dao/report_requests_dao.py
```diff
@@ -63,3 +76,14 @@ def dao_get_oldest_ongoing_report_request(
 def dao_update_report_request(report_request: ReportRequest) -> ReportRequest:
     db.session.add(report_request)
     return report_request
+
+
+def update_report_requests_status_to_deleted():
+    cutoff = datetime.utcnow() - timedelta(days=1)
```
klssmith, line 82: Can we change the cutoff to look at the last few days instead of just the last day? It shouldn't be needed, but means that if there are ever issues with the task we don't need to manually edit production data because things will get fixed the following night.

app/celery/nightly_tasks.py
```diff
@@ -394,3 +395,13 @@ def delete_unneeded_notification_history_for_specific_hour(start_datetime: str,
...
+def update_report_status_to_deleted():
+    try:
+        update_report_requests_status_to_deleted()
+        current_app.logger.info("Successfully updated report status to deleted.")
+    except SQLAlchemyError as e:
```
klssmith, line 405: Optional: without this error handling we'd still see the error and stack trace in the logs and know to investigate, so we can get away without catching and then re-raising to add an additional log line

app/dao/report_requests_dao.py
spatel033: Without `updated_at` also updated the date in db, but refactored for consistency.

app/dao/report_requests_dao.py
spatel033: The `cutoff` date is the starting point, and the query will find all the days before that date.
https://github.com/alphagov/notifications-api/blob/b2b11e9bb1a951e34a954a94686aeec7a1a3c89f/app/dao/report_requests_dao.py#L84-L87

klssmith: I'd wrongly read the cutoff as being the date we were looking _after_ and not before so can see why it is 1 day. It would be good to have two dates in the query though so that we're not checking the entire table every night. It could search for anything created 4 or 5 days ago up until the cutoff of 1 day, which will provide some contingency in case the task fails one night.

PR #4509, opened 2025-07-02
Add api_key_id index to notification_history

klssmith:
This is a temporary change to allow us to delete the old emergency alerts data. The `api_key_id` foreign key constraint on the `notification_history` table means that when attempting to delete the api_keys, Postgres also needs to check the `notification_history` table which is taking too long without an index.


risicle:
Ohhh yeah I remember now.

risicle (approved):
WFM dev-d

PR #4512, opened 2025-07-04
Add job_id index to notification_history

klssmith:
This is a temporary change to allow us to delete the old emergency alerts data. The `job_id` foreign key constraint on the `notification_history` table means that when attempting to delete the jobs, Postgres also needs to check the `notification_history` table which is taking too long without an index.

It was not possible to send a broadcast as a job, but there are still broadcast services in the database which have rows in the `jobs` table because in some cases, we converted an existing "normal" service to a broadcast service.

### DIFF
--- a/migrations/versions/0506_n_history_job_id_index.py
+++ b/migrations/versions/0506_n_history_job_id_index.py
@@ -1,0 +1,22 @@
+"""
+Create Date: 2025-07-04 09:28:38.029592
+"""
+
+from alembic import op
+
+revision = '0506_n_history_job_id_index'
+down_revision = '0505_n_history_api_key_index'
+
+
+def upgrade():
+    with op.get_context().autocommit_block():
+        op.execute(
+            "CREATE INDEX CONCURRENTLY IF NOT EXISTS ix_notification_history_job_id on notification_history (job_id)"
+        )
+
+
+def downgrade():
+    with op.get_context().autocommit_block():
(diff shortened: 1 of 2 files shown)


migrations/versions/0506_n_history_job_id_index.py
```diff
@@ -0,0 +1,23 @@
+"""
+Create Date: 2025-07-04 09:28:38.029592
+"""
+
+from alembic import op
+import sqlalchemy as sa
```
ipdigital24, line 6: do we need `import sqlalchemy as sa`
and in other pr you had `from sqlalchemy.dialects import postgresql`, which isn't here

quis (approved):
Checked this does the same thing as https://github.com/alphagov/notifications-api/pull/4509/files#diff-ea5a33ee04380a59c6a84479fb5231157b0b9372fbfe60a6c854e23467a4a5e7

👍🏻

migrations/versions/0506_n_history_job_id_index.py
klssmith: By default, new migrations come with these imports in the migration file
```python
from alembic import op
import sqlalchemy as sa
from sqlalchemy.dialects import postgresql
```

Neither of the last 2 imports are needed here (or in the previous PR) I just forgot to remove them. I've now removed both

quis (approved):

PR #4513, opened 2025-07-07
Add service_id index to notification_history 

klssmith:
This is a temporary change to allow us to delete the old emergency alerts data.
The service_id foreign key constraint on the notification_history table means
that when attempting to delete the services, Postgres also needs to check the
notification_history table which is taking too long without an index.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0506_n_history_job_id_index
+0507_n_history_service_id_index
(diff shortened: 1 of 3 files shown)


quis (approved):
Confirmed this one doesn’t exist in production:
```
Indexes:
    "notification_history_pkey" PRIMARY KEY, btree (id)
    "ix_notification_history_api_key_id" btree (api_key_id)
    "ix_notification_history_created_at" btree (created_at)
    "ix_notification_history_job_id" btree (job_id)
    "ix_notification_history_reference" btree (reference)
    "ix_notification_history_service_id_composite" btree (service_id, key_type, notification_type, created_at)
```

PR #4516, opened 2025-07-09
Remove emergency alerts data

klssmith:
This is a data migration (so no downgrade) that deletes broadcast services and all associated data - this includes obvious things like broadcast events and templates, but also everything connected to those services.

Although some of these theoretically shouldn't exist for broadcast services (eg returned_letters), we attempt to delete them out of an abundance of caution.

[This work is based off an older PR](https://github.com/alphagov/notifications-api/pull/4201), which was tested against a copy of the staging database. That PR was reverted because it was taking too long to delete from the `notification_history` table due to the foreign key constraints - since then, indexes have been added to notification_history to make deleting from the table faster.

Differences between this PR and the original are:
* the `service_inbound_api` table no longer exists, so has been removed from the migration
* this adds the `api_keys_history`, `service_callback_api_history`, `ft_billing` and `ft_notification_status` tables to the list we delete from
* this deletes from the `notifications` and `notification_history` tables first - these have various foreign key constraints on other tables which would cause issues if the data from those other tables is deleted first.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0508_n_history_templates_index
+0509_delete_broadcast_data
--- a/migrations/versions/0509_delete_broadcast_data.py
+++ b/migrations/versions/0509_delete_broadcast_data.py
@@ -1,0 +1,109 @@
+"""
+Create Date: 2025-07-07 15:46:53.673184
+"""
+
+from alembic import op
+from sqlalchemy.sql import text
+
+revision = '0509_delete_broadcast_data'
+down_revision = '0508_n_history_templates_index'
+
+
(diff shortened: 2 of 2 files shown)


klssmith:
Squawk is giving this error
```python
...
    res = results.fetchall()
          ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'fetchall' 
```

I have tested this locally and confirmed that `results` is a `sqlalchemy.engine.cursor.LegacyCursorResult`, and that calling its `fetchall` method returns a list. The same code has also been run on production before (the first time this was deployed) without issues, so this error seems to be specific to how the migrations are being tested with Squawk

PR #4517, opened 2025-07-09
Allow X-Forwarded-For overrides for the incoming ip address

corlettb:
https://trello.com/c/CjPJrFAk/1160-notifications-api-recording-all-remoteaddr-as-internal-ips

What
----

Allow X-Forwarded-For overrides for the incoming ip address

Why
----

This should show the X-Forwarded-For ips in our logs.


robinjam:
Ah cool, thanks Ben! Didn't expect you to reply 😅

PR #4521, opened 2025-07-17
Fix dvla rejected letter callback payload

spatel033:
When receiving a letter callback from the DVLA about the status of a letter, there are two possible values for jobStatus that we allow - DESPATCHED or REJECTED. Their documentation can be found here. The JSON schema we made to validate this data is defined here. This works for letters which are DESPATCHED, but we’ve found that for letters that are REJECTED, the data received does not validate according to the data in the schema and instead looks like this
```
{
    comment: Unable to download PDF  - Connection reset,
    jobId: 1efee8ae-93a8-789b-7f58-2946d0ee4a76,
    jobStatus: REJECTED,
    jobType: NOTIFY,
    templateReference: NOTIFY,
    transitionDate: 2025-06-23T17:17:04Z
}
```

This means that because the callback fails validation, we don’t give a validation error and don’t update the letter to technical-failure status as intended. It looks like we’ll need to change how we validate letters which have the REJECTED status so that the error we get is the one we intend to raise. Fix for this [card](https://trello.com/c/69mDVHiM/217-ensure-our-letter-callback-code-handles-letters-that-are-rejected-as-intended).


tests/app/notifications/test_notifications_letter_callbacks.py
```diff
@@ -218,6 +218,43 @@ def test_process_letter_callback_calls_process_letter_callback_data_task(
     )
 
 
+def test_process_rejected_letter_callback_calls_process_letter_callback_data_task(
```
CrystalPea, line 221: in the test above this one, there are cases for both `DESPATCHED` and `REJECTED` statuses, and page count is not 0. Does this mean we do different things with rejected letters based on different scenarios?

tests/app/notifications/test_notifications_letter_callbacks.py
```diff
@@ -218,6 +218,43 @@ def test_process_letter_callback_calls_process_letter_callback_data_task(
...
+    assert response.status_code == 204, response.json
+
+    mock_task.assert_called_once_with(
+        queue="letter-callbacks",
+        kwargs={
+            "notification_id": uuid.UUID("cfce9e7b-1534-4c07-a66d-3cf9172f7640"),
+            "dvla_status": "REJECTED",
+            "despatch_date": datetime.date(2025, 6, 23),
+            "page_count": 0,
+            "cost_threshold": "unsorted",
```
CrystalPea, line 253: Could we use the `LetterCostThreshold.unsorted` variable here, like the test above, for consistency?

CrystalPea (commented):
Hi, I have a question, in comment below:

PR #4532, opened 2025-07-30
readme change to trigger new commit SHA for image build

WesleyHindle:


quis:
Do we still need this?

PR #4202, opened 2024-10-01
Remove emergency alerts db schema

leohemsted:
Lots to read in the commits

PR #4536, opened 2025-08-07
Remove emergency alerts database tables

klssmith:
Now that there's no data that depends on these, we can delete the tables completely. This was important to do as separate steps, or we could end up in a case where have have fragments of broadcast services sitting in our database but without the broadcast-specific tables, which would leave us unsure as to what those services were.

This commit is taken from https://github.com/alphagov/notifications-api/pull/4202. The earlier work in that PR has already been merged, and this is the first stage of removing the database schema, which I'm splitting into separate PRs instead of doing in one.

### DIFF
--- a/migrations/versions/0510_delete_broadcast_tables.py
+++ b/migrations/versions/0510_delete_broadcast_tables.py
@@ -1,0 +1,184 @@
+"""
+Create Date: 2025-08-07 11:08:46.900469
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = "0510_delete_broadcast_tables"
+down_revision = "0509_delete_broadcast_data"
+
+
+def upgrade():
+    op.drop_table("service_broadcast_settings")
+    op.drop_table("broadcast_channel_types")
+    op.drop_table("broadcast_provider_message_number")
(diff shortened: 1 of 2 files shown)


spatel033 (approved):

migrations/versions/0510_delete_broadcast_tables.py
```diff
@@ -0,0 +1,184 @@
...
+        sa.ForeignKeyConstraint(
+            ["created_by_api_key_id"], ["api_keys.id"], name="broadcast_message_created_by_api_key_id_fkey"
+        ),
+        sa.ForeignKeyConstraint(["created_by_id"], ["users.id"], name="broadcast_message_created_by_id_fkey"),
+        sa.ForeignKeyConstraint(["service_id"], ["services.id"], name="broadcast_message_service_id_fkey"),
+        sa.ForeignKeyConstraint(["status"], ["broadcast_status_type.name"], name="broadcast_message_status_fkey"),
+        sa.ForeignKeyConstraint(
+            ["template_id", "template_version"],
+            ["templates_history.id", "templates_history.version"],
+            name="broadcast_message_template_id_fkey",
```
spatel033, line 107: optional: current foreign key name is based on id and version `broadcast_message_template_id_template_version_fkey`

PR #4539, opened 2025-08-11
Token bucket rate limiting

rparke:
Brings in the token bucket algorithm.

We have a bucket max of 1,000 and bucket min of 0. This  keeps the per minute rate limit "the same". A per minute limit of 3,000 is just translated to a 50 tokens per second replenishment rate, with some headroom.

For full context see https://docs.google.com/document/d/1Px21_Seg_kfaSktscGY4XWC80tdwbwxcX8OXIB65TUI/edit?tab=t.0#heading=h.y3i6qo1ruhuw

This will only affect services who have the permission (none at the moment).

***

Stages of rollout:
- [x] https://github.com/alphagov/notifications-utils/pull/1246
- [x] https://github.com/alphagov/notifications-api/pull/4548
- [ ] https://github.com/alphagov/notifications-api/pull/4539  **👈🏻 you are here**
- [ ] https://github.com/alphagov/notifications-api/pull/4552
- [ ] https://github.com/alphagov/notifications-api/pull/4558
- [ ] (remove the feature flag)

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -65,7 +65,7 @@ def token_bucket_rate_limit_exceeded(service, key_type):
     with REDIS_EXCEEDED_RATE_LIMIT_DURATION_SECONDS.labels(algorithm="token_bucket").time():
         return (
             redis_store.get_remaining_bucket_tokens(
-                key=f"{service.id}-{key_type}",
+                key=f"{service.id}-tokens-{key_type}",
                 replenish_per_sec=int(service.rate_limit / SECONDS_IN_1_MINUTE),
                 bucket_max=TOKEN_BUCKET_MAX,
                 bucket_min=TOKEN_BUCKET_MIN,
--- a/tests/app/notifications/test_validators.py
+++ b/tests/app/notifications/test_validators.py
@@ -520,7 +520,7 @@ def test_check_token_bucket_service_over_api_rate_limit_when_exceed_rate_limit_r
 
         assert app.redis_store.get_remaining_bucket_tokens.call_args_list == [
             mocker.call(
-                key=f"{str(service.id)}-{api_key.key_type}", replenish_per_sec=50, bucket_max=1_000, bucket_min=0
+                key=f"{str(service.id)}-tokens-{api_key.key_type}", replenish_per_sec=50, bucket_max=1_000, bucket_min=0
             )
         ]
         assert e.value.status_code == 429
@@ -546,7 +546,7 @@ def test_check_token_bucket_service_over_api_rate_limit_when_rate_limit_has_not_
         check_service_over_api_rate_limit(serialised_service, serialised_api_key.key_type)
         assert app.redis_store.get_remaining_bucket_tokens.call_args_list == [
             mocker.call(
-                key=f"{str(service.id)}-{api_key.key_type}", replenish_per_sec=50, bucket_max=1_000, bucket_min=0
+                key=f"{str(service.id)}-tokens-{api_key.key_type}", replenish_per_sec=50, bucket_max=1_000, bucket_min=0
             )
         ]
 


app/notifications/validators.py
```diff
@@ -48,9 +50,17 @@ def check_service_over_api_rate_limit(service, key_type):
...
         cache_key = rate_limit_cache_key(service.id, key_type)
         rate_limit = service.rate_limit
-        interval = 60
+        interval = 60  # keep this for now so that we have a like for like mapping of the per minute rate limit
+        rate_limit_in_seconds = rate_limit / interval
+        token_bucket_max = rate_limit_in_seconds * TOKEN_BUCKET_MAX_MULTIPLIER
+        token_bucket_min = rate_limit_in_seconds * TOKEN_BUCKET_MIN_MULTIPLIER
         with REDIS_EXCEEDED_RATE_LIMIT_DURATION_SECONDS.time():
-            if redis_store.exceeded_rate_limit(cache_key, rate_limit, interval):
+            if (
```
quis, line 58: This now needs to be conditional on `service.has_permission("token_bucket")`

app/notifications/validators.py
```diff
@@ -45,14 +48,32 @@
...
+    with REDIS_EXCEEDED_RATE_LIMIT_DURATION_SECONDS.time():
+        if rate_limit_exceeded(service, key_type):
+            current_app.logger.info("service %s has been rate limited for throughput", service.id)
+            raise RateLimitError(service.rate_limit, SECONDS_IN_1_MINUTE, key_type)
+
+
+def rate_limit_exceeded(service, key_type):
+    cache_key = rate_limit_cache_key(service.id, key_type)
+
+    if service.has_permission("token_bucket"):
```
rparke, line 65: This returns null of it's not in the table, yes? (vs an error)

quis: Yes: https://github.com/alphagov/notifications-api/blob/f460b181ae2ee3e938ce7ebc189ea95545c4eede/app/serialised_models.py#L128-L129

BlessedDev (approved):

app/notifications/validators.py
```diff
@@ -45,14 +48,32 @@
...
-                current_app.logger.info("service %s has been rate limited for throughput", service.id)
-                raise RateLimitError(rate_limit, interval, key_type)
+    if not current_app.config["API_RATE_LIMIT_ENABLED"]:
+        return
+    if not current_app.config["REDIS_ENABLED"]:
+        return
+
+    with REDIS_EXCEEDED_RATE_LIMIT_DURATION_SECONDS.time():
+        if rate_limit_exceeded(service, key_type):
+            current_app.logger.info("service %s has been rate limited for throughput", service.id)
```
quis, line 58: Should log different messages depending on token bucket or sliding window

app/notifications/validators.py
```diff
@@ -41,18 +41,46 @@
...
-                raise RateLimitError(rate_limit, interval, key_type)
+    if not current_app.config["API_RATE_LIMIT_ENABLED"]:
+        return
+    if not current_app.config["REDIS_ENABLED"]:
+        return
+
+    if service.has_permission("token_bucket"):
+        if token_bucket_rate_limit_exceeded(service, key_type):
+            current_app.logger.info("service %s has been rate limited for token bucket", service.id)
+            raise RateLimitError(service.rate_limit, SECONDS_IN_1_MINUTE, key_type)
```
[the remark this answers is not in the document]
quis: Unfortunately we can’t change the error message in case anyone has written code to expect that string.

Some decisions on guidance/documentation here: https://docs.google.com/document/d/1vvGDOsJduGayZ-T6wvtgf59IPz_035E-sHEI46NLocY/edit?pli=1&tab=t.0#heading=h.hi97ef8pxq7x

CrystalPea (approved):

PR #4552, opened 2025-09-01
Give new services the token bucket permission

quis:
We’re going to give all existing services this permission. We need to make sure that services created after we’ve done the migration also get the permission.

We can merge this any time before the migration. Merging further in advance of the migration means we can soft launch the feature by giving it only to new services (which are unlikely to start at very high rates of sending anyway).


rparke:
LGTM

PR #4580, opened 2025-09-26
sqlalchemy init: dispose engine after events are set up

risicle:
to drop all current connections and ensure all connections in our pool were initialized using the newly installed hooks


risicle:
Nope observed problem appears to happen when connection's "first" transaction is a rolled back.

PR #4582, opened 2025-10-01
Add resanitise letter attachment task

spatel033:
Add celery task to re-sanitise template letter attachments.

This task calls the  template preview task which recreates the PDF for a letter attachment which is already sanitised and in the letters-attachment bucket. The new file that is generated will then overwrite the existing letter in the letter-attachments bucket.


CrystalPea (approved):
LGTM, ready to go after the related template preview PR is merged 👍🏼

PR #4676, opened 2025-12-18
feat: Add queue grouping helper controlled by feature flag

joybytes:
## Summary 
Introduce a helper to generate MessageGroupId values based on queue-specific rules, improving fairness between services and traffic types.

Grouping is guarded by the ENABLE_SQS_FAIR_GROUPING feature flag to allow safe rollout and rollback, and includes initial tests for the new behaviour.

Resource for implementation of `MessageGroupId` [here](https://docs.celeryq.dev/en/main/getting-started/backends-and-brokers/sqs.html#:~:text=With%20FIFO%20queues%20it%20might%20be%20necessary%20to%20set%20additional%20message%20properties%20such%20as%20MessageGroupId%20and%20MessageDeduplicationId%20when%20publishing%20a%20message)

### Ticket


app/config.py
```diff
@@ -560,6 +551,7 @@ class Config:
 
     REPORT_REQUEST_NOTIFICATIONS_TIMEOUT_MINUTES = 30
     REPORT_REQUEST_NOTIFICATIONS_CSV_BATCH_SIZE = 2500
+    FAIR_QUEUE_ENABLED = False
```
DilwoarH:
```suggestion
    ENABLE_SQS_FAIR_GROUPING = os.getenv("ENABLE_SQS_FAIR_GROUPING", False)
```

app/celery/queue_utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+def get_message_group_id_for_queue(
+    *,
+    queue_name: QueueNames,
+    service_id: str | None = None,
+    notification_type: str | None = None,
+    origin: str | None = None,
+    key_type: str | None = None,
+) -> dict[str, str]:
+    if not current_app.config.get("ENABLE_SQS_FAIR_GROUPING", False):
```
DilwoarH:
```suggestion
    if not current_app.config.get("ENABLE_SQS_FAIR_GROUPING"):
```

DilwoarH (changes requested):

app/celery/queue_utils.py
```diff
@@ -0,0 +1,50 @@
...
+        "notification_type": notification_type,
+        "origin": origin,
+        "key_type": key_type,
+    }
+
+    parts: list[str] = []
+    for field in grouping_fields:
+        value = values.get(field)
+        if value:
+            parts.append(value)
```
spatel033:
```suggestion
        parts.append(str(value) if value else "unknown")
```
Instead of skipping missing values, this keeps the shape of group id consistent and logs would be easier to parse.

app/celery/queue_utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+    if not parts:
+        return {}
```
spatel033: It would be better to log a warning message, since we expect a group ID but it is missing.

DilwoarH (changes requested):
Minor changes

PR #4776, opened 2026-03-12
Use new DVLA endpoints when sending letters 

klssmith:
This changes the `send_letter` method of the `DVLAClient` to send the letter in a different way - by uploading the file before making the POST request.

Currently, sending a letter is done in a single request to the `/v1/print/jobs` endpoint that includes the file.

The new way of sending involves making 3 requests:
1. A GET request to `/v1/print/files/upload-url` to get a presigned URL
2. A PUT request to upload the letter file to the URL returned by step 1
3. A POST request to `/v1/print/jobs` with the `uploadId` from step 1 included in the request data

These three requests are all made consecutively from the same method, which means we don't need to store the URL and uploadId from step 1 - they are used immediately.

Any of the stages failing with a retryable error will result in the whole `deliver_letter` task being retried. It is possible for a letter to be uploaded successfully at stage 2, then a failure at stage 3 cause the task to retry and it to be uploaded again to a new presigned URL before being sent. If a letter is uploaded but there is no corresponding print request for the uploadId associated with the presigned URL, the file will not be sent by the DVLA and will get deleted.

Screenshot 2026-04-10 at 08 14 26

# Error handling at the new stages of sending a letter
**Stage 1** Getting a presigned URL requires authentication so has the same error handling as making the POST request.

**Stage 2** We don't handle the case of a URL having expired when uploading a file to it - URLs are valid for 1 hour and the delay between getting a new URL and using it is seconds at most. We treat rate limit errors (503s) from this endpoint like any other 5xx status code and retry. We don't expect to see rate limiting here because we'd hit rate
limiting in stage 1 first.

### DIFF
--- a/app/clients/letter/dvla.py
+++ b/app/clients/letter/dvla.py
@@ -1,4 +1,3 @@
-import base64
 import contextlib
 import secrets
 import string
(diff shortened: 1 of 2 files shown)


joybytes (approved):
this looks good to me and the tests are very thorough

joybytes:
will we be doing any manual tests with this change?
by manual I mean, sending actual letters?

klssmith:
> will we be doing any manual tests with this change? by manual I mean, sending actual letters?

There's currently ongoing testing of the change in test environments (sending from our test environment to the DVLA one), and that has confirmed that this change is sending the right data across and works. The plan before deploying this to production is to manually send a few real letters in production using this code earlier in the day before all letters get automatically sent in the evening. But I'll write down the planned steps etc. once this is ready to deploy so that people have a chance to review them

PR #4807, opened 2026-04-02
refactor: move notifications and jobs to read replica

joybytes:
## Summary
Directed queries from the following endpoints to read replica

- /service/{service_id}/job/scheduled-job-stats
- /service/{service_id}/job/{job_id}/notifications
- /service/{service_id}/job
- /service/{service_id}/notifications/count
- /service/{service_id}/notifications/monthly

###  Ticket
[DB Replica - Admin Dashboard: Notifications & Jobs](https://trello.com/c/rb1NBhGU/1737-db-replica-admin-dashboard-notifications-jobs)

### DIFF
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -8,6 +8,7 @@ from notifications_utils.letter_timings import (
     letter_can_be_cancelled,
 )
 from sqlalchemy import and_, asc, desc, func
+from sqlalchemy.orm import Session, scoped_session
 
 from app import db, redis_store
 from app.constants import (
@@ -31,7 +32,7 @@ from app.models import (
     ServiceDataRetention,
     Template,
 )
-from app.utils import midnight_n_days_ago
+from app.utils import midnight_n_days_ago, retryable_query
 
 
 def dao_get_notification_outcomes_for_job(job_id):
@@ -85,11 +86,13 @@ def dao_get_jobs_by_service_id(
     )
 
 
+@retryable_query()
 def dao_get_scheduled_job_stats(
     service_id,
+    session: Session | scoped_session = db.session,
 ):
     return (
-        db.session.query(
+        session.query(
             func.count(Job.id),
             func.min(Job.scheduled_for),
         )
--- a/app/job/rest.py
+++ b/app/job/rest.py
@@ -3,6 +3,7 @@ from datetime import UTC
 import dateutil
 from flask import Blueprint, current_app, jsonify, request
 
+from app import db
 from app.aws.s3 import get_job_metadata_from_s3
 from app.celery.tasks import process_job
 from app.config import QueueNames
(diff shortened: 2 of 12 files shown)


app/dao/jobs_dao.py
```diff
@@ -299,16 +306,28 @@ def find_missing_row_for_job(job_id, job_size):
 
 @redis_cache.set("job-{job_id}-notification-outcomes", ttl_in_seconds=timedelta(days=1).total_seconds())
 def get_possibly_cached_notification_outcomes_for_job(
```
spatel033, line 304: The data is cached here, so we don’t need to use the read replica.

app/job/rest.py
spatel033, line 67: `get_job_by_service_and_job_id` is used in `cancel_job`. After cancelling the job in the primary database, it immediately tries to fetch the job from the replica database. This can return an outdated status for the cancelled job.

It would be better not to move `get_job_by_service_and_job_id` to the replica for now, or to add extra logic to handle this edge case.

app/upload/rest.py
```diff
@@ -47,7 +48,10 @@ def get_paginated_uploads(service_id, limit_days, page):
         }
         if upload.upload_type == "job":
             upload_dict["statistics"] = get_possibly_cached_notification_outcomes_for_job(
```
spatel033, line 49: The data is cached here, so we don’t need to use the read replica.

app/dao/jobs_dao.py
joybytes: Hey, just checking this again

`get_possibly_cached_notification_outcomes_for_job` it only skips DB work on a Redis cache hit, on a cache miss, it still queries via the passed session.

Are you suggesting that if there is any sort of caching involved we should not call replica at all?

app/upload/rest.py
joybytes: same as this comment https://github.com/alphagov/notifications-api/pull/4807#discussion_r3080640484

app/dao/jobs_dao.py
joybytes: I think I understood the flow a bit better and just want to check with you before pushing the change


I think this commit should be removed then 389a9bb5a8b6aef6cd7d99e0fd94f8dc588de1a0

`cancel_job` -> `get_job_by_service_and_job_id` -> ` dao_get_notification_outcomes_for_job` (Like you mentioned - it makes sense to keep this in primary)

`dao_get_notification_outcomes_for_job` is also used in `get_possibly_cached_notification_outcomes_for_job` which you mentioned below that since it has cache involved, we should hit the primary

`get_possibly_cached_notification_outcomes_for_job` is also used in get_jobs_by_service(`get_paginated_jobs`) and `get_paginated_uploads`

`get_paginated_jobs` calls `dao_get_jobs_by_service_id` we should probably get that information from primary too to make the whole flow consistent.

what do you think?

joybytes: I removed the commit ✅

spatel033 (approved):

PR #4824, opened 2026-04-16
Remove `send_files_via_ui` service permission

rparke:
🚨 🚨 🚨 **DO NOT MERGE UNTIL https://github.com/alphagov/notifications-admin/pull/5858 is merged and deployed** 🚨 🚨 🚨 

remove "send_files_via_ui" service permission

removes the "send_files_via_ui" service permission when the send files via email feature is released.

### DIFF
--- a/migrations/versions/0549_remove_send_files_via_ui.py
+++ b/migrations/versions/0549_remove_send_files_via_ui.py
@@ -1,0 +1,21 @@
+"""
+
+Create Date: 2026-04-23 14:00:00.0
+Revision ID: 0549_remove_send_files_via_ui
+Revises: "0548_letter_rates_from_3_5_26"
+
+"""
+
+revision = "0549_remove_send_files_via_ui"
+down_revision = "0548_letter_rates_from_3_5_26"
+
+from alembic import op
+from sqlalchemy import text
+
+def upgrade():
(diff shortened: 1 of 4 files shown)


migrations/versions/0549_remove_send_files_via_ui.py
```diff
@@ -0,0 +1,24 @@
+"""
+
+Create Date: 2026-04-13 04:00:11.329567
+Revision ID: 0548_letter_rates_from_3_5_26
```
joybytes: `Revision ID: 0549_remove_send_files_via_ui
Revises: 0548_letter_rates_from_3_5_26`

migrations/versions/0549_remove_send_files_via_ui.py
```diff
@@ -0,0 +1,24 @@
...
+Revises: 0547_new_sms_allowance_n_rate
+
+"""
+
+revision = "0549_remove_send_files_via_ui"
+down_revision = "0548_letter_rates_from_3_5_26"
+
+import uuid
+from datetime import datetime
```
joybytes: I think we can remove some of these imports as they are not being used

joybytes (approved):

PR #4826, opened 2026-04-17
`get_all_templates_for_service`: reduce overfetching with `load_only`

risicle:
Similar to #4821 (which appeared to give us a >30% performance improvement), reducing the degree to which we overfetch fields should reduce the amount of (uninterruptible) time the ORM & db connector has to spend interpreting the wide result rows.

This view is one of api's next big sinners in hogging CPU time.

### DIFF
--- a/app/dao/templates_dao.py
+++ b/app/dao/templates_dao.py
@@ -2,13 +2,13 @@ from datetime import datetime
 
 from flask import current_app
 from sqlalchemy import asc, desc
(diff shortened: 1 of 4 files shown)


quis (approved):
Had to check https://docs.sqlalchemy.org/en/21/orm/queryguide/relationships.html#prevent-lazy-with-raiseload to understand what the flag is doing.

All looks good.

PR #4830, opened 2026-04-24
feat: remove old archived email files from S3

joybytes:
### Summary
Added a daily task to remove archived template email files from S3. By default, it cleans a 1-day slice of files that are at least 14 days old. You can pass `archived_from` to backfill older files.

### Testing
**Without argument**
Running the task without archived_from uses the default cleanup window:
- archived_to = now - 14 days (retention cutoff)
- archived_from = archived_to - 1 day So it only targets files archived in that 1-day window (between 15 and 14 days old), not all files older than 14 days.

```
docker exec -it notify-api-celery \
  celery -A run_celery.notify_celery call remove-archived-template-email-files-from-s3 \
  --queue=periodic-tasks
```

**With argument**
Providing archived_from widens/narrows the start of the window for backfill:
- lower bound becomes your provided archived_from
- upper bound is still archived_to = now - 14 days This means retention is still enforced + files newer than 14 days are never included.

```
docker exec -it notify-api-celery \
  celery -A run_celery.notify_celery call remove-archived-template-email-files-from-s3 \
  --queue=periodic-tasks \
  --kwargs='{"archived_after":"2026-04-17"}'
```

### Ticket
[Automatically remove stale email files from S3](https://trello.com/c/xnuOn2bB/538-automatically-remove-stale-email-files-from-s3)


app/celery/nightly_tasks.py
```diff
@@ -80,6 +83,98 @@ def _remove_csv_files(job_types):
...
+
+    # we only delete files once retention is reached.
+    # eg: now Apr 28, retention 14 days -> archived_to is Apr 14.
+    archived_to = datetime.now(UTC) - timedelta(days=DEFAULT_TEMPLATE_EMAIL_FILE_ARCHIVE_RETENTION_DAYS)
+    # optional manual backfill start date:
+    # - if `archived_from` is passed, use it.
+    # - otherwise run a 1-day default window.
+    #   eg: if archived_to is Apr 14, delete files archived from Apr 13 up to Apr 14.
+    cleanup_window_start = (
+        _parse_archived_from(archived_from) if archived_from is not None else archived_to - timedelta(days=1)
```
CrystalPea: if the task runs once a day, could our window be three days? That would give us some grace if the task fails/doesn't run for some reason.

app/celery/nightly_tasks.py
```diff
@@ -80,6 +83,98 @@ def _remove_csv_files(job_types):
...
 
 
+@notify_celery.task(name="remove-archived-template-email-files-from-s3")
+@cronitor("remove-archived-template-email-files-from-s3")
+def remove_archived_template_email_files_from_s3(archived_from=None):
+    # `archived_from` is optional.
+    # Valid values:
+    # - "YYYY-MM-DD" (eg: "2026-01-01")
+    # - ISO datetime (eg: "2026-01-01T12:00:00+00:00" or "2026-01-01T12:00:00Z")
+    def _parse_archived_from(value):
```
CrystalPea: do we need to accept all three types of values? what would be the case for calling the task? Probably like this?
- task called programmatically by the app - in this case we don't pass anything in
- task called manually - it would be fair to only allow one format for the argument - as this shouldn't happen much, and wouldn't be run in urgency

If we only allow one format for the argument, that will make our code simpler without losing much.

app/dao/template_email_files_dao.py
```diff
@@ -52,6 +54,35 @@ def dao_get_template_email_file_by_id(template_email_file_id):
     return TemplateEmailFile.query.filter(TemplateEmailFile.id == template_email_file_id).one()
 
 
+@retryable_query()
+def dao_get_archived_template_email_files_older_than(
+    session: Session | scoped_session = db.session,
+    archived_from=None,
+    archived_to=None,
```
CrystalPea: [suggestion] we could name these args `archived_after` and `archived_before` - might make it easier to understand at a glance from which side the date limits the query?

app/dao/template_email_files_dao.py
```diff
@@ -52,6 +54,35 @@ def dao_get_template_email_file_by_id(template_email_file_id):
...
+            *(() if archived_from is None else (TemplateEmailFile.archived_at >= archived_from,)),
+            TemplateEmailFile.archived_at <= archived_to,
+        )
+        .order_by(
+            TemplateEmailFile.archived_at.asc(),
+            TemplateEmailFile.id.asc(),
+        )
+        .limit(limit)
+        # skip rows already processed by earlier batches in this run.
+        .offset(offset)
```
CrystalPea: it might not matter for this query, as it will rarely, if ever, have to offset, but offset generally is rather slow:

https://www.cybrosys.com/research-and-development/postgres/why-large-offset-queries-are-slow-in-postgresql-and-what-to-use-instead

I remember refactoring it out in queries that were struggling to boost performance.

tests/app/dao/test_template_email_files_dao.py
```diff
@@ -191,3 +194,202 @@ def test_dao_archive_template_email_file(sample_email_template, sample_template_
...
+
+@freeze_time("2026-04-23 12:00:00")
+@pytest.mark.parametrize(
+    "session,expected_bind_key",
+    (
+        (db.session, None),
+        (db.session_bulk, "bulk"),
+    ),
+)
+def test_dao_get_archived_template_email_files_older_than_supports_limit_and_offset(
```
CrystalPea: this test is pretty complex - it could use a few comments at keys steps to explain what the set up means, and what you're asserting

tests/app/dao/test_template_email_files_dao.py
```diff
@@ -191,3 +194,202 @@ def test_dao_archive_template_email_file(sample_email_template, sample_template_
...
+    "session,expected_bind_key",
+    (
+        (db.session, None),
+        (db.session_bulk, "bulk"),
+    ),
+)
+def test_dao_get_archived_template_email_files_older_than_filters_by_archived_from(
+    sample_email_template, session, expected_bind_key
+):
+    old_archived_file = create_template_email_file(
```
CrystalPea: maybe this could be named `too_old_archived_file`? Then we know it shouldn't be caught by the query.

tests/app/dao/test_template_email_files_dao.py
```diff
@@ -191,3 +194,202 @@ def test_dao_archive_template_email_file(sample_email_template, sample_template_
...
+
+@freeze_time("2026-04-23 12:00:00")
+@pytest.mark.parametrize(
+    "session,expected_bind_key",
+    (
+        (db.session, None),
+        (db.session_bulk, "bulk"),
+    ),
+)
+def test_dao_get_archived_template_email_files_older_than_archived_from_does_not_bypass_retention(
```
CrystalPea: This test seems to be a duplicate with `test_dao_get_archived_template_email_files_older_than_filters_by_archived_to` - do we need both?

app/constants.py
```diff
@@ -345,6 +345,8 @@ class ServiceCallbackTypes(enum.StrEnum):
 
 # Document download constants
 DEFAULT_DOCUMENT_DOWNLOAD_RETENTION_PERIOD = "26 weeks"
+DEFAULT_TEMPLATE_EMAIL_FILE_ARCHIVE_RETENTION_DAYS = 14
```
CrystalPea: I don't think those two consts need `DEFAULT` in the name, as we don't provide alternative values.

So they could be:
`ARCHIVED_TEMPLATE_EMAIL_FILE_RETENTION_DAYS`
and
`ARCHIVED_TEMPLATE_EMAIL_FILE_S3_CLEANUP_BATCH_SIZE`

what do you think?

app/celery/nightly_tasks.py
```diff
@@ -80,6 +83,98 @@ def _remove_csv_files(job_types):
...
+    candidates = deleted = failed = 0
+
+    current_app.logger.info(
+        ("Starting archived template email file s3 cleanup with archived_from %s and archived_to %s"),
+        cleanup_window_start.isoformat(),
+        archived_to.isoformat(),
+    )
+
+    while True:
+        # dao enforces BOTH limits:
```
CrystalPea: is this comment needed? As in - isn't this how it usually works?

app/celery/nightly_tasks.py
```diff
@@ -80,6 +83,98 @@ def _remove_csv_files(job_types):
...
+            retry_attempts=2,  # type: ignore
+            archived_from=cleanup_window_start,
+            archived_to=archived_to,
+            limit=DEFAULT_TEMPLATE_EMAIL_FILE_S3_CLEANUP_BATCH_SIZE,
+            offset=offset,
+        )
+        if not archived_files_batch:
+            break
+
+        candidates += len(archived_files_batch)
```
CrystalPea: maybe we could name this `scoped`? Then it fits with deleted and failed (all adjectives)

CrystalPea (commented):
Hi, good work on this, well done!

I left some comments. This is not the full review yet - Github seems to be playing up and losing some of my comments, so I want to submit to see what came through and avoid repeating myself. I will finish the review in a second swipe now.

CrystalPea:
I reviewed the rest of the PR and I have no more comments 🙌🏼

app/celery/nightly_tasks.py
joybytes: I have updated the deletion window to be 3 days instead of 1
https://github.com/alphagov/notifications-api/pull/4830/changes/2631cdd5430099fdaef92d9878c1f96441d50ee4

The task still receives an argument of `archived_after` to enable us to delete all archived files on the first run

app/celery/nightly_tasks.py
joybytes: The task now expects string type for "YYYY-MM-DD", so the other checks were dropped
https://github.com/alphagov/notifications-api/pull/4830/changes/2631cdd5430099fdaef92d9878c1f96441d50ee4

app/dao/template_email_files_dao.py
joybytes: Renamed all references to `archived_after` and `archived_before`

tests/app/dao/test_template_email_files_dao.py
joybytes: I tried to simplify all tests by creating a helper function `create_archived_template_email_file`
https://github.com/alphagov/notifications-api/pull/4830/changes/2631cdd5430099fdaef92d9878c1f96441d50ee4

I was unsure on a helper vs a fixture. I chose helper.

The reason behind it is that the flow we're trying to test, is quite specific to this task and I am not sure other functions in the repo would make use of the fixture. Since it seemed quite specific, I decided to go for helper, but please let me know if that's not appropriate

Before, all the logic of creating and archiving files on a specific date was reproduced in every test, which made it quite lenghty and complex. With this helper now, we can focus on the logic of recently or older archived files, and I think it makes it easier to understand what's happening. This goes for both the nightly_task and template_emails tests.

tests/app/dao/test_template_email_files_dao.py
joybytes: I use now `too_old_archived_file` and `in-window-archived-file` for files that are too old to be deleted, and files that are within the window of deletion

tests/app/dao/test_template_email_files_dao.py
joybytes: Since I updated the test to a keyset/cursor style, this test has been removed

app/constants.py
joybytes: agreed, I also added a third constant for the window `TEMPLATE_EMAIL_FILE_S3_CLEANUP_CATCH_UP_WINDOW_DAYS`

https://github.com/alphagov/notifications-api/pull/4830/changes/2631cdd5430099fdaef92d9878c1f96441d50ee4

app/celery/nightly_tasks.py
joybytes: removed

app/celery/nightly_tasks.py
joybytes: renamed candidates to scoped, which sounds better

app/celery/nightly_tasks.py
CrystalPea: nice one!

tests/app/dao/test_template_email_files_dao.py
CrystalPea: This helper is great, and I see you also did some renaming. Great work, the tests are now much clearer 👏🏼 .

Since the helper is used by two files, maybe I would put it in `tests.app.db` - as this is where we usually import similar functions from, but I don't think it's hugely important.

joybytes: great idea, I moved the helper to `tests.app.db`

joybytes:
@CrystalPea could you give a final look to the implementation of `dao_get_archived_template_email_files_older_than` please? 🙏 

I did some refactoring, and I am now using older_than and .scalar()

I noticed that the other keyset/cursor functions in our repo seems to use this pattern, so I tried to follow the same
(eg.`dao_get_paginated_inbound_sms_for_service_for_public_api`, `get_notifications_for_service`)

I also decided to use `PAGE_SIZE` config variables, which is what inbound notifications seem to use, so I just decided to follow the same limit for sizing

CrystalPea (approved):

PR #4832, opened 2026-04-27
Revert "Add support for sanitised database restore inside the fixtures"

WesleyHindle:
Reverts alphagov/notifications-api#4806

PR #4806, opened 2026-04-02
Add support for sanitised database restore inside the fixtures

corlettb:
https://trello.com/c/PqKyYsTG/1485-ensure-fixtures-apply-to-a-sanitised-database-and-functional-tests-pass

What
----

Handle sanitised database restores in functional/performance fixtures
    
Make fixture creation resilient when running against a restored sanitised database, while keeping canonical fixture names stable (Functional Tests Org, Functional Tests, Performance Tests).
    
Key changes:
    - Detect and rotate sanitised org/service rows that occupy canonical fixture names.
    - Recreate canonical fixture org/service rows deterministically after rotation.
    - Reclaim unique resources sanitised data may still own (domain, inbound number, inbound-linked SMS sender).
    - Repair invalid API key secrets to prevent BadSignature failures in auth/serialisation paths.
    - Keep fixture setup idempotent and safe to run repeatedly across environments.
    
Also updates fixture tests to cover sanitised-row rotation, resource reclaim behaviour, and API key repair logic.

Why
----

After restoring a sanitised db we need the functional and performance tests to pass

PR #4834, opened 2026-04-27
Add support for sanitised database restore inside the fixtures

WesleyHindle:
Originally https://github.com/alphagov/notifications-api/pull/4806, then reverted in https://github.com/alphagov/notifications-api/pull/4832.

Reverted because:
- The breaking PR changed the name of the service used in the FTs, resulting in a permissions issue, causing the tests to fail. 
- Reverting this PR didn't fix the issue as the service name was cached, giving a mismatch between service names
- Flushing the cache fixed this issue.

---
https://trello.com/c/PqKyYsTG/1485-ensure-fixtures-apply-to-a-sanitised-database-and-functional-tests-pass

What
Handle sanitised database restores in functional/performance fixtures

Make fixture creation resilient when running against a restored sanitised database, while keeping canonical fixture names stable (Functional Tests Org, Functional Tests, Performance Tests).

Key changes:
- Detect and rotate sanitised org/service rows that occupy canonical fixture names.
- Recreate canonical fixture org/service rows deterministically after rotation.
- Reclaim unique resources sanitised data may still own (domain, inbound number, inbound-linked SMS sender).
- Repair invalid API key secrets to prevent BadSignature failures in auth/serialisation paths.
- Keep fixture setup idempotent and safe to run repeatedly across environments.

Also updates fixture tests to cover sanitised-row rotation, resource reclaim behaviour, and API key repair logic.

Why
After restoring a sanitised db we need the functional and performance tests to pass

---

**When merging you must pause the job after the FTs. If the FTs pass, re run them and check they pass a second time. If they do not, this shows the cache must be cleared after each FT run. This PR must be reverted and amended to add this functionality.**

### DIFF
--- a/app/functional_tests_fixtures/__init__.py
+++ b/app/functional_tests_fixtures/__init__.py
@@ -4,8 +4,10 @@ from uuid import uuid4
 
 import boto3
 from flask import current_app
+from itsdangerous.exc import BadSignature
 from sqlalchemy.exc import NoResultFound
 
+from app import db
 from app.constants import (
     EDIT_FOLDER_PERMISSIONS,
     EMAIL_AUTH,
@@ -33,7 +35,6 @@ from app.dao.organisation_dao import (
     dao_create_organisation,
     dao_get_organisation_by_id,
     dao_get_organisations_by_partial_name,
-    dao_update_organisation,
 )
 from app.dao.permissions_dao import permission_dao
 from app.dao.service_callback_api_dao import get_service_callback_api_by_callback_type, save_service_callback_api
@@ -64,6 +65,7 @@ from app.dao.templates_dao import (
 )
 from app.dao.users_dao import get_user_by_email, save_model_user
 from app.models import (
+    Domain,
     InboundNumber,
     InboundSms,
     Organisation,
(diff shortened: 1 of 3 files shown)


WesleyHindle:
**When merging you must pause the job after the FTs. If the FTs pass, re run them and check they pass a second time. If they do not, this shows the cache must be cleared after each FT run. This PR must be reverted and amended to add this functionality.**

app/functional_tests_fixtures/__init__.py
```diff
@@ -102,6 +110,96 @@ def _upload_env_to_ssm(ssm_upload_path, env_config, description):
...
+    old_name = candidate.name
+    candidate.name = _build_sanitised_renamed_name(candidate.name)
+    db.session.add(candidate)
+    db.session.commit()
+    current_app.logger.info("Renamed sanitised fixture service '%s' to '%s'", old_name, candidate.name)
+
+
+def _clear_sanitised_fixture_caches():
+    # Sanitisation rotates canonical fixture entities; clear short-lived memoised lookups so they repopulate cleanly.
+    reset_memos()
```
whpearson, line 195: As far as I can tell memoised lookups are only for clients, not for database entities

whpearson (approved):
Makes sense

PR #4837, opened 2026-04-30
`Notification` model: replace "failed" partial index with "nondelivered" partial index

risicle:
`delivered` notifications make up over 95% of the `notifications` table. When we need to find notifications of *any* of the other statuses, this index will allow the query planner to scan in in `created_at` order, skipping the vast majority of the rows, provided we're also filtering by `service_id` and `notification_type`.

Judging from the current contents of the `notifications` table, this should barely add 20% to the size of the index compared to the existing "failed statuses" partial index.

The original motivation for this is speeding up `get_notifications` queries for the `sent` status.

For this PR, the "migration isolation" check can be ignored because the changes made to the app code don't actually have any runtime effect (the `Index` objects are for documentation as much as anything - the definitions of the indexes are managed entirely by the migrations.

We also ignore the "squawk" check - it always throws up these errors for `CREATE INDEX CONCURRENTLY` and we've decided this is the way we want to handle this sort of index addition.

Successfully tested on a dev env that this index is indeed still usable for the "failed notifications" queries it was originally created for.

### DIFF
--- a/app/constants.py
+++ b/app/constants.py
@@ -18,7 +18,6 @@ NOTIFICATION_RETURNED_LETTER = "returned-letter"
 
 # Raw notification status values grouped into types
 NOTIFICATION_STATUS_TYPES_FAILED = [
-    # if changing this group, update ix_notifications_failed_service_id_composite to match
     NOTIFICATION_TECHNICAL_FAILURE,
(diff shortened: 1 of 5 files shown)


CrystalPea (approved):
LGTM 👍🏼

PR #4838, opened 2026-04-30
Include sanitised content info in send_email V2 endpoint api response

CrystalPea:
When a request to send an email has sanitise_content_for arg, we check the content of specified placeholders for potentially malicious content.

During the user research, service users requested to be notified when this is activated, so that they could investigate.

This PR sends the information about the content we changed, and how it looks before and after.

The information is included in the response. "sanitised_content" key is included in the response for all send_email requests (so it's less error-prone for our service users), but not for send_sms and send_letter requests.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -128,8 +128,17 @@ def post_notification(notification_type):
 
     check_service_has_permission(authenticated_service, notification_type)
 
-    personalisation = _prepare_personalisation_for_post_notification(
-        personalisation=form.get("personalisation", {}), sanitise_content_for=form.get("sanitise_content_for", [])
+    unsanitised_personalisation = form.get("personalisation", {})
+    personalisation = (
+        _prepare_personalisation_for_post_notification(
+            personalisation=unsanitised_personalisation, sanitise_content_for=form.get("sanitise_content_for", [])
+        )
+        if notification_type == "email"
+        else unsanitised_personalisation
+    )
+
+    sanitised_content = (
+        _get_sanitised_content(unsanitised_personalisation, personalisation) if notification_type == "email" else {}
     )
 
     template, template_with_content = validate_template(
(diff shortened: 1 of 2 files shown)


app/v2/notifications/post_notifications.py
```diff
@@ -128,8 +128,17 @@ def post_notification(notification_type):
...
     check_service_has_permission(authenticated_service, notification_type)
 
-    personalisation = _prepare_personalisation_for_post_notification(
-        personalisation=form.get("personalisation", {}), sanitise_content_for=form.get("sanitise_content_for", [])
+    unsanitised_personalisation = form.get("personalisation", {})
+    personalisation = (
+        _prepare_personalisation_for_post_notification(
+            personalisation=unsanitised_personalisation, sanitise_content_for=form.get("sanitise_content_for", [])
+        )
+        if notification_type == "email"
```
joybytes, line 136: is there a specific reason that we are only keeping this for emails?
Before it seems like it was running for all notification types

CrystalPea: The "Sanitise placeholder content" feature we are developing is currently email-specific. We will enable it for send_email endpoints in the clients, and it will be explained in "Send email message" section in the API docs.

I think it makes most sense for emails, as it prevents Markdown injection, and for SMS we do not do markdown (SMS is plain text), and letters don't have clickable links.

I wrote the previous PR on this, and not validating that it's for email only was an omission that I spotted now. I should have done a better job explaining it from the get go 😳 

Still, if we would like to offer removing links from selected placeholders for sms and letter, I'm not against it - but it's something we should decide as a team.

joybytes: I see, that makes a lot of sense. Thanks for the context.

joybytes (approved):

CrystalPea:
Just to add - I have tested this locally by making following API requests with Python client, modified to know about `sanitise_content_for` arg:
- `send_email_notification` with a placeholder to sanitise
- `send_email_notification` without a placeholder to sanitise
- `send_sms_notification`

It worked as expected on all counts, and the response was as expected.

PR #4845, opened 2026-05-08
sqlalchemy: set `connect_timeout` to 4 seconds

risicle:
We're currently seeing some unusual 5 second waits in api-web that aren't yielding the greenthread. If it's _this_ 5s timeout that's responsible, changing it to 4s should cause us to start seeing 4s waits instead.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -182,7 +182,7 @@ class Config:
         "pool_timeout": 30,
         "pool_recycle": 300,
         "connect_args": {
-            "connect_timeout": "5",  # seconds
+            "connect_timeout": "4",  # seconds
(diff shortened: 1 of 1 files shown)


CrystalPea (commented):
Are we changing the timeout as a performance improvement, or as a part of an investigation, or both?

robinjam:
> Are we changing the timeout as a performance improvement, or as a part of an investigation, or both?

Just part of an investigation I think

robinjam (approved):

risicle:
Looks like we're still getting the 5s pauses, so this probably wasn't the culprit. Will revert once we've captured a few more examples.

PR #4848, opened 2026-05-12
Adding two protected_sender_ids to fixtures

whpearson:
This PR adds two protected_sender_ids to the fixtures.

They are mainly to be used in game days.

Adding real protected sender ids is hard because we don't want to make them public.

### DIFF
--- a/app/functional_tests_fixtures/__init__.py
+++ b/app/functional_tests_fixtures/__init__.py
@@ -6,6 +6,7 @@ import boto3
 from flask import current_app
 from sqlalchemy.exc import NoResultFound
 
+from app import db
 from app.constants import (
     EDIT_FOLDER_PERMISSIONS,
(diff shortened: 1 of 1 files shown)


app/functional_tests_fixtures/__init__.py
```diff
@@ -406,6 +408,9 @@ def _create_db_objects(
     current_app.logger.info("--> Ensure dummy inbound SMS objects exist")
     _create_inbound_sms(service, 3)
 
+    _create_protected_sender_id("TEAM_DOVE")
```
klssmith, line 412: Optional: You could add a log line before adding the protected sender IDs to say what is being done, to match the other log lines of  `--> Ensure ... exist`

klssmith (approved):

klssmith (approved):

PR #4856, opened 2026-05-20
`check_for_services_with_high_failure_rates_or_sending_to_tv_numbers`: use `session_bulk`

risicle:
Looks like we forgot to actually specify the db session to use the replica back in #4731 when we made these queries bulk-capable.


klssmith (approved):
Whoops - good spot 😅

PR #4862, opened 2026-05-22
`get_all_templates_for_service`: make serialization interruptible

risicle:
Depending on https://github.com/alphagov/notifications-utils/pull/1356

Marshmallow serialization is not particularly fast, for large template lists we can spend quite a while doing this, so don't make this block async.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-# This file was automatically copied from notifications-utils@115.3.1
+# This file was automatically copied from notifications-utils@115.5.0
 
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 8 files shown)


quis:
Given we only serialize these fields…
https://github.com/alphagov/notifications-api/blob/77cc1e385b8a2d3889ac4837d06ce5b3c545f40e/app/schemas.py#L442-L448

…we could probably drop Marshmallow entirely here. Lots of precedence for adding something like a `serialize_no_detail` method on the `Template` model. For example this from the `Service` model:
https://github.com/alphagov/notifications-api/blob/77cc1e385b8a2d3889ac4837d06ce5b3c545f40e/app/models.py#L701-L707

risicle:
🤷 I wouldn't dare weigh in on such a matter

quis:
- [x] https://github.com/alphagov/notifications-api/pull/4864

***

Happy to rebase on top of this PR if you want to get it in first

risicle:
If yours does the job this hack won't be needed.

risicle:
Does the job.

PR #4863, opened 2026-05-26
`process_ses_results`: extract more bounce data into structured log fields

risicle:
Should make it easier to analyze our bounce reports in kibana.

### DIFF
--- a/app/celery/process_ses_receipts_tasks.py
+++ b/app/celery/process_ses_receipts_tasks.py
@@ -92,6 +92,7 @@ def process_ses_results(  # noqa: C901
             return
 
         if bounce_message:
+            bounce_message_bounce = bounce_message.get("bounce") or {}
             current_app.logger.info(
                 "SES bounce for notification ID %s",
                 notification.id,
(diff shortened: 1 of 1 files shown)


app/celery/process_ses_receipts_tasks.py
```diff
@@ -100,6 +101,9 @@ def process_ses_results(  # noqa: C901
                     "bounce_message": json.dumps(bounce_message),
                     "bounced_at": delivery_dt,
                     "bounced_ago": (uniform_now - delivery_dt).total_seconds() if delivery_dt is not None else None,
+                    "bounce_message_type": bounce_message_bounce.get("bounceType"),
+                    "bounce_message_sub_type": bounce_message_bounce.get("bounceSubType"),
+                    "bounce_message_remote_mta_ip": bounce_message_bounce.get("remoteMtaIp"),
```
WesleyHindle, line 106: Would it be beneficial to have `reportingMTA` as well as / instead of `remoteMtaIp` so it's more explicit which MTA attempted delivery?

risicle: I thought it looked like quite noisy data, given most the time it's one of SES's servers. And sometimes the strings can be big. Not having an immediate use for it, I didn't include it.

WesleyHindle (approved):

PR #4864, opened 2026-05-26
Don’t use Marshmallow to serialize all templates

quis:
There can be a lot of templates, this can take time, we can remove a bit of overhead by not using Marshmallow.

Benchmarked with 1,000 templates.

# Before

Screenshot 2026-05-26 at 11 24 18

# After

Screenshot 2026-05-26 at 11 17 47

7&times; faster

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -105,6 +105,7 @@ from app.models_types import (
     SerializedServiceSmsSender,
     SerializedTemplateEmailFile,
     SerializedTemplateFolder,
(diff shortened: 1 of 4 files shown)


risicle:
Need to ruff it I think.

risicle (approved):

quis:
Looks like this change (abea0fd, dark purple) has brought down the p95 response time by at least 67%:

image

PR #4867, opened 2026-05-28
`process_ses_results`, `process_sms_client_response`: supply receipt timestamp to `send_delivery_status_to_service`

risicle:
Second half of https://trello.com/c/kof0w1IT/1529-record-delivery-receipt-callback-latency-as-a-metric, which sits on top of #4868 (so review that first).


robinjam (approved):
Shout if you need a re-approval after rebasing this

emmetdev (approved):

PR #4868, opened 2026-05-29
`send_delivery_status_to_service`: accept `receipt_iso_timestamp`

risicle:
And use for `service_callback_forward_duration` metric.

This is the first half of https://trello.com/c/kof0w1IT/1529-record-delivery-receipt-callback-latency-as-a-metric, which needs to be deployed before the second half goes out because we need to be able to accept the new `receipt_iso_timestamp` kwarg in `send_delivery_status_to_service` before any sending code starts supplying it.

Note how this is recording the latency *excluding* the callback time itself, because we note the `start_dt` just before we begin the request. It will also only emit a metric for successful callbacks. It's likely we'll want to filter the resultant metric for `callback.attempt == 0` in any SLOs so we're not counting retry wait-time, but I do expect information invlving retries will be useful in the future so I'm not excluding retried callbacks entirely.

Note also how I've designed this as a general service callback metric, intending to use it for inbound sms latency too (suitably labeled).

Have tested this along with its other half #4867 and its `-aws` requisite https://github.com/alphagov/notifications-aws/pull/2993 on dev-f.

### DIFF
--- a/app/celery/service_callback_tasks.py
+++ b/app/celery/service_callback_tasks.py
@@ -1,6 +1,7 @@
 import json
 import logging
 from contextvars import ContextVar
+from datetime import datetime
 
 import requests
 from flask import current_app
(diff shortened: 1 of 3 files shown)


robinjam (approved):

risicle:
Ok, this new version I've pushed records a metric both for successful and unsuccessful attempts. Unsuccessful attempts have an extra `error.type` attribute.

emmetdev (approved):

PR #4871, opened 2026-06-02
Alert on letters in technical-failure status via Zendesk

spatel033:
Letters stuck in `technical-failure` can be silently left unresolved. During a recent incident, letters in technical-failure that were unrelated to the incident added noise and made triage harder. By alerting on this state, we ensure every letter in technical-failure is either fixed or explicitly moved to a different failure state. [Card](https://trello.com/c/EdTYjeJU/1819-create-a-zendesk-alert-for-letters-that-go-into-technical-failure) for more details.
Test zendesk ticket on the card.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -59,6 +59,7 @@ from app.dao.jobs_dao import (
 )
 from app.dao.notifications_dao import (
     SlowProviderDeliveryReport,
+    dao_letters_in_technical_failure,
     dao_old_letters_with_created_status,
     dao_precompiled_letters_still_pending_virus_check,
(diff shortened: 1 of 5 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -429,6 +430,36 @@ def check_if_letters_still_in_created():
             )
 
 
+@notify_celery.task(name="check-if-letters-in-technical-failure")
+def check_if_letters_in_technical_failure():
+    technical_failed_letters = dao_letters_in_technical_failure(session=db.session_bulk)
```
klssmith: Did you want to add `retry_attempts` in here just for consistency with the other functions with use the replica?

klssmith (approved):

PR #4875, opened 2026-06-08
Add celery beat task, `archive_pending_files` to automatically archives files that have been created but not made live after 24 hours.

rparke:
This is vital cleanup work that ensures that these stale files don't persist and are properly cleaned up, eventually being removed from S3 too when they are picked up by the scheduled task: `remove-archived-template-email-files-from-s3`.

24 hours has been selected because that's the session timeout for our users and should be long enough that if they were going to make the file live, they'd have done it.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -76,7 +76,7 @@ from app.dao.services_dao import (
     dao_find_services_sending_to_tv_numbers,
     dao_find_services_with_high_failure_rates,
 )
-from app.dao.template_email_files_dao import dao_get_template_email_files_by_template_id
+from app.dao.template_email_files_dao import dao_archive_pending_files, dao_get_template_email_files_by_template_id
 from app.dao.templates_dao import dao_get_template_by_id
 from app.dao.users_dao import delete_codes_older_created_more_than_a_day_ago, get_users_for_research
 from app.letters.utils import generate_letter_pdf_filename
@@ -106,6 +106,19 @@ def run_scheduled_jobs():
             current_app.logger.info("Job ID %s added to process job queue", job.id, extra={"job_id": job.id})
     except SQLAlchemyError:
         current_app.logger.exception("Failed to run scheduled jobs")
+        raise
+
+
+@notify_celery.task(name="archive-pending-files")
+def archive_pending_files():
+    try:
+        number_of_files_archived = dao_archive_pending_files()
+        current_app.logger.info(
+            "Archived %(number_of_files_archived)s files created more than 24 hours ago that are still in pending",
+            extra={"number_of_files_archived": number_of_files_archived},
+        )
+    except SQLAlchemyError:
+        current_app.logger.exception("Failed to archive pending files")
         raise
 
 
--- a/app/config.py
+++ b/app/config.py
@@ -231,6 +231,8 @@ class Config:
     MAX_VERIFY_CODE_COUNT = 5
     MAX_FAILED_LOGIN_COUNT = 10
 
+    TEMPLATE_EMAIL_FILE_ARCHIVE_PERIOD_IN_HOURS = 24
+
     # these should always add up to 100%
     SMS_PROVIDER_RESTING_POINTS = {"mmg": 51, "firetext": 49}
 
@@ -483,6 +485,11 @@ class Config:
             "weekly-user-research-email": {
                 "task": "weekly-user-research-email",
                 "schedule": crontab(hour=10, minute=0, day_of_week="wed"),
+                "options": {"queue": QueueNames.PERIODIC},
+            },
+            "archive-pending-files": {
+                "task": "archive-pending-files",
+                "schedule": crontab(hour=3, minute=33),
                 "options": {"queue": QueueNames.PERIODIC},
             },
             # first tuesday of every month
--- a/app/dao/template_email_files_dao.py
+++ b/app/dao/template_email_files_dao.py
@@ -137,9 +137,22 @@ def dao_make_pending_template_email_file_live(template_email_file: TemplateEmail
 @version_class(
     VersionOptions(TemplateEmailFile, history_class=TemplateEmailFileHistory),
 )
-def dao_archive_template_email_file(file_to_archive, archived_by_id, template_version):
+def dao_archive_template_email_file(file_to_archive, archived_by_id, template_version=None):
+    if not template_version:
+        template_version = Template.query.get(file_to_archive.template_id).version
     if not file_to_archive.archived_at:
         file_to_archive.archived_at = datetime.datetime.utcnow()
         file_to_archive.archived_by_id = archived_by_id
         file_to_archive.template_version = template_version
         db.session.add(file_to_archive)
+
+
+def dao_archive_pending_files():
+    files_in_pending = TemplateEmailFile.query.filter(
+        TemplateEmailFile.pending,
+        datetime.datetime.utcnow() - TemplateEmailFile.created_at
+        > datetime.timedelta(hours=current_app.config.get("TEMPLATE_EMAIL_FILE_ARCHIVE_PERIOD_IN_HOURS")),
+    ).all()
+    for file in files_in_pending:
+        dao_archive_template_email_file(file, archived_by_id=file.created_by_id)
+    return len(files_in_pending)
--- a/tests/app/celery/test_scheduled_tasks.py
+++ b/tests/app/celery/test_scheduled_tasks.py
@@ -22,6 +22,7 @@ from app.celery.letters_pdf_tasks import get_pdf_for_templated_letter
 from app.celery.provider_tasks import deliver_email, deliver_sms
 from app.celery.scheduled_tasks import (
     _check_slow_text_message_delivery_reports_and_raise_error_if_needed,
+    archive_pending_files,
     change_dvla_api_key,
     change_dvla_password,
     check_for_low_available_inbound_sms_numbers,
@@ -64,6 +65,7 @@ from app.dao.annual_billing_dao import set_default_free_allowance_for_service
 from app.dao.jobs_dao import dao_get_job_by_id
 from app.dao.notifications_dao import SlowProviderDeliveryReport
 from app.dao.provider_details_dao import get_provider_details_by_identifier
+from app.dao.template_email_files_dao import dao_get_template_email_file_by_id
 from app.models import Event, InboundNumber, Notification
 from tests.app import load_example_csv
 from tests.app.db import (
@@ -72,6 +74,7 @@ from tests.app.db import (
     create_notification,
     create_organisation,
     create_template,
+    create_template_email_file,
     create_user,
 )
 from tests.conftest import _with_message_group_id, set_config, set_config_values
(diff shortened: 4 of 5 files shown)


app/config.py
```diff
@@ -231,6 +231,8 @@ class Config:
     MAX_VERIFY_CODE_COUNT = 5
     MAX_FAILED_LOGIN_COUNT = 10
 
+    TEMPLATE_EMAIL_FILE_ARCHIVE_PERIOD_IN_HOURS = int(os.getenv("TEMPLATE_EMAIL_FILE_ARCHIVE_PERIOD_IN_HOURS", 24))
```
rparke: I'm not sure yet whether I want to set this at the terraform level or not. 24 is probably okay for each environment so we can just make it 

`TEMPLATE_EMAIL_FILE_ARCHIVE_PERIOD_IN_HOURS = 24`

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -106,6 +109,34 @@ def test_should_update_scheduled_jobs_and_put_on_queue(mock_celery_task, sample_
     )
 
 
+@pytest.mark.parametrize("pending_expirary_exceeded", (True, False))
+def test_archive_pending_files_only_archives_pending_file_after_auto_archive_period(
```
rparke: any ideas for a better test name?

app/dao/template_email_files_dao.py
```diff
@@ -143,3 +143,14 @@ def dao_archive_template_email_file(file_to_archive, archived_by_id, template_ve
...
+
+def dao_archive_pending_files():
+    files_in_pending = TemplateEmailFile.query.filter(
+        TemplateEmailFile.pending,
+        datetime.datetime.utcnow() - TemplateEmailFile.created_at
+        > datetime.timedelta(hours=current_app.config.get("TEMPLATE_EMAIL_FILE_ARCHIVE_PERIOD_IN_HOURS")),
+    ).all()
+    for file in files_in_pending:
+        template = Template.query.get(file.template_id)
+        dao_archive_template_email_file(file, archived_by_id=file.created_by_id, template_version=template.version)
```
rparke: `template_email_files is a relatively small table because 1 file --> many notifications. It will always be strictly smaller than the templates table. We could future proof this with some pagination, but it's arguably code that may never run.

app/dao/template_email_files_dao.py
```diff
@@ -143,3 +143,14 @@ def dao_archive_template_email_file(file_to_archive, archived_by_id, template_ve
...
+
+
+def dao_archive_pending_files():
+    files_in_pending = TemplateEmailFile.query.filter(
+        TemplateEmailFile.pending,
+        datetime.datetime.utcnow() - TemplateEmailFile.created_at
+        > datetime.timedelta(hours=current_app.config.get("TEMPLATE_EMAIL_FILE_ARCHIVE_PERIOD_IN_HOURS")),
+    ).all()
+    for file in files_in_pending:
+        template = Template.query.get(file.template_id)
```
rparke: To consider - this method is not wrapped up as a single transaction. Only the file archive step is. The templates version _could_ shift during before it gets to archiving the file (unlikely, edge case and doesn't hugely matter - the template version is just a required field for the archive method and these files were never actually associated with a specific template version).

app/config.py
CrystalPea: I agree, we can have it hardcoded in config file

tests/app/celery/test_scheduled_tasks.py
CrystalPea: maybe like this? `test_archive_pending_files_only_scopes_stale_files`

app/dao/template_email_files_dao.py
CrystalPea: It could theoretically grow bigger than templates, as one template can have many email files, and also we keep archived files in the table forever (and that includes former pending files, meaning it could grow faster).

Still, in practice I also don't think it will grow that big in foreseeable future. And since we run the task every day, there shouldn't ever be enough pending files created in the last 1-2 days to require pagination.

app/celery/scheduled_tasks.py
```diff
@@ -109,6 +109,15 @@ def run_scheduled_jobs():
         raise
 
 
+@notify_celery.task(name="archive-pending-files")
+def archive_pending_files():
+    try:
+        dao_archive_pending_files()
+        current_app.logger.info("Archiving files created more than 24 hours ago that are still in pending")
```
CrystalPea: would we maybe want the dao method to return how many files it archived, and then to include that in the log?

CrystalPea: This could be in past tense as we already did the archiving at that point.

app/dao/template_email_files_dao.py
```diff
@@ -143,3 +143,14 @@ def dao_archive_template_email_file(file_to_archive, archived_by_id, template_ve
...
         db.session.add(file_to_archive)
+
+
+def dao_archive_pending_files():
+    files_in_pending = TemplateEmailFile.query.filter(
+        TemplateEmailFile.pending,
+        datetime.datetime.utcnow() - TemplateEmailFile.created_at
+        > datetime.timedelta(hours=current_app.config.get("TEMPLATE_EMAIL_FILE_ARCHIVE_PERIOD_IN_HOURS")),
+    ).all()
+    for file in files_in_pending:
```
CrystalPea, line 156: Could we archive all pending files in a single transaction, instead of looping? That will make the query much much faster.

app/config.py
```diff
@@ -485,6 +487,11 @@ class Config:
                 "schedule": crontab(hour=10, minute=0, day_of_week="wed"),
                 "options": {"queue": QueueNames.PERIODIC},
             },
+            "archive-pending-files": {
+                "task": "archive-pending-files",
+                "schedule": crontab(hour=3, minute=33),
```
CrystalPea, line 492: Good idea to do this in the middle of the night - not only because it is a quiet period, but also because it means in practice almost all files will be a bit older than 24 hours, and that's a good safety net.

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -106,6 +109,34 @@ def test_should_update_scheduled_jobs_and_put_on_queue(mock_celery_task, sample_
...
+    sample_email_template, pending_expirary_exceeded
+):
+    with freeze_time("2016-01-01 01:00:00.000000"):
```
CrystalPea, line 114: wow, this date probably goes back to when Notify was first developed :o

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -106,6 +109,34 @@ def test_should_update_scheduled_jobs_and_put_on_queue(mock_celery_task, sample_
...
+            created_at=datetime.utcnow(),
+        )
+        live_file = create_template_email_file(
+            template_id=sample_email_template.id,
+            created_by_id=sample_email_template.created_by_id,
+            pending=False,
+            created_at=datetime.utcnow(),
+        )
+    with freeze_time("2016-01-02 01:00:01.00000" if pending_expirary_exceeded else "2016-01-02 00:00:00.00000"):
+        archive_pending_files()
```
CrystalPea, line 128: we could also assert the logline is triggered successfully n this test, then we know if it gets broken in the future.

CrystalPea (commented):
Nice work! I left a few comments 🙌🏼

app/celery/scheduled_tasks.py
rparke: can do

app/dao/template_email_files_dao.py
rparke: We can leave it for now.

app/dao/template_email_files_dao.py
rparke: I decided that putting this in a single transaction is a small change and is probably better anyway.

tests/app/celery/test_scheduled_tasks.py
rparke: it's used elsewhere as a datetime stamp in our code 😆. Keeps it somewhat consistent.

CrystalPea: yeah, more of observation than suggestion - I usually use a somewhat recent date, but it doesn't really make any difference 😄 . Just pondering passage of time 😝

app/dao/template_email_files_dao.py
rparke: I don't know how much faster this will make the query because we don't get thousands of new files a day (at the moment) can update if needed.

app/celery/scheduled_tasks.py
```diff
@@ -109,6 +109,18 @@ def run_scheduled_jobs():
         raise
 
 
+@notify_celery.task(name="archive-pending-files")
+def archive_pending_files():
+    try:
+        number_of_files_archived = dao_archive_pending_files()
+        current_app.logger.info(
+            "Archived %(number_of_files_archived) files created more than 24 hours ago that are still in pending",
```
joybytes: I think the formatting for this log is wrong, you need an `s` and the end

`%(number_of_files_archived)` and you're missing parms for it 

check out `delete_verify_codes()` below

tests/app/db.py
```diff
@@ -212,6 +213,7 @@ def create_template_email_file(
         "validate_users_email": validate_users_email,
         "template_id": template_id,
         "created_by_id": created_by_id,
+        "created_at": created_at if not None else datetime.utcnow(),
```
joybytes:
```suggestion
        "created_at": created_at if created_at is not None else datetime.utcnow(),
```

otherwise it's always None

app/celery/scheduled_tasks.py
```diff
@@ -109,6 +109,18 @@ def run_scheduled_jobs():
...
+            extra={"number_of_files_archived": number_of_files_archived},
+        )
+    except SQLAlchemyError:
+        current_app.logger.exception("Failed to archive pending files")
+
```
joybytes, line 123: should we add `raise` here like other functions (delete_verify_codes)?
otherwise celery will mark this as successful even if DB operations fail

tests/app/db.py
joybytes: https://www.onlineide.pro/playground/share/e10854d1-454d-40e4-bfa6-13529729ef35

app/dao/template_email_files_dao.py
```diff
@@ -137,9 +137,22 @@ def dao_make_pending_template_email_file_live(template_email_file: TemplateEmail
...
         file_to_archive.archived_at = datetime.datetime.utcnow()
         file_to_archive.archived_by_id = archived_by_id
         file_to_archive.template_version = template_version
         db.session.add(file_to_archive)
+
+
+def dao_archive_pending_files():
+    files_in_pending = TemplateEmailFile.query.filter(
+        TemplateEmailFile.pending,
+        datetime.datetime.utcnow() - TemplateEmailFile.created_at
```
joybytes, line 153: nit-pick and non-blocking:
consider flipping the check

instead of : `utcnow() - created_at > timedelta(hours=...)`

use `created_at < utcnow() - timedelta(hours=...)`

I've noticed that in our repo we tend to use the second pattern
few examples:
`delete_codes_older_created_more_than_a_day_ago()`
`delete_invitations_created_more_than_two_days_ago()`

app/dao/template_email_files_dao.py
```diff
@@ -137,9 +137,22 @@ def dao_make_pending_template_email_file_live(template_email_file: TemplateEmail
...
+        > datetime.timedelta(hours=current_app.config.get("TEMPLATE_EMAIL_FILE_ARCHIVE_PERIOD_IN_HOURS")),
+    ).all()
+    for file in files_in_pending:
+        dao_archive_template_email_file(file, archived_by_id=file.created_by_id)
+    return len(files_in_pending)
```
joybytes, line 158: just checking, this return value, is the number of values matched and not the number of files actually archived. Is this expected?

tests/app/db.py
rparke: good spot!

app/celery/scheduled_tasks.py
rparke: yes we should!

app/dao/template_email_files_dao.py
joybytes: also just a quick question about this query

the filter uses > rather than >=, so files exactly 24 hours old won't be archived yet. Is that intentional?

rparke: That's intentional to have it run for files that are strictly older than 24 hours ... it doesn't hugely matter but strict inequalities are just easier to talk about. "We archive anything older than a day" vs "we archive anything that is at least 24 hours old". In practice we will never see a value excluded/included by use of a strict inequality.

app/dao/template_email_files_dao.py
rparke: It's an inference based on the assumption that `dao_archive_template_email_file` doesn't fail. If it does fail and raises a SQLAlchemy exception we escape this loop anyway.

We could write handling logic that goes "okay this is how many you have managed to archive before I stopped", but I don't know if we need to?

app/celery/scheduled_tasks.py
rparke: fixed

app/dao/template_email_files_dao.py
joybytes: looking at it again, there is no need for it
the current return is fine

joybytes (approved):

PR #4876, opened 2026-06-08
Remove statsd_mapping.yml

robinjam:
https://trello.com/c/25A3gpXh/1358-replace-statsd-instrumentation-across-notify-with-otel

I'm not entirely sure why it's here to begin with, but we don't use StatsD anymore so I think it's safe to say it's not needed.


quis (approved):
Only thing left in this repo when [searching for ’statsd’](https://github.com/search?q=repo%3Aalphagov%2Fnotifications-api%20statsd&type=code)

PR #4877, opened 2026-06-11
Fix excessive calls to dao get organisation domains

rparke:
Remove calls to `dao_get_organisation_domains()` from for loops where possible.

These calls to `dao_get_organisation_domains()` are redundant (but also crucially aren't cached). For organisations with a large number of team members in their services, this has resulted in timeouts when they have tried to do things like updating and deleting a team member from their service.

By only calling this method once before the loop/list comprehension, it should be much more efficient.

### DIFF
--- a/app/dao/users_dao.py
+++ b/app/dao/users_dao.py
@@ -199,7 +199,10 @@ def dao_archive_user(user):
 
 
 def _count_gov_users_with_manage_settings(service):
-    active_government_users = [u for u in service.users if u.state == "active" and is_gov_user(u.email_address)]
+    organisation_domains = dao_get_organisation_domains()
+    active_government_users = [
+        u for u in service.users if u.state == "active" and is_gov_user(u.email_address, organisation_domains)
+    ]
 
     return sum(MANAGE_SETTINGS in u.get_permissions(service_id=service.id) for u in active_government_users)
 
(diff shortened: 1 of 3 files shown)


joybytes (approved):

app/dao/users_dao.py
```diff
@@ -227,13 +230,15 @@ def users_permissions_can_be_changed(user, service, new_permissions):
...
-    if not is_gov_user(user.email_address):
+    if not is_gov_user(user.email_address, dao_get_organisation_domains()):
         return True
 
     return _can_remove_gov_user_with_manage_settings(service)
 
 
-def user_can_be_removed_from_service(user, service):
+def user_can_be_removed_from_service(user, service, organisation_domains=None):
+    if not organisation_domains:
```
ipdigital24, line 240: do we need this check here?

ipdigital24: ah. i see this fn being called from [DELETE] remove_user_from_service.

may be, in that case, could we make both user_can_be_archived() and remove_user_from_service() behave in similar way?

rparke: The check means that we dynamically create an instance (not in a loop) if organisation_domains isn't set.

You can't set a function call as a default value. If we don't use the default value everywhere we call this though we could just remove the default.

ipdigital24: understood. thanks

ipdigital24 (approved):

PR #4887, opened 2026-06-19
Add reason to provider change

whpearson:
This will allow us to store a reason for a provider ratio change.

So that we know why something was changed.

Changes the database/model and the rest interface.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0552_update_letter_pricing
+0553_add_reason_to_provider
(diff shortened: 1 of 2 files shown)


quis:
Doesn’t have to be part of this PR but it would be awesome if [`dao_reduce_sms_provider_priority`](https://github.com/alphagov/notifications-api/blob/a26c135d6ff55850c5fff4320696cf4069903e82/app/dao/provider_details_dao.py#L93) could take a `reason` argument as well. Then for all the automatic updates we’d know why the system decided to make a change.

spatel033 (approved):

PR #4889, opened 2026-06-22
feat: add migration for updated_by_id (revoker)

joybytes:
### Summary

- Added `updated_by_id` field on API Keys and API Keys History table
- Added index for API Key table for this field, but not for the History table as that table is only required by `service_id`

Ticket: https://trello.com/c/BwLKBEc4/645-api-key-history-shows-incorrect-data

### DIFF
--- a/migrations/versions/0552_api_keys_updated_by_id.py
+++ b/migrations/versions/0552_api_keys_updated_by_id.py
@@ -1,0 +1,28 @@
+"""
+Revision ID: 0552_api_keys_updated_by_id
+Revises: 0551_drop_ntfcns_failed_idx
+Create Date: 2026-06-22
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "0552_api_keys_updated_by_id"
+down_revision = "0551_drop_ntfcns_failed_idx"
+
+
+def upgrade():
+    op.add_column("api_keys", sa.Column("updated_by_id", postgresql.UUID(as_uuid=True), nullable=True))
+    op.create_index(op.f("ix_api_keys_updated_by_id"), "api_keys", ["updated_by_id"], unique=False)
+    op.create_foreign_key("fk_api_keys_updated_by_id", "api_keys", "users", ["updated_by_id"], ["id"])
+
+    op.add_column("api_keys_history", sa.Column("updated_by_id", postgresql.UUID(as_uuid=True), nullable=True))
+
+
+def downgrade():
+    op.drop_column("api_keys_history", "updated_by_id")
+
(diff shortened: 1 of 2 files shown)


klssmith:
This is more a question about the general approach than this particular PR. With the other versioned tables (services and templates) we can see who changed or archived the object without needing an extra column to store the ID of the person. The `created_by_id` is not the person who created the service or template originally, it's the person who created "that version" of it. So `created_by_id` is storing `updated_by_id`.  If we were to update the `created_by_id` column when changing an API key, like we do for the other versioned models, this would work the same way as them. 

Wondering if that existing approach was considered vs this new way? I can see the pros and cons for both - for example, the existing way is consistent with how the rest of our code works, but it's less obvious what's going on

migrations/versions/0552_api_keys_updated_by_id.py
```diff
@@ -0,0 +1,28 @@
...
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "0552_api_keys_updated_by_id"
+down_revision = "0551_drop_ntfcns_failed_idx"
+
+
+def upgrade():
+    op.add_column("api_keys", sa.Column("updated_by_id", postgresql.UUID(as_uuid=True), nullable=True))
+    op.create_index(op.f("ix_api_keys_updated_by_id"), "api_keys", ["updated_by_id"], unique=False)
```
BlessedDev, line 17: When creating the index, it locks the table against writes locking the table, the warning in the squawk linter is to create the index concurrently so the index is built in the background without blocking other operations on the table.

We have a pattern here:
https://github.com/alphagov/notifications-api/blob/a26c135d6ff55850c5fff4320696cf4069903e82/migrations/versions/0468.py#L15

use  `with op.get_context().autocommit_block()` to wrap the transactions with `postgresql_concurrently=True,` within the index block.

joybytes:
> This is more a question about the general approach than this particular PR. With the other versioned tables (services and templates) we can see who changed or archived the object without needing an extra column to store the ID of the person. The `created_by_id` is not the person who created the service or template originally, it's the person who created "that version" of it. So `created_by_id` is storing `updated_by_id`. If we were to update the `created_by_id` column when changing an API key, like we do for the other versioned models, this would work the same way as them.
> 
> Wondering if that existing approach was considered vs this new way? I can see the pros and cons for both - for example, the existing way is consistent with how the rest of our code works, but it's less obvious what's going on

@klssmith  This comment actually answers something I saw when doing this story.
I did wonder why the existing API on admin was propagating the field for revoking the key as  `created_by`

From what I see, it seems like the intention would be to follow the same pattern as the others.
I actually did not realise that the api_keys table already had a version column in it. That being the case, to me it doesn't make sense to create another column `updated_by_id`

AND that means no migration is needed too 😄 

This is such a good catch ❤️

PR #4893, opened 2026-06-23
Add reason for provider change 2

whpearson:
Initial work on updating the model and the rest endpoints.

Also recording the reasons for the automated changes.

### DIFF
--- a/app/dao/provider_details_dao.py
+++ b/app/dao/provider_details_dao.py
@@ -189,6 +189,7 @@ def dao_get_provider_stats():
             ProviderDetails.priority,
             ProviderDetails.notification_type,
             ProviderDetails.active,
+            ProviderDetails.reason,
             ProviderDetails.updated_at,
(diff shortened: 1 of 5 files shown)


spatel033 (approved):

app/celery/scheduled_tasks.py
```diff
@@ -195,7 +195,9 @@ def switch_current_sms_provider_on_slow_delivery():
                     provider_name,
                     extra={"provider_name": provider_name},
                 )
-                dao_reduce_sms_provider_priority(provider_name, time_threshold=timedelta(minutes=5))
+                dao_reduce_sms_provider_priority(
```
klssmith: The `dao_reduce_sms_provider_priority` function doesn't have a parameter called `reason`, so calling the function with that argument gives an error

whpearson: Removed

klssmith (approved):

PR #4894, opened 2026-06-25
Delete archived letter attachments from s3

CrystalPea:
I used `remove_archived_template_email_files_from_s3()` task as a base for this new task.

I had to adjust it, as letter attachments are a bit different. I did some refactoring on the way, and de-duplicated the code shared between the tasks.

I have pushed these changes to dev-b, and functional tests pass: https://concourse.notify.tools/teams/dev-b/pipelines/deploy-notify/jobs/functional-tests/builds/520

I also did manual testing. 

I needed to make letter attachments S3 bucket available to api apps: https://github.com/alphagov/notifications-aws/pull/3045 (this needs to be merged / possibly terraform to be run first)

After that, the task worked as expected:
Screenshot 2026-07-09 at 16 08 40


app/config.py
```diff
@@ -565,6 +570,7 @@ class Config:
     S3_BUCKET_INVALID_PDF = os.environ.get("S3_BUCKET_INVALID_PDF")
     S3_BUCKET_TRANSIENT_UPLOADED_LETTERS = os.environ.get("S3_BUCKET_TRANSIENT_UPLOADED_LETTERS")
     S3_BUCKET_LETTER_SANITISE = os.environ.get("S3_BUCKET_LETTER_SANITISE")
+    S3_BUCKET_LETTER_ATTACHMENTS = os.environ.get("S3_BUCKET_LETTER_ATTACHMENTS", "local-letter-attachments")
```
klssmith, line 573: Can we follow the pattern of other buckets and have `os.environ.get("S3_BUCKET_LETTER_ATTACHMENTS"`) here, and then hardcode `development-letter-attachments` in the Development config and `test-letter-attachments` in the Test config?

It keeps it consistent, plus we don't have buckets with the prefix of `local-`, so there isn't a bucket called `local-letter-attachments`

app/celery/nightly_tasks.py
```diff
@@ -86,68 +87,111 @@ def _remove_csv_files(job_types):
...
+        older_than = _remove_batch_of_files_from_s3(
+            batch_of_files=archived_files_batch,
+            bucket_name=current_app.config["S3_BUCKET_TEMPLATE_EMAIL_FILES"],
+            location_prefix="",
         )
 
-    bucket_name = current_app.config["S3_BUCKET_TEMPLATE_EMAIL_FILES"]
+
+@notify_celery.task(name="remove-archived-letter-attachments-from-s3")
+@cronitor("remove-archived-letter-attachments-from-s3")
```
klssmith, line 114: I couldn't see a corresponding credentials PR, but we will need one when this is ready to be merged (https://github.com/alphagov/notifications-manuals/wiki/Monitoring-and-alerting#cronitor)

app/config.py
BlessedDev: The pattern for S3_BUCKET_TEMPLATE_EMAIL_FILES https://github.com/alphagov/notifications-api/blob/4d1273291b1f8b54a1d5e25c029c19dca600a44e/app/config.py#L561  is a good example

app/celery/nightly_tasks.py
```diff
@@ -86,68 +87,111 @@ def _remove_csv_files(job_types):
...
-            object_key = f"{service_id}/{template_email_file.id}"
 
-            try:
-                s3.remove_s3_object(bucket_name, object_key)
-                deleted += 1
-            except Exception:
-                failed += 1
+def _get_archived_before_and_after(archived_after):
+    archived_before = (
+        (datetime.now(UTC) - timedelta(days=ARCHIVED_FILE_RETENTION_DAYS)).astimezone(UTC).replace(tzinfo=None)
```
BlessedDev, line 142: Since `timedelta` is acting on datetime.now(UTC) which is already UTC aware, isn't `astimezone(UTC)` redundant here?

app/celery/nightly_tasks.py
```diff
@@ -86,68 +87,111 @@ def _remove_csv_files(job_types):
...
-        older_than = last_archived_file.id
+        archived_after = (
+            datetime.combine(parsed_date, datetime.min.time(), tzinfo=UTC).astimezone(UTC).replace(tzinfo=None)
+        )
+    else:
+        archived_after = archived_before - timedelta(days=ARCHIVED_FILE_DELETION_WINDOW_DAYS)
+
+    if archived_after > archived_before:
+        raise ValueError(
+            f"Invalid archived_after for s3 cleanup of archived files: "
```
BlessedDev, line 159: This aspect of the refactoring makes the error message less clear, it replaces ` "Invalid archived_after for archived template email file s3 cleanup: "` and `"Invalid archived_after for archived letter attachment s3 cleanup: "` with less specificity.

app/celery/nightly_tasks.py
BlessedDev: .ie would `timedelta(days=ARCHIVED_FILE_RETENTION_DAYS)).replace(tzinfo=None)` be sufficient

app/config.py
CrystalPea: sure thing, I will change it 🙌🏼

app/celery/nightly_tasks.py
CrystalPea: Ahh, good spot, I will follow the process in the manual 🙌🏼

app/celery/nightly_tasks.py
CrystalPea: I thought it wouldn't matter since this error message can only be seen when running the task manually - and then that person knows which task they are running.

BlessedDev: Ok. I see your point.

app/config.py
joybytes: https://github.com/alphagov/notifications-api/pull/4894/changes/b60e10b824c01c0cf33ffede9eef4fb307687c4d

app/celery/nightly_tasks.py
joybytes: https://github.com/alphagov/notifications-api/pull/4894/changes/92600ba71cf8fb65c17e2bdbaa245893e034c11b

app/celery/nightly_tasks.py
joybytes: created `GOV.UK Notify remove-archived-letter-attachments-from-s3` in cronitor
It's currently disabled and that should be changed after the first successful run

https://github.com/alphagov/notifications-credentials/pull/581

joybytes (approved):
Finished testing

I pushed some fixes and added the cronitor job.

Next steps are to:
1. merge credentials PR 
2. activate cronitor job after a first successful run 

I created a ticket to  track these and I don't mind following it through

Great job!

rparke (approved):

app/dao/letter_attachment_dao.py
```diff
@@ -0,0 +1,67 @@
...
+from app.dao.templates_dao import dao_update_template
+from app.models import LetterAttachment, TemplateHistory
+from app.utils import retryable_query
+
+
+def dao_archive_letter_attachment(letter_attachment, template, archived_by):
+    template.letter_attachment = None
+    letter_attachment.archived_at = datetime.datetime.utcnow()
+    letter_attachment.archived_by_id = archived_by
+    dao_update_template(template)
```
rparke, line 18: only `dao_update_template` is committed here

edit: actually I'm not sure if SQLAlchemy does something clever with relationships when you commit templates since letter_attachments are a declared attribute? This might be fine?

rparke:
**Please do not merge, there are a few issues where archiving won't be totally reliable.**

app/dao/letter_attachment_dao.py
joybytes: yes, the update happens for both letter_attachment and template

PR #4902, opened 2026-06-30
update letter pricing for 1st July

rparke:


BlessedDev (approved):

migrations/versions/0552_update_letter_pricing.py
```diff
@@ -0,0 +1,89 @@
...
+import uuid
+from datetime import datetime
+
+from alembic import op
+from sqlalchemy import text
+
+revision = "0552_update_letter_pricing"
+down_revision = "0551_drop_ntfcns_failed_idx"
+
+RATE_CHANGE_DATE = datetime(2026, 7, 1, 0, 0)
```
klssmith: We start new rates at midnight local time, but the database is in UTC so this needs be 2300 the day before

rparke: argh timezones!

migrations/versions/0552_update_letter_pricing.py
```diff
@@ -0,0 +1,89 @@
...
+            """
+            UPDATE
+                letter_rates
+            SET
+                end_date = :rate_change_date
+            WHERE
+                end_date IS NULL
+            """
+        ),
+        {"rate_change_date": RATE_CHANGE_DATE},
```
BlessedDev: This should be  `{ "end_date": RATE_CHANGE_DATE }`

rparke: rate_change_date is the postgresql parameter being passed in `:rate_change_date` on line 34.

BlessedDev: Yes, you are right. That was my mistake

BlessedDev: It seems an option is to change line 34 to `end_date = :end_date `and then line 39 to `{ "end_date": RATE_CHANGE_DATE } `to get past the linter

rparke: oh nvm, yes updating both for consistency.

quis (approved):

PR #4905, opened 2026-06-30
Add v2 returned letters endpoints

BlessedDev:


app/dao/returned_letters_dao.py
```diff
@@ -48,6 +50,11 @@ def insert_returned_letters(references):
         )
         db.session.connection().execute(stmt)
 
+    # clear the service-{service_id}-returned-letter-summary cache for each service
+
+    for service_id in service_ids:
```
quis: Use [`delete_by_pattern`](https://github.com/alphagov/notifications-utils/blob/3e28d8a75d0c429430faabdff9e477f85c7f7dc0/notifications_utils/clients/redis/redis_client.py#L162-L174) instead of looping over the list of services and potentially making hundreds of calls to Redis.

There’s an example pattern using a service ID here: https://github.com/alphagov/notifications-admin/blob/179706e1f8ca5bb28e36279e259f870c8026afef/app/notify_client/letter_jobs_client.py#L12

app/v2/returned_letters/get_returned_letters.py
```diff
@@ -0,0 +1,89 @@
...
+
+    return result
+
+
+@redis_cache.set("service-{service_id}-returned-letters-{report_date}", ttl_in_seconds=cache_expiration_time)
+def get_returned_letters_for_v2_endpoint(service_id: UUID | str, report_date: str) -> dict[Any]:
+    result = [
+        {
+            "notification_id": str(row["notification_id"]),
+            "reference": row["client_reference"],
```
quis, line 72: This is different to what we send in the callback:
https://github.com/alphagov/notifications-api/blob/ad0b987c97fc0b0788bc16058ad17626cb9b1f14/app/celery/service_callback_tasks.py#L36

We should make a `.serialize()` method to guarantee the callback and report are using the same schema and then we only have to document it once

BlessedDev: The callback data is pre processed here https://github.com/alphagov/notifications-api/blob/77f968030d160c61e326ffd9aec77311bdf486be/app/celery/service_callback_tasks.py#L248
So both the report and the callback data present "reference" to the end user instead of "client_reference".

I explored the `.serialize()` option initially but the data is collated from a few other tables as well so it is not a perfect fit.

I am thinking of a follow up PR to tie the callback and endpoints data together.

quis: The user receives a field called`reference` in both cases. The problem is that in one case the data is coming from the `client_reference` column in the database and in the other it’s coming from the `reference` column in the database. The former is the reference the user provides us, the latter is our internal reference.

BlessedDev: It looks like both `fetch_returned_letters` and `fetch_returned_letter_callback_data_dao` return the `client_reference` from the database:
https://github.com/alphagov/notifications-api/blob/77f968030d160c61e326ffd9aec77311bdf486be/app/dao/returned_letters_dao.py#L99
https://github.com/alphagov/notifications-api/blob/77f968030d160c61e326ffd9aec77311bdf486be/app/dao/returned_letters_dao.py#L150

However this shows that the current architecture is not very clear and I will look at refactoring what we have by implementing something similar to a `.serialize()` that will be used by the admin, callback and GET endpoint.

It will have to go on the back burner for now because of other priorities.

PR #4906, opened 2026-07-01
[DO NOT MERGE] Add migration to create logical replication slot for notify dashboard

DilwoarH:
**NOTE: DO NOT MERGE UNTIL DB HAS BEEN RESTARTED**

## What

Add migration to create logical replication slot for notify dashboard

## Why

This allows us to keep track of changes in the database

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0553_add_reason_to_provider
+0554_create_replacation_slot
--- a/migrations/versions/0554_create_replacation_slot.py
+++ b/migrations/versions/0554_create_replacation_slot.py
@@ -1,0 +1,43 @@
+"""
+Create Date: 2026-07-01T16:39:58
+"""
+
+from alembic import op
+from sqlalchemy import text
+
+revision = "0554_create_replacation_slot"
+down_revision = "0553_add_reason_to_provider"
+slot_name = "notify_dashboard_replication_slot"
+
+def upgrade():
+    bind = op.get_bind()
+    with op.get_context().autocommit_block():
+        bind.execute(
+            text(
+                f"""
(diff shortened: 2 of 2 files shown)


DilwoarH:
Needs https://github.com/alphagov/notifications-aws/pull/3035 to be merged in first

spatel033:
`max_slot_wal_keep_size` is set to `-1` on the local database, which means PostgreSQL will retain WAL indefinitely until a consumer connects and acknowledges them. Is there a deployment plan or any precautions in place to ensure the consumer is available and doesn't cause WAL to grow indefinitely?

DilwoarH:
> `max_slot_wal_keep_size` is set to `-1` on the local database, which means PostgreSQL will retain WAL indefinitely until a consumer connects and acknowledges them. Is there a deployment plan or any precautions in place to ensure the consumer is available and doesn't cause WAL to grow indefinitely?

Good point - I think we'll have to update this here:
https://github.com/alphagov/notifications-local/blob/main/docker-compose.yml#L8-L11

I'll get on this

spatel033:
> Good point - I think we'll have to update this here: https://github.com/alphagov/notifications-local/blob/main/docker-compose.yml#L8-L11
> 
> I'll get on this

The same issue could happen in the production database if the data is not consumed, unless the production database is configured differently.

DilwoarH:
> > Good point - I think we'll have to update this here: https://github.com/alphagov/notifications-local/blob/main/docker-compose.yml#L8-L11
> > I'll get on this
> 
> The same issue could happen in the production database if the data is not consumed, unless the production database is configured differently.

I don't think we want to restrict production as we'll be relying on the data to update the stats. Also WAL is used for other things so don't want to change any of these

DilwoarH:
superseded by #4936

PR #4908, opened 2026-07-02
Enforce NOT NULL constraint on notification_status

DilwoarH:
## What

Enforce NOT NULL constraint on notification_status

## Why

Without NOT NULL - we cannot create an index for notification_status column

Checked DB:

```sql
SELECT count(*)
 FROM notifications where notification_status is NULL;

Result was 0
```

## Context

This is needed as we need to implement the following replica identity:

```sql
ALTER TABLE notifications REPLICA IDENTITY USING INDEX ix_notifications_id_status;
```

Without adding this index we would have to put the replica on the whole table which will keep track of old values of every column which is not needed.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0553_add_reason_to_provider
+0556_add_notify_status_idx
--- a/migrations/versions/0554_add_notif_status_constraint.py
+++ b/migrations/versions/0554_add_notif_status_constraint.py
@@ -1,0 +1,23 @@
+"""
+Create Date: 2026-07-20T00:00:00
+"""
+
+from alembic import op
+
+revision = "0554_add_notif_status_constraint"
+down_revision = "0553_add_reason_to_provider"
+
+
+def upgrade():
+    op.execute(
+        "-- squawk-ignore require-timeout-settings\n"
+        "ALTER TABLE notifications "
+        "ADD CONSTRAINT ck_notifications_notification_status_not_null "
+        "CHECK (notification_status IS NOT NULL) NOT VALID;"
+    )
+
+
+def downgrade():
+    op.execute(
+        "ALTER TABLE notifications DROP CONSTRAINT IF EXISTS ck_notifications_notification_status_not_null;"
+    )
--- a/migrations/versions/0555_add_notify_status_not_null.py
+++ b/migrations/versions/0555_add_notify_status_not_null.py
@@ -1,0 +1,20 @@
+"""
+Create Date: 2026-07-20T00:00:00
+"""
+
+from alembic import op
+
+revision = "0555_add_notify_status_not_null"
+down_revision = "0554_add_notif_status_constraint"
+
+
+def upgrade():
+    op.execute(
+        "ALTER TABLE notifications "
+        "VALIDATE CONSTRAINT ck_notifications_notification_status_not_null;"
+    )
+    op.execute("ALTER TABLE notifications ALTER COLUMN notification_status SET NOT NULL;")
+
+
+def downgrade():
+    op.execute("ALTER TABLE notifications ALTER COLUMN notification_status DROP NOT NULL;")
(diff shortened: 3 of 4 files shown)


migrations/versions/0554_add_notification_status_idx.py
```diff
@@ -0,0 +1,43 @@
...
+            "ADD CONSTRAINT ck_notifications_notification_status_not_null "
+            "CHECK (notification_status IS NOT NULL) NOT VALID;"
+        )
+        op.execute(
+            "ALTER TABLE notifications "
+            "VALIDATE CONSTRAINT ck_notifications_notification_status_not_null;"
+        )
+        op.execute("ALTER TABLE notifications ALTER COLUMN notification_status SET NOT NULL;")
+
+    with op.get_context().autocommit_block():
```
spatel033: Notifications table only handles `INSERT` and `UPDATE` operations, PostgreSQL’s default replica identity already includes all required columns (`template_id`, `notification_status`, `created_at`) in the `wal2json` stream using our existing `id` primary key.
The `DELETE` operations are for archiving notifications to `notification_history` where `notification_status` counts must remain unchanged, so our aggregator can simply ignore all delete logs.
I think in this case we can skip this migration and it also saves write overhead on notifications table.

risicle: 😕 the default replica identity is simply `id`, is it not? And as such we don't get the *old* values for the `status` field in the `wal2json` stream, which we need for bookkeeping so we know which field in the status table to *decrement* as a notification moves to another status.

I agree it's a shame this increases our write load slightly, but I don't think there's another way.

DilwoarH: Yeah unfortunately, only returns ID so without this index there is no other way to get the old values.

DilwoarH:
The base branch for this has now been changed to `main` - this can be merge in independently

whpearson (approved):
Change makes sense.  Shouldn't lock the notifications table.Tested locally

risicle:
I don't understand the name of the migration

migrations/versions/0554_add_notification_status_idx.py
```diff
@@ -0,0 +1,50 @@
...
+
+from alembic import op
+
+revision = "0554_add_notification_status_idx"
+down_revision = "0553_add_reason_to_provider"
+
+
+def upgrade():
+    op.execute("SET lock_timeout = '1s';")
+    op.execute("SET statement_timeout = '5s';")
```
risicle: Why are we doing this here? We already set the `lock_timeout` to 1s in `migrations/env.py`.

DilwoarH: Happy to remove it - didn't know that we had done this already

migrations/versions/0554_add_notification_status_idx.py
```diff
@@ -0,0 +1,50 @@
...
+        op.execute("COMMIT;")
+
+    with op.get_context().autocommit_block():
+        op.execute("BEGIN;")
+        op.execute(
+            "ALTER TABLE notifications "
+            "VALIDATE CONSTRAINT ck_notifications_notification_status_not_null;"
+        )
+        op.execute("ALTER TABLE notifications ALTER COLUMN notification_status SET NOT NULL;")
+        op.execute("COMMIT;")
```
risicle: Ideally we shouldn't be doing multiple transactions for a single migration. If one operation succeeds and a later one fails we end up in a halfway state, neither the state of the previous migration or of this migration. And resuming from that state can be problematic if e.g. the db complains about the first operation being repeated we'll end up with a knot to untie.

Generally if you want things done in separate transactions they should be multiple migrations. And the migration's default transaction should be fine for this.

One exception to this is `CREATE INDEX CONCURRENTLY` which is picky about being done from a transaction.

migrations/versions/0554_add_notification_status_idx.py
DilwoarH: had to add it back in as it's causing linting issues

DilwoarH:
@risicle - migrations have been split up

risicle (approved):

spatel033 (approved):

PR #4909, opened 2026-07-02
[DO NOT MERGE] Enable replica identity using index on notifications

DilwoarH:
## What

Add migration to enable replica identity using index on notifications

## Why

This will allow us to listen to changes made to the date and status


DilwoarH:
superseded by #4936

PR #4910, opened 2026-07-02
Create `ft_service_stats` table

DilwoarH:
## What

Add migration to create `ft_service_stats` table with relevant indexes

## Why

We need the `ft_service_stats` table to store the count of statuses

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0556_add_notify_status_idx
+0557_create_ft_service_stats
--- a/migrations/versions/0557_create_ft_service_stats.py
+++ b/migrations/versions/0557_create_ft_service_stats.py
@@ -1,0 +1,51 @@
+"""
+Create Date: 2026-07-22T00:00:00
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = "0557_create_ft_service_stats"
+down_revision = "0556_add_notify_status_idx"
+
+
+def upgrade():
+    op.create_table(
+        "ft_service_stats",
+        sa.Column("bst_date", sa.Date(), nullable=False),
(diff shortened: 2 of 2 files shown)


migrations/versions/0556_create_ft_service_stats.py
```diff
@@ -0,0 +1,52 @@
...
+
+def upgrade():
+    op.create_table(
+        "ft_service_stats",
+        sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("service_id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("template_id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("notification_type", postgresql.ENUM(name="notification_type", create_type=False), nullable=False),
+        sa.Column("notification_status", sa.Text(), nullable=False),
+        sa.Column("date", sa.Date(), nullable=False, server_default=sa.text("CURRENT_DATE")),
```
klssmith: If this is the date in BST, we probably don't want this default since it won't be in BST. It could help avoid confusion or misunderstandings to call the column `bst_date` too, like we do for `ft_notification_status`

klssmith (commented):
Just making a note of the things we chatted about, one thing we could consider is to use a composite primary key, matching what we do with the existing ft tables, which means the ID column isn't necessary and it will automatically get a unique index.

We also don't want test notifications to show up on the dashboard, so unless we're specifically not writing test notifications to the new table, might need to add a column for `key_type` so we can filter these out

migrations/versions/0556_create_ft_service_stats.py
DilwoarH: done

klssmith (approved):
The plan to not write test notifications to this table means we don't need to have a column for key_type, so this is fine to not have that

PR #4912, opened 2026-07-09
fix: record API key revoker via created_by_id on revoke

joybytes:
### Summary

**Changes on this PR**
We never recorded who revoked a key, so in this PR we're just propagating that information to the database. We are doing this by following the existing pattern of  storing that information on `created_by_id` field.

**Note:** older historical revoke events will show the wrong user (the key creator), because we did not store the revoker separately in the past. This change fixes that for new revokes going forward, but it cannot retroactively fix old row.
The same will happen for `handle archive_service()` -> this is a platform admin functionality, so there is no user name being passed on as part of this flow.

A bit of context on how the tables for `api_keys` and `api_keys_history` work:

We store API keys in two places:
- api_keys = the current/latest state of each key (one row per key)
- api_keys_history = a snapshot history of changes (new row each time the key changes)

**So when a key is revoked:**
- the row in api_keys is updated to show it is now revoked
- a new row is added to api_keys_history to record that revoked version

On admin, audit events are built from the API’s history data (api_key_history), not from the live api_keys row.
So if you want to know what happened over time, look at history; if you want the current state now, look at api_keys.

https://trello.com/c/BwLKBEc4/645-api-key-history-shows-incorrect-data

### DIFF
--- a/app/dao/api_key_dao.py
+++ b/app/dao/api_key_dao.py
@@ -17,9 +17,11 @@ def save_model_api_key(api_key):
 
 @autocommit
 @version_class(ApiKey)
-def expire_api_key(service_id, api_key_id):
+def expire_api_key(service_id, api_key_id, created_by_id=None):
     api_key = ApiKey.query.filter_by(id=api_key_id, service_id=service_id).one()
     api_key.expiry_date = datetime.utcnow()
+    if created_by_id is not None:
+        api_key.created_by_id = created_by_id
     db.session.add(api_key)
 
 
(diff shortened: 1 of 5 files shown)


BlessedDev (commented):
Great work here. I tested the functionality and it worked as expected. I alternated the creator and revoker of the keys and the right user was displayed as the revoker.

I did notice that when adding a new row after an API key is revoked, that the new row was given the same timestamp as the original entry for the API key in the `api_key_history ` table resulting in two entries for the API key with the same created_at timestamp, however the version number is increased so it is clear which event was which. I guess we have to mindful of the discrepancy between the entry in the history table and that in the api_key table for that particular user id(the person who revoked the key)

BlessedDev (approved):

PR #4916, opened 2026-07-14
SMS providers, research mode: supply basic auth credentials with callbacks to delivery status endpoint

risicle:
The first active step of https://trello.com/c/94Q38sgQ/1546-implement-authentication-for-delivery-receipt-callback-endpoints

This causes `FiretextClient` and `MMGClient` to insert basic auth credentials (supplied via `FIRETEXT_DELIVERY_STATUS_CALLBACK_BASIC_AUTH_CREDENTIALS` & `MMG_DELIVERY_STATUS_CALLBACK_BASIC_AUTH_CREDENTIALS`) into the "callback url" we give to the outbound SMS requests. It depends on https://github.com/alphagov/notifications-aws/pull/3062 and needs to be rolled out _well before_ we start actually heeding/enforcing the basic auth at the endpoint (later PR). How long before? The maximum length of time we expect to wait for an SMS's receipt (is this 3 days?).

It also makes our "research mode" (used for test notifications) use basic auth credentials (supplied via `RESEARCH_MODE_SELF_CALLBACK_MMG_BASIC_AUTH_CREDENTIALS` & `RESEARCH_MODE_SELF_CALLBACK_FIRETEXT_BASIC_AUTH_CREDENTIALS`) for its "fake" delivery status callbacks.

It also does a small amount of unit test refactoring to make the use of fixtures for these areas less weird.

You'll possibly notice that this doesn't follow the same (weird) scheme by which username/passwords are supplied for the "inbound sms" system, and that's deliberate. Follow-up PRs will switch inbound sms basic auth to use this new mechanism, because it has fewer unusual quirks.

The PRs that form this story have been tested together in all-singing, all-dancing mode, operating correctly on dev-d.

### DIFF
--- a/tests/app/clients/test_firetext.py
+++ b/tests/app/clients/test_firetext.py
@@ -1,13 +1,22 @@
+from unittest.mock import Mock
 from urllib.parse import parse_qs
 
 import pytest
 import requests_mock
 from requests.exceptions import ConnectTimeout, ReadTimeout
 
-from app.clients.sms.firetext import (
-    SmsClientResponseException,
-    get_firetext_responses,
-)
+from app.clients.sms.firetext import FiretextClient, SmsClientResponseException, get_firetext_responses
+
+
+@pytest.fixture(scope="function")
+def mock_firetext_configured_app():
+    config = {
+        "FIRETEXT_URL": "https://example.com/firetext",
+        "FIRETEXT_API_KEY": "foo",
+        "FIRETEXT_INTERNATIONAL_API_KEY": "international",
+        "FROM_NUMBER": "bar",
+    }
+    return Mock(config=config)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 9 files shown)


tests/app/clients/test_firetext.py
```diff
@@ -1,13 +1,32 @@
...
+
+
+@pytest.fixture(scope="function")
+def mock_firetext_configured_app_with_receipts(mock_firetext_configured_app):
+    return mock_firetext_configured_app.config.update(
+        {
+            "FIRETEXT_RECEIPT_URL": "https://www.example.com/notifications/sms/firetext",
+            "FIRETEXT_DELIVERY_STATUS_CALLBACK_BASIC_AUTH_CREDENTIALS": None,
+        }
+    )
```
rparke: is this used anywhere?

tests/app/clients/test_mmg.py
```diff
@@ -52,16 +65,32 @@ def test_try_send_sms_successful_returns_mmg_response(notify_api, mocker):
     assert response_json["Reference"] == 12345678
 
 
-def test_try_send_sms_successful_with_custom_receipts(notify_api, mock_mmg_client_with_receipts):
+@pytest.mark.parametrize("basic_auth_credentials", (None, ("foo123", "foo321")))
```
rparke: small, verging on nitpicking - can we use `("user123", "pass321")` instead of `("foo123", "foo321")` (like we are for the research mode tests).

tests/app/clients/test_firetext.py
risicle: Uhh, no looks like I got rid of the only remaining use of it...

rparke (approved):

PR #4917, opened 2026-07-15
Add WAL2JSON support for concourse test db

DilwoarH:
## Why

Add WAL2JSON support for concourse test db

## Why

This is needed as concourse tests do not run on RDS - it creates it's own DB.


risicle (approved):
Should do.

PR #4920, opened 2026-07-21
squawk: ignore `require-statement-timeout` and `require-lock-timeout` rules

risicle:
These are not appropriate for us, because we already set these for each migration (though in a way squawk can't see - in the harness) and it prompts people to add these lines to migrations without really knowing what values to give them just to shut the CI up.


DilwoarH:
instead of this, should we just use a config file:

https://squawkhq.com/docs/cli#squawktoml-configuration-file

risicle:
Where do you think we should put it? Repo root?

DilwoarH:
> Where do you think we should put it? Repo root?

Yh repo root would be the ideal place or could even sit in the scripts dir

risicle:
Going to have to temporarily add a migration that would set these off..

risicle:
Oh and this was only split into two rules in 2.60.0 (https://github.com/sbdchd/squawk/blob/master/CHANGELOG.md#v2600---2026-07-18), so I should probably upgrade us to that at the same time.

uv.toml
```diff
@@ -1,6 +1,6 @@
 # This file was automatically copied from notifications-utils@121.0.0
 
-exclude-newer = "7 days"
+exclude-newer = "5 days"
```
quis: https://github.com/alphagov/notifications-manuals/wiki/Dependencies#if-you-need-a-very-fresh-version-of-a-dependency-released-in-the-last-7-days

risicle: Yup that is just a very temporary hack - in reality putting an exception here doesn't work very well because the `copy_config()` thing overwrites the `uv.toml` before it runs the compile for `requirements_for_test.txt`

risicle: Which raises an interesting point actually - the compile steps for `requirements.txt` and `requirements_for_test.txt` run against different `uv.toml` versions.

risicle: In theory what we should do is re-run the compile for `requirements.txt` after the `copy_config()` and repeat the process until nothing changes and we've therefore reached a "fixed point" (https://en.wikipedia.org/wiki/Fixed-point_combinator)

quis: I guess ideally the freeze command would:
- install specified version of utils
- run `copy_config()`
- freeze both `.txt` files and `sync`

risicle:
Success. Without the exclusions: https://concourse.notify.tools/teams/pull-requests/pipelines/pull-requests/jobs/build-notifications-api-pr/builds/1858#L6a62394d:1 , with the exclusions: https://concourse.notify.tools/teams/pull-requests/pipelines/pull-requests/jobs/build-notifications-api-pr/builds/1856#L6a625f99:1

uv.toml
risicle: Might work if you're certain utils is the only thing that affects the outcome.

quis (approved):

PR #4922, opened 2026-07-23
Use `RelaxedContainerJSONEncoder` & `FlaskRelaxedContainerJSONProvider` for json encoding where appropriate

risicle:
This bumps notifications-utils to 124.2.0 and uses `RelaxedContainerJSONEncoder` to encode json payloads we're sending via `requests`. This should address any problems arising from `InsensitiveDict` no longer directly inheriting from `dict`.

It also sets `FlaskRelaxedContainerJSONProvider` our flask app's `json_provider_class` so we get the same robustness when returning json via e.g. `jsonify`.

There are also a couple of other random places we used `json.dumps` which I've replaced.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-# This file was automatically copied from notifications-utils@121.0.0
+# This file was automatically copied from notifications-utils@124.2.0
 
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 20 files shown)


quis (approved):
There will be a way to write a test which uses `ast` to fail on finding instances of `json.dumps`.

This would prevent people reintroducing the bugs that this is intended to solve.

Don’t think it’s worth the afternoon’s work to write it though.

PR #4925, opened 2026-07-31
Add NHS Notify org type with no free allowance

klssmith:
This adds a new org type of `nhs_notify` and sets its default free allowance to 0.

### DIFF
--- a/migrations/versions/0559_set_nhs_notify_allowance.py
+++ b/migrations/versions/0559_set_nhs_notify_allowance.py
@@ -1,0 +1,29 @@
+"""
+Create Date: 2026-07-31 10:11:51.069316
+"""
+
+import uuid
+
+from alembic import op
+
+revision = '0559_set_nhs_notify_allowance'
+down_revision = '0558_add_nhs_notify_org_type'
+
+
+def upgrade():
+    """
+    The free allowance is set from 2016 even though it's a new org type for consistency and to avoid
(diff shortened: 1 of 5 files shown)


klssmith:
This fails the migration isolation warning, but one file changed is just a test, and in the other case we're only adding a new constant for an org type that isn't used yet, so the changes won't cause issues if deployed together.

migrations/versions/0559_set_nhs_notify_allowance.py
```diff
@@ -0,0 +1,29 @@
...
+
+def upgrade():
+    """
+    The free allowance is set from 2016 even though it's a new org type for consistency and to avoid
+    potential bugs if trying to view historic usage
+    """
+    op.execute(
+        "INSERT INTO default_annual_allowance"
+        "(id, valid_from_financial_year_start, organisation_type, allowance, notification_type) "
+        f"VALUES('{uuid.uuid4()}', 2016, 'nhs_notify', 0, 'sms')"
```
spatel033, line 21: Shouldn't the financial year be 2026?

klssmith: I went for 2016 as the `valid_from_financial_year_start` so that we have a value for all org types from the same start date - the start of Notify. I was worried there's the potential for a bug if we have "missing years" for an org type

spatel033 (approved):

PR #4926, opened 2026-07-31
Add NHS Notify org

klssmith:
This adds the NHS Notify org - this org will be a unique case, so we want to create it through a migration so that we have a set ID for it which we can then refer to in all environments.

**To do after deployment**
- [x] Clear the organisation cache


klssmith:
Squawk failing because it can't understand this syntax - like we've seen when changing letter rates but weren't able to get round easily

spatel033 (approved):

PR #4929, opened 2026-08-04
add document download count to notifications sent with files via the ui

rparke:


app/v2/notifications/post_notifications.py
```diff
@@ -253,6 +253,10 @@ def process_sms_or_email_notification(
         # We changed personalisation which means we need to update the content
         template_with_content.values = personalisation
 
+    if template_email_file_count := len(template.email_file_objects) > 0:
+        # the total count should be the number of files an api user has added + the number already in the template
+        document_download_count += template_email_file_count
```
quis: This will do the same thing in both the `0` and `> 0` case:
```suggestion
    # the total count should be the number of files an api user has added + the number already in the template
    document_download_count += len(template.email_file_objects)
```

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -57,7 +59,7 @@ def test_post_sms_notification_returns_201(api_client_request, sample_template_w
     assert notifications[0].status == NOTIFICATION_CREATED
     notification_id = notifications[0].id
     assert notifications[0].postage is None
-    assert notifications[0].document_download_count is None
+    assert notifications[0].document_download_count == 0
```
rparke, line 62: Setting this to 0 vs None is probably right, but I'm not sure?

This field is never used anywhere but it feels good to pick what we're setting this as and be consistent...

quis: What do we have in the database at the moment?

For:
- emails without file links
- text messages or letters which cannot have file links

quis:
```sql
select
    distinct(document_download_count, notification_type)
from
    notifications
where
    created_at > now() - interval '30 minutes'
```

| row              |
|------------------|
| `('1', 'email')`   |
| `('2', 'email')`   |
| `(None, 'email')`  |
| `(None, 'sms')`    |
| `(None, 'letter')` |

***

I think it’s better if we stick with `None` otherwise anyone doing a query needs to account for the `None` and `0` cases. Probably a better alternative than doing a migration to convert all the `None`s to `0`s

rparke: Yeh makes sense, they're all set to none.

I was thinking that retention period would mean that everything was eventually consistent because I don't _think_ notifications history records this?

quis (approved):

PR #4931, opened 2026-08-05
`process_mmg_response`, `process_firetext_response`: protect views with basic auth

risicle:
https://trello.com/c/94Q38sgQ/1546-implement-authentication-for-delivery-receipt-callback-endpoints

The existing pattern for auth-protecting views is via `before_request` functions applied to their blueprints, but this doesn't provide the opportunity for configuring each individually, and we need to do that to point each view at its respective config variable supplying "allowed" credentials. So while `requires_basic_auth` is _usable_ as a `before_request` hook (if you use `functools.partial` to supply its config variable), it's better used via its `view_requires_basic_auth` wrapper decorator.

Another notable design decision here is to accept bcrypt hashes of the basic auth passwords corresponding to each username as configuration (multiple username/password pairs are of course supported to enable smooth credential rotation). While this is of questionable value seeing as the wider application needs to have access to the plaintext passwords used in callbacks by both providers and "research mode", technically api-web *doesn't* need access to these. So keeping them as hashes adds a small amount of protection.

To mitigate this causing a performance problem (having to perform a deliberately-expensive password hash for each request is not the normal use-case for password hashing, but it's hard to avoid with basic auth), the `check_hash` function is wrapped in a least-frequently-used in-memory cache. This has some security tradeoffs because it means a plaintext password will end up being kept in memory once it has been used, but it should mean most requests with correct credentials have zero hashing cost.

This mustn't be merged until we're absolutely sure we've got all the other parts in place and working.

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -1,6 +1,9 @@
 import time
 import uuid
+from collections.abc import Callable
+from functools import wraps
 
+from cachetools.func import lfu_cache
 from flask import current_app, g, request
 from gds_metrics import Histogram
 from notifications_python_client.authentication import (
@@ -16,6 +19,8 @@ from notifications_python_client.errors import (
 )
 from sqlalchemy.orm.exc import NoResultFound
 
+from app import memo_resetters
+from app.hashing import check_hash
 from app.serialised_models import SerialisedService
 
 GENERAL_TOKEN_ERROR_MESSAGE = "Invalid token: make sure your API token matches the example at https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header"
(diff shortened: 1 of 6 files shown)


app/authentication/auth.py
```diff
@@ -191,3 +196,73 @@ def _get_token_issuer(auth_token):
...
+) -> None:
+    """
+    `credentials_config_key` is expected to be a key in the flask app config containing a mapping
+    of usernames to bcrypt hashes of their basic auth passwords. a very low bcrypt difficulty can
+    be used in the hashes if performance is a concern.
+
+    For use as a `before_request` function, apply `credentials_config_key` with `functools.partial`.
+    """
+    auth = request.authorization
+    creds_dict = current_app.config.get(credentials_config_key) or {}
```
rparke, line 234: does `app.config.get` work the same as dicts `get` method? If so can just refactor to:

`creds_dict = current_app.config.get(credentials_config_key, {})`

(if not ... obviously you can't and this is fine :P )

tests/app/authentication/test_authentication.py
```diff
@@ -356,3 +359,66 @@ def test_requires_internal_auth_sets_global_variables(
     request.headers = {"Authorization": f"Bearer {internal_jwt_token}"}
     requires_my_internal_app_auth()
     assert g.service_id == "my-internal-app"
+
+
+def test_requires_basic_auth_missing_auth(notify_api, client):
+    notify_api.config["FOO_BAR_BAZ"] = {}
```
rparke, line 367: Worth also testing this without setting the app config? It _should_ work the same for both.

app/authentication/auth.py
risicle: I'm considering the case where it's set, but set to `None`

rparke:
Looks good an the logging only means it _looks_ safe for rollout :)

rparke (approved):

PR #4937, opened 2026-08-17
Allow a full range of unicode characters in text messages

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/4935
- [x] https://github.com/alphagov/notifications-utils/pull/1443

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -60,7 +60,7 @@ def send_sms_to_provider(notification: Notification) -> None:
         )
 
         if non_gsm_characters := template.non_gsm_characters:
-            current_app.logger.warning(
+            current_app.logger.info(
                 "%s character(s) caused UTF-16 encoding in SMS content for service %s and notification %s",
                 len(non_gsm_characters),
                 service.id,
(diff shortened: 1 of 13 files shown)


app/delivery/send_to_providers.py
```diff
@@ -60,9 +59,9 @@ def send_sms_to_provider(notification: Notification) -> None:
             show_prefix=service.prefix_sms,
         )
 
-        if non_gsm_characters := SanitiseSMS.get_non_gsm_characters(template.unsanitised_content):
+        if non_gsm_characters := template.non_gsm_characters:
             current_app.logger.warning(
-                "%s character(s) replaced with ? in SMS content for service %s and notification %s",
+                "%s character(s) caused UTF-16 encoding in SMS content for service %s and notification %s",
```
CrystalPea, line 64: do we still want this log to be a warning? Since this will be an expected thing?

quis: I figure it will be useful initially to see adoption of the feature, but we can remove it eventually

CrystalPea: should we move the log type from warning to info though?

quis: Sure, can do, changed in 439c519

CrystalPea (approved):

rparke (approved):

PR #4943, opened 2026-08-26
inbound sms callback views: add `view_requires_basic_auth` decorator in `log_only` mode

risicle:
A followup to https://trello.com/c/94Q38sgQ/1546-implement-authentication-for-delivery-receipt-callback-endpoints to (eventually) switch the inbound sms basic auth to use the new mechanism added for delivery status callbacks.

This first PR, like #4931 before it did, only adds the new basic auth decorator in `log_only` mode so we can be certain it's working correctly and we've got the credentials correct. The existing basic auth mechanism remains in place doing the actual enforcing until we tear that out in a following PR (which also drops the old-style config variables and updates the unit tests to exercise the new mechanism more fully).

Of course, before I could get to any of that I had to fix the existing unit tests to work more correctly and sanely. The first commit addresses this, and the commit adding the new decorator doesn't itself introduce any test changes, because it shouldn't change any real behaviour.


tests/app/notifications/test_receive_notification.py
```diff
@@ -487,15 +473,18 @@ def test_mmg_inbound_sms_auth(notify_db_session, notify_api, client, mocker, aut
...
     data = {
         "ID": "1234",
-        "MSISDN": "07111111111",
+        "MSISDN": "07222222222",
         "Message": "Some message to notify",
         "Trigger": "Trigger?",
-        "Number": "testing",
+        "Number": "07111111111",
         "Channel": "SMS",
         "DateRecieved": "2012-06-27 12:33:00",
```
rparke, line 489: Why do we have MISDN differ from the number?

risicle: The `MSISDN` is the "from" number. Or at least that's how we interpret it in our handler.

rparke (approved):

PR #4945, opened 2026-08-27
log every persist notification call for SMS notifications if a user t…

rparke:
…ries to send to a uk mobile number in the reserved range

### DIFF
--- a/app/notifications/process_notifications.py
+++ b/app/notifications/process_notifications.py
@@ -9,7 +9,7 @@ from notifications_utils.recipient_validation.email_address import (
     format_email_address,
     validate_and_format_email_address,
 )
-from notifications_utils.recipient_validation.phone_number import UK_PREFIX
+from notifications_utils.recipient_validation.phone_number import UK_PREFIX, InvalidPhoneError, PhoneNumber
 from notifications_utils.template import (
     LetterPrintTemplate,
     PlainTextEmailTemplate,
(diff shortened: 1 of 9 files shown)


app/notifications/process_notifications.py
```diff
@@ -177,6 +177,18 @@ def persist_notification(
...
+            five_digit_prefix = str(number.number.national_number)[:5]
+            if number.is_uk_mobile_number() and number.get_carrier_info == "":
+                current_app.logger.info(
+                    "Service %s tried to send to UK mobile number in reserved range. Prefix without leading zero: %s",
+                    service.id,
+                    five_digit_prefix,
+                    extra={"service_id": service.id, "five_digit_prefix": five_digit_prefix},
+                )
+        except InvalidPhoneError:
+            current_app.logger.info("Could not parse number")
```
rparke, line 191: The try-except is mainly just extra caution. Nothing in the logging code should stop us creating the row as normal.

app/notifications/process_notifications.py
rparke, line 204: We could optionally get the notification_id returned to us from dao_create_notification and put that in the log message (removes the need for storing the prefix but does require some more manual effort cross referencing)

CrystalPea: I think the prefix will be easier for data analysis - and if we want to contact the service about particular notifications, we will be able to also find them by the prefix and sending time if they're within data retention.

CrystalPea (approved):

PR #4930, opened 2026-08-05
Process replication slot changes

DilwoarH:
** DO NOT MERGE UNTIL #4946 IS MERGED IN FIRST** 

## What

The application code for processing slot changes for notifications table

## Why

This is needed to process WAL logs in the database

## Test results

Ran a load test on `dev-c` - this is the result after 2400 emails.

```sql
SELECT
	(SELECT count(*) FROM pg_logical_slot_peek_changes('notify_dashboard_replication_slot', NULL, NULL)) AS still_processing,
    (SELECT SUM(notification_count) FROM ft_service_stats) AS ft_stats_count,
    COUNT(*) AS notif_count,
    (SELECT SUM(notification_count) FROM ft_service_stats where notification_status = 'created') AS ft_created_count,
    COUNT(CASE WHEN notification_status = 'created' THEN 1 END) AS notif_created_count,
    (SELECT SUM(notification_count) FROM ft_service_stats where notification_status = 'sending') AS ft_sending_count,
    COUNT(CASE WHEN notification_status = 'sending' THEN 1 END) AS notif_sending_count,
    (SELECT SUM(notification_count) FROM ft_service_stats where notification_status = 'delivered') AS ft_delivered_count,
    COUNT(CASE WHEN notification_status = 'delivered' THEN 1 END) AS notif_delivered_count,
    (SELECT SUM(notification_count) FROM ft_service_stats where notification_status = 'temporary-failure') AS ft_temp_fail_count,
    COUNT(CASE WHEN notification_status = 'temporary-failure' THEN 1 END) AS notif_temp_fail_count,
    (SELECT SUM(notification_count) FROM ft_service_stats where notification_status = 'permanent-failure') AS ft_perm_fail_count,
    COUNT(CASE WHEN notification_status = 'permanent-failure' THEN 1 END) AS notif_perm_fail_count
FROM
    public.notifications;
```

<table>
    <tr>
        <th>still_processing</th>
        <th>ft_stats_count</th>
        <th>notif_count</th>
        <th>ft_created_count</th>
        <th>notif_created_count</th>
        <th>ft_sending_count</th>
        <th>notif_sending_count</th>
        <th>ft_delivered_count</th>
        <th>notif_delivered_count</th>
        <th>ft_temp_fail_count</th>
        <th>notif_temp_fail_count</th>
        <th>ft_perm_fail_count</th>
        <th>notif_perm_fail_count</th>
    </tr>
    <tr>
        <td>0</td>
        <td>2400</td>
[...]

PR #4946, opened 2026-08-28
Add migration to create logical replication slot

DilwoarH:
This creates a wal2json replication slot to allow us to listen to changes to the notifications table.

### DIFF
--- a/migrations/.current-alembic-head
+++ b/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-0563_confirmed_service_name_col
+0564_create_replacation_slot
--- a/migrations/versions/0564_create_replacation_slot.py
+++ b/migrations/versions/0564_create_replacation_slot.py
@@ -1,0 +1,43 @@
+"""
+Create Date: 2026-08-17 00:00:00
+"""
+
+from alembic import op
+from sqlalchemy import text
+
+revision = "0564_create_replacation_slot"
+down_revision = "0563_confirmed_service_name_col"
+slot_name = "notify_dashboard_replication_slot"
+
+def upgrade():
+    bind = op.get_bind()
+    with op.get_context().autocommit_block():
+        bind.execute(
+            text(
+                f"""
+                SELECT pg_create_logical_replication_slot('{slot_name}', 'wal2json')
+                WHERE NOT EXISTS (
+                    SELECT 1
+                    FROM pg_replication_slots
+                    WHERE slot_name = '{slot_name}'
+                )
+                """
+            )
+        )
+
+
+def downgrade():
+    bind = op.get_bind()
+    with op.get_context().autocommit_block():
+        bind.execute(
+            text(
+                f"""
+                SELECT pg_drop_replication_slot('{slot_name}')
+                WHERE EXISTS (
+                    SELECT 1
+                    FROM pg_replication_slots
+                    WHERE slot_name = '{slot_name}'
+                )
+                """
+            )
+        )


migrations/versions/0562_create_replacation_slot.py
```diff
@@ -0,0 +1,43 @@
...
+slot_name = "notify_dashboard_replication_slot"
+
+def upgrade():
+    bind = op.get_bind()
+    with op.get_context().autocommit_block():
+        bind.execute(
+            text(
+                f"""
+                SELECT pg_create_logical_replication_slot('{slot_name}', 'wal2json')
+                WHERE NOT EXISTS (
```
joybytes: have you tried?
```
FROM (SELECT 1) AS dummy
WHERE NOT EXISTS (
```

klssmith:
Can we test this on a dev env? It seems like we're going to run into the problem of `output_plugin_libraries` now needing to include wal2json. [This PR](https://github.com/alphagov/notifications-api/pull/4950) added that to the Concourse test image, but I'd expect us to need the same change in our environments

DilwoarH:
> Can we test this on a dev env? 

Good point, Will run another test.

> It seems like we're going to run into the problem of `output_plugin_libraries` now needing to include wal2json. [This PR](https://github.com/alphagov/notifications-api/pull/4950) added that to the Concourse test image, but I'd expect us to need the same change in our environments

I noticed that I missed the default ones, it should have been:

```
pgoutput, test_decoding, wal2json
```

I'll have a look at adding this in.

risicle:
Dude before we _consider_ enabling this for production we need to choose and set a `max_slot_wal_keep_size`. I think it's currently set to "infinite", which means this would just start eating up our disk space forever if nothing were consuming the slot.

DilwoarH:
> Dude before we _consider_ enabling this for production we need to choose and set a `max_slot_wal_keep_size`. I think it's currently set to "infinite", which means this would just start eating up our disk space forever if nothing were consuming the slot.

Yeah - I'm avoiding merging this in before the processing code (#4930).

DilwoarH:
> > Can we test this on a dev env?
> 
> Good point, Will run another test.
> 
> > It seems like we're going to run into the problem of `output_plugin_libraries` now needing to include wal2json. [This PR](https://github.com/alphagov/notifications-api/pull/4950) added that to the Concourse test image, but I'd expect us to need the same change in our environments
> 
> I noticed that I missed the default ones, it should have been:
> 
> ```
> pgoutput, test_decoding, wal2json
> ```
> 
> I'll have a look at adding this in.

A PR was merged to fix the concourse test dbs - #4955

DilwoarH:
> Can we test this on a dev env? It seems like we're going to run into the problem of `output_plugin_libraries` now needing to include wal2json. [This PR](https://github.com/alphagov/notifications-api/pull/4950) added that to the Concourse test image, but I'd expect us to need the same change in our environments

@klssmith additional test ran on `dev-d` - worked fine.

DilwoarH:
Todo before merging this in:
- Set max size limit
- Set output_plugin_libraries

DilwoarH:
@risicle - looks like there is a default limit set of 2GB already:

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.Parameters.html#:~:text=max_wal_size,SHOW%20max_wal_size%3B

Screenshot 2026-09-03 at 12 03 32 AM
(from dev d)

klssmith:
I realised this will work in our environments now because we're using Postgres 15.17, and the restriction on `output_plugin_libraries` was only introduced in 15.19. But we will presumably have issues when we do upgrade Postgres - it feels like it would be best just to update `output_plugin_libraries` now before we're using wal2json and in case we don't remember we need to do that before upgrading Postgres. That wouldn't block this PR but might be something to discuss

risicle:
> @risicle - looks like there is a default limit set of 2GB already:

That's not the same config setting.

DilwoarH:
> > @risicle - looks like there is a default limit set of 2GB already:
> 
> That's not the same config setting.

PR added: https://github.com/alphagov/notifications-aws/pull/3094

DilwoarH:
[#3094](https://github.com/alphagov/notifications-aws/pull/3094) has been merged in but requires reboot to apply changes

DilwoarH:
> [#3094](https://github.com/alphagov/notifications-aws/pull/3094) has been merged in but requires reboot to apply changes

Restart not required: https://github.com/alphagov/notifications-aws/pull/3096

DilwoarH:
@risicle @klssmith - are you both happy with this now?

klssmith:
It doesn't block this, but I still think we need a definite plan or PR for when we'll change `output_plugin_libraries` to include `wal2json`. Otherwise it will get forgotten about and cause an incident when the Postgres version is upgraded and our code no longer works

DilwoarH:
> It doesn't block this, but I still think we need a definite plan or PR for when we'll change `output_plugin_libraries` to include `wal2json`. Otherwise it will get forgotten about and cause an incident when the Postgres version is upgraded and our code no longer works

PR added to support this:

https://github.com/alphagov/notifications-aws/pull/3100

PR #4948, opened 2026-08-28
chore: `dao_get_template_email_file_by_id` look up the file by service and template ID

joybytes:
## Summary
`dao_get_template_email_file_by_id` now takes in service and template ID as arguments.

This came from the scan report which said: The endpoint looked files up by file ID alone, so the service and template in the URL were ignored. The admin app already does the authorisation checks, so this is not exploitable and was triaged as low, but we want the DAO to filter by service ID like it does elsewhere.
Checking the template ID as well is worth having because of restricted template folders: within one service a user can be allowed to see one template but not another, so the file should have to belong to the template in the URL, not just the service.

Additionally, I added a read replica to the dao function, as it can be used by primary database or replica depending on the flow of higher functions

https://trello.com/c/NXqWA36L/755-template-email-file-endpoints-do-not-bind-template-file-ids-to-the-route-service-template

### DIFF
--- a/app/dao/template_email_files_dao.py
+++ b/app/dao/template_email_files_dao.py
@@ -51,10 +51,17 @@ def dao_get_template_email_files_by_template_id(template_id, template_version=No
     ).all()
 
 
-@autocommit
-def dao_get_template_email_file_by_id(service_id, template_id, template_email_file_id):
+@retryable_query()
+def dao_get_template_email_file_by_id(
+    session: Session | scoped_session = db.session,
+    *,
+    service_id,
+    template_id,
+    template_email_file_id,
+):
     return (
-        TemplateEmailFile.query.join(Template, Template.id == TemplateEmailFile.template_id)
+        session.query(TemplateEmailFile)
+        .join(Template, Template.id == TemplateEmailFile.template_id)
         .filter(
             TemplateEmailFile.id == template_email_file_id,
             Template.service_id == service_id,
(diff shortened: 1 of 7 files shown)


app/dao/template_email_files_dao.py
```diff
@@ -52,8 +52,16 @@ def dao_get_template_email_files_by_template_id(template_id, template_version=No
 
 
 @autocommit
-def dao_get_template_email_file_by_id(template_email_file_id):
-    return TemplateEmailFile.query.filter(TemplateEmailFile.id == template_email_file_id).one()
+def dao_get_template_email_file_by_id(service_id, template_email_file_id, template_id=None):
```
joybytes: check if this function should be through replica

app/dao/template_email_files_dao.py
```diff
@@ -51,9 +51,24 @@ def dao_get_template_email_files_by_template_id(template_id, template_version=No
     ).all()
 
 
-@autocommit
-def dao_get_template_email_file_by_id(template_email_file_id):
-    return TemplateEmailFile.query.filter(TemplateEmailFile.id == template_email_file_id).one()
+@retryable_query()
+def dao_get_template_email_file_by_id(
+    session: Session | scoped_session = db.session,
```
risicle: Only suggestion I'd make is put this after the `*` with everything else as we tend to have `session` as a final argument elsewhere - but I'm not sure it's important.

risicle (approved):

app/dao/template_email_files_dao.py
joybytes: https://github.com/alphagov/notifications-api/pull/4948/changes/ce897a16680019aebdadb24443dc7d9a2a2f0e8a

risicle (approved):

PR #4951, opened 2026-09-01
inbound sms: remove old ad-hoc basic auth code, activate enforcement by new decorator

risicle:
https://trello.com/c/94Q38sgQ/1546-implement-authentication-for-delivery-receipt-callback-endpoints

This includes removal of old-style config variables used by old code.

We can be quite sure this will work because neither staging or prod are currently emitting any error messages from the decorator in `log_only` mode.

After this, the only thing left for this story is tidying up of old variables/credentials etc.

### DIFF
--- a/tests/app/functional_test_fixtures/test_functional_test_fixtures.py
+++ b/tests/app/functional_test_fixtures/test_functional_test_fixtures.py
@@ -42,8 +42,7 @@ def test_create_db_objects_sets_db_up(notify_api, notify_service):
         with set_config_values(
             notify_api,
             {
-                "MMG_INBOUND_SMS_USERNAME": ["test_mmg_username"],
-                "MMG_INBOUND_SMS_AUTH": ["test_mmg_password"],
+                "MMG_INBOUND_SMS_CALLBACK_BASIC_AUTH_CREDENTIALS": ["test_mmg_username", "test_mmg_password"],
                 "INTERNAL_CLIENT_API_KEYS": {"notify-functional-tests": ["functional-tests-secret-key"]},
                 "ADMIN_BASE_URL": "http://localhost:6012",
                 "API_HOST_NAME": "http://localhost:6011",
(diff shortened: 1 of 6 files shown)


tests/app/functional_test_fixtures/test_functional_test_fixtures.py
```diff
@@ -42,8 +42,7 @@ def test_create_db_objects_sets_db_up(notify_api, notify_service):
         with set_config_values(
             notify_api,
             {
-                "MMG_INBOUND_SMS_USERNAME": ["test_mmg_username"],
-                "MMG_INBOUND_SMS_AUTH": ["test_mmg_password"],
+                "MMG_INBOUND_SMS_CALLBACK_BASIC_AUTH_CREDENTIALS": ["test_mmg_username", "test_mmg_password"],
```
CrystalPea, line 45: relying on order of elements to know which is which is a bit dicey - maybe we could make this into a dictionary instead of a list, so instead of assigning like `current_app.config["MMG_INBOUND_SMS_CALLBACK_BASIC_AUTH_CREDENTIALS"][0]`, we could assign like `current_app.config["MMG_INBOUND_SMS_CALLBACK_BASIC_AUTH_CREDENTIALS"]["username"]`?

CrystalPea (commented):
Nice one! One small suggestion about data structure 🙌🏼

tests/app/functional_test_fixtures/test_functional_test_fixtures.py
risicle: We're already using this configuration pattern for the other endpoints using basic auth, and I have already generated, committed and uploaded all the credentials for doing this for all our environments. The _only_ thing not being done with this configuration pattern is actually denying requests - it's mostly a tidyup PR at this point.

I'd be very surprised if someone expected a username/password combination to be listed as password, username - I can't think of another place in computing things are specified that way round. Someone could equally feed in a dict with `user` instead of `username` and/or `pass` instead of `password`.

CrystalPea (approved):

PR #4952, opened 2026-09-01
add `get_banded_slow_text_message_delivery_reports_by_provider`, record result in `generate_sms_delivery_stats`

risicle:
https://trello.com/c/1Q7z3GH3/1568-investigate-and-decide-what-to-do-next-re-balancer-behaviour-during-vodafone-network-outage

This is a candidate building block for a new sms balancer slow delivery metric that is capable of measuring delivery delay in a series of "bands" that each trade off severity of delay measured with fast responses to interruption.

It also notably uses `sent_at` rather than `created_at` to calculate delivery delay, preventing it from being sensitive to delivery delays at our end.

Here we simply expose the separate components of this measure as an otel metric, allowing us to experiment with real data for determining e.g. methods off combining the bands into a single representative figure which we might later be able to make balancing decisions based on.

Sending these queries to the replica to avoid any possibility these could seriously affect the primary database.

In this PR I've chosen bands systematically rather than using "human-feeling" boundaries of e.g. 1m, 5m, 15m. The smallest band I've chosen counts how many notifications took more than 30s to deliver in the time window 30s ago to 1m ago. Each successive band is double the size of the previous, and has a delivery threshold similarly twice the size, up to a band counting messages that took more than 8m to deliver looking back up to 16m ago.

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -2,6 +2,7 @@ import csv
 import io
 from collections import defaultdict
 from datetime import date, datetime, timedelta
+from itertools import pairwise
 
 import jinja2
 import sentry_sdk
@@ -62,6 +63,7 @@ from app.dao.notifications_dao import (
     dao_letters_in_technical_failure,
     dao_old_letters_with_created_status,
     dao_precompiled_letters_still_pending_virus_check,
+    get_banded_slow_text_message_delivery_reports_by_provider,
     get_slow_text_message_delivery_reports_by_provider,
     is_delivery_slow_for_providers,
     letters_missing_from_sending_bucket,
@@ -95,6 +97,7 @@ from app.notifications.process_notifications import persist_notification, send_n
 from app.otel_metrics.provider import (
     record_info,
     record_priority,
+    record_sms_banded_not_delivered_within,
     record_sms_legacy_not_delivered_within,
     record_updated_at,
 )
@@ -258,6 +261,9 @@ def _check_slow_text_message_delivery_reports_and_raise_error_if_needed(reports:
         redis_store.set(CacheKeys.NUMBER_OF_TIMES_OVER_SLOW_SMS_DELIVERY_THRESHOLD, 0)
 
 
+_sms_slow_delivery_bands = tuple(pairwise(timedelta(seconds=30) * 2**i for i in range(6)))
+
+
 @notify_celery.task(name="generate-sms-delivery-stats")
 def generate_sms_delivery_stats():
     for delivery_interval in (1, 5, 10):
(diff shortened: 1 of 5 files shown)


risicle:
Screenshot 2026-09-02 at 12 48 26

Here is an example of it running in a dev environment - showing `..._total` because we don't really get enough slow deliveries in dev to make even the 30s band show much useful. But here you can see the bursts of notifications created by the FTs falling into successive bands as time progresses, each band being over a longer scale and therefore more "stretched out".

app/celery/scheduled_tasks.py
```diff
@@ -258,6 +261,9 @@ def _check_slow_text_message_delivery_reports_and_raise_error_if_needed(reports:
         redis_store.set(CacheKeys.NUMBER_OF_TIMES_OVER_SLOW_SMS_DELIVERY_THRESHOLD, 0)
 
 
+_sms_slow_delivery_bands = tuple(pairwise(timedelta(seconds=30) * 2**i for i in range(6)))
```
robinjam: This is very elegant

tests/app/celery/test_scheduled_tasks.py
```diff
@@ -288,35 +350,167 @@ def test_generate_sms_delivery_stats(slow_delivery_config_option, expect_check_s
...
-            mocker.call(0.4, {"provider.name": "mmg", "time_window.evaluation": 900, "time_window.delivery": 600}),
-            mocker.call(0.8, {"provider.name": "firetext", "time_window.evaluation": 900, "time_window.delivery": 600}),
+            call(0.4, {"provider.name": "mmg", "time_window.evaluation": 900, "time_window.delivery": 60}),
+            call(0.8, {"provider.name": "firetext", "time_window.evaluation": 900, "time_window.delivery": 60}),
+            call(0.4, {"provider.name": "mmg", "time_window.evaluation": 900, "time_window.delivery": 300}),
+            call(0.8, {"provider.name": "firetext", "time_window.evaluation": 900, "time_window.delivery": 300}),
+            call(0.4, {"provider.name": "mmg", "time_window.evaluation": 900, "time_window.delivery": 600}),
+            call(0.8, {"provider.name": "firetext", "time_window.evaluation": 900, "time_window.delivery": 600}),
+        ),
+        key=lambda c: sorted(c.args[1].items()),
```
robinjam, line 360: I'm wondering if it would make sense to either extract this lambda to an appropriately-named function (perhaps `sorted_metric_labels`), or even lift the whole `sorted(..., key=lambda c: sorted(c.args[1].items()))` pattern into something like `sorted_by_metric_labels`.

robinjam (approved):
Looks good, just one nitpick, you can decide if you want to address it or not

tests/app/celery/test_scheduled_tasks.py
risicle: Meh I don't love leaving my mess (named helper functions etc) around for other tests to trip over unless I really need to.

PR #4953, opened 2026-09-01
Send a new email to users with manage settings when a service goes live

klssmith:
We currently send an email to all team members of a service when the service is made live. We we now want to send two emails - one to all team members, and one to team members with the manage settings permission.

The new template, for users with the manage settings permission, has already [been created in all environments](https://github.com/alphagov/notifications-api/pull/4944).

The existing template that is currently sent to all users will have its content modified to remove the content about out of hours emergencies and admin user responsibilities. We will deploy this change, then immediately change the content of the current email. It won't be an issue if people receive the existing email with its current content before we've had a chance to change it - they will just see a tiny bit of duplication of content in both emails they receive.

[Trello card](https://trello.com/c/LO75LC8Y/1962-go-live-emails-1-3-add-new-email)


tests/app/dao/test_services_dao.py
```diff
@@ -1140,6 +1141,34 @@ def test_dao_fetch_active_users_for_service_returns_active_only(notify_db_sessio
     users = dao_fetch_active_users_for_service(service.id)
 
     assert len(users) == 1
+    assert users[0].id == active_user.id
+
+
+def test_dao_fetch_active_users_with_manage_settings_for_service(notify_db_session):
```
CrystalPea, line 1147: in this test it would be good to have a user who is also part of the right service, but the permissions don't line up - to show that this user is not scoped.

CrystalPea (commented):
Nice one! One little comment 🙌🏼

tests/app/dao/test_services_dao.py
klssmith: Good idea - I've added `active_user_3` to this test and only given them `manage_templates` permission

CrystalPea (approved):
:shipit:

PR #4956, opened 2026-09-03
Add `confirmed_service_name` column to services table

klssmith:
This adds a new boolean column, `confirmed_service_name`, to the `services` and `services_history` tables with a default of `False`.

We currently have a page users see when requesting to go live which talks about the service name and free allowance and populates the `is_unique` column. We want to split this into two separate pages - one that asks the user to confirm the service name is understandable, and one which asks them to confirm they meet the terms of the free allowance.

We will use the existing `is_unique` column to store whether they have confirmed they meet the terms of the free allowance, and this new column will be used to store whether they have confirmed the service name is clear.

**After deployment**
- [x] Clear services cache

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -607,6 +607,7 @@ class Service(db.Model, Versioned):
     go_live_user = db.relationship("User", foreign_keys=[go_live_user_id])
     go_live_at = db.Column(db.DateTime, nullable=True)
     has_active_go_live_request = db.Column(db.Boolean, default=False, nullable=False)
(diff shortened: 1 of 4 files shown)


klssmith:
This fails the (suggested) check of not having a migration and model change in the same file, which is not a problem in this case. It's adding a new column, adding it to the database first (since the migration runs first) then adding it to the code (when the apps get deployed after). So there won't be a case where the code is referencing something that doesn't exist.

joybytes (approved):

PR #4959, opened 2026-09-07
migration: Add document_download_filepath column to notifications table

rparke:
As part of the work to block files that have sent in jobs by document download, we need a way of storing the one or many files that may be sent using the document download via UI feature (we may also want to store this for documents uploaded via the api, but this is a less pressing concern - but we may just want to have it there for consistency).

Storing as a JSONB type because a template with files can have 0, 1 or many files, and our past experience has shown that we need the ability to selectively get the object names for specific files

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1585,6 +1585,8 @@ class Notification(db.Model):
 
     unsubscribe_link = db.Column(db.String, nullable=True)
 
+    document_download_filepath = db.Column(JSONB(none_as_null=True), nullable=True)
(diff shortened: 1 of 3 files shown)


risicle:
Would it not be better to have a separate table?

rparke:
> Would it not be better to have a separate table?

Yes

rparke:
Make this a separate table - it's naturally relational data that naturally fits in a table, let's put it in one.

PR #4961, opened 2026-09-08
migration: create table notification_to_file_id

rparke:
to be used to allow lookups of the unique file_id (and so object name in s3) of files sent via document download. We currently don't have any way of associating a file id (and so its object name in s3) with a notification without trying to parse the personalisation dict, knowing the filename we're looking for.

This is bad for a few reasons:
1) it's cumbersome
2) it means that the ability to associate a notification_id with a specific file lasts only as long as the services retention period (personalisation gets wiped).

Storing that relationship here fixes both those issues.

Be aware that the notification_id is not and cannot be a foreign key here, it has to point both to notifications and notifications_history and an explicit foreign key constraint would either require us to prune this table (that doesn't contain PII) alongside the notifications table for notifications older than the retention period, or it would just outright block the celery beat deletion task from being able to clean up notifications.

[Trello Ticket](https://trello.com/c/8M2hqh5P/723-blocking-files-sent-via-the-ui-in-error)


app/models.py
```diff
@@ -2889,3 +2889,11 @@ def serialize(self) -> SerializedReportRequest:
             created_at=self.created_at.strftime(DATETIME_FORMAT),
             updated_at=get_dt_string_or_none(self.updated_at),
         )
+
+
+class NotificationToFileId(db.Model):
+    __tablename__ = "notification_to_file_id"
+    file_id = db.Column(UUID(as_uuid=True), primary_key=True)
+    filename = db.Column(db.String(255), unique=True, nullable=False)
```
risicle: What is this column?

rparke: name of the placeholder/filename of the file the user has uploaded.

A single notification id can map to 0, 1 or many files. We might want to just revoke access to a single file in a notification that has sent more than one.

risicle: I'm confused about why it's `unique=True`

rparke: it shouldn't be

migrations/versions/0564_notification_to_file_table.py
```diff
@@ -0,0 +1,30 @@
...
+        sa.Column("filename", sa.String(length=255), nullable=False),
+        sa.Column("notification_id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("service_id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.PrimaryKeyConstraint("file_id"),
+        sa.ForeignKeyConstraint(
+                    ["service_id"],
+                    ["services.id"],
+                ),
+    )
+    op.create_index(op.f("ix_notiifcation_to_file_id_service_id"), "notification_to_file_id", ["service_id"], unique=False)
```
risicle: `notiifcation` is misspelt

risicle: I think the weird name of this index suggests this ain't a great name for the table.

Really the table is the only place we register our document download files.. could it be named something more related "sent files"? I'd call it "documents" obviously but I gather that naming doesn't go very deep into the code :(

rparke: done

app/models.py
```diff
@@ -2889,3 +2889,11 @@ def serialize(self) -> SerializedReportRequest:
...
             updated_at=get_dt_string_or_none(self.updated_at),
         )
+
+
+class SentFiles(db.Model):
+    __tablename__ = "sent_files"
+    file_id = db.Column(UUID(as_uuid=True), primary_key=True)
+    filename = db.Column(db.String(255), nullable=False)
+    service_id = db.Column(UUID(as_uuid=True), db.ForeignKey("services.id"), index=True, nullable=False)
+    notification_id = db.Column(UUID(as_uuid=True))  # we can't have a strict foreign key relationship here
```
risicle: 🤔 We'll probably want an index on this won't we

risicle:
Other than that this is probably looking about right.

app/models.py
rparke: on the notification id?

app/models.py
```diff
@@ -2889,3 +2889,13 @@ def serialize(self) -> SerializedReportRequest:
             created_at=self.created_at.strftime(DATETIME_FORMAT),
             updated_at=get_dt_string_or_none(self.updated_at),
         )
+
+
+class SentFiles(db.Model):
```
BlessedDev: Should this be added in a subsequent PR? 
Only the migration script should be in this PR.

migrations/versions/0566_notification_to_file_table.py
```diff
@@ -0,0 +1,32 @@
...
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = "0566_notification_to_file_table"
+down_revision = "0565_add_api_key_usage_table"
+
+
+def upgrade():
+    op.create_table("sent_files",
+        sa.Column("file_id", postgresql.UUID(as_uuid=True), nullable=False),
```
ipdigital24: convention seems to be id column is named "id" in primary table and prefixed in referending table

app/models.py
rparke: Squawk thinks so. I don't agree.

The model itself isn't code that runs until it is called, but it's the only way we want the api to query the new table so I normally bundle the two up in the same PR + commit (as a practical point it makes it much easier to review because you can check the diff and check that the migration and the models types and attributes all match up).

If you're ever in a situation where you want to revert a change, you want to undo both changes, because you want the table gone AND you want the thing that queries the table gone.

migrations/versions/0566_notification_to_file_table.py
CrystalPea: is it? in our other tables, like user_to_service, we name both ids - I think it's clearer when we name them both

rparke: to be honest we should really be calling it `document_id` because [that's what document_download_api calls it](https://github.com/alphagov/document-download-api/blob/main/app/download/views.py#L67-L69).

`file_id` (or `document_id` - I can be persuaded to rename it to that) shouldn't ever be a foreign key in another table (this will do joins on the notification_id to the notifications or notification_history table).

This attribute maps to the object name in S3 given by document download so I wanted to be consistent.

ipdigital24: > is it? in our other tables, like user_to_service, we name both ids - I think it's clearer when we name them both

in that case, should this table be named notification_to_document or notification_to_file instead?

rparke: Exactly @CrystalPea it's similar to the user_to_service table in that way (except we aren't mapping to a table, it's to an S3 object).

We _could_ make this a composite key as we do in that table, but it's not really necessary since document_id will always be unique.

risicle: That leads to really confusing e.g. index names.

app/models.py
risicle: 🤷 

I think the situation that caused us to make this rule was when we were _removing_ something, we discovered that what will cause queries to be fired at a table is subtler than you think. It was to do with reverse-relationships.

rparke: Ah that's fair I didn't realise. I think I'd assumed we do it in case other application code is bundled in with the PR (dao methods new endpoints etc)

migrations/versions/0566_notification_to_file_table.py
CrystalPea: Hmm... actually I thought about it more - and I agree with Ishwar - this table it's not just a simple mapping table with two columns.

It is a table where the main object is a sent file, that has associated objects, like notification and service, and some properties of its own, like filename.

In this case naming the object id simply `id` does make sense, and might be the best solution.

risicle: Everywhere else we have an `id` field (any exceptions?), it is a pseudorandomly generated uuid field where the table is the "authority" on whether it exists or not. This is not that. It is a shadow of another data source somewhere else, and so it matters what it's referring _to_.

CrystalPea (approved):

risicle:
~Another possibility re performance worries is we could introduce this for a less heavily called endpoint first so it will have less impact.~ wrong PR

risicle:
~(also I noticed the new table could benefit from some extended statistics, see #4975)~ wrong PR

CrystalPea (approved):

PR #4962, opened 2026-09-08
Add 0565_add_api_key_usage_table

BlessedDev:
This adds a new api_key_usage table to log a service's API keys usage.
 It has a composite primary key derived from its columns: service_id, api_key_id, endpoint and usage_timestamp.

The details can be found in this card https://trello.com/c/nV0j5Mst/751-create-hourly-log-table-for-api-keys


migrations/versions/0564_add_api_key_usage_table.py
```diff
@@ -0,0 +1,28 @@
...
+
+revision = '0564_add_api_key_usage_table'
+down_revision = '0563_confirmed_service_name_col'
+
+
+def upgrade():
+    op.create_table('api_key_usage',
+    sa.Column('service_id', sa.UUID(), nullable=False),
+    sa.Column('api_key_id', sa.UUID(), nullable=False),
+    sa.Column('usage_timestamp', sa.DateTime(timezone=True), nullable=False),
```
joybytes, line 16: This column is meant to be the hour bucket(as I understand), not the exact time of the request.

Could we name it to make that obvious (`hourly_usage` as in the story, or `usage_hour`)? And store it as a normal DateTime with no timezone, always truncated to the hour in UTC? I think this is the pattern we usually do

If the Redis flush writes the real request time, we will get one row per request again, which is what Redis is meant to avoid.

But to be completely honest I don't really understand too much of Redis so please ignore if that doesn't make sense

migrations/versions/0564_add_api_key_usage_table.py
```diff
@@ -0,0 +1,28 @@
...
+    sa.Column('endpoint', sa.String(), nullable=False),
```
joybytes, line 17: I think we should make this strict since we already know the endpoint routes we should be expecting
Perhaps a table like `notification_status_types` would be good. What do you think?

migrations/versions/0564_add_api_key_usage_table.py
```diff
@@ -0,0 +1,28 @@
...
+    sa.ForeignKeyConstraint(['api_key_id'], ['api_keys.id'], ),
```
joybytes, line 18: Do we need to handle the delete for this table too?

delete_service_and_all_associated_db_objects (and the user delete that removes API keys) doesn’t clear api_key_usage first. Once Redis is writing rows, deleting a service or key will fail because these foreign keys will still be pointing at them.

Should we add this table to those delete functions, or use ON DELETE CASCADE?

joybytes:
Great work 🎉 
I left a couple of comments to understand it better (I am a little out of the loop), but it all looks good

migrations/versions/0564_add_api_key_usage_table.py
BlessedDev: The column records only one request each hour and that data will be retrieved from the service' s API key usage hourly bucket in redis.
I considered usage_hour, but left it as usage_timestamp to give flexibility.

The hourly buckets in redis which would be redis sets would contain the time of the first request each hour for a combination of `service_id`, `api_key_id` and `endpoint`. This is what would be flushed to the database after the hour has elapsed, so there won't be duplication.

sqauwk rules mandate the use of `timestampz` https://squawkhq.com/docs/prefer-timestamptz, so it has to be `sa.DateTime(timezone=True)`

migrations/versions/0564_add_api_key_usage_table.py
BlessedDev: I don't think that is necessary at this stage. A determination of how the endpoints would be recorded ie route name or view function name. That will be determined during the kick off for recording the data at the endpoints, so I think it will be good to keep as much flexibility at this stage.

migrations/versions/0564_add_api_key_usage_table.py
BlessedDev: That is a good catch. I will add `ON DELETE CASCADE`.

BlessedDev: Thinking about this again, I will follow the normal pattern and the table to the delete functions in a separate PR.

migrations/versions/0564_add_api_key_usage_table.py
BlessedDev: I will also update the name to `usage_hour`.

BlessedDev:
I am converting this back to draft for now. There was a mistake in implementation, `redis` will not be used to flush data to the database.
I am updating all the cards accordingly.

migrations/versions/0564_add_api_key_usage_table.py
risicle: You might be able to go one better and add a `CheckConstraint("date_trunc('hour', usage_hour) = usage_hour")` to the column to enforce values will always be on the hour.

migrations/versions/0564_add_api_key_usage_table.py
joybytes: yeah that makes sense

migrations/versions/0565_add_api_key_usage_table.py
```diff
@@ -0,0 +1,31 @@
...
+
+def upgrade():
+    op.create_table('api_key_usage',
+    sa.Column('service_id', sa.UUID(), nullable=False),
+    sa.Column('api_key_id', sa.UUID(), nullable=False),
+    sa.Column('usage_hour', sa.DateTime(timezone=True), nullable=False),
+    sa.Column('endpoint', sa.String(), nullable=False),
+    sa.ForeignKeyConstraint(['api_key_id'], ['api_keys.id'], ),
+    sa.ForeignKeyConstraint(['service_id'], ['services.id'], ),
+    sa.PrimaryKeyConstraint('service_id', 'api_key_id',  'endpoint', 'usage_hour'),
```
risicle: Suggest ordering `service_id`, `api_key_id`, `usage_hour`, `endpoint`. And then you don't need the other index.

migrations/versions/0565_add_api_key_usage_table.py
```diff
@@ -0,0 +1,27 @@
...
+revision = '0565_add_api_key_usage_table'
+down_revision = '0564_block_ofcom_protected'
+
+
+def upgrade():
+    op.create_table('api_key_usage',
+    sa.Column('service_id', sa.UUID(), primary_key=True, nullable=False),
+    sa.Column('api_key_id', sa.UUID(), primary_key=True, nullable=False),
+    sa.Column('usage_hour', sa.DateTime(timezone=True), primary_key=True, nullable=False),
+    sa.Column('endpoint', sa.String(), primary_key=True, nullable=False),
```
risicle: It does actually respect the order the columns are added as the primary key order does it?

migrations/versions/0565_add_api_key_usage_table.py
```diff
@@ -0,0 +1,27 @@
...
+
+    sa.ForeignKeyConstraint(['api_key_id'], ['api_keys.id'], ),
+    sa.ForeignKeyConstraint(['service_id'], ['services.id'], ),
+    sa.CheckConstraint("date_trunc('hour', usage_hour) = usage_hour",
+                                       name="ck_api_key_usage_usage_hour")
```
risicle: Assume you've checked this does actually do what we think it does? I didn't try it..

BlessedDev: Yes, the constraint works.

migrations/versions/0565_add_api_key_usage_table.py
BlessedDev: It should. I took out the explicit declaration of the primary key composition because of ordering the columns to also aid B-Tree indexing.
However I have added it back in to be on the safe side.

risicle (approved):

PR #4970, opened 2026-09-17
Enable api key usage logging

BlessedDev:
This PR adds an `ApiKeyUsage` model and a `record_service_api_key_hourly_usage_at_endpoint` function.
This is is to facilitate recording API Key usage at endpoints as directed in this [card](https://trello.com/c/eiKDOLcB/764-cache-api-key-usage).
`record_service_api_key_hourly_usage_at_endpoint` records the first usage of an API Key at an endpoint every hour.
By caching the first entry, subsequent uses of the API key at that endpoint , that hour are not written to the database.
For example if the first usage of the API key is at 10:05:45am, 10:00:00am is recorded and no other usage from then on till 10:59:59am is recorded. A cache value of "1" is assigned as the purpose is just to mark the presence of a record in the database for the hour in question and prevent any other records being added that hour.
It is not practical to record every time an API key is used, so hourly usage is the preferred option. We just want to know if the API key has been used that hour.
In this PR 3 endpoints have been updated to use `record_service_api_key_hourly_usage_at_endpoint` as a test run.
Other endpoints will be updated in a separate PR. That PR will also determine how `record_service_api_key_hourly_usage_at_endpoint`  will be applied across all the endpoints.

I have tested this on `dev-c` but didn't run a load test, another consideration is wether one should be run at this stage or when all the endpoints are updated and implementation finalised.
given that most API key usage is for sending notifications, it seemed practical to use the `v2_notifications.post_notification` for the test run.


app/v2/api_key_usage.py
```diff
@@ -0,0 +1,36 @@
...
+    _record_service_api_key_hourly_usage(api_key_id, service_id, hourly_bucket, endpoint, usage_hour)
+
+
+@redis_cache.set("api-key-usage-marker:{service_id}:{api_key_id}:{hourly_bucket}:{endpoint}", ttl_in_seconds=10800)
+def _record_service_api_key_hourly_usage(api_key_id, service_id, hourly_bucket, endpoint, usage_hour):
+    # create record in the database
+    create_api_key_hourly_usage_record_dao(
+        service_id=service_id, api_key_id=api_key_id, endpoint=endpoint, usage_hour=usage_hour
+    )
+    return RequestCache.CacheResultWrapper(value=1, cache_decision=True)
```
risicle, line 36: I don't think you _actually_ have to use `CacheResultWrapper` here if the `cache_decision` is always true. Just returning the raw value should behave identically.

risicle:
This looks great. Wondering how hard we need to performance-test this before deploying.

rparke:
> This looks great. Wondering how hard we need to performance-test this before deploying.

How realistic do we need to make this? There shouldn't be a huge difference between a single key hitting one endpoint 1,000,000 times vs 10 keys hitting a single endpoint 100,000 times each etc etc.

And I'd imagine that a few high volume services POSTing to `/v2/notifications` is the real high volume case we want to test works (although some services do some weirdly inefficient stuff when it comes to GETing redundant information from other endpoints when they send so ... maybe we want to check that too).

risicle:
That is all true as long as all our assumptions are correct.

risicle:
As far as "how do we name our endpoints" goes, using the flask `endpoint` is very systematic. The only thing that concerns me is whether we'll all remember that our endpoint names have a need to be somewhat stable now.

risicle:
That said, all the v2 endpoints are very sensibly named.

app/v2/api_key_usage.py
BlessedDev: Good point. I have tested it out and will fix it.

BlessedDev:
> As far as "how do we name our endpoints" goes, using the flask `endpoint` is very systematic. The only thing that concerns me is whether we'll all remember that our endpoint names have a need to be somewhat stable now.

The hope is that this will help encourage a recognisable and stable  pattern in naming endpoints. Also it seemed the best of the available options and a way to make it consistent.

BlessedDev:
> That is all true as long as all our assumptions are correct.

I have been considering this as well. The `POST /notifications` endpoints are the key ones to test.
I will look into running some load tests next week to see if there are perceptible performance hits and take it from there.

risicle:
Another possibility re performance worries is we could introduce this first for a less heavily called endpoint first so it will have less impact.

rparke:
> Another possibility re performance worries is we could introduce this first for a less heavily called endpoint first so it will have less impact.

oh that's a good idea. Maybe `v2_template.get_template_by_id` - I don't think that's super frequently used but can check the logs to confirm.

BlessedDev:
> Another possibility re performance worries is we could introduce this first for a less heavily called endpoint first so it will have less impact.

A good candidate could be the one of the "Get message data" endpoints, probably precompiled letter status. It has to have just enough traffic to be a meaningful test case.

app/v2/api_key_usage.py
```diff
@@ -0,0 +1,36 @@
...
+    that entry. The existence of the marker prevents recording any further uses of the api key at that
+    endpoint for that hour ie from 10:00:00 to 10:59:59.
+    """
+    service_id = authenticated_service.id
+    api_key_id = api_user.id
+    usage_hour = datetime.now(UTC).replace(minute=0, second=0, microsecond=0)
+    endpoint = endpoint
+    # hourly_bucket is of the format "%Y%m%d%T%H", ie the hourly_bucket for 10:00:00 to 10:59:59 UTC on 08/09/2026
+    # would be 20260908T10
+    hourly_bucket = usage_hour.strftime("%Y%m%d%H")
```
quis: Could we make this a bit more readable?
```suggestion
    hourly_bucket = usage_hour.strftime("%Y-%m-%d%H")
```

risicle:
Could even do `get_notifications`

app/v2/api_key_usage.py
BlessedDev: Yes. I will update it.

BlessedDev:
> Could even do `get_notifications`

The question seems to be what level of traffic would be meaningful: tens, hundreds or thousands requests a day?

risicle:
What's the second-most-heavily-hit endpoint?

BlessedDev:
> What's the second-most-heavily-hit endpoint?

`get_notification_by_id` seems to see a lot of traffic apart from the `POST notifications/<notification-type>` endpoints.
I am also looking at  `post_template_preview`, almost 185K hits in the past 24 hours.

BlessedDev:
`get_pdf_for_notification` registered about 8K  in the past 24hours. Looks like a good candidate

BlessedDev:
I am considering both `get_pdf_for_notification` and `post_template_preview`. They see enough traffic (8K, 185K in the past 24 hours) to be useful tests but much less than the big ones : `post_notifications`, `get_templates` etc and also less "important".

risicle:
Do you mind if I rebase this and push my commit that adds the model-and-test counterpart of #4975?

BlessedDev:
> Do you mind if I rebase this and push my commit that adds the model-and-test counterpart of #4975?

No. It would be very useful to see if it adds performance gains.

risicle:
I wouldn't get your hopes up, I can only really see it helping once the table becomes massive and we start querying the table.

BlessedDev:
> I wouldn't get your hopes up, I can only really see it helping once the table becomes massive and we start querying the table.

Fair point.

PR #4975, opened 2026-09-18
`api_key_usage` table: add extended statistics

risicle:
The `service_id` and `api_key_id` fields should have very high dependency (one `api_key_id` value will only ever appear with one `service_id` value.

Declaring this dependency as an extended statistic will allow the query planner to make better result estimates and therefore better plans.

This will be followed up by another PR that adds this on the model side as an `__extended_statistics__` attribute.


BlessedDev (approved):
This makes sense.

